PR #5, opened 2015-10-28
Windows & Visual Studio build scripts

Stinkfist0:
To-dos & known missing features:
- CMakeLists.txt for IfcMax - after this the manually maintained solution files could be deleted. Should we also move CMakeLists.txt to the project root which seems to be typical for CMake-based projects?

Notes & Stuff for the future:
- it would appear that CHECK_ADD_OCE_OCC_DEF() macro stuff could be removed altogether?
- fix annoying LNK4049 warning spam coming from OpenCascade
- Allow building ICU from the sources if user has Cygwin (which allows the next item)
- Allow choosing whether to build the deps and IFCOS against static or dynamic MSVC run-time
- How to enable whole program optimization & link-time code generation (ENABLE_BUILD_OPTIMIZATIONS ) for static libraries (IfcParse, IfcGeom) from CMakeLists.txt?

Summary:
- Windows & Visual Studio build scripts for very effortless IfcOpenShell development environment setup. Greatly inspired the realXtend Tundra Windows & Visual Studio build scripts I've done work with. Ideally the project would probably have a single shared build script for all platforms (maybe written in Python?). 
- Debug and release builds of the dependencies can co-exist, allowing effortless switching between debug and release IfcOpenShell builds
- ICU downloaded as prebuilt package (didn't really want to add Cygwin dependency that would be required for building ICU on Windows) that is built using static MSVC run-time.
  **This means all other dependencies and IfcOpenShell itself is built accordingly too**.
- OCE and oce-win-bundle used from Tridify's forks (for now) in order to be able to build against static MSVC run-time
- Added Boost.Regex and Thread (which require System and DateTime subsequently) to the required Boost libraries
- Tested during the last few months on Windows 8.1 & 10 and VS 2013 & 2015 (both 32-bit and 64-bit builds), Python 2 & 3.

* new CMake build options:
    * UNICODE_SUPPORT & COLLADA_SUPPORT for explicit control of whether the aforementioned features are wanted
    * ENABLE_BUILD_OPTIMIZATIONS for enforcing more aggressive optimizations (MSVC-only for now)
    * USE_IFC4 (defines USE_IFC4, otherwise USE_IFC2x3 is defined)
    * BUILD_IFCPYTHON & BUILD_EXAMPLES for more fine-grained control what the user wants in the project
    * fail fast instead of elusive warning prints if optional feature (Unicode, COLLADA, Python) requested and the required dependency is not found

screenshot 2015-11-19 18 10 23
screenshot 2015-11-19 18 13 40


aothms:
Hi,

Many thanks again for your amazing contribution. Shall I wait for the additions you mentioned or just merge it in?

My main comment would be that I am not a big fan of the binaries in the utils/ folder. But I don't think such considerations have to play a big role now. 

I think I will replace them with small python scripts some time after merging. What do you think of this idea? It should be fairly straightforward to emulate wget and 7z (for PKzip) in small python scripts, colored printing would be less straightforward unless a module like colorama is installed.

Kind regards,
Thomas

Stinkfist0:
I would like to get the effortless Debug/Release switching in and after that this should be good for merging. Can you take care of the CMakeLists.txt for IfcMax (which should be trivial probably) as I don't have 3ds Max available? After this I think the old manually solution files should/could be removed for good.

Ah yes, small Python scripts could be used to replace 7zip and  wget, if wanting to keep the script filly automatized, how would we automatically install Python? PowerShell could be one option too. For colored printing I'll see if I can find a non-executable replacement. If not, the whole coloring biz can be naturally omitted.

aothms:
Sure, I will take care of IfcMax. I think to install Python manually as a prerequisite isn't a big problem, after all the compiler needs to be there as well.

Stinkfist0:
Pending small cleanup and after that I think this should be good for a merge!

aothms:
Wonderful :)

Stinkfist0:
Should be good to go, but I'll replace the wget, 7zip and cecho binaries with PowerShell snippets (build script itself doesn't need to be run in PowerShell). PowerShell comes preinstalled beginning from Windows 7, so it shouldn't be unreasonable prerequisite for the scripts IMO.

Stinkfist0:
OK, I think I can get rid of wget and cecho binaries quite easily by utilizing PowerShell, but 7-zip is a bit more tricky (ICU download is unfortunately 7-zip file) as it would require all kinds of extension at it could not be guaranteed to run OK on older PowerShell versions.

aothms:
This is really amazing work, thanks again. I have a couple of small suggestions / questions and run into an error when running the script.

- Somehow, the CTRL+C doesn't work for me on the pause, it just starts downloading boost
- The link for MSVC in the build-deps is incorrect
- Can we check for the tools to be installed? I have written following, but I wouldn't call myself a .bat expert, maybe you have a better idea, otherwise just paste it in, there also seems to exist a where.exe, but not sure how reliable:
~~~
    set COMMAND_NAMES[0]=PowerShell
    set COMMAND_NAMES[1]=git
    set COMMAND_NAMES[2]=CMake

    set COMMAND_CHECKS[0]=powershell /?
    set COMMAND_CHECKS[1]=git --version
    set COMMAND_CHECKS[2]=cmake --version

    for /L %%i in (0,1,2) do (
        call %%COMMAND_CHECKS[%%i]%% 1> NUL 2> NUL || call echo Required component '%%COMMAND_NAMES[%%i]%%' not installed or not added to PATH && exit /B 1
    )
~~~

- Can the usage info from lines 112 be shown also if there is missing tools / env variables?
- Compiling the dependencies on a fresh system I get the following error when building OpenCollada. Is there something still missing to setup libxml2?
~~~
    "deps\OpenCOLLADA\build-vs2015-x64\OPENCOLLADA.sln" (Build target) (1) ->
    "deps\OpenCOLLADA\build-vs2015-x64\ALL_BUILD.vcxproj.metaproj" (default target) (2) ->
    "deps\OpenCOLLADA\build-vs2015-x64\GeneratedSaxParser\GeneratedSaxParser_static.vcxproj.metaproj" (default target) (3) ->
    "deps\OpenCOLLADA\build-vs2015-x64\GeneratedSaxParser\GeneratedSaxParser_static.vcxproj" (default target) (20) -> (ClCompile target) ->
     deps\OpenCOLLADA\GeneratedSaxParser\src\GeneratedSaxParserLibxmlSaxParser.cpp(207): error C3861: 'xmlSAX2GetLineNumber': identifier not found [deps\OpenCOLLADA\build-vs2015-x64\GeneratedSaxParser\GeneratedSaxParser_static.vcxproj]
     deps\OpenCOLLADA\GeneratedSaxParser\src\GeneratedSaxParserLibxmlSaxParser.cpp(212): error C3861: 'xmlSAX2GetColumnNumber': identifier not found [deps\OpenCOLLADA\build-vs2015-x64\GeneratedSaxParser\GeneratedSaxParser_static.vcxproj] 
~~~

Anyway this is really great. I have something similar for Linux/OSX and will be polishing it conform to your structure. Also, this is really a big step forwards to automated daily builds, which will be a big help for the community.

Stinkfist0:
Thanks, I'll look into those notes/issues. For checking if PS/git/cmake is available, `where.exe` might be the most convenient way.

Stinkfist0:
Usage will be printed upon error situations:
screenshot 2015-11-19 17 57 47
screenshot 2015-11-19 17 56 37

Stinkfist0:
OK, all the issues/notes with the exception of OpenCOLLADA + libxml2 are now handled.

Stinkfist0:
Would you like to me to create a clean PR so that we could get rid of the unnecessary wget and cecho binaries in the history?

Stinkfist0:
OK, I think I found the reason why I don't have the LibXml2 problems you're having:
```
-- Could NOT find LibXml2 (missing:  LIBXML2_LIBRARIES LIBXML2_INCLUDE_DIR)
WARNING: Native LibXml2 not found, taking LibXml from ./Externals
```

aothms:
Not necessary to create a new PR / rewrite history. If the repo becomes bloated due to history we can git-filter-branch later. For example, the IFC models will probably also be separated into a different repo some time.

I'll check the build log tomorrow, it's not my regular dev machine, so it's not immediately obvious where it would find a libxml somewhere, but never know for sure.

aothms:
Yep, you were right

~~~
-- Found LibXml2: D:/downloads/gnuwin32/GetGnuWin32/gnuwin32/lib/libxml2.lib (found version "2.4.12") 
-- SUCCESSFUL: LibXml2 found
-- Found PCRE: D:/downloads/gnuwin32/GetGnuWin32/gnuwin32/include  
-- SUCCESSFUL: PCRE found
~~~

I don't understand how cmake is able to find this. Neither in FindLibXml2.cmake nor  OpenCOLLADA/CMakeLists.txt I see easy ways to force cmake to stick the one that ships with OpenCOLLADA.

Any ideas?

Stinkfist0:
Maybe make the script to set LIBXML2_LIBRARIES and LIBXML2_INCLUDE_DIR to empty before CMake for OpenCOLLADA to enforce that the embedded version of LibXml2 is used?

aothms:
Yep, you're completely right. This works for me:
~~~
call :RunCMake -DCMAKE_INSTALL_PREFIX="%INSTALL_DIR%\OpenCOLLADA" -DUSE_STATIC_MSVC_RUNTIME=1 -DCMAKE_DEBUG_POSTFIX=d -DLIBXML2_LIBRARIES="" -DLIBXML2_INCLUDE_DIR="" -DPCRE_INCLUDE_DIR="" -DPCRE_LIBRARIES=""
~~~

Stinkfist0:
OK, I'll add those.

Stinkfist0:
OK, ready to go! I'll provide a new PR for possible further developments.

aothms:
This is an amazing contribution. IfcOpenShell builds flawlessly. Thanks for all your efforts.

Stinkfist0:
Hmm, the formatting on the README.md's Compiling on Windows section seems a bit off.

Stinkfist0:
BTW which Windows and VS versions you tested these with?

aothms:
I tested with VS2015 CE on Windows 7.

aothms:
Hey, made a small change to your script. I observed that `git pull` returned a non-zero exit code (at least my version on my system did) when there were no new changes. Made the script no longer halt in that case.

Also made some changes to CMake, but as far as my testing goes, that only affects *nix.

Check out the *nix buildscript btw in case you're interested.

https://github.com/aothms/IfcOpenShell/commit/18f8d7b3a3d11cf3a9e548a7161fbf2bbb884532

Stinkfist0:
Nice! BTW regarding Boost libs, `system` will be required for `thread` and `date_time` for `regex `. I'm using regex at the old branch, but not thread yet (I started toying around threading IfcConvert but had no time the get anything concrete running). So, those can be omitted now, but I'll be bringing back regex and date_time in the near future.

PR #11, opened 2015-12-04
build-ifcopenshell & install-ifcopenshell batch files

Stinkfist0:


Stinkfist0:
Though these might be handy. Don't have any more commits pending on this matter.

aothms:
Works like a charm as always. Thank you!

PR #16, opened 2015-12-30
Windows build scripts: OCE & OpenCOLLADA updates

Stinkfist0:
- use OCE & oce-win-bundle from the official repos
- work around OpenCOLLADA issue with Debug build


aothms:
Clever use of `git apply`, didn't know of its existence. Thanks for your continued effort.

PR #29, opened 2016-02-03
Windows build script improvements

Stinkfist0:
Most notably support for automatically deducing the CMake generator and allowing usage user-friendly CMake generator shorthands, e.g. vs2015-x64 instead of the lengthy "Visual Studio 14 2015 Win64"


Stinkfist0:
Hmm, just noticed that this doesn't work:
```
>run-cmake.bat "Visual Studio 14 2015 Win64" -DUSE_IFC4=1


Script configuration:
  Generator = "Visual Studio 14 2015 Win64"
  Arguments = Studio 14 2015 Win64" -DUSE_IFC4=1
```

Stinkfist0:
OK, ready to go.

Stinkfist0:
Alaso added support for Visual Leak Detector (https://vld.codeplex.com/) and started hunting some memory leaks.

Stinkfist0:
Don't have any more work pending for this for now.

aothms:
Nice, sorry for the delay. Really like the build-all, that further simplifies things a lot imho.

But what's the reason for disabling the prompt? https://github.com/Tridify/IfcOpenShell-New/blob/win_build_script_improvements/win/build-all.cmd#L34 

And can't you do it by `echo "y" | call .\build-deps.bat`? Or does that only work in my own special windows shell?

In particular also thanks for fixing some memory leaks. Do you happen to have numbers on an amount of bytes being leaked for a typical IFC file? I've heard reports that the triangulation (https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcgeom/IfcGeomRepresentation.h#L151) causes memory to leak as well. `BRepMesh::Clean()` should be called in that case. I have not actually verified this. Does your tool detect anything related to this?

Stinkfist0:
Ah yes, `echo y | ...` works just fine, don't know why I didn't try that. I'll fix that.

Regarding memory leaks, a plenty still remains, had not time to inspect them in full detail. I can maybe provide a report as a separate issue later on when the time permits.

aothms:
Thanks a lot. You're continued dedication to improving the build script is tremendously appreciated. Looking forward to your memory leak issue if you find the time.

win/build-deps.cmd
```diff
@@ -115,8 +119,13 @@ call cecho.cmd 0 14 "Warning: You will need roughly 8 GB of disk space to procee
 echo.
 
 call cecho.cmd black cyan "If you are not ready with the above, press Ctrl-C to abort!"
```
aothms: @Stinkfist0 Sometimes you use `utils\cecho.cmd`, sometimes just `cecho.cmd`. I never observed any problems with this, but now I'm running this on Windows 10 (for my build farm) and I get many complaints about paths not found. Same for `7za`. The only way I can get this to work reliably is to use `%~dp0\utils\cecho.cmd` and `%~dp0\utils\7za`. Did you test on Windows 10? Isn't it weird this actually works with all the `pushd`s and `popd`s flying around? Does it makes sense to change this to use full paths?

Stinkfist0:
Yeah, I use Windows 10 on a daily basis. The idea is that once vs-cfg.cmd is successful, the utils (cecho, 7z et al.) are in path, but in erroneous cases I've used full path for the utils as the utils are not yet set in path.

aothms:
Ok, my bad! I am hacking your script to allow MinGW as a generator. Makes sense that vs-cfg.cmd does not run completely. Didn't know it also augmented PATH. Thanks for the speedy reply!

PR #31, opened 2016-02-15
Add option to compile (external) schema by cmake

aothms:
The idea is to simplify generating IfcOpenShell code for custom developed schemas, e.g [1] [2]. The added CMake code checks for dependencies in the python executable (pyparsing) and invokes the two steps necessary to compile an express schema into cpp. Generated files are added to the IfcParse project.

[1] https://bitbucket.org/Vertexwahn/bug-reports/downloads/IFC4_P6_longform_draft5_withMinorSyntaxCorrection.exp
[2] https://raw.githubusercontent.com/DURAARK/IFCPointCloud/master/tool/schemas/IFC2X3_TC1_PC.exp


aothms:
@Stinkfist0 thoughts? It touches some of your recent changes. Maybe you have some ideas?

Stinkfist0:
Sure, I can see this might be handy for some. IMHO, the PYTHON_EXECUTABLE biz could be removed and we could assume (and instruct, if this was not instructed yet) that the correct Python is set to path prior to the CMake run.

Stinkfist0:
BTW just noticed that when I added removal of unused IFC revision source files from the project they're still copied as a part of the installation phase.

aothms:
Well, the `PYTHON_EXECUTABLE` is anyway for making sure that IfcOpenShell gets installed into the correct directory. Right now [1] just a random Python is picked from the path, which may be completely unrelated to the `PYTHON_INCLUDE_DIR` and `PYTHON_LIBRARY` provided during the build. In fact, let me make an issue for that [2].

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcwrap/CMakeLists.txt#L58
[2] https://github.com/IfcOpenShell/IfcOpenShell/issues/32

Thanks for noting the installation files issue [3]. I will fix it in this branch to prevent conflicts.

[3] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/cmake/CMakeLists.txt#L355

Stinkfist0:
In the original script I have used a "standard" (I think official Python docs instruct to set this) PYTHONPATH for the location of python.exe. However, PYTHONPATH is appended to PATH, not prepended, which might be better. I don't mind having explicit control for setting the executable (I think in some projects I have seen problems at least on OS X for picking Python 2 vs 3) as long as it's used in cross-platform manner.

aothms:
I thought the `PYTHONPATH` env var is related to python modules, but I must admit I don't know the details, especially if several different interpreters are installed on the system.

The cmake `find_library` can be a can of worms, but I am under the impression that it should take care of resolving this in a cross-platform manner. It does seem I need to find the `PythonInterp` [1] package though (I will do that). I was under the impression that `PythonLibs` already would set the `PYTHON_EXECUTABLE`. Unfortunately the docs are not very elaborate on the inner workings of `find_package(PythonInterp)` neither do I find the source [2] very illustrative. pythonOCC seems to rely on cmake as well to find python interpreter [3].

[1] https://cmake.org/cmake/help/v3.0/module/FindPythonInterp.html
[2] https://github.com/Kitware/CMake/blob/master/Modules/FindPythonInterp.cmake
[3] https://github.com/tpaviot/pythonocc-core/blob/master/CMakeLists.txt#L117

Stinkfist0:
Seems that I have mixed up PYHONPATH & PYHONHOME. The latter seems to be "official" way for telling the location of python.exe, at least on Windows, so maybe better to switch to using that.

aothms:
@Stinkfist0 I see, shall I commit a s/PYTHONPATH/PYTHONHOME/ ? Or does it need more adaptations?

Stinkfist0:
@aothms Sure, these places should be altered: https://github.com/IfcOpenShell/IfcOpenShell/search?utf8=✓&q=PYTHONPATH

aothms:
Ok, doesn't seem github's search is very reliable or my patch is over-enthusiastic. Your search has 11 results, but my patch has 17 changes.

Stinkfist0:
Note for Windows users: remember to edit existing `BuilDepsCache-xYY.txt` to contain `PYTHONHOME` instead of `PYTHONPATH`, or alternatively re-run `build-deps.cmd`

PR #34, opened 2016-02-19
[fix] CmakeLists.txt now works on ubuntu 16.04 without the need to compile oce

yilativs:
[fix] CmakeLists.txt now works on ubuntu 16.04 without the need to compile oce
instead oce and opencollada dev packages should be installed with the command below:

apt-get install git swig cmake gcc g++ libftgl-dev libtbb2 libtbb-dev libboost-all-dev libgl1-mesa-dev libfreetype6-dev liboce-modeling-dev opencollada-dev liboce-foundation-dev


aothms:
Hi,

Thanks for this. I think this is an improvement but let me elaborate a bit. The `/usr/include/opencascade/` path is the installation path of the "official" Open Cascade repository [1]. Over time I came to prefer the community edition [2] due its convenient cmake build system and more open character (occt didn't even have a public git repo at that time). It would be good if this is reflected in the IfcOpenShell/CMake file. Secondly, note that these are merely defaults. Users can override these with `cmake -DOCC_INCLUDE_DIR=/path/` or with env vars. I am happy though that the defaults are set correctly for Debian/Ubuntu.

[1] http://git.dev.opencascade.org/gitweb/?p=occt.git
[2] https://github.com/tpaviot/oce

Kind regards,
Thomas

PR #35, opened 2016-02-19
Bring IfcMax up to date

### LINKED ISSUE #13: CMakeLists.txt for IfcMax
Marking this down as an FYI/public to-do if people wonder when they don't find IfcMax in the solution after running the Windows build scripts.

Stinkfist0:
Tested with VS 2015 and 3ds Max 2016 + SDK by importing acad2010_objects.ifc. Closes #13 
Having CMakeLists.txt complete for all projects makes the old manually crafted VS solution and project files obsolete so I removed them.

screenshot 2016-02-19 21 32 56


Stinkfist0:
solutionfolders

Stinkfist0:
If I may suggest, cmake/CMakeLists.txt could be maybe moved to the project root which seems to be typical for CMake-based projects.

Stinkfist0:
Documented  GCC 4.7 as the minimum requirement due to OpenCOLLADA requirements.

aothms:
Hi, many thanks for this. Sorry for the slow reply. I do have some small points for improvements, but for me they are of low priority as I expect the intersection of 3dsmax users and people compiling their own software rather small. I created a separate issue to track them [1] and will get to it in good time. Major thing is that everything is covered from within cmake, which is a major step forward. Thanks for your efforts.

[1] https://github.com/IfcOpenShell/IfcOpenShell/issues/40

Issue #38, opened 2016-02-25
Fix build scripts to known good revisions of OpenCOLLADA and OCE


Stinkfist0:
These libraries introduce breaking changes from time to time.

berndhahnebach:
Just tried a clean build with latest master branch, which is commit 480 (https://github.com/IfcOpenShell/IfcOpenShell/commit/90f72c4b2c3981f6bcfaa1fc1614a4de764f3fa8) and it builds like a charm on windows !!!

again thanks for all the afford you did stinkfist !

Stinkfist0:
Any opinions on OCE revision to use? It would appear that 0.18 should be out in the near future so maybe the release would a good point to use?

aothms:
Maybe just checkout a tag and make this configurable with parameters?

Alternatively, wonder whether it would be feasible to automatically checkout the latest tag conforming to some naming scheme. Would be possible using bash, but not sure if windows shell scripting is sufficient.

Stinkfist0:
What kind of bash trick you have in mind? Something like `git rev-list --tags --max-count=1` could work for getting the latest tag.

aothms:
I thought more magic was needed (like `grep | sort | head -n1` ), but it
seems `rev-list` comes batteries included. Only thing to add would be to
filter out the tags made in the process of merging OCCT official releases
such as [1]. I guess that can be accomplished with `--tags=OCE-*`. Neat.

[1] https://github.com/tpaviot/oce/tree/official/6.8.0

Stinkfist0:
Will you want to handle this or should I? Oh and it seems that 0.17.1 was just recently released so 0.18 won't be out soon.

aothms:
That's odd. None of the git version I tried on Windows (apparently I had 3) have this `--max-count` option for `rev-parse`. Does yours have it? Apparently `git describe --tags --abbrev=0` would do the same. That is, find the closest tag reachable from HEAD.

Does the following makes sense to you? Capturing stdout into a variable seems to be rather awkward in a bat file.

~~~
:: GitCloneAndCheckoutRevision - Clones a Git repository and checks out the latest tag reachable from HEAD
:: Params: %1 gitUrl, %2 destDir
:: F.ex. call :GitCloneAndCheckoutRevision https://github.com/tpaviot/oce.git "%DEPENDENCY_DIR%"
:GitCloneOrPullRepositoryAndCheckoutLatestTag
[...]

PR #39, opened 2016-02-25
Use a fixed revision of OpenCOLLADA...

### LINKED ISSUE #36: IfcConvert generates DAE that crashes Collada Importers
VS 2015 x64 RelWithDebInfo build:
`IfcConvert.exe ..\..\test\input\acad2010_objects.ifc acad2010_objects.dae`
Both ViewFBX and Unity will crash when trying to open the file. The crash seems to occur with all possible input files. OBJ works fine. OpenCOLLADAValidator tells the file to be valid.

I have been using mostly my old SourgeForge-based fork and noticed this now when testing the official repo/fork as I wanted to test the fixes in IfcMappedItem surfaces styles.

Stinkfist0:
in order to prevent introducing breaking changes. Closes #36, related to #38

### DIFF
--- a/win/build-deps.cmd
+++ b/win/build-deps.cmd
@@ -205,7 +205,8 @@ IF EXIST "%DEPS_DIR%\icu-55.1-vs%VS_VER%\". (
 :: Note OpenCOLLADA has only Release and Debug builds.
 set DEPENDENCY_NAME=OpenCOLLADA
 set DEPENDENCY_DIR=%DEPS_DIR%\OpenCOLLADA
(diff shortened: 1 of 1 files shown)


Stinkfist0:
The *nix build script seems to use pretty old revision - maybe good to update it to the same revision if possible?

aothms:
This didn't update the repo on my local machine because it was already cloned. Can you do a `git checkout` in the ELSE branch as well?

Stinkfist0:
Ah, yes this assumed nuking the old clone completely.

Stinkfist0:
Does this look OK now?

aothms:
Sorry for the slow replies these days. Thanks again!

Issue #20, opened 2016-01-11
IfcConvert enhancements


Stinkfist0:
Hi. Thought to open up a discussion of the enhancements I have made to IfcConvert in https://github.com/Tridify/IfcOpenShell-SourceForge before I start to reimplementing and cleaning up these to the new fork. Here's what I have done:
- new functionality for IfcConvert (taken from the `--help` output):

```
  --names arg                     A list of names or wildcard patterns that 
                                  should be included in or excluded from the 
                                  geometrical output, depending on whether 
                                  --exclude or --include is specified. The 
                                  names are handled case-sensitively. Cannot be
                                  placed right before input file argument.
  --no-normals                    Disables computation of normals. Saves time 
                                  and file size and is useful in instances 
                                  where you're going to recompute normals for 
                                  the exported model in other modelling 
                                  application in any case.

  --use-names                     Use entity names instead of unique IDs for 
                                  naming objects and materials upon 
                                  serialization Applicable for .obj and .dae 
                                  output.
  --use-guids                     Use entity GUIDs instead of unique IDs for 
                                  naming objects upon serialization. Overrides 
                                  possible usage of --use-names for objects but
                                  not for materials.Applicable for .obj and 
                                  .dae output.
  --center-model                  Centers the models upon serialization by the 
                                  applying the center point of the scene bounds
                                  as an offset. Applicable only for .dae output
                                  currently.
  --generate-uvs                  Generates UVs by using simple box projection.
                                  Requires normals. Applicable only for .dae 
                                  output currently.
```
- new `IFCCONVERT_DOUBLE_PRECISION` build option: typedef real_t either to double (ON) or to float (OFF) and use real_t throughout the IfcConvert codebase

In hindsight, `--no-normals` might not be that useful after all. Also while I started adding new options, I did this: https://github.com/Tridify/IfcOpenShell-SourceForge/commit/b5725e2b4703a87a03585e869e791a2cd3cfec78 Does this seem good idea to you? Also, thinking more about it, majority of the new options are not really geometry iterator settings, but serializer settings, so maybe we should add a new class or enum/field for the serializer settings that are used only in IfcConvert?

aothms:
Hi,

Thanks for raising this discussion. As you have read, people are already eager to get their hands on these improvements.

--names | good addition I think. Maybe the wildcard expansion can be added to --entity as well, so that one can type --include --type *, to work around infamous issue here [1]

[...]

Issue #41, opened 2016-03-05
IfcGeomServer instructions?


Stinkfist0:
There seems to be lacking any kind of instructions for trying out IfcGeomServer.

aothms:
Done: https://github.com/IfcOpenShell/IfcOpenShell/tree/master/src/ifcgeomserver

Issue #8, opened 2015-11-19
Error when converting ifc file using IfcConvert


klokoy:
I have tried this in both v4 and v5 with same result.

When I convert the attached ifc file It seems like some windows and opening elements are skewed a bit. Opening the file in solibri shows a different result. Do you have any suggestion on what is happening, is there anything I can do to fix this?

![screen shot 2015-11-19 at 15 25 11](https://cloud.githubusercontent.com/assets/973082/11273106/d65816a0-8ed1-11e5-96d0-267eece1683d.png)

This is a screenshot from solibri showing the ifc file.
![screen shot 2015-11-19 at 15 25 33](https://cloud.githubusercontent.com/assets/973082/11273107/d65ddc0c-8ed1-11e5-9e20-77cf6982e7c0.png)

This is the ifc file:

[wallwindow.ifc](https://files.zenhub.io/564ddc1680486a947e5e3b6b)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

aothms:
Thanks for your report. I will look into it.

Kind regards,
Thomas

klokoy:
The first ifc file I added was IFC4, attached you will find a 2x3 file with the same 
problem[wallwindow.ifc](https://files.zenhub.io/56547e7b5b75e1023f093404)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

The file works in BimServer. I was under the impression that they use the IfcConvert, is that correct?  Is there someway I can use the same code as BimServer to convert from ifc to collada?

Stinkfist0:
Testing with the 2x3 version, the model seems to have a very wild offset (-114509.8, 0, -1185980). Using a --center-model command line switch (my own addition, not yet in official IfcOpenShell/IfcConvert) the wall seems fine:
![screenshot 2015-11-24 17 43 40](https://cloud.githubusercontent.com/assets/227876/11371774/e6c8f3ce-92d2-11e5-873a-8985ff24a07a.png)

klokoy:
Thanks @Stinkfist0 .  If we center the model in revit it works as expected.  Will the `center-model` command line switch be merged to master?

Stinkfist0:
Yes, eventually (assuming aothms will accept that). In the meantime you can build a version of IfcConvert with the feature (and couple other new options) from https://github.com/Tridify/IfcOpenShell-SourceForge

aothms:
[...]

PR #42, opened 2016-03-10
IfcConvert enhancements

Stinkfist0:
Features discussed in #20. Opening up for an initial review, couple more commits pending.


src/ifcconvert/ColladaSerializer.cpp
```diff
@@ -19,48 +19,91 @@
...
-	id.reserve(s.size());
-	for (std::string::const_iterator it = s.begin(); it != s.end(); ++it) {
-		const std::string::value_type c = *it;
-		if ((c >= '0' && c <= '9') || (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c == '_') || ( c == '-')) {
-			id.push_back(c);
-		}
-	}
-	return id;
+/// @todo Very simple impl. Assumes that input vertices and normals match 1:1.
+static std::vector<real_t> box_project_uvs(const std::vector<real_t> &vertices, const std::vector<real_t> &normals)
```
aothms: Nice. Ideally this would be moved to `IfcGeom::Representation::Triangulation` and uvs added as member vector. Then it would be trivial to re-use this for obj/blender/max/etc.

src/ifcconvert/IfcConvert.cpp
```diff
@@ -373,30 +455,63 @@ int main(int argc, char** argv) {
...
-		old_progress = progress;
-	
-	} while (context_iterator.next());
+        IfcGeom::Element<real_t> *geom_object = context_iterator.get(true); // true == take ownership, we will clean up ourselves
+        geometries.push_back(geom_object);
+        const int progress = context_iterator.progress() / 2;
+        if (old_progress != progress) Logger::ProgressBar(progress);
+        old_progress = progress;
+
+        if (center_model) {
```
aothms: Of course this is a valuable feature. But considering it is only based on placement/insertionpoint and not on actual geometry bounding box. I think it is better to precompute this based on the product object placements prior to doing the geometry conversion loop. Converting them into a `gp_Trsf` beforehand is trivial, shouldn't take much time and the result is cached. This way we do not need to keep elements in memory and the streaming serializers such as OBJ can start right away. If you prefer I can do this after the merge though.

src/ifcgeom/IfcGeomRepresentation.h
```diff
@@ -176,13 +176,15 @@ namespace IfcGeom {
 							std::vector<std::pair<int,int> > edges_temp;
 
 							const TColgp_Array1OfPnt& nodes = tri->Nodes();
+                            /// @todo can these UVs be used for texture coordinates as is?
```
aothms: I think your box projection method is more convenient to end-users. These UVs probably have all different scales depending on their underlying geometric form. At least for curved vs planar surfaces.

aothms:
Nice. Just a few in-line comments.

src/ifcgeom/IfcGeomRepresentation.h
Stinkfist0: Roger.

src/ifcconvert/IfcConvert.cpp
Stinkfist0: I'll take a stab, but if I fail miserably I can leave it to you. :)

Stinkfist0:
BTW in SVG the elements are named currently like this:
default:
<g id="product-product-0b74b3fa-1a92-405e-9ac9-d59067be1d42-body">
`--use-element-names`:
<g id="product-A102">
`--use-element-guids`:
<g id="product-0BTBFw6f90Nfh9rP1dlXr2">

Was thinking that maybe the "product" prefix could be dropped?

aothms:
Well, I assume nobody really cares, but XML identifiers are not supposed to start with a number [1]. This is the reason for the `product-` prefix. I guess the same applies to collada, as most names and ids are of type `xsd:NCName`. Therefore I would propose to have some sort of prefix in collada and svg output.

[1] http://www.datypic.com/sc/xsd/t-xsd_ID.html

Stinkfist0:
I see, in that case some kind of prefix could be justified.

BTW was thinking to change the default behavior of overwriting output file always to fail instead and having a `--overwrite` switch or something. Does this sound good?

aothms:
1) I think a temporary file is a good idea regardless, upon completion this would then be moved. Moving a file is an atomic operation, so any process checking the existence of the file knows that if it exists the process has finished and the entire file can be read.

2) As for overwriting, I can see where this would be annoying. An alternative would be a prompt that you can silence with `-y`. I think this is more common at least in the linux world. Besides, you also have to consider users that simply drag a file over the executable in order to convert it. The console would disappear immediately so they cannot read what had been going on.

Stinkfist0:
Aye, prompt sounds good.

Stinkfist0:
Just a status update: I have been looking into writing UVs for OBJ which turned out to be not 100% trivial but I should able to figure out by looking some existing exporters for pointers.

aothms:
Ok, let me know where I can assist.

Stinkfist0:
OK. Could you give a few pointers how you would like the scene bounding box to be precomputed? Should this occur in IfcGeom::Iterator<T>::initialize()?

aothms:
I think that would be a good place. Question also is what products to include. If only including products that will be emitted, you have the risk of users processing types separately being unable to overlay their models geometrically. So, I'm guessing something like this would work:

				double minx = std::numeric_limits<double>::infinity();
				...
				IfcSchema::IfcProduct::list::ptr products = ifc_file->entitiesByType<IfcSchema::IfcProduct>();
				for (IfcSchema::IfcProduct::list::it it = products->begin(); it != products->end(); ++it) {
					IfcSchema::IfcProduct* product = *it;
					if (product->hasObjectPlacement) {
						// This is a bit buggy, but it's important that the trsf is newly initialized every time, or they will be concatenated (and wrongly stored in the cache!).
						gp_Trsf trsf;
						if (kernel.convert(product->ObjectPlacement(), trsf)) {
							const gp_XYZ& position = trsf.TranslationPart();
							minx = (std::min)(minx, position.X());
							...
						}
					}
				}

Stinkfist0:
Is it possible to get a reliable IfcBoundingBox for a product?

aothms:
Some models contain explicit BoundingBox representations for products. For most other representations it can potentially be straightforward to compute them, without too much computation, except for things like boolean operands where computing the boundingbox is basically equivalent to the actual geometry processing.

If the goal is mostly to workaround issues like https://github.com/IfcOpenShell/IfcOpenShell/issues/8 and maximizing floating point precision it shouldn't really matter whether to use the bounding box or placement origin. Most of these cases come from a very large translation at the IfcSite or IfcBuilding level. For example in the case of https://github.com/IfcOpenShell/IfcOpenShell/issues/8 . If this offset would be introduced within the geometries the model would be seriously flawed.

~~~
#1421= IFCCARTESIANPOINT((114502766.318516,1185944009.,0.));
#1423= IFCDIRECTION((0.995788774458495,0.0916772417912352,0.));
#1425= IFCAXIS2PLACEMENT3D(#1421,#19,#1423);
#1426= IFCLOCALPLACEMENT($,#1425);
#1427= IFCSITE('3o91zj$Gr6cQ_i1EP5O_03',#41,'0214 41 1 00',$,'',#1426,$,'14323',.ELEMENT.,(59,39,52,987060),(10,47,40,726547),62050.,'0214 41 1 00',$);
~~~

From that point of view an alternative way of alleviating this problem would be an option to ignore Site and Building-level placements. That is also quickly to implement and might even work more predictably to users.

aothms:
fyi, I am also opening a pull request https://github.com/IfcOpenShell/IfcOpenShell/pull/47. I will wait with merging as our PRs will conflict.

It contains some performance improvements that I guess are also of interest to you. Most notably in collada files, geometrical instances (IfcMappedItems) can now re-use the same geometry definition. Also, it seems that this caching is not doing us any good. Would appreciate a second eye testing this.

Stinkfist0:
Great, those improvements sound good for us. This PR should be now ready to go.

Stinkfist0:
Or, I could try squeezing in the --yes/-y switch and/or temporary output file if we're not in a hurry.

Stinkfist0:
Added usage of temporary file and file overwrite switch. Let me know what you think.

src/ifcconvert/ColladaSerializer.h
```diff
@@ -150,7 +150,9 @@ class ColladaSerializer : public GeometrySerializer
 	public:
 		ColladaExporter(const std::string& scene_name, const std::string& fn, ColladaSerializer *_serializer)
 			: filename(fn.c_str())
-			, stream(filename)
+            /// @todo We could use the double precision feature here, but it seems to bloat the output file quite a bit
```
aothms: Ow, I didn't know about this `bool doublePrecision = false` argument! Well, we should probably check here if these trailing 9s can somehow be traced back to something in the IFC file, or whether it is something produced inside IfcConvert.

I traced opencollada's double writing all the way to some formatting library:
https://github.com/KhronosGroup/OpenCOLLADA/blob/f4b31b4327b0eaaf539f57c618c39be83331aebd/common/libftoa/src/Commondtoa.cpp

9s could also be an artefact of that.

aothms: On first sight, it seems to me though we should enable this.

Stinkfist0: Yeah, looking the numbers a bit more closely, if `doublePrecision = true` is used, the values in DAE seem to be closer to what they are in IFC.

Stinkfist0:
This can be merged, if deemed OK. However, I also started looking into "include/exclude with children" functionality as an acute need emerged for that (complex model with many buildings but I only need geometry for one or two storeys).

aothms:
If you don't mind I would like to go ahead and merge. Please create a new PR when ready. Thanks for your great work.

PS: Don't forget to test the file on my feature branch #47 as I'm sure the geometry instancing will cut a great deal from your DAE file.

Stinkfist0:
OK, great. I already have a working version of `--with-children` switch (`e.g. --include --with-children --names "Level 1"`) and I'll provide a PR of this after some testing.

src/ifcgeom/IfcGeomIterator.h
```diff
@@ -330,6 +371,14 @@ namespace IfcGeom {
 									break;
 								}
 							}
+
+                            foreach(const boost::regex& r, names_to_include_or_exclude) {
+                                if (boost::regex_match((*jt)->Name(), r)) {
```
aothms: Ah, fyi, when merging my own PR I found one small oversight. `Name` is an optional attribute so its existence to be checked with `hasName()` first. This should be urgently addressed on a schema level in 0.6 so that optional attributes are of type `boost::optional<T>` (#50).

aothms:
> I'll provide a PR of this after some testing. Is the name OK or should it be more longer/descriptive?

Sounds good. I think it's descriptive enough. I'm not sure at this point what is easier to implement in terms of `--include-with-children` or `--include --with-children`. The latter would need additional precautions to make sure `--with-children` is not specified by itself. I can live with both. Perhaps we are stretching what can be reasonably accomplished with `boost::program_options` and we almost need some sort of grammar what is a valid command line.

> I'm also open for suggestion for enhancing the file overwrite feature (maybe ask for new file name if the file already exists etc.)

I am fine with the way it is, but am also open to changing it.

One last remark, if you try the feature branch mentioned above, be sure to checkout `59d2d227d9a13b352f20367a20572ed74fe0bf0d`, as after the merge it still broken as mentioned in https://github.com/IfcOpenShell/IfcOpenShell/pull/47#discussion_r57005581

Stinkfist0:
I'm quite busy with a project but I should have some time for testing on Thursday.

PR #51, opened 2016-03-23
Fix build script under os x

shsteven:
Building under OS X El Cap (10.11) fails with the follow error:

```
Undefined symbols for architecture x86_64:
  "_lzma_auto_decoder", referenced from:
      _xz_head in libxml2.a(xzlib.o)
  "_lzma_code", referenced from:
      _xz_decomp in libxml2.a(xzlib.o)
  "_lzma_end", referenced from:
      ___libxml2_xzclose in libxml2.a(xzlib.o)
  "_lzma_properties_decode", referenced from:
      _is_format_lzma in libxml2.a(xzlib.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Fixed the issue by disabling lzma when building libxml.


aothms:
Thanks! Good catch!

PR #53, opened 2016-03-24
Fix build script under os x

shsteven:
Fix boost related errors under OS X.


aothms:
Thanks again!

For reference, here is a ticket describing the issue with boost 1.55:
https://svn.boost.org/trac/boost/ticket/9610

PR #65, opened 2016-05-05
separated nested argument lists to avoid an error when compiling with…

iras:
… gcc.


aothms:
Thanks, yeah, it's unfortunate that this doesn't result in an error in MS Visual Studio. Automated builds will help spot these issues in the future.

Stinkfist0:
This should be valid C++11 syntax. Newer GCC should compile this AFAIK.

aothms:
But then you would need to compile with `-std=c++11` which we do not do (thusfar).

Stinkfist0:
Yup, just an observation.

PR #71, opened 2016-05-19
Geometry serialization

aothms:
#60 

Serialization of arbitrary Open Cascade boundary representation (for example, from STEP data) to IFC. Due to the difference in required semantics, a 1-1 automatic conversion is not possible. The python script below creates the required IFC semantics (project->site->building decomposition) and transfers the STEP shape to IFC.

STEP
step file
IFC
ifc file

STEP
step file
IFC
ifc -> obj

### Python script
~~~
import ifcopenshell
import ifcopenshell.geom

import time
import uuid
import tempfile

import OCC.gp
import OCC.BRepPrimAPI
import OCC.BRepAlgoAPI

from OCC.STEPControl import STEPControl_Reader
from OCC.IFSelect import IFSelect_RetDone, IFSelect_ItemsByEntity

# Code below inspired by:
# http://academy.ifcopenshell.org/creating-a-simple-wall-with-property-set-and-quantity-information/

O = 0., 0., 0.
X = 1., 0., 0.
Y = 0., 1., 0.
Z = 0., 0., 1.

def get_shape_from_step(fn):
    step_reader = STEPControl_Reader()
    status = step_reader.ReadFile(fn)
    if status == IFSelect_RetDone:
        step_reader.TransferRoot(1)
        if step_reader.NbShapes() >= 1:
            return step_reader.Shape(1)

# Creates an IfcAxis2Placement3D from Location, Axis and RefDirection specified as Python tuples
def create_ifcaxis2placement(ifcfile, point=O, dir1=Z, dir2=X):
    point = ifcfile.createIfcCartesianPoint(point)
    dir1 = ifcfile.createIfcDirection(dir1)
    dir2 = ifcfile.createIfcDirection(dir2)
    axis2placement = ifcfile.createIfcAxis2Placement3D(point, dir1, dir2)
    return axis2placement

# Creates an IfcLocalPlacement from Location, Axis and RefDirection, specified as Python tuples, and relative placement
def create_ifclocalplacement(ifcfile, point=O, dir1=Z, dir2=X, relative_to=None):
    axis2placement = create_ifcaxis2placement(ifcfile,point,dir1,dir2)
    ifclocalplacement2 = ifcfile.createIfcLocalPlacement(relative_to,axis2placement)
    return ifclocalplacement2
    
# Creates a base64 encoded GlobalId string
create_guid = lambda: ifcopenshell.guid.compress(uuid.uuid1().hex)

# IFC template creation

filename = "step_serialize.ifc"
timestamp = time.time()
[...]

PR #75, opened 2016-05-29
Make PythonOCC optional

yorikvanhavre:
I propose to make PythonOCC optional, while it's still time ;)


aothms:
Thanks Yorik. I thought I did something similar within `occ_utils`, but this looks good as well. Just the `print` statement wouldn't work in Python 3. Can you do something similar as in https://github.com/IfcOpenShell/IfcOpenShell/pull/71/files#diff-1fd3208fcbccdf94b95859832c2baef9R20 ? Thanks.

yorikvanhavre:
Oops indeed...  I'll close this and do it again

PR #76, opened 2016-05-29
IfcParse as DLL + minor fixes

stgatilov:
These changes make it possible to build IfcParse as shared library (DLL).
Note that the build system is left almost unchanged. In order to build shared IfcParse you have to manually change CMakeLists.txt.

Please **pay attention** that now it is necessary to define IFCPARSE_STATIC_DEFINE macro in order to build and use IfcParse as static library. Change it if you don't like it. For example, we can force shared users to define IFCPARSE_SHARED_DEFINE in order to build and use it (i.e. vice versa).

Minor other changes:
1. It is possible to build IfcParse without dependency on boost.regex by defining macro IFCPARSE_NO_REGEX.
2. Fixed three memory leaks: in IfcCharacterDecoder.cpp, IfcParse.cpp, IfcWrite.cpp.


aothms:
Thanks for this. I am fine with `IFCPARSE_STATIC_DEFINE`. This is how many of the libraries do things, for example ICU has its `U_STATIC_IMPLEMENTATION`. And for an LGPL library it is necessary to have good dynamic linking support. 

Since the contribution from https://github.com/IfcOpenShell/IfcOpenShell/commit/42188904afcc04bd8581d4560135a527598ae5a7#diff-f808c5dde690cfe904903e625de2aa48R33 there is already a CMake option to have dynamic libs. I will quickly tie them together after merging.

Stinkfist0:
We should also make this for IfcGeom and I have some other notes and issues but we can sort them out later.

aothms:
I agree, registered https://github.com/IfcOpenShell/IfcOpenShell/issues/77

Stinkfist0:
Some notes/issues about this PR:
1. Currently the Windows MSVC build uses static VC runtime which is not recommended to use if building redistributable DLLs. We should add a note/warning about this. An option for controlling whether dynamic or static VC runtime is used would be great, but that's another issue.
2. We should use a single build option BUILD_IFCOS_AS_DYNAMIC/STATIC, or similar, to control the build type of the current and possible new libraries, as is typical for many libraries. IfcParse_EXPORT could be renamed to IFCOS_API or similar.
3. IFCPARSE_STATIC_DEFINE - this does not need to leak to the whole project, but can be done in the specific library build's scope, I can provide a fix.
4. IFCPARSE_NO_REGEX - not a fan of this. Modes mean harder maintainability and IFCOS uses Boost widely and if one uses IFCOS, one must use Boost too. However, these kind of utility functions could be moved to their own library. Or if the <boost/regex.hpp> inclusion leaking to client code  is a problem, we should forward declare the class in the header and move header inclusion to cpp file.
5. __declspec and __attribute__ are compiler features, not platform features, so the check needs to be fixed e.g. when MinGW support is added.

stgatilov:
(3) Unfortunately, IFCPARSE_STATIC_DEFINE **has** to leak to all the users of IfcParse. Otherwise classes would be marked as dllimport on MSVC, leading to linking errors.
If you reverse it (like IFCPARSE_SHARED_DEFINE), then everyone who uses IfcParse as shared library would have to define the macro.

(4) Regarding IFCPARSE_NO_REGEX, only a single function in IfcParse uses boost.regex. Which is in fact not called in IfcParse at all, and is even not related to it (looks like a general helper). So it is more convenient to remove it from compilation than to use boost static libraries because of it.
Since it is not controlled by the CMake file in the repo, I hope no one would think that it is really maintained =)

Stinkfist0:
I might have used a bit poor wording. What I meant was this: https://github.com/Tridify/IfcOpenShell/commit/c13eb9e13112c465886ce297703c4e0b8521f8ba

Stinkfist0:
Hmm, I'm trying to build IfcOpenShell master (run-cmake.bat vs2015-x64 -DBUILD_SHARED_LIBS=1) and the DLL build seems to be completely broken. Did you test this?

Stinkfist0:
I will provide a PR that fixes this PR.

stgatilov:
I am very sorry for all the problems that have surfaced.

However, I do not see how my PR could have broken any builds.
I did not seriously alter the official CMake file with my PR (because I'm pretty sure I have too little knowledge to do it right). I use my own version of CMake file locally, and it works fine for my needs (that is: builds a usable IfcParse.dll).
I have just tried to build IfcOS on MSVC2013 x64 with BUILD_SHARED_LIBS enabled. Two versions: at the merge commit of this PR (https://github.com/IfcOpenShell/IfcOpenShell/commit/b5c4eb3b6f6af3aeeaa50b683638c18c3cb6c255) and just before that (https://github.com/IfcOpenShell/IfcOpenShell/commit/a298903fddd472565a0ce6d6b416ffb680745d03). The build is broken in both cases in rather similar ways: OpenCascade symbols not found by IfcGeom project. In fact, the commit just before PR also has a problem with link directories not set.
Also, I must note that commit https://github.com/IfcOpenShell/IfcOpenShell/commit/83243ec5141a13deb1d5653832bdfe69d13e0854 just after merge is not fully correct: IfcParse project is switched to using dllimport/dllexport, but it remains a static library.

Again, my apologies.

As for comment (2), I don't think it is possible to rename IfcParse_EXPORT to just IFCOS_API and use it in every other project. The export macro must be different for each project on MSVC. Because when you build IfcParse project, the export macro in its headers must turn into dllexport, and when you build IfcGeom, the same macro must turn into dllimport, while some other (e.g. IfcGeom_EXPORT) must become dllexport.

Perhaps the whole confusion here is because we think about different build models. Since I need a DLL version of IfcParse with as few dependencies as possible, I would be happy to see the model "each project in its own DLL". However, it seems to me that you are thinking of building a single DLL of the whole bunch of projects, forcing users to use the whole framework even if they need only a part of it.

Stinkfist0:
No worries, I have a fix coming very soon. When you to changes for a specific use case, you must also remember to build and test for the usual use case.

> As for comment (2), I don't think it is possible to rename IfcParse_EXPORT to just IFCOS_API...

It's very possible, each library projects simply will have its own uniques XXX_EXPORT*S*, see Tridify@c13eb9e

>However, it seems to me that you are thinking of building a single DLL of the whole bunch of projects, forcing users to use the whole framework even if they need only a part of it.

Nope, I referring exactly that we would have a single build option that controls whether or not all libraries of the projects are build as static or dynamic libraries.

Stinkfist0:
I also noticed that the INSTALL target was forgotten and it does not copy the library files fully to their locations.

Stinkfist0:
#79

PR #78, opened 2016-05-29
MSYS2 Mingw64 compilation

Fredounet:
Compiling IfcOpenShell using MSYS2  shell and mingw64 toolchain.
Some of the changes were suggested to me by some of the members here. There is one I created that 'works' but could be incorrect. This is the change to ColladaSerializer.cpp. Could it be due to the version of boost used in MSYS2 (1.60)?
For make, I encountered various issues for which work-arounds are listed here

- when compiling IfcParse.cpp:
    - Problem: `error: 'locale_t' does not name a type`
    - Solution: add the proposed code covering the MINGW64 case.

- when compiling ColladaSerializer.cpp:
    - Problem: `no matching function for call to 'COLLADASW::Scene::Scene(COLLADASW::StreamWriter*&, COLLADASW::URI)' COLLADASW::Scene scene (mSW, COLLADASW::URI ("#" + scene_id));`
    - Solution: add `COLLADASW::URI ("#" + scene_id)` as 3rd parameter of `COLLADASW::Scene scene ();` (same as 2nd parameter).  
	
-  when compiling IfcGeomServer.cpp:
    - Error: `error: no matching function for call to 'std::basic_ostream<char>::setf(const openmode&)'`
    - Solution: change `!defined(__CYGWIN__)` to `defined(__CYGWIN__)` on line 30 of IfcGeomServer.cpp (note that removing it entirely didn't work).

Note: I only tried compilation for IFC2x3. I also didn't build IfcMax nor shared libraries.

Note: when compiling IfcGeomFunctions.cpp, I got the following message. I don't know if it is specific to compilation with MSYS2 though:
   `C:/msys64/IfcOpenShell/src/ifcgeom/IfcGeomFunctions.cpp:139:2: warning:  #warning "You are linking against linking against an older version of Open CASCADE. Version 6.9.0 introduces various improvements with relation to boolean operations. You are advised to upgrade." [-Wcpp]`


aothms:
Thanks much. I will have a look and get back to you. As for the `COLLADASW::Scene::Scene()`, it seems you were unlucky enough to catch this moment in the OpenCollada repository. This was actually addressed by me some time ago: https://github.com/KhronosGroup/OpenCOLLADA/commit/42fef784b376b98ab01521998af306b937938f9e

Fredounet:
Thanks for the openCollada link. Indeed, it would solve the problem.

Fred-
_____________________________
Dr. Frédéric Bosché
Associate Professor
Heriot-Watt University | School of Energy, Geoscience, Infrastructure and Society
RAE Centre of Excellence in Sustainable Building Design
Associate Editor of the international journal of Automation in Construction<http://www.journals.elsevier.com/automation-in-construction/>.
Tel: +44 (0)131 451 4659
Web: http://web.sbe.hw.ac.uk/fbosche/ (CyberBuild Lab)

Heriot-Watt Univeristy Construction Project Management and Quantity Surveying<http://www.hw.ac.uk/schools/energy-geoscience-infrastructure-society/undergraduate/cpmqs.htm> – Ranked #1 in Scotland and #3 in UK (NSS 2015)

Stinkfist0:
Thanks for the initial work @Fredounet. Do you happen to have any kind of build script for the dependencies or did you simply build each by hand?

Fredounet:
With MSYS2, things are pretty simple. Everything is already built and only needs to be installed from the repository (not all packages are there in the default installation). The only thing I had to build is OCE. FOr this, I did the following (I could put this somewhere else on this website:

Using MSYS2, follow the "Building" procedure at https://github.com/tpaviot/oce/wiki/Build-%28MinGW-w64%29. 
You don't need to use the bundle as all those libraries are already available under MSYS2.

For what follows, please note that I compiled with mingw64, not mingw32. So, adapt as needed. So, I installed the `mingw-x86_64-toolchain` and also add to install `make` as well as a few libraries that don't get installed by default such as boost (the command is simply `pacman -S mingw-w64-x86_64_boost`; apply the same for all packages missing for the compilation such as opencollada and swig, and possibly others). 

For Cmake, use `MSYS Makefiles`, and set `CMAKE_MAKE_PROGRAM` to `mingw64/bin/mingw32-make` (but during the make stage, I had to use the command `make` instead; this confuses me... I must have done something wrong here, but it still worked...). 
You must also add `CMAKE_NEED_RESPONSE=YES` to the CMakeLists file or pass to cmake. If you correct this after the error has occurred ("linking error, argument list too long"), Compilation will restart from scratch. So, I recommend adding this upfront, and not wait for the error. I recommend using cmake-gui (installed using MSYS2 and launched from its command line)

Like always with OCE, compilation takes forever (but it's a good idea to remember to use `make -j4`, or more if you have). But it went smoothly.

Fredounet:
I could write a script for this, indeed. But, I always seem to experience slightly different issues. Anyways, I'll put this on my todo list.

Stinkfist0:
Thanks for the info.

aothms:
I've created a new PR based on this #84. With some additional changes. Shuffled around the preprocessor macros a bit. The description also contains a small script. Overall building with MSYS2 is a breeze. I'll close this one and leave the other one open for feedback for a while.

PR #79, opened 2016-05-30
Fix #76

Stinkfist0:
- a bit cleaner ("the usual") implementation for MSVC __declspec(dllimport/dllexport) related macros
- fix Windows/MSVC compilation of the whole project when BUILD_SHARED_LIBS=1
- fix installation of IfcParse DLL
- I generated new IFC4 Add1 code using IFC4_ADD1.exp from http://www.buildingsmart-tech.org/specifications/ifc-releases/ifc4-add1-release and I noticed some changes in the result's code. Can you check this? Which file did you use to generate the IFC4 Add1 code earlier?


stgatilov:
I have also noticed that generating cpp code from IFC4_ADD1.exp did not produce same results as in the repo.
Also, in the _Ifc4*.cpp_ files it is still mentioned that "IFC4.exp" was used to generate them. Perhaps change this to "IFC4_ADD1.exp" to make it clear?

Stinkfist0:
Yeah, that should be updated. It would be nice if the Python script would write the license information and notification which would include the filename automatically directly.

aothms:
Thanks a lot Ali. For bookkeeping purposes: It seems this actually addresses issues that stem from #67 as well on Windows. I will check and merge soon.

Stinkfist0:
@aothms No prob. If could pay special attention to the new generated code for IFC 4 Add 1 so that there will not be anything broken in changes related to that.

Stinkfist0:
As this is not merged yet, I could also do some renaming for the macros. I'd propose that `IfcParse_EXPORT` would be renamed to `IFC_PARSE_API` and we would introduce `IFC_PARSE_PRIVATE` (`__attribute__ ((visibility ("hidden"))` on GCC, empty on MSVC) to explicitily prevent unwanted symbols to be included in a DLL. When adding the export macros, I think I saw maybe one or two classes that could be considered private/internal to IfcParse but for now I simply used `IfcParse_EXPORT` for all classes.

aothms:
The *nix build script compiles with `-fvisibility=hidden` by default (and I think this is common practice?). So it wouldn't be necessary to add this additional private symbol IMHO.

I'm a bit worried about conflicts between all open pull requests atm (my bad I know), especially #82. Maybe save the rename for a different PR (or I can do a quick sed run myself after merging).

Stinkfist0:
Yeah, having -fvisibility=hidden by default would be even better (AFAIK it's not the de facto practice currently/yet, but could be wrong), although I would suggest putting it to the the CMakeLists.txt that it would be the default (and clearly visible) when not building via the build script.

Yeah, the renaming biz can be postponed, no prob.

cmake/CMakeLists.txt
```diff
@@ -429,7 +444,17 @@ endforeach()
 
 set(IFCPARSE_FILES ${IFCPARSE_CPP_FILES} ${IFCPARSE_H_FILES})
 
-ADD_LIBRARY(IfcParse STATIC ${IFCPARSE_FILES})
+if (BUILD_SHARED_LIBS)
+    add_library(IfcParse SHARED ${IFCPARSE_FILES})
+    set(IFCBINS "${IFCBINS};IfcParse")
```
aothms: I don't fully understand: `IfcParse` is added to `IFCBINS` now on non-MSVC platforms. With the `INSTALL` statement below this means that `IfcParse.lib` would go to `/usr/bin`, whereas I think the convention is to put it in `/usr/lib`? Somehow this also triggers a CMake error in Linux: `install TARGETS given no LIBRARY DESTINATION for shared library target "IfcParse"`, potentially because of the empty `${IFCLIBS}` at the `INSTALL` command, not sure.

Stinkfist0:
Might have made some mistake there. I can check it tomorrow.

aothms:
Thanks

Stinkfist0:
Rechecking this and this should be I how meant it to be (if BUILD_SHARED_LIBS=0, building static libs and these should go to libs and if BUILD_SHARED_LIBS=1, building DLLs and these should go to bin) so there are probably some platform-specific differences for the install command. We need to probably add LIBRARY install command in addition to the ARCHIVE and RUNTIME commands. Oh and BTW noticed that BUILD_SHARED_LIBS is actually a global CMake option that affects all add_library calls if no lib type is specified and it might be nicer not to use this as we probably want to control this behavior explicitly.

Stinkfist0:
So - maybe adding this to the end of file would solve the issue?
```
if (NOT WIN32)
    INSTALL(TARGETS ${IFCBINS} LIBRARY DESTINATION ${IFCBINS})
endif()
```

aothms:
Confusing. Even though `BUILD_SHARED_LIBS` is an existing flag, the docs mention it is often added as an option [1]. So apparently this is recommended.
[1] https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html 

But then indeed this `if` statement [2] is unnecessary, or at least the `add_library()` in it.
[2] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/cmake/CMakeLists.txt#L443

What I read from the CMake documentation is [3] that we can simply let CMake figure this out:

~~~
install(TARGETS myExe mySharedLib myStaticLib
        RUNTIME DESTINATION bin
        LIBRARY DESTINATION lib
        ARCHIVE DESTINATION lib/static)
~~~

will install myExe to <prefix>/bin and myStaticLib to <prefix>/lib/static. On non-DLL platforms mySharedLib will be installed to <prefix>/lib and /some/full/path. On DLL platforms the mySharedLib DLL will be installed to <prefix>/bin and /some/full/path and its import library will be installed to <prefix>/lib/static and /some/full/path.

As "All Windows-based systems including Cygwin are DLL platforms."

[3] https://cmake.org/cmake/help/v3.0/command/install.html 

Thoughts?

aothms:
Based on our discussion I came up with the following changes:

~~~
diff --git a/cmake/CMakeLists.txt b/cmake/CMakeLists.txt
index 37dc123..7d5331e 100644
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -60,7 +60,7 @@ IF(NOT IS_ABSOLUTE ${LIBDIR})
 ENDIF()
 MESSAGE(STATUS "LIBDIR: ${LIBDIR}")
 
-set(IFCDIRS "") # for *nix rpaths
+set(IFCOPENSHELL_LIBARY_DIR "") # for *nix rpaths
 
 if (BUILD_SHARED_LIBS)
     add_definitions(-DBUILD_SHARED_LIBS)
@@ -70,7 +70,7 @@ if (BUILD_SHARED_LIBS)
         # There will be couple hundreds of these so suppress them away, https://msdn.microsoft.com/en-us/library/esew7y1w.aspx
         add_definitions(-wd4251)
     endif()
-    set(IFCDIRS "${LIBDIR}")
+    set(IFCOPENSHELL_LIBARY_DIR "${LIBDIR}")
 endif()
 
 # Create cache entries if absent for environment variables
@@ -427,8 +427,7 @@ if(NOT WIN32)
 	LINK_DIRECTORIES(${LINK_DIRECTORIES} /usr/lib /usr/lib64 /usr/local/lib /usr/local/lib64)
 endif()
 
-SET(IFCLIBS "")
-SET(IFCBINS "")
+SET(IFCOPENSHELL_LIBRARIES IfcParse IfcGeom)
 
 # IfcParse
 file(GLOB IFCPARSE_H_FILES ../src/ifcparse/*.h)
@@ -444,16 +443,7 @@ endforeach()
 
 set(IFCPARSE_FILES ${IFCPARSE_CPP_FILES} ${IFCPARSE_H_FILES})
 
-if (BUILD_SHARED_LIBS)
-    add_library(IfcParse SHARED ${IFCPARSE_FILES})
-    set(IFCBINS "${IFCBINS};IfcParse")
-    if (MSVC)
-        set(IFCLIBS "${IFCLIBS};IfcParse") # import lib for the DLL
-    endif()
-else()
-    add_library(IfcParse STATIC ${IFCPARSE_FILES})
-    set(IFCLIBS "${IFCLIBS};IfcParse")
-endif()
+add_library(IfcParse ${IFCPARSE_FILES})
 set_target_properties(IfcParse PROPERTIES COMPILE_FLAGS -DIfcParse_EXPORTS)
 
 IF(UNICODE_SUPPORT)
@@ -465,17 +455,10 @@ file(GLOB IFCGEOM_H_FILES ../src/ifcgeom/*.h)
 file(GLOB IFCGEOM_CPP_FILES ../src/ifcgeom/*.cpp)
 set(IFCGEOM_FILES ${IFCGEOM_CPP_FILES} ${IFCGEOM_H_FILES})
 IF(BUILD_SHARED_LIBS)
-    if (MSVC)
-        message(WARNING "Building IfcGeom as DLL not currently supported on Windows/MSVC!")
+	message(WARNING "Building IfcGeom as shared library not currently supported")
         add_library(IfcGeom STATIC ${IFCGEOM_FILES})
-        set(IFCLIBS "${IFCLIBS};IfcGeom")
-    else()
-        add_library(IfcGeom SHARED ${IFCGEOM_FILES})
-        set(IFCBINS "${IFCBINS};IfcGeom")
-    endif()
 ELSE()
-    ADD_LIBRARY(IfcGeom STATIC ${IFCGEOM_FILES})
-    set(IFCLIBS "${IFCLIBS};IfcGeom")
+        add_library(IfcGeom ${IFCGEOM_FILES})
 ENDIF()
 
 TARGET_LINK_LIBRARIES(IfcGeom IfcParse)
@@ -488,7 +471,6 @@ ADD_EXECUTABLE(IfcConvert ${IFCCONVERT_FILES})
 if (IFCCONVERT_DOUBLE_PRECISION)
     set_target_properties(IfcConvert PROPERTIES COMPILE_FLAGS -DIFCCONVERT_DOUBLE_PRECISION)
 endif()
-set(IFCBINS "${IFCBINS};IfcConvert")
 
 # Make sure cross-referenced symbols between static OCC libraries get
 # resolved. Also add thread and rt libraries.
@@ -501,9 +483,9 @@ if("${libTKernelExt}" STREQUAL ".a")
     set(OPENCASCADE_LIBRARIES ${OPENCASCADE_LIBRARIES} ${OPENCASCADE_LIBRARIES} ${OPENCASCADE_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} ${LIB_RT} dl)
 endif()
 
-TARGET_LINK_LIBRARIES(IfcConvert ${IFCLIBS} ${OPENCASCADE_LIBRARIES} ${Boost_LIBRARIES} ${OPENCOLLADA_LIBRARIES} ${ICU_LIBRARIES})
+TARGET_LINK_LIBRARIES(IfcConvert ${IFCOPENSHELL_LIBRARIES} ${OPENCASCADE_LIBRARIES} ${Boost_LIBRARIES} ${OPENCOLLADA_LIBRARIES} ${ICU_LIBRARIES})
 if (NOT WIN32)
-    SET_INSTALL_RPATHS(IfcConvert "${IFCDIRS};${OCC_LIBRARY_DIR};${Boost_LIBRARY_DIRS};${OPENCOLLADA_LIBRARY_DIR};${ICU_LIBRARY_DIR}")
+    SET_INSTALL_RPATHS(IfcConvert "${IFCOPENSHELL_LIBARY_DIR};${OCC_LIBRARY_DIR};${Boost_LIBRARY_DIRS};${OPENCOLLADA_LIBRARY_DIR};${ICU_LIBRARY_DIR}")
 endif()
 
 # IfcGeomServer
@@ -511,11 +493,10 @@ file(GLOB CPP_FILES ../src/ifcgeomserver/*.cpp)
 file(GLOB H_FILES ../src/ifcgeomserver/*.h)
 set(SOURCE_FILES ${CPP_FILES} ${H_FILES})
 ADD_EXECUTABLE(IfcGeomServer ${SOURCE_FILES})
-TARGET_LINK_LIBRARIES(IfcGeomServer ${IFCLIBS} ${OPENCASCADE_LIBRARIES} ${Boost_LIBRARIES} ${ICU_LIBRARIES})
+TARGET_LINK_LIBRARIES(IfcGeomServer ${IFCOPENSHELL_LIBRARIES} ${OPENCASCADE_LIBRARIES} ${Boost_LIBRARIES} ${ICU_LIBRARIES})
 if (NOT WIN32)
-    SET_INSTALL_RPATHS(IfcGeomServer "${IFCDIRS};${OCC_LIBRARY_DIR};${Boost_LIBRARY_DIRS};${ICU_LIBRARY_DIR}")
+    SET_INSTALL_RPATHS(IfcGeomServer "${IFCOPENSHELL_LIBARY_DIR};${OCC_LIBRARY_DIR};${Boost_LIBRARY_DIRS};${ICU_LIBRARY_DIR}")
 endif()
-set(IFCBINS "${IFCBINS};IfcGeomServer")
 
 IF(BUILD_IFCPYTHON)
 	ADD_SUBDIRECTORY(../src/ifcwrap ifcwrap)
@@ -530,7 +511,16 @@ IF(BUILD_IFCMAX)
 ENDIF()
 
 # CMake installation targets
-INSTALL(FILES ${IFCPARSE_H_FILES} DESTINATION ${INCLUDEDIR}/ifcparse)
-INSTALL(FILES ${IFCGEOM_H_FILES} DESTINATION ${INCLUDEDIR}/ifcgeom)
-INSTALL(TARGETS ${IFCBINS} RUNTIME DESTINATION ${BINDIR})
-INSTALL(TARGETS ${IFCLIBS} ARCHIVE DESTINATION ${LIBDIR})
+INSTALL(FILES ${IFCPARSE_H_FILES} 
+	DESTINATION ${INCLUDEDIR}/ifcparse
+)
+
+INSTALL(FILES ${IFCGEOM_H_FILES} 
+	DESTINATION ${INCLUDEDIR}/ifcgeom
+)
+
+INSTALL(TARGETS IfcParse IfcGeom IfcConvert IfcGeomServer
+	ARCHIVE DESTINATION ${LIBDIR}
+	LIBRARY DESTINATION ${LIBDIR}
+	RUNTIME DESTINATION ${BINDIR}
+)
diff --git a/src/examples/CMakeLists.txt b/src/examples/CMakeLists.txt
index b7f96c2..6e9c53b 100644
--- a/src/examples/CMakeLists.txt
+++ b/src/examples/CMakeLists.txt
@@ -18,13 +18,9 @@
 ################################################################################
 
 ADD_EXECUTABLE(IfcParseExamples IfcParseExamples.cpp)
-IF(BUILD_SHARED_LIBS)
-    TARGET_LINK_LIBRARIES(IfcParseExamples ${IFCLIBS})
-ELSE()
-    TARGET_LINK_LIBRARIES(IfcParseExamples IfcParse)
-ENDIF()
+TARGET_LINK_LIBRARIES(IfcParseExamples IfcParse)
 set_target_properties(IfcParseExamples PROPERTIES FOLDER Examples)
 
 ADD_EXECUTABLE(IfcOpenHouse IfcOpenHouse.cpp)
-TARGET_LINK_LIBRARIES(IfcOpenHouse ${IFCLIBS} ${OPENCASCADE_LIBRARIES})
+TARGET_LINK_LIBRARIES(IfcOpenHouse ${IFCOPENSHELL_LIBRARIES} ${OPENCASCADE_LIBRARIES})
 set_target_properties(IfcOpenHouse PROPERTIES FOLDER Examples)
~~~

Does this make sense to you? Didn't test extensively on Windows.

Stinkfist0:
Yeah, looks OK. It should be fine on Windows/MSVC if BUILD_SHARED_LIBS=1 and when installing you'll see IfcParse.dll going to bin and IfcGeom.lib and IfcParse.lib going to lib and  when BUILD_SHARED_LIBS=0  IfcGeom.lib and IfcParse.lib go to lib.

aothms:
Perfect. Thanks for checking. Created #89. Closing this one.

Issue #21, opened 2016-01-12
Coding conventions?


Stinkfist0:
The coding conventions of IfcOpenShell seem rather mixed. What would be the "official" conventions to follow when I start to contribute?

aothms:
I agree it's quite horrible. Witness the result of 5 years of changes in taste and even wanting to mimick the IfcCamelCasing at some point, which was a bad idea. Also OpenCascade's peculiar conventions played a role I guess.

I intended to unify this as part of the upcoming 0.6.0 efforts [1]. But we can settle on something before that. In general I can agree to anything K&R'ish with some modernization. Only identifiers verbatim from the IFC schema would be camel cased to prevent confusion. Open Cascade names can remain as is of course, but preferably remain under the surface for most end-users so that eventually other geometry kernels can be used as well without impacting most of clients' own code.

[1] https://sourceforge.net/p/ifcopenshell/discussion/1782718/thread/b5e3ee85/#b6b2

Do you have strong preferences? I have always thought that mozilla provides [2] a reasonable and succinct guideline. Although 1) I prefer 4-width indentation 2) indenting the labels of a switch statement can be difficult to achieve in some editors 3) return types do not necessarily need go on their own line for me, but I can live with that.

[2] https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style

Stinkfist0:
Hi, sorry for taking a long time to reply. As IFCOS is your baby, pretty much everything goes as long as it's consistent. :) But here follows a few of my opinions and preferences. Also in general, in my opinion, different languages don't need different conventions, so it would be nice that the conventions would be similar in different languages (C++, Python, CMake).

- K&R style for braces sounds good. I also I like the modification that this style is applied for control flow but not for classes, enums and functions.

- I prefer spaces over tabs, but as the whole codebase uses tabs consistently (with the exception of generated code for Python which has some inconsistency), can't think of a good reason for changing this.

- For classes, structs, enums, anything "constant-like" (one doesn't see assigment calls for the types, with the exception of initialization), I like `UpperCamelCase`. How about consecutive capital letters? `XMLRPCCall` vs. `XmlRpcCall`? Personally I prefer the latter, as the first option can lead to `ALLCAPS` which is already a typical convetion for macros.

[...]

Issue #10, opened 2015-11-30
Erroneous space materials 


klokoy:
When using the IfcConverter to convert the duplex house (with only space included) some spaces are not transparent. I see the same in other models as well. In solibri I see that theese spaces are a different colour then the other (I think they use a different geometry) but they are transparent.

<img width="734" alt="screen shot 2015-11-30 at 15 51 47" src="https://cloud.githubusercontent.com/assets/973082/11474809/63dd29d8-977a-11e5-92ff-6dae9ff3aa7f.png">

aothms:
The reason for this is that these spaces do not have any colour (IfcSurfaceStyle) associated with their representations. Solibri is kind enough to give you a transparent default color, but IfcOpenShell doesn't. You can clearly identify colours that are derived from the IFC file by the name they are assigned in the OBJ or DAE file.

PR #82, opened 2016-06-02
Performance improvement of IfcParse tokenizer

stgatilov:
In my opinion, most of the performance issues in IfcParse are due to too many dynamic memory allocations.
This PR is dedicated to removing memory allocations in tokenization process, at least for numbers and operators.

As a result, opening a 47Mb file in our IfcParse-based project now takes 9.5 seconds instead of 12.5 seconds. According to MSVC profiler, samples count of IfcParse.dll has reduced from 3600 to 2600. Built with MSVC2013 x64 with additional optimizations enabled.

In the new tokenization system, most of the information for each token is determined at the moment it is created, and it is saved within the Token struct itself. In particular:
1. startPos - index of first char in token (was saved before).
2. endPos - index of the after-the-last char of token (I thought it would be useful, but it is not actually used).
3. type - operator/keyword/integer/bool/whatever.
4. value - only for some types of tokens, stored in union.
For simple tokens, their values are determined on creation without any memory allocations. This allows to avoid costly creation of std::string-s in IfcSpfLexer::TokenString function.

There are several points I would like to discuss about these changes:

(1) [According to STEP specification](http://www.steptools.com/library/standard/p21e3_dis_paris.html#clause-5-6), EOL chars may appear in any tokens. So breaking a real number into two lines is considered a correct STEP. This is rather unpleasant for tokenization, because it is not possible to parse numbers with standard libraries "in-place" (unless all EOLs are removed even before tokenization).
In order to support such weird tokens, I copy each token into a 64-byte buffer on stack, removing all EOL in process (see function RemoveTokenSeparators). This allows to avoid dynamic allocation, but limits length of any integer or real number to 63 characters. I wonder if it is really a problem and if anyone knows a way to fix it (without dynamic allocation).
For example, the following valid numbers would screw tokenization process in the PR:
- integer: 00000...[repeat 70 times]...0000078
- real: 14.1039...[write 70 digits]...23569

(2) IfcParse library lacks an assert macro.
Throwing exception is the right thing to do on ill-formatted inputs. But sometimes it is desirable to check for conditions which must be true by construction, in order to simplify debugging process. And throwing exception in a small getter function is a waste of program time and space. See functions asInt, asFloat and other for example.

### DIFF
--- a/src/ifcparse/IfcParse.cpp
+++ b/src/ifcparse/IfcParse.cpp
@@ -339,23 +339,24 @@ Token IfcSpfLexer::Next() {
 // Reads a std::string from the file at specified offset
 // Omits whitespace and comments
 //
-std::string IfcSpfLexer::TokenString(unsigned int offset) {
+void IfcSpfLexer::TokenString(unsigned int offset, std::string &buffer) {
 	const bool was_eof = stream->eof;
 	unsigned int old_offset = stream->Tell();
 	stream->Seek(offset);
-	std::string buffer;
-	buffer.reserve(128);
+	buffer.clear();
 	while ( ! stream->eof ) {
 		char c = stream->Peek();
 		if ( buffer.size() && (c == '(' || c == ')' || c == '=' || c == ',' || c == ';' || c == '/') ) break;
 		stream->Inc();
 		if ( c == ' ' || c == '\r' || c == '\n' || c == '\t' ) continue;
-		else if ( c == '\'' ) return *decoder;
+		else if ( c == '\'' ) {
+			buffer = *decoder;
+			return;
+		}
 		else buffer.push_back(c);
 	}
 	if ( was_eof ) stream->eof = true;
 	else stream->Seek(old_offset);
-	return buffer;
 }
 
 //Note: according to STEP standard, there may be newlines in tokens
@@ -503,15 +504,23 @@ double TokenFunc::asFloat(const Token& t) {
 	return t.value_double;
 }
 
+const std::string &TokenFunc::asStringRef(const Token& t) {
+	std::string &str = t.lexer->GetTempString();
+	t.lexer->TokenString(t.startPos, str);
+	if (isString(t) || isEnumeration(t) || isBinary(t)) {
+		//remove start+end characters in-place
+		str.pop_back();
+		str.erase(str.begin());
+	}
+	return str;
+}
+
 std::string TokenFunc::asString(const Token& t) {
-	if ( isOperator(t,'$') ) return "";
-	else if ( isOperator(t) ) throw IfcException("Token is not a string");
-	std::string str = t.lexer->TokenString(t.startPos);
-	return isString(t) || isEnumeration(t) || isBinary(t) ? str.substr(1,str.size()-2) : str;
+	return asStringRef(t);
 }
 
 boost::dynamic_bitset<> TokenFunc::asBinary(const Token& t) {
-	const std::string str = asString(t);
+	const std::string &str = asStringRef(t);
 	if (str.size() < 1) {
 		throw IfcException("Token is not a valid binary sequence");
 	}
@@ -541,8 +550,9 @@ boost::dynamic_bitset<> TokenFunc::asBinary(const Token& t) {
 }
 
 std::string TokenFunc::toString(const Token& t) {
-	if ( isOperator(t) ) return std::string ( (char*) &t.startPos	, 1 );
-	else return t.lexer->TokenString(t.startPos);
+	std::string result;
+	t.lexer->TokenString(t.startPos, result);
+	return result;
 }
 
 
--- a/src/ifcparse/IfcParse.cpp
+++ b/src/ifcparse/IfcParse.cpp
@@ -815,7 +815,7 @@ Entity::Entity(unsigned int i, IfcFile* f) : args(0), _id(i) {
 	file = f;
 	Token datatype = f->tokens->Next();
 	if ( ! TokenFunc::isKeyword(datatype)) throw IfcException("Unexpected token while parsing entity");
-	_type = IfcSchema::Type::FromString(TokenFunc::asString(datatype));
+	_type = IfcSchema::Type::FromString(TokenFunc::asStringRef(datatype));
 	offset = datatype.startPos;
 }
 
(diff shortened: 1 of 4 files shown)


src/ifcparse/IfcParse.cpp
```diff
@@ -70,8 +70,8 @@ void init_locale() {
 // Opens the file, gets the filesize and reads a chunk in memory
 //
 IfcSpfStream::IfcSpfStream(const std::string& fn)
-    : stream(0)
```
Stinkfist0: BTW the convention will be to use spaces instead of tabs, hence I've already started to use spaces, https://github.com/IfcOpenShell/IfcOpenShell/issues/21

Stinkfist0:
Good stuff!

stgatilov:
Thank you!
I hope someone familiar with the original code would find time to review the changes.

aothms:
Indeed many thanks for this, especially considering we all know this is not the best maintained part of the code base. I will have time for a closer look probably tomorrow. 

But indeed, 

> EOL chars may appear in any tokens

this is indeed the main reason why there are these `std::string` allocations in the parser code base now. First of all, my impression is that random new lines within tokens will break many of the parsers out there. Also, numbers of more than 63 digits I find a bit of a pathological example. But how about the other option you mention, stripping all carriage returns and newlines before tokenization, you should be able to do this in-place without too much overhead am I right? Another monstrous option of course is to provide drop in replacements for `ato_()` that skip over newline characters there, surely there must be reliable sources for this we can adapt. So to summarize:

- Allow newline to break parser
- Use buffer on stack (This PR)
- Use dynamic buffer (master)
- Pre-process to remove newlines
- Alternative `ato_()` implementations that ignore newlines
- More...?

Any thoughts? We can defer a final decision to the 0.6 release and stick to this PR for now.

Regarding the exceptions: 5 years ago many IFC files were semantically and syntactically invalid in various ways and we felt that a very permissive parser, one that would just keep churning along until some exception was thrown, was most desirable. Still at this time, many IFC files have their subtle issues, such as issues with cardinalities. I am not entirely sure how this relates to your proposal of an assertion macro. I do agree that either a strict parsing mode or an earlier indication of whether the file conforms to at least the subset of the schema you are interested in would be valuable and would also keep the code (especially the geometry interpretation which is full of try catch blocks for this reason) cleaner.

stgatilov:
Stripping all EOLs before parsing is trivial to do, but there is one problem.
It seems that there exists a streaming mode in IfcSpfStream, it can be enabled by defining a macro BUF_SIZE. And it is not possible to add input preprocessing with this feature enabled. Well, at least it is not easy to do: it is necessary to maintain correspondence between real positions in input file and virtual positions in the resulting preprocessed buffer.
Other from that, it should be simple and relatively fast. Moreover, it can perhaps simplify parsing even further, especially the TokenString function.
BTW, IfcPlusPlus strips all EOLs and comments before parsing.

I also feel that newlines inside numbers is never generated and not supported by many programs. But a standard is a standard, unfortunately. A very long integer is indeed a highly unlikely case, but a very long float seems to be possible.

As for writing custom atoi and atod equivalents, I was also considering this option. But it is quite a lot of nontrivial code to write and debug.

Another option is to create a single std::string object, and use it as a temporary location for all the tokens (and shrink only when finished). It can be trivially achieved by making a **static** local variable in IfcParse::GeneralTokenPtr function. However, this would immediately break multithreaded parsing. This temporary buffer can be placed into some object like IfcSpfStream (or IfcFile), then it would work safely under condition that no IfcSpfStream object is used by several threads simultaneously.

aothms:
Yep, good point. This `BUF_SIZE` has not been used as the intermittent seeking was very detrimental to performance. However, with this pull request and the most prominent token values actually represented in memory, it might be worthwhile to reconsider this streaming parsing.

Of course there are good examples of `atoi` and `atod`, e.g [1] we don't necessarily need to reinvent the wheel. But I too am not very much in favor of this option.
[1] ftp://ftp.irisa.fr/pub/OpenBSD/src/sys/lib/libsa/strtol.c

Multi-threaded parsing seems quite far out to me still. I am also fine with static buffer that can grow. I am sure we can we can somehow instantiate different buffers per thread later if that ever becomes imminent.

stgatilov:
I have implemented the last way: added a _tempString field to IfcSpfLexer, which is used in several cases as local storage. So numbers can have unlimited length now. It also allows to handle keywords without allocations.

P.S. I'm sorry for the EOL style: it seems that I have messed it in the first bunch of commits.

aothms:
Thanks again. I think that's a good approach.

Now, I ran IfcConvert compiled with this PR and master and compared the perfomance. It's not really conclusive or anything as it is hard to eliminate e.g hard drive caching of the file. It does seem overall the performance increases nicely. IfcConvert is not the best executable to test this PR as most of the time is spent inside Open Cascade modules. Memory usages increases a bit too, which is not unexpected. I wonder if we can reduce it a bit though. One potential option would be to store aggregates not as separate tokens, but already aggregate them into vectors at the parsing phase.

clinic_mep_20110906_optimized ifc

With this python script I isolated a bit more on the performance of the parser

~~~
import sys, time, ifcopenshell
li = []
t0 = time.clock()
f = ifcopenshell.open(sys.argv[1])
t1 = time.clock()
for pt in f.by_type("IfcCartesianPoint"):
    li.append(pt.Coordinates)
t2 = time.clock()
print t2 - t1, t1 - t0
~~~

The differences are quite significant:

~~~
|        | t2-t1 | t1-t0 | 
| pr     | 6.7   | 12.4  | 
| master | 9.5   | 14.6  |
~~~

I'll be ready to merge it in soon.

aothms:
With regards to the exception / assertion, I agree things were not ideal. However, by not throwing exception, invalid data will seep through the cracks now. For example:

~~~
#11=IFCGEOMETRICREPRESENTATIONCONTEXT('Plan','Model',3,.T.,#10,#6);
                                                        ^ Precision should be REAL
~~~

`f.by_id(11).Precision` will now happily return `*(float*) (&0xcccccc01)` (on a msvc debug build, otherwise 3 bytes of garbage). Therefore I vote for re-enabling the exceptions in the `asFloat()` and similar functions until we have something better, as that is what the rest of the codebase is currently expecting.

I agree a check on construction is preferable, for example in the `IfcSchema::SchemaEntity()` call. As that is were the untyped IfcParse instance is associated to a typed instance from the schema. If it fails that check, the instance would simply not be inserted in the maps of `IfcFile`. But that would be for a different PR I would say. The question then still remains what checks to do: attribute types, cardinalities or potentially even inverse cardinalities and where rules, the latter two being astronomically more complicated of course.

In case I don't hear back I will simply add the exceptions and otherwise merge as is.

stgatilov:
I agree with you.
I was under impression that caller always checks the type with isFloat function before calling asFloat. But it turns out that it is not so. Committed a small fix.

As for reducing memory consumed, I think implementing special treatment for tessellation entities should improve it a lot (along with performance). I'm not sure: what's the best place to discuss this topic?

aothms:
Ok, perfect, thanks.

Best place to discuss would be to open an issue here on github and label it improvement, that way we have a reference to the discussion when it gets implemented.

PR #88, opened 2016-06-07
Windows batch script enhancements

Stinkfist0:
Make possible to invoke run-cmake.bat, build-ifcopenshell.bat, and install-ifcopenshell.cmd directly from File Explorer or command prompt without VS env vars set, if BuildDepsCache file exists, by utilizing CMake's --build functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Prerequisites
 * Git
 * CMake (2.6 or newer)
 * Windows: Visual Studio 2008 or newer with C++ toolset, MinGW not supported currently
-* *nix: GCC 4.7 or newer, or Clang (any version should work, but not tested)
+* *nix: GCC 4.7 or newer, or Clang (any version)
 
 Dependencies
 -------------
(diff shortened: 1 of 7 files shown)


Stinkfist0:
Oops, accidentally deleted this branch.

aothms:
I like these changes, this will simplify things quite a bit. Thanks!

1) I would recommend not to read cached variables from both files, as I am bit worried that variables would be mixed between the two architectures if they are absent in `x64`, how about using a loop sorted on atime/mtime, e.g:

~~~
@echo off

for /f "tokens=*" %%f in ('dir BuildDepsCache-*.txt /o:-n /t:a /b') do (
    echo %%f
    goto :break_loop
)

:break_loop
~~~

2) Just out of curiosity, is this a typo or does this trailing dot have some significance: `if exist BuildDepsCache-x64.txt.`

3) Maybe it is an idea to show an error message if `%GEN_SHORTHAND%` is not defined. This is what would happen when a user runs the script by clicking in their explorer without invoking it from the cli first, am I right?

Stinkfist0:
Yeah, I'll do these enhancements. Regarding the trailing dot, it is intentional (it is "best practice" of some sort, even `if /?` uses it in the example, can't recall the actual reason though...) and actually the line that misses it has a typo.

Stinkfist0:
The fixes/enhancements are in.

aothms:
Great, thanks again Ali!

Issue #77, opened 2016-05-29
Define export symbol to have IfcGeom as a dynamic lib


aothms:
As per https://github.com/IfcOpenShell/IfcOpenShell/pull/76#issuecomment-222358392

Stinkfist0:
I could take care of this. Do you have preferences for the file and macro names for IfcGeom's import/export business?

aothms:
Well, there are some acronyms established as how to refer to IfcOpenShell: `IFCOS_`, `IOS_`, etc. You personally seem to use `IFCOS_`. What would your opinion be on simply `IFC_`? Similar to ICU uses `U_`. People hopefully wouldn't use multiple IFC libraries in their project.

Copying ICU once again, we would get `IFC_STATIC_IMPLEMENTATION`. Possibly `IFC_PARSE_STATIC_IMPLEMENTATION` and `IFC_GEOM_STATIC_IMPLEMENTATION` if two separate ones make sense (no clue, could be conditionally derived from one project-wide macro). It's a little lengthy, I am fine with `IFC_*_STATIC_LIB` for example as well.

Thoughts?

Stinkfist0:
IFC_STATIC_* sounds good. However, as building IFCOS libs as static libraries has been and currently is the default behavior and linking to static libraries requires no macro gymnastic, it would feel more natural to have the definition for the special case, i.e. IFC_DYNAMIC/SHARED_* which requires all these tricks. Also, I think it would be nice not to require client code the add the IFC_WHATEVER definition when using IFCOS, but instead when IFCOS is deployed, we would deploy a small configuration header which automatically has these definitions baked in. A quick pseudo draft:
```c++
// ifc_build_config.h
#pragma once
#define IFC_STATIC_BUILD
#define IFC_SCHEMA_2X3
// and so on...
```

jobermayr:
Why was it neccessary to build ifcparse SHARED and ifcgeom STATIC?

I introduced it with commit 4218890 vice versa since SHARED ifcgeom super lib worked on openSUSE ...

jobermayr:
Maybe you want to use a version-script:
https://cmake.org/pipermail/cmake/2008-August/023202.html

aothms:
[...]

PR #91, opened 2016-06-14
Bring back logic for BUILD_SHARED_LIBS.

jobermayr:
Also remove C++11 code.

An already implemented warning should be enough ...

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -432,7 +432,11 @@ if(NOT WIN32)
 	LINK_DIRECTORIES(${LINK_DIRECTORIES} /usr/lib /usr/lib64 /usr/local/lib /usr/local/lib64)
 endif()
 
-SET(IFCOPENSHELL_LIBRARIES IfcParse IfcGeom)
+if(BUILD_SHARED_LIBS)
+    SET(IFCOPENSHELL_LIBRARIES IfcGeom)
+else()
+    SET(IFCOPENSHELL_LIBRARIES IfcParse IfcGeom)
+endif()
 
 # IfcParse
 file(GLOB IFCPARSE_H_FILES ../src/ifcparse/*.h)
@@ -448,7 +452,7 @@ endforeach()
 
 set(IFCPARSE_FILES ${IFCPARSE_CPP_FILES} ${IFCPARSE_H_FILES})
 
-add_library(IfcParse ${IFCPARSE_FILES})
+add_library(IfcParse STATIC ${IFCPARSE_FILES})
 set_target_properties(IfcParse PROPERTIES COMPILE_FLAGS -DIfcParse_EXPORTS)
 
 IF(UNICODE_SUPPORT)
@@ -461,7 +465,7 @@ file(GLOB IFCGEOM_CPP_FILES ../src/ifcgeom/*.cpp)
 set(IFCGEOM_FILES ${IFCGEOM_CPP_FILES} ${IFCGEOM_H_FILES})
 IF(BUILD_SHARED_LIBS)
 	message(WARNING "Building IfcGeom as shared library not currently supported")
-        add_library(IfcGeom STATIC ${IFCGEOM_FILES})
+	add_library(IfcGeom SHARED ${IFCGEOM_FILES})
 ELSE()
         add_library(IfcGeom ${IFCGEOM_FILES})
 ENDIF()
@@ -524,7 +528,7 @@ INSTALL(FILES ${IFCGEOM_H_FILES}
 	DESTINATION ${INCLUDEDIR}/ifcgeom
 )
 
-INSTALL(TARGETS IfcParse IfcGeom IfcConvert IfcGeomServer
+INSTALL(TARGETS ${IFCOPENSHELL_LIBRARIES} IfcConvert IfcGeomServer
 	ARCHIVE DESTINATION ${LIBDIR}
 	LIBRARY DESTINATION ${LIBDIR}
 	RUNTIME DESTINATION ${BINDIR}
--- a/src/ifcparse/IfcParse.cpp
+++ b/src/ifcparse/IfcParse.cpp
@@ -559,9 +559,9 @@ double TokenFunc::asFloat(const Token& t) {
 const std::string &TokenFunc::asStringRef(const Token& t) {
 	std::string &str = t.lexer->GetTempString();
 	t.lexer->TokenString(t.startPos, str);
-	if (isString(t) || isEnumeration(t) || isBinary(t)) {
+	if ((isString(t) || isEnumeration(t) || isBinary(t)) && !str.empty()) {
 		//remove start+end characters in-place
-		str.pop_back();
+		str.erase(str.end()-1);
 		str.erase(str.begin());
 	}
 	return str;


src/ifcparse/IfcParse.cpp
```diff
@@ -559,9 +559,9 @@ double TokenFunc::asFloat(const Token& t) {
 const std::string &TokenFunc::asStringRef(const Token& t) {
 	std::string &str = t.lexer->GetTempString();
 	t.lexer->TokenString(t.startPos, str);
-	if (isString(t) || isEnumeration(t) || isBinary(t)) {
+	if ((isString(t) || isEnumeration(t) || isBinary(t)) && !str.empty()) {
 		//remove start+end characters in-place
-		str.pop_back();
+		str.erase(str.end()-1);
```
aothms: Thanks for this

cmake/CMakeLists.txt
```diff
@@ -448,7 +452,7 @@ endforeach()
 
 set(IFCPARSE_FILES ${IFCPARSE_CPP_FILES} ${IFCPARSE_H_FILES})
 
-add_library(IfcParse ${IFCPARSE_FILES})
+add_library(IfcParse STATIC ${IFCPARSE_FILES})
```
aothms: Is this a typo?

cmake/CMakeLists.txt
```diff
@@ -461,7 +465,7 @@ file(GLOB IFCGEOM_CPP_FILES ../src/ifcgeom/*.cpp)
 set(IFCGEOM_FILES ${IFCGEOM_CPP_FILES} ${IFCGEOM_H_FILES})
 IF(BUILD_SHARED_LIBS)
 	message(WARNING "Building IfcGeom as shared library not currently supported")
-        add_library(IfcGeom STATIC ${IFCGEOM_FILES})
+	add_library(IfcGeom SHARED ${IFCGEOM_FILES})
```
aothms: It would be sufficient to move `add_library()` outside of the if statement. `BUILD_SHARED_LIBS=1` sets `add_library` default to `STATIC`.

cmake/CMakeLists.txt
```diff
@@ -432,7 +432,11 @@ if(NOT WIN32)
 	LINK_DIRECTORIES(${LINK_DIRECTORIES} /usr/lib /usr/lib64 /usr/local/lib /usr/local/lib64)
 endif()
 
-SET(IFCOPENSHELL_LIBRARIES IfcParse IfcGeom)
+if(BUILD_SHARED_LIBS)
```
aothms: I don't understand the distinction here. Please clarify?

jobermayr: You see except IfcParseExamples all targets will use either both libraries (IfcParse and IfcGeom) or none of them. To have only one library to load all should be in IfcParse.

$ grep -r -n -e "TARGET_LINK_LIBRARIES" -A1
cmake/CMakeLists.txt:459:       TARGET_LINK_LIBRARIES(IfcParse ${ICU_LIBRARIES} ${Boost_LIBRARIES})
cmake/CMakeLists.txt-460-ENDIF()
--
cmake/CMakeLists.txt:473:TARGET_LINK_LIBRARIES(IfcGeom IfcParse)
cmake/CMakeLists.txt-474-
--
cmake/CMakeLists.txt:495:TARGET_LINK_LIBRARIES(IfcConvert ${IFCOPENSHELL_LIBRARIES} ${OPENCASCADE_LIBRARIES} ${Boost_LIBRARIES} ${OPENCOLLADA_LIBRARIES} ${ICU_LIBRARIES})
cmake/CMakeLists.txt-496-if (NOT WIN32)
--
cmake/CMakeLists.txt:505:TARGET_LINK_LIBRARIES(IfcGeomServer ${IFCOPENSHELL_LIBRARIES} ${OPENCASCADE_LIBRARIES} ${Boost_LIBRARIES} ${ICU_LIBRARIES})
cmake/CMakeLists.txt-506-if (NOT WIN32)
--
src/examples/CMakeLists.txt:21:TARGET_LINK_LIBRARIES(IfcParseExamples IfcParse)
src/examples/CMakeLists.txt-22-set_target_properties(IfcParseExamples PROPERTIES FOLDER Examples)
--
src/examples/CMakeLists.txt:25:TARGET_LINK_LIBRARIES(IfcOpenHouse ${IFCOPENSHELL_LIBRARIES} ${OPENCASCADE_LIBRARIES})
src/examples/CMakeLists.txt-26-set_target_properties(IfcOpenHouse PROPERTIES FOLDER Examples)
--
src/ifcmax/CMakeLists.txt:31:TARGET_LINK_LIBRARIES(IfcMax ${IFCLIBS} Comctl32.lib zlibdll.lib bmm.lib core.lib CustDlg.lib edmodel.lib expr.lib
src/ifcmax/CMakeLists.txt-32-    flt.lib geom.lib gfx.lib gup.lib imageViewers.lib ManipSys.lib maxnet.lib Maxscrpt.lib
--
src/qtviewer/CMakeLists.txt:55:TARGET_LINK_LIBRARIES (QTviewer ${IFCLIBS} ${QT_LIBRARIES} ${OPENCASCADE_LIBRARIES})

aothms: Ok, but `IFCOPENSHELL_LIBRARIES` is also used in the `INSTALL()` statement. Meaning that IfcParse is not installed with `BUILD_SHARED_LIBS`.

aothms:
Merged with addition of the following changes:
https://github.com/IfcOpenShell/IfcOpenShell/pull/97/commits/478bda30bcf64dc97a85a2443be16dd12d632bc6

jobermayr:
> Merged with addition of the following changes:
> https://github.com/IfcOpenShell/IfcOpenShell/pull/97/commits/478bda30bcf64dc97a85a2443be16dd12d632bc6
Thanks for **destroying** my logic again ...

aothms:
@jobermayr sorry to hear that. I asked some questions about the contents of your PR and got only one reply I didn't completely understand. Since people were waiting for the C++11 fix I decided not to wait for another iteration of comments.

For the sake of completeness here is an overview of my changes:

~~~
 -add_library(IfcParse STATIC ${IFCPARSE_FILES})
+add_library(IfcParse ${IFCPARSE_FILES})
~~~

IMHO this is a necessary change. We have spent effort defining symbols to be externally visible in `IfcParse`. Therefore the library type can and should be inferred by the value of `BUILD_SHARED_LIBS`.

~~~
 IF(BUILD_SHARED_LIBS)
-	message(WARNING "Building IfcGeom as shared library not currently supported")
-	add_library(IfcGeom SHARED ${IFCGEOM_FILES})
-ELSE()
-        add_library(IfcGeom ${IFCGEOM_FILES})
+	message(WARNING "Building IfcGeom as shared library currently not supported depending on platform and compiler options")
 ENDIF()
~~~

As I said in my earlier comment, IMHO functionally this is equivalent since `BUILD_SHARED_LIBS` implies `SHARED` as a default. Only the warning message has been elaborated slightly.

~~~
 +if(WIN32)
+    add_library(IfcGeom STATIC ${IFCGEOM_FILES})
+else()
+    add_library(IfcGeom ${IFCGEOM_FILES})
+endif()
~~~

Only on Windows IfcGeom is forced to be static as it definitely does not work there as a dynamic library. On nix it works depending on compiler options. Which as mentioned, when compiled with the recommended `-fvisibility=hidden` (which our build script does) it does not, but alas. Maybe this should be `MSVC` rather than `WIN32` not sure about symbol visibility in e.g mingw.

~~~
 -INSTALL(TARGETS ${IFCOPENSHELL_LIBRARIES} IfcConvert IfcGeomServer
+INSTALL(TARGETS IfcParse IfcGeom IfcConvert IfcGeomServer
~~~

Both `IfcParse` and `IfcGeom` are installed, since we have a library and an archive destination this only makes sense to me.

-

Those were the changes I made. If some of them somehow broke something for you, that is unfortunate and we can discuss that. At this moment I don't know which of the changes broke something for you, what it is you are trying to accomplish and what justifies your hostile comment, but feel free to clarify. Frankly, the option to keep `IfcGeom` as a dynamic library on linux has only been kept as a courtesy to you, because IMHO it is broken until #77 is fixed.

PR #112, opened 2016-08-04
cmake and travis

berndhahnebach:
https://github.com/IfcOpenShell/IfcOpenShell/issues/111

https://sourceforge.net/p/ifcopenshell/discussion/1782718/thread/8710a757/

https://travis-ci.org/berndhahnebach/IfcOpenShell/builds/149752734

If travis is activated for PullRequests, any PullRequest will be checked on travis by default. See as an example: https://github.com/FreeCAD/FreeCAD/pull/234

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,5 +35,5 @@ script:
   - cd cmake
   - mkdir build
   - cd build
-  - cmake -DCOLLADA_SUPPORT=False -DUSE_IFC4=False -DBUILD_IFCPYTHON=True -DUNICODE_SUPPORT=True  ..
+  - cmake -DCOLLADA_SUPPORT=False -DUSE_IFC4=False -DBUILD_IFCPYTHON=True -DUNICODE_SUPPORT=True -DOCC_LIBRARY_DIR=/usr/lib/x86_64-linux-gnu  ..
   - make -j2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,39 @@
+language: cpp
+compiler: gcc
+os: linux
+dist: trusty
+sudo: required
+
+
+before_install:
+  - sudo apt-get update -qq
+
+
+install:
+  - sudo apt-get install -qq gcc-4.8 g++-4.8
+
+  - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 90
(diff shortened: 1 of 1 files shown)


berndhahnebach:
see: https://github.com/IfcOpenShell/IfcOpenShell/issues/111#issuecomment-237549548

Stinkfist0:
Great, this a good starting point and I even used some of this to enhance the *nix build instructions already. Naturally we want to the Travis build to include all features (OpenCOLLADA support) and have a full build matrix (MSVC, Clang too(?), static and shared libs, IFC2x3 and IFC4, etc.).

berndhahnebach:
Microsoft builds we do with Appveyor at FreeCAD since Travis does not support windows AFAIK.
https://ci.appveyor.com/project/yorikvanhavre/freecad/build/1.0.48

Stinkfist0:
Ah, right, was mixing those two together.

berndhahnebach:
occt library problem has been solved by the latest commits you guys did. As a result I need to specify the occ ibrary path on travis. Means on a system with only one oce from package manager oce librarys are not found anymore.

berndhahnebach:
By setting occt library path travis builds fine ... 
https://travis-ci.org/berndhahnebach/IfcOpenShell/builds/150841767

aothms:
Ok, so what's the plan with this, is it ready to be merged in?

For me this is good enough. I am still working on my own build system that will do visual diffs on IfcConvert output. So for me travis does not need to be feature complete and can lack osx/windows, as that will all be covered by my system.

Stinkfist0:
IMO the OpenCOLLADA support should be added (simply code-paste from the README), but otherwise this should be OK.

berndhahnebach:
It is far from being full featured, it is just a start. But most things start like this. 

It works, it shows if a new commit compiles on ubuntu gcc. I'd be happy to see it in master. I may improve it but I'm not gone promising anything ... 

BTW: If you (thomas) activate it for PullRequests any PullRequest will be compiled on travis.

aothms:
> By setting occt library path travis builds fine ... 

Perhaps we need to come up with more sensible default paths to try so that the `-DOCC_LIBRARY_DIR=/usr/lib/x86_64-linux-gnu` is not necessary anymore. This is kind of the default for packages from the package manager after debian introduced multiarch. But that's another story, frankly I'm a bit tired of fighting with CMake.

I will just merge it in then. Additions always welcome.

PR #113, opened 2016-08-06
Bash on Ubuntu on Windows instructions

Stinkfist0:
Just tested and works perfectly.


aothms:
Perfect! Thanks for checking and for your diligence towards the readme docs :)

PR #120, opened 2016-08-18
*nix build instructions tweaks

Stinkfist0:


README.md
```diff
@@ -88,14 +88,14 @@ for additional information):
...
-If not, you will need to compile Open CASCADE yourself:
+If not, you will need to compile Open CASCADE yourself (note that the build takes a long time):
 
     $ sudo apt-get install libftgl-dev libtbb2 libtbb-dev libgl1-mesa-dev libfreetype6-dev
     $ git clone https://github.com/tpaviot/oce.git
     $ cd oce
     $ mkdir build && cd build
     $ cmake ..
-    $ make
+    $ make -j
```
aothms: Is this a good suggestion, wouldn't it spawn too many processes when using `-j` without an integer? I have no experience with this.

README.md
```diff
@@ -107,30 +107,24 @@ For building IfcConvert with COLLADA (.dae) support (on by default), OpenCOLLADA
...
-        $ export OCC_LIBRARY_PATH
-    If building with COLLADA support (path might vary):
-        $ OPENCOLLADA_INCLUDE_DIR="/usr/local/include/opencollada"
-        $ OPENCOLLADA_LIBRARY_DIR="/usr/local/lib/opencollada"
-        $ export OPENCOLLADA_INCLUDE_DIR
-        $ export OPENCOLLADA_LIBRARY_DIR
-    $ cmake ../cmake
-    $ make
+    $ mkdir build && cd build
+    $ cmake ../cmake -DOCC_LIBRARY_DIR=/usr/lib/x86_64-linux-gnu/ \
```
aothms: This is a good suggestion. I think we need to phase out the possibility to use env vars, as it is not so clear to me how this would need to work in conjunction with the CMake cache. Any thoughts? (wrt the win build script and in general)

README.md
Stinkfist0: The default behavior would appear to be number of cores + 1, and by being the default behavior, I'd assume it should not be a "bad" thing at least.

README.md
Stinkfist0: My main motivation for this was simply to make the instructions more concise. I think env vars are OK, one just needs to remember clear the CMakeCache.txt, but that's recommended to to be done pretty much every time in any case when changes to the build configuration are done (the run-cmake.bat does this already).

aothms: Ok, fair enough. Thanks again for you effort.

README.md
aothms: Hm. interesting. That's exactly what the nix build script tries to determine the hard way.
https://github.com/IfcOpenShell/IfcOpenShell/blob/master/nix/build-all.sh#L64

PR #124, opened 2016-08-25
Windows/MSVC + OCCT 7.0.0

### LINKED ISSUE #121: Windows/MSVC: possibility to build official latest Open Cascade (7.0.0) instead of OCE
FYI this is incoming soon.

### LINKED ISSUE #123: MSVC debug build broken
Due to #114.

Stinkfist0:
- not enable yet by default, but we probably could after some testing period
- closes #121
- also got rid of the hundreds of linker warnings regarding OCE/OCCT
- patches implement using file copying so that no new tools are required, PowerShell could be an option though
- also fixes #123


Stinkfist0:
Oh, I am testing with slicing with the Duplex model and I notice it crashes with OCCT 7.0.0. Did you happen to test slicing Duplex by any change on Linux?

aothms:
Great! Many thanks. I am having a look now.

> Did you happen to test slicing Duplex by any change on Linux with OCCT 7.0.0?

No, I will. It doesn't crash on Windows or you just didn't happen to check?

Is there a reason why you copied over the entire files you want to patch rather than submit patches and `git apply` them like you did before?

Stinkfist0:
Frankly, I was simply tired after hours of OCCT & OCE building to research into if/how ´git apply´ could be used for non version controlled files. I tried quickly creating patch files with WinMerge but those turned out not to be in format `git apply` woudl accept. If this is doable, I'm happy to provide a cleaner PR (some accidental merge commits already sneaked in).

aothms:
Well, one of the reasons I asked is that I get unexplainable `%ERRORLEVEL%=1` from the `copy` command. Even though the script prints `1 file(s) copied.` and the file indeed seems updated.

Then later on I get a CMake error:

~~~
CMake Error at adm/cmake/occt_macros.cmake:43 (math):
  math cannot parse the expression: "32 + 32*(/8)": syntax error, unexpected
  exp_DIVIDE, expecting exp_OPENPARENT or exp_NUMBER (10)
~~~

meaning that `${CMAKE_SIZEOF_VOID_P}` is empty.

# COMPILER_BITNESS variable
macro (OCCT_MAKE_COMPILER_BITNESS)
  math (EXPR COMPILER_BITNESS "32 + 32*(${CMAKE_SIZEOF_VOID_P}/8)")
endmacro()

Strange errors. Not sure what all is happening here.

Stinkfist0:
OK, sounds weird, I'll retry building OCCT in a sec. The CMake problems sound weired - what's your version BTW? I'm using 3.6.1.

Stinkfist0:
OK, I'm getting the same copy error.

Stinkfist0:
Fix committed. I'll look into cleaning up this PR tomorrow.

aothms:
This is fine for me. Little reason to assume the `copy` would fail. Perhaps using `xcopy` or `robocopy` would also have been an option.

Stinkfist0:
A swift research sheds no light on using `git diff/apply` for non version controlled files. Would you like be to use PowerShell for patching or are the file copy patches OK?

Stinkfist0:
I'll take that back: `git diff` works OK but I cannot figure out how to use `git apply` for untracked files: the command runs successfully but it doesn't alter the file at all.

Stinkfist0:
So, if this satisfies you as is, it's ready for merging. I can squash the commits if wanted.

aothms:
Thanks, works like a charm, as always.

> I wonder why on earth OCCT has changed usage of bool to unsigned int, oh well...

Yeah, this is a bit peculiar. Maybe a better way to go about this would have been to use `!!b` rather than `b!=0`. As the first one would be less awkward when OCCT decides to typedef `Standard_Boolean` back to `bool`. Or doesn't it get rid of the warning? Well, doesn't matter.

Just have one small thing relating to this ever annoying PCRE library dir. I will commit it after merging.

> >    Did you happen to test slicing Duplex by any change on Linux with OCCT 7.0.0?
>
> No, I will. It doesn't crash on Windows or you just didn't happen to check?

I can confirm the same happens on Windows. Will have a look.

PR #144, opened 2016-10-18
Geometry generated is incorrect if there are nested IfcMappedItem representations 

carmenfan:
see https://github.com/3drepo/IfcOpenShell/issues/1


aothms:
Thanks. Do you by any chance have an example IFC file you can share?

aothms:
Created a test file: https://github.com/IfcOpenShell/IfcOpenShell/commit/1e86a31b814a0fad32a33f1ad8b9802a7e367015

### Before and after:
schermafdruk 2016-11-06 16 07 01schermafdruk 2016-11-06 16 07 04

PR #147, opened 2016-10-19
More elaborate wire ordering check and shell sewing code

aothms:


jf---:
@aothms , a current PR at `tigl` also deals with this aspect... [might be interesting](https://github.com/DLR-SC/tigl/blob/master/src/common/tiglcommonfunctions.cpp)?

aothms:
@jf-- interesting, thanks. Can you point me to it directly? I couldn't find it in your link.

src/ifcgeom/IfcGeomFaces.cpp
```diff
@@ -307,11 +298,47 @@ bool IfcGeom::Kernel::convert(const IfcSchema::IfcFace* l, TopoDS_Shape& face) {
...
+				
+				face = TopoDS::Face(sfs.Face());				
 			}
 		}
 	}
 
 	if (success) {
+		// If the wires are reversed the face needs to be reversed as well in order
+		// to maintain the counter-clock-wise ordering of the bounding wire's vertices.
+		if (num_bounds == 1 || true) {
```
aothms: Note to self: change this condition.

jf---:
@aothms nothing specific, i glanced over it and it seemed relevant, non-trivial, but easy to follow code, tangential to this topic, possible...

aothms:
@jf---, maybe we should establish some common best practices and code snippets, also re the discussions in https://github.com/tpaviot/pythonocc-core/issues/351 and similar to what has been made available @ http://opencascade.wikidot.com/recipes 

I would happily lend http://academy.ifcopenshell.org for this purpose, but perhaps the github OCE wiki page is a more neutral environment.

PR #149, opened 2016-10-20
Add layer information (IfcPresentationLayerAssignment) to IfcConvert'…

### LINKED ISSUE #141: Add layer information to IfcConvert's XML output?
I've been noticing that the XML file has gained various new tidbits recently. Would the layer information also a good applicant to be included in the file?

Stinkfist0:
…s XML output. Closes #141

### DIFF
--- a/src/ifcconvert/XmlSerializer.cpp
+++ b/src/ifcconvert/XmlSerializer.cpp
@@ -33,7 +33,9 @@
 using boost::property_tree::ptree;
 using namespace IfcSchema;
 
-static std::map<std::string, std::string> argument_name_map;
+namespace {
+
+std::map<std::string, std::string> argument_name_map;
 
 // Format an IFC attribute and maybe returns as string. Only literal scalar 
 // values are converted. Things like entity instances and lists are omitted.
@@ -105,10 +107,8 @@ boost::optional<std::string> format_attribute(const Argument* argument, IfcUtil:
 	return value;
 }
 
-// Formats an entity instances as a ptree node, and insert into the DOM. Recurses
-// over the entity attributes and writes them as xml attributes of the node.
-ptree& format_entity_instance(IfcUtil::IfcBaseEntity* instance, ptree& tree, bool as_link = false) {
-	ptree child;
+// Appends to a node with possibly existing attributes
+ptree& format_entity_instance(IfcUtil::IfcBaseEntity* instance, ptree& child, ptree& tree, bool as_link = false) {
 	const unsigned n = instance->getArgumentCount();
 	for (unsigned i = 0; i < n; ++i) {
 		const Argument* argument = instance->getArgument(i);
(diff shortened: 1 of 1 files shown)


Stinkfist0:
Oh, noticed an inconsistency: I'm prefixing the layer ID's with #, but this is not done for types' or properties' IDs. Maybe we should simply remove the `id` which is duplicate to `Name` in any case. I.e.
` <IfcPresentationLayerAssignment xlink:href="#A-WALL-MBNI"/>` would stay as is but the ` <IfcPresentationLayerAssignment id="#A-WALL-MBNI" Name="A-WALL-MBNI"/>` would become ` <IfcPresentationLayerAssignment Name="A-WALL-MBNI"/>`

src/ifcconvert/XmlSerializer.cpp
```diff
@@ -328,10 +368,20 @@ void XmlSerializer::finalize() {
 		}
 	}
 
+    // Layer assignments
+    IfcPresentationLayerAssignment::list::ptr layer_assignments = file->entitiesByType<IfcPresentationLayerAssignment>();
+    for (IfcPresentationLayerAssignment::list::it it = layer_assignments->begin(); it != layer_assignments->end(); ++it) {
+        // IfcPresentationLayerAssignments don't have GUIDs (only optional Identifier) so use name as the ID
+        ptree node;
+        node.put("<xmlattr>.id", "#" + (*it)->Name());
```
aothms: The `"#"` is not part of the `id` (so should be removed here), but as part of the `href` it signifies that the link points to a node in the same document. The xlink w3c spec is a bit too verbose to me now, but the example here [1] illustrates this.

[1] https://en.wikipedia.org/wiki/XLink#Simple_links

aothms:
What about duplicate `IfcPresentationLayerAssignment.Name`s? There is nothing in the IFC spec that would argue they are unique. But XML id attributes better be unique. That's why I in the #141 suggested a set of string with layer names.

It might be a bit of a stretch, but I can imagine there might be files in which the layer assignment name is not unique. For example when multiple files get merged into one, or perhaps there are simply some exporters that emit the multiple assignments to the same string label.

Stinkfist0:
Aye, will fix this shorty.

Stinkfist0:
^

aothms:
I don't think postfixing is necessary. The layer assignment name I think is what matters, not really the assignment instance.

Stinkfist0:
OK, should be fixed now.

Stinkfist0:
Hi. Does this look OK now?

aothms:
Suppose so :) sorry for the delay and many thanks as always!

Stinkfist0:
FYI, I just encountered a file that used IfcPresentationLayerWithStyle and this seemed to handle also them just fine:
```xml
<IfcPresentationLayerWithStyle id="something" Name="something" LayerOn="true" LayerFrozen="false" LayerBlocked="false"/>
...
<IfcPresentationLayerWithStyle xlink:href="#something"/>
```

PR #150, opened 2016-10-21
IfcConvert: --bulding-storey filter

andersgb:
When converting large .ifc files to .obj, I wanted to select specific storeys. Is this a feature of general interest?

Especially the second commit is somewhat hacky, let me know if you want me to clean this up.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -128,6 +128,7 @@ int main(int argc, char** argv) {
 		("output-file", boost::program_options::value<std::string>(), "output geometry file");
 
     std::vector<std::string> entity_vector, names;
+    std::string building_storey;
     double deflection_tolerance;
     boost::program_options::options_description geom_options("Geometry options");
 	geom_options.add_options()
@@ -188,6 +189,8 @@ int main(int argc, char** argv) {
             "A list of names or wildcard patterns that should be included in or excluded from the "
             "geometrical output, depending on whether --exclude or --include is specified. "
             "The names are handled case-sensitively. Cannot be placed right before input file argument.")
+        ("building-storey", boost::program_options::value<std::string>(&building_storey),
+            "Only include the specified building storey")
         ("no-normals",
             "Disables computation of normals. Saves time and file size and is useful "
             "in instances where you're going to recompute normals for the exported "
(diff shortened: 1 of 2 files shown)


Stinkfist0:
Hi. I made this a while back: https://github.com/Tridify/IfcOpenShell/commit/91051f020a0763a74313585ecb34bf739b8d33ec (the original code here might contain some bugs). I've been planning to submit it to here but never really got around to it yet.

src/ifcconvert/IfcConvert.cpp
```diff
@@ -429,6 +432,29 @@ int main(int argc, char** argv) {
 
     IfcGeom::Iterator<real_t> context_iterator(settings, input_filename);
 
+    if (!building_storey.empty()) {
+        bool found_storey = false;
+        IfcParse::IfcFile *ifc_file = context_iterator.getFile();
+        IfcEntityList::ptr storeys = ifc_file->entitiesByType("IfcBuildingStorey");
```
aothms: Better use the templated version: `ifc_file->entitiesByType<IfcSchema::IfcBuildingStorey>()`, than you get a typed list back.

src/ifcconvert/IfcConvert.cpp
```diff
@@ -429,6 +432,29 @@ int main(int argc, char** argv) {
...
+
+        for (IfcEntityList::it i = storeys->begin(); i != storeys->end(); ++i) {
+            if ((*i)->getArgument(2)->toString().find(building_storey) != std::string::npos) {
```
aothms: With the typed instances, you can simply say `storey->Name()`. Beware that the attribute is optional, so check first with `storey->hasName()`.

src/ifcconvert/IfcConvert.cpp
```diff
@@ -429,6 +432,29 @@ int main(int argc, char** argv) {
...
+                std::cout << "Selecting " << building_storey << " for filtering..." << std::endl;
```
aothms: Better use `Logger::Message(Logger::LOG_NOTICE, ` so that users can set the verbosity

src/ifcconvert/IfcConvert.cpp
```diff
@@ -429,6 +432,29 @@ int main(int argc, char** argv) {
...
+                IfcSchema::IfcBuildingStorey* selected = (*i)->as<IfcSchema::IfcBuildingStorey>();
+                IfcTemplatedEntityList< IfcSchema::IfcRelContainedInSpatialStructure >::ptr in_storey__ =
+                    selected->ContainsElements();
+                std::cout << "IfcRelContainedInSpatialStructure size " << in_storey__->size() << std::endl;
+                IfcSchema::IfcRelContainedInSpatialStructure* in_storey = *(selected->ContainsElements()->begin());
+                context_iterator.includeProducts(in_storey->RelatedElements());
```
aothms: This needs to be recursive. So that decomposed elements are also included.

aothms:
Thanks for this. I made some comments about the code in-line.

The proposal by @Stinkfist0 is more general and I can imagine it is useful to consider for other cases, for example, to process a single wall with its openings and windows. Also the fact that it specifically selects by `Name` and has an option to select by `GlobalId` is nice to have.

Is there anyway we can unite these efforts?

Stinkfist0:
I plan to provide a PR in the near future of my implementation, I'm planning also adding `--guid` option (e.g. `--include --guid 1UhvOtJ6T8Tgo1W$J1I_D$ --with-children`).

andersgb:
Great, thanks for the review. I don't really know what I'm doing, just hammered on this until I got something working. It seems the proposal of @Stinkfist0 is more general and it covers my specific need here as well. Maybe just close this PR for now?

And yeah, the --guid option would be very useful.

aothms:
> Maybe just close this PR for now?

Ok, better luck next time, thanks for your effort though.

Stinkfist0:
@aothms BTW, I'm not really a fan of --with-children name - would you have some better suggestion, something a bit more "IFC-like"?

aothms:
Well, I agree.

It would be a grouping of the following relationships if I'm not mistaken: `IsDecomposedBy`, `HasOpenings`, `FillsVoid`, `ContainedInStructure`. I cannot really think of any umbrella term that matches this. Something like "relation" is obviously too broad.

I am also fine with something like `--recurse` if you like that better, but it isn't necessarily more informative.

Stinkfist0:
Perhaps something like `containment` would sound informative.

aothms:
> Perhaps something like containment would sound informative.

IMHO it would cover `HasOpenings`, `FillsVoid`, `ContainedInStructure`. `IsDecomposedBy` is really a different beast. E.g typically you have an `IfcCurtainWall` or `IfcWindow` with `IfcMembers` and `-Plates`, the curtain wall exists by nature of their decomposing elements, but does not really contain them I would argue.

My preference is for something general (recurse, children) and to document it properly in the help message.

Stinkfist0:
Maybe I'll stick to the original name then for now 😄 I'll start working on providing a PR soon.

PR #159, opened 2016-11-11
IfcConvert/IfcGeom --traverse option

Stinkfist0:
e.g. `IfcConvert Duplex_A_20110907.ifc level1.dae --include --traverse --names "Level 1"` will produce a DAE that contains only the objects of the 1st floor:

image

Note that the results is slightly different compared to e.g. Solibri:
image

image

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -197,10 +197,10 @@ int main(int argc, char** argv) {
         ("generate-uvs",
             "Generates UVs (texture coordinates) by using simple box projection. Requires normals. "
             "Not guaranteed to work properly if used with --weld-vertices.")
-        ("with-children",
+        ("traverse",
             "Applies --include or --exclude also to the decomposition and/or containment (IsDecomposedBy, "
-            "HasOpenings, FillsVoid, ContainedInStructure) of the filtered element, e.g. "
-            "--include --with-children --names \"Level 1\" includes element with name \"Level 1\" and all of its children.");
+            "HasOpenings, FillsVoid, ContainedInStructure) of the filtered entity, e.g. "
+            "--include --traverse --names \"Level 1\" includes entity with name \"Level 1\" and all of its children.");
 
     std::string bounds;
     boost::program_options::options_description serializer_options("Serialization options");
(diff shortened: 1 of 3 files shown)


aothms:
Great work. Couple of additional observations just of reference:

* By specifying `--include --traverse --names "Level 1"` we cannot have a filter on entity types anymore, in particular as is visible in your screen-shot the opening elements are visible. In all fairness this was to be expected, but thinking naively it is still a bit surprising, as users would think it is orthogonal to the implicitly hidden types `{IfcOpeningElement, IfcSpace}`. No easy solution for this I think.
* So the door is on "Level 2", but still we wouldn't get it if we ask for `--include --traverse --names "Level 2"`. Wouldn't be that much of an issue as long as the assignment to building storeys is consistent.  This is simply a consequence of the way `get_decomposing_entity()` works. At least it seems there is no overlap in the output.
* `--include --traverse --entities IfcWall` is also a rather nice example as you get walls with windows and doors, or combined with your earlier enhancement: `--include --traverse --names "*Exterior*"` really really awesome.

So, I think these issues are all fine, what I hope we can do someday is turn this into a simple grammar to make it a bit more expressive and predictable.

exterior

Stinkfist0:
About your first point: yeah, this is something I'd like to solve quite soon. Maybe if there's no much objection, we could change `include` and `exclude` to accept type parameters so we could do e.g. `--include=names "Level 1" --exclude=entities IfcOpeningElement IfcSpace` (this would be the default behavior when `--include=names` would be used).

aothms:
No objection at all from my side. As I have been saying, in the end we may need some sort of grammar with arbitrary nesting of boolean operators, augmented with IfcProperty selections perhaps, but I think what you are suggesting is a reasonable middle ground.

Stinkfist0:
Great, I'll try to look into this right away (tomorrow).

Stinkfist0:
OK, finally got time to really start working with this (ability to do e.g. `--include=names "Level 1" --exclude=entities IfcOpeningElement IfcSpace`) . I'll try to provide a PR tomorrow.

aothms:
Great

Stinkfist0:
I'm making some progress but I won't have anything concrete by today.

PR #160, opened 2016-11-18
Fix build when USE_IFC4=1

Stinkfist0:
BTW would be nice to configure Travis to build the project using both configurations. That would help to catch these early on.


aothms:
Good idea. Not the first time this has happened. Let's see what travis says: #161

PR #166, opened 2016-11-24
scripting pane and a property panel tab to the PyQt geom.app viewer 

jakob-beetz:
Initial pull request for an interactive viewer with scriptability options as an extension to the PyQT app

see https://youtu.be/Jxot2ERo-zU for a short demo

- added a property panel shows all properties from IfcPropertySets attached to the currently selected object


aothms:
A big thank you is in place here! Integrated into main app and small clean ups in #167.

PR #167, opened 2016-11-28
Interactive console and Property panel tab

aothms:
screenshot
Picture says it all, amazing contribution by @jakob-beetz. Closes #166.


thorade:
Possibly silly question: 
To start this viewer, I would just `cd` to the `IfcOpenShell\src\ifcopenshell-python\ifcopenshell\geom` directory and then run `python app.py` !?
(Asking, because that did not work for me...)

aothms:
@thorade no indeed, it could be made to work like that if you add this `if __name__ == '__main__' business and make sure imports work from packages as well as directly. Pull requests welcome btw!

For now see:
https://github.com/IfcOpenShell/IfcOpenShell/commit/1d1fb6d27a5bb1f221aed4dc290c1dc99a324e0f#commitcomment-17161165

Issue #171, opened 2016-12-20
Windows build script and spaces in path


LukasMHD:
Sorry if this is a stupid issue, but everytime i try to compile the project for windows i get this error message.
I installed git and cmake and added them to path, which brought me a step further but now this happens: 

> "this script fetches and builds all IfcOpenShell dependencies
> 
> Generator not passed but VisualStudioVersion=14.0 environment variable detected:
> using `Visual Studio 14 2015` as the generator.
> build-type-cfg.cmd: Warning: BUILD_CFG not specified - using the default RelWithDebInfo
> 
> Script configuration:
> * CMake Generator       = `Visual Studio 14 2015`
>   - Passed to CMake -G option.
> * Target Architecture   = x86
>   - Whether were doing 32-bit (x86) or 64-bit (x64) build.
> * Dependency Directory  = C:\Users\...\Documents\IfcOpenShell-master 1 \IfcOpenShell-master\deps
>   - The directory where IfcOpenShell dependencies are fetched and built.
> * Installation Directory = C:\Users\...\Documents\IfcOpenShell-master 1 \IfcOpenShell-master\deps-vs2015-x86-installed
>   - The directory where IfcOpenShell dependencies are installed.
> * Build Config Type     = RelWithDebInfo
>   - The used build configuration type for the dependencies.
>     Defaults to RelWithDebInfo if not specified.
> * Build Type            = Build
>   - The used build type for the dependencies (Build, Rebuild, Clean).
>     Defaults to Build if not specified. Rebuild/Clean also uninstalls Python (if it was installed by this script).
> * IFCOS_USE_OCCT        = FALSE
>   - Use the official Open CASCADE instead of the community edition.
> * IFCOS_INSTALL_PYTHON  = TRUE
>   - Download and install Python.
>     Set to something other than TRUE if you wish to use an already installed version of Python.
> * IFCOS_USE_PYTHON2     = FALSE
>   - Use Python 2 instead of 3.
>     Set to TRUE if you wish to use Python 2 instead of 3. Has no effect if IFCOS_INSTALL_PYTHON is not TRUE.
> * IFCOS_NUM_BUILD_PROCS = 4
>   - How many MSBuild.exe processes may be run in parallel.
>     Defaults to NUMBER_OF_PROCESSORS. Used also by other IfcOpenShell build scripts.
> 
> Der Befehl "C:\Users\...\Documents\IfcOpenShell-master" ist entweder falsch geschrieben oder
[...]

PR #173, opened 2016-12-30
Windows/MSVC: support spaces in the IfcOpenShell build directory's pa…

Stinkfist0:
…th. re: #171

I also tested a path with spaces and non-ASCII characters and non-ASCII characters caused problems for find_file(), but e.g. find_package(Boost) worked OK (I was able to build all of the dependencies fine though).

Thanks @klacol for pointers regarding the batch files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,9 @@ Dependencies
 
 Building IfcOpenShell
 ---------------------
+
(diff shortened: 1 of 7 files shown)


aothms:
Great! Thanks.

PR #177, opened 2017-01-04
log exceptions on `get_info` method 

jf---:
log exceptions on `get_info` method
these exceptions currently are actively being hidden, which is
intransparent / could make debugging challenging

improved formatting tot pep8


aothms:
Wonderful, many thanks!

Issue #122, opened 2016-08-23
Add progressbar for XML serializer


Stinkfist0:
There's seems to be a really long pause (can be tens of seconds, maybe even minute, depending how large the file is) between "Done scanning file" and "Creating geometry..." phases. Should maybe be inspected if there is a real performance issue somewhere here.

aothms:
Could it be because of: https://github.com/IfcOpenShell/IfcOpenShell/commit/562f38756f8f2d2ecea8879112019d10dd9a2154 Maybe we need to check first if `--center-model` is set prior to iterating over all product placements.

A minute sounds like a whole lotta time for processing the product placements too. Maybe there is still something left to optimize there.

aothms:
@Stinkfist0 I tried this on `DC_Riverside_Bldg-LOD_300.ifc` (279mb) and I cannot observe anything beyond a fraction of a second. Are we talking about much larger files here, something you can share perhaps?

Stinkfist0:
I'll try to dig up some files to test again this.

Stinkfist0:
OK, so revisiting this, and this issue can be observed writing an XML output for a file that has a lot of property (and layer) information, so maybe at least add a simple progress print for that.

aothms:
Ok!

Moult:
PR seems to be merged, so this is solved and can be closed?

aothms:
Well we still don't have a progress bar.

> OK, so revisiting this, and this issue seem to be present when writing an XML output for a file that has a lot of property (and layer) information, so maybe at least add a simple progress print for that.

This is likely just the inefficiencies for calculating inverses we suffered from at the time.

Issue #73, opened 2016-05-26
Unicode characters in IfcConvert's XML output


Stinkfist0:
I think something has changed during the past year or so regarding IfcConvert's Unicode handling. In some older files generated by IfcConvert e.g. `J\X2\00E4\X0\tehuone` in IFC file has resulted in `J\u00e4tehuone` in XML. However, in more recent XML files I'm seeing `Jätehuone`. Any ideas what might have changed? How does these behave for you other users?

aothms:
The default encoder has changed during that time: https://github.com/IfcOpenShell/IfcOpenShell/commit/a58b2ede63e927b6e28afb58d508bebcb5d98728#diff-2
I think UTF-8 is a more natural default than this escaped unicode notation.

Stinkfist0:
OK, agreed in general it's better to have them human-readable straight away, although the escaped notation was more handy for my further processing. But I can escape these on my own no problem.

aothms:
Well, feel free to create a command line option. I'm sure there are others as well that prefer their output to be ASCII-safe.

Stinkfist0:
Yeah, command-line option could be nice.

Stinkfist0:
If/when this is made an option, it would be nice that the behavior would be consistent for all IfcConvert's output formats (XML, DAE, SVG, etc.), possible restrictions applied.

iras:
A command line option is a great idea. Tripped up by the line below:

`#11= IFCSURFACESTYLE('\X2\00C70130\X0\ZG\X2\0130\X0\-RRR',.BOTH.,(#10));`

Stinkfist0:
@aothms Doing this now. Do you have preferences what the command-line argument would be called?

aothms:
Well, maybe we need to think for a second what the meaningful options are, e.g. for the input `J\X2\00E4\X0\tehuone`. I can think of:

- `J\X2\00E4\X0\tehuone` (verbatim IFC)
- `J\u00e4tehuone` (old behaviour)
- `Jätehuone` (current behaviour when compiled with ICU)
- `J_tehuone` (current behaviour when compiled without ICU)
- `Jatehuone` (not sure if this is meaningful, Americans might think so, maybe Finnish people would disagree?) [[1](http://stackoverflow.com/questions/2992066/code-to-strip-diacritical-marks-using-icu#answer-13071166), [2](https://github.com/andrewrk/node-diacritics/blob/master/index.js)]

[...]

PR #179, opened 2017-01-11
IfcConvert enhancements

Stinkfist0:
--precision
--model-offset
--unicode

re: #20 #73 #122


Stinkfist0:
I could also throw in the --model-offset feature to the same pull request soon or provide a new PR with it, whatever suits you the best.

Stinkfist0:
I could also squeeze in `--unicode=escape|utf8` as it's pretty trivial, I should be done with in within an hour.

Stinkfist0:
OK, that should be it for now.

src/ifcconvert/IfcConvert.cpp
```diff
@@ -482,16 +517,25 @@ int main(int argc, char** argv) {
...
-	}
+    if (center_model || model_offset) {
+        double* offset = serializer->settings().offset;
+        if (center_model) {
+            gp_XYZ center = (context_iterator.bounds_min() + context_iterator.bounds_max()) * 0.5;
+            offset[0] = -center.X();
+            offset[1] = -center.Y();
+            offset[2] = -center.Z();
+        } else {
+            if (sscanf(offset_str.c_str(), "%lf;%lf;%lf", &offset[0], &offset[1], &offset[2]) != 3) {
```
aothms: Wonder if we need to do something with `setlocale()` here? I'm somewhat surprised that I need to use a dot as a decimal separator even though my OS is set to believe in the mainland European convention of using commas. It's a minor issue and I'm sure users can figure this out for themselves.

aothms:
Wonderful, just waiting for your opinion on the locale thing and then I will merge it.

Stinkfist0:
Hmm, I'm suspecting we never set the locale explicitly anywhere; the default implicit locale for for any C and C++ programs should be (IIRC)  the C locale.

aothms:
Yeah just I had to get too so much trouble to make 100% sure it is the C locale in the parser [1] that I got a bit allergic to this. But I suppose this is mostly for *nix users messing with `LC_ALL` that supposedly know what they are doing. Great!

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcparse/IfcParse.cpp#L48

PR #187, opened 2017-02-20
Update vs-cfg.cmd

carlesRT:
prepend win/utils path instead of append it.
It should force the build-deps.cmd to call 7za.exe within utils and not any other old-version that users might already have. 
Discussed in https://sourceforge.net/p/ifcopenshell/discussion/1782717/thread/53803197/


aothms:
Thanks!

Issue #104, opened 2016-07-18
Sub-meshes that share material are merged?


Stinkfist0:
Toying around with layer set slicing. I have e.g. a wall of which material specification looks like this in Solibri Model Viewer:
![image](https://cloud.githubusercontent.com/assets/227876/16911628/412f6e2c-4ce8-11e6-9910-86de1d9f8689.png)

However, in the resulted DAE, the mesh has only two sub-meshes and the both outer wall parts (Gypsum Wall Board) are part of the same sub-mesh and share material usage, and I cannot choose different materials for inner and outer wall parts. Is this a bug in the layer set slicing code?

aothms:
I would expect the layers to end up in different meshes. Do you have a small test file you can share?

Stinkfist0:
I'll try to see if I can segregate a small test file. I also try to research a bit more, maybe check if there's difference between OBJ and DAE, etc.

Stinkfist0:
Actually this issue can be reproduced with Duplex_A_20110907.ifc `: IfcConvert Duplex_A_20110907.ifc Duplex_A_20110907-layerset-sliced.dae --enable-layerset-slicing

In Solibri this wall has 7 materials:
![image](https://cloud.githubusercontent.com/assets/227876/16913096/f3a83cde-4cf0-11e6-892f-faab7b11ac8d.png)


But in Unity only 4
![image](https://cloud.githubusercontent.com/assets/227876/16913085/ec2e41c4-4cf0-11e6-952a-d2cb238a5079.png)

aothms:
This appears to be an optimization happening inside unity. While the object indeed has 4 materials assigned to it, the mesh has 7 streams of triangles.

~~~
>>> ifcopenshell.guid.split(ifcopenshell.guid.expand("2O2Fr$t4X7Zf8NOew3FKau"))
'{9808fd7f-dc48-478e-9217-628e833d4938}'

<node id="product-9808fd7f-dc48-478e-9217-628e833d4938-body" name="product-9808fd7f-dc48-478e-9217-628e833d4938-body" type="NODE">
  <matrix>0 1 0 4.400000000000001 -1 0 0 -0.4170000000000139 0 0 1 3.100000000000378 0 0 0 1</matrix>
  <instance_geometry url="#representation-5397">
    <bind_material>
      <technique_common>
        <instance_material symbol="surface-style-3944-plasterboard" target="#surface-style-3944-plasterboard"/>
[...]

PR #188, opened 2017-02-23
IfcConvert and IfcGeomIterator filtering improvements

### LINKED ISSUE #20: IfcConvert enhancements

Stinkfist0:
- possibility to --include and --exclude simultaneously: however, only single --include and --exclude supported for now
- "outsourced" filtering from IfcGeomIterator to any possible client code instead, default implementations for old and new filters provided
- filtering by Description and Tag
- `--<include|exclude> --<enties|layers|guids|names> <value1 value2 ...>` syntax changed to `--<include|exclude>=<entities|layers|arg> (<GlobalId|Name|Description|Tag>) <value1 value2 ...>`
- closes #20 
- there's some room for more cleanup and refactoring, will continue the work after the discussed (on IRC) merging of IfcBaseEntity and IfcLateboundEntity happens and after tackling #104

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,14 +13,14 @@ Prerequisites
 -------------
 * Git
 * CMake (2.6 or newer)
-* Windows: [Visual Studio] 2008 or newer with C++ toolset (or [Visual C++ Build Tools]) or [MSYS2] + MinGW
+* Windows: [Visual Studio] 2008 or newer (2017 not yet officially supported) with C++ toolset (or [Visual C++ Build Tools]) or [MSYS2] + MinGW
 * *nix: GCC 4.7 or newer, or Clang (any version)
 
 Dependencies
 -------------
 * [Boost](http://www.boost.org/)
-* Open Cascade - *optional*, but required for building IfcGeom
-  ([official](http://www.opencascade.org/getocc/download/loadocc/) or [community edition](https://github.com/tpaviot/oce))  
+* [Open Cascade](http://opencascade.org) - *optional*, but required for building IfcGeom
+  ([official](http://www.opencascade.org/getocc/download/loadocc/), "OCCT", or [community edition](https://github.com/tpaviot/oce), "OCE")  
   For converting IFC representation items into BRep solids and tesselated meshes
 * [ICU](http://site.icu-project.org/) - *optional*  
   For handling code pages and Unicode in the parser
@@ -58,6 +58,7 @@ Only) to deploy the headers and binaries into a single location if wanted/needed
 
 Alternatively, one can use the utility batch file(s) to build and install the project easily from the command-line
 (installing a project will build it also, if required):
+
     > install-ifcopenshell.bat
 
 #### Using MSYS2 + MinGW
@@ -84,12 +85,11 @@ meant to be used for a typical IfcOpenShell workspace setup.
 
     $ sudo apt-get install git cmake gcc g++ libboost-all-dev libicu-dev
 
-**2a)** Either use an Open CASCADE package from your operating system's software repository (see http://opencascade.org
-for additional information)
+**2a)** Either use an OCE package from your operating system's software repository
 
     $ sudo apt-get install liboce-foundation-dev liboce-modeling-dev liboce-ocaf-dev liboce-visualization-dev liboce-ocaf-lite-dev
 
-**2b)** or (if not available, or the latest code is wanted) compile Open CASCADE yourself (note that the build takes a long time):
+**2b)** or (if not available, or the latest code is wanted) compile OCE yourself (note that the build takes a long time):
 
     $ sudo apt-get install libftgl-dev libtbb2 libtbb-dev libgl1-mesa-dev libfreetype6-dev
     $ git clone https://github.com/tpaviot/oce.git
@@ -98,6 +98,8 @@ for additional information)
     $ cmake ..
     $ make -j
     $ sudo make install
+
+**2c)** or obtain and compile OCCT from http://www.opencascade.org/getocc/download/loadocc/
 
 **3)** For building IfcConvert with COLLADA (.dae) support (on by default), OpenCOLLADA is needed:
 
--- a/src/ifcconvert/ColladaSerializer.cpp
+++ b/src/ifcconvert/ColladaSerializer.cpp
@@ -338,6 +338,7 @@ void ColladaSerializer::ColladaExporter::endDocument() {
 	geometries.close();
 	for (std::vector<DeferredObject>::const_iterator it = deferreds.begin(); it != deferreds.end(); ++it) {
 		const std::string object_name = it->unique_id;
+        /// @todo redundant information using ID as both ID and Name, maybe omit Name or allow specifying what would be used as the name
 		scene.add(object_name, object_name, it->representation_id, it->material_references, it->matrix);
 	}
 	scene.write();
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -596,7 +596,7 @@ int main(int argc, char** argv)
 	}
 	Logger::Status(msg.str());
 
-    return successful ? 0 : 1;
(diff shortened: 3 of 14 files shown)


Stinkfist0:
Ideally I'd like to see Iterator refactored a bit more. It now reminds a bit Java-style iterators, but not quite there yet. I think something like this would be nice:

No filtering:
```c++
IfcGeom::Iterator it(...);
while(it.has_next()) {
    auto geom = it.next();
    // write geom
}
```

Filtering:
```c++
IfcGeom::Iterator it(...);
while(it.has_next()) {
    auto next_prod = it.peek_next();
    if (filter_matches(next_prod)) {
        it.next();
        auto geom = it.get();
        // write geom
    } else {
        it.skip();
    }
}
```

Stinkfist0:
While still at it, I could also throw in specifying the ID usage of serialized element (Unique ID or any of the currently supported inclusion/exclusion argument names (GlobalId, Name, Description, Tag).

aothms:
While I quite like your suggestion, how about making it more STL-like? (We should also really then consider renaming it to something else than iterator.)

~~~c++
IfcGeom::Iterator model(...);
auto filtered_model = ...;
for(auto it = filtered_model.begin(); it != filtered_model.end(); ++it) {
    auto geom = *it; // <- when dereferenced the geometry is interpreted? Is that too unconventional?
}

// range based for also fine of course

// or

for(auto it = model.begin(); it != model.end(); next_matching(filters, it)) {
    auto geom = *it;
}

// And then perhaps sth like:

ColladaSerializer s(...);
std::copy(model.begin(), model.end(), s.begin());
~~~

Not sure really, just thinking out loud. For example, would it be nice to `std::sort()` the geometries at some point? Or perhaps we can implement the filters by means of `boost::adaptors::filtered`. Or should we implement this triangulation using `std::transform()`? Just a feeling that if we keep calling this iterator it better matches with the conception that C++ programmers have of iterators.

> While still at it, I could also throw in specifying the ID usage of serialized element (Unique ID or any of the currently supported inclusion/exclusion argument names (GlobalId, Name, Description, Tag), https://sourceforge.net/p/ifcopenshell/discussion/1782717/thread/0ff477aa/#70f4

Yeah, sorry, I mixed up things in my head when typing that reply, I implied that that was already being worked on. I'm inclined to say let's first factor out this `IfcLateBoundEntity` before creating more dependencies on it, but maybe that's not really a consideration.

aothms:
I think everything is nicely thought out, but one thing was surprising to me

1. I tried first to filter out internal elements:
`IfcConvert Duplex_A_20110907_optimized.ifc --exclude=arg Name "*Interior*"`
2. This worked well, but left some floating doors from the internal walls, so, I added a `--traverse`. 
traverse
The surprising thing is that this is also applied to the default entity filter and **all** windows and doors disappear as they are considered children of `IfcOpeningElement`.

I don't see an immediately obvious solution to this, besides perhaps applying `--traverse` only to filters specified by users on the command line. A sec `IfcConvert Duplex_A_20110907_optimized.ifc --traverse` also doesn't make a lot of sense to me and the result is also counter intuitive.

Or, how about this, add a plus sign (or asterisks, but then what about shell completion) when filters need to traverse, so: `IfcConvert Duplex_A_20110907_optimized.ifc --exclude=arg+ Name "*Interior*"`? Thoughts? Would that be a major effort to implement?

Stinkfist0:
Your ideas look pretty good too, but perhaps let's continue discussing these in a separate issue. Does this PR look OK to you otherwise?

Stinkfist0:
Yeah, that was surprising also to me, at first I even though it was a bug in my code! Let me ponder for a minute what would be the most convenient way for specifying traversal on a per filter basis.

Stinkfist0:
Maybe `--include+` and `--exclude+` would be convenient to specify traversal on filter basis?

Stinkfist0:
OK! I Should be good to go. I think I'm going to postpone the work with specifying arbitrary element ID for now.

Stinkfist0:
Couple more minor commits incoming if you don't mind.

Stinkfist0:
Oh, and regarding your iterator ideas, those look quite good too, I especially like the look of `for(auto it = model.begin(); it != model.end(); next_matching(filters, it))`, let's discuss these in more detail in  a separate issue.

Stinkfist0:
One more pending thing: I just found out that IfcConvert crashes (null ptr access in TokenFunc::asStringRef) when trying to read an empty IFC file, so I'll see if that's easy to fix quickly.

src/ifcparse/IfcSpfHeader.cpp
```diff
@@ -36,19 +36,22 @@ static const char * const DATA				= "DATA";
 using namespace IfcParse;
 
 void IfcSpfHeader::readSemicolon() {
-	if (!TokenFunc::isOperator(_lexer->Next(), ';')) {
+    Token next = _lexer->Next();
+    if (next.lexer != 0 && !TokenFunc::isOperator(next, ';')) {
```
Stinkfist0: I did not look into this really in-depth as noticed the `IfcSpfLexer* lexer; //TODO: remove it from here` comment. Can be refactored properly when time permits.

aothms: Thanks for this, I had a look as well. This only protects the header parser. I think that in the data section this issue less likely to manifest itself, as there eof is detected and non complete instances are not registered in the map. Still, I think it might be better to resolve it in `asStringRef()` to reduce the likelihood of hard segfaults, by simply throwing an exception there it seems the issue is addressed as well (also, the check on lexer is a bit cryptic, better check on `type == NONE`).

I suggest the following instead at the start of `asStringRef()`:

~~~C++
	if (t.type == Token_NONE) {
		throw IfcParse::IfcException("Null token encountered, premature end of file?");
	}
~~~

Stinkfist0: OK, thanks, will do as suggested.

Stinkfist0:
OK, cleanup up the the fix.

aothms:
Works well, many thanks. There are some small things in the ifcconvert help text that needs updating, I will commit that right way.

PR #201, opened 2017-04-03
Fixes (?) to add_debug_variants

aothms:
@Stinkfist0 I ran into this problem, that if CMake runs against debug libraries of OCCT, that already have a `_d` postfix, it would try to link to `TKerneldd.lib`, i.e. double d.

Perhaps in this particular case, no release libs where available in the dir, so `FIND_LIBRARY(libTKernel NAMES TKernel TKerneld` can only return the `TKerneld.lib`.

Then in [1] (which I believe was added by me after your `add_debug_variants()`) this `_d` is not accounted for.

Somehow I ended making more changes than I anticipated, because I kept ending up with a structure like: `TKernel.lib GENERAL TKernel.lib DEBUG TKernel.lib`. So overcautiously I made sure `add_debug_variants()` does not append to a pre-existing list.

Does this look somewhat reasonable to you?

[1]
~~~
# Use the found libTKernel as a template for all other OCC libraries
foreach(lib ${OPENCASCADE_LIBRARY_NAMES})
    string(REPLACE TKernel "${lib}" lib_path "${libTKernel}")
    list(APPEND OPENCASCADE_LIBRARIES_RELEASE "${lib_path}")
endforeach()
~~~


Stinkfist0:
I'll have a deeper look into this: I also noticed that add_debug_variants() in malfunctioning in general and the Debug-Release switching is broken. I'll dig into this in the VS 2017 branch.

aothms:
@Stinkfist0 that would be great. Just fixed the PR to work on nix.

aothms:
@Stinkfist0 did you manage to have a look?

Stinkfist0:
Sorry, got derailed with this, I'll try to take a look tonight and report back.

aothms:
No worries, not urgent

Stinkfist0:
Unfortunately I didn't have still time to look into this properly, but I dig into the commit history and I think after this commit all the Release/Debug switching problems started: https://github.com/IfcOpenShell/IfcOpenShell/commit/e9bd18fd4bd3019d2c2fcfdd513f456b7d1b370a
IIRC, that commit was the reason I needed to introduce `add_debug_variants()`. Prior to that I think `link_directories()` and `link_libraries()` worked perfectly fine, at least on Windows/MSVC.

Stinkfist0:
OK, finally have the time to properly look into this. Should be done by tonight.

aothms:
> Unfortunately I didn't have still time to look into this properly, but I think the root of the Release/Debug switching problems is the point when usage of link_directories() and link_libraries() was removed.

Well, I agree the amount of custom code to do some simple library linking doesn't feel right. The commit you are referring to is https://github.com/IfcOpenShell/IfcOpenShell/commit/e9bd18fd4bd3019d2c2fcfdd513f456b7d1b370a based on this issue https://github.com/IfcOpenShell/IfcOpenShell/issues/111

The problem being solved there is that if you refer to libs as a pair of `<path, name>` rather than just `path/name` you run into issues if there are libraries installed into (default) library locations (possibly referred to by other deps). I.e. you no longer have full control over what the linker is going to pick when there are multiple versions of one dep installed (e.g. oce from package manager and manually compiled).

Also note the CMake docs recommending against `link_directories()` [1]

> Note that this command is rarely necessary. Library locations returned by find_package() and find_library() are absolute paths. Pass these absolute library file paths directly to the target_link_libraries() command. CMake will ensure the linker finds them.

So, with #217 this can be closed?

[1] https://cmake.org/cmake/help/v3.2/command/link_directories.html

Stinkfist0:
Yeah, upon more thought specifying the full filename of the library feels better than the path-filename pair. Maybe in the future we can cook up a bit nicer and cleaner unified approach for this as the current CMake script is becoming quite convoluted. If the new PR looks good enough (better than this PR), yeah, this can be closed.

Issue #200, opened 2017-03-31
Regressions with IfcOpenHouse and IfcAdvancedHouse examples : ground shape is not generated anymore.


sylvainmarie:
Commit [aebb9e3](https://github.com/IfcOpenShell/IfcOpenShell/commit/aebb9e324b0128503dde95bf9eb166c0ff9fb470#diff-6017059461ea51ae44daa8a8df4b6a42L268) includes a regression, causing a missing ground shape in IfcOpenHouse and IfcAdvancedHouse examples. 

Fixing that regression is easy, but commit [e8ae491](https://github.com/IfcOpenShell/IfcOpenShell/commit/00215867c06269b28a11acffb32baa730812a728) introduces another regression (causing the same problem) by shifting enum indices without updating ifcexpressparser/mapping.py (and others ??) and without regenerating the IFC headers using the EXPRESS parser.

More details on the [sourceforge forum](https://sourceforge.net/p/ifcopenshell/discussion/1782718/thread/337b2b46/).

aothms:
Many thanks for your detective work. The issue indeed was related to a misalignment between the enumeration value, not with the python code, but with the `boost::variant` container index for the assignable attributes. Ironically I didn't forget about this, but somehow I have only checked it in into the pull request I was somewhat simultaneously working on [1].

What happened is that when adding the `ground_representation` to the file, the `IfcProductDefinitionShape.Representations` was not recognized as a aggregate of entity instances (due to the misalignment), and therefore were not traversed and added to the file recursively.

[1] https://github.com/IfcOpenShell/IfcOpenShell/pull/190/files#diff-b74cff1eb78dd804ddc8227d13b2bcb8

FragColor:
Thanks for your response and your fix :)

However, I still have a question in mind : why don't you update `mapping.py` and generate IFC headers ? especially these lines :

```cpp
    supported_argument_types = set([
        'INT', 'BOOL', 'DOUBLE', 'STRING', 'BINARY', 'ENUMERATION', 'ENTITY_INSTANCE',
        'AGGREGATE_OF_INT', 'AGGREGATE_OF_DOUBLE', 'AGGREGATE_OF_STRING', 'AGGREGATE_OF_BINARY', 'AGGREGATE_OF_ENTITY_INSTANCE',
[...]

PR #205, opened 2017-04-07
Fix generation of IfcOpenHouse.ifc

FragColor:
Fix generation of IfcOpenHouse.ifc which doesn't have its ground representation. See #200


aothms:
Thanks! I checked, I thought it wasn't necessary to have this in, as the representation is reachable by means of the site and would be added on traversal. Doesn't hurt though, I will merge it in.

FragColor:
Unfortunately, it was necessary because it didn't work without it ! Moreover, IfcAdvancedHouse.cpp uses the same codeline to do exactly the same thing.

PR #210, opened 2017-04-25
Limit retries for boolean op on OCCT 6.9 and higher

aothms:
@Stinkfist0 can you have a look re your messages on IRC, seems like some oversight indeed that was not there for the OCCT < 6.9 specific code. Thanks.


Stinkfist0:
Great, thanks. Indeed I had start to experience stack overflow exceptions due to this code path when using newer OCE/OCCT. Tried with OCE 0.17.2 last Friday and it did not have this problem.

PR #211, opened 2017-04-25
Represent Ifc files hierarchy in Collada files

Skyndia:
# Represent Ifc files hierarchy in Collada files

## Context

We are working on an application based on the Unity game engine. In this application we have to display and manipulate IFC files.
In this purpose, we use IfcConvert to convert IFC files into Collada 3D files to automatically import them in the engine. To be able to apply some operations or transformations, we need to access the IFC hierarchy and identify each element type in the Collada file.

## Objectives
- [x] 1 - Retrieve the type of each element and put it in the element name
- [x] 2 - Organise the element by floors in the output files 
- [x] 3 - Retrieve additional information about the type (for example an IfcSlab can be a .ROOF. or a .FLOOR. )
- [x] 4 - Retrieve the full hierarchy (like in the Xml serializer)

### 1 - Retrieve the type of each element and put it in the element name
related option : --use-element-types

We added an option to use the type of an element as name, and handled the naming like it was done previously. Here is the main code : 

https://github.com/mitmadness/IfcOpenShell/blob/feature/hierarchy/src/ifcconvert/ColladaSerializer.cpp#L515-L517

Status : Completed.

### 2 - Organise the element by floors in the output files 
related option : --use-element-hierarchy

Outdated : We are now retrieving the entire hierarchy.
Status : Completed.

### 3 - Retrieve additional information about the type 
related option : --use-element-types

We want to know if for example an IfcSlab element is part of the floor or part of the roof. 
We simply add this information at the end of the element name. 

See the code here :

 https://github.com/mitmadness/IfcOpenShell/blob/feature/hierarchy/src/ifcconvert/ColladaSerializer.cpp#L538-L562

https://github.com/mitmadness/IfcOpenShell/blob/feature/hierarchy/src/ifcconvert/ColladaSerializer.cpp#L517

Status : Completed.

### 4 - Retrieve the full hierarchy
related option : --use-element-hierarchy

- For each element to display, retrieve a list of parents : 

https://github.com/mitmadness/IfcOpenShell/blob/feature/hierarchy/src/ifcgeom/IfcGeomIterator.h#L578-L623

- Store this list in an Element property, and then in a DeferredObject property. Sort the elements using the parents lists : 

https://github.com/mitmadness/IfcOpenShell/blob/feature/hierarchy/src/ifcconvert/ColladaSerializer.cpp#L575

https://github.com/mitmadness/IfcOpenShell/blob/feature/hierarchy/src/ifcconvert/ColladaSerializer.h#L137-L151

https://github.com/mitmadness/IfcOpenShell/blob/feature/hierarchy/src/ifcgeom/IfcGeomElement.h#L93-L109

- Finally update a parent stack while creating the collada nodes : 

https://github.com/mitmadness/IfcOpenShell/blob/feature/hierarchy/src/ifcconvert/ColladaSerializer.cpp#L592-L632

Status : Completed.

### Before

before

### After

after


aothms:
Hi, this is amazing, many thanks. It'll take me a couple more days until I can dive into this in detail, but the before-after pictures certainly look promising. Are the two remaining items intended to be added to this PR? Is there sth we can do help, or is this finalized now?

Skyndia:
Hi.

In our opinion the PR is mergeable as it is now. But we are opened to any suggestion in order to improve the implementation in the workflow.

The first two items were our main objectives to be able to process the IFC in Unity.

About the third item, with the last commit we can already identify the different types of IfcSlab. We don’t need any more information for the moment, but it may change in the future.

About the last item, we think that it can be really work intensive. In fact in order to retrieve the full hierarchy (all the different layers) we think that a modification of the whole workflow of the collada serializer is necessary.

The current state of the project allow us to continue working on our application but we may need to work on the last two points in the near future to be able to improve our processing workflow.


P.S: Our last commit changed the lines selected in the github link, but it is repaired.

Stinkfist0:
My two cents: the commit history looks really messy, but luckily the changes itself look rather clean. However, I'm not a fan of adding naming mechanic that's based on arbitrary attribute (PredefinedType) of arbitrary element (IfcSlab) - wouldn't it make more sense to use this naming e.g. on each element that has PredefinedType (there should several of these, IfcCovering most notably at least)? Or in the best case, as discussed with Thomas, IIRC, the user could specify the naming schema instead of providing `--use-element-guids`, `--use-element-names`, etc. we could have e.g. `--name-elements=IfcRoot.GlobalId` (for the `--use-element-guids` behavior).

aothms:
Thanks for having a look @Stinkfist0 . I still didn't manage to look into it yet, but hopefully today. I can imagine the line of thought to have PredefinedType influence the naming, especially for slabs, as it can differentiate between floors and roofs. But indeed preferably using a generic and reusable approach (and maybe not necessarily as part of this PR).

IFC4 adds predefined types for nearly all building elements (and unfortunately all in leaf entities, there is not a common ancestor). The way IfcOpenShell is structured with naked C++ enums for the EXPRESS enumerations, makes it somewhat difficult to deal with this generically.

aothms:
Quite neat, I like most of the approach, but I would argue that with the 4th point included, results would be a bit more complete and predictable, especially if there (for example) are multiple `IfcBuilding`s in the file.

Just thinking out loud here, but would it make sense to give `DeferredObject` a `std::vector<IfcGeom::Element*>` with the `parent_ids()` instead of the proposed `parent`? Similar to the current approach the objects can then be sorted, but using a `std::lexicographical_compare()` on their parents vectors. Upon serialization you would need to keep a stack of `COLLADASW::Node`s instead of just `current_node`, but I think overall the changes would be rather minimal. It would make this `SetFloor()` and `storey()` obsolete, which are hard to generalize to a full solution for all hierarchy nodes later on.

Anyway, many thanks again for your work. I can finish this up if my proposal above is too elaborate to your liking.

PS: Something I need to check, but if you add a matrix to the parent nodes, doesn't its inverse needs to be multiplied into the child nodes? Not sure how that works in collada.

Stinkfist0:
Also, why are only 3 out of 6 enum values (which don't handle by far most the common value, at least to my experience, FLOOR) handled while converting the PredefinedType into string?

Skyndia:
Hi,

The 3 last commits were added to the PR by mistake, there still are errors as you noticed Stinkfist0, sorry for that. I'm going to fix them. 

Using the elevation to sort the ifcBuildingStorey if set would be nice I will take a look. 
About the 4th point, your approach is interesting! I have to estimate again the time it could take to implement that, and then I will discuss with my colleagues about our needs. 
I will let you know in a few days if I will be able to work on this topic.

src/ifcconvert/ColladaSerializer.cpp
```diff
@@ -301,33 +340,67 @@ void ColladaSerializer::ColladaExporter::startDocument(const std::string& unit_n
...
+	IfcSlab* slab = (IfcSlab*)o->product();
+	switch (slab->PredefinedType()){
+			case (IfcSlabTypeEnum::IfcSlabType_ROOF):
+				return "_Roof";
+				break;
+			case (IfcSlabTypeEnum::IfcSlabType_LANDING):
+				return "_Landing";
+				break;
+			case (IfcSlabTypeEnum::IfcSlabType_BASESLAB):
+				return "_BasesLab";
```
Skyndia: true

src/ifcconvert/ColladaSerializer.cpp
```diff
@@ -373,20 +373,30 @@ void ColladaSerializer::ColladaExporter::write(const IfcGeom::TriangulationEleme
...
+			return "_Roof";
+			break;
+		case (IfcSlabTypeEnum::IfcSlabType_LANDING):
+			return "_Landing";
+			break;
+		case (IfcSlabTypeEnum::IfcSlabType_BASESLAB):
+			return "_BaseSlab";
+			break;
+		case (IfcSlabTypeEnum::IfcSlabType_USERDEFINED):
+			return "_" + slab->ObjectType();
```
aothms: `ObjectType` is an optional attribute so you'll need to call `HasObjectType()` first. Perhaps this case can be moved to the bottom and fall through to the default in case ObjectType is Null.

aothms:
> The 3 last commits were added to the PR by mistake

These are the commits `ced9d8c`, `93ece6b`, `ff64ce1`?

> I will let you know in a few days if I will be able to work on this topic.

Any updates? Again, I don't mind wrapping this up myself.

Skyndia:
I mean 1d8f6ad, 78b904d and e3e4fba, where Stinkfist0 found some problems. It wasn't tested yet.

Finally I'm working on it. I have implemented most of the workflow. I just have a problem with the element positions, as you said I probably need to perform some matrix calculation. 

I hope that I will be able to finish and test this today or tomorrow.

Skyndia:
I'm done with the implementation and waiting for your feedback. 
I have edited my first comment

aothms:
Hi, really wonderful. I'll make some changes to it over time, but in general and especially the functionality is a really great contribution.

While I appreciate your effort not to introduce another dependency. I'll switch the ublas matrix multiplication for Open Cascade. Open Cascade can invert these matrices more efficiently as they are tagged (i.e. it is known when a matrix is identity/translation) so you might not need a LU factorization, elimination or cofactor stuff. FYI, you don't need to worry about these matrices being singular, they are created from axis2_placements [1].

I agree with @Stinkfist0 next time maybe try to keep the commit history a bit tidier. Normally I can squash the commits, but I'd rather keep them as is, especially since multiple people apparently worked on this.

There still seems to be an open issue with the sorting. Elements with geometry are not necessarily leaf nodes (e.g. IfcSite, IfcWall with windows...). But since elements in the vector of parents are not compared to the objects them selves this may result into duplicates. These DeferredObjects stem from a time when IfcGeom::TriangulationElement were not copy-constructable. We should probably look into storing them in a vector directly and then benefit from the fact that IfcGeom::TriangulationElement is a subclass of IfcGeom::Element and the same ordering predicate can be applied to them. See image below where both DAE nodes represent the same elements, once as a IfcGeom::Element and once as a DeferredObject.

dae-same-elements

[1] http://www.buildingsmart-tech.org/ifc/IFC2x3/TC1/html/ifcgeometryresource/lexical/ifcaxis2placement3d.htm

Skyndia:
Hi.
In fact we still have this problem, I haven't noticed it, but I won't have the time to fix for the moment. 
I found another little issue here :
 https://github.com/mitmadness/IfcOpenShell/blob/feature/hierarchy/src/ifcgeom/IfcGeomElement.h#L105
I got to work with a file where all the IfcBuildingStorey have an elevation set to 0. Here we just have to take this case into account 
capture

ani-rudh:
Hi guys,
this looks like an amazing project. I have infact been involved in a similar project to import IFC files to UNITY for VR visualizations. I have been able to accomplish this by using XML, both for geometry and object meta-data (properties, quantities).
Anyway, I have never used ifcOpenShell before,but, it looks promising to do it in a simpler manner. My only issue is that I am not good at c++, maybe I have to look into python wrapper. 
Is it possible to some basic documentation and sample code to get started with?

thorade:
This functionality would be very interesting for us, so I compiled the master branch at ee16deabe57ade22e8706d1eb9a37c85c02f2559, but whenever I try to use `--use-element-hierarchy` IfcConvert.exe just crashes. Using `--use-element-types` works and I can see the names in Unity afterwards.
Any ideas? 

Here is the IFC file that I used for testing.
[7_7_17_OneZone_19.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/1302673/7_7_17_OneZone_19.zip)

aothms:
That appears to be my bad. When adding log statements something wrong appears to have been copy pasted.

https://github.com/IfcOpenShell/IfcOpenShell/commit/aa0a6aa5d13ceb8bf14f45b608472164f17d87ea?diff=split#diff-4a1897ad94043892537a522bdd7fafd5L593

Fix: 8c26037f84316355e88fd8a666f8824e88fcea76

thorade:
Thanks for that quick fix! I just compiled the master at that commit and now ifcconvert seems to be working nicely.

Stinkfist0:
@Skyndia could you check out #397

PR #216, opened 2017-05-11
Authenticated proxy support for DownloadFile

arzezniczak:
This will allow to download required files even if user is behind authenticated proxy and has proper proxy setup in system settings.


aothms:
Thanks! Didn't test this with any proxy enabled, but assume this to work.

PR #217, opened 2017-05-12
MSVC Debug/Release CMake fixes

Stinkfist0:
Supersedes #201, thanks for the pointers.


aothms:
Doesn't seem to work on Linux:

    make[2]: *** No rule to make target `/usr/lib/x86_64-linux-gnu/TKernel', needed by `examples/IfcOpenHouse'.  Stop.

Perhaps you need to include `get_filename_component(OCE_LIB_EXT ${libTKernel} EXT)`?

Stinkfist0:
Hmm, weird. Maybe simply better I set up the Windows-Ubuntu dev env to test this properly instead of trying to fix bindly.

Stinkfist0:
Ah, right, missing the lib-prefix for the filenames.

Stinkfist0:
Should be good to go.

aothms:
Great, many many thanks as always!

PR #222, opened 2017-05-23
IfcConvert --filter-file & --no-progress

Stinkfist0:
I found myself often crafting the filtering commands in notepad and copy-pasting them to command-line so I thought why not offer a support for reading them from a file.  The second new option can be very useful when directing the output of IfcConvert to a log file.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -155,7 +155,8 @@ int main(int argc, char** argv)
 		("help,h", "display usage information")
 		("version", "display version information")
         ("verbose,v", "more verbose output")
-        ("yes,y", "answer 'yes' automatically to possible confirmation queries (e.g. overwriting an existing output file)");
+        ("yes,y", "answer 'yes' automatically to possible confirmation queries (e.g. overwriting an existing output file)")
+        ("no-progress", "Suppress possible progress bar type of prints that use carriage return.");
 
     po::options_description fileio_options;
 	fileio_options.add_options()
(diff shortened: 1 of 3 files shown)


aothms:
I agree that writing to stdout there is an awkward choice. But it could already be toggled with an (obscure) existing method [1]. I.e. `if (...) { Logger::SetOutput(NULL, &log_stream); }` at [2]. Not saying it is particularly elegant that way, but no need for these boolean flags either.

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcparse/IfcLogger.cpp#L34
[2] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcconvert/IfcConvert.cpp#L410

Stinkfist0:
Thanks for the feedback, I'll rework the solution as per your suggestion. Does `--filter-file` look OK?

aothms:
> Does --filter-file look OK?

Looks ok to me. I can imagine sth like a reusable configuration would come in handy if you need work with IfcConvert a lot. Ideally I think these commands would use the same flag, sth like (purely hypothetical) `--filter "include(entity(IfcWall) or attribute(ObjectType=seinä))"` and `--filter @my_file.txt`. Especially if you can apply these filter flags to IFC output files as well, i.e get subsets of IFC models, it gets even more useful perhaps. Anyway, just random thoughts.

Stinkfist0:
Yeah, I've toyed around with similar ideas in my head. Also being able to specify multiple output files, e.g. one per representation layer or a storey, in a single go would be really nice at some point.

Stinkfist0:
Suppressing IfcFile::Init() prints done suggested.

aothms:
Thanks, sorry for the delay

Issue #230, opened 2017-06-14
IfcConvert with building local coordinates


andersgb:
I am using IfcConvert to convert from IFC to OBJ. My IFC files typically have very large X and Y values for the insertion point (the ObjectPlacement) of the `IfcSite` and the axis directions are typically not at right angles with the building walls.

Using `--center-model` or `--model-offset` I can put the OBJ vertices closer to origo, but the orientation will still be according to the global axes.

I essentially want all geometry to be converted with relative placements up to the `IfcBuilding` level, but skip the `IfcSite` level. 

Is this achievable with IfcConvert?

A workaround would be to modify the IFC file by overriding the ObjectPlacement of the `IfcSite` to origo/identity, I guess.

aothms:
> A workaround would be to modify the IFC file by overriding the ObjectPlacement of the IfcSite to origo/identity, I guess.

I guess that's the most viable approach. It's fairly trivial with a small python script. E.g. something like below:

~~~py
import ifcopenshell
f = ifcopenshell.open(...)

for site in f.by_type("IfcSite"):
    if site.ObjectPlacement is not None and site.ObjectPlacement.is_a("IfcLocalPlacement"):
        site.ObjectPlacement.RelativePlacement = f.createIfcAxis2Placement3D(f.createIfcCartesianPoint((0.,0.,0.)))

f.write(...)
~~~

I think it's a fairly common problem though, quite a few authoring tools seem to insert arbitrarily large offsets at some point in the hierarchy, if you want to work on this personally I don't mind a pull request for IfcConvert in C++...

andersgb:
This is my attempt at scripting it in python (I had to edit the fields in-place, if not it doesn't affect the relative placement hierarchy already existing):
```python
from __future__ import print_function
import ifcopenshell
import argparse


def ifc_site_at_origo(ifc_in, ifc_out):
    print("Opening {}...".format(ifc_in))
    ifc = ifcopenshell.open(ifc_in)
    site = ifc.by_type('IfcSite')[0]

    print("Modifying ObjectPlacement...")
    placement = site.ObjectPlacement.RelativePlacement
    placement.Axis.DirectionRatios = (0., 0., 1.) # Z
    placement.RefDirection.DirectionRatios = (1., 0., 0.) # X
    placement.Location.Coordinates = (0., 0., 0.) # origo

    print("Writing new file {}...".format(ifc_out))
    ifc.write(ifc_out)


if __name__ == "__main__":
[...]

PR #240, opened 2017-07-16
Site local placement

andersgb:
The problem:
```
Some IFC authoring tools may put the global origo at a large distance
away from the actual building geometry. Also, the global X- and Y-axes
may be at arbitrary angles with natural building axes. Using
--center-model or --model-offset may place the converted geometry closer
to origo, but will still use the global IFC definition of axis
directions. Also, in a scenario with multiple IFC files for one
construction site, using --center-model on each of them would not
generate consistent geometry across OBJ or DAE files.
```
See also discussion in issue #230 

This is an initial attempt at supporting site local placement in `IfcConvert`. It should be working now for OBJ conversion, but it's not completed and should not be merged as-is. Since I don't really know this codebase I need some feedback here. Questions:
- Does the generalization of `offset` -> `transform` in `SerializerSettings` make sense? It feels a bit hacky to first generate geometry in global coordinates, then partly reverse it by essentially applying `IfcSite.ObjectPlacement.inverse` to all geometry before writeout. But I tried doing it this way in order to keep the feature implemented mostly inside `src/ifconvert`.
- Is there a way to avoid exposing `IfcGeom::Iterator.kernel` through the public `getKernel` method? I.e. can we find the `ObjectPlacement` of `IfcSite` in a better way?
- Is `--site-local-placement` a good name for this? `--building-local-placement`? `--strip-site-coords`?
- I need some help on the DAE conversion. I'm not even compiling with COLLADA support currently and the last commit is very much work-in-progress. The unit conversion stuff included in `IfcGeom::Transformation` and `IfcGeom::Matrix` may complicate things, not sure.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -293,6 +293,9 @@ int main(int argc, char** argv)
             "all placements as an offset. Applicable for OBJ and DAE output.")
         ("model-offset", po::value<std::string>(&offset_str),
             "Applies an arbitrary offset of form 'x;y;z' to all placements. Applicable for OBJ and DAE output.")
+		("site-local-placement",
+			"Place elements locally in the IfcSite coordinate system, instead of placing "
+			"them in the IFC global coords. Applicable for OBJ and DAE output.")
         ("precision", po::value<short>(&precision)->default_value(SerializerSettings::DEFAULT_PRECISION),
             "Sets the precision to be used to format floating-point values, 15 by default. "
             "Use a negative value to use the system's default precision (should be 6 typically). "
@@ -364,6 +367,7 @@ int main(int argc, char** argv)
     const bool no_normals = vmap.count("no-normals") != 0 ;
     const bool center_model = vmap.count("center-model") != 0 ;
     const bool model_offset = vmap.count("model-offset") != 0 ;
+	const bool site_local_placement = vmap.count("site-local-placement") != 0 ;
     const bool generate_uvs = vmap.count("generate-uvs") != 0 ;
 
 #ifdef HAVE_ICU
@@ -496,6 +500,7 @@ int main(int argc, char** argv)
     settings.set(IfcGeom::IteratorSettings::NO_NORMALS, no_normals);
     settings.set(IfcGeom::IteratorSettings::GENERATE_UVS, generate_uvs);
 	settings.set(IfcGeom::IteratorSettings::SEARCH_FLOOR, use_element_hierarchy);
+	settings.set(IfcGeom::IteratorSettings::SITE_LOCAL_PLACEMENT, site_local_placement);
 
     settings.set(SerializerSettings::USE_ELEMENT_NAMES, use_element_names);
     settings.set(SerializerSettings::USE_ELEMENT_GUIDS, use_element_guids);
@@ -602,6 +607,11 @@ int main(int argc, char** argv)
     if (center_model || model_offset) {
         double* offset = serializer->settings().offset;
         if (center_model) {
+			if (site_local_placement) {
+				Logger::Error("Cannot use --center-model together with --site-local-placement");
+				delete serializer;
+				return EXIT_FAILURE;
+			}
             gp_XYZ center = (context_iterator.bounds_min() + context_iterator.bounds_max()) * 0.5;
             offset[0] = -center.X();
             offset[1] = -center.Y();
--- a/src/ifcgeom/IfcGeomIterator.h
+++ b/src/ifcgeom/IfcGeomIterator.h
@@ -755,6 +755,9 @@ namespace IfcGeom {
             kernel.setValue(IfcGeom::Kernel::GV_MAX_FACES_TO_SEW, settings.get(IteratorSettings::SEW_SHELLS) ? 1000 : -1);
             kernel.setValue(IfcGeom::Kernel::GV_DIMENSIONALITY, (settings.get(IteratorSettings::INCLUDE_CURVES)
                 ? (settings.get(IteratorSettings::EXCLUDE_SOLIDS_AND_SURFACES) ? -1. : 0.) : +1.));
+			if (settings.get(IteratorSettings::SITE_LOCAL_PLACEMENT)) {
+				kernel.set_conversion_placement_rel_to(IfcSchema::Type::IfcSite);
+			}
 		}
 
 		bool owns_ifc_file;
--- a/src/ifcgeom/IfcGeom.h
+++ b/src/ifcgeom/IfcGeom.h
@@ -99,6 +99,10 @@ private:
 	std::map<int, SurfaceStyle> style_cache;
 
 	const SurfaceStyle* internalize_surface_style(const std::pair<IfcSchema::IfcSurfaceStyle*, IfcSchema::IfcSurfaceStyleShading*>& shading_style);
+
+	 // For stopping PlacementRelTo recursion in convert(const IfcSchema::IfcObjectPlacement* l, gp_Trsf& trsf)
+	IfcSchema::Type::Enum placement_rel_to;
+
 public:
 	Kernel()
 		: deflection_tolerance(0.001)
@@ -109,6 +113,7 @@ public:
 		, ifc_planeangle_unit(-1.0)
 		, modelling_precision(0.00001)
 		, dimensionality(1.)
+		, placement_rel_to(IfcSchema::Type::UNDEFINED)
 	{}
 
 	Kernel(const Kernel& other) {
@@ -312,6 +317,8 @@ public:
 		cache = Cache(); 
 #endif
 	}
+
+	void set_conversion_placement_rel_to(IfcSchema::Type::Enum type);
 
 #include "IfcRegisterGeomHeader.h"
 
--- a/src/ifcgeom/IfcGeomHelpers.cpp
+++ b/src/ifcgeom/IfcGeomHelpers.cpp
@@ -357,6 +357,10 @@ bool IfcGeom::Kernel::convert(const IfcSchema::IfcAxis2Placement2D* l, gp_Trsf2d
 	return true;
 }
 
+void IfcGeom::Kernel::set_conversion_placement_rel_to(IfcSchema::Type::Enum type) {
+	placement_rel_to = type;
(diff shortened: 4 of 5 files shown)


src/ifcconvert/IfcConvert.cpp
```diff
@@ -589,30 +593,62 @@ int main(int argc, char** argv)
...
-    if (center_model || model_offset) {
-        double* offset = serializer->settings().offset;
+        std::stringstream msg;
         if (center_model) {
             gp_XYZ center = (context_iterator.bounds_min() + context_iterator.bounds_max()) * 0.5;
-            offset[0] = -center.X();
-            offset[1] = -center.Y();
-            offset[2] = -center.Z();
-        } else {
+			serializer->settings().transform.SetTranslation(center);
```
aothms: Appreciate how you made this work for the translation case as well 👍

src/ifcconvert/IfcConvert.cpp
```diff
@@ -589,30 +593,62 @@ int main(int argc, char** argv)
...
                 return EXIT_FAILURE;
             }
-        }
-
-        std::stringstream msg;
-        msg << "Using model offset (" << offset[0] << "," << offset[1] << "," << offset[2] << ")";
+			msg << "Using model offset (" << offset[0] << "," << offset[1] << "," << offset[2] << ")";
+			serializer->settings().transform.SetTranslation(gp_XYZ(offset[0], offset[1], offset[2]));
+        } else {
+			IfcParse::IfcFile *ifc_file = context_iterator.getFile();
```
aothms: Perhaps this can be moved to a method on `IfcGeom::Iterator`? Cleans up this file a bit and removes the need to expose the `getKernel()` method.

src/ifcgeom/IfcGeomIterator.h
```diff
@@ -317,6 +317,8 @@ namespace IfcGeom {
 
 		IfcParse::IfcFile* getFile() const { return ifc_file; }
 
+		Kernel* getKernel() { return &kernel; }
```
aothms: The `getKernel()` method breaks the SWIG Python wrapper, because SWIG is not aware of the `Kernel` type. If we need to keep it, you can ignore the function prototype in the interface file [1].

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcwrap/IfcGeomWrapper.i#L117

~~~
/home/travis/build/IfcOpenShell/IfcOpenShell/cmake/build-ifc2x3/ifcwrap/IfcPythonPYTHON_wrap.cxx: In function ‘PyObject* _wrap_iterator_single_precision_getKernel(PyObject*, PyObject*)’:

/home/travis/build/IfcOpenShell/IfcOpenShell/cmake/build-ifc2x3/ifcwrap/IfcPythonPYTHON_wrap.cxx:6242:3: error: ‘Kernel’ was not declared in this scope

   Kernel *result = 0 ;
~~~

src/ifcconvert/ColladaSerializer.cpp
```diff
@@ -213,11 +213,27 @@ void ColladaSerializer::ColladaExporter::ColladaScene::add(
...
 		{ 0, 0, 0, 1 }
 	};
-	
-	/// @todo: TFK: Rather than applying this offset to all leafs (which might be undesirable) should this offset be applied to a node higher up in the hierarchy?
-    matrix_array[0][3] += serializer->settings().offset[0];
-    matrix_array[1][3] += serializer->settings().offset[1];
-    matrix_array[2][3] += serializer->settings().offset[2];
+
+	gp_Trsf pos_trsf;
+	pos_trsf.SetValues(matrix_array[0][0], matrix_array[0][1], matrix_array[0][2], matrix_array[0][3],
```
aothms: The prototype of `SetValues()` depends on the OCCT version. See [1].

I wonder if it is necessary to reconstruct a `gp_Trsf` manually. Can't we pick up the `gp_Trsf` from `transformation_towrite->data()`?

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/8ce4bbdf6194649cbaca4dff2ec1f53342cf5d39/src/ifcgeom/IfcGeomShapes.cpp#L729

aothms:
Hi,

Wow thanks! This is a pretty good start. I think it has been a good idea to keep the changes contained within IfcConvert and not 'pollute' the conversion functions IfcGeom namespace with this. The dirtier hack would have been to introduce something like (in pseudo-code) `if (ignore_site_placement && placement->PlacesObject()->is<IfcSchema::IfcSite>()) { /* return identity */ }` here [1]. However, the PlacesObject inverse relationship has been changed into a set with unbounded length in IFC4. Therefore placements might no longer uniquely affect a site.

Some thoughts:

* One concern would be the potential precision loss of multiplying its inverse rather than simply not including the site placement in the calculation. With double floating point precision the translation probably needs to be very large, like 1e12 (?), for this effect to be noticeable.
* Another concern is that you cannot generalize this approach to other placements in the hierarchy because they are not on the 'outside' of the equation. `B^-1 A B C` is not `A C`. Not sure whether this functionality also makes sense to apply to the building placement for example.
* Lastly, as you also mention this approach only works for files containing a single IfcSite, which is probably fine, you would end up with potentially 

None of these concerns are dealbreakers for me, but we should probably be aware of them.

An alternative approach would be to no longer convert the full ObjectPlacement in its entirety, but unroll the multiplication loop [2] when the element geometry is converted [3] when on a per element basis can be decided which placements in the hierarchy should be included. There are downsides to this approach as well. It would leave the serializer code more unaffected though. I'll leave it up to you whether to investigate this approach.

I made some inline remarks. I think those are the only issues that cause the build to fail currently.

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcgeom/IfcGeomHelpers.cpp#L361
[2] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcgeom/IfcGeomHelpers.cpp#L367
[3] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcgeom/IfcGeomFunctions.cpp#L1143

Kind regards,
Thomas

andersgb:
Hi,
Thanks for your comments. I force-pushed a new version of the branch now, where I tried something more along the lines of your alternative suggestion, stopping the `PlacementRelTo` recursion when reading the geometry. This is approach is certainly less code, and without any need to dabble inside the OBJ and DAE specific classes.

As you mention, this has the advantage of avoiding round-off errors. I like this approach better, but it's a bit scary to change logic inside `IfcGeom::Kernel::create_brep_for_representation_and_product`, especially since there are quite a few different users of the `convert` function using `ObjectPlacement`. Do you see any potential errors here? At least the logic should be the unchanged as long as `SITE_LOCAL_PLACEMENT` is disabled.

Instead of creating a `convert_placement_rel_to_site` I tried generalizing this by stopping at first occurence of a certain `IfcSchema::Type`. Does that make sense? Still, we can't "skip" an ObjectPlacement and continue at the parent, but we should be able to follow the same logic if we want to stop at another level of (e.g. `C D` instead of `A B C D`). That should enable conversion relative to `IfcBuilding` or anything else.

aothms:
> especially since there are quite a few different users of the convert function using ObjectPlacement. Do you see any potential errors here? 

Excellent point, didn't think about that. This will cause problems with at least openings (possibly layersets). Did a quick grep and found some other uses of `ObjectPlacement()` listed below. 

Since they all just call `IfcGeom::Kernel::convert()`, what do you think of making the `IfcSchema::Type::Enum` to `convert_placement_rel_to()` a member of `Kernel` instead, that is set somewhere (either in constructor, or elsewhere, can be default initialized to `IfcSchema::Type::UNDEFINED`) and then use one general method that converts relatively based on the value of this member?

Looks like all uses of `IfcGeom::Kernel::convert(IfcObjectPlacement*)` will need to use the updated conversion.

> I tried generalizing this by stopping at first occurence of a certain IfcSchema::Type. Does that make sense?

I think so :)

---

**convert_openings()**
~~~
IfcGeomFunctions.cpp:269:			if (fes->hasObjectPlacement()) {
IfcGeomFunctions.cpp:271:					convert(fes->ObjectPlacement(),opening_trsf);
IfcGeomFunctions.cpp:438:			if (fes->hasObjectPlacement()) {
IfcGeomFunctions.cpp:440:					convert(fes->ObjectPlacement(),opening_trsf);
IfcGeomFunctions.cpp:513:			if (fes->hasObjectPlacement()) {
IfcGeomFunctions.cpp:515:					convert(fes->ObjectPlacement(),opening_trsf);
~~~
**create_brep()**
~~~
IfcGeomFunctions.cpp:1143:		convert(product->ObjectPlacement(),trsf);
IfcGeomFunctions.cpp:1235:		convert(product->ObjectPlacement(),trsf);
~~~
**fold_layers()**
~~~
IfcGeomFunctions.cpp:1701:	if (!convert(wall->ObjectPlacement(), local)) {
IfcGeomFunctions.cpp:1757:			if (!convert(it->second->ObjectPlacement(), other)) {
IfcGeomFunctions.cpp:1792:		if (!convert(other_wall->ObjectPlacement(), other)) {
~~~
**bounds_min_, bounds_max_**
~~~
IfcGeomIterator.h:285:                if (product->hasObjectPlacement()) {
IfcGeomIterator.h:290:						success = kernel.convert(product->ObjectPlacement(), trsf);
~~~
**getObject()**
~~~
IfcGeomIterator.h:655:						kernel.convert(ifc_product->ObjectPlacement(), trsf);
~~~

---

This is what duplex looks now when I introduced a rotation at the IfcSite and converted with site-local. You can see that the openings are in the original location and geometry for mapped items is not always in the correct location (due to `create_brep_for_processed_representation()`).

duplex_site_local

andersgb:
Summer and other stuff came in the way here, will update in a week or two 👍

aothms:
> Summer and other stuff came in the way here, will update in a week or two 👍

No worries. Thanks for the headsup and have a good summer!

andersgb:
Force pushed a new revision now. Introduced `placement_rel_to` as a member of `Kernel `, such that all users of `convert()` get the same result. Works on a single example case on my side. Should work with openings etc now as well?

aothms:
Looks good to me :) many thanks for your continued effort.

Issue #22, opened 2016-01-13
Using mmap 


jetablejfc:
Hi,
  I changed code in order to use mmap for file handling under linux (not tested under windows).
This should help when communicating with outside programs (bimserver for example).

I'll post the patch later ...

Regards.

jetablejfc:
The following patch should use an mmap'ed file as input  in IfcParse. It replaces the  FILE* stream  used in IfcSpfStream.h by  boost::iostreams::mapped_file_source .....
[IfcOpenShell_mmap_patch.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/89736/IfcOpenShell_mmap_patch.txt)

jetablejfc:
A file mmapped in two processes is shared between the process (less memory is used). See for example the following code in java ....

import java.io.File;
import java.io.RandomAccessFile;
import java.nio.MappedByteBuffer;
import java.nio.channels.FileChannel;
 

//lsof : IfcConvert 23122  user  mem    REG    8,5 52923163  6034972 BOREL_IFC.ifc

public class Test {
	/*
	 * This Java application should be called with two args :
	 * args[0] is the pathname of IfcConvert
	 * args[1] is the pathname of the file to convert 
	 * 
	 * If the ifc file is big enough you'll see with lsof (lsof -f -p pid)
	 * that the file is mapped in memory in both  java and
	 * IfcConvert. 
	 * Moreover this memory segment is shared between the two processes as 
	 * their address are the same in each process.
	 * Memory is thus allocated only one time in the mmap scenario instead of two in the file/open 
        * one.
	 */
			
	 public static void main(String[] args) throws Exception {
		 RandomAccessFile file =  new RandomAccessFile(args[1],"r");
		 MappedByteBuffer mmap = file.getChannel().map(FileChannel.MapMode.READ_ONLY,0,file.length());
		 
		 
		 Runtime runtime = Runtime.getRuntime();
		 Process p = runtime. exec(args);
		 p.waitFor();
		 System.out.println(p.exitValue());
		 
		 
	 }
  
}

aothms:
Thanks for this! I will have a look and see how the mmap paging impacts performance. Might even be more efficient regardless of other processes using the data.

PR #242, opened 2017-07-23
Introduce memory-mapped file option

aothms:
See #22

It's rather messy, a lot of ifdefs and conditional code, but probably a good idea to have this option without breaking too much existing build environments.

After 0.5 this will be cleaned up.

In addition, this manual paging (the `BUF_SIZE` macro, now removed) should probably also be revisited now that many terminals (float, int, ...) are no longer parsed lazily, so likely a lot less seaking is necessary.

Anyway @jetablejfc apologies for the delay in merging this.


aothms:
I did a quick test [1] on a linux vm on a meager laptop and didn't find any difference in performance with or without --mmap. I will keep it as a an option for developers and will not release binaries with this option enabled. The build scripts build the required boost libs but do not enable USE_MMAP either.

[1] `~/IfcOpenShell/build/Linux/x86_64/install/ifcopenshell/bin/IfcConvert DC_Riverside_Bldg-LOD_300.ifc 300.dae -y --mmap? --include=entities IfcSlab`

PR #246, opened 2017-08-10
Changes the behavior of IfcFile::Init

LukeB42:
You probably don't want to merge this but to instead get the implementation of `Dup` to compile and then uncomment the `read` method in `IfcParseWrapper.i`.

If you can get `Dup` to compile without throwing a "missing member name" error then removing line 1267 from `IfcParse.cpp` reinstates the current `IfcFile::Init` behaviour.

I'd throw more time at perfecting this myself but declaring the method name in the `IfcFile.h` doesn't seem to suffice and I haven't got the patience to pretend the error makes any sense.

Best of luck.


aothms:
Hi, thanks for this. I admit something like this is more complex then one would think due to the various layers of Python and C++. 

I don't think changes are necessary to the stuff in `ifcparse/`. There is a `IfcFile::Init(void* data, int len)` that you can use. See [1] how you can get a pointer and length from a `std::string`. You need to create a copy (`memcpy`) for the data pointed to to live beyond the life of the string.

Then, another layer of complexity is added by the fact that the generated wrapper is not directly exposed into the python `ifcopenshell` module. There is handwritten code to add aditional syntactic sugar and more idiomatic interfaces [2]. So to the file class [3] you can add for example a `@staticmethod def from_string(s)` that calls the added `ifcopenshell_wrapper.read(std::string)` that you commented out.

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcgeomserver/IfcGeomServer.cpp#L488
[2] https://github.com/IfcOpenShell/IfcOpenShell/tree/master/src/ifcopenshell-python/ifcopenshell
[3] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcopenshell-python/ifcopenshell/file.py#L26

LukeB42:
@aothms thanks for pointing a tenable route out. I've remove all the code relating to `IfcFile::Dup` and used `IfcFile::Init` from the Python wrapper as you've suggested, in this commit.

There is an error around freeing the data once it's loaded but this works gracefully from the Python wrapper as long as you call `reference_to_IFC_instance.this.disown()` before exiting, otherwise there's a segfault there at the moment. If you let me know what I should be `free()`ing I've got no problem submitting a PR containing that.

Regards,
Luke

src/ifcwrap/IfcParseWrapper.i
```diff
@@ -394,14 +394,21 @@ private:
...
 		IfcParse::IfcFile* f = new IfcParse::IfcFile();
-		f->Init(s);
+		f->Init(fn);
+		return f;
+	}
+    IfcParse::IfcFile* read(const std::string& data) {
+		IfcParse::IfcFile* f = new IfcParse::IfcFile();
+        std::string copiedData;
+        copiedData = data;
+		f->Init((void *)copiedData.c_str(), data.length());
```
aothms: Hi, thanks for the updates, but this way of creating a copy is not sufficient. The issue with calling `std::string::c_str()` is that you loose all perks that come with C++ RAII or automatically garbage collected languages. Once the function `read()` is terminated `copiedData` will be freed. This means that the data pointed to by `c_str()` is also no longer available. Maybe it is not immediately apparent because after `IfcFile::Init()` the majority of data is parsed and present in memory. Strings and enumerations are parsed lazily and would probably crash in this approach. Additionally, once the IfcFile is freed it will try to clean up the `c_str()` which will not work, because it already has been freed or because it was allocated in a different way internally by the `std::string`. This is likely the error you report about. Long storey short, you really need to use the `memcpy()` approach used here: https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcgeomserver/IfcGeomServer.cpp#L490

LukeB42:
@aothms Thanks a tonne for linking to the technique necessary to get this working. The PR now hopefully contains useful changes.

Regards,
Luke

aothms:
Thanks for your continued efforts. I will cleanup the changes that are no longer necessary in another commit. Also, I will add the `ifcopenshell.file.from_string()`.

PR #251, opened 2017-08-22
Fix #250

### LINKED ISSUE #250: VS 2017 (15.3) and OCCT 7.1.0: unresolved external symbol 'recv'
Probably easiest to fix simply by adding linking to ws2_32.lib.

Stinkfist0:
Also print OCC version when `IfcConvert --version` invoked.


aothms:
Oh, great! Many thanks as always.

Should we investigate why OCC suddenly wants to open sockets?

PR #190, opened 2017-03-04
ifcparse refactoring

aothms:
Objectives:

- [x] Merge the various attribute containers (`IfcAbstractEntity`, `IfcWritableEntity`, `IfcParse::Entity`) -> `IfcEntityInstanceData`.
- [x] Provide latebound attribute access directly on `IfcBaseEntity` and remove `IfcLateboundEntity`
- [ ] ~~Make naming consistent~~ (not attempted in order not to make breaking changes in 0.5 branch)
- [ ] ~~Create `spf` namespace~~ (not attempted in order not to make breaking changes in 0.5 branch)

PR #254, opened 2017-08-30
don't assign ID to Simple types

elmirjagudin:
When adding entities, don't assign ID to Simple types.
Otherwise that type will be serialized incorrectly.

This fixes the issue with IFCPLANEANGLEMEASURE being serilzed
incorrectly when used as attribute to IFCMEASUREWITHUNIT.

With ID assigned, it will look like this:

 `#10=IFCMEASUREWITHUNIT(#9=IFCPLANEANGLEMEASURE(0.017453293),#8);`

The correct way to serialze is:

`#10=IFCMEASUREWITHUNIT(IFCPLANEANGLEMEASURE(0.017453293),#8);`

### DIFF
--- a/src/ifcparse/IfcParse.cpp
+++ b/src/ifcparse/IfcParse.cpp
@@ -1603,24 +1603,26 @@ IfcUtil::IfcBaseClass* IfcFile::addEntity(IfcUtil::IfcBaseClass* entity) {
 		}
 	}
 
-	int new_id = -1;
-	if (!new_entity->entity->file) {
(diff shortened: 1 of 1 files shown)


aothms:
Many thanks! Seems to be related to a specific condition here [1], perhaps that should have been `&&` instead. Introduced by #190. Yours should do the trick as well though, better not assign (and increment the counter) in the first place.

[1] https://github.com/elmirjagudin/IfcOpenShell/blame/b65a2a957ee529c56d181986020bf722bc0f3b36/src/ifcparse/IfcParse.cpp#L924

PR #258, opened 2017-09-13
IfcConvert Windows unicode support

aothms:
The idea is that all `std::string` file paths are assumed to be UTF-8 encoded. Therefore function signatures remain unchanged and behaviour is the same on nix and win. On nix UTF-8 paths work by default, on Windows this requires a constant back-and-forth conversion from UCS2/UTF-16 to UTF-8. Therefore the majority of changes are in IfcConvert that does need to have a wchar_t entry point for windows to accept arguments in unicode.

There are explicit `IfcUtil::path::to/from_utf8()` calls scattered throughout the code and a typedef `path_t` based on preprocessor directives. Perhaps `path_t` can be a class instead with implicit constructors from and casts to `std::string` and `std::wstring`. This to make the code in IfcConvert more readable.


Stinkfist0:
Great! For MinGW support I think we need to use _WIN32 instead of _MSC_VER for the Windows-specific code blocks.

aothms:
> For MinGW support 

Does MinGW have this `std::ofstream(wchar_t*)` constructor [1]? Because that is what happens with this  `(std::wstring IfcUtil::path::from_utf8(std::string&)).c_str()` on MSVC now.

[1] https://msdn.microsoft.com/en-us/library/7z434859.aspx#basic_ofstream__basic_ofstream

Stinkfist0:
Hmm, it might very well be that the wchar constructor is MSVC-specific extension indeed.

Stinkfist0:
Hi. Here's how I approached the problem in my little hobby project, this example only handles `std::ifstream/ofstream` though, for `std::cout/cerr` I added dedicated functions also:
https://github.com/Stinkfist0/PoGoCmp/blob/master/src/DbGen/PoGoDbGen.cpp#L38
https://github.com/Stinkfist0/PoGoCmp/blob/master/src/Lib/Utf8.cpp#L117
https://github.com/Stinkfist0/PoGoCmp/blob/master/src/Lib/Utf8.h#L49

aothms:
@Stinkfist0 thanks for sharing! I think it's a good approach, in line with http://utf8everywhere.org/ and with what I tried to accomplish

Don't know really if there's anyway of opening non-localized char paths then on MinGW (without opening Windows.h and just abandon std lib)

PR #260, opened 2017-09-20
CMakeLists.txt: add icui18n library for linux compilation

thomas-guenet:
To solve following error message:
```
Linking CXX executable IfcConvert
/usr/bin/ld: warning: libicui18n.so.54, needed by /**/lib/libboost_regex.so, not found (try using -rpath or -rpath-link)
```
I had to add `icui18n` library to the linking flags. This pull request updates the `CMakelists.txt` to solve the linking issue (linux CentOS).


Stinkfist0:
Boost.Regex can use ICU, but this should not be the default behavior. However, I don't think it's worth the trouble to start sprinkling our CMakeListst.txt with some BOOST_WITH_ICU build flags, so should be fine as-is.

aothms:
> Boost.Regex can use ICU, but this should not be the default behavior. However, I don't think it's worth the trouble to start sprinkling some BOOST_WITH_ICU build flags to our CMakeListst.txt, so, the fix suggested here looks OK.

I agree. In the long run we might get rid of ICU as a dependency all together though (but the same with boost if we upgrade to C++17 `std::optional` and `variant`).

@thomas-guenet many thanks!

PR #264, opened 2017-09-28
IfcConvert: Fix bug with --bounds for SVG

andersgb:
`if(bounding_width)` evaluates to true as long as `bounding_width` is
non-zero (including negative values). Since `bounding_width` and
`bounding_height` are default initialized to -1.0, we were erroneously
setting a bounding rectangle of (-1.0, -1.0) here. This means the SVG
output has been broken/malformed if we didn't specify --bounds.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -384,7 +384,8 @@ int main(int argc, char** argv)
     }
 #endif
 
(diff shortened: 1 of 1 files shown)


Stinkfist0:
Doh, this was a mistake by me. Thanks!

aothms:
Ok, many thanks!

PR #271, opened 2017-10-15
Validate mesh input

ideasman42:
This report comes from a but reported to Blender, where invalid geometry is apparently created by this importer.

https://developer.blender.org/T52921


aothms:
Thanks!

So there is a duplicate vertex reference in one of the polygons. I've asked to upload the original IFC file here as well, so that we can pinpoint the underlying issue as well.

~~~
BKE_mesh_validate_arrays: verts(96), edges(257), loops(489), polygons(163)
    Poly 156 has duplicated vert reference at corner (1)
    Loop 468 is unused.
    Loop 469 is unused.
    Loop 470 is unused.
~~~

PR #276, opened 2017-10-23
py 2 and 3

thorade:
http://python3porting.com/noconv.html#handling-exceptions

this is the first error i get when i try to start th eviewer from python 3
using the two-line from https://github.com/IfcOpenShell/IfcOpenShell/commit/1d1fb6d27a5bb1f221aed4dc290c1dc99a324e0f#commitcomment-17161165


aothms:
From the quoted document:

> The above syntax also works in Python 2.6

Good enough for me.

Thanks again!

PR #277, opened 2017-10-23
update SWIG to latest release

thorade:
because SWIG 3.0.7 is two years old and the intermediate releases all mention minor Python fixes.
http://www.swig.org/

I have used this for the last couple of builds and did not notice any negative effects.


thorade:
Closing for now, because I suddenly get an error about missing pthread,
I'll check whether this is due to the SWIG update, or to some other change in the master.

thorade:
The problems were not related to the SWIG update, so I am re-opening this PR.

aothms:
Thanks.

PR #284, opened 2017-11-01
geom/app.py: ConfigParser py2/py3 fix

andersgb:
When running the viewer app in python3, I got the following crash:
```
  File "geom_app.py", line 2, in <module>
    ifcopenshell.geom.app.application().start()
  File "/usr/lib/python3/dist-packages/ifcopenshell/geom/app.py", line 534, in __init__
    self.editor = code_edit(self.canvas, configuration().options('snippets'))
  File "/usr/lib/python3/dist-packages/ifcopenshell/geom/app.py", line 68, in __init__
    config = Cfg()
TypeError: 'ConfigParser' object is not callable
```

The branching for ConfigParser (for py2) and configparser (for py3)
should set the same type on Cfg. This fixes python3 on my box, haven't
tested for python2 though.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/geom/app.py
+++ b/src/ifcopenshell-python/ifcopenshell/geom/app.py
@@ -51,7 +51,7 @@ class configuration(object):
             Cfg = ConfigParser.RawConfigParser
         except:
             import configparser
-            Cfg = configparser.ConfigParser(interpolation=None)
+            Cfg = lambda: configparser.ConfigParser(interpolation=None)
             
         conf_file = os.path.expanduser(os.path.join("~", ".ifcopenshell", "app", "snippets.conf"))
         if conf_file.startswith("~"):


thorade:
What about using one of the two following imports (untested)
```python
from ConfigParser import RawConfigParser as Cfg
import ConfigParser.RawConfigParser as Cfg
```

andersgb:
From master:
```python
        try:
            import ConfigParser
            Cfg = ConfigParser.RawConfigParser
        except:
            import configparser
            Cfg = configparser.ConfigParser(interpolation=None)
```
Hm, I guess the issue here is that there is an argument used in the py3 variant (`interpolation`). If that wasn't the case, we could do something like
```python
        try:
            import ConfigParser.RawConfigParser as Cfg
        except:
            import configparser.ConfigParser as Cfg
```

thorade:
Or upgrade the configparser on Python 2 using pip?
http://python-future.org/compatible_idioms.html#configparser

andersgb:
Sure, I can change this PR to use that pattern instead.

andersgb:
See force-pushed commit. Note that I haven't tested this for python2

src/ifcopenshell-python/ifcopenshell/geom/app.py
```diff
@@ -46,13 +46,9 @@
...
 class configuration(object):
     def __init__(self):
-        try:
-            import ConfigParser
-            Cfg = ConfigParser.RawConfigParser
-        except:
-            import configparser
-            Cfg = configparser.ConfigParser(interpolation=None)
-            
+        # python2 users need to 'pip install configparser'
```
aothms: Ugh that is a bit rough on our py2 users. While I tend to agree with @thorade to follow documented best practices, I'd like to avoid requiring all users to install yet another module for something fairly trivial as this.

I can no longer see your earlier fix. What was it? You're right about the current state that the py2 branch assigns a type object that can be instantiated and the py3 branch assigns a instance. This is clearly wrong. What about changing:

~~~diff
- Cfg = configparser.ConfigParser(interpolation=None)
+ Cfg = lambda: configparser.ConfigParser(interpolation=None)
~~~

to get sth callable into `Cfg`?

thorade:
The previous fix was using a function call in both parts of the try/except block
and then using Cfg without parantheses aftwerwards.

```python
        try:
            import ConfigParser
            Cfg = ConfigParser.RawConfigParser()
        except:
            import configparser
            Cfg = configparser.ConfigParser(interpolation=None)
```

aothms:
> function call in both parts

I see, I prefer the other way around as I suggested to make Cfg callable as it is called twice and I'm not sure about the side effects of reusing the same config parser twice.

andersgb:
Agree that sounds like a good approach. Pushed a third iteration here.

Btw, previous two versions are at:
https://github.com/andersgb/IfcOpenShell/commit/32f40548b8963cbb749dddabf0c776def44d3118 
https://github.com/andersgb/IfcOpenShell/commit/b07a413d9975682f15d01860687868a8618941e4

aothms:
Great, thanks guys.

PR #287, opened 2017-11-01
Issue281 future import

### LINKED ISSUE #281: from __future__ import
Some (or all?) of the handwritten python files have the line
```python
from __future__ import print_function
```
at the top.

I suggest to also import absolute_import and division :
```python
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
```

Should be quite harmless and help to avoid regressions.
http://python-future.org/quickstart.html

thorade:
This closes #281

Opening the pull request so the changes can easily be seen, but more testing on Py2 is necessary.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
+++ b/src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
@@ -17,6 +17,10 @@
 #                                                                             #
 ###############################################################################
 
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
 import random
 import operator
 from collections import namedtuple, Iterable
--- a/src/ifcopenshell-python/ifcopenshell/__init__.py
+++ b/src/ifcopenshell-python/ifcopenshell/__init__.py
@@ -17,6 +17,8 @@
 #                                                                             #
 ###############################################################################
 
+from __future__ import absolute_import
+from __future__ import division
 from __future__ import print_function
 
 import os
--- a/src/ifcopenshell-python/ifcopenshell/entity_instance.py
+++ b/src/ifcopenshell-python/ifcopenshell/entity_instance.py
@@ -16,6 +16,9 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.        #
 #                                                                             #
 ###############################################################################
+
+from __future__ import absolute_import
+from __future__ import division
 from __future__ import print_function
 
 import functools
--- a/src/ifcopenshell-python/ifcopenshell/file.py
+++ b/src/ifcopenshell-python/ifcopenshell/file.py
@@ -17,6 +17,10 @@
 #                                                                             #
 ###############################################################################
 
+from __future__ import absolute_import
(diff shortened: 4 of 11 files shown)


src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
```diff
@@ -17,6 +17,10 @@
 #                                                                             #
 ###############################################################################
 
+from __future__ import absolute_import
+from __future__ import division
```
thorade: Check line 185
https://github.com/IfcOpenShell/IfcOpenShell/pull/287/files#diff-a4812f80978da2638af11368d5da84b5R185

aothms: You mean because of the usage of `map()` which returns a list / iterable in py2 / 3? I think it is fine in both versions, due to the `*` the elements get expanded into positional arguments to `gp_Pnt()` regardless.

thorade: I was looking at that line, because as a test I ran the futurize script and this line was the only one where it detected a division. But I think that also here the new division is what really was intended, so it is ok as it is.

What really needs to be tested is whether any imports break on Python 2.

aothms:
> py3 only knows next() function 

Actually the `it` in `it.next()` is not really a Python iterator, it's this beast [1], which always has a `next()` method.
[1] https://github.com/thorade/IfcOpenShell/blob/9ee87fd23bea8d995645f66a88d685ad638e09a1/src/ifcgeom/IfcGeomIterator.h#L496

The python docs [2] say for 2 and 3 respectively

> Retrieve the next item from the iterator by calling its `next()` method

> Retrieve the next item from the iterator by calling its `__next__()` method

[2] https://docs.python.org/3.5/library/functions.html#next

So I think this would work on py2 but not py3, since it only has a method `next()` and not `__next__()`. Did you test this? If this is indeed an issue perhaps it can be fixed with a simple `__next__ = next` here [3].

[3] https://github.com/thorade/IfcOpenShell/blob/9ee87fd23bea8d995645f66a88d685ad638e09a1/src/ifcopenshell-python/ifcopenshell/geom/main.py#L77

thorade:
That line was also detected by the futurize script, but as it seems that script was wrong here, so I'll revert the second commit (or you can cherry pick just the first commit and close the PR).

thorade:
I removed the commit with the `next()`

src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
aothms: Ah, I get it now, it's about the division and not about the map(). Both operands are floating point so `from __future__ import division` doesn't affect anything.

aothms:
> but as it seems that script was wrong here

Well, to be fair, detailed static analysis would be required to truly understand the type of `it`. I am not surprised there would be false positives in such a script.

> I removed the commit with the next()

Thanks.

PR #290, opened 2017-11-06
autopep8

### LINKED ISSUE #286: autopep8
The Python code could be made pep8 compliant using autopep8:
```
python -m pip install -U autopep8
cd src\ifcopenshell-python
autopep8 --in-place --recursive  --max-line-length=200 ifcopenshell
autopep8 --in-place --recursive  --max-line-length=200 --aggressive --aggressive --aggressive ifcopenshell
```
The changes are mostly whitespace, depending on aggressiveness level.

thorade:
This would close #286,
but once again, I am not sure whtether all the changes are welcome.
The output is produced by the autopep8 script, so it was not much work at all
and I do not mind if the PR is rejected/closed right away.

pep8 recommends to use def over lambda (rule E731)
https://stackoverflow.com/questions/25010167/e731-do-not-assign-a-lambda-expression-use-a-def
That rule could be ignored:
```
autopep8 --in-place --recursive  --max-line-length=200 --ignore=E731 ifcopenshell
```

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
+++ b/src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
@@ -30,44 +30,47 @@ shape_tuple = namedtuple('shape_tuple', ('data', 'geometry', 'styles'))
 handle, main_loop, add_menu, add_function_to_menu = None, None, None, None
 
 DEFAULT_STYLES = {
-    "DEFAULT"            : (.7 , .7, .7     ),
-    "IfcWall"            : (.8 , .8, .8     ),
-    "IfcSite"            : (.75, .8, .65    ),
-    "IfcSlab"            : (.4 , .4, .4     ),
-    "IfcWallStandardCase": (.9 , .9, .9     ),
-    "IfcWall"            : (.9 , .9, .9     ),
-    "IfcWindow"          : (.75, .8, .75, .3),
-    "IfcDoor"            : (.55, .3, .15    ),
-    "IfcBeam"            : (.75, .7, .7     ),
-    "IfcRailing"         : (.65, .6, .6     ),
-    "IfcMember"          : (.65, .6, .6     ),
-    "IfcPlate"           : (.8 , .8, .8     )
+    "DEFAULT": (.7, .7, .7),
+    "IfcWall": (.8, .8, .8),
+    "IfcSite": (.75, .8, .65),
+    "IfcSlab": (.4, .4, .4),
+    "IfcWallStandardCase": (.9, .9, .9),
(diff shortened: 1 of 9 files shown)


thorade:
To ignore (some?) whitespace changes, append `?w=1` to the URL:
https://github.com/IfcOpenShell/IfcOpenShell/pull/290/files?w=1
https://github.com/IfcOpenShell/IfcOpenShell/pull/290/files

src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
```diff
@@ -30,85 +30,93 @@
 handle, main_loop, add_menu, add_function_to_menu = None, None, None, None
 
 DEFAULT_STYLES = {
```
thorade: this change might also be undesired

src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
```diff
@@ -179,47 +191,53 @@ def set_shape_transparency(ais, t):
...
+
+if not brep_data:
+    return shape_tuple(brep_object, None, styles)
+
+try:
+    ss = OCC.BRepTools.BRepTools_ShapeSet()
+    ss.ReadFromString(brep_data)
+    occ_shape = ss.Shape(ss.NbShapes())
+except BaseException:
+    pass
```
thorade: Did this change the behavior? What is the intended behavior?

src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
aothms: Well, it's not a big deal to me. These aligned indents are a pain to maintain anyway when updates take place.

src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
```diff
@@ -179,47 +191,48 @@ def set_shape_transparency(ais, t):
...
     brep_data, occ_shape, styles = None, None, ()
-    
+
     is_product_shape = True
     try:
         brep_data = brep_object.geometry.brep_data
         styles = brep_object.geometry.surface_styles
-    except: 
-        try: 
+    except BaseException: try:
```
thorade: this is wrong, I'll hcorrect it, and file an issue with autopep8

thorade: This was causes by an autopep8 bug:
https://github.com/hhatto/autopep8/issues/320
After updating to the latest autopep8 version, the reformated code is valid.

aothms:
Thanks again!

PR #293, opened 2017-11-08
Create docstrings for generated python code.

civilx64:
[http://www.swig.org/Doc1.3/Python.html#Python_nn65](http://www.swig.org/Doc1.3/Python.html#Python_nn65)

### DIFF
--- a/src/ifcwrap/IfcPython.i
+++ b/src/ifcwrap/IfcPython.i
@@ -86,6 +86,9 @@
 	#include <BRepTools_ShapeSet.hxx>
 %}
 
+// Create docstrings for generated python code.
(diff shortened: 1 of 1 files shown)


aothms:
>  It is helpful as a starting point but would be much better with corresponding documentation of the C++ API

Sad but true. All help welcome also there. But thanks for this as well.

aothms:
Also, many docstrings would probably be hidden in the way the user-written ( It is helpful as a starting point but would be much better with corresponding documentation of the C++ API) interacts with the generated code.

Using update_wrapper() would be an idea to retain more docstrings. https://docs.python.org/2/library/functools.html#functools.update_wrapper

Issue #292, opened 2017-11-07
Circular shapes not supported in SVG conversion


andersgb:
I'm using `IfcConvert` to get SVG 2D geometry from my IFC, but I'm struggling with circular shapes. I have boiled the problem down to [this IFC file](https://github.com/IfcOpenShell/IfcOpenShell/files/1450549/cylinders.ifc.zip). The converted SVG result is included below and will typically render nothing in browsers, or will give an error if loading into a validating SVG library. 

Since `SvgSerializer.cpp` uses `<path>` Arcs for circular segments, it will not give valid SVG output for *complete* circles. The SVG standard says `If the endpoints (x1, y1) and (x2, y2) are identical, then this is equivalent to omitting the elliptical arc segment entirely.` [source](https://www.w3.org/TR/SVG/implnote.html#ArcOutOfRangeParameters)

I guess complete circles should use `<circle>` instead of `<path>`. The same goes for a complete ellipse.

-----

`./IfcConvert --bounds 1024x1024 cylinders.ifc cylinders.svg`
```xml
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <g id="product-34d87c32-c3cf-11e7-8e5b-80193481049d">
        <g id="product-product-34d87c2c-c3cf-11e7-8e5b-80193481049d-body">
            <path style="stroke:black; fill:none;" d="M0,0 A256,256 0 0,0 0,6.27019e-14"/>
        </g>
        <g id="product-product-34d87c2d-c3cf-11e7-8e5b-80193481049d-body">
            <path style="stroke:black; fill:none;" d="M0,512 A256,256 0 0,0 0,512"/>
        </g>
        <g id="product-product-34d87c2e-c3cf-11e7-8e5b-80193481049d-body">
            <path style="stroke:black; fill:none;" d="M0,1024 A256,256 0 0,0 0,1024"/>
        </g>
        <g id="product-product-34d87c2f-c3cf-11e7-8e5b-80193481049d-body">
            <path style="stroke:black; fill:none;" d="M1024,0 A256,256 0 0,0 1024,6.27019e-14"/>
        </g>
    </g>
</svg>
```

[...]

PR #294, opened 2017-11-08
SVG conversion improvements

andersgb:
Two first commits are general improvements to the SVG conversion code. 

Third commit is a bit more opinionated, adding new logic for cross section height.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -262,7 +262,7 @@ int main(int argc, char** argv)
     std::string unicode_mode;
 #endif
     short precision;
-
+	double section_height;
     po::options_description serializer_options("Serialization options");
     serializer_options.add_options()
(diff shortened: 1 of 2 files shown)


andersgb:
Added a fourth commit for CLI argument `--section-height`.

A potential problem with using storey elevation + 1 meter is if the storey elevation is not really representing the actual elevation of the geometry. I have now seen an example where the storey elevation is `-0.5`, but all geometry contained is in the range `[1.5, 6]`

aothms:
Many thanks!

> A potential problem with using storey elevation + 1 meter is if the storey elevation is not really representing the actual elevation of the geometry

I agree, good addition. Also, Elevation might be missing altogether as it is an optional attribute. 

How this would work with `--section-height` and multiple stories is still a bit vague, but I guess the `--include ... traverse` could help there to process storeys separately with an explicit section height.

So, more as a note to self, this is unrelated to #292.

PR #302, opened 2017-11-23
SVG closed circles

### LINKED ISSUE #292: Circular shapes not supported in SVG conversion

andersgb:
This is my attempt to resolve #292 

Fully closed circles will now use `<circle>` instead of arc segments, this will now work:

`IfcConvert --exclude entities IfcSpace IfcOpeningElement  --section-height "1.0" --bounds 1024x1024 cylinders.ifc cylinders.svg`

I also implemented similar treatment of closed ellipses, but I haven't been able to test the implementation, since the `ellipses.ifc` will not end up generating `Geom_Ellipse` type geometry. Instead, the ellipses end up as `Geom_BSplineCurve` types. I don't have any full sized IFC with ellipse geometry either.


src/ifcconvert/SvgSerializer.cpp
```diff
@@ -66,14 +68,77 @@ void SvgSerializer::write(path_object& p, const TopoDS_Wire& wire) {
...
+        Handle(Standard_Type) ty = curve->DynamicType();
+        bool conical = (ty == STANDARD_TYPE(Geom_Circle) || ty == STANDARD_TYPE(Geom_Ellipse));
+        bool closed = ALMOST_THE_SAME(u1 + PI2, u2);
+        if (conical && closed) {
+            if (first) {
+                if (ty == STANDARD_TYPE(Geom_Circle)) {
+                    Handle(Geom_Circle) circle = Handle(Geom_Circle)::DownCast(curve);
+                    double r = circle->Radius();
+                    gp_Circ c = circle->Circ();
+                    gp_Pnt center = c.Location();
```
aothms: Small note: `gp_Cirl::Location()` returns type `const gp_Pnt &`. `gp_Pnt center = c.Location();` will make a copy, while `const gp_Pnt& center = c.Location();` will more efficiently point to the same reference. In practice though, the compiler might be smart enough not to make a copy anyway.

src/ifcconvert/SvgSerializer.cpp
```diff
@@ -66,14 +68,77 @@ void SvgSerializer::write(path_object& p, const TopoDS_Wire& wire) {
...
 	for (TopExp_Explorer edges(wire, TopAbs_EDGE); edges.More(); edges.Next()) {
 		const TopoDS_Edge& edge = TopoDS::Edge(edges.Current());
 
 		double u1, u2;
 		Handle(Geom_Curve) curve = BRep_Tool::Curve(edge, u1, u2);
 
-		const bool reversed = edge.Orientation() == TopAbs_REVERSED;
+        Handle(Standard_Type) ty = curve->DynamicType();
+        bool conical = (ty == STANDARD_TYPE(Geom_Circle) || ty == STANDARD_TYPE(Geom_Ellipse));
+        bool closed = ALMOST_THE_SAME(u1 + PI2, u2);
```
aothms: The question is whether `u1` is always smaller than `u2`. Frankly I don't really know. I think that's why I was using `fmod()` earlier.

The docs say

> if p2 < p1 the edge will be REVERSED.

https://www.opencascade.com/doc/occt-6.9.1/refman/html/class_b_rep_builder_a_p_i___make_edge.html

But I'm not sure in what order they are returned by `BRep_Tool::Curve()` later on.

aothms:
Many thanks. My comments are not important, don't need to be addressed. Let me check if I can somehow get that ellipse example working.

src/ifcconvert/SvgSerializer.cpp
aothms: I created some hollow circle profile defs, to see if the reversed orientation there would result in `u1 > u2`. But it didn't. Let's leave it the way it is.

aothms:
> Let me check if I can somehow get that ellipse example working.

I couldn't the ellipses are maintained all the way up to the extrusion. But the OCCT section creates a bspline out of it. I guess this could be fixed upstream in OCCT, but I don't have the time to report it there.

I'll commit code to approximate all other curve types with straight line segments. That should be good enough.

Thanks again for your effort here!

PR #306, opened 2017-12-04
SvgSerializer: Fix bug in ellipse writeout

andersgb:
Add a missing "=" to create a valid SVG for ellipses.

I accidentally stumbled upon a way to actually get an ellipse in SVG by
tilting a cylinder. If I tilt the cylinder via the `ObjectPlacement` of
the `IfcProduct` I get an ellipse. If I set the `ExtrudedDirection` of
the `IfcExtrudedAreasolid`, however, I get a sampled path.

---

This is a fixup for #302


aothms:
Oh thanks. Nice 👍

PR #314, opened 2017-12-18
Corrected name of *nix build script.

civilx64:
It looks like at some point this changed from a shell script to a python script.  I corrected the link so as to avoid a 404.


aothms:
Ah thanks for that :)

PR #316, opened 2017-12-22
BaseQuantities support in IfcConvert XmlSerializer

stefkeB:
Added the export of BaseQuantities in the XML export of IfcConvert.
Very similar to the export of properties.


aothms:
Thanks, this is great!

PR #322, opened 2018-01-24
Make IFCConvert and IFCGeomServer optional

carmenfan:
For users like me, who only uses the libraries, it saves a lot of time and trouble if there are configuration flags available to skip the compilation of the executables


aothms:
Thanks, probably the `IfcGeom` target [1] needs to be installed as well, you're not using that?

https://github.com/3drepo/IfcOpenShell/blob/623e95cefe462a1fa6e58a1480a991049502a67b/cmake/CMakeLists.txt#L582

carmenfan:
I was using that on a previous version[v0.4] and I'm trying to upgrade to the latest version. Since the library spilled out 2 geom with different IFC Versions I just assumed that you guys have split up the library into 2... Is this not the case? If I use build ifcgeom does it support both v2.3 and v4?

We are mainly using it for v2.3 at the moment so I was just linking it against 2x3

aothms:
> If I use build ifcgeom does it support both v2.3 and v4?

Yes exactly. If you use this `IfcGeom::Iterator` it selects the correct library based on schema identifier in the file [1].

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom_schema_agnostic/IfcGeomIterator.h#L88

PR #323, opened 2018-01-24
Adding ifcgeom_schema_agnostic headers in include directory

carmenfan:
ifcgeom_schema_agnostic  headers are not being added in the include directory, which makes the library unusable as a library


aothms:
Ah thanks, this is all work in progress, more rigorous changes will follow after [1] and [2] are merged. Thanks for having a look if you have other ideas in general on the schema agnostic code let me know.

[1] https://github.com/sanderboer/IfcOpenShell/tree/multithreading
[2] https://github.com/aothms/IfcOpenShell_CGAL/

PR #327, opened 2018-02-14
IfcConvert: Add --building-local-placement argument

andersgb:
Similar to --site-local-placement. Motivated by dealing with at least
two examples where IFC files for the same construction site have
inconsistent ObjectPlacement hierarchy:

- MEP models exported from MagiCAD for Revit:
   ObjectPlacement.PlacementRelTo of IfcBuilding is empty
- Regular construction models exported from Revit:
   ObjectPlacement.PlacementRelTo of IfcBuilding is pointing to
   IfcSite's placement

----

This feels a bit like a hack. If I had known about this mess before adding `--site-local-placement` #240 , maybe exposing a `--placement-rel-to IFC_TYPE` (e.g. `--placement-rel-to IfcSite`) would have been a better choice. 

But anyway this solved my immediate problem, I can now run
```
IfcConvert --building-local-placement
```

on a bunch of IFC files (originating from both MagiCAD and Revit exports) and get the OBJ files into common local coord system like I want.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -298,6 +298,8 @@ int main(int argc, char** argv)
 		("site-local-placement",
 			"Place elements locally in the IfcSite coordinate system, instead of placing "
 			"them in the IFC global coords. Applicable for OBJ and DAE output.")
+		("building-local-placement",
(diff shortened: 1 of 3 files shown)


aothms:
Ok, great thanks. If and when this extends to even more possibilities we might consider something like `--relative-placement IfcBuilding` instead to keep the amount of options under control.

PR #352, opened 2018-03-13
[CMAKE] Enable the use of IFC 4 by default.

frozar:
As IFC 4 is the most recent file format, I would say why not enable it by default. Opinions?

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -25,7 +25,7 @@ OPTION(UNICODE_SUPPORT "Build IfcOpenShell with Unicode support (requires ICU)."
 OPTION(COLLADA_SUPPORT "Build IfcConvert with COLLADA support (requires OpenCOLLADA)." ON)
 OPTION(ENABLE_BUILD_OPTIMIZATIONS "Enable certain compiler and linker optimizations on RelWithDebInfo and Release builds." OFF)
 OPTION(IFCCONVERT_DOUBLE_PRECISION "IfcConvert: Use double precision floating-point numbers." ON)
-OPTION(USE_IFC4 "Use IFC 4 instead of IFC 2x3 (full rebuild recommended when switching this)" OFF)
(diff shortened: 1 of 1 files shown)


Stinkfist0:
I think IFC 2x3 is still much more prevalent in the industry and hence a more logical default choice. However, the upcoming 0.6 branch will handle both 2x3 and 4 versions without any compile-time configuraion, so this build option will mostly likely become obsolete in the future.

frozar:
Ok, that's fine. And just to know, should I use the master branch or the 0.6 branch? What is the difference between this two? Would you merge the 0.6 branch into master in the future?

Stinkfist0:
I think 0.6 branch is still work in progress and I haven't tried it myself. Master branch should be pretty stable, although there seems to  plethora of regression issues with the latest OCE/OCCT (can be mitigated using the older version).

frozar:
Ok, thank you for these answers.

So if I well understand, if want to read IFC4 file I should compile with `USE_IFC4=ON` and if I want to read IFC2X3 file I should compile with `USE_IFC4=OFF`, souldn't I? So to be able to handle both format, I should have the project compiled twice with these different compiling option?

Stinkfist0:
Yes, one must compile the project twice currently and save the binaries of the previous build, so something like this (assuming using MSVC, semi-pseudo instructions):
0. cd win
1. build-deps
2. run-cmake (IFC2x3 by default)
3. install-ifcopenshell
4. copy `..\installed-vs2017-x64 ` as `..\installed-vs2017-x64-ifc2x3` 
5. del CMakeCache.txt just to be sure
5. run-cmake -DUSE_IFC4=1
6. install-ifcopenshell
7. `..\installed-vs2017-x64 ` should now contain the IFC4 binaries

frozar:
@Stinkfist0 : Thank you very much for your answers.

This PR should not be open anymore :)

Issue #369, opened 2018-04-12
Reuse detection takes very long time with a big model


otsoa:
Hello,

We have a big (~900M) Ifc file which we are converting to collada. It seems that reuse detection takes very long time (i left it running for hours on debug and it was still doing 'reuse_ok' calls). If I either set use_woorld_coords flag, which by passes the reuse detection code, or revert to commit just before 6c4fee102 (Color elements by material association. Reimplement reuse detection. Fix #265) the conversion completes in couple of minutes.

otsoa:
It seems that in our case this is also related to heavy use of mapped representations. If I add keep track of reuse_ok results in those cases, so that the somewhat heavy reuse_ok operation is not done multiple times for same 'mapping target', the prosessing times are again reasonable.

```cpp
IfcSchema::IfcRepresentation* representation_mapped_to = kernel.representation_mapped_to(representation);
if (representation_mapped_to) {
    // Check if this represenation has (or will be) processed as part its mapped representation
    representation_processed_as_mapped_item = ok_mapped_representations->contains(representation_mapped_to) || reuse_ok_(kernel.products_represented_by(representation_mapped_to));
}

if (representation_processed_as_mapped_item) {
    ok_mapped_representations->push(representation_mapped_to);
    _nextShape();
    continue;
}
```

aothms:
That's very interesting, thanks for sharing. Naively I had not expect IFC attribute traversal times to be significant wrt geometry interpretation. It might be related to some of the inefficiencies of looking up inverse attributes. Would it be possible to share a pull request?

otsoa:
Indeed the heaviest stacktrace leads to IfcParse::IfcFile::entityById(int). I now also changed entity_by_id_t from map to unordered_map but i need to profile a bit more. I will create a pull request after some more testing.

aothms:
> I now also changed entity_by_id_t from map to unordered_map

That's something I had been considering as well. Alternatively just a vector if you can get a reasonable estimate whether the integer names are somewhat contiguous. Or something like a sparse vector impl https://www.boost.org/doc/libs/1_35_0/libs/numeric/ublas/doc/vector_sparse.htm ?

STEPcode uses judy arrays https://github.com/stepcode/stepcode/blob/master/src/base/judy/src/judy.h

[...]

PR #371, opened 2018-04-16
Topic/optimise reuse detection

otsoa:
Optimisations for reuse detection on big models. issue #369

### DIFF
--- a/src/ifcgeom/IfcGeomIterator.h
+++ b/src/ifcgeom/IfcGeomIterator.h
@@ -116,6 +116,9 @@ namespace IfcGeom {
 		IfcSchema::IfcProduct::list::ptr ifcproducts;
 		IfcSchema::IfcProduct::list::it ifcproduct_iterator;
 
+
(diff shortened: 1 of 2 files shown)


aothms:
Many thanks! I am going to have a quick look on the impact on performance of both commits individually on large and small files.

aothms:
Sorry for the delay. Wanted to do a more elaborate check and see if we could see the log vs linear 
instance storage in the map. Didn't find that time for that and didn't want to delay merging your work any longer.

Merged master onto both commits and ran a conversion from Clinic MEP [1] to .dae. Significant improvement. I can imagine there are files with heavy use of mapped items where the difference is more pronounced, clinic mep is only a moderately complex file. Anyway this is really the enjoyable part of open source software development for me, that somebody comes along and in 10 lines of code contributes something significant. Thanks.

version | time | %
---|---|---
master | 337.953249931 | 100%
1f297f0 | 266.664919853 | 78.9%
c8895c0 | 248.890305042 | 73.6%

[1] https://www.nibs.org/?page=bsa_commonbimfiles#project3

PR #375, opened 2018-04-23
update SF.net download URL

### LINKED ISSUE #374: win build script: SWIG download fails
Not sure whether this is just temporary, or maybe sourceforge has changed the URL a bit!?
https instead of http ??

https://github.com/IfcOpenShell/IfcOpenShell/blob/master/win/build-deps.cmd#L429-L444

https://sourceforge.net/projects/swig/files/swigwin/swigwin-3.0.12/

EDIT: Downloading manually works, the URL used was:
https://datapacket.dl.sourceforge.net/project/swig/swigwin/swigwin-3.0.12/swigwin-3.0.12.zip

EDIT: Documentation of the mirrors is here:
https://sourceforge.net/p/forge/documentation/Downloading%20files%20via%20the%20command%20line/

thorade:
This closes #374

### DIFF
--- a/win/build-deps.cmd
+++ b/win/build-deps.cmd
@@ -488,7 +488,7 @@ exit /b %IFCOS_SCRIPT_RET%
 pushd "%2"
 if not exist "%~3". (
     call cecho.cmd 0 13 "Downloading %DEPENDENCY_NAME% into %~2."
-    powershell -Command "$webClient = new-object System.Net.WebClient; $webClient.Proxy.Credentials = [System.Net.CredentialCache]::DefaultNetworkCredentials; $webClient.DownloadFile('%1', '%3')"
+    powershell -Command "[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12; $webClient = new-object System.Net.WebClient; $webClient.Proxy.Credentials = [System.Net.CredentialCache]::DefaultNetworkCredentials; $webClient.DownloadFile('%1', '%3')"
     REM Old wget version in case someone has problem with PowerShell: wget --no-check-certificate %1
 ) else (
     call cecho.cmd 0 13 "%DEPENDENCY_NAME% already downloaded. Skipping."
--- a/win/build-deps.cmd
+++ b/win/build-deps.cmd
@@ -432,7 +432,7 @@ set DEPENDENCY_NAME=SWIG %SWIG_VERSION%
 set DEPENDENCY_DIR=N/A
 set SWIG_ZIP=swigwin-%SWIG_VERSION%.zip
 cd "%DEPS_DIR%"
-call :DownloadFile https://datapacket.dl.sourceforge.net/project/swig/swigwin/swigwin-%SWIG_VERSION%/%SWIG_ZIP% "%DEPS_DIR%" %SWIG_ZIP%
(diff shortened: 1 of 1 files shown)


thorade:
Note: This points to a certain mirror, so I will just close it, and possibly re-open once I found how to point to a generic/permanent URL.

thorade:
One has to use https with TLS1.2 now to download files from Sourceforge.

thorade:
It seems that only Powershell >= 5.0 is able to do TLS 1.2,
so this possibly breaks the script for some users.
Whenever that happens, these users would have to manually download and install an up-to-date PowerShell!

thorade:
As an alternative, one could just assume SWIG is installe dand on the path, just like git and CMake.

Stinkfist0:
Could https://osdn.net/projects/sfnet_swig/releases/ be trusted for obtaining SWIG.

thorade:
You could also download from here:
https://anaconda.org/conda-forge/swig/files
https://anaconda.org/anaconda/swig
But both OSDN and Anaconda also use https, so one would have to test what TLS level they require.

aothms:
Thanks, sorry for the late reply. If powershell 5 is required. According to the table here [1] this would break the script for everybody not running Win 10. That sounds a bit harsh. Where did you read about the powershell version required for TLS 1.2?

[1] https://4sysops.com/wiki/differences-between-powershell-versions/

Alternatively we should probably at least detect this early on and prevent the same confusion you experienced. E.g. `powershell -c "exit $PSVersionTable.PSVersion.Major -lt 5"`

thorade:
I think I read that in one the github issues or stackoverflow, e.g. here:
https://stackoverflow.com/a/36266139/874701
Apparently it depends on the .NET version used by powershell!?
But that is not "official", I can try to find an official source, or try to test the script on an older computer.

Also, it is possible to install the latest Powershell version also on older Windows, but it does not come preinstalled.

aothms:
> I can try to find an official source, or try to test the script on an older computer.

Yeah I'm not sure how to proceed on this. Don't want to waste much time on this, but also don't want to upset users with cryptic error messages. Let's just merge this and see what happens. Thanks!

Stinkfist0:
It would be nice to have the required Powershell documented here, and perhaps provide a download link it, https://github.com/IfcOpenShell/IfcOpenShell/blob/master/win/build-deps.cmd#L592

aothms:
I agree and/or a check to see whether the required version is found: `powershell -c "exit $PSVersionTable.PSVersion.Major -lt 5"`

thorade:
Download links are provided in the blog post that Thomas linked above:
https://4sysops.com/wiki/differences-between-powershell-versions/
https://www.microsoft.com/en-us/download/details.aspx?id=54616

Not sure where to put that check and corresponding output, otherwise I would send a PR.

thorade:
Just tested on Windows 7, and the error message people would see now is similar to:
```powershell
Exception setting "SecurityProtocol": "Cannot convert null to type "System.Net.SecurityProtocolType" due to invalid enumeration value
s. Specify one of the following enumeration values and try again. The possible enumeration values are "Ssl3, Tls"."
At line:2 char:35
+ [System.Net.ServicePointManager]:: <<<< SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12;
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : PropertyAssignmentException
```
So, at least it is a bit more informative than before.

PR #393, opened 2018-05-30
minor cosmetic cleanup

thorade:
just two minor things my IDE compalined about:
use triple double-quotes for the docstring,
and do not use python builtin names/keywords for variable names, because otherwise the builtin will be shadowed


aothms:
Thanks :)

PR #399, opened 2018-06-08
Fixed compilation on Ubuntu 18.04

Victor-Haefner:


aothms:
Thanks, sorry for the delay. I might implement the a conditional switch to C++11/c++0x based on ICU version to still enable compilation on older systems.

PR #404, opened 2018-06-17
Add ICU header and namespace in IfcCharacterDecoder

stefkeB:
To compile on macOS with a recent ICU, we have to add the correct header and refer to the right namespace in IfcCharacterDecoder


Stinkfist0:
Update on the required CMake version should be reflected in the README. I would have though this version would be too new for the Travis CI script (uses pretty old Ubuntu 14.04) but apparently it build everything just fine.

stefkeB:
I wasn't even aware that this "push" to my fork was actually being transmitted to the original respository as wall (I'm still new in this git-based workflows). Is C++ 11 otherwise a strict requirement? I've only had to set this to compile with an up-to-date ICU. So far, it works fine on my Mac, but I don't have any idea on the impact on other platforms.

stefkeB:
I've had to add the same code corrections in the v0.6 branch to compile on macOS High Sierra. I was able to reload my project with v0.6 libraries and at first sight, I could load Ifc2x3 and IFC4 files with no major problems. Have to test more in detail, though.

aothms:
Thanks @stefkeB I will merge these changes with 0.6 and try to keep the 0.5 branch on c++03 (conditionally based on occt and icu version)

aothms:
> Thanks @stefkeB I will merge these changes with 0.6 and try to keep the 0.5 branch on c++03 (conditionally based on occt and icu version)

Plans changed a bit. master is now also on C++11 and v0.6.0 has ICU removed. Many thanks for the PR anyway. Best to focus on V0.6 now.

PR #427, opened 2018-08-02
SegFault when exporting xml - Fix #419

### LINKED ISSUE #419: Segfault / IndexError when exporting to xml
I've got a `segfault` (on `0.5.0.p2` release) and `IndexError` (on `master`) exception when exporting the IFC below to xml (remove `.txt`):

[LakesideRestaurant.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/2207579/LakesideRestaurant.ifc.txt)

Investigating the cause, I've discovered that happens (here)[https://github.com/femesq/IfcOpenShell/commit/e27d7a7b2594431b46a48039ebb92673ca9b0996#diff-80da8c2080af3e87a888ebc500a77ec7L141] and made a `try..catch` to workaround it.

It seems that `instance->getArgumentCount()` is returning an argument-count higher than the instance really have, but couldn't imagine where to start to fix it (I'm new to the code base).

Can I have some direction to contribute on this?

femesq:


Stinkfist0:
Instead of changing 400 lines of code, you probably want to submit only the first commit (remove .gitignore changes)?

femesq:
Sorry about that... seems like my IDE auto-formatted the code, changing more lines than it should.. Will try to fix it...

femesq:
Couldn't make it work... Will close this PR and open a new one with necessary lines only..

PR #428, opened 2018-08-05
Workaround wrong argument count. Fix #419

### LINKED ISSUE #419: Segfault / IndexError when exporting to xml

femesq:
Making new PR with necessary lines-change only


aothms:
Thanks for the PR and the tidied version.

aothms:
I changed the `continue` into a `break` statement as subsequent attribute numbers are only going to result in the same error being reported.

Issue #410, opened 2018-07-09
Done scanning file


hujiaxin0:
The executable program is generated by compiling the source code. After “Done scanning file”, the executable program has no output for a long time.

Stinkfist0:
Related issue I made in the past, apparently the issue was forgotten open although I think at the time the problem was resolved, https://github.com/IfcOpenShell/IfcOpenShell/issues/122

aothms:
> Related issue I made in the past, apparently the issue was forgotten open although I think at the time the problem was resolved, #122

Seems to be applicable to XML only now.

Could also be that the first geometry item to interpret takes a very long time I think?

> is generated by compiling the source code

Did you enable compiler optimizations? What platform are you on.

What kind of file are we talking about in terms of file size / complexity? And what are you converting to?

Jeffrey-Heart:
My goal is to convert IFC files into OBJ files. The size of the IFC file is 197M.  

If I use your compiled executable program "IfcOpenShell-0.5.0-preview1", then it can run quickly. 

When I use your compiled executable program "IfcOpenShell-0.5.0-preview2",  the program will stop running after output "##............................"( only output two "#") .

When I  compile the source code to get an executable program.
The  program executes this statement `"context_iterator.initialize()"`
`
`            IfcSchema::IfcProduct::list::ptr products = ifc_file->entitiesByType<IfcSchema::IfcProduct>();
            for (IfcSchema::IfcProduct::list::it iter = products->begin(); iter != products->end(); ++iter) {
                IfcSchema::IfcProduct* product = *iter;
                if (product->hasObjectPlacement()) {
					// Use a fresh trsf every time in order to prevent the result to be concatenated
                    gp_Trsf trsf; 
					bool success = false;
					
					try {
						success = kernel.convert(product->ObjectPlacement(), trsf);
					} catch (const std::exception& e) {
						Logger::Error(e);
					} catch (...) {
						Logger::Error("Failed to construct placement");
					}
					
					if (!success) {
						continue;
					}

                    const gp_XYZ& pos = trsf.TranslationPart();
                    bounds_min_.SetX(std::min(bounds_min_.X(), pos.X()));
                    bounds_min_.SetY(std::min(bounds_min_.Y(), pos.Y()));
                    bounds_min_.SetZ(std::min(bounds_min_.Z(), pos.Z()));
                    bounds_max_.SetX(std::max(bounds_max_.X(), pos.X()));
[...]

PR #440, opened 2018-08-20
IfcGeomIterator: compute model's bounding box only if needed (IfcConv…

Stinkfist0:
…ert --center-model used).

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -296,7 +296,7 @@ int main(int argc, char** argv)
 			"Applicable for DAE output.")
         ("center-model",
             "Centers the elements upon serialization by applying the center point of "
-            "all placements as an offset. Applicable for OBJ and DAE output.")
+            "all placements as an offset. Applicable for OBJ and DAE output. Can take several minutes on large models.")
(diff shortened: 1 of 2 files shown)


Stinkfist0:
Potentially fixes #410

aothms:
Many thanks!

If this really takes time perhaps we need to look into another library for these matrix multiplications. Maybe `boost::ublas` is quicker?

Stinkfist0:
Also, I wonder if --site-local-placement and/or  --building-local-placement could replace --center-model and --model-offset use cases (working around huge offsets in the model) altogether.

aothms:
> working around huge offsets in the model

I don't think these are the only use cases. I think there are valid reasons to ignore the site or building placement when the actual precise positions are somehow relevant to you either because you're overlaying multiple models (although --model-offset could be used for that as well) or because you're relating information back to the original models or other information sources (2d drawings perhaps) or one of these nodes introduces a rotation that you want to ignore to keep things orthogonal...

I do feel that all these flags and other could probably live in some other executable or script that preprocesses IFC files and emits a new IFC file.

Stinkfist0:
Yeah, e.g. a set of Python DAE/XML post-processing scripts could be great instead of semi-hacks in IfcConvert.

Issue #308, opened 2017-12-09
Recent ICU headers require C++11


aothms:
Detect ICU header version and set accordingly? Or always enable and risk upsetting older compilers?

aothms:
Also, the `using namespace icu;` has been removed in ICU 61, so it needs to be added to IfcOpenShell code base or types need to be qualified including the icu: namespace

stefkeB:
I had to add "#include "unicode/unistr.h" in IfcCharacterDecoder.h to compile ifcOpenShell with a recent ICU version (ICU 60.2)

aothms:
@stefkeB ok thanks for the headsup. I think `unicode/unistr.h` is part of earlier versions of ICU as well, but apparently needs to be explicitly included from now, feel free to send a PR to keep your codebase in sync.

BIMvoice:
Moot on current `v0.8.0`: the build sets `CMAKE_CXX_STANDARD 17` and errors below 17 (`cmake/CMakeLists.txt:21-24`), and ICU itself has been removed as a dependency. The C++11-for-ICU question no longer applies. Can be closed as obsolete.

PR #453, opened 2018-09-06
IfcCharacterDecoder.cpp: Explicitly include unicode.h

andersgb:
In order to build on Ubuntu 18.04, I had to do this change. This is pretty much issue #308, where this change is suggested in a comment. For my case here, adding `-DCMAKE_CXX_STANDARD=11` is not necessary as long as we use this change.

Using this Dockerfile I can now build with both ubuntu 16.04 and 18.04 as base image
```
FROM ubuntu:18.04
RUN apt-get update && apt-get install -y \
    apt-utils \
    supervisor \
    git \
    cmake \
    build-essential \
    locales \
    libboost-all-dev \
    wget \
    unzip \
    libicu-dev \
    swig \
    liboce-foundation-dev \
    liboce-modeling-dev \
    liboce-ocaf-dev \
    liboce-visualization-dev \
    liboce-ocaf-lite-dev \
    libpcre3-dev \
    libxml2-dev

RUN mkdir -p /deploy/libs && \
    cd /deploy/libs && \
    git clone https://github.com/andersgb/IfcOpenShell.git && \
    cd IfcOpenShell && \
    git checkout 193556966205 && \
    pwd && \
    cd cmake && \
    mkdir build && \
    cd build && \
    cmake -DCOLLADA_SUPPORT=False -DUSE_IFC4=False -DBUILD_IFCPYTHON=True -DUNICODE_SUPPORT=True -DOCC_LIBRARY_DIR=/usr/lib/x86_64-linux-gnu .. && \
    make -j2 && \
    make install
```

I have no idea whether this may break other systems though.

### DIFF
--- a/src/ifcparse/IfcCharacterDecoder.cpp
+++ b/src/ifcparse/IfcCharacterDecoder.cpp
@@ -27,6 +27,7 @@
 #include <string>
 #include <sstream>
 #include <iomanip>
+#include <unicode/unistr.h>
 
 #include "../ifcparse/IfcCharacterDecoder.h"
 #include "../ifcparse/IfcException.h"


aothms:
Hi thanks, but it seems that header is already there: https://github.com/IfcOpenShell/IfcOpenShell/blame/master/src/ifcparse/IfcCharacterDecoder.cpp#L73 am I missing something?

#404 was merged into v0.6, but github doesn't realize it.

andersgb:
@aothms totally right, I was building on an outdated master branch, my bad. Due to commit 11d0c6ef569 current master builds cleanly. Closing this PR.

PR #469, opened 2018-09-18
V0.6.0

SOCfperrier:
Fix dll export for Windows


SOCfperrier:
Forget this, mishandling

aothms:
Thanks, I merged master into v0.6.0 so your fix is also applied there.

PR #472, opened 2018-09-21
Implement support for custom default materials

hypirion:
Hey! Thanks for the good tool :)

I'd like to use a custom set of default materials instead of the default ones, and I thought it would be more convenient to do that through an input file instead of applying a patch every time master is updated.

I have implemented a way to read json files with material defaults. Consider a file `material-defaults.json`, for example with the following data:

```javascript
{
  "IfcWall": {
    "diffuse": [0.964, 0.913, 0.729]
  },
  "IfcSlab": {
    "diffuse": [0.8, 1, 1]
  },
  "IfcSpace": {
    "diffuse": [0.6, 0.8, 0],
    "transparency": 0.8
  }
}
```

Then, though a new CLI flag, you can refer to that file by using the new option `--default-material-file material-defaults.json`. Except for the toplevel default value, all the default values will be replaced by this.

I don't think this is ready to be merged down yet though, as I have not yet added documentation, and I'm pretty sure you'd like to have better error handling (as of now, there isn't any at all). I'd need some help with the error handling, as this is literally my first lines of C++ and I'm not sure how to do that idiomatically.

I'm also curious if you think using normalised RGB values is better than hexadecimal RGB values or not. I just implemented normalised RGB ones as that was the easiest, but that could be changed.


aothms:
Thanks this is great!

One minor issue would be that it seems that items not found in the json mapping, are now rendered pure black. Because that color never get's initialized any more [1].

I guess the proper way to do this, is not global static variables, but static variables within function scope and returned as references. Or some other way that it cannot be referenced without being initialized first.

~~~C++
const IfcGeom::SurfaceStyle& get_default_material() {
  static IfcGeom::SurfaceStyle m;
  // ideally only executed once though of course
  m.Diffuse().reset(IfcGeom::SurfaceStyle::ColorComponent(0.7, 0.7, 0.7));
  return m;
}
~~~

On the other hand, maybe it would also be nice though if it were possible to set this catch-all default? Maybe with `"*"` as a json key? Although 0.7;0.7;0.7 is a pretty safe choice I guess.

Ideally I'd have the 0.5 branch still on C++98. So your usage of `auto` keywords and `range-based for` do not comply with that. This is not caught in continuous integration currently and recent version of both ICU and OCCT require C++11 these days. So it's not a major issue.

Re error handling: there are probably at least a few exceptions you want to catch such as `boost::property_tree::ptree_bad_data` for the templated `get_value()`. There already is a `file_exists()` [2] you can reuse. Your choice on whether to terminate or run with the partly initialized defaults.

> not sure how to do that idiomatically

Opinions differ on that I guess, but we'll just have to play along with what boost has in mind for us. This isn't user-supplied data in the sense that it needs to be super scrutinized, I think catching `std::exception` (all boost exceptions probably inherit from this) in IfcConvert.cpp and passing that to IfcOpenShell's `Logger::Error()` is good enough to give users a sense where their json file needs improvement.

Right now with the `double rgb[3]` and for loop you can corrupt the stack or return uninitialized memory if `i` does not end up at exactly 3. Maybe check if there is a way to get a container out of a ptree directly, or initialize the array and keep an eye on `i`.

If you can tolerate the C++ syntax for member function pointers you can probably eliminate the redundancies in your function.

~~~C++
// declare and init function pointer named setter
boost::optional<IfcGeom::SurfaceStyle::ColorComponent>& (SurfaceStyle::*setter)() = &SurfaceStyle::Diffuse;
// call
(default_materials[name].*(setter))().reset(IfcGeom::SurfaceStyle::ColorComponent(rgb[0], rgb[1], rgb[2]));
~~~

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/c072ccc423fb4d9a4f4714c0e0721f6a9df43b47/src/ifcgeom/IfcGeomRenderStyles.cpp#L172
[2] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcconvert/IfcConvert.cpp#L99

hypirion:
> Thanks this is great!
> 
> One minor issue would be that it seems that items not found in the json mapping, are now rendered pure black. Because that color never get's initialized any more [1].

Oh dang, that's a bug – I still intended to have default objects as 0.7;0.7;0.7.

> On the other hand, maybe it would also be nice though if it were possible to set this catch-all default? Maybe with `"*"` as a json key? Although 0.7;0.7;0.7 is a pretty safe choice I guess.

That's a great suggestion. As mentioned above, I still think it makes sense to have objects as 0.7;0.7;0.7 by default, but being able to override the default through `"*"` sounds nice.
 
> Ideally I'd have the 0.5 branch still on C++98. So your usage of `auto` keywords and `range-based for` do not comply with that. This is not caught in continuous integration currently and recent version of both ICU and OCCT require C++11 these days. So it's not a major issue.

Oh, I wasn't aware, sorry about that. I could fix that if you'd like to keep backwards compatibility. (There's some usage of `auto` in `ifcgeom/IfcMax.cpp` as well, by the way.)

> Re error handling: _[...]_
> _[...]_ If you can tolerate the C++ syntax for member function pointers you can probably eliminate the redundancies in your function.

Great, thanks for the pointers! I'll read up and have a stab at this, I guess I'll be able to get back to this with something more refined by early next week.

hypirion:
Alright, I was a bit more busy than expected the last weeks, but I finally got around to fix these issues. Right now, I have fixed the following

- [X] Doesn't use auto anymore
- [X] Handles errors "gracefully" and avoid memory corruptions
- [X] Fixes the bug with default fallback material not being handled
- [X] Picks up and uses `"*"` as a parameter for default fallback material

There is one thing that I found out in hindsight, and that is that I use spaces instead of tabs, which looks ugly if you don't use the same indentation as I do (of course, I prefer 2 spaces instead of Stinkfist's 4). I will rebase that to avoid introducing spaces into the codebase a bit later.

As for error handling with respect to JSON parsing: I couldn't find out how to use `Logger::Error`, as it didn't spit out anything on stderr for me. I looked around in the codebase and found out that `cerr` was used in other places, so I presume it is fine to use that one as well. Also, I couldn't find examples nor documentation on how to read JSON data into existing structures through the boost library, so the RGB colour components are read in through a function. And some of the error messages are pretty horrible right now, but at least it fails if you send in some data that cannot be properly parsed, and gives you an indication of which file that failed.

Stinkfist0:
Just a minor note: `set_default_style(filename)` and `get_default_style(ifc_type)` is a rather inconsistent API, maybe `set_default_style_file` would be nicer.

hypirion:
Good point, I have added a commit to fix that.

I tried to fixup tabs vs. spaces by rebasing this branch, but it turns out to be a bit more nasty than I hoped for. I'm not sure what you folks prefer – I can tabify the lines I've added by adding a commit, or just leave it as is.

Stinkfist0:
We discussed the coding conventions a while back and this is what we came up: https://github.com/IfcOpenShell/IfcOpenShell/wiki/Coding-Conventions These are not enforced yet, but probably at some point (when 0.6.0 becomes the master?).

Personally I try to use the convention-to-be (spaces) when writing completely new code, but use tabs when altering existing code.

hypirion:
Ahh, thanks for the pointers. I thought the tabs were the intended convention, as that was what I saw the most of in the codebase. After reading through #21, I saw that https://github.com/IfcOpenShell/IfcOpenShell/issues/21#issuecomment-177879806 suggests that there will be a cleanup through some autoformatter at some point, so I will not worry too much about the current commits.

PR #475, opened 2018-09-25
Use /permissive- introduced in VS 2017 to enforce standards-conformance.

Stinkfist0:
https://docs.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=vs-2017

Should alleviate the problem of breaking GCC/Clang builds occasionally.


aothms:
Thanks this might be very helpful indeed

PR #476, opened 2018-09-25
Optimize/fix geometry filtering and reusability checks

Stinkfist0:
I'll try to provide exact test/profiling measurements later today, but some weeks ago we were working with a huge IFC file (1,1 GB file with 36.7 million STEP entities) and including only a subset (250 objects) for the conversion. With these fixes the conversion reduced approx. from 60 minutes to 6 minutes and the resulting DAE file size reduced from 935 MB to 48 MB.


Stinkfist0:
OK, RelWithDebInfo build and ENABLE_BUILD_OPTIMIZATIONS=1, with these fixes the conversion took 100 minutes 23 seconds, (file parsing and shutdown take several minutes also on a file of this caliber), yielding a 1,6 GB DAE. I'll leave the conversion without the fixes running for the night.

Stinkfist0:
Results for the conversion without these fixes: `Conversion took 273 minutes 57 seconds` and 5.562 GB DAE.

Issue #480, opened 2018-09-29
build_all.py on Fedora 27 - - - Stop at OpenCollada


paullee0:
Hi, I attempt again compiling - - - aim to use IfcOpenShell-python for FreeCAD IfcExport / Import only.

I try the nix/build_all.py and it stop at OpenCollada.

I am not familiar with Python, appreciate any ideas could be offered.

```
Configuring OpenCOLLADA...
Building OpenCOLLADA...   
----------------------------------------------------------------------
/run/media/paullee/80G_Data/Download/Download_IfcOpenShell/git/IfcOpenShell/build/Linux/x86_64/build/OpenCOLLADA/Externals/UTF/src/ConvertUTF.c: In function ‘ConvertUTF16toUTF8’:
/run/media/paullee/80G_Data/Download/Download_IfcOpenShell/git/IfcOpenShell/build/Linux/x86_64/build/OpenCOLLADA/Externals/UTF/src/ConvertUTF.c:270:65: warning: this statement may fall through [-Wimplicit-fallthrough=]
      case 4: *--target = (UTF8)((ch | byteMark) & byteMask); ch >>= 6;
                                                              ~~~^~~~~
/run/media/paullee/80G_Data/Download/Download_IfcOpenShell/git/IfcOpenShell/build/Linux/x86_64/build/OpenCOLLADA/Externals/UTF/src/ConvertUTF.c:271:6: note: here
      case 3: *--target = (UTF8)((ch | byteMark) & byteMask); ch >>= 6;
      ^~~~
/run/media/paullee/80G_Data/Download/Download_IfcOpenShell/git/IfcOpenShell/build/Linux/x86_64/build/OpenCOLLADA/Externals/UTF/src/ConvertUTF.c:271:65: warning: this statement may fall through [-Wimplicit-fallthrough=]
      case 3: *--target = (UTF8)((ch | byteMark) & byteMask); ch >>= 6;
                                                              ~~~^~~~~
/run/media/paullee/80G_Data/Download/Download_IfcOpenShell/git/IfcOpenShell/build/Linux/x86_64/build/OpenCOLLADA/Externals/UTF/src/ConvertUTF.c:272:6: note: here
      case 2: *--target = (UTF8)((ch | byteMark) & byteMask); ch >>= 6;
      ^~~~
[...]

PR #481, opened 2018-10-01
Changing opencollada version to "v1.6.63"

paullee0:
Discussion:
https://github.com/IfcOpenShell/IfcOpenShell/issues/480

Summary:
"... Compiling OpenCOLLADA  fail on fedora 27... changing  opencollada version to "v1.6.63" (since some time now they do proper tagging over there, the commit we track now is from 19 Dec 2013, probably worth updating as newer compilers might rightfully complain)...

Probably best to rm -rf ... IfcOpenShell/build/Linux/x86_64/build/OpenCOLLADA before restarting."


aothms:
Thanks!

PR #482, opened 2018-10-02
Ifcfile entity look-up + Kernel::get_layers optimizations

Stinkfist0:
Recently we reran conversion for a file ("MEP File") that to our best of recollection was converted in somewhat orderly fashion a long time ago. This time, the XML conversion seemed to take forever, so I started to looking into the biggest time spenders in XML serialization code, which lead mostly to IfcFile::getInverse(). This lead me to optimize entity look-ups and to remove one particular slow part of layer retrieval code that doesn't seem to yield any results ever.

```
Timing includes only the conversion ("-layer code" == unnecessary(?) layer code removed, no other optimizations)
Conference center (http://openifcmodel.cs.auckland.ac.nz/_models/20160124OTC-Conference%20Center.ifc):
before:         1 m 15 s
-layer code:         8 s
after:               7 s

"Massive File" (1.1 GB)
before:         21 m 2 s
-layer code:    19 m 46 s
after:           6 m 58 s

"MEP File" (180 MB)
before:         did not finish (hours?)
-layer code:    14 m 58 s
after:           3 m 25 s
```

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -116,6 +116,7 @@ bool rename_file(const std::string& old_filename, const std::string& new_filenam
 
 static std::stringstream log_stream;
 void write_log(bool);
+std::string format_duration(time_t start, time_t end);
 
 /// @todo make the filters non-global
 IfcGeom::entity_filter entity_filter; // Entity filter is used always by default.
@@ -473,11 +474,15 @@ int main(int argc, char** argv)
         int exit_code = EXIT_FAILURE;
         try {
             if (init_input_file(input_filename, ifc_file, no_progress || quiet, mmap)) {
+                time_t start, end;
+                time(&start);
                 XmlSerializer s(output_temp_filename);
                 s.setFile(&ifc_file);
                 Logger::Status("Writing XML output...");
                 s.finalize();
-                Logger::Status("Done!");
+                time(&end);
+                Logger::Status("Done! Conversion took " +  format_duration(start, end));
+
                 rename_file(output_temp_filename, output_filename);
                 exit_code = EXIT_SUCCESS;
             }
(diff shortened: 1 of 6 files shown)


aothms:
Thanks this is brilliant, but there two problems I can think of.

First, I thought of using a vector, but it's not unlikely to have huge outliers in the instance name ids. What do you think of the idea to first scan through the file and try to obtain a measure for how sparse the instance name ids are and then decide on `unordered_map` vs `vector`. I can work on that.

Second, caching the instance refs need to somehow take into account that the files can be mutated. This is already a huge pain in the current code base. Or shall we set a flag that a file is going to be immutable and only in that case cache the instance refs?

Stinkfist0:
I did some testing regarding map vs. unordered_map vs. vector. Vector was the fastest by a wide margin and (naturally) used the least memory. Although the the memory usage reduction was negated by the cached map. Map and unordered_map did not have much difference, sometimes one was faster than the other. Ah yes, I did not take into consideration the mutability of the file fully. Yeah, maybe it would be the best (and easiest) to allow the user of the IfcFile to (re)compute the cache with a public function when the user knows the file will not be mutated.

Stinkfist0:
Also ideally, maybe one "entities by ref" map would suffice, and I tried to achieve that, but IfcEntityList made the task rather awkward, so I dediced not to pursue that for now.

Stinkfist0:
I'll try to to continue with this PR this week.

Stinkfist0:
I'll finish this this week.

Stinkfist0:
So, I'm thinking about the simplest ways to handle the entity cache invalidation. Could the modified entity perhaps easily mark the owning file as dirty so the next time the cache is used it would be cleared?

aothms:
Perhaps the entire file doesn't need to be marked as dirty? It should be possible to navigate to exactly the correct entries. But agree, let's keep that as a follow up task, as manipulating data and inspecting inverses are usually not done simultaneously. File-wide flag should be fine :+1.

aothms:
> Clearing the whole cache should not really affect the performance of typical use cases where the file is never modified though.

I agree. 

> one particular slow part of layer retrieval code that doesn't seem to yield any results ever

I ran through all layer assignments in the small set of test files, and at least there this relationship between layer and individual item was never encountered.

One reason why this 2nd step would be slow is that a LOT OF things in IFC are representation items. Looking at a IfcShellBasedSurfaceModel, all referenced instances: IfcClosedShell - IfcFace - IfcFaceBound - IfcPolyLoop - IfcCartesianPoint are all subclasses of IfcRepresentationItem.

Perhaps an alternative approach would be to set a limit (max_level=2) to `prod->entity->file->traverse(prod->Representation())` so that only IfcProductDefinitionShape - IfcShapeRepresentation - and the immediate items are considered. 

I think that would be a minimal performance difference. I can imagine how for this e.g. MEP file, there are probably huge graphs of items hidden under these representations. But the top one should be a single shell most of the time.

I mean it just feels odd to eliminate code that the schema so explicitly requires: 

~~~
TYPE IfcLayeredItem = SELECT
  ( IfcRepresentationItem
  , IfcRepresentation);
END_TYPE; 
~~~

aothms:
@Stinkfist0 is there still work pending on this? Sorry, lost track a bit.

Stinkfist0:
Ah, right, for some reason I thought this was already merged! Does the minimal cache invalidation look okay to to you? If yes, I won't be continuing further with this.

aothms:
Yes, cache invalidation seems fine to me. There is it a minimal functional difference now, that before a null shared pointer would be returned and now a pointer is returned to an empty list. Of course the later is much better behaviour (my previous self was sometimes too focussed on needless allocations). But I do think there might be parts of the code base that do directly check the shared pointer truth value rather than checking for an empty list.

aothms:
> There is it a minimal functional difference now

I will take care of that.

PR #483, opened 2018-10-03
Check shell validity before passing it over to TOPODB::Shell()

Stinkfist0:
A shell validity check that was found to be useful earlier this year.


aothms:
Thanks. Out of curiosity, do you have a test file you can share (also privately at models@ifcopenshell.org) where this helped?

Stinkfist0:
I'll contact there original author of this code if he can remember some specific file that he had problems with.

Stinkfist0:
Unfortunately we could not remember the specific test files. No bad feelings if this is rejected, can be easily resubmitted if we find the actual problematic files.

aothms:
Yeah I think I have to let this one slip. There's been too many changes in the whole TopoDS_Shell handling in the 0.6 branch for this to make sense. Thanks anyway,

PR #484, opened 2018-10-08
abort IFC file parsing on unexpected schema + small IfcConvert tweaks/fixes

Stinkfist0:


aothms:
Some users might complain about this, as they might depend on the relatively large overlap between 2x3 and 4. Or use IfcConvert to convert old 2x2 files. But yeah, better be explicit about this.

PR #489, opened 2018-10-23
IFC Relation Parents / Child

Archimen:
Hello,
Nice jobs with the code.
I am a french developper and I will do my best to expose my question in english.

I am really new to IfcOpenShell and Ifc file in general. I would like to know if your IfcOpenShell module is able to looking for parents/child entities of an object in Ifc graph EXPRESS data, or if this question can't be answered in IfcOpenShell because it's a behavior of native Ifc hierarchy file.  
I need to traverse entities to find their connection and I don't know how to do it.

I tried the .traverse() method but I didn't find a lot of documentation and I don't really know how to use it, and the expected result isn't very clear for me.
I got every inversed properties to find connections around an object but I would like to know the connections in the Express Data graph and not in the structural element graph to generate a graphdb with elements of my Ifc file.

I really think IfcOpenShell can help me to solve this problem and i hope an answer is possible.

It would be very appreciated to have some help.


aothms:
I don't think you intended to open a pull request, I answered your question over here: https://sourceforge.net/p/ifcopenshell/discussion/1782716/thread/ce96fadec3/?limit=25#c39d

Issue #117, opened 2016-08-09
Generate schema documentation from IfcDoc baseline files


Stinkfist0:
Looks like some kind of encoding bug in the code generator?

aothms:
Yes, the comments come from an old MS Access dump that used to be online. These days however the documentation is provided officially as a part of IfcDoc, with documentation strings encoded in an IFC-like (10303-21) encoded file.

The code generator should probably be rewritten to use this new documentation, which would also be a good point to look into these encoding issues. There is a good chance they were introduced when I exported the MS Access db to .csv.

aothms:
Available here

http://www.buildingsmart-tech.org/specifications/specification-tools/ifcdoc-tool/ifcdoc-baselines

Moult:
I suspect this bug is no longer relevant as the new IFC documentation has entirely new tooling, of which IfcOpenShell plays a part. So... maybe this can be closed?

aothms:
God yes

PR #505, opened 2018-11-19
Update documentation.py

SigveMartin:
On macOS Mojave v10.14. Using python3.6 conda environment (named ifcbridge). 

Got this error:  

(ifcbridge) Sigves-MacBook-Pro:ifcexpressparser Sigve$ python bootstrap.py express.bnf > express_parser.py && python express_parser.py /Users/Sigve/Projects/bSN/ifcbridge-test/ifc4x2-draft/IFC4x2.exp
  File "express_parser.py", line 460, in <module>
    import header
  File "/Users/Sigve/Projects/bSN/ifcbridge-test/IfcOpenShell/src/ifcexpressparser/header.py", line 24, in <module>
    import documentation
  File "/Users/Sigve/Projects/bSN/ifcbridge-test/IfcOpenShell/src/ifcexpressparser/documentation.py", line 51, in <module>
    for oid, name, desc in csv.reader(f, delimiter=';', quotechar='"'):
  File "/Users/Sigve/anaconda/envs/ifcbridge/lib/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 6602: invalid start byte

*** 
Found a solution through this: https://stackoverflow.com/questions/21504319/python-3-csv-file-giving-unicodedecodeerror-utf-8-codec-cant-decode-byte-err

Changed line 50 in documentation.py from: 

*    with open(fn) as f:

to

*    with open(fn,'r', encoding='mac_roman', newline='') as f:

It now generates these successfully: 

Ifc4x2-latebound.cpp		
Ifc4x2-latebound.h		
Ifc4x2.cpp			
Ifc4x2.h		
Ifc4x2enum.h

Haven't tested on other platforms, so don't know if it will work or break somehting for other platforms. Create a pull request..

### DIFF
--- a/src/ifcexpressparser/documentation.py
+++ b/src/ifcexpressparser/documentation.py
@@ -31,7 +31,7 @@ import re
 import os
 import csv
 
-from schema import OrderedCaseInsensitiveDict 
+from schema import OrderedCaseInsensitiveDict
 
 try: from html.entities import entitydefs
 except: from htmlentitydefs import entitydefs
@@ -47,7 +47,7 @@ regices = list(zip([re.compile(s,re.M) for s in [r'<[\w\n=" \-/\.;_\t:%#,\?\(\)]
 definition_files = ['DocEntity.csv', 'DocEnumeration.csv', 'DocDefined.csv', 'DocSelect.csv']
 definition_files = map(make_absolute, definition_files)
 for fn in definition_files:
-    with open(fn) as f:
+    with open(fn,'r', encoding='mac_roman', newline='') as f:
         for oid, name, desc in csv.reader(f, delimiter=';', quotechar='"'):
             name_to_oid[name] = oid
             oid_to_name[oid] = name


aothms:
Hi Sigve, I think there are definitely encoding errors in these files. But I would more in favour of adding `errors='ignore'` to `open()` then assuming this mac-specific encoding. Could you try that as well and see if that solves the issue you had?

You didn't accidentally modify these files right (by opening in an application for example)? i.e. `git status` is clean.

Reading CSV files in Python is different btw for python 2 and python 3. This guy has a work-around to make this version agnostic: https://stackoverflow.com/questions/41585078/how-do-i-read-and-write-csv-files-with-python?answertab=votes#tab-top

The current code base seems to follow the Python 2 approach. Maybe that is part of the problem as well, because I don't see any `0xff` byte at position 6602 in any of the CSV files.

aothms:
PS. This whole documentation part needs to be changed anyway. These CSV files are based on an export of the old MS Access databases.

This information is now contained in some custom stp files that IfcDoc uses: http://www.buildingsmart-tech.org/specifications/specification-tools/ifcdoc-tool/ifcdoc-baselines

See #117

aothms:
Hi @SigveMartin did you have a look whether adding `errors='ignore'` to `open()` also fixes the issues you had?

SigveMartin:
sorry for not updating after the test @aothms! As far as I can remember I got it to work by the changes you proposed. However, I didn't get time to finish what I started and haven't picked it up yet :-/

aothms:
Ok, no worries, I'll go ahead and close this then for the time being, but let me know once you pick this up again!

PR #528, opened 2019-01-02
Fix compilation using occt version < 7

tpaviot:
The ```DumpErrors``` method is available only for occt versions >= 7


tpaviot:
No, I only know that this method is unavailable in 6.9.1

PR #542, opened 2019-01-29
Fix C++17 compilation errors

lamorlette-cintoo:
std::bind1st and std::mem_fun are deprecated since C++11 and removed in C++17.
I also fixed a warning preventing me to compile my project with VCC.


aothms:
Thanks. Initially the aim was to keep 0.5 on C++98, but I think the world has moved on and there is no point in doing that. ICU and OCCT 7 both require C++11 after all.

PR #543, opened 2019-02-06
basic update to blender2.8, use of existing mesh if id match,

tibicen:
- I've cleaned up th code to python 3.7.
- updated to blender 2.80, **breaks comaptibility with earlier versions**.
- each import creates separate collection (group) named after a file (TODO: get file creation timestamp too)
- relations have own collection too (IFC Site, ifc project ifc floor etc, all with no geometry)
- if there was alredy imported mesh with the same id it use the previously imported mesh for new object (ex. all doors can be edited at once).
- updated material creation to Eevee viewport.

TODO: This is step one, just to make it work, now I will try to get a littlebit deeper, and later add more funcionality.


aothms:
Thanks this is wonderful.

I don't have 2.8 installed yet. I'll merge first test later :)

> updated to blender 2.80, **breaks comaptibility with earlier versions**.

Yes, but anyway the update to python37 will break compatibility and I don't plan on releasing too many builds of the add-on, so I'm perfectly fine with this.

> get file creation timestamp too

~~~
f = ifcopenshell.open(fn)
f.header.file_name.time_stamp
~~~

Will give you a file-based timestamp. I think that just focussing on the GlobalId attributes (guid) will be enough for revision management though. You never really know where the file will come from, maybe it's better not to focus on file-specific things. The industry is trying to move to more service / transaction approaches for quite some time now...

PR #550, opened 2019-02-13
process IfcIndexedPolyCurve without segments

### LINKED ISSUE #549: IfcIndexedPolyCurve without segments
There is a minor issue with `IfcIndexedPolyCurve`.

According to the standard, the `Segments` attribute of `IndexedPolyCurve` is optional, but currently it is treated as mandatory in IfcOpenShell, such that representation items without the attribute set will be ignored. Apparently there was an error in the [html documentation](http://www.buildingsmart-tech.org/ifc/IFC4/Add1/html/link/ifcindexedpolycurve.htm) of the entity type, omitting the ? in the attribute table. This documentation error has been fixed in subsequent IFC4 releases, but it really was only a typo. The Express specification for this entity type has not changed and the verbal documentation of semantics when `Segments` is omitted seems also unaltered.

Here is a test file, a modified version of the [IFC4 slab example](http://www.buildingsmart-tech.org/ifc/IFC4/Add1/html/link/slab-standard-case.htm) with segments removed: 
[IfcIndexedPolyCurve.ifc](https://github.com/IfcOpenShell/IfcOpenShell/files/2763119/IfcIndexedPolyCurve.ifc.txt).

_Originally posted by @hlg in https://github.com/IfcOpenShell/IfcOpenShell/issues/503#issuecomment-454684379_

hlg:
resolves #549

### DIFF
--- a/src/ifcgeom/IfcGeomWires.cpp
+++ b/src/ifcgeom/IfcGeomWires.cpp
@@ -912,7 +912,7 @@ bool IfcGeom::Kernel::convert(const IfcSchema::IfcIndexedPolyCurve* l, TopoDS_Wi
 				throw IfcParse::IfcException("Unexpected IfcIndexedPolyCurve segment of type " + IfcSchema::Type::ToString(segment->type()));
 			}
 		}
-	} else {
+	} else if (points.begin() < points.end()) {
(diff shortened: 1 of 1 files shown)


aothms:
Thanks. Really appreciate the contribution.

PR #558, opened 2019-03-08
Update Python to 2.7.16 - So Fedora 29 Compile

paullee0:


aothms:
Thanks for following up on this!

Issue #488, opened 2018-10-19
IfcGeomIterator segfault on next() function call in loop


hylandnp:
Reading IFC files (commonly IfcBuildingElementProxies, i.e. potentially badly-formed third-party model elements) with _IfcGeomIterator_ crashes with segmentation fault on next() function at some point during an iteration loop.

Debugging led me to the **create_shape_model_for_next_entity()** function, where in the code:
```cpp
IfcSchema::IfcProduct* product = *ifcproduct_iterator;
Logger::SetProduct(product);
```
Has a null pointer for _ifcproduct_iterator_, so when it's passed to the logger it tries to access an invalid memory location.

![screenshot from 2018-10-19 12-44-00](https://user-images.githubusercontent.com/3501643/47216477-f5fa9880-d39c-11e8-9139-b2847cc56ce3.png)

This appears to happen when OpenCASCADE fails to create the geometry.
Have tested this with OCC v7.3 and v6.9, as well as OCE latest version. All have the same behaviour.

Fixing the model element (via capping holes in the geometry and ensuring proper/correct 3D modelling etc... is done) does not cause this, however a better way of reporting bad geometry (if it fails to generate a model element) would be preferable to the whole program crashing with a segfault.

Cannot provide problem IFC files to replicate, are not property of mine.

aothms:
Thanks for reporting this. I am not completely certain of what's going on exactly.

The callstack you posted shows the `create_brep_for_representation_and_product()` function. That is called after [1] the `Logger::SetProduct(product);` call you refer to.

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/ffa282af9d50023a484db327e766ebf6e75d3c6d/src/ifcgeom/IfcGeomIterator.h#L477

Is it perhaps a combination of issues with this file? Exceptions during `boolean_operation()` (also in the callstack) are quite not that uncommon frankly. So there are quite a few try/catch statements throughout this callstack.

Sometimes Open CASCADE doesn't throw an exception, but really segfaults. Is that what is happening? What is the exact error in `NearestPoint()`? Is it a catchable exception?

hylandnp:
Thanks -- I think I misread the line the error was occurring on. It's probably occurring there, during geometry creation.

And yes, I am getting it reported as a segfault, no exception info, whether debugging in my IDE (debug and release builds) or running my code via the terminal, thus I cannot capture any more detailed information about the error/crash.

[...]

PR #573, opened 2019-04-17
Check builder safety

YouYue123:
Fix related to https://github.com/IfcOpenShell/IfcOpenShell/issues/488

### DIFF
--- a/src/ifcgeom/IfcGeomFunctions.cpp
+++ b/src/ifcgeom/IfcGeomFunctions.cpp
@@ -3726,7 +3726,9 @@ bool IfcGeom::Kernel::boolean_operation(const TopoDS_Shape& a, const TopTools_Li
 	builder->SetArguments(s1s);
 	copy_operand(b, B);
 	builder->SetTools(B);
-	builder->Build();
+	if(builder->BuilderCanWork()) {
+	 builder->Build();
+	}
 	if (builder->IsDone()) {
 		TopoDS_Shape r = *builder;
 


aothms:
Thanks for this, but I looked into this and this function does not appear to do what the name suggests. It returns `myBuilderCanWork` which is a private member only set to true once actually calling `Build()` [0]. So wrapping a call to `Build()` with an if statement like this will prevent all boolean operations.

[0] https://git.dev.opencascade.org/gitweb/?p=occt.git;a=blob;f=src/BRepAlgoAPI/BRepAlgoAPI_BooleanOperation.cxx;h=d381153f477f345e62c66380f1e4da64cca16178;hb=42da0d5115bff683c6b596e66cdeaff957f81e7d#l369

YouYue123:
I see. Thanks for finding this and this could be quite tricky for us to prevent all boolean operations. But calling Build() function itself will segment fault for some models.  What is your idea to prevent from it

aothms:
>  What is your idea to prevent from it

In the end we're dealing with C++ from external parties and we have to take into account the possibility of these kind of segmentation faults or stack overflows.

My long term idea is to make the IfcConvert process reentrant. Instead of directly writing to OBJ/DAE, the process would have a log file of the guids processed and store the geometry in some intermediate format on disk. When the process terminates with an error the next time the process continues where it left off, skipping over the element that caused issues.

YouYue123:
Hi Thomas any progress on this move? I recently found the file issue become quite often for 	builder->Build() funcitons

Issue #509, opened 2018-11-20
Styles of LayerSet


ninichki:
Hi,

![image](https://user-images.githubusercontent.com/1086380/48788223-a506f880-ecea-11e8-8bd4-412e478b8c26.png)

In the line 1937 the order of the surfaces is changed. Should'nt be the order of the styles changed, too?
Because I experience some wrong colors later in the 3D View.

kind regards,

Chris

ninichki:
Hi,
is it possible to upload somewhere an IFC file as example?
I wonder if
const int axis = usage->LayerSetDirection();
!= Axis2
should be captured and maybe return false at the moment?
kind Regards

aothms:
>  is it possible to upload somewhere an IFC file as example?

Preferably here: https://github.com/IfcOpenShell/files/upload/master

ninichki:
Uploads are disabled.
File uploads require push access to this repository.

aothms:
> Uploads are disabled.
> File uploads require push access to this repository.

Aw that's a bit of a inconsistency in the Github user interface. For edits they do silently create forks.

You'll have to press the [Fork] button manually then (top of page) and then press [Upload files] on your personal copy. Sorry for the inconvenience.

aothms:
Thanks for the file, it's here now as well: https://github.com/IfcOpenShell/files/blob/master/Allplan%202017%20Test%20Bauteil-Oberfl%C3%A4chen.ifc

I just now reworked the layerset operations. It should be quite a bit more performant. Next step is to investigate your file

aothms:
Ok sorry for taking a while, you're absolutely correct. Ideally I hope you can create PR for this to retain authorship information.


Regarding my earlier statements:

> The same would apply for thicknesses then as well probably.

`thicknesses` needs to be reversed as well. It's used during fold_layers() to offset the Axis from the neighbouring walls.

> I have some concerns as well on whether the orientation of the surfaces need to be changed in that case

That didn't seem to be necessary, both using the new `BRepAlgoAPI_Splitter` as well as the old iterative slicer.

> whether computing surface orientation by means of cross product is reliable enough or whether the placement y-axis should be used instead.

It seems the documentation explicitly mentions to use direction the Axis representation [1]. So the idea behind the code seems correct.


[1] http://www.buildingsmart-tech.org/ifc/IFC2x3/TC1/html/ifcmaterialresource/lexical/ifcmateriallayersetusage.htm

[...]

PR #594, opened 2019-05-27
reverse styles also when reversing surfaces

### LINKED ISSUE #545: Layerset slicing reverses order of materials on some walls
I have encountered an issue with layerset slicing that makes materials on some walls reversed in generated collada files. In archicad the walls have the same materials and the order is correct.

image

Output viewed in Unity. The colors should be in the same order on both walls.

The sample file used: 
[Hansaviertel 17-12 (2 wall sample).zip](https://github.com/IfcOpenShell/IfcOpenShell/files/2844952/Hansaviertel.17-12.2.wall.sample.zip)

Material layerset usages in the IFC:
```
#256= IFCMATERIALLAYERSET((#222,#239,#255),'405 Insulated Concrete Single Plastered');
#261= IFCMATERIALLAYERSETUSAGE(#256,.AXIS2.,.POSITIVE.,-0.);
#359= IFCMATERIALLAYERSETUSAGE(#256,.AXIS2.,.NEGATIVE.,0.);
```

#509 might be the same issue.

otsoa:
fixes #545 and possibly #509

### DIFF
--- a/src/ifcgeom/IfcGeomFunctions.cpp
+++ b/src/ifcgeom/IfcGeomFunctions.cpp
@@ -2025,6 +2025,7 @@ bool IfcGeom::Kernel::convert_layerset(const IfcSchema::IfcProduct* product, std
 	}
 
 	if (positive) {
+		std::reverse(styles.begin(), styles.end());
 		std::reverse(surfaces.begin(), surfaces.end());
 	}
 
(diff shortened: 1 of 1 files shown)


aothms:
Thanks, I accidentally forgot about this but it comes right in time as I have been planning to do new releases. As discussed, can you also reverse the thickness vector? https://github.com/IfcOpenShell/IfcOpenShell/issues/509#issuecomment-449366713

aothms:
Many thanks!

PR #627, opened 2019-06-18
Added support for the ifc4x2-draft

atibaut:
Support for the IFC version 4.2 Draft, IFC Bridge proposed extension

For generation of the five new C++ files the IFC 4.2 schema (IFC4x2.exp) was used from:
https://buildingsmart.sharefile.com/share/view/s619ceb7623d4b099/fod4202e-f039-49d7-8e4e-cddabe6efd3c


aothms:
Many thanks, I will take #628 though because that one is for v0.6.0 which allows multiple schemas simultaneously at runtime,

PR #631, opened 2019-06-20
catch Standard_ConstructionError when serialising xml attributes

otsoa:
F.e. If there is a zero length direction vector for IFCAXIS2PLACEMENT3D.


aothms:
Thanks, probably `Standard_ConstructionError` is appropriate. The more general case Standard_Failure would also be an option but I don't know if the code paths from XML serialization can trigger anything else besides construction errors.

Issue #608, opened 2019-06-06
Build under macOS 10.14.5 fails


kurb70:
Tried to build, but got only

```
This script fetches and builds IfcOpenShell and its dependencies

Script configuration:


* Target Architecture    = x86_64
 - Whether 32-bit (i686) or 64-bit (x86_64) will be built.
* USE_OCCT               = True
 - Compiling against official Open Cascade
* Dependency Directory   = /Users/kuba/dev/projects-bim/IfcOpenShell/build/Darwin/x86_64/10.9
 - The directory where IfcOpenShell dependencies are installed.
* Build Config Type      = RelWithDebInfo
 - The used build configuration type for the dependencies.
   Defaults to RelWithDebInfo if not specified.
* IFCOS_NUM_BUILD_PROCS  = 5
 - How many compiler processes may be run in parallel.

Building: pcre python libxml2 swig json boost occ IfcParse OpenCOLLADA IfcGeom IfcGeomServer IfcOpenShell-Python IfcConvert
using command log file '/Users/kuba/dev/projects-bim/IfcOpenShell/build/Darwin/x86_64/10.9/logs/20190606.log'
Collecting dependencies:
Found existing cmake-3.4.1, skipping
Found existing pcre-8.39, skipping
Found existing swig, skipping
Fetching occt-7.3.0p3...   
directory '/Users/kuba/dev/projects-bim/IfcOpenShell/build/Darwin/x86_64/10.9/build/occt' exists, skipping cloning
----------------------------------------------------------------------
error: pathspec 'V7_3_0p3' did not match any file(s) known to git

----------------------------------------------------------------------
Traceback (most recent call last):
  File "./build-all.py", line 532, in <module>
    revision="V" + OCCT_VERSION.replace('.', '_')
  File "./build-all.py", line 359, in build_dependency
    git_clone(download_url, target_dir=os.path.join(build_dir, download_name), revision=revision)
  File "./build-all.py", line 323, in git_clone
    run([git, "checkout", revision], cwd=target_dir)
  File "./build-all.py", line 281, in run
    raise Exception("Command `%s` returned exit code %d" % (" ".join(cmds), proc.returncode))
Exception: Command `git checkout V7_3_0p3` returned exit code 1

```

BTW The build works if the dependencies are already loaded locally.

aothms:
[...]

Issue #582, opened 2019-05-06
[WIP] conda-forge package


looooo:
Would be nice if a package is provided for conda-forge. I created a staged-recipe here: https://github.com/conda-forge/staged-recipes/pull/8277 but this still needs a lot of work.

I never used ifcopenshell, but freecad-users needs it. Maybe someone can help with making the build pass.

looooo:
ci's are stuck with:

linux, osx: no idea cmake 166: `unable to find header files`
windows: occ not found

aothms:
Thanks this is great. Did you see the conda recipe in the 0.6.0 branch? https://github.com/IfcOpenShell/IfcOpenShell/tree/v0.6.0/conda We chose to publish 0.6 https://anaconda.org/ifcopenshell/ifcopenshell because it allows multiple schemas at runtime. But I think freecad is still on master/0.5. 

Where does the occt dependency come from?

looooo:
> Did you see the conda recipe in the 0.6.0 branch? 

yes I used it as the base for this recipe

> We chose to publish 0.6 https://anaconda.org/ifcopenshell/ifcopenshell because it allows multiple schemas at runtime. But I think freecad is still on master/0.5.

There is  no 0.6.0 on the release page (github). Where can I find the source?

> Where does the occt dependency come from?

I would like to use the same occ-library for ifcopenshell and freecad. Therefor I would like to use occt.

aothms:
> Therefor I would like to use occt.

Sounds good to me. Especially if it's a recent version. I looked it up on conda-forge and it seems the include files are in a folder `opencasade/` but the cmake command you provide searches for `occt/` https://github.com/conda-forge/staged-recipes/pull/8277/files#diff-cf0a4024c796226f0664f8b7552cc02bR10

> Which compiler should I use for windows?

Preferably the compiler that Python was built with, but this is always an issue on Windows. Not sure how conda-forge manages that. But I think it's MSVC 2015 for the recent versions of Python3.

> I see cases where windows use static libs, would it be possible to make this more consistent with other platforms to link dynamically.

Yes, this is slowly getting fixed in the 0.6 branch.

aothms:
@looooo are you planning to continue working on this? It aligns a bit with my aims to create new tags in the master (0.5.0-rc1) and 0.6 (0.6-beta1) branches soon. So if we coordinate this a bit we might be able to push something recent and stable to conda-forge. Let me know what I can do to help.

[...]

PR #633, opened 2019-06-23
Updated *nix build script to also support git pull repository for dependencies + Support for IFC4x1 added

atibaut:
Motivated by issue #608 to improve the method def git_clone(...) to the def git_clone_or_pull_repository(...). 
Also, dependencies versions were updated.


aothms:
Thanks this looks really comprehensive. I have some doubts though about the update to Swig 4. It seems there are some assumptions in the code that won't work with that version, see https://github.com/IfcOpenShell/IfcOpenShell/issues/582#issuecomment-497789849 Were there specific reasons for that particular update or can you revert that?

atibaut:
@aothms reverted to SWIG_VERSION=4.0.0.

aothms:
Thanks, in addition I cannot find the libxml2 version you specified at ftp://xmlsoft.org/libxml2/ so the build currently fails. Where did you find it? `2.9.9` is there, but not `2.9.9_2`.

aothms:
I'll change it back to `2.9.9` after merging. Thanks again for your efforts on this.

PR #638, opened 2019-07-02
Multithreading base work

YouYue123:
This pull request is related to the multithreading work, this is based on Sander's previous work in version 0.5 and Airsquire changed to version 0.6

Known issue:
The 2x3 kernel is crashing while using multi-threading. Thomas, please provide your change about how you fix the kernel issue.


aothms:
Thanks, see #639 for my fixes.

PR #639, opened 2019-07-03
Multithreaded

aothms:


aothms:
Frankly results are not what we had hoped for. At least on Windows I'm not seeing great improvements beyond two threads. Not unlikely this is due to poor performance in multithreaded memory management. We should see if using https://github.com/intel/tbb alleviates this.

YouYue123:
Hi Thomas,

I will test it this week on Linux and also will try TBB if I got time

YouYue123:
Hi Thomas,

There is a crashing of segment fault for my test IFC model.

I realised your version is based on 0.6.0b0. Our work is based on 0.6.0a0. 

May I know the difference between 0.6.0a0 and 0.6.0b0?  And would that be the reason which cause segment fault in the geometry creation we are facing?

Thanks

YouYue123:
Hi,

I found the problem. The segment fault is related to 

https://github.com/aothms/IfcOpenShell/blob/3ebca0516e8484b7369e03737da23dcba31ce45f/src/ifcconvert/IfcConvert.cpp#L807

Sometimes the "geom_object pointer" is nullptr which creates issues of the Brep or Trianglelation. A safety pointer check fix the issue for me

Hope this helps.

Thanks

YouYue123:
Fernando [3:36 PM]
Performance wise,

The multithreading does help us to increase speed without TBB.

For the geometry processing part,

```c++
    collect();
    process_concurrently();
```

the benchmark is as below by using the IFC file I shared with you

- 2 cores -- 106 seconds
- 16 cores -- 34 seconds

The only bottleneck part is the serialisation for each element  in the iterator do-while loop. Is that possible to multithread it as well?

aothms:
> 2 cores -- 106 seconds
> 16 cores -- 34 seconds

I think that matches what I observed. 8x more threads 3x better performance. Not linear.

> The only bottleneck part is the serialisation for each element in the iterator do-while loop. Is that possible to multithread it as well?

I don't think it'll make a big difference. We have to profile the code to be sure. But (a) most serializes are probably not thread safe. (b) Performance is most likely bound by the I/O speed.

Amoki:
It's 16 real cores or 16 threads on an 8 cores CPU with SMT/HT?
If HyperThreaded cores are not much useful, 4x more threads for 3x better performance is not that bad.

aothms:
> It's 16 real cores or 16 threads on an 8 cores CPU with SMT/HT?
>  If HyperThreaded cores are not much useful, 4x more threads for 3x better performance is not that bad.

Thanks, good point! Probably computation involves mostly floating point math, which indeed is probably what is shared between two cores.

> Sometimes the "geom_object pointer" is nullptr which creates issues of the Brep or Trianglelation. A safety pointer check fix the issue for me

Thanks I will use a `copy_if()` to build the full vector and filter out the nullptr there.

There's something else going on as well in the multi-threaded code path, as for one file, 2 threads is almost twice as slow as single threaded. 

TBB is indeed faster, but doesn't change multithreading efficiency fundamentally. After building with TBB, you need to set an env var as well to have OCCT actually use it. `set MMGT_OPT=2`. 

Conversion times on linux on an EC2 machine. Note that there is a bug in the current PR, the num_threads_ is not actually used to determine size of the threadpool.

exe|ifc file|n threads|wall clock time
---|---|---|---
./IfcConvert|Office_A_20110811_optimized.ifc|1|5.51661503315
./IfcConvert|Office_A_20110811_optimized.ifc|2|4.33341443539
./IfcConvert|Office_A_20110811_optimized.ifc|4|2.71475338936
./IfcConvert|Office_A_20110811_optimized.ifc|8|1.98269963264
./IfcConvert|Office_A_20110811_optimized.ifc|16|1.64179551601
./IfcConvert|Clinic_MEP_20110906_optimized.ifc|1|118.392883539
./IfcConvert|Clinic_MEP_20110906_optimized.ifc|2|206.934801459
./IfcConvert|Clinic_MEP_20110906_optimized.ifc|4|114.919862032
./IfcConvert|Clinic_MEP_20110906_optimized.ifc|8|75.8054435253
./IfcConvert|Clinic_MEP_20110906_optimized.ifc|16|64.2108564377
./IfcConvert-tbb|Office_A_20110811_optimized.ifc|1|5.33473241329
./IfcConvert-tbb|Office_A_20110811_optimized.ifc|2|4.0364664793
./IfcConvert-tbb|Office_A_20110811_optimized.ifc|4|2.56830000877
./IfcConvert-tbb|Office_A_20110811_optimized.ifc|8|1.89747595787
./IfcConvert-tbb|Office_A_20110811_optimized.ifc|16|1.56096851826
./IfcConvert-tbb|Clinic_MEP_20110906_optimized.ifc|1|112.832552552
./IfcConvert-tbb|Clinic_MEP_20110906_optimized.ifc|2|190.200834513
./IfcConvert-tbb|Clinic_MEP_20110906_optimized.ifc|4|104.962541461
./IfcConvert-tbb|Clinic_MEP_20110906_optimized.ifc|8|70.1025134325
./IfcConvert-tbb|Clinic_MEP_20110906_optimized.ifc|16|58.9066236019

aothms:
~~~
./IfcConvert|Office_A_20110811_optimized.ifc|1|5.60256052017
./IfcConvert|Office_A_20110811_optimized.ifc|2|3.52646994591
./IfcConvert|Office_A_20110811_optimized.ifc|4|2.93937051296
./IfcConvert|Clinic_MEP_20110906_optimized.ifc|1|121.545045495
./IfcConvert|Clinic_MEP_20110906_optimized.ifc|2|73.8505574465
./IfcConvert|Clinic_MEP_20110906_optimized.ifc|4|61.6651084423
~~~

I located the issue of worse performance with two threads. It was due to triangulations not being properly reused in case of geometrical instances (IfcMappedItem).

Results printed below are on a much less powerful machine, which explains the diminished results for 4 threads.

Performance is now as expected. Parsing and serialization are single threaded. It's not that times for these are insignificant, but there is likely little to gain as these are likely I/O bound.

Unless one of you wants to invest additional effort, I'll likely merge this as is.

YouYue123:
@aothms I agree with you. I/O bound would be a hard constrain for that part. Feel free to merge this back. Thanks for the effort to make this happen.

ssg3d:
I tried to use the multi threading functionality by instantiating IfcGeom::Iterator<real_t> with -1. When in single threaded mode, IfcGeom::Iterator:: initialize is fast and defers all work to IfcGeom::Iterator::next. In MT mode, IfcGeom::Iterator:: initialize finishes everything and hence we have to wait till everything is processed before we can access the geom with IfcGeom::Iterator::next. I was hoping for each geometry to be processed independently so that I can spawn off multiple threads and immediately start accessing geom as and when its thread finishes. I am ok to do the MT framework myself, but am looking for some interface by which I can call some function that creates geom for an element independently in a thread.

aothms:
@ssg3d can you create a new issue for this so that we can discuss.

Now that we're in the 0.7 branch we can also open up the discussion a bit with respect to C++ version and e.g async and coroutines to get a better API. In the 0.6 branch we didn't want to move to a newer C++ standard just for this feature. Also we wanted to have the same API regardless of threads, but as you said it does mean that the workloads in the functions have shifted.

I didn't give this a lot of thought. But it could perhaps also be possible in the current code to call initialize() on a new thread. And then inspect the `tasks_` variable (currently private) [0] to get incremental progress?

Or, as said, we can have a fresh look at what's possible these days with co routines.

[0] https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcgeom/IfcGeomIteratorImplementation.h#L178

PR #641, opened 2019-07-16
Blender ifc 2.80rc1 collections enhancement

elmopl:
This fix helps with importing by creation nested collections according to layers that can be retrieved from IFC file.

image

### DIFF
--- a/src/ifcblender/io_import_scene_ifc/__init__.py
+++ b/src/ifcblender/io_import_scene_ifc/__init__.py
@@ -43,6 +43,7 @@ if "bpy" in locals():
 
 import bpy
 import mathutils
+from collections import defaultdict
 from bpy.props import StringProperty, IntProperty, BoolProperty
 from bpy_extras.io_utils import ImportHelper
 
@@ -74,7 +75,7 @@ def import_ifc(filename, use_names, process_relations, blender_booleans):
     if not valid_file:
         return False
     print("Done reading file")
-    id_to_object = {}
+    id_to_object = defaultdict(list)
     id_to_parent = {}
     id_to_matrix = {}
     openings = []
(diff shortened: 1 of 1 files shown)


aothms:
Many thanks for this! Sorry for the delay. I hope to be able to have a look in the coming week.

Moult:
I have tested a quick import and it seems to work great! See attached import.

2019-09-03-213145_1214x1048_scrot

It also seems as though we have both figured that Blender's collections feature does a great job of spatial decomposition! See: https://github.com/IfcOpenShell/IfcOpenShell/tree/v0.6.0/src/ifcblenderexport

If @aothms is OK with merging this, I'd be happy to do so, along with test it on a real project that we're doing in Blender right now :) And perhaps implement more tricks such as reusing mesh instances, and an option for putting the IFC class in the name as well as a custom property (it isn't necessary, but we found when it helps modelers stay organised and navigate the outliner)

aothms:
> If @aothms is OK with merging this

Yes, yes, yes.

> And perhaps implement more tricks such as reusing mesh instances

There is a bug related to that. #661 https://sourceforge.net/p/ifcopenshell/discussion/1782718/

elmopl:
> And perhaps implement more tricks such as reusing mesh instances

Looking at code it seems like it does that when loading (or am I missing something?)

        # MESH CREATION
        # Depending on version, geometry.id will be either int or str
        mesh_name = 'mesh-%r' % ob.geometry.id
        if mesh_name in bpy.data.meshes:
            me = bpy.data.meshes[mesh_name]
        else:

> IFC class in the name as well as a custom property

I was wondering if that would be helpful. This is the first time I am exposed to IFC files at all, so I don't really know all the quirks and relations you can store in that file.
I was wondering how much information is lost and whether it would be possible to pull out all data that IFC file contains so that you could re-save it without losing any data (or ideally without changing internal structure at all).

From other things I was also thinking about adding extra level  collection `geometry`.
Currently I found it really annoying when I loaded a model of house that IIRC I couldn't easily hide ceiling bits as they were in structure above collection with walls.
I could make that change before this gets merged (or after, don't really mind).

Moult:
Also cherry picked the commits into v0.6.0 branch.

PR #658, opened 2019-08-26
Casting default timestamp template to integer

gpioto:
The documentation of IFC indicates that the timestamp must be of type Integer. [IfcTimeStamp](https://standards.buildingsmart.org/IFC/DEV/IFC4_2/FINAL/HTML/link/ifctimestamp.htm)
If not, it cause some problems importing the generated IFC file in some platforms


aothms:
Thanks for this! I also have to look into the underlying problem though. Generally it's not possible to supply incorrectly typed attributes from Python, but there is some compatibility between integers and floats. But this should not be one of the outcomes.

PR #662, opened 2019-09-04
Fix reusing meshes and mesh naming

elmopl:
This fixes bug I introduced in 7c99294ba815f58176a4858e60defefa68fee0a1

Below view of editing a mesh that is used multiple times within same model. This verifies that indeed mesh is shared.

Model is from http://openifcmodel.cs.auckland.ac.nz/Model?sortField=Name&sortDirection=ascending&page=4 (AISC_Sculpture_param.ifc)

image


Moult:
Cherry picked into v0.6.0 branch.

PR #672, opened 2019-09-13
Use model-offset in IfcGeom::Kernel and add model-rotation

birgerbr:
Applying model-offset in IfcGeom::Kernel makes it work independently
of the serializer.

`model-rotation` was also added to support rotation in the same
manner as `model-offset` supports offset.

### DIFF
--- a/src/ifcconvert/GeometrySerializer.h
+++ b/src/ifcconvert/GeometrySerializer.h
@@ -56,11 +56,11 @@ public:
     SerializerSettings()
         : precision(DEFAULT_PRECISION)
     {
-        memset(offset, 0, sizeof(offset));
+        memset(serializer_offset, 0, sizeof(serializer_offset));
     }
 
     /// Optional offset that is applied to serialized objects, (0,0,0) by default.
-    double offset[3];
+    double serializer_offset[3];
 
     /// Sets the precision used to format floating-point values, 15 by default.
     /// Use a negative value to use the system's default precision (should be 6 typically).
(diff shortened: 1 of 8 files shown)


aothms:
> Applying model-offset in IfcGeom::Kernel makes it work independently of the serializer.

Thanks that's indeed better.

> model-rotation was also added to support rotation in the same

Cool thanks for this.

It appears there are some compile errors on gcc wrt brace initialization https://travis-ci.org/IfcOpenShell/IfcOpenShell/builds/584613362#L2125

I think it'd be better to try and merge this into v0.6.0. We can switch to C++14 there (if not already). Would that be an option for you?

birgerbr:
> > Applying model-offset in IfcGeom::Kernel makes it work independently of the serializer.
> 
> Thanks that's indeed better.
> 
> > model-rotation was also added to support rotation in the same
> 
> Cool thanks for this.
> 
> It appears there are some compile errors on gcc wrt brace initialization https://travis-ci.org/IfcOpenShell/IfcOpenShell/builds/584613362#L2125
> 
> I think it'd be better to try and merge this into v0.6.0. We can switch to C++14 there (if not already). Would that be an option for you?

Glad to hear that you like the changes :blush:  I've created a fixup that works with the gcc 4.8.
You can decide if it should go to master or v0.6.0.
Thank you for your work on this project. Really appreciate it :+1:

aothms:
Superseded by #788. As expressed there. Many thanks for this effort and sorry for the delay in merging.

PR #694, opened 2019-10-17
fix: python-occ color length check

Demetrio92:
should be OR, not AN;  color is spelled without U everywhere else; exception call itself was breaking


aothms:
Ha thanks, good catch!

PR #727, opened 2019-11-19
IfcPolygonalFaceSet management

stephassoria:
added IfcPolygonalFaceSet convertion to TopoDS_Shape


aothms:
Many thanks, but coincidently this was just merged as part of #714.

Regarding you 2nd commit. Do you know if the inheritance from unary- and binary_function in schema.h actually necessary? Or is the operator() on those structs sufficient? These definitions are probably removed from std for a reason after all.

stephassoria:
You are right, it seems that removing the inheritance from unary- and binary_function in IfcSchema.h is correct. I have submit a new commit.

stephassoria:
The travis-ci build have failed because of the merge #714, I let you see if you can just merge my last commit and close the pull request.

PR #731, opened 2019-11-25
IfcSchema.h c++ 17 compatibility

stephassoria:
removed obsolete unary_function and binary_function inheritance


aothms:
Many thanks!

Issue #736, opened 2019-12-06
Conflict with OCE 7.4.0 in geom module


derbosper:
There is a function `has_occ()` in geom/main.py that checks if OCC is installed in the environment, which tries to `import OCC.BRepTools` and returns `False` when `ModuleNotFoundError` occurs. Now it failed even if I had the OCC installed. It seems like the old import structure is deprecated and now it supports only `import OCC.Core.BRepTools`.  I've updated all the necessary imports by adding the `.Core` and it basically solved the problem, but I wanted to let you know here :)

aothms:
Great thanks for reporting. Do you think this is something that can made compatible? We'll probably need to use `import * from OCC.Core.gp` kind of imports for both with and without `Core` with `try catch`. What do you think? Pull requests welcome but don't feel obliged.

mieszk:
Yes, if we want to support the old version, they both can be wrapped (with double `try-catch` for sure, maybe there is another possibility). Also, I can submit a PR.

aothms:
try-catch is probably the 'pythonic' way to do this. Personally I would probably have a preference for something `importlib.import_module()` to have import based on dynamic strings, i.e. with and without `.Core.`.

CyrilWaechter:
@derbosper Seeing related merges. Should this issue be closed now ?

derbosper:
@CyrilWaechter Yes, I guess, seems like it's no longer an issue.

PR #737, opened 2019-12-08
Fix OCC imports in Python

mieszk:
Here is a proposed solution to #736. I changed the `has_occ()` and imported selected modules from `OCC` depending on what's currently available. I also think it's a good idea to move all the imports to the top, because with each import the whole `OCC.Core` module is being loaded anyway (except for `OCC.Display` maybe, because it's a separate one).

And also, `importlib.import_module()` does the same thing as `import` and afaik it does't provide any extra features that might be useful here.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/geom/__init__.py
+++ b/src/ifcopenshell-python/ifcopenshell/geom/__init__.py
@@ -21,5 +21,25 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-from . import occ_utils as utils
+def _has_occ():
+    try:
+        import OCC.Core.BRepTools
+        return True
+    except ImportError:
+        pass
+
+    try:
+        import OCC.BRepTools
+        return True
+    except ImportError:
+        pass
+
+    return False
+
+
+has_occ = _has_occ()
+
+if has_occ:
+    from . import occ_utils as utils
+    
 from .main import *
--- a/src/ifcopenshell-python/ifcopenshell/geom/main.py
+++ b/src/ifcopenshell-python/ifcopenshell/geom/main.py
@@ -33,13 +33,13 @@ def has_occ():
     try:
         import OCC.Core.BRepTools
         return True
-    except ModuleNotFoundError:
+    except ImportError:
         pass
 
     try:
         import OCC.BRepTools
         return True
-    except ModuleNotFoundError:
+    except ImportError:
         pass
 
     return False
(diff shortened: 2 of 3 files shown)


aothms:
Many thanks!

tpaviot:
@mieszk The pythonocc package structure has changes a long time ago (I would say one year and a half), since the previous release 0.18.1. Now, in the latest 7.4.0 release (currently beat 2 actually), the base modules are located in ```OCC.Core```, at the same level than ```OCC.Display``` or  ```OCC.Extend```.

Porting up to 7.4.0beta is quite easy : replace all ```from OCC.xxx import``` with ```from OCC.Core.xxx```, as you did in your PR.

I noticed you try to maintain a kind of backward compatibility. Rather than testing imports using ```try/except``` blocks, you can also use the ```OCC.VERSION```:
```
>>> import OCC
>>> OCC.VERSION
'7.4.0-beta2'
>>> dir(OCC)
['PYTHONOCC_VERSION_DEVEL', 'PYTHONOCC_VERSION_MAJOR', 'PYTHONOCC_VERSION_MINOR', 'PYTHONOCC_VERSION_PATCH', 'VERSION', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'pythonoccVersionNumberError', 'require_pythonocc_version', 'test_require_pythonocc_version', 'unittest']
>>> OCC.PYTHONOCC_VERSION_MAJOR
7
[...]
```

Important Note: in the last pythonocc release, we definitely dropped python2 support.

tpaviot:
@aothms This branch was branch into ```master``` which seems to be not up-to-date with latest devs. If i test the ```v0.6.0 branch```, I can't benefit from this fix. Could you please rebase your ```v0.6.0``` branch on top of master ? or, if too difficult, duplicate the commit in the v0.6.0 branch ?

mieszk:
@tpaviot Thanks, fair points. I tried to make it compatible with old version as discussed in #736. Also, @aothms, what's you opinion on python2 support? Do you think it's necessary or can it be dropped some day (as it will reach EOL in three weeks)?

aothms:
@tpaviot Of course, just merged master into v0.6.0

@mieszk re py2 compatibility. Well let's keep things the way they are for 0.5.0 and 0.6.0. God knows who still depends on this. New code in 0.7.0 can be py3 only from my point of view.

SWIG4 compatibility has higher priority for me, as brought up somewhere else by @tpaviot

PR #743, opened 2019-12-16
Fix: No XML serializer registered for IFC4X1

AdriScho:


AdriScho:
Note : It would be nice to add some test files in IfcOpenShell/files. I will check in the examples given in the official 4X1 spec (mostly alignement examples)

aothms:
Many thanks, looks comprehensive! Indeed, would be nice to test the infra schemas as well.

PR #752, opened 2020-01-06
Clash testing on 1 file

hanslammerts01:
Make it possible to clash detect within a single IFC file on different objects.


Moult:
Re-created this request as a bugreport, not a PR: https://github.com/IfcOpenShell/IfcOpenShell/issues/753

Issue #768, opened 2020-01-20
IfcGeom::Kernel::wire_intersections fails for occt 7.4.0


tpaviot:
latest ifcopenshell python from v0.6.0 branch fails at loading/displaying any ifc file. No debug information, only a segfault.

After an investigation using valgrind on Unbuntu 18.04, the method
 ```c++
IfcGeom::Kernel::wire_intersections
```

seems to be the cause. valgrind reports the failure at the line 

```c++
GeomAPI_ExtremaCurveCurve ecc(
				BRep_Tool::Curve(wd->Edge(i + 1), u11, u12),
				BRep_Tool::Curve(wd->Edge(j + 1), u21, u22)
			);
```
I modified this function to return ```false``` in any case, and could load any 3d representation as it used to work with previous versions.

sanderboer:
I think I fixed it by changing the class instantiation to:

`
static Handle(Geom_Curve) c0 = BRep_Tool::Curve(wd->Edge(i + 1), u11, u12);
static Handle(Geom_Curve) c1 = BRep_Tool::Curve(wd->Edge(j + 1), u21, u22);
GeomAPI_ExtremaCurveCurve ecc(c0, c1);
`

I had this issue crash blender and freecad when importing a door I made in freecad.

[segfault_door.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/4148331/segfault_door.zip)

aothms:
Interesting. I haven't had a chance yet to look into 7.4.

According to the Docs:
> Curve()
> Returns the 3D curve of the edge. May be a Null handle. In <First> and <Last> the parameter range. It can be a copy if there is a Location.

The copy in case of location could be a clue, but since the Handle should behave like a smart pointer it shouldn't make a difference.

@sanderboer does the `static` make a difference or is a normal local variable sufficient?

@tpaviot would be great if you can verify the fix from @sanderboer on your files

sanderboer:
occt refs:

`GeomAPI_ExtremaCurveCurve (const Handle< Geom_Curve > &C1, const Handle< Geom_Curve > &C2)
`

`static Handle< Geom_Curve > Curve (const TopoDS_Edge &E, Standard_Real &First, Standard_Real &Last)
`

I tried omitting the 'static' storage class specifiers alltogether, segfault, replaced it with 'const', segfault. Only the 'thread_local' and 'static' specifiers don't segfault. 

[...]

PR #784, opened 2020-02-07
Proposed fix for #768

sanderboer:
This is a duplicate of a previous PR, I am deleting the first one. My v0.6.0 branch will a merge of a number of fixes and this pr is cherrypicked.


aothms:
Thanks I need to understand this a bit better. This is a recursive function so making this static or thread_local means these curve handles are shared between recursion levels (?) That is counter-intuitive to me.

sanderboer:
Maybe it is better to do this ?
```
thread_local Handle(Geom_Curve) c0;
c0 = BRep_Tool::Curve(wd->Edge(i + 1), u11, u12);
thread_local Handle(Geom_Curve) c1;
c1 = BRep_Tool::Curve(wd->Edge(j + 1), u21, u22);
```
I am unclear if c++ allocates new memory if you initialize a static (or thread_local) variable with the result of a function. I would think the code above is equivalent with the patch, but it's still c++.

sanderboer:
Wait, I didn't read your comment right, I was worried about memory leaking. 
It is possible that the last value of c0 and c1 carry over to the next recursion level. But that value is clobbered with the new value of BRep_Tool::Curve.

src/ifcgeom/IfcGeomFunctions.cpp
```diff
@@ -3691,6 +3691,8 @@ bool IfcGeom::Kernel::wire_intersections(const TopoDS_Wire& wire, TopTools_ListO
 	}
 	
 	bool intersected = false;
+        thread_local Handle(Geom_Curve) c0;
+        thread_local Handle(Geom_Curve) c1;
```
sanderboer: Turns out, this is a not smart move. This reintroduces the segfault.

aothms:
This was addressed by the check on `IsParallel`. But many thanks for the investigations.

PR #788, opened 2020-02-11
Use model-offset in IfcGeom::Kernel and add model-rotation

Blocovin:
I merge the pull request #672 into v0.6.0

And fix --center-model to makes it work independently of the serializer.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -345,8 +345,8 @@ int main(int argc, char** argv) {
 			"Order the elements using their IfcBuildingStorey parent. "
 			"Applicable for DAE output.")
         ("center-model",
-            "Centers the elements upon serialization by applying the center point of "
-            "all placements as an offset. Applicable for OBJ and DAE output. Can take several minutes on large models.")
+            "Centers the elements by applying the center point of all placements as an offset."
+            "Can take several minutes on large models.")
         ("model-offset", po::value<std::string>(&offset_str),
             "Applies an arbitrary offset of form 'x;y;z' to all placements.")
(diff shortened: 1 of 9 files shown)


Stinkfist0:
Minor note: this make the option/setting a geometry setting now instead of serializer option so the code and documentation would be good to be adjusted accordingly.

Blocovin:
> Minor note: this make the option/setting a geometry setting now instead of serializer option so the code and documentation would be good to be adjusted accordingly.

I've updated the documentation

Stinkfist0:
Great, also note that it's good to move this under `geom_options` instead of `serializer_options` in the code.

Blocovin:
> Great, also note that it's good to move this under `geom_options` instead of `serializer_options` in the code so that they appear under the right section when `--help` is used.

Oh, sorry ! I didn't realize there were two sections. So I moved this under `geom_options`.

Blocovin:
@aothms Is there a problem for accepting the merge request ?

aothms:
@Blocovin sorry for the delay. This is definitely a wonderful PR, adding functionality and cleaning up existing (std::array usage, conversion to gp_Trsf in kernel for more general usage). It's just that it took me some time to review as I have the bad habbit of going through the PRs top to bottom in github (so in reverse order) and I have been too busy.

The only issue I can think of is that it doesn't somehow fail on 0;0;0;0. I'd expected OCCT to throw a ConstructionError but it actually happily creates a gp_Mat with NaNs ending up all the way in the serialized output of the matrices. Perhaps that should be fixed upstream in OCCT. I don't think it's a blocker for this PR.

Wonderful work. Hope to welcome your contributions more.

PR #799, opened 2020-02-26
Bugfix/multi material obj

AlaaB2020:
choosing first material in multi-layer material object in case --enable-layerset is not configured.


aothms:
Thank you very much for this. Sorry for the delay. This is now merged in the v0.6.0 branch [0] but is behind a setting `--layerset-first`.

[0] https://github.com/IfcOpenShell/IfcOpenShell/commits/v0.6.0?author=AlaaB2020

Issue #744, opened 2019-12-17
seg fault on ifcopenshell.geom.create_shape(settings, product) 


berndhahnebach:
I got hands on a huge ifcfile (ca 100 MB) exported from Cadmatic. I have never had files from them before. A lot of products make ifcopenshell ends up in a segmentation fault on  ifcopenshell.geom.create_shape(settings, product) 

As a resulte of https://github.com/IfcOpenShell/IfcOpenShell/issues/564 I was able to extract just one of the results in a separate IFC file. I will go for a PR on files repo and post the file link here afterwards. 

code to reproduce the crash:
```
import ifcopenshell
from ifcopenshell import geom
settings = ifcopenshell.geom.settings()
settings.set(settings.USE_BREP_DATA,True)
settings.set(settings.SEW_SHELLS,True)
settings.set(settings.USE_WORLD_COORDS,True)

ifcfile = 'C:/Users/BHA/Downloads/743-seg-fault-geometry.ifc'
f = ifcopenshell.open(ifcfile)

p = f[1944]
print(p)
cr = ifcopenshell.geom.create_shape(settings, p)

```

BTW: ifc++ gave the following warning
Warning: triangulateLoops, face_loops.size() == 0, IFC entity: #1941=IfcExtrudedAreaSolid

link to topic on FreeCAD forum
https://forum.freecadweb.org/viewtopic.php?f=39&t=41600&p=354486#p354483

cheers bernd

berndhahnebach:
more information:
if 
`#1942=IFCSHAPEREPRESENTATION(#11,'Body','SolidModel',(#488,#936,#1400,#1848,#1855,#1941));`
is exchanged with
#1942=IFCSHAPEREPRESENTATION(#11,'Body','SolidModel',(#488,#936,#1400,#1848,#1855));
ifcopenshell and FreeCAD do import the geometry. 

Thus the problem seams to be on the IFCEXTRUDEDAREASOLID #1941, the one ifc++ claimed about to have a zero face_loops.size()

@aothms 
If there is a problem in one of the representations of a product, would it be possible to just ignore this? This is what ifc++ and BIMCollabZoom seam to do.

sanderboer:
I found that the issue is in IfcGeomFaces.cpp around line 475, OpenCascade segfaults when trying to add a hole of very small size.

I created a stopgap in that function that checks for the length of that wire and ignores it if it is smaller than 0.0001.
Think this check is better suited in convert_wire ?
[...]

PR #801, opened 2020-03-01
proposed fix for #744

sanderboer:
Issue with #744 is that some programs export full circles as trimmed circles, with trimparams at 0 and 2*pi. But due to floating point precision this trim could flip from full circle to an infinitesimally small line segment.


berndhahnebach:
For some reason travis fails.

I am not sure what caused travis to fail ?!?

aothms:
Many thanks for this @sanderboer. Sorry for the long delay in replying. I'm not entirely sure, but I'm guessing [0] the return of fmod() could be negative in this case, so we probably need to wrap it in std::abs() (which is what ALMOST_THE_SAME also would have done.

[0] http://cpp.sh/9a6s2

aothms:
I went with a slightly different fix, but many thanks for the investigations. d81196316d1a215f3a9395170479ab9ecf3dac81

PR #817, opened 2020-03-30
Make validate module write to logger

Moult:
This changes two things:

 1. Adds a try/catch for validating inverse relationships, otherwise you only get the first error.
 2. Remove hardcoded print statement and use a logger instead. This allows you to use it as a function in a library with an error log or otherwise. Note that by default if you run it from a command line it will still write to stream output.


aothms:
Thanks, perfect. I ran into this uncaught exception as well, surprised I didn't seem to have pushed that.

Issue #755, opened 2020-01-08
Create documentation for IfcOpenShell, BlenderBIM, et all


Moult:
I have just started a Sphinx docs for BlenderBIM: https://github.com/IfcOpenShell/IfcOpenShell/commit/e927ac49ac4afaed6a58c4fbad4a27d025f7e045

I will need to write docs for BlenderBIM's various features. However, I will also need to write documentation on how to use utilities like IfcClash, IfcDiff, IfcCOBie, BIMTester... so I figured maybe it is worthwhile proposing that instead of me just creating documentation for these alone, I should create a new Sphinx docs for IfcOpenShell, with subcategories for BlenderBIM, IfcMax, IfcConvert, IfcClash, BIMServer-plugin, IfcOpenShell-python, etc - then the IfcOpenShell website could be updated with a link to this central documentation location, and people can start contributing.

Thoughts? Good / bad idea to combine docs? Should docs be hosted on `ifcopenshell.org/docs/`?

In addition, I'm not sure where the source code is for the IfcOpenShell.org site, perhaps it could be updated with IfcBlender changed to "BlenderBIM" and linking to blenderbim.org to prevent confusion?

aothms:
Sphinx is only Python (am I correct?) but obviously a great starting point. Or are you suggesting to use https://breathe.readthedocs.io/en/latest/ ? 

> IfcBlender changed to "BlenderBIM" and linking to blenderbim.org to prevent confusion?

I will.

Moult:
I guess there's two aspects - there's code documentation and user documentation. For user documentation, Sphinx just parses `.rst` files which you write yourself. There is a ton to write about from a user perspective for BlenderBIM and its utilities. This is what I had started to write.

For code documentation, Sphinx I believe out of the box parses docstrings which is Python only, but yes the Breathe project looks like a very cool Sphinx extension that parses doxygen output. That would help document the C++ side of things :)

If you think it is a good idea to start writing these docs I can get started :) Also let me know if you think it's a good idea to combine all apps under IfcOpenShell's umbrella into a single Sphinx-based docs, or keep them separate?

Moult:
Docs hosted now on https://blenderbim.org/docs/ (still work-in-progress)

If you think it makes sense to combine the docs into a single master doc, I can restructure the pages and it makes more sense to host it on the ifcopenshell website, not the blenderbim website.

htlcnn:
[...]

PR #822, opened 2020-04-02
Add sphinx docstrings to ifcopenshell python binding

Moult:
I tried to follow the relevant sphinx docstring conventions and PEP conventions when doing this.

See the results here:

https://blenderbim.org/docs/ifcopenshell-python/api-documentation.html

Please let me know if I made any mistakes in the documentation :) I also didn't quite understand what the add() function does.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/file.py
+++ b/src/ifcopenshell-python/ifcopenshell/file.py
@@ -101,27 +101,29 @@ class file(object):
             return entity_instance(self.wrapped_data.by_guid(str(key)))
 
     def by_id(self, id):
-        """Return IFC objects filtered by IFC ID.
+        """Return an IFC entity instance filtered by IFC ID.
 
         :param id: STEP numerical identifier
         :type id: int
-        :returns: Returned objects are unwrapped IFC objects.
+        :returns: An ifcopenshell.entity_instance.entity_instance
         :rtype: ifcopenshell.entity_instance.entity_instance
         """
         return self[id]
 
     def by_guid(self, guid):
-        """Return IFC objects filtered by IFC GUID.
+        """Return an IFC entity instance filtered by IFC GUID.
 
         :param guid: GlobalId value in 22-character encoded form
         :type guid: string
-        :returns: Returned objects are unwrapped IFC objects.
+        :returns: An ifcopenshell.entity_instance.entity_instance
         :rtype: ifcopenshell.entity_instance.entity_instance
         """
         return self[guid]
 
     def add(self, inst):
-        """Undocumented function"""
+        """Adds an entity including any dependent entities to an IFC file.
+        
+        If the entity already exists, it is not re-added."""
         inst.wrapped_data.this.disown()
         return entity_instance(self.wrapped_data.add(inst.wrapped_data))
 
@@ -138,7 +140,7 @@ class file(object):
         return [entity_instance(e) for e in self.wrapped_data.by_type(type)]
 
     def traverse(self, inst, max_levels=None):
-        """Get a list of all referenced entities for a particular entity
+        """Get a list of all referenced instances for a particular instance including itself
 
         :param inst: The entity instance to get all sub instances
         :type inst: ifcopenshell.entity_instance.entity_instance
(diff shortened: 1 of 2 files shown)


Moult:
Related to #755

Moult:
I'd like to add documentation to the add() function, but I'm not sure the best way to describe it. Something like "adds an entity including any dependent entities to an IFC file. If the entity already exists, it is not re-added" or something - would you have a better wording @aothms ?

aothms:
@Moult I'd say that's accurate. You can mention that it returns the added entity.

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -86,42 +101,75 @@ def __getitem__(self, key):
...
+        :type inst: ifcopenshell.entity_instance.entity_instance
+        :returns: A list of ifcopenshell.entity_instance.entity_instance objects
+        :rtype: list
+        """
         return [entity_instance(e) for e in self.wrapped_data.get_inverse(inst.wrapped_data)]
 
     def remove(self, inst):
+        """Deletes an IFC object in the file.
+
+        Other entities that reference the deleted object will be set to null.
```
aothms, line 167: I would say *Attribute values in other entity instances that ...*. Also in case of a list/set of references, the reference to the deleted will simply get removed from the aggregate.

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -86,42 +101,75 @@ def __getitem__(self, key):
...
+
+        :param type: The case insensitive type of IFC class to return.
+        :type type: string
+        :returns: A list of ifcopenshell.entity_instance.entity_instance objects
+        :rtype: list
         """
         return [entity_instance(e) for e in self.wrapped_data.by_type(type)]
 
     def traverse(self, inst, max_levels=None):
+        """Get a list of all referenced entities for a particular entity
```
aothms, line 141: Including the instance `inst`. Generally (and I am a big part in this, we should try to establish consistent wording for instance/class/entity). I would say the entity is the class definition in the schema (although I agree it could mean the instance just as well). Instances (or entity instances) are then the instantiations of these beasts.

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -86,42 +101,75 @@ def __getitem__(self, key):
...
+        :type id: int
+        :returns: Returned objects are unwrapped IFC objects.
+        :rtype: ifcopenshell.entity_instance.entity_instance
         """
         return self[id]
 
     def by_guid(self, guid):
-        """
-        Return IFC objects filtered by IFC GUID.
+        """Return IFC objects filtered by IFC GUID.
```
aothms, line 114: This returns *a single object/instance*, no?

Also what's up with `Returned objects are unwrapped IFC objects`? Is that a bug that needs to be fixed?

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -86,42 +101,75 @@ def __getitem__(self, key):
             return entity_instance(self.wrapped_data.by_guid(str(key)))
 
     def by_id(self, id):
-        """
-        Return IFC objects filtered by IFC ID.
+        """Return IFC objects filtered by IFC ID.
```
aothms, line 104: Also singular I'd say

aothms:
Thanks this is a wonderful effort. Some small comments inline

src/ifcopenshell-python/ifcopenshell/file.py
Moult: Fixed

src/ifcopenshell-python/ifcopenshell/file.py
Moult: Fixed. Also made terminology consistent :)

src/ifcopenshell-python/ifcopenshell/file.py
Moult: Fixed :)

No idea why I wrote that ;) Removed.

src/ifcopenshell-python/ifcopenshell/file.py
Moult: Fixed

Moult:
All comments actioned :)

aothms:
Thanks a million!

PR #826, opened 2020-04-08
Add missing IFC_PARSE_API

kukalaka:
Following up on our discussion on https://sourceforge.net/p/ifcopenshell/discussion/1782717/thread/6b28330744/ 

This is the commit to add the missing IFC_PARSE_API macro to the classes in the IfcSchema.h, which were not exported in the dll.


aothms:
Thanks for this. Much appreciated!

PR #827, opened 2020-04-08
Settings factory

s-leger:
Use a static "factory" method to keep init logic near settings definition.
```python
logger = logging.getLogger('ExportIFC')
            logging.basicConfig(
                filename=context.scene.BIMProperties.data_dir + 'process.log',
                filemode='a', level=logging.DEBUG)
            output_file = bpy.path.ensure_ext(self.filepath, '.ifc')
            ifc_export_settings = export_ifc.IfcExportSettings.factory(context, output_file, logger)
            ifc_parser = export_ifc.IfcParser(ifc_export_settings)
            ifc_export_settings.logger.info('Starting export')
```


s-leger:
Use a static factory method for ifcExportSettings

PR #828, opened 2020-04-08
Use ifcExportSettings.factory()

s-leger:


s-leger:
Use IfcExportSettings.factory() in exporter

PR #848, opened 2020-05-10
Fix visualization for pythonocc 7.4.0

didi-ux:
Following up the issue disscused here: https://sourceforge.net/p/ifcopenshell/discussion/1782717/thread/06e3103d93/

According to '2. transparent handles' in the link below all the calls 'GetObject' and 'GetHandle' can be removed.
https://github.com/tpaviot/pythonocc-core/wiki/Porting-from-0.18.x-to-7.4.0

The directions are now V3d_TypeOfOrientation enumerations (randomly chosen by me) instead of tuples of integers. Not sure if these changes cause problems for other use cases.


aothms:
Thanks for this. Much appreciated.

Issue #829, opened 2020-04-09
Proposal to create a new IfcOpenShell-python util module


Moult:
Ping @berndhahnebach and @yorikvanhavre and @aothms 

IfcOpenShell is used in a lot of utilities, which can lead to duplicate code if the same things are implemented again and again.

I propose to create a new `util` module, which can be imported as `import ifcopenshell.util`. It will contain methods that can be reused across the many implementations.

Here are four functions I propose to add first, as I suspect they can be instantly reused in FreeCAD/BlenderBIM Add-on/IFCCSV/BIMTester/any mini-script written by someone/etc.

 - `ifcopenshell.util.get_pset_property(element, pset_name, prop_name)` returns the wrappedValue if it is a simple property definition, else, for now, return is undefined.
 - `ifcopenshell.util.get_qto_property(element, qto_name, prop_name)` - same, but for Qto
 - `ifcopenshell.util.dd2dms(dd)` - converts from decimal degrees to degrees minutes seconds. Returns a tuple of (d, m, s). This is required for almost all georeferencing implementations
 - `ifcopenshell.util.dms2dd(dms)` - the other way around

We can add more and tweak as we go along, but I suspect these 4 have little debate around them, so would be a good starting point.

aothms:
Thanks for this initiative. Great idea.

Regarding properties and quantities: how about just enumerating all properties and quantities into a `defaultdict(dict)`. So that users do something like 

`util.get_all_properties(elem)["PSet_WallCommon"].get("IsExternal")`.

Advantages:

* Users can store the entire dict: more efficient when multiple props are required (in the other case one can short circuit on the pset name, but I doubt it's significant)
* Useful for serializing to JSON or User Interface

Is it useful to distinguish type/instance prop associations?

Basically there is code for this defined here: https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcopenshell-python/ifcopenshell/geom/app.py#L320

Indeed we have to think about the return type. Even for PSingleValue there can be a Unit. And if Unit is null we probably need to apply the IfcProject Unit (as that's the default for geometry too). So that e.g. the volume/surface measures correspond. Perhaps it can be some custom class to encapsulate all of this (it can inherit from `str` or `float` so that users can use it easily).

Moult:
[...]

PR #857, opened 2020-05-17
fix util/geolocation.py imports and dms2dd

carlopav:
ref https://github.com/IfcOpenShell/IfcOpenShell/issues/829#issuecomment-629766301


aothms:
Thanks. Not sure if we want to go with python 2 compatibility (no opinion), but in that case we probably should change `60` (, ...) to `60.0` so that the division is not turned into integer division. In python 3 there is `/` and `//` for float and integer division respectively.

carlopav:
@aothms why not, I can change that :)

PR #888, opened 2020-06-06
Fix #883 by adding correct number of edges

### LINKED ISSUE #833: Exclude entities problem js
### Overview
Hello, I am having a trouble using **IfcConvert 0.6** on my Ubuntu machine.
I am developing a web app with nodejs(12.16.1) and don't manage to use the following option : 
`--exclude=entities IfcOpeningElement`
The command works in a shell but not  when i call it in my script.

### Log
```
Error: Command failed: /path/to/IfcConvert/binary/IfcConvertbinary --use-element-guids /path/to/ifc/file/Test-Maison.ifc /path/to/dae/file/Test-Maison.dae --exclude=entities IfcOpeningElement
[Error] Invalid usage of '--exclude': the argument for option '--exclude' is invalid
```

### Example used:

[Test-Maison.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/4456068/Test-Maison.zip)

If details are needed about the script I can give them.

### LINKED ISSUE #883: [BlenderBIM] Pressing undo after creating a DIM crashes Blender
- Blender version v2.83.
- Platform OSX Mojave.
- BlenderBIM version 0.0.200525

robertguetzkow:
The mesh structure was invalid because the vertex indices for only one edge were set. The other edge had the same index for both vertices. See [T77537](https://developer.blender.org/T77537) for the ticket on Blender's bug tracker.

### DIFF
--- a/src/ifcblenderexport/blenderbim/bim/annotation.py
+++ b/src/ifcblenderexport/blenderbim/bim/annotation.py
@@ -62,7 +62,7 @@ class Annotator:
             obj.data.vertices.add(2)
             obj.data.vertices[-2].co = co1
             obj.data.vertices[-1].co = co2
-            obj.data.edges.add(2)
+            obj.data.edges.add(1)
             obj.data.edges[-1].vertices = (obj.data.vertices[-2].index, obj.data.vertices[-1].index)
         if isinstance(obj.data, bpy.types.Curve):
             polyline = obj.data.splines.new('POLY')


robertguetzkow:
Typo in the commit title, it's issue #883

Moult:
You are absolutely correct! Thank you very much for your help! I could not reproduce the issue so I assumed it was on Blender's end - thanks for the correction.

PR #905, opened 2020-06-22
Added USERSPACE_PYTHON_PREFIX cmake variable

yorikvanhavre:
Added USERSPACE_PYTHON_PREFIX cmake variable to allow install IfcPython in userspace instead of system-wide.

This allows you to install IfcOpenShell entirely in user-space, for ex. by setting CMAKE_INSTALL_PREFIX to /$USER/SomeDir and USERSPACE_PYTHON_PREFIX on, which installs IfcPython to $USER/.local/lib/python3/site-packages (as provided by python's built-in site module) instead of /usr/lib/python3/site-packages


aothms:
Thanks Yorik!

PR #916, opened 2020-07-14
IfcSiPrefix and IfcHierarchyHelper changes

sf8321:
IfcSiPrefix.cpp changes for MSVC 2013
IfcHierarchyHelper.h:176,185 changes in addRelatedObject about throw exception
exception comes from

ifcparse/IfcHierarchyHelper.h line 176 get_parent_of_relation(rel)
if (get_parent_of_relation(rel) == relating_object) {

ifcparse/IfcHierarchyHelper.h line 47
IfcUtil::IfcBaseClass* get_parent_of_relation(IfcUtil::IfcBaseClass* t)

ifcparse/IfcSchema.h
ptrdiff_t attribute_index(const std::string& attr_name) const {

ptrdiff_t index = -1;

-1 return but getArgument(..) Parameter is unsigned int
in the methode -1 is ca. 42123423423423 and throw IfcParse::IfcAttributeOutOfRangeException
we catch this exception and all works as expected and before in 0.5-rc1

### DIFF
--- a/src/ifcparse/IfcHierarchyHelper.h
+++ b/src/ifcparse/IfcHierarchyHelper.h
@@ -173,13 +173,16 @@ public:
 		bool found = false;
 		for (typename T::list::it i = li->begin(); i != li->end(); ++i) {
 			T* rel = *i;
-			if (get_parent_of_relation(rel) == relating_object) {
-				IfcEntityList::ptr products = get_children_of_relation(rel);
-				products->push(related_object);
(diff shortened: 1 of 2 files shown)


aothms:
Thanks! I'll add a `defined(_MSC_VER) && ` after merging to the preprocessor check, because otherwise in cases where `_MSC_VER` is not defined because undefined directives will be treated as 0.

PR #920, opened 2020-07-20
Add center-model-geometry

BlocInBloc:
On some IFC files (e.g. exported from Civil 3D), ObjectPlacement of IfcProduct is (0,0,0) and the geometry is far from the origin.
It's complicated to ask the software company to fix the export.
So we add new option "-center-model-geometry" in IfcConvert that uses a calculated bounds from geometry to center the model.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -256,6 +256,8 @@ int main(int argc, char** argv) {
 		("center-model",
             "Centers the elements by applying the center point of all placements as an offset."
             "Can take several minutes on large models.")
+		("center-model-geometry",
+            "Centers the elements by applying the center point of all mesh vertices as an offset.")
         ("model-offset", po::value<std::string>(&offset_str),
             "Applies an arbitrary offset of form 'x;y;z' to all placements.")
 		("model-rotation", po::value<std::string>(&rotation_str),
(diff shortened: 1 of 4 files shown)


Moult:
This sounds similar to IfcPatch's "ResetAbsoluteCoordinates" [here](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcpatch/recipes/ResetAbsoluteCoordinates.py). Would it make sense to make this function reusable?

`ifcpatch -i foo.ifc -r ResetAbsoluteCoordinates`

aothms:
Thanks you. I'm fine with merging this, in terms of code it's a reasonable solution and I'm sure it's not less useful then some of the other command line args.

It is a bit wasteful though two compute the geometry twice.  Ultimately what I would like is if the whole of IfcOpenShell would play a bit nicer along with STL. So we could do something like for the current approach:

~~~
context_iterator it;
serializer s;

std::for_each(it.begin(), it.end(), s);
~~~

For the functionality in this PR:

~~~
context_iterator it;
serializer s;

std::vector<geom_object> vec, vec_transformed;
std::copy(it.begin(), it.end(), std::back_inserter(vec));
std::accumulate(...); // somehow join bounds
std::transform(vec.begin(), vec.end(), std::back_inserter(vec_transformed), [](){...});
std::for_each(vec_transformed.begin(), vec_transformed.end(), s);
~~~

So like I said I'm fine with merging this, but maybe we should reflect a bit on the double processing time?

BlocInBloc:
Thanks for your feedback.
Indeed we are aware of the double processing time.

We made the decision to use an option "center-model-geometry" because we don't want to alter the ifc file of our customers (we don't want to patch it).

We're not sure how to compute bounds if ObjectLocalPlacement is (0,0,0) without calculating the geometry (which results in double processing time).
Is there another lead ?

We didn't fully understand and we don't know where to use your suggestion.
Can you detail ?

aothms:
> We're not sure how to compute bounds  ... without calculating the geometry (which results in double processing time).

No that's definitely true. That's the only way.

> We didn't fully understand and we don't know where to use your suggestion.

It's not a suggestion that can be used now, it's just a reflection on how if ifcopenshell geom iterator behaved better with STL, we could just buffer everything in a vector, offset coordinates and then do serialization without recomputing the geometry.

I'll merge this as is. Thanks for contributing this!

Issue #656, opened 2019-08-19
Using the IfcOpenShell Python interface


s-khini:
The first line of code doesn't seem to work, anyone an idea whats up?
$ wget -O duplex.zip http://projects.buildingsmartalliance.org/files/?artifact_id=4278
![Screen Shot 2019-08-19 at 4 35 32 PM](https://user-images.githubusercontent.com/29233225/63274040-631dd980-c29f-11e9-907b-3010a550edbd.png)

aothms:
> The first line of code

First line of what? Is this really about IfcOpenShell?

Try below, they updated their URLs

https://portal.nibs.org/files/wl/?id=4DsTgHFQAcOXzFetxbpRCECPbbfUqpgo

s-khini:
Dear aothms,
Thankl you very much, yes this is about IfcOpenShell in Conda
https://github.com/IfcOpenShell/IfcOpenShell#installing-ifcopenshell-with-conda

PR #922, opened 2020-07-20
fixed broken url

YoniChechik:
fixed broken url (thanks https://github.com/IfcOpenShell/IfcOpenShell/issues/656)


aothms:
Thanks!

PR #925, opened 2020-07-21
Use cerr_ instead of std::cerr in IfcConvert

HeySora:
It was crashing on `_UNICODE` builds (at least on VS2017) when `std::cerr` was written to. (e.g. when the program is launched without args).

Old behaviour *(wrong characters are being printed + an exception is thrown)* :
image

new behaviour *(the program executes just fine through the end, errors are printed correctly)* :
image


aothms:
Thanks! Indeed quite an oversight, I think the official builds are with `_UNICODE`, so this is very helpful.

PR #931, opened 2020-07-26
[BlenderBIM] Support for pythonocc-core 7.4.0

CyrilWaechter:
Proposal to support both pythonocc-core 0.18.1 and pythonocc-core 7.4.0


Moult:
This is more than a proposal, this is awesome work! Thanks so much!

Issue #937, opened 2020-07-29
shared library files have incomplete soname (patch)


brunopostle:
On a unix/linux system shared libraries are expected to be installed using a versioned [soname](https://en.wikipedia.org/wiki/Soname), with the unversioned libXXX.so name existing only as a symlink. The attached patch to CMakeLists.txt fixes this for the v0.6.0 branch. The result looks something like this: 

`lrwxrwxrwx. 1 root root       15 Jul 29 23:53 /usr/lib64/libIfcGeom.so -> libIfcGeom.so.0`
`lrwxrwxrwx. 1 root root       19 Jul 29 23:53 /usr/lib64/libIfcGeom.so.0 -> libIfcGeom.so.0.0.0`
`-rwxr-xr-x. 1 root root  4581336 Jul 29 23:53 /usr/lib64/libIfcGeom.so.0.0.0`
`lrwxrwxrwx. 1 root root       16 Jul 29 23:53 /usr/lib64/libIfcParse.so -> libIfcParse.so.0`
`lrwxrwxrwx. 1 root root       20 Jul 29 23:53 /usr/lib64/libIfcParse.so.0 -> libIfcParse.so.0.0.0`
`-rwxr-xr-x. 1 root root 32844392 Jul 29 23:53 /usr/lib64/libIfcParse.so.0.0.0`
`lrwxrwxrwx. 1 root root       19 Jul 30 00:14 /usr/lib64/libSerializers.so -> libSerializers.so.0`
`lrwxrwxrwx. 1 root root       23 Jul 30 00:14 /usr/lib64/libSerializers.so.0 -> libSerializers.so.0.0.0`
`-rwxr-xr-x. 1 root root  1001416 Jul 30 00:14 /usr/lib64/libSerializers.so.0.0.0`

[ifcopenshell-soname.patch](https://github.com/IfcOpenShell/IfcOpenShell/files/4997935/ifcopenshell-soname.patch.txt)

aothms:
Thanks, I think that makes sense. Any reason you didn't make this a PR? Should the version number be 0.6 or something?

brunopostle:
I'd forgotten how to do a pull request, I'll do one.

The soname version numbers tend not to follow release numbers: the first number increments whenever there is a binary API change, so it may change between IfcOpenShell 0.6 and 0.7 for example. The second and third numbers are for API compatible major/minor releases. So I started at 0.0.0.

[...]

PR #941, opened 2020-07-31
cmake to use linux/unix soname conventions for shared libraries

brunopostle:
initial soversion for v0.6.0 branch is 0


aothms:
Thanks. @kkremitzki @CyrilWaechter would this disrupt your packaging work? Any thoughts?

kkremitzki:
It would not cause disruption on my end. Depending on how closely you follow semantic versioning practices, you may want to use `0.6` for the soversion instead of `0`. Regardless of choice for that you can also use the actual `0.6.0` for the version, too. But using all zeroes like this is OK too if the version choice explicitly is making no promises.

CyrilWaechter:
After reading wikipedia and #937 I would also vote for 0.6 for soname and 0.6.0 for version as 0.7 is also around (not yet packaged but maybe later). There was some API change between 0.5 and 0.6. I assume it is the same between 0.6 and 0.7.

aothms:
Thanks all for your comments. This is really not my expertise, but naively I would say that a version that matches the branch will be more recognizable. There are quite a bit of novice users of IfcOpenShell that might get confused from a low incrementing number.

Should the version number also be applied to the SWIG generated python wrapper module https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcwrap/CMakeLists.txt#L56 ?

brunopostle:
> Should the version number also be applied to the SWIG generated python wrapper module

This doesn't seem to be a problem, since `_ifcopenshell_wrapper.so` is a private library and not intended to be linked-to by anything else. None of the .so files within python wrappers on this fedora system have a versioned soname, but all the 'normal' shared libraries do.

PR #960, opened 2020-08-26
Fix collections deprecation warning

CyrilWaechter:
# What it solves
```python
ifcopenshell/geom/occ_utils.py:28
  /home/cyril/git/BIMxBEM/ifcopenshell/geom/occ_utils.py:28: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import namedtuple, Iterable
```
(see [python3 doc](https://docs.python.org/3/library/collections.html?highlight=collections#module-collections))
Note : python 3.9 final should be released on [05.10.2020](https://www.python.org/dev/peps/pep-0596/)
# Potential issues
Concerning python3 I think it is not an issue as python 3.3 is not supported anymore.
Concerning python2, there is still many `from __future__ import *`. This new code will probably fail using python2. Is it an issue as support for python2 has been dropped on 01.01.2020 ?

### DIFF
--- a/src/ifcblenderexport/occ_utils.py
+++ b/src/ifcblenderexport/occ_utils.py
@@ -25,7 +25,11 @@ import random
 import operator
 import warnings
 
-from collections import namedtuple, Iterable
+from collections import namedtuple
+try:  # python 3.3+
+    from collections.abc import Iterable
+except ModuleNotFoundError: # python 2
+    from collections import Iterable
 
 try:
     from OCC.Core import TopoDS, gp, Quantity, BRepTools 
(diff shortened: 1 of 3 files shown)


CyrilWaechter:
Just a little update : python 3.9 has been released. It is not currently even in Arch testing so I assume we still have some time. Is there something wrong with the PR ?

aothms:
Sorry I had mistakenly assumed this was solely about BlenderBIM, so waited for @Moult to handle it.

IfcOpenShell has these long release cycles, but I prefer that v0.6.0 remains compatible with 2.7. So if it's not too much trouble I prefer this to be try catched with a fallback to the old import.

CyrilWaechter:
I modified the commit and rebased it on v0.6.0 to allow fast forward.

aothms:
Thanks so much @CyrilWaechter and again sorry for the delay.

Issue #981, opened 2020-09-06
BlenderBIM: unit conversion slightly off


theoryshaw:
Might be missing something, but it looks like the unit conversions are slightly off

file: https://github.com/OpeningDesign/OD_Library/tree/052d57a9a0c7647fb4d033344511ef49f2ed9803/BIM/BlenderBIM

![image](https://user-images.githubusercontent.com/507113/92334817-07172600-f057-11ea-8039-2c9573563b1f.png)

PR #982, opened 2020-09-07
Fixes the areas of #981, but not the volumes

theoryshaw:


theoryshaw:
I tried to increase the accuracy of 'cubic foot', but it didn't seem to change anything. 

image

theoryshaw:
For the volume it seem like it's multiplying the width and length (in feet) by the height (in meters).

Moult:
Fixed in bugreport.

PR #1033, opened 2020-10-09
[BlenderBIM] Import materials psets and display it

CyrilWaechter:
Import psets similarly to objects.
Display in the UI similarly to objects.
Basic handling for ConstituentSet and MaterialList.

### DIFF
--- a/src/ifcblenderexport/blenderbim/bim/ui.py
+++ b/src/ifcblenderexport/blenderbim/bim/ui.py
@@ -700,6 +700,15 @@ class BIM_PT_material(Panel):
             row = layout.row(align=True)
             row.prop(pset, 'name', text='')
             row.operator('bim.remove_material_pset', icon='X', text='').pset_index = index
+            op = row.operator('bim.copy_property_to_selection', icon='COPYDOWN', text='')
+            for prop in pset.properties:
+                row = layout.row(align=True)
+                row.prop(prop, 'name', text='')
+                row.prop(prop, 'string_value', text='')
+                op = row.operator('bim.copy_property_to_selection', icon='COPYDOWN', text='')
+                op.pset_name = pset.name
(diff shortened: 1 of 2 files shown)


CyrilWaechter:
Is it necessary to have 2 different class for psets ? `BIMMaterialProperties` seems a bit redundant with `BIMObjectProperties`.
Also according to Blender API docs on [PropertyGroup](https://docs.blender.org/api/current/bpy.props.html) and all other examples in blender property group are always assigned to a name with no capital case to differentiate class from group instance.
`bpy.types.Material.my_settings = bpy.props.PointerProperty(type=MaterialSettings)`

I would imagine assigning `BIMObjectProperties` to both `Material` and `Object`.

src/ifcblenderexport/blenderbim/bim/ui.py
```diff
@@ -700,6 +700,15 @@ def draw(self, context):
             row = layout.row(align=True)
             row.prop(pset, 'name', text='')
             row.operator('bim.remove_material_pset', icon='X', text='').pset_index = index
+            op = row.operator('bim.copy_property_to_selection', icon='COPYDOWN', text='')
```
Moult, line 703: This operator doesn't currently work for materials

Moult:
Thanks so much for the PR! Materials is on the list to receive a lot of upcoming fixes!

`BIMMaterialProperties` and `BIMObjectProperties` are two different classes as they represent attributes that can be stored to `IfcMaterial` and `IfcElement`, which do not share the same parent class. If, in the future, `IfcMaterial` becomes rooted in IFC, then I would be happy to merge. For now, would prefer to keep separate.

I wasn't aware that there was a naming scheme required for property group. I think you're right. Happy to change (a little search and replace task)!

PR #1070, opened 2020-11-01
'foot' not 'feet'.  Also case seems to matter

theoryshaw:


Moult:
Rejecting PR because according to docs the lowercase is more correct (see [here](https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/link/ifcconversionbasedunit.htm)).

But you are absolutely right that `foot` is the right one, not `feet`. Can you submit another PR?

PR #1079, opened 2020-11-05
gcc and clang warning fixes

tpaviot:
Fix a bunch of compiler warnings occurring on Ubuntu Linux 18.04 and higher (both gcc and clang). commits are atomic and can be cherry-picked if ever the PR is not merged. Tested with opencascade 7.4.0 and 7.4.0p1

### DIFF
--- a/src/ifcgeom/IfcGeom.h
+++ b/src/ifcgeom/IfcGeom.h
@@ -224,6 +224,11 @@ private:
 	double modelling_precision;
 	double dimensionality;
 	double layerset_first;
+
+	// For stopping PlacementRelTo recursion in convert(const IfcSchema::IfcObjectPlacement* l, gp_Trsf& trsf)
+	const IfcParse::declaration* placement_rel_to;
+
+	faceset_helper* faceset_helper_;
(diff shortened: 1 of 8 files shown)


aothms:
Thanks, this is great. I will just merge it as is. Some of these functions might be used later, but we have version control for that. Thanks!

tpaviot:
I could also have commented out unused functions, I didn't know if they ware outdated or here just in case. Note that I could not fix another "initialization warning":
```c++
[...]IfcOpenShell/src/serializers/SvgSerializer.cpp:415:96: warning: missing initializer for member ‘horizontal_plan::next_elevation’ [-Wmissing-field-initializers]
 storage.push_back(horizontal_plan{ data.storey, data.storey_elevation,  +1. });
```
Indeed ```horizontal_plan``` expects 4 arguments whereas only 3 are passed to the initializer:
```c++
struct horizontal_plan {
	IfcUtil::IfcBaseEntity* storey;
	double elevation, offset, next_elevation;
};
```
In the SvgSerializer algorithm, I don't know how to set the ```next_elevation``` value in the initializer.

PR #1080, opened 2020-11-05
add flake8 to travis-ci

htlcnn:
@aothms please have a look


tpaviot:
I'm not sure that including a pyflake check in the travis build is such a good idea. It drastically increases verbosity and it is a choice that can be balanced in regard with other solutions (there are plenty of linters and other code analysis tools). IMHO, as ifcopenshell includes both c++ and python code, it would be better to register an account somewhere (codacy for instance) or setup gh actions to trigger a code review each time a PR is opened.

Moult:
@tpaviot good point! I checked with @htlcnn on IRC (##architect on freenode) and it's now reverted. https://github.com/IfcOpenShell/IfcOpenShell/commit/adf9ac52b2422b89d2b68677a2cdde5b780b7903

PR #1106, opened 2020-11-13
Suggested implementation for IfcPresentationLayerAssignment support in BlenderBIM

Krande:
Here is the rough draft for supporting IfcPresentationLayerAssignment in BlenderBIM. 

One thing I noted when preparing to make a pull request is that I changed the location of the original `presentation_layer` PointerProperty from the `BIMMeshProperties` to the `BIMObjectProperties`. So as of now there is a presentation_layer attribute on both. I can't recall why I did that (maybe it was due to the way I use the object name\guid to recognize each object on import from IFC or something), but it is certainly something I am open to change. I should at the very least remove the attribute from one or the other. 

Oh, and I added a .gitignore file to the lib\site\packages folder primarily to have my dependencies located in that folder. It is not strictly necessary if it messes up something on your end, so feel free to remove it.

Anyways, let me know what you need to make this good enough for a merge into master (or should I say v0.6.0) :)

Best Regards
Kristoffer


Moult:
Cheers! I've merged it. I'll leave some comments that you can ignore in the PR, and I'll commit the changes so you can review. Let me know if you think any of my comments don't make sense and we can revert :)

src/ifcblenderexport/blenderbim/bim/ui.py
```diff
@@ -917,6 +959,56 @@ def draw(self, context):
...
+    bl_space_type = "PROPERTIES"
+    bl_region_type = "WINDOW"
+    bl_context = "scene"
+
+    def draw(self, context):
+        layout = self.layout
+        layout.use_property_split = True
+        props = context.scene.BIMProperties
+
+        layout.row().prop(props, "presentation_layers")
```
Moult, line 975: I removed this line simply because this convention isn't shown anywhere else in the line

src/ifcblenderexport/blenderbim/bim/ui.py
```diff
@@ -917,6 +959,56 @@ def draw(self, context):
...
+        layout.row().operator("bim.add_presentation_layer")
+
+        if props.presentation_layers:
+            layout.template_list(
+                "BIM_UL_presentation_layers", "", props, "presentation_layers", props, "active_presentation_layer_index"
```
Moult, line 980: This can use BIM_UL_generic as there is nothing special about this list.

src/ifcblenderexport/blenderbim/bim/__init__.py
```diff
@@ -312,6 +320,7 @@
         ui.BIM_UL_generic,
         ui.BIM_UL_clash_sets,
         ui.BIM_UL_constraints,
+        ui.BIM_UL_presentation_layers,
```
Moult, line 323: Deleted, as BIM_UL_generic can be used instead.

src/ifcblenderexport/blenderbim/bim/ui.py
```diff
@@ -1987,6 +2079,15 @@ def draw_item(self, context, layout, data, item, icon, active_data, active_propn
             layout.label(text="", translate=False)
 
 
+class BIM_UL_presentation_layers(bpy.types.UIList):
```
Moult, line 2082: Deleted entire class, as BIM_UL_generic can be used instead

src/ifcblenderexport/blenderbim/libs/site/packages/.gitignore
```diff
@@ -0,0 +1,5 @@
+*
+!.gitignore
+!ifcopenshell.pth
+!OCC.pth
+!svgwrite.pth
```
Moult, line 5: Deleting this file

src/ifcblenderexport/blenderbim/bim/ui.py
```diff
@@ -917,6 +959,56 @@ def draw(self, context):
...
+        if props.presentation_layers:
+            layout.template_list(
+                "BIM_UL_presentation_layers", "", props, "presentation_layers", props, "active_presentation_layer_index"
+            )
+            pres_layer = props.presentation_layers[props.active_presentation_layer_index]
+
+            row = layout.row(align=True)
+            row.prop(pres_layer, "name")
+            if pres_layer.name in bpy.context.scene.BIMProperties.presentation_layers:
+                pres_layer = bpy.context.scene.BIMProperties.presentation_layers[pres_layer.name]
```
Moult, line 987: This will not work if two layers have the same name. You can trigger this bug by adding two layers and not changing their name, and then editing the description of one but not the other. A better approach is to use props.active_presentation_layer_index

src/ifcblenderexport/blenderbim/bim/ui.py
```diff
@@ -917,6 +959,56 @@ def draw(self, context):
...
+                row = layout.row()
+                row.prop(pres_layer, "identifier")
+                row = layout.row()
+                row.prop(pres_layer, "layer_on")
+                row = layout.row()
+                row.prop(pres_layer, "layer_frozen")
+                row = layout.row()
+                row.prop(pres_layer, "layer_blocked")
+                row = layout.row()
+                row.prop(pres_layer, "layer_styles")
```
Moult, line 1002: Removed layer_styles, since it is not yet implemented, and needs further consideration prior to adding it to the UI. These styles could be complex. It _could_ be linked to a greasepencil style? Maybe? Not 100% sure yet.

src/ifcblenderexport/blenderbim/bim/ui.py
```diff
@@ -804,6 +801,51 @@ def draw(self, context):
...
+
+        if elem_props.name != "":
+            layout.label(text=f'Object is part of Presentation layer "{elem_props.name}"')
+            layout.row().operator("bim.remove_from_presentation_layer")
+        else:
+            layout.label(text="Not included in any presentation layer")
+
+        if scene_props.presentation_layers:
+            layout.template_list(
+                "BIM_UL_presentation_layers",
```
Moult, line 834: Switched to BIM_UL_generic. In the future, when we create a more special UL which has the typical UI like an "eye" icon for visibility and snow icon for freeze we can switch this back.

src/ifcblenderexport/blenderbim/bim/ui.py
```diff
@@ -804,6 +801,51 @@ def draw(self, context):
...
+                "",
+                scene_props,
+                "presentation_layers",
+                scene_props,
+                "active_presentation_layer_index",
+            )
+            pres_layer = scene_props.presentation_layers[scene_props.active_presentation_layer_index]
+            if pres_layer.name in bpy.context.scene.BIMProperties.presentation_layers:
```
Moult, line 842: Probably better to use the layer index, as name may be non-unique.

src/ifcblenderexport/blenderbim/bim/prop.py
```diff
@@ -797,6 +797,16 @@ class ClashSet(PropertyGroup):
...
 
 
+class PresentationLayer(PropertyGroup):
+    name: StringProperty(name="Name")
+    description: StringProperty(name="Description")
+    identifier: StringProperty(name="Identifier")
+    layer_on: BoolProperty(name="LayerOn", default=True)
+    layer_frozen: BoolProperty(name="LayerFrozen", default=False)
+    layer_blocked: BoolProperty(name="LayerBlocked", default=False)
+    layer_styles: EnumProperty(items=[], name="LayerStyles")
```
Moult, line 807: Removed for now, as implementation is uncertain.

src/ifcblenderexport/blenderbim/bim/prop.py
```diff
@@ -1716,7 +1729,7 @@ class BIMMeshProperties(PropertyGroup):
     is_swept_solid: BoolProperty(name="Is Swept Solid")
     swept_solids: CollectionProperty(name="Swept Solids", type=SweptSolid)
     is_parametric: BoolProperty(name="Is Parametric", default=False)
-    presentation_layer: StringProperty(name="Presentation Layer")
+    presentation_layer: PointerProperty(name="Presentation Layer", type=PresentationLayer)
```
Moult, line 1732: Removed for now, as you don't use this, instead you use the one assigned to BIMObjectProperties.

However, I _think_ it is more correct for it to belong here, because layers in IFC are actually assigned to representation items (i.e. Meshes in the Blender world), _not_ to the object itself. If you store it on the object, there is currently no way to distinguish the theoretical situation where an object has two representations, both assigned to two different layers.

This can fix this in the future.

src/ifcblenderexport/blenderbim/bim/operator.py
```diff
@@ -4316,6 +4316,80 @@ def execute(self, context):
         return {"FINISHED"}
 
 
+class AddToPresentationLayer(bpy.types.Operator):
```
Moult, line 4319: Renamed to "Assign" instead of "AddTo" to be consistent with the rest of the code.

src/ifcblenderexport/blenderbim/bim/operator.py
```diff
@@ -4316,6 +4316,80 @@ def execute(self, context):
...
+        for obj in bpy.context.selected_objects:
+            obj.BIMObjectProperties.presentation_layer.name = presentation_layer.name
+            elem_name = obj.BIMObjectProperties.attributes["Name"].string_value
+            obj.hide_set(not presentation_layer.layer_on)
+            print(f'Adding "{elem_name}" to View/Presentation Layer "({self.index} - {presentation_layer.name})"')
+
+        return {"FINISHED"}
+
+
+class RemoveFromPresentationLayer(bpy.types.Operator):
```
Moult, line 4338: Renamed to "Unassign" instead of "RemoveFrom" to be consistent with the rest of the code.

src/ifcblenderexport/blenderbim/bim/operator.py
```diff
@@ -4316,6 +4316,80 @@ def execute(self, context):
...
+    index: bpy.props.IntProperty()
+
+    def execute(self, context):
+        presentation_layer = bpy.context.scene.BIMProperties.presentation_layers[self.index]
+
+        # vl = bpy.context.scene.view_layers[presentation_layer.name]
+
+        for obj in bpy.context.selected_objects:
+            obj.BIMObjectProperties.presentation_layer.name = presentation_layer.name
+            elem_name = obj.BIMObjectProperties.attributes["Name"].string_value
```
Moult, line 4331: This will break if the layer name changes after assigning. We'll leave it for now but will come up with a better solution later.

src/ifcblenderexport/blenderbim/bim/export_ifc.py
```diff
@@ -2023,10 +2022,26 @@ def create_styled_item(self, item, representation_item=None):
...
+            pl = scene_props.presentation_layers[name]
+            print("Presentation Export: ", name, pl.identifier, pl.layer_on)
+
+            with_style = False
+            if pl.layer_on is False:
+                with_style = True
+
+            if with_style is False:
+                self.file.createIfcPresentationLayerAssignment(
+                    name, pl.description, [i["ifc"].MappedRepresentation for i in assigned_items], pl.identifier
```
Moult, line 2038: We should export description and identifier as None if it is blank

src/ifcblenderexport/blenderbim/bim/export_ifc.py
```diff
@@ -2023,10 +2022,26 @@ def create_styled_item(self, item, representation_item=None):
...
+                )
+            else:
+                self.file.createIfcPresentationLayerWithStyle(
+                    name, pl.description, [i["ifc"].MappedRepresentation for i in assigned_items], pl.identifier,
```
Moult, line 2042: We should export description and identifier as None if it is blank

src/ifcblenderexport/blenderbim/bim/import_ifc.py
```diff
@@ -1309,6 +1310,25 @@ def add_project_to_scene(self):
             self.type_collection.name
         ].hide_viewport = True
 
+    def get_presentation_layers(self):
+        for f in self.file.by_type("IfcPresentationLayerAssignment"):
+            pl = bpy.context.scene.BIMProperties.presentation_layers.add()
+            pl.name = f.Name
+            pl.description = f.Description
+            pl.identifier = f.Identifier
```
Moult, line 1318: Description and identifier can be null, so we need to check that.

src/ifcblenderexport/blenderbim/bim/import_ifc.py
```diff
@@ -1309,6 +1310,25 @@ def add_project_to_scene(self):
...
+            if f.is_a() == "IfcPresentationLayerWithStyle":
+                pl.layer_on = f.LayerOn if f.LayerOn is not None else True
+                pl.layer_frozen = f.LayerFrozen if f.LayerFrozen is not None else False
+                pl.layer_blocked = f.LayerBlocked if f.LayerBlocked is not None else False
+
+            for item in f.AssignedItems:
+                guid = item.OfProductRepresentation[0].ShapeOfProduct[0].GlobalId
```
Moult, line 1325: `OfProductRepresentation` only works for `IfcRepresentation`, not `IfcRepresentationItem`. This means the import will not work for IFCs exported from the BlenderBIM Add-on, as the export links to the representation items. I just realised this part is actually harder than it seems, see bug #1109. I'll leave the implementation as is for now, but make a note of it that it probably needs to be rewritten.

Moult:
Cheers! Again, amazing work! You can see my changes to it here: https://github.com/IfcOpenShell/IfcOpenShell/commit/90807446695f0e2bf6443f2a1b298049aab91bd2

I realised that supporting the import of presentation layers could be much trickier than I originally envisioned, see #1109 - so there might be quite a bit more work to do. I'll have a think about it :) However, the support is certainly much improved from where it was previously! Great job!

PR #1115, opened 2020-11-19
IfcWall duplicate in default styles deleted

LaurensJN:
@Moult Can you check if I did a proper pull request?


Moult:
Looks great! Thank you so much!

PR #1117, opened 2020-11-19
Added default constructor to IfcHierarchyHelper

mo0nless:
Easily get the correct IFC file schema definition by name.

### DIFF
--- a/src/ifcparse/IfcHierarchyHelper.h
+++ b/src/ifcparse/IfcHierarchyHelper.h
@@ -100,6 +100,7 @@ namespace {
 template <typename Schema>
 class IFC_PARSE_API IfcHierarchyHelper : public IfcParse::IfcFile {
 public:
+	IfcHierarchyHelper(const IfcParse::schema_definition* schema) :IfcFile(schema) {}
 	template <class T> 
 	T* addTriplet(double x, double y, double z) {
 		std::vector<double> a; a.push_back(x); a.push_back(y); a.push_back(z);


aothms:
Thanks. I guess hierarchy helper is supposed to be used in a templated fashion. Couple of questions: How do you envision the constructor argument to be used with the template argument? Any reason you're not working off of v0.6.0, 0.7 is a bit experimental and quite a bit of commits of 0.6 need still to be applied to 0.7 I guess. 

See the default constructor in 0.6 for example: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcparse/IfcHierarchyHelper.h#L171

aothms:
@mo0nless did you get my previous comment.

See the v0.6.0 version of the prototype you're changing:

```c++
template <typename Schema>
class IFC_PARSE_API IfcHierarchyHelper : public IfcParse::IfcFile {
public:
	IfcHierarchyHelper() : IfcParse::IfcFile(&Schema::get_schema()) {}
```

mo0nless:
Hi @aothms!! The change was made due to an issue that I had with the templated function. I couldn't use a different IFC schema rather than the [one specified](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcparse/IfcFile.h#L115) initially set to IFC2x3. 
Currently I'm using the HierarchyHelper in this way: 

`IfcHierarchyHelper<Ifc4>* _file = new IfcHierarchyHelper<Ifc4>(IfcParse::schema_by_name("IFC4"));`

I've been using the v0.7.0 because with the v0.6.0 we were stuck and we had some problem while building with MSVC, mostly with some of the dependencies like ICU. Also we faced some problems trying to have dynamic linking instead of static.

aothms:
Ok, thanks for clarifying. I will just merge this as is, but I would really recommend to stick to v0.6.0 as it has much more active development at the moment. v0.6.0 also doesn't rely on ICU anymore btw. I can imagine there might be issues though with dynamic linking (because none of the devs rely on this). Once in a while we get a PR that fixes this. If v0.7 is fine then no worries, but otherwise you might want to have a look again at the current state of v0.6.

Issue #1119, opened 2020-11-20
Regression: Failed to convert face of covering (example file)


johanrd:
1. Convert [1119--covering--failedToConvertFace--augmented.ifc](https://github.com/IfcOpenShell/files/blob/master/1119--covering--failedToConvertFace--augmented.ifc) with 808ef1a19ecea3653d7a6c101bafe232e9897f39 (OCCT 7.5.0 or 7.3.9p4) and and experience error:
```
#36149=IfcCovering('0zbeqN_I965rBWQ8kgPRAl',$,'','',$,#36150,#36183,$,.INSULATION.)
[Error] {0zbeqN_I965rBWQ8kgPRAl} Failed to convert:
#35975=IfcPolygonalFaceSet(#35904,.F.,(#95,#17936,#702,#35656,#35905,#709,#35906,#17570,#35907,#2296,#35908,#17943,#27901,#18378,#35909,#27905,#35910,#35911,#14002,#35912,#35913,#35914,#1368,#3365,#17951,#4168,#35915,#142,#35916,#943,#17953,#35917,#35918,#...
```
2. Convert the same file with b08346cc6451b46b6bd88b93da1dbf2f3c169eca and experience successful conversion:
<img width="284" alt="Screenshot 2020-11-20 at 14 41 08" src="https://user-images.githubusercontent.com/4601554/99806587-838db100-2b3e-11eb-930a-ec2f631a7332.png">

Suspected commit: 09847513ab1346aba287fcbb0b301ea547089450

aothms:
I cannot reproduce this on Windows and Linux

```
~/IfcOpenShell/build/Linux/x86_64/install/ifcopenshell/bin/IfcConvert -yv 1119--covering--failedToConvertFace--augmented.ifc 1119.dae
IfcOpenShell IfcConvert 0.6.0b0 (OCC 7.5.0)
Scanning file...
Done scanning file
Parsing input file took 0 seconds
Creating geometry...
Done creating geometry (1 objects)

Log:
[Notice] exclude entities IfcOpeningElement IfcSpace
[Warning] {0zbeqN_I965rBWQ8kgPRAl} No material and surface styles for:
#36149=IfcCovering('0zbeqN_I965rBWQ8kgPRAl',$,'','',$,#36150,#36183,$,.INSULATION.)


Conversion took 0 second
```

johanrd:
Yes, this can be closed after 37b7d76, thanks!

PR #1120, opened 2020-11-20
Fix for #1119 

johanrd:
Fixes regression introduced by 0984751


aothms:
Thanks, good catch @johanrd

PR #1122, opened 2020-11-23
bimtester: some restructure work

berndhahnebach:
to be disscussed. see https://community.osarch.org/discussion/322/bimtester-improvements-as-well-as-standalone-gui-and-freecad-module

In addition to the changes made this has to be disscuessed to ... https://github.com/berndhahnebach/IfcOpenShell/blob/18cb8df3e6a6a2362532093e9206e12fbc3289d6/src/ifcbimtester/bimtester.py#L25  Otherwise  current standalone bimtester script might have trouble. I did not test BlenderBIMTester with this PR code.


berndhahnebach:
How about the licence header. Does ifcopenshell uses them or not. If yes which one? Is there some template?

Moult:
I like to merge first, ask questions later :D 

It looks like a step in the right direction, though I haven't tested your code.

PR #1134, opened 2020-11-26
bimtester: steps, improve imports

berndhahnebach:


Moult:
Btw we've recently established a convention to run "black" on the code with a line length of 120

PR #1137, opened 2020-12-01
A start for smart clash grouping

vinnividivicci:
Hello,

Before I continue development on this feature I would like to merge what I have so far into the main branch, since it's getting tedious to keep my blender version up to date for testing.

You will notice that the dependencies for the smart grouping in ifclash are imported at the function level. This way, if the user does not have them installed it will not break the whole ifclash module.

It's obviously not perfect yet but my understanding is that BlenderBIM is currently in an unstable release cycle, so I will improve it and merge the improvements in here later.


Moult:
Awesome! Do you mind cleaning this up on top of the latest v0.6.0 HEAD so that it can be merged automatically instead of resolving conflicts?

In addition, can you make sure the commits only touch blenderbim files. e.g don't touch the  `test/input` submodule.

Happy to merge after that :)

vinnividivicci:
Sounds good, it may take me a while to do that so I'll close this pull request in the meantime.

PR #1166, opened 2020-12-13
Changes to the build scripts to support Visual Studio 2019

ahladik:
Changes to the build scripts to support Visual Studio 2019, to optionally use a toolset different from the default and to build correctly when more than a single Visual Studio version is installed.

I have updated the Boost version from 1.67 to 1.71, since this is the first version to support VS 2019.

I also changed EOL to CR/LF instead of LF, since using Unix EOL style in a Windows batch file occasionally causes incomprehensible errors (which I have experienced in this case).

However, errors remain.

It is impossible to successfully complete the build of the Debug version of the IfcOpenShell solution, as the Visual Studio projects are incorrectly set to link to ...\OpenCOLLADA\lib\opencollada\xml.lib instead of xmld.lib, which is the version produced by the scripts when the 'Debug' parameter is used. I guess this is not noticeable in case the release version has been built as well or only the release version.

Building with Visual Studio 2012, for some reason the installation position of the OCCT libraries is wrong. The correct location would be:

...\deps-...-installed\opencascade-7.3.0p3\win32\lib

while with the above compiler it becomes:

...\deps-...-installed\opencascade-7.3.0p3\win32\vc11\libd

and consequently the call to build-CMake fails.

Finally, build-CMake doesn't use the 'BuildDepsCache' file correctly (if that's what it's supposed to do) and errs in determining the location of the OCCT library when running a second build by changing compiler or architecture.

Addendum:

I have discovered that branch 0.6.0 cannot succesfully compile with Visual C++ 2012 (and so I would say with 2010 and 2008 too) due to C++ constructs that are not supported by that compiler. I would then suggest to remove support for those 3 Visual Studio versions from the scripts.
The problematic files are:
IfcParse.h
IfcParse.cpp
IfcSchema.cpp
IfcGeomIteratorSettings.h
parse_ifcxml.cpp


aothms:
Thanks! It appears this PR changes line endings of the files so it makes it hard to inspect the changes being applied as the entire file appears changed. Would be great if you can correct this. Otherwise I will add a commit to reverse the line endings before applying.

ahladik:
If you have read my comments, there is a definite reason for changing the line endings, please don't change them.

Also there is no purpose to keep using Boost 1.67 instead of 1.71 since Boost 1.71 compiles smoothly with VS 2012, but IfcOpenShell 0.6.0 instead needs at least VS 2013.

So what I would do instead is to provide Visual Studio support only from 2013 onwards.

aothms:
Ok fair enough posting the changes ignoring whitespace then for reference.

> git diff -w v0.6.0..origin/pr/1166 -- win

```diff
diff --git a/win/build-deps.cmd b/win/build-deps.cmd
index b0a3a2fc..ffe124d2 100644
--- a/win/build-deps.cmd
+++ b/win/build-deps.cmd
@@ -55,6 +55,14 @@ if not %ERRORLEVEL%==0 (
 :: Set up variables depending on the used Visual Studio version
 call vs-cfg.cmd %1
 IF NOT %ERRORLEVEL%==0 GOTO :Error
+
+:: Set up the BuildDepsCache.txt filename
+IF DEFINED VS_TOOLSET (
+    set BUILDDEPTHCACHE=BuildDepsCache-%VS_PLATFORM%-%VS_TOOLSET%.txt
+) ELSE (
+    set BUILDDEPTHCACHE=BuildDepsCache-%VS_PLATFORM%.txt
+)
+
 call build-type-cfg.cmd %2
 IF NOT %ERRORLEVEL%==0 GOTO :Error
 
@@ -111,6 +119,10 @@ call cecho.cmd 0 13 "* CMake Generator`t= '`"%GENERATOR%`'`t
 echo   - Passed to CMake -G option.
 call cecho.cmd 0 13 "* Target Architecture`t= %TARGET_ARCH%"
 echo   - Whether were doing 32-bit (x86) or 64-bit (x64) build.
+call cecho.cmd 0 13 "* Target Platform`t= %VS_PLATFORM%"
+echo   - Passed to CMake -A option.
+call cecho.cmd 0 13 "* Target Toolset`t= %VS_TOOLSET%"
+echo   - Passed to CMake -T option.
 call cecho.cmd 0 13 "* Dependency Directory`t= %DEPS_DIR%"
 echo   - The directory where %PROJECT_NAME% dependencies are fetched and built.
 call cecho.cmd 0 13 "* Installation Directory = %INSTALL_DIR%"
@@ -146,7 +158,7 @@ set /p do_continue="> "
 if "%do_continue%"=="n" goto :Finish
 
 :: Cache last used CMake generator for other scripts to use
-if defined GEN_SHORTHAND echo GEN_SHORTHAND=%GEN_SHORTHAND%>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
+if defined GEN_SHORTHAND echo GEN_SHORTHAND=%GEN_SHORTHAND%>"%~dp0\%BUILDDEPTHCACHE%"
 
 echo.
 set START_TIME=%TIME%
@@ -160,13 +172,13 @@ cd "%DEPS_DIR%"
 :: by modifying this file and using goto.
 :Boost
 :: NOTE Boost < 1.64 doesn't work without tricks if the user has only VS 2017 installed and no earlier versions.
-set BOOST_VERSION=1.67.0
+set BOOST_VERSION=1.71.0
 :: Version string with underscores instead of dots.
 set BOOST_VER=%BOOST_VERSION:.=_%
 :: DEPENDENCY_NAME is used for logging and DEPENDENCY_DIR for saving from some redundant typing
 set DEPENDENCY_NAME=Boost %BOOST_VERSION%
 set DEPENDENCY_DIR=%DEPS_DIR%\boost_%BOOST_VER%
-set BOOST_LIBRARYDIR=%DEPENDENCY_DIR%\stage\%VS_PLATFORM%\lib
+set BOOST_LIBRARYDIR=%DEPENDENCY_DIR%\stage\%GEN_SHORTHAND%\lib
 :: NOTE Also zip download exists, if encountering problems with 7z for some reason.
 set ZIP_EXT=7z
 set BOOST_ZIP=boost_%BOOST_VER%.%ZIP_EXT%
@@ -183,7 +195,7 @@ if not exist "%DEPENDENCY_DIR%\project-config.jam". (
     IF NOT EXIST "%DEPENDENCY_DIR%\boost.css" GOTO :Error
     cd "%DEPENDENCY_DIR%"
     call cecho.cmd 0 13 "Building Boost build script."
-    call bootstrap msvc
+    call bootstrap %BOOST_BOOTSTRAP_VER%
     IF NOT %ERRORLEVEL%==0 GOTO :Error
 )
 
@@ -192,14 +204,10 @@ set BOOST_LIBS=--with-system --with-regex --with-thread --with-program_options -
 cd "%DEPENDENCY_DIR%"
 call cecho.cmd 0 13 "Building %DEPENDENCY_NAME% %BOOST_LIBS% Please be patient, this will take a while."
 IF EXIST "%DEPENDENCY_DIR%\bin.v2\project-cache.jam" del "%DEPS_DIR%\boost\bin.v2\project-cache.jam"
-:: BOOST_VC_VER can be empty (or needs to be) for newer VS versions
-set BOOST_VC_VER=
-if %VS_VER% LSS 2017 (
-    set BOOST_VC_VER=-%VC_VER%.0
-)
 
-call .\b2 toolset=msvc%BOOST_VC_VER% runtime-link=static address-model=%ARCH_BITS% -j%IFCOS_NUM_BUILD_PROCS% ^
-    variant=%DEBUG_OR_RELEASE_LOWERCASE% %BOOST_LIBS% stage --stagedir=stage/vs%VS_VER%-%VS_PLATFORM% 
+call .\b2 toolset=%BOOST_TOOLSET% runtime-link=static address-model=%ARCH_BITS% -j%IFCOS_NUM_BUILD_PROCS% ^
+    variant=%DEBUG_OR_RELEASE_LOWERCASE% %BOOST_WIN_API% %BOOST_LIBS% stage --stagedir=stage/%GEN_SHORTHAND% 
+
 IF NOT %ERRORLEVEL%==0 GOTO :Error
 
 :JSON
@@ -237,10 +245,10 @@ if %IFCOS_USE_OCCT%==FALSE goto :OCE
 set OCCT_VERSION=7.3.0p3
 SET OCCT_VER=V%OCCT_VERSION:.=_%
 
-set OCC_INCLUDE_DIR=%INSTALL_DIR%\opencascade-%OCCT_VERSION%\inc>>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
-set OCC_LIBRARY_DIR=%INSTALL_DIR%\opencascade-%OCCT_VERSION%\win%ARCH_BITS%\lib>>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
-echo OCC_INCLUDE_DIR=%OCC_INCLUDE_DIR%>>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
-echo OCC_LIBRARY_DIR=%OCC_LIBRARY_DIR%>>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
+set OCC_INCLUDE_DIR=%INSTALL_DIR%\opencascade-%OCCT_VERSION%\inc>>"%~dp0\%BUILDDEPTHCACHE%"
+set OCC_LIBRARY_DIR=%INSTALL_DIR%\opencascade-%OCCT_VERSION%\win%ARCH_BITS%\lib>>"%~dp0\%BUILDDEPTHCACHE%"
+echo OCC_INCLUDE_DIR=%OCC_INCLUDE_DIR%>>"%~dp0\%BUILDDEPTHCACHE%"
+echo OCC_LIBRARY_DIR=%OCC_LIBRARY_DIR%>>"%~dp0\%BUILDDEPTHCACHE%"
 
 :: OCCT has many dependencies but FreeType is the only mandatory
 set DEPENDENCY_NAME=FreeType
@@ -284,7 +292,7 @@ call :BuildSolution "%DEPENDENCY_DIR%\%BUILD_DIR%\OCCT.sln" %BUILD_CFG%
 if not %ERRORLEVEL%==0 goto :Error
 call :InstallCMakeProject "%DEPENDENCY_DIR%\%BUILD_DIR%" %BUILD_CFG%
 if not %ERRORLEVEL%==0 goto :Error
-:: Use a single lib directory for for release and debug libraries as is done with OCE
+:: Use a single lib directory for release and debug libraries as is done with OCE
 if not exist "%OCC_LIBRARY_DIR%". mkdir "%OCC_LIBRARY_DIR%"
 :: NOTE OCCT (at least occt-V7_0_0-9059ca1) directory creation code is hardcoded and doesn't seem handle future VC versions
 set OCCT_VC_VER=%VC_VER%
@@ -303,10 +311,10 @@ del "%INSTALL_DIR%\opencascade-%OCCT_VERSION%\*.bat"
 goto :Python
 
 :OCE
-set OCC_INCLUDE_DIR=%INSTALL_DIR%\oce\include\oce>>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
-set OCC_LIBRARY_DIR=%INSTALL_DIR%\oce\Win%ARCH_BITS%\lib>>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
-echo OCC_INCLUDE_DIR=%OCC_INCLUDE_DIR%>>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
-echo OCC_LIBRARY_DIR=%OCC_LIBRARY_DIR%>>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
+set OCC_INCLUDE_DIR=%INSTALL_DIR%\oce\include\oce>>"%~dp0\%BUILDDEPTHCACHE%"
+set OCC_LIBRARY_DIR=%INSTALL_DIR%\oce\Win%ARCH_BITS%\lib>>"%~dp0\%BUILDDEPTHCACHE%"
+echo OCC_INCLUDE_DIR=%OCC_INCLUDE_DIR%>>"%~dp0\%BUILDDEPTHCACHE%"
+echo OCC_LIBRARY_DIR=%OCC_LIBRARY_DIR%>>"%~dp0\%BUILDDEPTHCACHE%"
 
 set DEPENDENCY_NAME=Open CASCADE Community Edition
 set DEPENDENCY_DIR=%DEPS_DIR%\oce
@@ -346,8 +354,8 @@ set PYTHON_INSTALLER=python-%PYTHON_VERSION%%PYTHON_AMD64_POSTFIX%.msi
 :: NOTE/TODO 3.5.0 doesn't use MSI any longer, but exe: set PYTHON_INSTALLER=python-%PYTHON_VERSION%%PYTHON_AMD64_POSTFIX%.exe
 IF "%IFCOS_INSTALL_PYTHON%"=="TRUE" (
     REM Store Python versions to BuildDepsCache.txt for run-cmake.bat
-    echo PY_VER_MAJOR_MINOR=%PY_VER_MAJOR_MINOR%>>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
-    echo PYTHONHOME=%PYTHONHOME%>>"%~dp0\BuildDepsCache-%TARGET_ARCH%.txt"
+    echo PY_VER_MAJOR_MINOR=%PY_VER_MAJOR_MINOR%>>"%~dp0\%BUILDDEPTHCACHE%"
+    echo PYTHONHOME=%PYTHONHOME%>>"%~dp0\%BUILDDEPTHCACHE%"
 
     cd "%DEPS_DIR%"
     call :DownloadFile https://www.python.org/ftp/python/%PYTHON_VERSION%/%PYTHON_INSTALLER% "%DEPS_DIR%" %PYTHON_INSTALLER%
@@ -502,7 +510,13 @@ pushd %BUILD_DIR%
 :: TODO make deleting cache a parameter for this subroutine? We probably want to delete the 
 :: cache always e.g. when we've had new changes in the repository.
 IF %BUILD_TYPE%==Rebuild IF EXIST CMakeCache.txt. del CMakeCache.txt
-cmake .. -G %GENERATOR% %*
+
+IF DEFINED VS_TOOLSET (
+    cmake .. -G %GENERATOR% -A %VS_PLATFORM% -T %VS_TOOLSET% %*
+) ELSE (
+    cmake .. -G %GENERATOR% -A %VS_PLATFORM% %*
+)
+
 set RET=%ERRORLEVEL%
 popd
 exit /b %RET%
diff --git a/win/readme.md b/win/readme.md
index e8e4c510..31ada3fe 100644
--- a/win/readme.md
+++ b/win/readme.md
@@ -18,36 +18,36 @@ of the shell scripts before using them. Note that contrary to MSVC, with MSYS al
 built or used as static libraries. Currently Release build is used for all libraries.
 
 ### MSVC
-Execute `build-deps.cmd` to fetch, build and install the dependencies. The batch file will print the requirements for
-a successful execution. The script allows a few user-configurable build options which are listed in the usage
-instructions. Either edit the script file or set these values before running the script.
-
-`build-deps.cmd` expects a CMake generator as `%1` and a build configuration type (`RelWithDebInfo`, `Release`,
-`MinSizeRel`, or `Debug`, defaults to `RelWithDebInfo`) as `%2`. If the generator is not provided, the generator is
-deduced from the MSVC environment variables. User-friendly VS generator shorthands are supported, e.g.
-`vs2013-x86` or `vs2015-x64`, and these are converted to the appropriate CMake ones by the scripts. A build type
-(`Build`, `Rebuild`, or `Clean`, defaults to `Build`) can be provided as `%3`. See `vs-cfg.cmd` if you wish to change
-the defaults. The batch file will create `deps\` and `deps-vs<VERSION>-<ARCHITECTURE>-installed\` directories to the
-project root. Debug and release builds of the dependencies can co-exist by simply running
+Launch the proper Visual Studio command prompt, cd to the 'win' directory inside the IfcOpenShell directory and execute `build-deps.cmd` to fetch, build and install the dependencies. The batch file will print the requirements for a successful execution. The script allows a few user-configurable build options which are listed below.
+
+`build-deps.cmd` expects a CMake generator as `%1` and a build configuration type (`RelWithDebInfo`, `Release`, `MinSizeRel`, or `Debug`, defaults to `RelWithDebInfo`) as `%2`. If the generator is not provided, the generator is deduced from the MSVC environment variables.
+
+User-friendly CMake Visual Studio generator shorthands are supported. They are converted to the appropriate CMake generators and options. Shorthands are indeed the preferable way to specify the generator, since they allow a more accurate platform and toolset configuration. Here are some examples:
+```
+"vs2013"             => cmake -G "Visual Studio 12 2013" -A Win32
+"vs2013-x86"         => cmake -G "Visual Studio 12 2013" -A Win32
+"vs2015-x64"         => cmake -G "Visual Studio 14 2015" -A x64
+"vs2017-ARM64"       => cmake -G "Visual Studio 15 2017" -A ARM64
+"vs2019-x86-v141_xp" => cmake -G "Visual Studio 16 2019" -A Win32 -T v141_xp
+```
+Of course not all Visual C++ compilers support any platform or toolset, refer to the Visual Studio and CMake documentation for this. If you do not specify a toolset, the compiler will use the default toolset for the version, i.e. vs2019 will use the v142 toolset.
+
+A build type (`Build`, `Rebuild`, or `Clean`, defaults to `Build`) can be provided as `%3`.
+
+See `vs-cfg.cmd` if you wish to change the defaults. The batch file will create `deps\` and `deps-vs<VERSION>-<PLATFORM>[-<TOOLSET>]-installed\` directories to the project root. Debug and release builds of the dependencies can co-exist by simply running:

     > build-deps.cmd <GENERATOR> Debug
     > build-deps.cmd <GENERATOR> <Release|RelWithDebInfo|MinSizeRel>

 
-After the dependencies are build, execute `run-cmake.bat`. The batch file expects a CMake generator as `%1` and the
-rest of possible parameters are passed as is. If a generator is not provided, the generator is read from the
-BuildDepsCache file, or tried to be deduced from the location of `cl.exe`. If passing build options for the script,
-the generator must be always passed as the first option:
+After the dependencies are build, execute `run-cmake.bat`. The batch file expects a CMake generator as `%1`, that is interpreted just like the `build-deps.cmd` script, and the rest of possible parameters are passed as is. If a generator is not provided, the generator is read from the BuildDepsCache file, or tried to be deduced from the location of `cl.exe`. If passing build options for the script, the generator must be always passed as the first option:

    > run-cmake.bat vs2015-x64 -DUSE_IFC4=1 -DBUILD_IFCPYTHON=0
 
-**If you wish to use any library from a custom location, modify the paths in `run-cmake.bat` accordingly**. The batch
-script will create a folder of form `build-vs<VERSION>-<ARCHITECTURE>\` which will contain the solution and project
-files for MSVC.
+**If you wish to use any library from a custom location, modify the paths in `run-cmake.bat` accordingly**. The batch script will create a folder of form `build-vs<VERSION>-<PLATFORM>[-<TOOLSET>]\` which will contain the solution and project files for MSVC.
 
-Note that building IfcOpenShell as 64-bit is recommended as many of real life IFC files has been observed to take
-easily more than 2 GBs of RAM while converting.
+Note that building IfcOpenShell as 64-bit is recommended as many of real life IFC files has been observed to take easily more than 2 GBs of RAM while converting.
 
 After this, one can build the project using the `IfcOpenShell.sln` file in the build folder. Build the `INSTALL` project
 if wanted. Convenience batch files `build-ifcopenshell.bat` and `install-ifcopenshell.bat` can also be used. The batch
diff --git a/win/run-cmake.bat b/win/run-cmake.bat
index 8154b3a4..eb2b5977 100755
--- a/win/run-cmake.bat
+++ b/win/run-cmake.bat
@@ -52,18 +52,18 @@ if not (%1)==() (
 )
 
 pushd ..
-set CMAKE_INSTALL_PREFIX=%CD%\installed-vs%VS_VER%-%TARGET_ARCH%
+set CMAKE_INSTALL_PREFIX=%CD%\installed-%GEN_SHORTHAND%
 popd
 
 IF NOT EXIST ..\%BUILD_DIR%. mkdir ..\%BUILD_DIR%
 pushd ..\%BUILD_DIR%
 
 :: tfk: todo remove duplication
-set BOOST_VERSION=1.67.0
+set BOOST_VERSION=1.71.0
 set BOOST_VER=%BOOST_VERSION:.=_%
 
 set BOOST_ROOT=%DEPS_DIR%\boost_%BOOST_VER%
-set BOOST_LIBRARYDIR=%BOOST_ROOT%\stage\vs%VS_VER%-%VS_PLATFORM%\lib
+set BOOST_LIBRARYDIR=%BOOST_ROOT%\stage\%GEN_SHORTHAND%\lib
 if not defined OCC_INCLUDE_DIR set OCC_INCLUDE_DIR=%INSTALL_DIR%\oce\include\oce
 if not defined OCC_LIBRARY_DIR set OCC_LIBRARY_DIR=%INSTALL_DIR%\oce\Win%ARCH_BITS%\lib
 set OPENCOLLADA_INCLUDE_DIR=%INSTALL_DIR%\OpenCOLLADA\include\opencollada
@@ -82,6 +82,8 @@ set JSON_INCLUDE_DIR=%INSTALL_DIR%\json
 echo.
 call cecho.cmd 0 10 "Script configuration:"
 echo   Generator    = %GENERATOR%
+echo   Architecture = %VS_PLATFORM%
+echo   Toolset      = %VS_TOOLSET%
 echo   Arguments    = %ARGUMENTS%
 echo.
 call cecho.cmd 0 10 "Dependency Environment Variables for %PROJECT_NAME%:"
@@ -107,7 +109,13 @@ set CMAKELISTS_DIR=..\cmake
 :: Delete CMakeCache.txt if command-line options were provided for this batch script.
 if not (%1)==() if exist CMakeCache.txt. del /Q CMakeCache.txt
 call cecho.cmd 0 13 "Running CMake for %PROJECT_NAME%."
-cmake.exe %CMAKELISTS_DIR% -G %GENERATOR% -DCMAKE_INSTALL_PREFIX="%CMAKE_INSTALL_PREFIX%" %ARGUMENTS%
+
+IF DEFINED VS_TOOLSET (
+	cmake.exe %CMAKELISTS_DIR% -G %GENERATOR% -A %VS_PLATFORM% -T %VS_TOOLSET% -DCMAKE_INSTALL_PREFIX="%CMAKE_INSTALL_PREFIX%" %ARGUMENTS%
+) ELSE (
+	cmake.exe %CMAKELISTS_DIR% -G %GENERATOR% -A %VS_PLATFORM% -DCMAKE_INSTALL_PREFIX="%CMAKE_INSTALL_PREFIX%" %ARGUMENTS%
+)
+
 IF NOT %ERRORLEVEL%==0 GOTO :Error
 
 echo.
diff --git a/win/vs-cfg.cmd b/win/vs-cfg.cmd
index 992de2ef..4b461883 100644
--- a/win/vs-cfg.cmd
+++ b/win/vs-cfg.cmd
@@ -17,15 +17,23 @@
 ::                                                                             ::
 ::::::::::::::::::::::::::::::::::::::::::::::::::::::: ::::::::::::::::::::::::::
 
-:: This script initializes various Visual Studio related environment variables needed for building.
-:: the dependencies. This batch file expects a CMake generator as %1. If %1 is not provided, it is
-:: deduced from the VisualStudioVersion environment variable and from the location of cl.exe.
-:: User-friendly VS generators are allowed (e.g. "vs2013-x86") and converted to the appropriate CMake ones.
-
-:: NOTE This batch file expects the generator string to be CMake 3.0.0 and newer format, i.e.
-:: "Visual Studio 10 2010" instead of "Visual Studio 10". However, one can use this batch file
-:: also with CMake 2 as the generator will be converted into the older format if necessary.
-
+:: This script initializes various Visual Studio related environment variables.
+::
+:: It expects a CMake generator as %1. If %1 is not provided, it is deduced from
+:: the VisualStudioVersion environment variable and from the location of cl.exe.
+::
+:: The generator string must be in the CMake 3.0 and newer format, i.e.
+:: "Visual Studio 10 2010" instead of "Visual Studio 10"
+::
+:: User-friendly VS generators are preferred, since they permit a more accurate
+:: platform and toolset configuration. Some examples:
+::
+::   "vs2013"             => cmake -G "Visual Studio 12 2013" -A Win32
+::   "vs2013-x86"         => cmake -G "Visual Studio 12 2013" -A Win32
+::   "vs2015-x64"         => cmake -G "Visual Studio 14 2015" -A x64
+::   "vs2017-ARM64"       => cmake -G "Visual Studio 15 2017" -A ARM64
+::   "vs2019-x86-v141_xp" => cmake -G "Visual Studio 16 2019" -A Win32 -T v141_xp
+::
 :: NOTE: The delayed environment variable expansion needs to be enabled before calling this.
 
 @echo off
@@ -33,34 +41,42 @@
 set GENERATOR=%1
 
 :: Supported Visual Studio versions:
-set GENERATORS[0]="Visual Studio 9 2008 Win64"
 set GENERATORS[1]="Visual Studio 9 2008"
-set GENERATORS[2]="Visual Studio 10 2010 Win64"
-set GENERATORS[3]="Visual Studio 10 2010"
-set GENERATORS[4]="Visual Studio 11 2012 Win64"
-set GENERATORS[5]="Visual Studio 11 2012"
-set GENERATORS[6]="Visual Studio 12 2013 Win64"
-set GENERATORS[7]="Visual Studio 12 2013"
-set GENERATORS[8]="Visual Studio 14 2015 Win64"
-set GENERATORS[9]="Visual Studio 14 2015"
-:: NOTE VC version for VS 2017 is not 15 but 14.1: have to wait and see
-:: if CMake generator string is updated to reflect this.
-set GENERATORS[10]="Visual Studio 15 2017 Win64"
-set GENERATORS[11]="Visual Studio 15 2017"
-set LAST_GENERATOR_IDX=11
-
-set STEP=2
+set GENERATORS[2]="Visual Studio 10 2010"
+set GENERATORS[3]="Visual Studio 11 2012"
+set GENERATORS[4]="Visual Studio 12 2013"
+set GENERATORS[5]="Visual Studio 14 2015"
+set GENERATORS[6]="Visual Studio 15 2017"
+set GENERATORS[7]="Visual Studio 16 2019"
+set LAST_GENERATOR_IDX=7
+
 :: Is generator shorthand used?
 set GEN_SHORTHAND=!GENERATOR:vs=!
+
 if not "!GEN_SHORTHAND!"=="" if !GEN_SHORTHAND!==!GENERATOR! goto :GeneratorShorthandCheckDone
-set START=%LAST_GENERATOR_IDX%
+set VS_PLATFORM=Win32
 :: "echo if" trick from http://stackoverflow.com/a/8758579
-echo(!GEN_SHORTHAND! | findstr /c:"-x86" >nul && ( set START=1 )
-echo(!GEN_SHORTHAND! | findstr /c:"-x64" >nul && ( set START=0 )
-set VS_VER=!GEN_SHORTHAND:-x86=!
-set VS_VER=!VS_VER:-x64=!
+echo(!GEN_SHORTHAND! | findstr /c:"-x86"     >nul && ( set "VS_PLATFORM=Win32" )
+echo(!GEN_SHORTHAND! | findstr /c:"-x64"     >nul && ( set "VS_PLATFORM=x64" )
+echo(!GEN_SHORTHAND! | findstr /c:"-ARM"     >nul && ( set "VS_PLATFORM=ARM" )
+echo(!GEN_SHORTHAND! | findstr /c:"-ARM64"   >nul && ( set "VS_PLATFORM=ARM64" )
+
+echo(!GEN_SHORTHAND! | findstr /c:"-v90"     >nul && ( set "VS_TOOLSET=v90" )     && ( set "BOOST_TOOLSET=9.0" )
+echo(!GEN_SHORTHAND! | findstr /c:"-v100"    >nul && ( set "VS_TOOLSET=v100" )    && ( set "BOOST_TOOLSET=10.0" )
+echo(!GEN_SHORTHAND! | findstr /c:"-v110"    >nul && ( set "VS_TOOLSET=v110" )    && ( set "BOOST_TOOLSET=11.0" )
+echo(!GEN_SHORTHAND! | findstr /c:"-v110_xp" >nul && ( set "VS_TOOLSET=v110_xp" ) && ( set "BOOST_TOOLSET=11.0" )
+echo(!GEN_SHORTHAND! | findstr /c:"-v120"    >nul && ( set "VS_TOOLSET=v120" )    && ( set "BOOST_TOOLSET=12.0" )
+echo(!GEN_SHORTHAND! | findstr /c:"-v120_xp" >nul && ( set "VS_TOOLSET=v120_xp" ) && ( set "BOOST_TOOLSET=12.0" )
+echo(!GEN_SHORTHAND! | findstr /c:"-v140"    >nul && ( set "VS_TOOLSET=v140" )    && ( set "BOOST_TOOLSET=14.0" )
+echo(!GEN_SHORTHAND! | findstr /c:"-v140_xp" >nul && ( set "VS_TOOLSET=v140_xp" ) && ( set "BOOST_TOOLSET=14.0" )
+echo(!GEN_SHORTHAND! | findstr /c:"-v141"    >nul && ( set "VS_TOOLSET=v141" )    && ( set "BOOST_TOOLSET=14.1" )
+echo(!GEN_SHORTHAND! | findstr /c:"-v141_xp" >nul && ( set "VS_TOOLSET=v141_xp" ) && ( set "BOOST_TOOLSET=14.1" )
+echo(!GEN_SHORTHAND! | findstr /c:"-v142"    >nul && ( set "VS_TOOLSET=v142" )    && ( set "BOOST_TOOLSET=14.2" )
+
+SET VS_VER=%GEN_SHORTHAND:~0,4%
+
 echo(!GENERATOR! | findstr /c:"vs20" >nul && (
-    for /l %%i in (!START!,!STEP!,%LAST_GENERATOR_IDX%) do (
+    for /L %%i in (0,1,%LAST_GENERATOR_IDX%) do (
         echo(!GENERATORS[%%i]! | findstr /c:"!VS_VER!" >nul && (
             set GENERATOR=!GENERATORS[%%i]!
             goto :GeneratorShorthandCheckDone
@@ -77,8 +93,8 @@ set START=%ERRORLEVEL%
 
 IF "!GENERATOR!"=="" IF NOT "%VisualStudioVersion%"=="" (
     set VC_VER=%VisualStudioVersion:.0=%
-    FOR /l %%i in (%START%,%STEP%,%LAST_GENERATOR_IDX%) DO (
-        REM NOTE add space before VC_VER so that e.g. "12" doesn't match with "2012"
+    FOR /L %%i in (%START%,1,%LAST_GENERATOR_IDX%) DO (
+        :: NOTE add space before VC_VER so that e.g. "12" doesn't match with "2012"
         echo(!GENERATORS[%%i]! | findstr /c:" !VC_VER!" >nul && (
             set GENERATOR=!GENERATORS[%%i]!
             call utils\cecho.cmd black cyan "Generator not passed, but VisualStudioVersion=%VisualStudioVersion% environment variable detected:"
@@ -87,6 +103,7 @@ IF "!GENERATOR!"=="" IF NOT "%VisualStudioVersion%"=="" (
         )
     )
 )
+
 :: Check that the used CMake version supports the chosen generator
 set GENERATOR_CHECK=%GENERATOR: Win64=%
 cmake --help | findstr /c:%GENERATOR_CHECK% >nul
@@ -95,21 +112,23 @@ call utils\cecho.cmd 0 12 "%~nx0: The used CMake version does not support '`"!GE
 exit /b 1
 )
 
-FOR /l %%i in (0,1,%LAST_GENERATOR_IDX%) DO (
+FOR /L %%i in (0,1,%LAST_GENERATOR_IDX%) DO (
     IF !GENERATOR!==!GENERATORS[%%i]! GOTO :GeneratorValid
 )
 call utils\cecho.cmd 0 12 "%~nx0: Invalid or unsupported CMake generator string passed: '`"!GENERATOR!`'"- cannot proceed."
 echo Supported CMake generator strings:
-FOR /l %%i in (0,1,%LAST_GENERATOR_IDX%) DO (
+FOR /L %%i in (0,1,%LAST_GENERATOR_IDX%) DO (
     echo !GENERATORS[%%i]!
 )
 exit /b 1
 
 :GeneratorValid
-:: Figure out the build configuration from the CMake generator string.
+
+IF DEFINED VS_PLATFORM goto :PlatformDefined
+
+:: If we do not have defined a platform yet,
+:: figure out the build configuration from the CMake generator string.
 :: Are we building 32-bit or 64-bit version.
-set ARCH_BITS=32
-set TARGET_ARCH=x86
 :: Visual Studio platform name, Win32 (i.e. x86) or x64.
 set VS_PLATFORM=Win32
 
@@ -122,12 +141,56 @@ FOR %%i IN (%GENERATOR_SPLIT%) DO (
     IF !LEN!==1 set VC_VER=%%i
     IF !LEN!==2 set VC_VER=%%i
     IF !LEN!==4 set VS_VER=%%i
-    REM Are going to perform a 64-bit build?
+    :: Are going to perform a 64-bit build?
     IF %%i==Win64 (
+        set VS_PLATFORM=x64
+    )
+)
+
+:PlatformDefined
+
+:: determine the Visual C++ default version
+IF %VS_VER%==2008 ( set "VC_VER=9.0" )
+IF %VS_VER%==2010 ( set "VC_VER=10.0" )
+IF %VS_VER%==2012 ( set "VC_VER=11.0" )
+IF %VS_VER%==2013 ( set "VC_VER=12.0" )
+IF %VS_VER%==2015 ( set "VC_VER=14.0" )
+IF %VS_VER%==2017 ( set "VC_VER=14.1" )
+IF %VS_VER%==2019 ( set "VC_VER=14.2" )
+
+:: determine the argument for Boost bootstrap
+set BOOST_BOOTSTRAP_VER=vc%VC_VER%
+set BOOST_BOOTSTRAP_VER=%BOOST_BOOTSTRAP_VER:.=%
+
+:: determine the toolset and winapi for Boost b2
+IF DEFINED VS_TOOLSET (
+    set BOOST_TOOLSET=msvc-%BOOST_TOOLSET%
+    if "!VS_TOOLSET:~-3!"=="_xp" (
+        set BOOST_WIN_API=define=BOOST_USE_WINAPI_VERSION=0x0501
+        )
+) ELSE (
+    set BOOST_TOOLSET=msvc-%VC_VER%
+    set BOOST_WIN_API=
+)
+
+IF %VS_PLATFORM%==Win32 (
+    set ARCH_BITS=32
+    set TARGET_ARCH=x86
+)
+
+IF %VS_PLATFORM%==x64 (
     set ARCH_BITS=64
     set TARGET_ARCH=x64
-        set VS_PLATFORM=x64
 )
+
+IF %VS_PLATFORM%==ARM (
+    set ARCH_BITS=32
+    set TARGET_ARCH=ARM
+)
+
+IF %VS_PLATFORM%==ARM64 (
+    set ARCH_BITS=64
+    set TARGET_ARCH=ARM64
 )
 
 :: Check CMake version and convert possible new format (>= 3.0) generator names to the old versions if using older CMake for VS <= 2013,
@@ -142,7 +205,25 @@ set GENERATOR=%GENERATOR: 2012=%
 set GENERATOR=%GENERATOR: 2010=%
 :CMake3AndNewer
 
-set GEN_SHORTHAND=vs%VS_VER%-%TARGET_ARCH%
+    :: check variables for debugging
+     echo GENERATOR:           [!GENERATOR!]
+     echo VS_VER:              [!VS_VER!]
+     echo VS_PLATFORM:         [!VS_PLATFORM!]
+     echo VS_TOOLSET:          [!VS_TOOLSET!]
+     echo VC_VER:              [!VC_VER!]
+     echo ARCH_BITS:           [!ARCH_BITS!]
+     echo TARGET_ARCH:         [!TARGET_ARCH!]
+     echo BOOST_BOOTSTRAP_VER: [!BOOST_BOOTSTRAP_VER!]
+     echo BOOST_TOOLSET:       [!BOOST_TOOLSET!]
+     echo BOOST_WIN_API:       [!BOOST_WIN_API!]
+     pause
+
+IF DEFINED VS_TOOLSET (
+    set GEN_SHORTHAND=vs%VS_VER%-%VS_PLATFORM%-%VS_TOOLSET%
+) ELSE (
+    set GEN_SHORTHAND=vs%VS_VER%-%VS_PLATFORM%
+)
+
 :: VS project file extension is different on older VS versions
 set VCPROJ_FILE_EXT=vcxproj
 IF %VS_VER%==2008 set VCPROJ_FILE_EXT=vcproj
@@ -158,7 +239,7 @@ set PATH=%~dp0utils;%PATH%
 pushd ..
 set DEPS_DIR=%CD%\deps
 set INSTALL_DIR=%CD%\deps-%GEN_SHORTHAND%-installed
-REM set INSTALL_DIR=%CD%\deps-vs%VS_VER%-%TARGET_ARCH%-%DEBUG_OR_RELEASE_LOWERCASE%-installed
+:: set INSTALL_DIR=%CD%\deps-vs%VS_VER%-%TARGET_ARCH%-%DEBUG_OR_RELEASE_LOWERCASE%-installed
 :: BUILD_DIR is a relative build directory used for CMake-based projects
 set BUILD_DIR=build-%GEN_SHORTHAND%
 popd
```

aothms:
> It is impossible to successfully complete the build of the Debug version of the IfcOpenShell solution, as the Visual Studio projects are incorrectly set to link to ...\OpenCOLLADA\lib\opencollada\xml.lib instead of xmld.lib, which is the version produced by the scripts when the 'Debug' parameter is used. I guess this is not noticeable in case the release version has been built as well or only the release version.

xml.lib is chosen here 

```
set LIBXML2_LIBRARIES=%INSTALL_DIR%\OpenCOLLADA\lib\opencollada\xml.lib
```

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/win/run-cmake.bat#L72

I think you can simply postfix it with %POSTFIX_D% according to https://github.com/IfcOpenShell/IfcOpenShell/blob/ac40b90f2769d550c153eff13ff23ee5fbd41469/win/build-type-cfg.cmd#L57

> So what I would do instead is to provide Visual Studio support only from 2013 onwards.

I am fine with that

> However, errors remain.

Are those all. Again thanks for your effort on this, it looks really comprehensive.

ahladik:
> xml.lib is chosen here
> I think you can simply postfix it with %POSTFIX_D%

This does not seem correct to me. Besides the fact that POSTFIX_D is never used in the scripts, to initialize POSTFIX_D, 'build-type-cfg.cmd' must be called inside 'run-cmake.bat' and this is not expected.
In fact in this case also the build target should be specified when calling 'run-cmake.bat'.

But if I add it nonetheless and use
`set LIBXML2_LIBRARIES=%INSTALL_DIR%\OpenCOLLADA\lib\opencollada\xml%POSTFIX_D%.lib`
I get a few warnings from CMake regarding library names and the result is the same.

It looks like that the 'd' suffix problem is managed inside CMakeLists.txt, where I see
```
IF(MSVC)
	add_debug_variants(OPENCOLLADA_LIBRARIES "${OPENCOLLADA_LIBRARIES}" d)
ENDIF()
```
So I have done something similar:
```
IF(MSVC)
	add_debug_variants(LIBXML2_LIBRARIES "${LIBXML2_LIBRARIES}" d)
ENDIF()
```
This seems to work.

But there are still a couple of wrong library links to fix.

aothms:
Ok, sorry for pointing you in the wrong direction

ahladik:
I have pushed my last changes. All seems to work now except for one thing. If I try to compile with VS 2019 using a different toolset than the default, for example with:
```
build-deps.cmd vs2019-x64-v141_xp Debug
run-cmake.bat vs2019-x64-v141_xp -DBUILD_IFCPYTHON=0
```
the two steps are completed without errors, but when I finally try to compile the entire IfcOpenShell solution, I get the following errors:
```
2>LINK : fatal error LNK1104: impossibile aprire il file 'libboost_regex-vc142-mt-sgd-x64-1_71.lib'
2>Compilazione progetto "IfcGeomServer.vcxproj" NON COMPLETATA.
1>LINK : fatal error LNK1104: impossibile aprire il file 'libboost_regex-vc142-mt-sgd-x64-1_71.lib'
1>Compilazione progetto "IfcConvert.vcxproj" NON COMPLETATA.
```
The other projects are successfully compiled. That library doesn't exist, because instead there is the one compiled with the right toolset:
`libboost_regex-vc141-mt-sgd-x64-1_71.lib`
that all projects correctly refer to.

None of the Visual Studio projects inside the IfcOpenShell solution contain an explicit link to that library, only to the correct one. Nevertheless, inside the binary .lib files and in some .obj files of the IfcOpenShell solution I have found that the non-existent library is referenced with:
`/DEFAULTLIB:"libboost_regex-vc142-mt-sgd-x64-1_71.lib"`

I have not been able to figure out how such an instruction is inserted and frankly have run out of ideas. Is someone please able to explain this?

aothms:
> I have not been able to figure out how such an instruction is inserted and frankly have run out of ideas. Is someone please able to explain this?

Well firstly thanks again for your effort, I don't think we need to solve everything at once.

Linking boost is quite magical on Windows. One thing maybe to look at for clues is this `BOOST_ALL_NO_LIB` preproc directive that we set in case we were compiling for anaconda. https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/cmake/CMakeLists.txt#L150 The default is on and the boost header will add a pragma directive to link to the correct libraries based on compiler flags (debug, msvc version) maybe there is something faulty here for the toolset. I do see we also explicitly link to Boost_LIBRARIES in CMake which is maybe redundant. Maybe that's an explanation for what you're seeing?

ahladik:
You're right, that's the point.
This is the piece of code in CMakeLists.txt that works for me.
```
IF(WIN32 AND ("$ENV{CONDA_BUILD}" STREQUAL ""))
    SET(Boost_USE_STATIC_LIBS ON)
    SET(Boost_USE_STATIC_RUNTIME ON)
    SET(Boost_USE_MULTITHREADED ON)
    # Disable Boost's autolinking as the libraries to be linked to are supplied
    # already by CMake, and wrong libraries would be asked for when code is
    # compiled with a toolset different from default.
    if(MSVC)
        ADD_DEFINITIONS(-DBOOST_ALL_NO_LIB)
        # Necessary for boost version >= 1.67
        SET(BCRYPT_LIBRARIES "bcrypt.lib")
    ENDIF()
ELSE()
    # Disable Boost's autolinking as the libraries to be linked to are supplied
    # already by CMake, and it's going to conflict if there are multiple, as is
    # the case in conda-forge's libboost feedstock.
    ADD_DEFINITIONS(-DBOOST_ALL_NO_LIB)
    IF(WIN32)
        # Necessary for boost version >= 1.67
        SET(BCRYPT_LIBRARIES "bcrypt.lib")
    ENDIF()
ENDIF()
```
I'm not sure if it can be cleaned up. What do you think?

ahladik:
One more thing, if you agree I would bump the Boost version up to 1.74.

aothms:
When I use the **x64** native tools. I get a 32bit build. I'll also investigate in the mean time.

afbeelding

win/vs-cfg.cmd
```diff
@@ -1,175 +1,247 @@
...
+set GENERATORS[2]="Visual Studio 14 2015"
+set GENERATORS[3]="Visual Studio 15 2017"
+set GENERATORS[4]="Visual Studio 16 2019"
+set LAST_GENERATOR_IDX=4
+
+:: Is generator shorthand used?
+set GEN_SHORTHAND=!GENERATOR:vs=!
+
+if not "!GEN_SHORTHAND!"=="" if !GEN_SHORTHAND!==!GENERATOR! goto :GeneratorShorthandCheckDone
+set VS_PLATFORM=Win32
```
aothms, line 54: Should we rather than defaulting to `Win32`, use %VSCMD_ARG_TGT_ARCH% which should be set by the dev tools command prompt (and maybe when that is empty) default to Win32?

ahladik: Good idea, done.

ahladik:
Unfortunately, I found another problem.

Visual C++ 2019 is unable to fully compile OCCT for the x86 architecture in 'Release' configuration.

The compilation remains stuck indefinitely trying to compile the 'TKBO' library. The 'TKTopAlgo' and 'TKV3d' libraries also fail to compile, probably because they are later in the order. Also the installation step is obviously not performed.

However, I have verified that by interrupting the process and manually copying the remaining already compiled libraries, the IfcOpenShell solution successfully compiles the 'Release' configuration as well. So such libraries seem superfluous, at least if you just compile IfcGeom and IfcParse (I don't use IfcConvert or IfcGeomServer).

Compiling in x86 Debug and x64 Debug and Release configurations shows no problems.

This happens using "build-deps.cmd vs2019-x86-v141_xp Release". I will do an experiment to see if toolset v142 behaves differently.

ahladik:
The problem also occurs with toolset v142 and is a known issue, fixed with OCCT version 7.5.0.
See https://tracker.dev.opencascade.org/view.php?id=31628

ahladik:
Actually there is a workaround, which is to set the whole program optimization (/GL option) for TKBO, TKTopAlgo and TKV3d projects. But how to do that within the build scripts? Note that the Visual Studio projects get regenerated each time build-deps gets called and thus this option is reset.

ahladik:
I found a way to apply the workaround for the conflict between OCCT and Visual C++ in Release x86 compilation.
I finally managed to compile both x86 and x64 platforms in both Debug and Release configurations with Visual Studio 2019.

aothms:
Wow thanks for your continued effort and investigations.

Yes I think we should move to 7.5 it appears to fix quite a few issues that users have been reporting.

But at the same time it's really great to still be able to move back to older OCCT releases to check for regressions.

So what's the status now from your end, can it be merged?

ahladik:
You're welcome, I'm very interested in IfcOpenShell, it seems to me the only option to handle the complexity of IFC in a C++ application. Thanks to you for developing this project.

Yes, I think it is possible to merge. I hope that the workaround suggested by the OCCT support, i.e. setting the global optimization for the library, will not manifest incompatibilities when linked with code that was not compiled with the same option. We'll see. 

Personally, however, I'm struggling with a compile-time error in my code. If I simply include "ifcparse\IfcFile.h" in one of my sources, I get:

```
1>C:\dev\Dev_NEWIFC\SG\trunk24\IFCOS\IfcOpenShell-0.6.0\_deps\boost_1_74_0\boost\variant\variant.hpp(1390): error C2061: errore di sintassi: identificatore 'storage_'
1>C:\dev\Dev_NEWIFC\SG\trunk24\IFCOS\IfcOpenShell-0.6.0\_deps\boost_1_74_0\boost\variant\variant.hpp(1375): note: durante la compilazione della funzione membro 'boost::variant<boost::blank,IfcWrite::IfcWriteArgument::Derived,int,bool,double,std::string,boost::dynamic_bitset<unsigned long,std::allocator<DWORD>>,IfcWrite::IfcWriteArgument::EnumerationReference,IfcUtil::IfcBaseClass *,IfcWrite::IfcWriteArgument::empty_aggregate_t,std::vector<int,std::allocator<int>>,std::vector<double,std::allocator<double>>,std::vector<std::string,std::allocator<std::string>>,std::vector<boost::dynamic_bitset<unsigned long,std::allocator<DWORD>>,std::allocator<boost::dynamic_bitset<unsigned long,std::allocator<DWORD>>>>,IfcEntityList::ptr,IfcWrite::IfcWriteArgument::empty_aggregate_of_aggregate_t,std::vector<std::vector<int,std::allocator<int>>,std::allocator<std::vector<int,std::allocator<int>>>>,std::vector<std::vector<double,std::allocator<double>>,std::allocator<std::vector<double,std::allocator<double>>>>,IfcEntityListList::ptr>::variant(void) noexcept' di modello classe
1>C:\dev\Dev_NEWIFC\SG\trunk24\IFCOS\IfcOpenShell-0.6.0\src\ifcparse\IfcSpfHeader.h(62): note: vedere il riferimento all'istanza 'boost::variant<boost::blank,IfcWrite::IfcWriteArgument::Derived,int,bool,double,std::string,boost::dynamic_bitset<unsigned long,std::allocator<DWORD>>,IfcWrite::IfcWriteArgument::EnumerationReference,IfcUtil::IfcBaseClass *,IfcWrite::IfcWriteArgument::empty_aggregate_t,std::vector<int,std::allocator<int>>,std::vector<double,std::allocator<double>>,std::vector<std::string,std::allocator<std::string>>,std::vector<boost::dynamic_bitset<unsigned long,std::allocator<DWORD>>,std::allocator<boost::dynamic_bitset<unsigned long,std::allocator<DWORD>>>>,IfcEntityList::ptr,IfcWrite::IfcWriteArgument::empty_aggregate_of_aggregate_t,std::vector<std::vector<int,std::allocator<int>>,std::allocator<std::vector<int,std::allocator<int>>>>,std::vector<std::vector<double,std::allocator<double>>,std::allocator<std::vector<double,std::allocator<double>>>>,IfcEntityListList::ptr>::variant(void) noexcept' della funzione modello di cui è in corso la compilazione
1>C:\dev\Dev_NEWIFC\SG\trunk24\IFCOS\IfcOpenShell-0.6.0\src\ifcparse\IfcWrite.h(116): note: vedere il riferimento all'istanza 'boost::variant<boost::blank,IfcWrite::IfcWriteArgument::Derived,int,bool,double,std::string,boost::dynamic_bitset<unsigned long,std::allocator<DWORD>>,IfcWrite::IfcWriteArgument::EnumerationReference,IfcUtil::IfcBaseClass *,IfcWrite::IfcWriteArgument::empty_aggregate_t,std::vector<int,std::allocator<int>>,std::vector<double,std::allocator<double>>,std::vector<std::string,std::allocator<std::string>>,std::vector<boost::dynamic_bitset<unsigned long,std::allocator<DWORD>>,std::allocator<boost::dynamic_bitset<unsigned long,std::allocator<DWORD>>>>,IfcEntityList::ptr,IfcWrite::IfcWriteArgument::empty_aggregate_of_aggregate_t,std::vector<std::vector<int,std::allocator<int>>,std::allocator<std::vector<int,std::allocator<int>>>>,std::vector<std::vector<double,std::allocator<double>>,std::allocator<std::vector<double,std::allocator<double>>>>,IfcEntityListList::ptr>' di modello classe di cui è in corso la compilazione
```
This doesn't happen with `IfcParseExamples`,  which works normally, even after modifying it to handle different schemas.
Do you happen to have any idea why this happens?
In the `variant.hpp` file there is this note:
```
        // NOTE TO USER :
        // Compile error from here indicates that the first bound
        // type is not default-constructible, and so variant cannot
        // support its own default-construction.
        //
        new( storage_.address() ) internal_T0();
```

aothms:
> We'll see.

Exactly, regardless it's a massive improvement as before this PR we don't have msvc19 to begin with.

> Compile error from here indicates  ...

I would bet that's related to `internal_T0();` as that's the first type, and it's being placement new'ed into storage_. And not a Syntax error for storage (but I could be wrong).

The first type is `boost::blank`. I don't see it being anywhere being included, so maybe it comes implicitly with some other headers that are then not included in your minimal example. Maybe try `#include <boost/blank.hpp>` somewhere. If it works please submit a PR for that include on the .h file that defines the variant in IfcOpenShell.

PR #1171, opened 2020-12-14
bimtester: start translation into German and French with one step as well as the report

berndhahnebach:


berndhahnebach:
you are merging before I press the ok button for the PR. This is only possible if you are faster than the speed of light ... :-)

PR #1201, opened 2020-12-22
Selection representation context

Noranius:
For the translation of ifc files to others, the selection of the geometry context would be helpful. The idea is to select the geometry representation context(s) by their ID(s). Hence, I have added this option to the `IfcConverter`. 

In short, I have extended the `initialize` method `IfcGeom:Iterator` with an overload that it takes a set of strings as argument. During the initialization this set of geometry representation names is respected to select related representations.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -156,7 +156,7 @@ struct geom_filter
 {
     geom_filter(bool include, bool traverse) : type(UNUSED), include(include), traverse(traverse) {}
     geom_filter() : type(UNUSED), include(false), traverse(false) {}
-    enum filter_type { UNUSED, ENTITY_TYPE, LAYER_NAME, ENTITY_ARG };
+    enum filter_type { UNUSED, ENTITY_TYPE, LAYER_NAME, ENTITY_ARG};
     filter_type type;
     bool include;
     bool traverse;
@@ -169,6 +169,7 @@ struct inclusion_filter : public geom_filter { inclusion_filter() : geom_filter(
 struct inclusion_traverse_filter : public geom_filter { inclusion_traverse_filter() : geom_filter(true, true) {} };
 struct exclusion_filter : public geom_filter { exclusion_filter() : geom_filter(false, false) {} };
 struct exclusion_traverse_filter : public geom_filter { exclusion_traverse_filter() : geom_filter(false, true) {} };
+struct geometricRepresentationContext_filter : public geom_filter { geometricRepresentationContext_filter() : geom_filter(true, false) {} };
 
 size_t read_filters_from_file(const std::string&, inclusion_filter&, inclusion_traverse_filter&, exclusion_filter&, exclusion_traverse_filter&);
 void parse_filter(geom_filter &, const std::vector<std::string>&);
@@ -202,6 +203,7 @@ int main(int argc, char** argv) {
 	inclusion_traverse_filter include_traverse_filter;
 	exclusion_filter exclude_filter;
 	exclusion_traverse_filter exclude_traverse_filter;
+	geometricRepresentationContext_filter include_geometricRepresentationContext_filter;
 	path_t filter_filename;
 	path_t default_material_filename;
 	path_t log_file;
@@ -304,7 +306,7 @@ int main(int argc, char** argv) {
 			"2) 'layers': the instances that are assigned to presentation layers of which names "
 			"match the given values should be included.\n"
 			"3) 'attribute <AttributeName>': products whose value for <AttributeName> should be included\n. "
-			"Currently supported arguments are GlobalId, Name, Description, and Tag.\n\n"
+				"Currently supported arguments are GlobalId, Name, Description, and Tag.\n\n"
 			"The values for 'layers' and 'arg' are handled case-sensitively (wildcards supported)."
 			"--include and --exclude cannot be placed right before input file argument and "
 			"only single of each argument supported for now. See also --exclude.")
@@ -318,6 +320,9 @@ int main(int argc, char** argv) {
 		("exclude+", po::value<exclusion_traverse_filter>(&exclude_traverse_filter)->multitoken(),
 			"Same as --exclude but applies filtering also to the decomposition and/or containment "
 			"of the filtered entity. See --include+ for more details.")
+		("geometricRepresentationContext", po::value<geometricRepresentationContext_filter>(&include_geometricRepresentationContext_filter)->multitoken(),
+				"Select the IfcGeometricRepresentation(Sub)Context for the geometries, which shall be exported. "
+				"If empty, all contexts are used for export.")
 		("filter-file", new po::typed_value<path_t, char_t>(&filter_filename),
 			"Specifies a filter file that describes the used filtering criteria. Supported formats "
 			"are '--include=arg GlobalId ...' and 'include arg GlobalId ...'. Spaces and tabs can be used as delimiters."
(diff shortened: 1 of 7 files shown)


Moult:
Oh wow! I've been looking for a way to run create_shape or the iterator that _only_ targets a particular context + subcontext + target_view - does this solve that problem?

Noranius:
In case you mean to export a shape of an IFC file, which has a specific context, then you can try this. You can select multiple contexts. Just run `IfcConvert` with the -h parameter to get the help. There is a new option `geometricRepresentationContext`.

 I haven't worked with target views, yet. Maybe you can provide some information about that?

Moult:
I see in IfcConvert, I was hoping that this could also affect the Python bindings for create_shape or geometry iterator.

`TargetView` is an attribute of a representation _subcontext_. It has a similar purpose to the context name and the subcontext name. So, especially when creating drawings, it isn't enough to filter by context + subcontext (e.g. Model/Body or Plan/Annotation), it is also really important to filter by context + subcontext + target view (e.g. Plan/Annotation/PLAN_VIEW vs Plan/Annotation/SECTION_VIEW)

Noranius:
Thank for that information. I'm working at my PhD currently and such information is very helpful 🙂.

Regarding the python bindings: many changes are in classes, which are not specified to IfcConvert, maybe you check the changes and ask again if necessary.

aothms:
Hi @Noranius thanks for this. Sorry for the delays in replying. This is great. We kind of already had --plan and --model but it mixed various things and I think it's good to have more control over rep context.

Did you have a look at the logs on Travis:

https://travis-ci.org/github/IfcOpenShell/IfcOpenShell/builds/751025974

It appears that the PR introduced some issues on files:

```
0.12s$ /usr/bin/python2.7 tests.py

  File "tests.py", line 98, in <module>

    t = ifcopenshell.geom.tree(f, tree_settings)

  File "/usr/lib/python2.7/dist-packages/ifcopenshell/geom/main.py", line 134, in __init__

    ifcopenshell_wrapper.tree.__init__(*args)

  File "/usr/lib/python2.7/dist-packages/ifcopenshell/ifcopenshell_wrapper.py", line 556, in __init__

    this = _ifcopenshell_wrapper.new_tree(*args)

RuntimeError: Token $ at 2325 invalid string

The command "/usr/bin/python2.7 tests.py" exited with 1.

0.00s$ cd input

The command "cd input" exited with 0.

0.19s$ /usr/local/bin/IfcConvert -yv acad2010_walls.ifc acad2010_walls.glb

IfcOpenShell IfcConvert 0.6.0b0 (OCC 7.3.0)

Scanning file...

Done scanning file   

Parsing input file took 0 seconds

terminate called after throwing an instance of 'IfcParse::IfcInvalidTokenException'

  what():  Token $ at 2325 invalid string

/home/travis/.travis/functions: line 109:  9760 Aborted                 (core dumped) /usr/local/bin/IfcConvert -yv acad2010_walls.ifc acad2010_walls.glb
```


> RuntimeError: Token $ at 2325 invalid string

Token 2325 in acad_2010_walls.ifc is

~~~
#10=IFCGEOMETRICREPRESENTATIONCONTEXT($,'PLAN',3,1.E-005,#9,$);
                                      ^
~~~

And indeed IfcGeometricRepresentationContext.ContextIdentifier is optional, so you better first check with `hasContextIdentifier()`. Unfortunately we never really went with boost::optional<std::string> in the attribute signature. But I will make that change in v0.7

Also instead of `geometricRepresentationContext` I'd prefer the simpler and shorter `context` for the command line arg.

> I see in IfcConvert, I was hoping that this could also affect the Python bindings for create_shape or geometry iterator.

It would for iterator (more or less automatically I think, for that purpose maybe we should change the signature from `std::set<std::string>` to `std::vector<std::string>` less pure semantically, but the SWIG typemaps are nicesly setup only for vector). But not for create_shape() but that already has an option to specify the IfcRepresentation directly.

Noranius:
Hi Aothms,
I'll have a look on that. Unfortunately, I have several other things on my desk currently. I'll get to this If the other stuff is ready.

Greetings
Noran

Moult:
Bump on this PR :) I wonder if the new functionality in set_context_ids has any relevance on this?

Moult:
Bump :)

theoryshaw:
Hi @Noranius 

src/ifcgeom/IfcGeomIteratorImplementation.h
```diff
@@ -307,8 +313,14 @@ namespace IfcGeom {
 						if (allowed_context_types.find(context_type) == allowed_context_types.end()) {
 							Logger::Warning(std::string("ContextType '") + context->ContextType() + "' not allowed:", context);
 						}
+
 						if (context_types.find(context_type) != context_types.end()) {
-							filtered_contexts->push(context);
+							std::string context_name = context->ContextIdentifier();
```
CyrilWaechter, line 318: `ContextIdentifier` is optional (`ContextIdentifier : OPTIONAL IfcLabel`). Calling it directly throws `IfcInvalidTokenException` when the value is `$` — this is exactly the `Token $ at 2325 invalid string` crash @aothms reproduced on `acad2010_walls.ifc` (`IFCGEOMETRICREPRESENTATIONCONTEXT($,'PLAN',...)`). Guard with `context->hasContextIdentifier()` before dereferencing.

src/ifcgeom/IfcGeomIteratorImplementation.h
```diff
@@ -330,19 +342,32 @@ namespace IfcGeom {
 			for (it = filtered_contexts->begin(); it != filtered_contexts->end(); ++it) {
 				IfcSchema::IfcGeometricRepresentationContext* context = *it;
 
-				representations->push(context->RepresentationsInContext());
-				try {
-					if (context->hasPrecision() && context->Precision() < lowest_precision_encountered) {
-						lowest_precision_encountered = context->Precision();
-						any_precision_encountered = true;
+				std::string context_name = context->ContextIdentifier();
```
CyrilWaechter, line 345: Same unguarded optional access as line 318. This is also a duplicate of the filtering already done when building `filtered_contexts`; if the settings-based approach isn't adopted, consolidate the two filters.

src/ifcgeom/IfcGeomIteratorImplementation.h
```diff
@@ -330,19 +342,32 @@ namespace IfcGeom {
...
+					}
+					catch (const std::exception& e) {
+						Logger::Error(e);
 					}
-				} catch (const std::exception& e) {
-					Logger::Error(e);
 				}
-
+				
+				//Hier muss eingegriffen werden. Falls eine Liste mit erlaubten Kontexten �bergeben --> Filtern
```
CyrilWaechter, line 361: Please drop this leftover TODO. It is in German and contains a mangled non-UTF8 byte, and commented-out/aspirational notes don't belong in a submit-ready change.

src/ifcgeom/IfcGeomIteratorImplementation.h
```diff
@@ -330,19 +342,32 @@ namespace IfcGeom {
...
 				IfcSchema::IfcGeometricRepresentationSubContext::list::ptr sub_contexts = context->HasSubContexts();
 				for (jt = sub_contexts->begin(); jt != sub_contexts->end(); ++jt) {
-					representations->push((*jt)->RepresentationsInContext());
+					std::string sub_context_name = (*jt)->ContextIdentifier();
+					if (allowed_context_identifiers.empty()
+						|| allowed_context_identifiers.find(context_name) != allowed_context_identifiers.end()
```
CyrilWaechter, line 366: Selecting a *sub*context identifier alone cannot work: the parent context was already filtered out at line 318 unless its own (often null) `ContextIdentifier` is in the allow-set. The representation-level filter here is never reached in that case. Filter per representation, or always let a parent through when any of its subcontexts match.

src/ifcgeom_schema_agnostic/Kernel.cpp
```diff
@@ -225,6 +243,27 @@ std::map<std::string, IfcUtil::IfcBaseEntity*> IfcGeom::Kernel::get_layers(IfcUt
 	}
 }
 
+std::map<std::string, IfcUtil::IfcBaseEntity*> IfcGeom::Kernel::get_representationContexts(IfcUtil::IfcBaseEntity* inst) {
+	if (inst->as<Ifc2x3::IfcProduct>()) {
+		return get_layers_impl<Ifc2x3>(inst->as<Ifc2x3::IfcProduct>());
```
CyrilWaechter, line 248: Copy-paste bug: this dispatches to `get_layers_impl` instead of the newly added `get_representationContext_impl`. As written it returns layer assignments, not representation contexts. Note that `get_representationContexts` is never called anywhere, and `get_representationContext_impl` is likewise dead — both can be dropped rather than fixed.

src/ifcconvert/IfcConvert.cpp
```diff
@@ -169,6 +169,8 @@ struct inclusion_filter : public geom_filter { inclusion_filter() : geom_filter(
 struct inclusion_traverse_filter : public geom_filter { inclusion_traverse_filter() : geom_filter(true, true) {} };
 struct exclusion_filter : public geom_filter { exclusion_filter() : geom_filter(false, false) {} };
 struct exclusion_traverse_filter : public geom_filter { exclusion_traverse_filter() : geom_filter(false, true) {} };
+//struct geometricRepresentationContext_filter : public geom_filter { geometricRepresentationContext_filter() : geom_filter(true, false) {} };
+std::vector<std::string> include_geometricRepresentationContext_filter;
```
CyrilWaechter, line 173: Global mutable option state. This should be local to `main` (or carried through `settings`) rather than a file-scope `std::vector<std::string>`.

src/ifcconvert/IfcConvert.cpp
```diff
@@ -318,6 +334,10 @@ int main(int argc, char** argv) {
 		("exclude+", po::value<exclusion_traverse_filter>(&exclude_traverse_filter)->multitoken(),
 			"Same as --exclude but applies filtering also to the decomposition and/or containment "
 			"of the filtered entity. See --include+ for more details.")
+		("geometricRepresentationContext", po::value<std::vector<std::string>>(&include_geometricRepresentationContext_filter)->multitoken(),
```
CyrilWaechter, line 337: Naming per @aothms: prefer `context` over `geometricRepresentationContext` for the CLI arg. Also fix the help-text typo "constexts" on line 339.

CyrilWaechter (commented):
After reading I do not think this PR is relevant anymore. Feel free to reopen.
AI analysis just as an informative addition:
```
## Review: PR #1201 — Selection representation context

Thanks for the contribution. Static review summary below.

**Verdict:** comment · **Relevance:** superseded

This PR targets the long-dead `v0.7.0` branch, is a draft and `CONFLICTING`. More importantly, representation-context selection has since landed via `settings::ContextIds` (already on the v0.8 line, `ConversionSettings.h`) and `settings::ContextPriorities` (v0.9.0, `conversion_settings.h:231`), with `mapping.cpp::addRepresentationsFromContextIds` and Python/IfcConvert exposure (`context-ids`, `context-priorities`). @aothms explicitly redirected this PR toward that settings-based design in the discussion and later noted the overlap. A rebase of this implementation is not the right path.

Outstanding maintainer concerns (from @aothms) that remain unaddressed:
- Optional `IfcGeometricRepresentationContext.ContextIdentifier` must be guarded with `hasContextIdentifier()`; direct access throws `IfcInvalidTokenException` (`Token $ ... invalid string`) — this is reproduced on `acad2010_walls.ifc`.
- Preferred command-line arg name is `context`, not `geometricRepresentationContext`.
- Signature should use `std::vector<std::string>` (SWIG typemaps) rather than `std::set<std::string>` if this were pursued.

Blocking issues:
1. Superseded by `settings::ContextIds` / `ContextPriorities` (commit 514c5eb and follow-ups).
2. Crash on the optional `ContextIdentifier` attribute (unguarded access).
3. `Kernel::get_representationContexts` calls `get_layers_impl` and is never called; the new `get_representationContext_impl` is dead too.
4. Subcontext selection is inconsistent: parent contexts are dropped by identifier before subcontexts are considered.
5. Unrelated scope: `win/readme.md` plus formatting/whitespace churn.
6. German TODO (non-UTF8 byte) and commented-out code.

No tests were added for the new option.
```

PR #1221, opened 2020-12-31
Rework python pset module

CyrilWaechter:
New features
============
* Handle all syntax for ApplicableEntity as defined in IFC documentation. (drawback it increase request time by ~2/3)
* Allow to get applicable psets directly from template definition without caching.
This allow more general purpose usages. (drawback it increase request time by ~2.4)

To notice / discuss
===================
* `ifcwrap/CmakeLists` was installing only `.py` and `.bnf` files from `ifcopenshell-python`.
Why ? It now copy all files. Is it an issue ?
* property template file is now stored in ifcopenshell python module to allow usage
from every softwares. It was before only available for blenderbim addon.
* psets and qtos are now optionnally cached at object instead of module level.
This allow to work with multiple schema in the same python interpreter.
(eg. might help conversion between schemas)
* in blenderbim addon psetqto are now stored in `schema.psetqto`.


CyrilWaechter:
I wanted to add some pytest but did not now where to place them. test folder do not contain many test and do not seem to correspond to test per module. Where to place test with naming convention and structure ?

Moult:
Everything looks good to me! However, I'll let @aothms approve the PR because it touches this file, and I'm not quite sure the impact of it:

```
src/ifcwrap/CMakeLists.txt 
```

aothms:
Sure, looks good to me. Wondering though if the `@functools.lru_cache` decorator couldn't have been used to implement caching using less lines of code?

aothms (approved):

aothms:
> I wanted to add some pytest but did not now where to place them. test folder do not contain many test and do not seem to correspond to test per module. Where to place test with naming convention and structure ?

Any proposal will work probably as it beats the current approach of one monolithic script (not really well structured isolated test cases).

CyrilWaechter:
> Sure, looks good to me. Wondering though if the `@functools.lru_cache` decorator couldn't have been used to implement caching using less lines of code?

@aothms You were absolutely right. It not only simplify code but also make request incredibly faster as many entities have common supertypes. It can handle 10+ times more requests in 4 less time. I will prepare another pull request to benefit from this improvement. Thank a lot for this suggestion.

PR #1225, opened 2021-01-03
Use lru_cache for pset module

CyrilWaechter:
New features
===========
* `get_by_name` : get pset template by name
* `is_templated` : allow to check if a pset template correspond to given name

Enhancement
============
* Use of lru_cache simplify code and enhance performance largely
(10+ times more requests in 4 less time)
* Take advantages of new caching system and new features in blenderbim


CyrilWaechter:
Forgot to mention https://github.com/IfcOpenShell/IfcOpenShell/pull/1221#issuecomment-752902996 in commit message.

PR #1259, opened 2021-01-14
Refactoring of BIMTester Module

ihabelaghoury:
Please check for merging


src/ifcblenderexport/blenderbim/bim/module/bimtester/ui.py
```diff
@@ -0,0 +1,61 @@
...
+        if bimtester_properties.features_dir:
+            row = layout.row(align=True)
+            row.prop(bimtester_properties, "features_file")
+
+            row = layout.row(align=True)
+            row.prop(bimtester_properties, "scenario")
+        else:
+            return
+
+        if context.scene.BimTesterProperties.features_file != '': # To handle the error when no .feature file exists in the folder
```
ihabelaghoury, line 35: Here, I need to access the STRING value of the EnumProperty, knowing that if nothing is selected it might return a 0 ...
What should I do in this case as I do not know all the EnumProperty Methods

Moult: Code looks good. There actually are no enumproperty methods. You only get access to the value of the currently selected enums.

PR #1268, opened 2021-01-18
[read carefully before decide merge] !!! normalize line endings of src files

berndhahnebach:
@aothms @Moult 

git has various possibilities to handle line endings. In FreeCAD I have found simplest way to solve it, is to normalize all line endings of source code files in the git repo to LF (unix). If ever someone pushes a file (with allowed file association) with CRLF it will be converted automatically. 

With this we will never ever end up with mixed line endings in one file, which is a nightmare because editors out there handles mixed line endings differently and thus this results in line ending commits. 

See


aothms:
I'm fine with this. @Moult do you see any issues? Better merge this asap as it's likely to trigger conflicts.

Moult:
All good! Nice to see which files I created on my work computer! :)

berndhahnebach:
:+1:

Issue #1258, opened 2021-01-13
Segfault when creating geometry


FreakTheMighty:
While testing `IfcConvert` from this build https://github.com/IfcOpenBot/IfcOpenShell/commit/e1b68dafa469e1f7549f4df500266075db2030a0#commitcomment-45887584 I'm seeing segfaults for various files. I haven't been able to isolate the problematic object so I'm going to email you an entire file that I'd rather not share publicly.

Moult:
@FreakTheMighty by the way, I've found using the BlenderBIM Add-on quite useful for isolating problematic objects. There is a feature which supports whitelisting, elements, and sometimes I print the name of the object being generated. There is also an "IFC Debug" panel that lets you dig into the shape representation tree, and one by one run `create_shape` on each subitem of the tree - especially useful for finding highly nested boolean operations where something causes a segfault.

If you'd like I can give you a little demo on it :)

aothms:
Recent versions of IfcConvert also have the -vv flag. That prints info on
the element currently being processed.

Op wo 13 jan. 2021 22:15 schreef Dion Moult <notifications@github.com>:

FreakTheMighty:
Thanks @Moult @aothms those are really helpful tips. 

I can confirm that `32uecXxLHAvBNzQKHB18uk` is the offending product. When I run with `-vv` this is the last log. 
```
[Notice] {32uecXxLHAvBNzQKHB18uk} Adjusted edge end-point with distance 0.001587499999999872 on:
#438924=IfcCompositeCurve((#438722,#438729,#438736,#438744,#438751,#438759,#438766,#438774,#438781,#438789,#438796,#438804,#438811,#438819,#438826,#438833,#438840,#438848,#438855,#438863,#438870,#438878,#438885,#438893,#438900,#438908,#438915,#438923),.F.)
```
Also, in a simple python script calling `create_shape` on that product, I get the segfault. If, however, I try to extract that one element (I use BIMVision + plugin) the extracted example does not segfault which means that large file I emailed is still the smallest reproducible sample I have.

Moult:
Nice! Didn't know about -vv, that's super useful!

[...]

PR #1270, opened 2021-01-20
Adding openings from extracted elements

FreakTheMighty:
Copies openings when extracting elements.

Closes #1270


Moult:
Merging, but I _think_ you only need the one add(opening) not add(opening.RelatedOpeningElement) (since the first one should detect the second). But it doesn't hurt :)

FreakTheMighty:
@Moult I tried without the second add and I wasn't reproducing the bug in #1258, for some reason including both elements caused the segfault. . . yay 👏

Moult:
Ah well :) Let's move on! :D

Issue #816, opened 2020-03-27
SweptSolid missing faces


FreakTheMighty:
Apologies for [double posting](https://sourceforge.net/p/ifcopenshell/discussion/1782717/thread/7a7839a1ba) from the forum, but I think it makes more sense to track this as an issue as I've found some existing tickets that seem related.

---

## Overview
I'm finding that an IfcBeam is missing faces. It should be a U shape, but `create_shape` is generating an L instead. I've tested with `IfcConvert` 0.6.0 as well as through some custom python code.

`IfcOpenShell IfcConvert 0.6.0b0 (OCC 7.3.0)`

## Sample Images
Here's what it looks like with the missing faces:
![Missing Faces](https://i.ibb.co/6YT6cnb/Annotation-2020-03-26-085454-tif.jpg)

And this was generated using assimp and contains the full geometry.
![Correct Model](https://i.ibb.co/3rGnjrq/Capture.png)

## Logs
```
[Notice] exclude entities IfcOpeningElement IfcSpace
[Warning] {00i6wSobrD$B7A4a3x0PB8} Self-intersections with 2 cycles detected
[Warning] {00i6wSobrD$B7A4a3x0PBA} Self-intersections with 2 cycles detected
[Warning] {00i6wSobrD$B7A4a3x0Qri} Self-intersections with 2 cycles detected
[Warning] {00i6wSobrD$B7A4a3x0Qrk} Self-intersections with 2 cycles detected
[Warning] {00i6wSobrD$B7A4a3x0Qs3} Self-intersections with 2 cycles detected
[Warning] {00i6wSobrD$B7A4a3x0Qs5} Self-intersections with 2 cycles detected

```
## Sample IFC
[small-framing.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/4394406/small-framing.zip)

FreakTheMighty:
I'm not sure if these are related, but there seems to have been recent issues related to `wire_intersections`. In this issue, nothing crashes, but it does seem like intersection detection may not be handling things in an ideal way.

Two things come to mind.

1. When it detects an intersection, it grabs the "largest" section. Is there a way it could process both section?
1. Is it possible that these elements are so thin that it is detecting an intersection where there it shouldn't?

https://github.com/IfcOpenShell/IfcOpenShell/issues/768

FreakTheMighty:
[...]

PR #1278, opened 2021-01-27
Remove wire intersection check to preserve faces

FreakTheMighty:
Relates to #816 

Before
image

After
image


aothms:
Hi @FreakTheMighty, frankly I don't know whether it's still necessary, but let's find out :)


The issue is not so much the  wire interesection check in principal, that's fine and useful, the issue is more the tolerance extension. 

E.g input in black will get you the red output after wire intersection check (assuming the black dashed line has an extent less then the tolerance in use at that point)

afbeelding

And also this is not so much the issue, that's probably also necessary.

The issue is that the tolerance might increase after a boolean operation, so what's valid at the time of the wire intersection check, might later become invalid after the boolean operation (and potentially hard crashes resulting from it).

For that reason there is a markup on the tolerance value https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom/IfcGeomFunctions.cpp#L3839

In another conversion we also talked about these tolerances, so that's also here a way to control ifcopenshell's functioning, if you patch the file to have a lower tolerance, your issue would likely be gone for this file as well.

But in general IfcOpenShell already calculates with a more conservative precision than what's in the model https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom/IfcGeomIteratorImplementation.h#L353

And maybe that's what we need to re-evaluate. This is from 6 years ago

afbeelding

That's roughly around the release of OCCT 6.9 which greatly increases OCCT boolean ops using fuzziness. https://dev.opencascade.org/index.php?q=node/1056

And also before we introduced the "adaptive tolerance handling" in IfcOpenShell (!) https://github.com/IfcOpenShell/IfcOpenShell/commit/8ce4bbdf6194649cbaca4dff2ec1f53342cf5d39

--

These kind of changes obviously affect the processing of all files, so at this point we can only merge these kind of invasive things when they are behind a flag that's off by default. I would suggest the following flags:

`--strict-tolerance`: removes the  * 10 markup https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom/IfcGeomIteratorImplementation.h#L353

`--no-wire-intersection-tolerance`: sets `eps` to zero https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom/IfcGeomFunctions.cpp#L3839

`--no-wire-intersection-check` your PR but behind a flag.

I don't want to dump this all on your shoulders, if you don't get to it at some point I will :) but any help is appreciated :)

FreakTheMighty:
@aothms thank you for all the tips here. I'll dig more. I'm trying to sort through some build issues that are preventing me from testing, but hopefully once those are sorted I can figure out if there's something I can do that's worth contributing back such as the flags you mentioned.

One silly question, when you say "markup", what are you referring to?

FreakTheMighty:
@aothms :P I understand markup now - my brain went to markup language rather than simply some factor. So yeah, you're right simply removing that factor fixes things in my case.

I'm down for adding this behind a flag, but I'm curious, how do you envision such a flag being passed to this function. The `IteratorSettings` object is not yet passed this deep into the call stack. Is there a less disruptive way to control this? Here's the first frame in the call stack where settings is still available.

image

aothms:
Yes good observation this is indeed the pain. The `Kernel` class only has this `setValue(enum, float)`, so there's duplicate work to implement it there as well. 

There's a couple of them now that are propagated like that. It's ugly. https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom/IfcGeomIteratorImplementation.h#L1001

FreakTheMighty:
🤷 I don't think it seems that messy. Now I see that all of these methods are part of the geometry kernel which as various options. I've started to try adding one and will push an update when I've added `strict-tolerance`

FreakTheMighty:
@aothms I've made updates to put my pull request behind a flag, does this look about right? Also, I haven't dug much, but what do I have to do to expose this flag to the python wrapper?

aothms:
> does this look about right?

Looks good to me.

> Also, I haven't dug much, but what do I have to do to expose this flag to the python wrapper?

Nothing. IteratorSettings is ifcopenshell.geom.settings, so the flag should be there once rebuilt. You don't have to use this ifcopenshell.geom.settings.set anymore, you can supply them directly as keyword args. ifcopenshell.geom.settings(NO_WIRE_INTERSECTION_CHECK=True)

FreakTheMighty:
Great! Should I add the flag to this `convert_wire_to_face` method as well? https://github.com/IfcOpenShell/IfcOpenShell/blob/8681610b79265309d8c02d5a1ff2a1e6ed227c7f/src/ifcgeom/IfcGeomFunctions.cpp#L1103

Also, would you be willing to accept this pull request without the other options? We could add you other suggestions as necessary.

aothms:
But isn't to remove the 10 * tolerance a better fix for your issue? If you run this on files with actual self intersecting faces with booleans you will likely get crashes.

aothms:
PS: If you want you can cover all cases by doing the check within wire_intersections() itself. Having it in the parent call site like you did is clearer in intention.

FreakTheMighty:
I'll have to take your word on that :) Ok, let me work on that next flag.

> PS: If you want you can cover all cases by doing the check within wire_intersections() itself. Having it in the parent call site like you did is clearer in intention.

I'd leave that call up to you, what would you prefer?

aothms:
Well having it in the function is at least a guarantee the flag is always respected, I guess that would be my preference.

FreakTheMighty:
@aothms added `--strict-tolerance` option, let me know what you think of the implementation. The change in precision has improved quality in other ways. Removing the tolerance factor not only fixes the missing face, it also prevents some of these vertexes from collapsing which is what the original IFC contains.  

image
image

src/ifcconvert/IfcConvert.cpp
```diff
@@ -339,7 +339,9 @@ int main(int argc, char** argv) {
         ("default-material-file", new po::typed_value<path_t, char_t>(&default_material_filename),
             "Specifies a material file that describes the material object types will have"
             "if an object does not have any specified material in the IFC file.")
-		("validate", "Checks whether geometrical output conforms to the included explicit quantities.");
+		("validate", "Checks whether geometrical output conforms to the included explicit quantities.")
+		("no-wire-intersection-check", "Skip wire intersection check")
+		("strict-tolerance", "Use strict tolerance for detecting wire intersections");
```
FreakTheMighty, line 344: @aothms I'm not sure I'm capturing what this does accurately. It seems like this precision change affects more than just wire intersections . . . do you have a better description?

aothms:
> do you have a better description

Yes it's more general.
https://standards.buildingsmart.org/IFC/RELEASE/IFC2x2/FINAL/HTML/ifcrepresentationresource/lexical/ifcgeometricrepresentationcontext.html
describes it as the "the tolerance under which two given points are still
assumed to be identical".

I'd say something like: "Use exact tolerance from model. Default is a 10
times increase for more permissive edge curves and less artefacts after
boolean operations, at the expense of a potential loss of geometric detail
due to vertex collapsing and wire intersection fuzziness." But it's a bit
long... if you can phrase it a bit more succinct feel free.

Op za 6 feb. 2021 om 17:31 schreef Jesse Vander Does <
notifications@github.com>:

FreakTheMighty:
@aothms I think this is ready for final review. I've added all three flags that you recommended. Let me know if there's anything you'd like changed.

aothms:
I think it's perfect. Many thanks! Which flags did you end up enabling by default?

FreakTheMighty:
@aothms so far `--strict-tolerance` seems like the best results, but we're still working through some build issues so that we can test this more broadly.

PR #1285, opened 2021-02-01
Some more fixes to the Windows build procedure

ahladik:
- fixed an error in run_cmake.bat that selected the wrong OCCT libraries path if more than one BuildDepsCacheXXX.txt file was present
- changed name of variable BUILDDEPTHCACHE to a proper BUILD_DEPS_CACHE_PATH


ahladik:
I don't understand why there are conflicts, don't know how to fix them in github and I'm too old for all this git stuff. grrrrr

aothms:
Thanks @ahladik


Btw, I wanted to talk to you about the previous PR. The build output directories where changed

afbeelding

You added an _ underscore at the beginning and changed the postfix in `GEN_SHORTHAND` from `x86` to `Win32` (which is fine I guess as Win32 is what the user tends to see in MSVC).

It's just that the automated build script needs to be updated and the docs still refer to the old path:

https://github.com/IfcOpenShell/IfcOpenShell/#using-visual-studio


So before I make those changes I wanted to ask: This changes was intentational? And what's the reason for the underscore?

ahladik:
> So before I make those changes I wanted to ask: This changes was intentational? And what's the reason for the underscore?

Yes, that was intentional. I needed to easily distinguish which folders were produced by the build process and did not belong to the original distribution.

This makes it easier to delete them when necessary and also makes it easy to exclude them from a commit to a repository.

aothms:
Ok, thanks, I've also prefixed the install dir then with _ https://github.com/IfcOpenShell/IfcOpenShell/commit/ef5d672d4df869cc95c6208413a7214162b77d47

ahladik:
Great, thanks!

Issue #1298, opened 2021-02-06
Build (cmake) should check whether libxml2 dev support is installed/detected


plaes:
Seems like libxml2 existence is not properly checked in the cmake phase.

Moult:
This looks fixed, can it be closed?

PR #1299, opened 2021-02-06
build: Check for libxml2

plaes:
Fixes issue #1298

Signed-off-by: Priit Laes <plaes@plaes.org>

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -175,6 +175,7 @@ if(USE_MMAP)
 endif()
 
 if (IFCXML_SUPPORT)
+    find_package(LibXml2 REQUIRED)
     add_definitions(-DWITH_IFCXML)
 endif()
 


aothms:
Thanks, perhaps the build scripts (in `./nix` `/.win`) need to be updated to instruct FindLibXml2 to the correct dir? Did you use them or call cmake directly?

plaes:
> Thanks, perhaps the build scripts (in `./nix` `/.win`) need to be updated to instruct FindLibXml2 to the correct dir? Did you use them or call cmake directly?

I just followed README and called cmake directly (`mkdir -p build && cd build && cmake ../cmake`). Another "gripe" with current build setup is the need to pass various OCCT/OCE/OpenCollada paths instead of autodetecting them (like FreeCAD for example).

aothms:
Sorry for the delay. It took me a while to check with the win and nix build scripts, but it works nicely, thanks again.

> OCCT/OCE/OpenCollada paths instead of autodetecting them

I think this has worked at some point, e.g see the hardcoded oce here https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/cmake/CMakeLists.txt#L227 but probably they no longer are sane defaults. Contributions very welcome.

Issue #1186, opened 2020-12-19
IFC Rail and BIMTester


Moult:
Pinging @aothms and @rbertucat !

Firstly, this is great! It's excellent that BIMTester can help in the implementation testing for IFC Rail! I've merged your PR, and I'm liking the improvements. A couple of things I thought I'd create an issue on.

 1. The PR deletes the submodule mvd and input. I've reverted those two files so hopefully it'll be fixed (@aothms can confirm this). In future PRs maybe double check that changes are local to the codebase.
 2. Have you seen https://wiki.osarch.org/index.php?title=Using_MicroMVDs_for_exchange_requirements ? In general, each set of definitions in BIMTester correlate to a MicroMVD (see the list of MicroMVDs). Would you be interested in developing MicroMVD usecases of exchanging IFC Rail related information? If so, this is a task that requires continual refinement, so maybe it'd be best if you got direct commit access?
 3. Have you seen the cool work @berndhahnebach has been doing in creating a standalone GUI, increasing portability of BIMTester, as well as translation into French and German (and there is a volunteer available for Dutch too)? Maybe there is room for collaboration?

Just a few ideas :)

berndhahnebach:
I have problems with this new code ... this does not work for me on latest ifcopenshell 

https://github.com/IfcOpenShell/IfcOpenShell/blob/05d835410daa23c2116d691db3d277e61bfea5e6/src/ifcbimtester/bimtester/features/steps/utils.py#L2

```
$ python
Python 2.7.16 (default, Oct 10 2019, 22:02:15) 
[GCC 8.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import ifcopenshell.express
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named ifcopenshell.express
```

aothms:
You need a fairly recent version of ifcopenshell. Recently the parser was
moved into the module to help the rail team quickly evaluate the RCs. Is
that the reason?

Op za 19 dec. 2020 14:04 schreef Bernd Hahnebach <notifications@github.com>:

berndhahnebach:
I pulled latest source before compile and install ifcopenshell.

berndhahnebach:
https://github.com/IfcOpenShell/IfcOpenShell/blob/05d835410daa23c2116d691db3d277e61bfea5e6/src/ifcopenshell-python/ifcopenshell/express/__init__.py#L11

[...]

PR #1309, opened 2021-02-09
BuildingSmart IFCRail changes to BIMTester

rbertucat:
All,
After Dion's changes (https://github.com/IfcOpenShell/IfcOpenShell/issues/1186#issuecomment-772337199), I've ported our step implementations here.

I've now installed black and run by default on Save as Dion mentionned:
> It was agreed upon a while ago to run "black" on all Python code for standardisation


Moult:
Looks awesome! Thanks so much!

PR #1334, opened 2021-02-20
Hdf5

johltn:


cmake/CMakeLists.txt
```diff
@@ -377,6 +384,40 @@ IF(COLLADA_SUPPORT AND BUILD_CONVERT)
...
+        # An additional lib- prefix is added by HDF5 to denote static libraries
+        SET(HDF5_LIBRARIES 
+            "${HDF5_LIBRARY_DIR}/libhdf5_cpp.lib"
+            "${HDF5_LIBRARY_DIR}/libhdf5.lib"
+            "${HDF5_LIBRARY_DIR}/libzlib.lib"
+			"${HDF5_LIBRARY_DIR}/libsz.lib"
+			"${HDF5_LIBRARY_DIR}/libaec.lib")
+    ELSE()
+        SET(HDF5_LIBRARIES 
+            "${HDF5_LIBRARY_DIR}/hdf5_cpp.lib"
```
aothms, line 412: The extension for library files on non-windows is not .lib (it's .a for static, .so for dynamic, .dylib for dynamic on OSX). For now maybe just change to .a

src/ifcconvert/IfcConvert.cpp
```diff
@@ -624,8 +626,15 @@ int main(int argc, char** argv) {
 	);
 
     path_t output_temp_filename = output_filename + IfcUtil::path::from_utf8(TEMP_FILE_EXTENSION);
+	
+	std::vector<std::wstring> tokens;
```
aothms, line 630: You cannot just use wstring, but use path_t which correctly toggles between string and wstring depending on platform and settings https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcconvert/IfcConvert.cpp#L69

src/ifcconvert/IfcConvert.cpp
```diff
@@ -794,7 +804,14 @@ int main(int argc, char** argv) {
 	} else if (output_extension == SVG) {
 		settings.set(IfcGeom::IteratorSettings::DISABLE_TRIANGULATION, true);
 		serializer = boost::make_shared<SvgSerializer>(IfcUtil::path::to_utf8(output_temp_filename), settings);
-	} else {
+	}
+	else if (output_extension == HDF) {
+		const path_t mtl_filename = change_extension(output_filename, MTL);
```
aothms, line 809: mtl_filename isn't needed I guess

src/serializers/HdfSerializer.cpp
```diff
@@ -0,0 +1,209 @@
...
+		}
+		positionsDataset.write(double_data_container.data(), mtype2);
+		double_data_container.clear();
+
+
+		for (std::vector<real_t>::const_iterator it = mesh.normals().begin(); it != mesh.normals().end(); ) {
+			const real_t x = *(it++);
+			const real_t y = *(it++);
+			const real_t z = *(it++);
+			int_data_container.push_back((int)x);
```
aothms, line 185: why do you convert the normals to int?

src/serializers/HdfSerializer.cpp
```diff
@@ -0,0 +1,209 @@
...
+			} else {
+				double_data_container.push_back(x);
+				double_data_container.push_back(y);
+				double_data_container.push_back(z);
+
+			}
+
+		
+		}
+		positionsDataset.write(double_data_container.data(), mtype2);
```
aothms, line 177: In theory you cannot do this. You cannot feed HDF5 a vector<x>. While you constructed a compound datatype that mirrors `s1_t`. If you want to do this it needs to be a `std::vector<s1_t>`. Because you do not have the guarantee that the compiler chooses to layout those things equivalently in memory. Coincidently there is a good chance it will, because a float is rather big, but it could well be that a compiler adds some padding after `s1_t.c` for ex.

My recommendation is that you make it (vcount, 3) dimension dataset with just simple floats. Same for indices.

aothms: That way all of the struct/compound logic can be removed (a shame I know, but we can save the complex datatypes for the voxels probably :))

src/serializers/HdfSerializer.cpp
```diff
@@ -0,0 +1,209 @@
...
+		dimsfaces[1] = 1;
+		H5::DataSpace occt_dataspace(RANK_OCCT, dimsocct);
+		OCCTDataset = OCCTGroup.createDataSet(DATASET_NAME_OCCT, str_type, occt_dataspace);
+		OCCTDataset.write(brep_data, str_type);
+
+		indicesDataset = meshGroup.createDataSet(DATASET_NAME_INDICES, mtype1, face_dataspace);
+		positionsDataset = meshGroup.createDataSet(DATASET_NAME_POSITIONS, mtype2, dataspace);
+		normalsDataset = meshGroup.createDataSet(DATASET_NAME_NORMALS, mtype3, dataspace);
+
+		for (std::vector<real_t>::const_iterator it = mesh.verts().begin(); it != mesh.verts().end(); ) {
```
aothms, line 158: In this for loop you're basically just copying mesh.verts() into double_data_container. You can construct the dataset directly from mesh.verts()

aothms: Well, except for isyup bit, but that can be ignored as far as I'm concerned

src/serializers/HdfSerializer.cpp
```diff
@@ -0,0 +1,127 @@
...
+	H5::Group meshGroup;
+	H5::DataSet positionsDataset;
+	H5::DataSet normalsDataset;
+	H5::DataSet indicesDataset;
+
+	H5::Group OCCTGroup;
+	H5::DataSet OCCTDataset;
+	
+	const IfcGeom::Representation::BRep& brepmesh = o->geometry();
+	const IfcGeom::Representation::Serialization *serialization = new IfcGeom::Representation::Serialization(brepmesh);
```
aothms, line 72: Ah one more thing, you create a new pointer (heap allocation) but you don't free (delete) the result.

In this particular case it'd be sufficient to do `const IfcGeom::Representation::Serialization serialization(brepmesh)`, meaning the element will be on the stack and automatically freed when it goes out of scope. Same for the triangulation element below.

aothms:
Thanks @johltn !!

PR #1335, opened 2021-02-21
Update occ_utils.py

### LINKED ISSUE #1326: Python 2 doenst have the ModuleNotFoundError, which cause the program to crash with a NameError, using ImportError works
https://github.com/IfcOpenShell/IfcOpenShell/blob/a0f79cc0db2404a77d1ff794cdd3b87fe822aa3c/src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py#L32

RobinVanTol:
fixes #1326


aothms:
Indeed ModuleNotFound is apparently introduced later and is a subclass of ImportError, so this looks good to me.

PR #1307, opened 2021-02-09
add github action support 

prudhomm:
this PR implements #1306

PR #1354, opened 2021-03-04
Automated Blenderbim releases using github actions ci

Krande:
A suggestion for a Github actions workflow to automatically produce a new blenderbim addon release upon push to repo.

It will trigger the build process only when it detects a change in the `src/ifcblenderexport/blenderbim/*` subfolder or in the .yml file describing the workflow itself `.github/workflows/ci-blenderbim-matrix.yml`.

I have tested this on my local repo and created a test release [here](https://github.com/Krande/IfcOpenShell/releases/tag/blenderbim-210304)

image

@Moult Last time I showed you the release sizes were too big. Does the release sizes shown here make more sense?

blender28-bim-210304-linux.zip 160 MB
blender28-bim-210304-macos.zip 144 MB
blender28-bim-210304-win.zip 81.7 MB

Anyways, take a look and let me know if there are any changes\improvements on the suggested workflow?

Best Regards
Kristoffer

### DIFF
--- a/.github/workflows/ci-blenderbim-matrix.yml
+++ b/.github/workflows/ci-blenderbim-matrix.yml
@@ -10,7 +10,6 @@ env:
   major: 0
   minor: 0
   name: blenderbim
-  pyver: py37
 
 jobs:
   activate:
@@ -23,11 +22,12 @@ jobs:
       run: echo ok go
 
   build:
-    name: ${{ matrix.config.name }}
+    name: ${{ matrix.config.name }}-${{ matrix.pyver }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
+        pyver: [py37, py39]
         config:
           - {
             name: "Windows Build",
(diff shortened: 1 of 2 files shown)


Moult:
Ping @aothms in case you have any issues with this - if not, more than happy to merge!

Moult:
In particular, I believe this may need a manual merge with #1307, which probably touches the same files. Probably not a big deal, but would prefer to have a second pair of eyes before I merge :)

Krande:
Sounds like a good idea. How about we wait and merge this after the other pull request is complete?

Moult:
I'm OK with either approach. Ping @aothms

aothms:
I don't think there is a conflict. I just merged the other PR.

Krande:
Okay, I will merge my branch with the latest changes done in v0.6 upstream and see if something new comes up before we consider merging.

Krande:
So I got an error in the Makefile 

```
# Required by BIMTester
mkdir dist/working
mkdir: cannot create directory ‘dist/working’: File exists
Makefile:48: recipe for target 'dist' failed
make: *** [dist] Error 1
Error: Process completed with exit code 2.
```

Which is related to

https://github.com/IfcOpenShell/IfcOpenShell/blob/059c405652aacf69c596a76c449c9d24939f7143/src/ifcblenderexport/Makefile#L233-L243

it seems like the files related to downloading and unzipping "python_fcl" is never removed before running another `mkdir` for the BIMTester build.

So I made a hack and removed the "dist/working" dir prior to `mkdir`

```python
cd dist/working && unzip python_fcl*
cp -r dist/working/fcl dist/blenderbim/libs/site/packages/

# Cleanup after python_fcl
rm -rf dist/working

# Required by BIMTester
mkdir dist/working
```

Which seems to resolved it for the Windows and Linux builds. However for macOS I got the following error:

```
cp -r dist/working/lark-parser-0.8.5/lark dist/blenderbim/libs/site/packages/
rm -rf dist/working
# Required by IFCClash
mkdir dist/working
cd dist/working && unzip python_fcl*
unzip:  cannot find or open python_fcl*, python_fcl*.zip or python_fcl*.ZIP.

No zipfiles found.
make: *** [dist] Error 9
Makefile:48: recipe for target 'dist' failed
Error: Process completed with exit code 2.
```

As it seems there is no `python_fcl`  for macOS? 

Anyways. At least 4 out of 6 releases are successful (py37 and py39 for windows and linux). But the issue I mentioned above regarding macOS and `python_fcl` remains.

For now I included a trigger-flag so that you need to include the string "[release]" in your commit message.

Let me know if this is okay to merge?

Let me know if this is accep

Moult:
Yeah there isn't a python fcl for macOS. This has been a problem since the past release: https://github.com/BerkeleyAutomation/python-fcl/issues/17 - if you have to know how on how to make automated builds of python-fcl cross-platform as well as Python 3.7 + 3.9, that would be super cool.

I recently changed some code in the makefile around fcl because i was going to update it for 3.9 but then I ended up just breaking the whole build as you've discovered :) I'll unbreak it.

Jesusbill:
@Krande this is really great!
Any thoughts why it did not build new versions during the last two days?

Krande:
Currently a new release is built only when the commit involves a change in the blenderbim directory and the commit message contains the string "[release]".


The idea was to reduce the number of builds to a minimum. Maybe I could add a nightly build release as well? Or another option is to create a fork and just push a commit with the aforementioned prerequisites to trigger a build on your fork? (you might have to change the repo organisation from IfcOpenShell to your name though).

Moult:
I think nightly is a good pace.

Jesusbill:
Thanks @Krande I agree with @Moult that nightly builds are good.

Krande:
For the moment we can see how the latest change by @Moult with testing a new release on every commit to the blenderbim repo. 

Maybe that is the best course of action. It will still just build 1 release per day. That means for each new commit within the same day the release assets will be replaced. The build process does not take long so let's just see how it works:)

Moult:
@Krande did I mess something up? I assumed my change would be on every commit, but it doesn't seem to be the case: https://github.com/IfcOpenShell/IfcOpenShell/releases

Krande:
That might be a mistake I did. Try changing `src/ifcblenderexport/*` to `src/ifcblenderexport/**`. That would include all subdirectories (if I read [this] (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-file-paths) correctly.)

PR #1359, opened 2021-03-06
Opencascade cmake improvements

plaes:
These changes should hopefully eliminate the need to specify `OCC_{HEADER,LIBRARY}_DIR` on distros where OpenCascade development packages are installed via package management.
Also some whitespace changes (tabs -> spaces) in the OpenCascade section. ;)

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -220,48 +220,57 @@ endfunction()
 if(BUILD_IFCGEOM)
 
 IF(MSVC)
-	add_debug_variants(LIBXML2_LIBRARIES "${LIBXML2_LIBRARIES}" d)
+    add_debug_variants(LIBXML2_LIBRARIES "${LIBXML2_LIBRARIES}" d)
 ENDIF()
 
-# Open CASCADE headers
+# Open CASCADE
 IF("${OCC_INCLUDE_DIR}" STREQUAL "")
-        FIND_PATH(OCC_INCLUDE_DIR Standard_Version.hxx
-                 [PATHS
-                        /usr/include/occt
-                        /usr/include/oce
-                        /usr/include/opencascade
-                 ]
-                 REQUIRED
-        )
-        IF(OCC_INCLUDE_DIR)
-                MESSAGE(STATUS "Found Open CASCADE include files in: ${OCC_INCLUDE_DIR}")
-        ELSE()
-                MESSAGE(FATAL_ERROR "Unable to find Open CASCADE include directory, specify OCC_INCLUDE_DIR manually.")
-        ENDIF()
+    FIND_PATH(OCC_INCLUDE_DIR Standard_Version.hxx
+        [PATHS
+            /usr/include/occt
+            /usr/include/oce
+            /usr/include/opencascade
+        ]
+        REQUIRED
+    )
+    IF(OCC_INCLUDE_DIR)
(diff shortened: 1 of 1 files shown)


aothms:
Looks great. This is exactly the kind of simplification we need.

Just a question, instead of the ` IF("${OCC_INCLUDE_DIR}" STREQUAL "")` block, can that be folded directly into the `PATHS` block (I wasn't familiar with that CMake syntax block quotes yet), like:

```
FIND_PATH(OCC_INCLUDE_DIR Standard_Version.hxx
    [PATHS
        "${OCC_INCLUDE_DIR}"   # <-- here
        /usr/include/occt
        /usr/include/oce
        /usr/include/opencascade
    ]
    REQUIRED
)
```

Also, the `MESSAGE(FATAL_ERROR ...` can probably be removed since you have specified REQUIRED in FIND_PATH?

plaes:
> Looks great. This is exactly the kind of simplification we need.
> 
> Just a question, instead of the ` IF("${OCC_INCLUDE_DIR}" STREQUAL "")` block, can that be folded directly into the `PATHS` block (I wasn't familiar with that CMake syntax block quotes yet), like:
> 
> ```
> FIND_PATH(OCC_INCLUDE_DIR Standard_Version.hxx
>     [PATHS
>         "${OCC_INCLUDE_DIR}"   # <-- here
>         /usr/include/occt
>         /usr/include/oce
>         /usr/include/opencascade
>     ]
>     REQUIRED
> )
> ```
It could be done, but then I would have to fiddle with variable names, because once user submits "required" `OCC_INCLUDE_DIR` variable, it's detected as being present even if the directory is wrong.

> 
> Also, the `MESSAGE(FATAL_ERROR ...` can probably be removed since you have specified REQUIRED in FIND_PATH?
Fatal message short-circuits the configuration process fail, without it would run to the end and show more confusing error:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.                                                                                                                                        
Please set them or make sure they are set and tested correctly in the CMake files:                                
OCC_INCLUDE_DIR                                                                                                   
   used as include directory in directory IfcOpenShell/cmake
   used as include directory in directory IfcOpenShell/cmake
   ... ++ around 40 extra lines ...
-- Configuring incomplete, errors occurred!
....

aothms:
> It could be done, but then I would have to fiddle with variable names, because once user submits "required" OCC_INCLUDE_DIR variable, it's detected as being present even if the directory is wrong.

I see your point.

> REQUIRED
> 
> New in version 3.18.
> 
> Stop processing with an error message if nothing is found, otherwise the search will be attempted again the next time find_path is invoked with the same variable.

That's now how I read it in the docs, but ok, I guess you're right that we can make the error clearer to users

plaes:
> > It could be done, but then I would have to fiddle with variable names, because once user submits "required" OCC_INCLUDE_DIR variable, it's detected as being present even if the directory is wrong.
> 
> I see your point.
> 
> > REQUIRED
> > New in version 3.18.
> > Stop processing with an error message if nothing is found, otherwise the search will be attempted again the next time find_path is invoked with the same variable.
> 
> That's now how I read it in the docs, but ok, I guess you're right that we can make the error clearer to users

3.18 is quite new - https://repology.org/project/cmake/versions

StefanBruens:
Why don't you use `find_package(OpenCASCADE REQUIRED COMPONENTS ...)`?

PR #1364, opened 2021-03-08
Shorten Boost pathnames and fix IfcParse::IfcSIPrefixToValue()

ahladik:
Shortening the pathnames generated by the b2 utility to avoid breaking length limits on Windows and a fix to the IfcParse::IfcSIPrefixToValue function, where the 'DECA' conversion looks wrong.


aothms:
> where the 'DECA' conversion looks wrong.

Good eye. Thanks!

PR #1380, opened 2021-03-13
Misc cleanups for build-all.py script

plaes:
Drops really obsolete Python versions and bumps minor versions.


aothms:
Merged to 0.7 as #1777

PR #1381, opened 2021-03-13
build: actions: Don't hardcode parallel builds to 10 in release workflow

plaes:
And convert from deprecated `` `backtick` `` syntax to POSIX-compatible `$(...)`


aothms:
Thanks!

PR #1384, opened 2021-03-14
actions: enable ccache and minor speed improvements

plaes:
* Add ccache support (from ~20m of regular build to ~4 minutes with 100% cache hits)
* Other misc small improvements regarding dependency installations - install everything in one go and reduce unneeded packages.


aothms:
Awesome thanks!

PR #1385, opened 2021-03-14
Ivccsv fixes

plaes:
Fix one obvious error in ifccsv (would be really nice to if there was a testsuite..)

### DIFF
--- a/src/ifccsv/ifccsv.py
+++ b/src/ifccsv/ifccsv.py
@@ -6,7 +6,6 @@ import ifcopenshell.util.selector
 import ifcopenshell.util.element
 import ifcopenshell.util.schema
 import csv
-import lark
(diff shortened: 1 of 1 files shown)


Moult:
Thanks! Absolutely right there should be a test suite. I need to write a simple dummy IFC generator to be used in a test suite. The test suite would then be used for not just IFCCSV, but also IFCDiff, IfcPatch, IfcClash, BlenderBIM Add-on, etc.

Moult:
Btw the unused import was there as a lazy way for pyinstaller to catch it as a dependency for building a distributable, from memory. There's probably an argument to pyinstaller to do it properly :)

htlcnn:
It's `hiddenimport`

PR #1387, opened 2021-03-15
Travis updates: Python3 / Ubuntu 20.04 / ccache

plaes:
* Enable ccache and use `$(nproc)`
* Distro upgraded to 20.04 / Focal
* Switch to Python3

Build now takes ~15 minutes in case we don't touch any c++ files. Previously it took ~35m.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,27 +2,27 @@ language: cpp
 compiler: gcc
 cache: ccache
 os: linux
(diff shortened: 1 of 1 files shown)


plaes:
Current pull request is finished, please review :)

aothms:
Perfect, thanks, all looks good to me.

PR #1398, opened 2021-03-20
build: Use -Wno-deprecated-copy for GCC 9+

plaes:
OpenCascade spews a lot of warnings due to -Wdeprecated-copy.


aothms:
Thanks, should this if statement also test for COMPILER_ID then?

plaes:
> Thanks, should this if statement also test for COMPILER_ID then?

Yeah, I tested it with both clang and gcc - with clang there is not that much warning spam. Will post an updated patch.

aothms:
Thanks again @plaes

PR #1400, opened 2021-03-22
BlenderBIM - allow to run without OCC

yorikvanhavre:
This is a bit of a selfish PR... BlenderBIM currently doesn't run without PythonOCC. However, PythonOCC is a rather delicate package that needs some manual tweaks everytime a new version of OCCT is released. The current available version of PythonOCC relies on OCCT 7.5.0, but the version  currently in Debian unstable & testing is 7.5.1. PythonOCC doesn't compile with it.

It seems to me PythonOCC is used only for two operations: boolena-cutting an IFC file and writing to SVG. BlenderBIM could therefore be able to run without it, simply disabling these two functions?

This PR just disables the OCC imports at BlenderBIM launch. There is probably further changes needed in the UI too, so the correspinding tools get disabled to the user. So I'm leaving this PR in Draft mode for now for @Moult to orient me please ;)


yorikvanhavre:
Oops sorry about the ghost commits. Will repack it all later on

aothms:
> Oops sorry about the ghost commits. Will repack it all later on

We can also just squash it on our end. 


I will leave the merge to @Moult but the long term aim is to rely on IfcOpenShell for the SVG writing anyway.

Moult:
Holding off on merging this, as I reckon the entire import statement should be delayed, not just wrapped in a try/except. This will also fix the issue where Topologic users want to use their own OCC. We should only import when the user attempts to trigger one of these drawing features, and as soon as IfcConvert can read the view bounds and output with those boundaries / scale as per #1153 I can fully drop the dependency.

I'll find a bit of time and commit a temporary workaround, and hopefully soon I can delete the code altogether :D

Moult:
Ping @wassimj - this should also fix your woes when I get around to committing something.

yorikvanhavre:
okay :) don't hesitate to kill this one if you come up with something better

wassimj:
Great. I have used Topologic with if open shell on Dynamo so hopefully no conflicts.

Moult:
Closing as commit should resolve this. Users have to explicitly run the operator to trigger import.

Issue #1437, opened 2021-04-17
Possible bugs IfcEntityInstanceData constructors


RickBrice:
I'm not entirely familiar with IfcEntityInstanceData but it looks like there could be a couple of bugs in the constructors.

Item 1 - In the second constructor attributes_ is an array of Argument* pointers but the pointers are not initialized. 

```
IfcEntityInstanceData(IfcParse::IfcFile* file_, size_t size)
		: file(file_), id_(0), type_(0), attributes_(new Argument*[size]), offset_in_file_(0)
	{
             for (size_t i = 0; i < size; ++i) { attributes_[i] = 0; } //<--- Add something like this to initialize the elements of the array
        }
```

Item 2 - The class member offset_in_file_ is uninitialized in the third constructor.
```
   IfcEntityInstanceData(const IfcParse::declaration* type)
		: file(0), id_(0), type_(type), attributes_(new Argument*[getArgumentCount()])
/* ADD SOMETHING LIKE THIS --> */ , offset_in_file_(0)
	{
		for (size_t i = 0; i < getArgumentCount(); ++i) {
			attributes_[i] = 0;
		}
	}
```

aothms:
Hi, many thanks, unfortunately there are still not that many people that offer 2nd pair of eyes on the lower level C++ parts. So this is genuinely appreciated.

I think both issues will not lead to real bugs in current usage, but they definitely lead to confusion and potentially bugs in the future.

### 1st issue:

I think that prototype is only referenced here https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcparse/IfcSpfHeader.cpp#L39 in which case either the header instance is loaded from file or the array is set to zeros. I guess I was trying to eliminate unnecessarily clearing memory. But I agree leaving them uninitialized for the other constructor to clear them is very error prone. 

### 2nd issue:

offset_in_file_() is only read when calling load() which is only called when attributes_ == nullptr. But nevertheless it's sloppy.

https://github.com/IfcOpenShell/IfcOpenShell/blob/5961253c29f20b0c606bead448c417c6ead6af81/src/ifcparse/IfcParse.cpp#L1056


Note since C++11 I believe we can just do `new Argument*[getArgumentCount()]{0}` which save a for loop in code.


Would you be so kind to submit both suggestions as a Pull Request here on Github?

RickBrice:
[...]

PR #1438, opened 2021-04-18
Fixes initialization issue in IfcEntityInstanceData and Crash problem in debug builds

RickBrice:
Fixes data member initialization issues with IfcEntityInstanceData as discussed in #1437.

Also fixes crash problem in debug builds related to dereferencing the end iterator for empty std::string objects. See discussion thread https://sourceforge.net/p/ifcopenshell/discussion/1782716/thread/2930921968/


aothms:
Ah you're the same person. I hadn't noticed yet. Many many thanks.

PR #1445, opened 2021-04-27
More fixes to object initialization, fixes bug in valid_binary_string, adds parsing for Ifc4x3_rc2

RickBrice:
### IfcEntityInstanceData
Fixes test for empty string

### ArguementList
Initializes Argument pointers to nullptr

### IfcUtil::valid_binary_string
Fixes empty strings returning true

### Ifc4x3_rc2
Enabled parsing support for Ifc4x3rc2 schema


aothms:
Thanks again @RickBrice just a comment regarding:

> IfcUtil::valid_binary_string
> Fixes empty strings returning true

`valid_binary_string()` is a function that's actually only applicable to Python: https://github.com/IfcOpenShell/IfcOpenShell/search?q=valid_binary_string 

In Python if you want to supply a binary value you have to supply a string of 0 and 1.

`valid_binary_string()` is not related to the encoding of binary data in the IFC-SPF model file.

In the parser, it's already checked that a SPF binary is not length 0 https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcparse/IfcParse.cpp#L569

Can you clarify what you were aiming to accomplish with that change https://github.com/IfcOpenShell/IfcOpenShell/pull/1445/commits/ee96a7c4bafa0d730ba4f6649a90a1a30ac2069d

RickBrice:
@aothms I was reading through your source code to better understand how the parser worked. Looking at `valid_binary_string()` in isolation it appears to have a bug. I assumed the intent of the method is to return true when a string contains a sequence of characters consisting of only '0' or '1'. Any other string would return false. The boundary case of an empty string is not considered and the method returns true. This seems like a bug to me. All I was aiming to accomplish was to make the method a little more bulletproof.

From the description in your comment, it seems the implementation of `valid_binary_string()` depends on knowledge of its callers. Seems like the implementation could be improved a bit by making it independent of its callers.

aothms:
> From the description in your comment, it seems the implementation of valid_binary_string() depends on knowledge of its callers. Seems like the implementation could be improved a bit by making it independent of its callers.

Well, not exactly

> I assumed the intent of the method is to return true when a string contains a sequence of characters consisting of only '0' or '1'.

It's more the opposite: "return false when the string contains a character outside of {0, 1}". I'll add it as a comment. Merge your PR and then reverse that one commit about the binary string thing.

Some more background info: IfcOpenShell-Python doesn't have a type for binary. Just string of {0,1} so in order to be able to assign an empty binary to an attribute it should be possible to pass an empty string.

RickBrice:
> Some more background info: IfcOpenShell-Python doesn't have a type for binary. Just string of {0,1} so in order to be able to assign an empty binary to an attribute it should be possible to pass an empty string.

That makes more sense.

PR #1447, opened 2021-04-29
Add PredefinedType support

atomczak:
**What:**
- implemented support for IFC PredefinedType in IDS validation
- fixed typo in 'applicabiliy'
- modified output message to be more clear to users

**Why:**
Now it's possible to validate if IFC objects have PredefinedType required by the IDS file.

**How to test:**
Run ids.py file with IDS (XML) and IFC input files. Both inputs should have PredefinedType attribute to test it.


src/ifcopenshell-python/ifcopenshell/ids.py
```diff
@@ -76,17 +79,18 @@ class entity(facet):
...
     
-    # @todo predefinedtype
-
     def __call__(self, inst, logger):
-        logger.debug("Testing %s == %s", inst.is_a(), self.name)
         # @nb with inheritance
-        # return inst.is_a() == self.name
-        return facet_evaluation(inst.is_a(self.name), self.message % {"name": inst.is_a()})
-
+        if self.predefinedtype and hasattr(inst, "PredefinedType"):
```
aothms, line 86: * I guess if `not hasattr(inst, "PredefinedType")` it means the IDS is invalid (for this particular IFC schema at least) and it should be reported somehow. The question is if we do a IDS validity check before running the model compliance check. ifcopenshell also has the schema data, so you can check whether an IDS is valid without actually opening a model

* PredefinedType can also be defined on the IfcTypeObject entity. (IsDefinedBy/RelatingType, IsTypedBy/RelatingType)

* In the IFC4.3 schema the PredefinedType attribute can be a bit more complex. It can be a SELECT (union type) of several ENUMs. In that case the attribute is handled slightly different in IfcOpenShell. At some point let's check if there is an example https://github.com/buildingSMART/Sample-Test-Files/tree/master/IFC%204.3

src/ifcopenshell-python/ifcopenshell/ids.py
```diff
@@ -52,12 +52,15 @@ def __init__(self, node):
...
 
     def __getattr__(self, k):
-        v = self.node.getElementsByTagName(k)[0]
-        elems = [n for n in v.childNodes if n.nodeType == n.ELEMENT_NODE]
-        if elems:
-            return restriction(elems[0])
-        else:
-            return v.firstChild.nodeValue.strip()
+        try:
+            v = self.node.getElementsByTagName(k)[0]
```
aothms, line 56: I guess the `IndexError` is only about the first line `self.node.getElementsByTagName(k)[0]`. My preference then would be to only try-catch that line so that we don't by accident hide more errors later.

atomczak: ok, I narrowed it down to that particular line.

src/ifcopenshell-python/ifcopenshell/ids.py
atomczak: from the IDS XSD I understood that IDS doesn't neceserilly need to have PredefinedType in entity:
`<xs:element name="predefinedtype" type="xs:anyType" minOccurs="0">` 
so it would still be a valid IDS file, right?

On the other hand, if IDS requires a particular PredefinedType and IFC (inst) has none, it should fail the validation, currently, it skips the validation. What do you think?

aothms: > it should fail the validation, currently, it skips the validation. What do you think?

There's a subtle but important distinction here: (a) The attribute can be absent on the entity class or (b) the attribute can be unset (null, $).

I was hinting at these possibilities, like:

* A predefinedtype that isn't possible for an entity: like IfcWall SKYLIGHT
* A predefinedtype in an IDS for the IFC2X3 schema where the entity doesn't have a PredefinedType attribute.

In those cases I'd say the IDS itself is invalid.

It's also possible that the PredefinedType is unset or has value NOTDEFINED / USERDEFINED. I think those cases are properly handled (USERDEFINED is tricky perhaps) because you're simply comparing the value (can be `None`) to the value in the IDS XML.

aothms:
Many thanks for this @atomczak! Let's merge this in and keep the remaining issues as possible future work.

PR #1458, opened 2021-05-11
Update Boost download address

ahladik:
Boost download address has changed, so `build-deps.cmd` needs to be updated.


aothms:
I'm a bit confused, because the link was already changed, but it doesn't hurt to have some comments there https://github.com/IfcOpenShell/IfcOpenShell/commit/e98771ae0904f72d987d4bb49ce888ffe6d02bdc

ahladik:
Sorry, I didn't check before making the request! Please go ahead and delete the comment if you feel it's appropriate.

PR #1464, opened 2021-05-13
A few fixes of types and a typo

ahladik:
I corrected a few VC++ [ C4267 warnings](https://docs.microsoft.com/it-it/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4267?view=msvc-160) and what I think was an oversight.


aothms:
Thanks, good catch on typo...

PR #1466, opened 2021-05-15
disable boolean for mesh evaluation

s-leger:
Disabling boolean modifier on screen instead of remove / add is enough to get proper mesh representation.


Moult:
Thanks so much! Yes this is a much better way of doing it!

Issue #842, opened 2020-05-04
IFC4x3 support?


Moult:
Just curious what the process is to start supporting the IFC4x3 entities? There's a lot of good infrastructure additions.

aothms:
Is there a public version? I have some, but not sure if I can publish them.
Only downside is that compilation time is more or less linear with amount
of schemas. 0.7.0 addresses that. And alignment geometry from 4x1 is not
even implemented yet :(

Sent from a mobile device. Excuse my brevity. Kind regards, Thomas

Op di 5 mei 2020 00:09 schreef Dion Moult <notifications@github.com>:

Moult:
It was released a week or so ago :)

https://standards.buildingsmart.org//IFC/DEV/IFC4_3/RC1/HTML

Alignment geometry is one of the fun ones :) When you say alignment geometry isn't implemented, do you mean it isn't parsed to be faceted? If so, that's OK - because they are highly parametric, and should actually not be converted into a faceted curve if possible.

aothms:
> Alignment geometry is one of the fun ones :) When you say alignment geometry isn't implemented, do you mean it isn't parsed to be faceted? If so, that's OK - because they are highly parametric, and should actually not be converted into a faceted curve if possible.

Can you elaborate on that? Ideally we'd convert them into some sort of OCCT BRep / Wire / Curve / Adaptor. Some of the components are not implemented though such as clothoid and the other transitioning curves. What other approaches exist besides faceting? I looked around and couldn't find a quick and easy conversion from clothoid to nurbs for example. If you know someone who would like to take a stab at this for example would be great https://www.sciencedirect.com/science/article/pii/S0377042704000135

Moult:
@aothms oh sorry, I didn't phrase that quite right - I meant that even if IfcOpenShell could facet it - as I am building the BlenderBIM Add-on to be an authoring package too, I will likely have to parse each alignment segment separately, so that I can re-export it out of Blender with specific parameters changed and retain its full description. The faceted curve is good for visualisation, but not sufficient for full authoring, which is what I'm doing :)

aothms:
Ok, but for visualization you still need something faceted. And the higher upstream it is implemented the better I guess (i.e. IfcOpenShell C++). But feel free to begin implementing things and we can later move things to C++ also if needed.

I added the 4x3. Keep in mind the schema identifier is IFC4X3_RC1, in case you want to create a new file.

civilx64:
[...]

PR #1470, opened 2021-05-17
#842 - IfcAlignment - Clothoid transition implementation

civilx64:
Initial implementation of alignment transition curves, per discussion on #842.
At this time, only the Clothoid has been confirmed to return the correct X & Y coordinates.


aothms:
Awesome, thanks @civilx64! I'm just going to merge this as is, and then try to piece things together so that the ifcopenshell.geom.create_shape() function will direct either to this set of functions or to the C++ based conversion functions depending on content. What do you think @Moult is that a quick way to work towards support in BlenderBIM?

Moult:
@aothms sure thing :) From your description, it sounds as though I need to make no changes and create_shape will "just work". Will the iterator also "just work"?

civilx64:
@aothms @Moult one potential obstacle before it will "just work" is that an IfcAlignment is topologically just a series of edges between vertices joined into a wire. Looking at the BlenderBIM importer code it appears that Blender meshes are built from vertices and faces, but these entities would be vertices only. If in fact this is an issue then one solution might be to extrude a unit square maybe 10mm per side along the alignment to provide the faces for mesh creation in Blender. (Circular profile extrusion would work also but be much heavier in terms of face count.)

Also as you might have noticed the code is returning a list of gp_Pnts rather than a wire. Feel free to adjust as necessary. When testing it was easier for me to print and review x,y coordinates on the points rather than using something like Topo_Explorer to get a vertex list from the wire.

Moult:
The code I think you're looking at is the default for 3d bodies, but there is code which handles 2d (or 3d minus faces). Should be solvable :) @aothms can you let me know what's required to test IFC 4.3 and up? Which express file should I get? I assume I can just so "load from express" prior to loading the file? @civilx64 have you got any test files I can try?

aothms:
@Moult rc1 and rc2 are baked into the binaries. Only for rc3 and newer you need to supply and parse the exp schema file.

aothms:
@Moult I dropped them an email to invite you to the test file repos. Also see #1480

civilx64:
>  @civilx64 have you got any test files I can try?

I have two test files that I have been working with but they are 4.1 as I exported from commercial software.  I was thinking I would manually edit both to align with 4.3 rc3 and store both versions of both files in the [files repo](https://github.com/IfcOpenShell/files).

aothms:
But also, all three of us have access to the IFCRail/IFC-Rail-Unit-Test repo. Isn't there something that we can use? RC3 is not final either. There is going to be an RC4. And who knows what comes after that.

civilx64:
Yes, I had not been to the unit test repo in a while but after reviewing just now it should work nicely for verifying the IFCOS / BlenderBIM implementation.

PR #1476, opened 2021-05-19
Fix to IfcHierarchyHelper.cpp

ahladik:
Failure to pass the 'owner_hist' parameter caused unnecessary entities to be created.


aothms:
Thanks!

PR #1480, opened 2021-05-23
first rough partial draft of alignment handling in python

aothms:
@civilx64 please have a look

I tried to work my way from the other end, from the alignment definitions connecting to your transition curve definitions. I didn't get very far. 

Some points for discussion:
- I used numpy instead of OCCT. numpy ships with Blender these days and I think we can get more eyes on the code if we settle on numpy, but for now it's not a major concern
- Since RC3 there is this separation between semantics and geometry. I tried to look at both, but I'm not entirely sure what to use. Also, with the nesting of the alignments you can either get the composite curve at the Alignment(Hor|Ver|Cant) or look one decomposition level deeper at IfcAlignmentSegments (which is I think is optional).
- Maybe I went a bit overboard with a functional approach of functors for the curves and higher order functions for the placements. I'm not sure if it really helps to hide everything behind a function interface (curve = f(u)) because you need more information perhaps on the outside to reasonable interpolate the segment.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/alignment.py
+++ b/src/ifcopenshell-python/ifcopenshell/alignment.py
@@ -1,0 +1,191 @@
+from dataclasses import dataclass
+
+import numpy
+
+import ifcopenshell                             
+import ifcopenshell.geom                            
+import ifcopenshell.express
+import ifcopenshell.transition_curve
+
+# geometric primitives
+
+# @notes
+# - not sure if the separation of geometric primitives make sense
+#   does it make handling the variety of distance expressions and
+#   interpolation harder?
+
+@dataclass
+class line:
+    start_point: numpy.ndarray
+    direction_vector: numpy.ndarray
+    
+    def __call__(self, u):     
+        return self.start_point + \
+            self.direction_vector * u
+
+
+@dataclass
+class circle:
+    radius: numpy.ndarray
+    
+    def __call__(self, u):        
+        return numpy.array([
+            self.radius * numpy.cos(u),
+            self.radius * numpy.sin(u)
+        ])
+
+        
+def place(matrix, func):
+    """
+    Higher order function for application of a 3x3 matrix
+    to a 2D point. Assumes a functor such as line or circle.
+    """
+    def inner(*args):
+        v = func(*args)
+        # homogenize
+        v = numpy.insert(v, v.shape[-1], 1, axis=-1)
+        return (matrix @ v)[0:2]
+    return inner
+
+
+# mapping functions from IFC entities
+
+def map_inst(inst):
+    """
+    Looks up one of the implementation functions below in the global namespace
+    """
+    return globals()[f"impl_{inst.is_a()}"](inst)
+
+
+def impl_IfcLine(inst):
+    return line(
+        numpy.array(inst.Pnt.Coordinates),
+        numpy.array(inst.Dir.Orientation.DirectionRatios) * inst.Dir.Magnitude
+    )
+
+
+def impl_IfcCircle(inst):
+    return place(map_inst(inst.Position), circle(
+        inst.Radius
+    ))
+
+
+def impl_IfcClothoid(inst):
+    # @todo
+    # place = map_inst(inst.Position)
+    # ifcopenshell.transition_curve.TransitionCurve(
+    #     StartPoint          = place.T[2]
+    #     StartDirection      = numpy.arctan2(place.T[0][1], place.T[0][0]),
+    #     SegmentLength       = 
+    #     IsStartRadiusCCW    = 
+    #     IsEndRadiusCCW      = 
+    #     TransitionCurveType = 
+    #     StartRadius         = 
+    #     EndRadius           = 
+    # )
+    return lambda *args: numpy.array((0.,0.))
+
+
+def impl_IfcAxis2Placement2D(inst):
+    arr = numpy.eye(3)
+    
+    if inst is None:
+        return arr
+        
+    arr.T[2, 0:2] = inst.Location.Coordinates
+    
+    if inst.RefDirection is None:
+        return arr
+        
+    arr.T[0, 0:2] = inst.RefDirection.DirectionRatios
+    arr.T[0, 0:2] /= numpy.linalg.norm(arr.T[0, 0:2])
+    arr.T[1, 0:2] = -arr.T[0,1], arr.T[0,0]
+    
+    return arr
+        
+
+# conversion functions for semantic design parameters (not used atm)
+
+def convert(inst):
+    """
+    Looks up one of the conversion functions below in the global namespace
+    """
+    yield from globals()[f"convert_{inst.is_a()}_{inst.PredefinedType}"](inst)
+    
+
+def convert_IfcAlignmentHorizontalSegment_LINE(data):
+    xy = numpy.array(data.StartPoint.Coordinates)
+    yield xy
+    di = numpy.array([
+        numpy.cos(data.StartDirection),
+        numpy.sin(data.StartDirection)
+    ])
+    yield xy + di * data.SegmentLength
+    
+# Two approaches, either DesignParameters or Representation  
+
+def interpret_linear_element_semantics(settings, crv):
+    # traverse decomposition
+    for rel in crv.IsNestedBy:
+        for obj in rel.RelatedObjects:
+            yield from interpret_linear_element_semantics(settings, obj)
+            
+    # lookup design parameters and dispatch to conversion function
+    if crv.is_a("IfcAlignmentSegment"):
+        dp = crv.DesignParameters
+        yield from convert(dp)
+
+    
+def interpret_linear_element_geometry(settings, crv):
+    for segment in crv.Representation.Representations[0].Items[0].Segments:
+    
+        print(segment)
+        print(segment.ParentCurve)
+        print()        
+    
+        func = place(
+            map_inst(segment.Placement), 
+            map_inst(segment.ParentCurve)
+        )
+
+        for u in numpy.linspace(
+            segment.SegmentStart[0],
+            segment.SegmentStart[0] + segment.SegmentLength[0],
+            num=32
+        ):
+            yield func(u)
+            
+    
+interpret_linear_element = interpret_linear_element_geometry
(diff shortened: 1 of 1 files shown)


civilx64:
@aothms I had considered suggesting numpy previously so I'm onboard there. In terms of the code we should approach it in terms of calculating points at a specified distance along each of the elements. To get to 3D we need to calculate elevations as a function of distance along the alignment ("station" or "chainage"). That is not quite what I did for the clothoid implementation but will be required, I think. 

I don't think I follow "semantics" versus "geometry" yet for RC3... perhaps I have some research ahead RE: DesignParameter.

Lastly The approach for placement looks fine although I would think we would calculate all points for a given segment first and then transform with a 4x4 for translation and rotation. Or perhaps the specifics of transform versus rotation might vary depending on segment type (line versus arc versus transition).

aothms:
@civilx64 good point the correspondence between horizontal and vertical should probably be more central in the code. I think it could also be (but correct me if I'm wrong) that for horizontal and vertical there are completely different segments (i.e segments that don't correspond in distance along and length).

So instead of creating functions/functors for the segments we should concatenate them into piecewise function first so that we have a function for the entire length of the alignment?

civilx64:
> I think it could also be (but correct me if I'm wrong) that for horizontal and vertical there are completely different segments (i.e segments that don't correspond in distance along and length).

Exactly correct - horizontal and vertical (and cant) segments typically start and end at unrelated locations along the alignment.  Also it's common to have some length of 2D-only horizontal at the beginning and end of the alignment that are outside the range of the overall vertical alignment. I was taught that "you can never have too much horizontal alignment" and it's always easier to trim excess than extend an existing alignment.  Murphy's Law and all...  This also helps with showing additional long-section information for context and confirmation that there are no discontinuities in elevation and grade at the limits of the vertical alignment.
 
> So instead of creating functions/functors for the segments we should concatenate them into piecewise function first so that we have a function for the entire length of the alignment?

We certainly want a single ndarray for the whole alignment but keeping separate functions for the individual segments makes the most sense IMHO.  Here's my thought on approach:

1. Parse all alignment data to determine the locations for generating 3D geometry coordinates.  (2D would be the same data structure with NaN for the z value).  In addition to regular intervals along the alignment (as a whole, not each segment) this would also include a point at the start/end of each horizontal / vertical / cant segment, as well as points of interest such as vertical crest (local maxima) and sag (local minima) points.
2. Initialize the overall ndarray to store the points to be generated.
3. Parameterize and normalize the distance along the entire alignment with variable "t" in the range of 0.0 to 1.0.
4. Parameterize the horizontal segments with variable "h" and normalized distance along each segment. Example for an alignment with three horizontal segments:
    - h=0.0 is the beginning of the first horizontal segment.  Also corresponds to t=0.0 as the beginning of the overall alignment.
    - h=1.0 is the end of the first segment and the beginning of the second.
    - h=2.0 is the end of the second segment and the beginning of the third.
    - h=3.0 is the end of the third horizontal segment. Also corresponds to t=1.0 as the end of the overall alignment.
 5. Parameterize the vertical segments with variable "v" and normalized distance along each segment.  Example for an alignment with two vertical segments:
     - v=0.0 is the beginning of the first vertical segment.  May or may not also be t=0.0.
     - v=1.0 is the end of the first vertical segment and the beginning of the second.
     - v=2.0 is the end of the second vertical segment.  May or may not also be t=1.0.
 6. Correlate values of "t" parameter to "h" parameter to calculate each x, y coordinate in the overall ndarray, piecewise per horizontal segment
 7. Correlate values of "t" parameter to "v" parameter to calculate each z value in the overall ndarray, piecewise per vertical segment
 
I realize this got pretty wordy so my next step will likely be a worked example in a jupyter notebook similar to [this exercise](https://www.heavycivilbigdata.com/modeling-the-st-louis-arch.html#modeling-the-st-louis-arch) where I had some fun with computational geometry for the St. Louis Gateway Arch.  Not only could this help guide development now but it could prove useful in the long term as part of the IfcOS and BlenderBIM documentation.

aothms:
> 2D would be the same data structure with NaN for the z value

I would say 0 for z? With NaN's in there you probably cannot do matrix multiplications on them without the NaNs propagating everywhere.

--

I follow what you're saying. I guess I would have just parametrized all over the range [0, l] with l total alignment length, but I trust your experience on this.

I think we can just introduce two concepts:

```
def reparametrized_curve(fn, a, b):
    return lambda u: fn(a*u + b)
```

(or instead potentially somehow just normalizing assuming we can get recover the total length from the functor)

```
def reparametrized_curve(fn):
    return lambda u: fn(u / length(fn))
```

and

```
def piecewise(*args):
    # takes a set of functors and returns a function f(u) that delegates to the correct segment
    def inner(u):
        u0 = 0
        for fn in args:
             u1 = u0 + length(fn)
             if u >= u0 and u <= u1:
                 return fn(u - u0)
             u0 = u1
    return inner
```

Then based on these constructs you can either create a piecewise() of n reparametrized_curve(fn) to get your h<=1, h<=2,h<=3 parametrization

Or create a piecewise() of the original segments to get to my 0<u<L parametrization.

civilx64:
> > 2D would be the same data structure with NaN for the z value
> 
> I would say 0 for z? With NaN's in there you probably cannot do matrix multiplications on them without the NaNs propagating everywhere.

NaN's actually a good thing IMO.  If you were to calculate a cross product of a 2D and 3D vector you would want the 3x3 to drop to 2x2 because the result is only valid in a plane.  Also elevation z = 0 is valid and common for coastal works near sea level.  Using NaN indicates explicitly that there is no z information.  Worst case we can use `numpy.nan_to_num()` to avoid a linear algebra catastrophe.

>  I guess I would have just parametrized all over the range [0, l] with l total alignment length, but I trust your experience on this.
 
0 > t > 1 is "more correct" in the academic sense but after some more thinking it would be better to use [0, L].  We would have to divide through by L to get to t, then multiply back again to get `distance_along` values anyway.  Let's stick with [0, L].

> I think we can just introduce two concepts:
 
TIL what a functor is. :smile: Yes, this makes sense and I like the functional approach.

> Or create a piecewise() of the original segments to get to my 0<u<L parametrization.

Yes, let's do that.

aothms:
> NaN's actually a good thing IMO. If you were to calculate a cross product of a 2D and 3D vector you would want the 3x3 to drop to 2x2 because the result is only valid in a plane. Also elevation z = 0 is valid and common for coastal works near sea level. Using NaN indicates explicitly that there is no z information. Worst case we can use numpy.nan_to_num() to avoid a linear algebra catastrophe.

On second thought I fully agree. In the main IfcOpenShell C++ code base we don't make use a lot of explicitly 2d entities, but rather use 3d with z=0, but there it makes sense. Because the 2d basis of an extrusion isn't different from the 3d bottom face after extrusion. In case of alignments it's obviously different.

May I ask then what's the advantage of [x, y, NaN] instead of just [x, y] ?

civilx64:
[x, y, Nan] is useful because it lets you use the same data structure across the entirety of the alignment regardless of vertical alignment status.  For instance, say we are parsing an alignment with total length 2000m but no vertical alignment is defined over the first and last 200m.  In that case we have a mix of 2d and 3d.  The list of coords generated by `piecewise()` would have [x, y, NaN] over the first and last 200m with [x, y, z] from station 0+200 to 1+800.

Theoretically we could parse the alignment ahead of time and use [x, y] if there are no vertical alignments defined.  But that adds an additional special case to think about.  This is also partially personal preference - the primary CAD platform I have used for years makes a distinction between 2D and 3D files.  It's not always clear which should be used in which instances.  My philosophy is that the world is 3D, so work in that environment and just drop z when not applicable.

aothms:
>  The list of coords generated by piecewise() would have [x, y, NaN] over the first and last 200m with [x, y, z] from station 0+200 to 1+800.

Makes sense! I didn't think of the partial vertical covering...

civilx64:
I refactored `transition_curves.py` based on our discussion and finished implementing the missing transition types.  I assume that scipy as a dependency is not a major concern.  It appears that sverchok utilizes it as well and `scipy.integrate` was invaluable for some numerical integration needed for specific transition curves. 

I ran some basic tests and confirmed that all transitions are returning reasonable values.  I anticipate more unit testing needed with the IFC-Rail datasets to confirm that all calculations have been implemented correctly.

I did not send a PR but you can find the updates on [my fork](https://github.com/civilx64/IfcOpenShell/commit/a8a296d3571aa307678896289e228b936358ba51).  Let me know if any changes are needed for you to be able to implement the `piecewise()` approach.  Thanks!

aothms:
@civilx64 great work. I don't think scipy is immediately available from Blender (at least not on the version I tried) but I also don't see an immediate alternative. I guess we can update @Moult BlenderBIM packaging script to include it. Or instruct users they need to download and run get-pip.py from Blender and then `pip install --user scipy`. I think this can be done lazily even if the functionality is required.

I have updated alignment.py to have the piecewise function and a bit of hackery to store the length on the functors by wrapping them in another class `trimmed_curve`. What should we do from here? Shall we merge the two separate .py files?

civilx64:
Yes, let's merge `transition_curve.py` into `alignment.py` as the transitions only make sense in the context of an alignment.

What version of the Ifc4x3_RC3 express schema are you using when running `alignment.py` as a script?  I get a pyparsing recursion depth error for both the version in the [buildingSmart repo](https://github.com/buildingSMART/IFC4.3.x-development/tree/master/reference_schemas) and the one in [buildingSmart technical docs](https://standards.buildingsmart.org/IFC/DEV/IFC4_3/RC3/HTML/).

Lastly, I assume we will target support of 4.1 in addition to 4.3 when it is finalized?  Most commercial packages that I currently utilize appear to support the 4.1 alignment schema.

aothms:
> What version of the Ifc4x3_RC3 express schema are you using when running alignment.py as a script? I get a pyparsing recursion depth error for both the version in the buildingSmart repo and the one in buildingSmart technical docs.

What version of pyparsing do you have? I have `pyparsing==2.4.7`.

> Lastly, I assume we will target support of 4.1 in addition to 4.3 when it is finalized? Most commercial packages that I currently utilize appear to support the 4.1 alignment schema.

Yes I think we should.

Issue #1481, opened 2021-05-23
blenderbim polygons not strictly coplanar


s-leger:
Ifc expect polygons to be strictly coplanar, but in blender context you may very easly get non strictly planar polygons by eg editing any vertex location of a simple box.

In order to fix this issue we should add a set of 2 modifiers on the fly before evaluating representation and remove once done.

1 Triangulate
2 Decimate in 'DISSOLVE'  mode with angle limit under 1° and {'MATERIAL'} as delimitation.

PR #1482, opened 2021-05-23
Ensure polygons are always coplanar

s-leger:
fix issue #1481
May adopt a bmesh approach instead ?

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/geometry/add_representation.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/geometry/add_representation.py
@@ -59,10 +59,26 @@ class Usecase:
             del bm
             self.settings["geometry"] = mesh
         else:
+            
+            if len(self.settings["blender_object"].data.polygons) > 1:
+                self.settings["blender_object"].modifiers.new("BIM-Triangulate", 'TRIANGULATE')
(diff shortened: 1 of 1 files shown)


s-leger:
Bmesh approach is way faster (and cleaner) than modifier based one.

Moult:
Fantastic approach! Great solution.

aothms:
while I like the approach and I'm sure it will work in a lot of cases. The tolerance in IFC is linear, not angular. So for large faces it may still pose problems.

Is there anyway you can make the `triangulate()` op work only on a subset of faces? For every polygon you can loop over the vertices and take the dot product with the polygon normal. Take min and max of those dot product values. If the difference between min and max is greater than the model tolerance, the face needs to be triangulated.

s-leger:
Make sense, triangulate() takes faces as input, so we may filter those faces. 
Agree that we must stronger approach to handle all cases given the fact that we are working in single precision.
Probably a any([abs(dot) > tol]) will do the trick. 
Looking for anybody able to implement this, my attempt to do such test rely on matrix multiplication (matrix located at first vertex with normal as z axis) @ vertex.coordinate absolute z distance, but maybe dot method will be faster - wich is a hard requirement as every single polybon must be checked.

PR #1489, opened 2021-05-26
Evaluate coplanar vertex in faces

s-leger:
The idea is to only triangulate non co-planar faces when required.

Timings for evaluation with matrix based algo.
IfcRoof/Roof found 3927 of 116178 polygons: 0.366638 sec
IfcWall/Wall.012 found 0 of 6 polygons: 0.000046 sec

So should be acceptable for common use cases
But with single precision in blender we may still be over occ precision with a 1e-6 epsilon.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/geometry/add_representation.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/geometry/add_representation.py
@@ -1,8 +1,12 @@
 import bpy
 import bmesh
 import ifcopenshell.util.unit
-from mathutils import Vector
+from mathutils import Vector, Matrix
 from blenderbim.bim.module.geometry.helper import Helper
+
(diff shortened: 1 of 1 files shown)


s-leger:
ping @aothms 
Does ifcOpenshell even use the IfcRepresentationContext precision value at import time ?

Moult:
Good to merge, or should I wait?

aothms:
>  Does ifcOpenshell even use the IfcRepresentationContext precision value at import time ?

Yes, but a bit in a weird way. A global value is determined from the smallest precision encountered in the file.

Nice implementation using m4. It's a bit different from what I suggested, ie not max-min, but comparing to verts[0], but that's why IfcOpenShell also enlarges the tolerance, because you never know how it is implemented exactly.

s-leger:
@aothms thanks for the clarification.
@Moult i guess we can merge now.

PR #1500, opened 2021-05-28
Allow global listeners

s-leger:
Add "global" listeners when ifc_file is None, listeners are run whatever the file.
This way allow to register listeners before a file exists, and "global" listeners.

Add listener returns the ifc_key
Remove listeners allow optional filtering by usecase_path / ifc_key.usecase_path.


s-leger:
Shouldn't such methods be documented using docstrings ?

Moult:
Yes, need docstrings :D I've just been lazy.

PR #1530, opened 2021-06-18
IDS - xmlschema, restrictions and requirements

atomczak:
- Switched from xml.dom to xmlschema.
- improved requirements with proper validation
- implemented missing restrictions


src/ifcopenshell-python/ifcopenshell/ids.py
```diff
@@ -97,17 +105,23 @@ class classification(facet):
     message = "a classification reference '%(value)s' from '%(system)s'"
 
     def __call__(self, inst, logger):
+        #TODO Location: 'type'/'instance'/'any'
         refs = []
         for association in inst.HasAssociations:
             if association.is_a("IfcRelAssociatesClassification"):
                 cref = association.RelatingClassification
-                refs.append((cref.ReferencedSource.Name, cref.ItemReference))
+                refs.append((cref.ReferencedSource.Name, cref.Identification))  # before was .ItemReference instead of .Identification
```
aothms, line 113: This still needs to be handled in a schema-agnostic way right? ItemReference for 2x3, Identification for IFC4.

aothms:
Looks really good I think. Couple of small remarks, some only for future work.

src/ifcopenshell-python/ifcopenshell/ids.py
atomczak: Of course. Left comment to myself and forgot to reflect both schemas in code. Just pushed a commit with that change.

aothms:
Thanks! Really cool to see the progress :)

aothms:
PS we're running the tests on Travis CI now as well so we get notified in case something breaks

afbeelding

https://app.travis-ci.com/github/IfcOpenShell/IfcOpenShell/builds/232381022#L1441

Issue #1533, opened 2021-06-21
IfcConvert: Xml double value is rounded


ketourneau:
We convert an georeferenced ifc (with use world coords) to xml with IfcConvert and we display "object placement" value from xml to our customer.

Double values in xml result is rounded, so we loose precision.

Ifc declaration :
```
#22= IFCCARTESIANPOINT((1350666.398080918006599,6240146.763804821297526,0.));
```

Xml result :
<img width="1227" alt="Capture d’écran 2021-06-21 à 15 42 00" src="https://user-images.githubusercontent.com/39763266/122772333-9076ac80-d2a7-11eb-9ad3-1f8f3f5429fe.png">

Is there a way to set IfcConvert to not round double value in xml ?

PR #1534, opened 2021-06-21
Fix xml double value is rounded #1533

ketourneau:
When we convert ifc to xml, double value is rounded so we loss position precision.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -944,7 +944,7 @@ int main(int argc, char** argv) {
         }
 
         std::stringstream msg;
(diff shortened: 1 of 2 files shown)


aothms:
Thanks, looks good to me. In other parts of the code base (spf serialization) we use digits10 instead of max_digits10, but maybe max_digits10 is indeed better. I'm not 100% of the details.

Note that the example in your issue appears to be roughly 23 digits of precision, so even more than the 17 digits offered by max_digits10. Theoretically IfcOpenShell could be modified so that it outputs the exact verbatim representation of the float when parsed because we retain the token positions. But I doubt your XML parser then would retain the precision after 17 decimals.

PR #1539, opened 2021-06-27
Experimental undo and redo Python prototype

Moult:
Don't merge, just created PR for prototyping review

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/file.py
+++ b/src/ifcopenshell-python/ifcopenshell/file.py
@@ -90,13 +90,18 @@ class Transaction:
                 self.file.remove(element)
             elif operation["action"] == "edit":
                 element = self.file.by_id(operation["id"])
-                element[operation["index"]] = self.unserialise_value(element, operation["old"])
+                try:
+                    element[operation["index"]] = self.unserialise_value(element, operation["old"])
+                except:
+                    # Catch discrepancy where IfcOpenShell creates but doesn't allow editing of invalid values
+                    pass
             elif operation["action"] == "delete":
                 e = self.file.create_entity(operation["value"]["type"], id=operation["value"]["id"])
                 for k, v in operation["value"].items():
                     try:
                         setattr(e, k, self.unserialise_value(e, v))
                     except:
+                        # Catch discrepancy where IfcOpenShell creates but doesn't allow editing of invalid values
                         pass
                 for inverse_id, data in operation["inverses"].items():
                     inverse = self.file.by_id(inverse_id)
@@ -119,6 +124,7 @@ class Transaction:
                     try:
                         setattr(e, k, self.unserialise_value(e, v))
                     except:
+                        # Catch discrepancy where IfcOpenShell creates but doesn't allow editing of invalid values
                         pass
             elif operation["action"] == "edit":
                 element = self.file.by_id(operation["id"])
--- a/src/ifcopenshell-python/ifcopenshell/file.py
+++ b/src/ifcopenshell-python/ifcopenshell/file.py
@@ -158,7 +158,7 @@ class file(object):
             args = filter(None, [schema])
             args = map(ifcopenshell_wrapper.schema_by_name, args)
             self.wrapped_data = ifcopenshell_wrapper.file(*args)
-        self.history_size = 5
+        self.history_size = 64
         self.history = []
         self.future = []
         self.transaction = None
--- a/src/ifcopenshell-python/ifcopenshell/file.py
+++ b/src/ifcopenshell-python/ifcopenshell/file.py
@@ -40,16 +40,14 @@ class Transaction:
         self.file = ifc_file
         self.operations = []
 
-    def serialise_entity_instance(self, info):
+    def serialise_entity_instance(self, element):
+        info = element.get_info()
         for key, value in info.items():
-            if isinstance(value, entity_instance):
-                info[key] = {"type": "single", "id": value.id()}
-            elif isinstance(value, tuple) and isinstance(value[0], entity_instance):
-                info[key] = {"type": "multiple", "ids": [v.id() for v in value]}
+            info[key] = element.walk(lambda v: isinstance(v, entity_instance), lambda v : {"id": v.id()}, value)
         return info
 
     def store_create(self, element):
-        self.operations.append({'action': 'create', 'value': self.serialise_entity_instance(element.get_info())})
+        self.operations.append({'action': 'create', 'value': self.serialise_entity_instance(element)})
 
     def store_edit(self, element, index, value):
         self.operations.append({'action': 'edit', 'id': element.id(), 'index': index, 'old': element[index], 'new': value})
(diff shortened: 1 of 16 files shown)


src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -81,11 +209,15 @@ def create_entity(self, type, *args, **kwargs):
             >>> #3=IfcPerson('Foobar',$,$,$,$,$,$,$)
         """
         e = entity_instance((self.schema, type))
+        # Hack, can a entity access its file?
```
aothms, line 212: yeah we need something for this, probably implemented in a different way so that not only create_entity() but also attribute access and other ways of obtaining an entity_instance have a way to access their file

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -35,6 +35,102 @@
...
+class Transaction:
+    def __init__(self, ifc_file):
+        self.file = ifc_file
+        self.operations = []
+
+    def serialise_entity_instance(self, info):
+        for key, value in info.items():
+            if isinstance(value, entity_instance):
+                info[key] = {"type": "single", "id": value.id()}
+            elif isinstance(value, tuple) and isinstance(value[0], entity_instance):
```
aothms, line 47: I usually do `set(map(type, value)) == {entity_instance}`.

We also need `isinstance(tuple) and set(map(type, value)) == {tuple} and set(map(type, sum(value, ())) == {entity_instance}` for LIST OF LIST OF x

We also have entity_instance.walk() for this, something like:

```
is_instance = lambda v: isinstance(v, entity_instance)
get_id = lambda inst: inst.id()
value = entity_instance.walk(value, is_instance, get_id)
```

This recursively visits the list of list of any other attribute value type and applies get_id() when is_instance() returns true. So something like a recursive combination of filter() and map()

aothms:
I think I'd do it slightly differently, but then again I didn't take the time to actually test this, but let's say that you start with

self.history = []

and then when you do begin_transaction() you do

self.history.append(Transaction())

the current transaction, if any, you access with

self.history[-1]

undo/rollback() then pops() from self.history and shifts this to another list (let's say self.future)

redo() pops from the front of self.future and appends to history

and, most importantly, any change invalidates the redo stack, sets `self.future[:] = []` 

---

Small comment, if you use Transaction terminology I would call it rollback() commit()

aothms:
And yes I guess I have to come up with a way to reliably return the file from any entity_instance. We have this in C++, but the difficult bit is that we don't want the C++ IfcFile object, but instead the decorated ifcopenshell.file (which has C++ IfcFile in .wrapped_data). This probably requires some sort of global mapping, so that we can navigate from C++ IfcFile to ifcopenshell.file.

Moult:
Happy to change to a pop history/future, instead of the index as I currently have it :) Let me know if you commit any fixes on the C++ side that I can clean this up with!

aothms:
```
>>> import ifcopenshell
>>> f = ifcopenshell.open()
TypeError: open() missing 1 required positional argument: 'fn'
>>> f = ifcopenshell.file()
>>> f.createIfcPerson(_id=6)
#6=IfcPerson($,$,$,$,$,$,$,$)
>>> f.createIfcPerson()
#7=IfcPerson($,$,$,$,$,$,$,$)
>>> f.remove(f[6])
>>> f.createIfcPerson(_id=6)
#6=IfcPerson($,$,$,$,$,$,$,$)
>>> f.createIfcPerson(_id=7)
  File "C:\Users\tkrij\Miniconda3\envs\py35\lib\site-packages\ifcopenshell\file.py", line 88, in create_entity
    self.wrapped_data.add(e.wrapped_data, eid)
  File "C:\Users\tkrij\Miniconda3\envs\py35\lib\site-packages\ifcopenshell\ifcopenshell_wrapper.py", line 1540, in add
    return _ifcopenshell_wrapper.file_add(self, entity, id)
RuntimeError: An instance with id 7 is already part of this file
```

src/ifcopenshell-python/ifcopenshell/file.py
Moult: Wow walk is amazing! I saw it but had no idea how to use it. I hope the last commit makes sense and fixes this!

Moult:
Last commit should resolve comments :) Wow! That was fast! I'll wait for the bot build :D

Moult:
Curious why `_id` was used and not just id? If it were just `id` you could do `f.create_entity(**foo)` where `foo = bar.get_info()` from where?

aothms:
Hm yes, that's indeed also asymmetric. The reason is, that I was worried about some attributes being named Id. And officially express is case insensitive for a lot of stuff (I didn't look up the specifics).

But yes, feel free to change _id to id in my latest commit.

Moult:
Last commit makes it much more stable. Here's a sample script - feel free to play around with multiple undos, redos, and transaction checkpoints. I think I'm now ready to start testing on bigger complex data via the Blender interface and existing files.

```
import ifcopenshell
import ifcopenshell.api

def debug(f, message):
    print('>>> ' + message)
    for e in f:
        print(e)

f = ifcopenshell.api.run("project.create_file")
project = ifcopenshell.api.run("root.create_entity", f, ifc_class="IfcProject", name="My Project")
ifcopenshell.api.run("unit.assign_unit", f)
ifcopenshell.api.run("context.add_context", f, context="Model")
ifcopenshell.api.run("context.add_context", f, context="Model", subcontext="Body", target_view="MODEL_VIEW")
ifcopenshell.api.run("context.add_context", f, context="Model", subcontext="Box", target_view="MODEL_VIEW")
f.begin_transaction()
ifcopenshell.api.run("context.add_context", f, context="Plan")
ifcopenshell.api.run("context.add_context", f, context="Plan", subcontext="Annotation", target_view="PLAN_VIEW")
site = ifcopenshell.api.run("root.create_entity", f, ifc_class="IfcSite", name="My Site")
building = ifcopenshell.api.run("root.create_entity", f, ifc_class="IfcBuilding", name="My Building")
storey = ifcopenshell.api.run("root.create_entity", f, ifc_class="IfcBuildingStorey", name="Ground Floor")
f.end_transaction()
f.begin_transaction()
ifcopenshell.api.run("aggregate.assign_object", f, product=site, relating_object=project)
ifcopenshell.api.run("aggregate.assign_object", f, product=building, relating_object=site)
ifcopenshell.api.run("aggregate.assign_object", f, product=storey, relating_object=building)
f.discard_transaction()

debug(f, 'see')
f.undo()
debug(f, 'see')
f.redo()
debug(f, 'see')
```

Moult:
@aothms do you mind if I merge this PR? I _think_ it won't impact regular users who don't begin a transaction, but it will help with the daily builds as I start attempting to sync the Blender undo system with the IfcOpenShell undo system (yikes!)

Moult:
The last commit doesn't affect ifcopenshell-python at all, but simply implements undo stack syncing between Blender's undo system (taking care of all Blender RNA), the BlenderBIM Add-on undo system (taking care of all Blender<->IFC sync state and linking, as well as oddities like IFC header data), and ifcopenshell-python undo system (taking care of all IFC data). It'll evolve over time but I think it's a good start and it seems to work well, including nested operations.

I'm committing to this branch simply because it depends on the ifcopenshell-python changes, but basically feel free to ignore that commit :) Let me know when you feel comfortable with merging this branch.

aothms:
> do you mind if I merge this PR?

If it all breaks down I'm sure we can fix it :)

Moult:
Hehehe that's what Git is for! Let's break things!

Issue #1029, opened 2020-10-07
ifcParse Link Error when build with Visual C++


NextBim:
When building ifcParse using Microsoft Visual Studio this error occurs : 
LNK1189 : The limit of 65535 objects or members in a library has been exceeded.

Nobody else is building ifcOpenShell using MSVC++ ?

aothms:
This is solved by adding `/bigobj` to the compiler command line. The cmake configuration takes care of this. How did you obtain your MSVC solution? https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/cmake/CMakeLists.txt#L407

NextBim:
Thank you @aothms for your feedback.
I used the given CMakeLists.txt and  bigobj is set, but when building with BUILD_SHARED_LIBS set to ON, the previous link error appear. 
With BUILD_SHARED_LIBS set to OFF, everything is fine. We will use static libraries to bypass the issue.

aothms:
Interesting, I wouldn't have guessed that's related, I'll have a look.

HeySora:
I can confirm this happens.

Exact instructions I did, for reproducing the issue:

```
git clone -b v0.6.0 --single-branch https://github.com/IfcOpenShell/IfcOpenShell.git
cd IfcOpenShell\win
build-deps vs2017 Release
run-cmake.bat vs2017 -DENABLE_BUILD_OPTIMIZATIONS=ON -DBUILD_SHARED_LIBS=ON -DBUILD_IFCPYTHON=OFF -DBUILD_EXAMPLES=OFF -DBUILD_GEOMSERVER=OFF -DBUILD_IFCMAX=OFF
(then, opening IfcOpenShell.sln, defining IfcConvert as the default project, and build the solution)
```
It indeed does not happen with `BUILD_SHARED_LIBS` unset.

carmenfan:
Also running into this problem when I try to upgrade our dependency from v0.4 to v0.6 :( 

We have other dependencies in the application which are shared libraries so unfortunately building it as static is not really an option. Looks like for now I will have to write a dynamic library to wrap around it. just wanted to bump this for considerations!

aothms:
>  LNK1189 : The limit of 65535 objects or members in a library has been exceeded.

> This is solved by adding /bigobj to the compiler command line.

I was wrong. /bigobj is during compilation. LNK1189 is about linking the actual dll, which apparently has a hard limit on 2^16 -1 symbols.

[...]

PR #1545, opened 2021-07-01
Fixes to enable building shared library in windows

carmenfan:
https://github.com/IfcOpenShell/IfcOpenShell/issues/1029 is still a problem when you try to build shared libraries in windows. This can be bypassed with reducing the number of schemas to compile. However there are still some linking errors:

1. IFCGeom needs `IfcParse.lib` as a dependency during compilation
2. Several functions are used by IfcGeom within IfcParse that are missing `IFC_PARSE_API` declaration

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -494,7 +494,7 @@ function(files_for_ifc_version IFC_VERSION RESULT_NAME)
     )
 endfunction()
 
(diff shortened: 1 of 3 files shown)


aothms:
Many thanks :)

Moult:
Just also want to chime in with a thank you @carmenfan ! It's really inspiring to see developers hacking especially on the C++ side!

PR #1548, opened 2021-07-04
Support for IFC 4.3 RC3 and RC4

RickBrice:
This pull request implements support for IFC 4.3 RC3 and RC4.

Implementation in IfcGeom::_getSurfaceStyle was a little tricky because IfcPresentationStyleAssignment was removed from the specification.

Also revised .gitignore to ignore build artifacts from multiple MSVC platforms.


aothms:
Really helpful, thanks.

> Implementation in IfcGeom::_getSurfaceStyle was a little tricky because IfcPresentationStyleAssignment was removed from the specification.

That's odd, to remove this in between RCs. It's not really related to the domain right? Well, it's a lot of indirections to assign styles, it won't be missed :)

PR #1580, opened 2021-07-22
Bimtester: Fixed attributes_psets step files

Nitschi:
I ported it to use the new modules.
Let me know if I misunderstood anything.


Moult:
Awesome work! Thanks so much!

PR #1583, opened 2021-07-23
GeomIterator is not exposed for use if the library is built dynamically on windows

carmenfan:
Iterator functions/classes were missing the IFC_GEOM_API definition, making it impossible to access the geometry if the library is built as a dll


aothms:
Thanks!

Issue #1584, opened 2021-07-23
head of v0.6.0 branch does not compile on windows


carmenfan:
Hi, apologies if I'm doing anything stupid here (I tried my best to see if I can resolve it but failed 😆)

I was using the branch and compiled fine in the beginning of July. Today I pulled in the latest changes from `v0.6.0` and it's now failing to compile (both static/dynamic library). The problem seems to be to do with the `MAKE_TYPE_NAME` macro is not concatenating properly:

![image](https://user-images.githubusercontent.com/11945337/126782539-dcfa6034-c2cb-4cc9-993d-acc8fc1d4083.png)

As you can see on the error log, it's returning `IfcGeom::KernelIfc2x3` instead of `IfcGeom::Kernel:Ifc2x3`

If I go back to the commit I was in (30f778e0a7b7e2f9dfd4c7466ddcdb2a74c93022) it compiles fine. Is there some configuration change I am missing?

aothms:
Hm, I also don't see what this could be. Asking the obvious, clean git work tree? did you try to clean the msvc project?

The concatenation is supposed to be like that. It just concatenates schema name directly after Kernel. The unmatched left brace { error is probably the easiest to verify. I see there have been some preprocessor changes involving braces in geom.h but it compiles fine for me.

carmenfan:
@aothms 

ok tracing the curly brace definitely helped! thanks!

the problematic line is this:
![image](https://user-images.githubusercontent.com/11945337/126990026-3db6359c-1678-4a3b-a91a-6074f644e0fb.png)

as you can see vc evaluated this to false even though it satisfied the first part. Maybe it's to do with the version of compiler not understanding the `or`? I'm using vc14.

Rewriting this to the following worked for me:

````c++
#if defined(SCHEMA_HAS_IfcStyleAssignmentSelect) || defined(SCHEMA_HAS_IfcPresentationStyleAssignment)
	   }
#endif
````

I shall make a PR. thanks for the help!

PR #1595, opened 2021-07-26
fix hash define logical operation

carmenfan:
The line that has changed was not being evaluated correctly in the previous version, resulting in compilation error on windows vc 14 compiler.

See details here https://github.com/IfcOpenShell/IfcOpenShell/issues/1584#issuecomment-886672675


aothms:
Thanks for tracking this down! I glanced over this, but didn't spot the error. I didn't open the C++ spec for this, but it looks suspicious to me, I'm surprised gcc and msvc2017 accepted this without a complaint.

PR #1607, opened 2021-07-29
Use passed context instead of bpy when available

Gorgious56:
I realize I may have bitten more than I can chew on this one but I think I pretty much covered it all. 
Do tell me if you rather I divide it into smaller chunks.


Moult:
Perfect! No need to break it down as this change is probably better seen as a single edit :)

PR #1612, opened 2021-07-30
Fix error trying to access unset ifc file

Gorgious56:
This fixes an error thrown in console when user opens the N panel when no ifc file is set (value is `None`).
I saw the fix was already present in one of the methods in the file (L43 
`def getIfcTypes`), but I figured it would be more efficient to test if file exists only after testing if we actually need to populate the enums. Please tell me if I need to change anything.

### DIFF
--- a/src/blenderbim/blenderbim/bim/ifc.py
+++ b/src/blenderbim/blenderbim/bim/ifc.py
@@ -184,9 +184,13 @@ class IfcStore:
     def execute_ifc_operator(operator, context):
         is_top_level_operator = not bool(IfcStore.current_transaction)
 
+        file = IfcStore.get_file()
+        if not file:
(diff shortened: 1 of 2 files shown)


Gorgious56:
I saw there are several instances where the code assumes there is an ifc file. Do I hunt down these instances and add a check ?

Moult:
I think it is better to instead not show any UI at all via the poll check if an IFC file is not available. Then we don't need this check in the operators.

Gorgious56:
@Moult Ok, so if we are executing the operator, we assume the file exists.

Just so I'm sure, do we add the check to the `poll` of every operator much like the `execute` that calls `IfcStore.execute_ifc_operator` on every operator or do we hide the panels where they are potentially drawn instead ? (or both :) )

PR #1613, opened 2021-07-30
Add check before showing potentially empty enum

Gorgious56:
This was not trivial since trying to access the enum to check if it is empty by regular means (`props.ifc_class`) would throw out the warning error anyway.

Found the solution there https://blender.stackexchange.com/a/35162/86891


Gorgious56:
I don't know how to change the PR to reflect it but this is not ready to merge yet, I've found other instances where I need to add a check

Gorgious56:
But I don't have data to see if this actually works. How does one populate these enums by regular means ?

Gorgious56:
Closing this because this workaround is wrong. Should have read the comments on the answer.

Issue #1609, opened 2021-07-29
BlenderBIM : Setting a person's middle names, prefixes or suffixes throws error if not formatted correctly


Gorgious56:
Using latest 0.6.0 branch, ifc 2x3
I'm trying to add a new person in the IFC People panel.
![image](https://user-images.githubusercontent.com/25156105/127520228-67b069b1-1b66-47c5-9c19-65bf7c8b47d5.png)

It works if only the name and surname are filled, but throws error and doesn't validate if either a middle name, prefix, or suffix is entered.

```
Error: Python: Traceback (most recent call last):
  File "C:\Users\ \AppData\Roaming\Blender Foundation\Blender\2.93\scripts\addons\blenderbim\bim\module\owner\operator.py", line 59, in execute
    return IfcStore.execute_ifc_operator(self, context)
  File "C:\Users\ \AppData\Roaming\Blender Foundation\Blender\2.93\scripts\addons\blenderbim\bim\ifc.py", line 195, in execute_ifc_operator
    result = getattr(operator, "_execute")(context)
  File "C:\Users\ \AppData\Roaming\Blender Foundation\Blender\2.93\scripts\addons\blenderbim\bim\module\owner\operator.py", line 70, in _execute
    "SuffixTitles": json.loads(props.person.suffix_titles) if props.person.suffix_titles else None,
  File "C:\Program Files (x86)\Steam\steamapps\common\Blender\2.93\python\lib\json\__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "C:\Program Files (x86)\Steam\steamapps\common\Blender\2.93\python\lib\json\decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "C:\Program Files (x86)\Steam\steamapps\common\Blender\2.93\python\lib\json\decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

location: <unknown location>:-1

```

Possibly related :

Throws the exact same error on trying to validate an organisation's adress for the fields : Telephone number, facs number, or emails. Doesn't complain for the other fields.

![image](https://user-images.githubusercontent.com/25156105/127521610-5c1639c3-9b01-435f-9290-9c6c8b12814f.png)

Moult:
[...]

PR #1617, opened 2021-07-31
Improve UI for multi input string fields

Gorgious56:
Following https://github.com/IfcOpenShell/IfcOpenShell/issues/1609
This PR aims to improve UX when using multi-input fields such as Middle Names and Organisation telephone number.
Added a + operator to add an item to the current multi-input field. If there is only one left, the user can't delete it. If there are more than one, a cross appears on the field's right to delete it.

BES_28

Possible improvements :
I'd very much like some feedback on possible UX/UI/Code improvements
Limit the number of possible items per CollectionProperty
Limit the number of character per field
Force using alphanumeric characters
Force using numeric characters on telephone number

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/owner/__init__.py
+++ b/src/blenderbim/blenderbim/bim/module/owner/__init__.py
@@ -2,6 +2,7 @@ import bpy
 from . import ui, prop, operator
 
 classes = (
+    operator.AddOrRemoveElementFromCollection,
     operator.EnableEditingPerson,
     operator.DisableEditingPerson,
     operator.AddPerson,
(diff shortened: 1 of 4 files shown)


Gorgious56:
Trying out a more compact UI
BES_31

Gorgious56:
Maybe even simpler, without using box layouts
image

Gorgious56:
Using minus icon instead of cross to remove unwanted elements :
image

Gorgious56:
Bringing inputs belonging to the name property closer together using a `column` layout

MarcelP102:
Nice eye for detail!

Moult:
Looks great!

PR #1626, opened 2021-08-03
Added BCF Server

TestPrab:


Moult:
We're going to take the James Bond approach to merging this PR, and then do a code review and tweaks during our meetup :)

Moult:
In the future, by the way, it is generally preferred to "commit early, commit often". This makes the review process more digestible, and when you are shaping up a PR, you can squash commits as necessary to make it easier (search for git squash).

PR #1627, opened 2021-08-04
Update selector.py

bbrangeo:
Add new inverse_relationship ==> boundedby

If Entity IfcRelSpaceBoundary exist then hasattr(element, "BoundedBy") is True and relationship.RelatedBuildingElement exist.

New Rule Lark :  "@@ .IfcSpace[Name *= \"Bedroom\"] & .IfcSlab"


Moult:
Thank you so much!

Issue #1632, opened 2021-08-07
BlenderBIM AttributeError (without any using API related with ifc/bim)


koji-dan:
Hi, i always appreciate for you great works !

What i want to try is just to import/export ifc file by python console of blender, and i installed BlenderBIM addon into blender.
After installing BlenderBIM, i always meet AttributeError on python script as the followings.
this message is always showing on starting python script, when not only before import/export call but also not using ifc import/export. this might be based on the version combination issue ??

<error message>
Blender 2.93.0 (hash 84da05a8b806 built 2021-06-02 11:34:30)
Read prefs: C:\Users\r00414643\AppData\Roaming\Blender Foundation\Blender\2.93\config\userpref.blend
Exception in module register(): C:\Users\r00414643\AppData\Roaming\Blender Foundation\Blender\2.93\scripts\addons\blenderbim\__init__.py
Traceback (most recent call last):
  File "C:\Program Files\Blender Foundation\Blender 2.93\2.93\scripts\modules\addon_utils.py", line 386, in enable
    mod.register()
  File "C:\Users\r00414643\AppData\Roaming\Blender Foundation\Blender\2.93\scripts\addons\blenderbim\bim\__init__.py", line 131, in register
    module.register()
  File "C:\Users\r00414643\AppData\Roaming\Blender Foundation\Blender\2.93\scripts\addons\blenderbim\bim\module\model\__init__.py", line 34, in register
    bpy.utils.register_tool(workspace.BimTool, after={"builtin.scale_cage"}, separator=True, group=True)
  File "C:\Program Files\Blender Foundation\Blender 2.93\2.93\scripts\modules\bpy\utils\__init__.py", line 888, in register_tool
    tool_converted = tool_from_class(tool_cls)
  File "C:\Program Files\Blender Foundation\Blender 2.93\2.93\scripts\modules\bpy\utils\__init__.py", line 885, in tool_from_class
    cls._km_action_simple(kc_default, kc, context_descr, tool_def.label, keymap_data)
  File "C:\Program Files\Blender Foundation\Blender 2.93\2.93\scripts\startup\bl_ui\space_toolsystem_common.py", line 471, in _km_action_simple
    km = kc.keymaps.get(km_idname)
AttributeError: 'NoneType' object has no attribute 'keymaps'
Traceback (most recent call last):
[...]

PR #1633, opened 2021-08-07
Update __init__.py

s-leger:
Should fix issue #1632
Do not register keymaps in background mode


Moult:
Ahh seems we commited at the same time! :D 

https://github.com/IfcOpenShell/IfcOpenShell/commit/a874a6b673c646ebe7ddd8061f77eda9e07d58d2

Issue #1622, opened 2021-08-02
BlenderBIM : IFC Patch panel : Improve UI/UX for arguments field


Gorgious56:
I've stumbled upon https://community.osarch.org/discussion/447/import-and-merge-ifc
Regarding
![image](https://user-images.githubusercontent.com/25156105/127863824-b2b59906-67ab-4348-b016-1c539376fa28.png)
The UI is pretty straightforward apart from the `Arguments` field which requires the user to manually input a file path with square brackets and quotations marks, and I assume needs to escape \  or use /. (converted afterwards using json deserializer)
`["c://path/to/file.ifc"]` 

- Might be better UX wise to instead provide the user with a file directory viewer (like the 2 other fields) to look for their file and take care of escaping characters and adding the extra brackets and quotations backstage
- The field is plural, meaning there might be multiple arguments ? In this case could be something akin to this solution for multi input string fields :
![image](https://user-images.githubusercontent.com/25156105/127770128-8c9d0b49-ef2e-475a-a71e-c64752ff0607.png)

I'd like to help on this topic if you think it's relevant to tackle this now.

Moult:
This is a great idea! By the way, I've sent you an invite for write access so you can commit directly instead of PRs. Feel free to commit throughout the Python portion of the code. If you are going to commit to the C++ portion, submit a PR so that @aothms can review first :)

So how IfcPatch works is that you can distribute a small contained IfcOpenShell script to manipulate a file and it's a very minimal wrapper that provides a standardised way to run, input a file, and args, and get a file out. This means that some IfcPatch scripts (which I've been calling recipes) have zero arguments, and some have quite a few.

Therefore, creating a JSON parsed args input was the lazy way to support it, but not very user friendly.

The more user friendly approach I would propose is to:

 1. Add docstrings and type hints to the IfcPatch recipes themselves so that arguments are well documented. Here's an example of docstrings and type hints used in a similar manner for the IfcOpenShell API: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcopenshell-python/ifcopenshell/api/project/create_file.py
 2. Write a function that parses these docstrings and type hints, if they exist. Here's similar code which extracts docs for the IfcOpenShell API code, which you can massage for this purpose: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcopenshell-python/ifcopenshell/api/__init__.py#L102-L149
[...]

PR #1655, opened 2021-08-15
Improve ifc patch arguments User Experience

Gorgious56:
See https://github.com/IfcOpenShell/IfcOpenShell/issues/1622

BES_32


src/blenderbim/blenderbim/bim/module/patch/operator.py
```diff
@@ -39,22 +41,55 @@ class ExecuteIfcPatch(bpy.types.Operator):
...
     file_format: bpy.props.StringProperty()
+    use_json_for_args: bpy.props.BoolProperty()
 
     @classmethod
     def poll(cls, context):
         input_file = context.scene.BIMPatchProperties.ifc_patch_input
         return os.path.isfile(input_file) and "ifc" in os.path.splitext(input_file)[1]
 
     def execute(self, context):
-        import ifcpatch
```
Moult, line 49: Lazy import because IfcPatch may pull in nasty dependencies which the user may not have.

src/blenderbim/blenderbim/bim/module/pset/ui.py
```diff
@@ -67,18 +68,8 @@ def draw_psetqto_ui(context, pset_id, pset, props, layout, obj_type):
...
-        row.prop(prop, "string_value", text=prop.name)
-    elif prop.data_type == "integer":
-        row.prop(prop, "int_value", text=prop.name)
-    elif prop.data_type == "float":
-        row.prop(prop, "float_value", text=prop.name)
-    elif prop.data_type == "boolean":
-        row.prop(prop, "bool_value", text=prop.name)
-    elif prop.data_type == "enum":
-        row.prop(prop, "enum_value", text=prop.name)
-    row.prop(prop, "is_null", icon="RADIOBUT_OFF" if prop.is_null else "RADIOBUT_ON", text="")
```
Moult, line 81: This is an exception, where notice how there is no if statement checking whether or not the prop is_optional before displaying the is_null toggle. I've fixed this and will commit.

src/blenderbim/blenderbim/bim/module/patch/operator.py
Gorgious56: OK, we only want to load it if the user is going to actually use the functionality, right ?

src/blenderbim/blenderbim/bim/module/pset/ui.py
Gorgious56: Alright, sorry for that, I figured there would be some exceptions somewhere, I'm glad you took the time to review it. :) Don't hesitate to ping me if you stumble upon similar tasks you think I could handle.

src/blenderbim/blenderbim/bim/module/patch/operator.py
Moult: I think so, yes. This is the same for all of the auxiliary utilities (IfcClash, IfcCSV, etc)

PR #1662, opened 2021-08-16
Make BCF project UI (more) foolproof

Gorgious56:
I've tried to make the BCF Project UI more foolproof.
Summary :
- Use `getBcfViewpoints` method to prevent the infamous empty dynamic enum throwing out cryptic error messages dozens of times per seconds.
- Use Poll method to prevent clicking on the operators at all if the conditions are not met
- Use `CollectionProperty.clear()` instead of `while...remove(0)` loop.
- Use `enumerate` to get incremented index in `for` loop
- Use more descriptive variables to prevent batch-renaming mistakes
- Remove L.256 & 257 since the next step in the operator is to clear the collection
- Various code fixes

Illustration :
BES_33

To go further :

- Is there really a use to storing all the displayed properties like `reference_link`, `related_topic`, `label`, etc on each and every topic ? They get overwritten and deleted on every collection update anyway, and only one of them is shown at any one time. It would make more sense to me to store them directly on the `BCFProperties` class like the Author and Project Name properties.
- It might be interesting to display a dynamic enum in place of the `related_topic` string field. This would prevent user from having to spellcheck the name, and allow topics with the same name to be chosen independently. Unless we don't want different topics with the same name ? In that case we should prevent that too ><

Food for thought :)

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/bcf/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/bcf/operator.py
@@ -7,6 +7,7 @@ import numpy as np
 import ifcopenshell
 import ifcopenshell.util.unit
 from . import bcfstore
+from blenderbim.bim.module.bcf.prop import getBcfViewpoints
 from blenderbim.bim.ifc import IfcStore
 from math import radians, degrees, atan, tan, cos, sin
 from mathutils import Vector, Matrix, Euler, geometry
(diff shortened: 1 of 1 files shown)


Moult:
Great work, and agree on both of your points on going further.

Personally, I reckon a system which just displays labels and doesn't store anything in Blender until the user hits a "edit" button, just like in IFC, would be much more efficient, but I never got around to building this for the BCF module.

> Use CollectionProperty.clear() instead of while...remove(0) loop.

Yes, in fact I'm an idiot and never used that :D Check out `cd /path/to/blenderbim && grep -ir "while len(" ./` and uh... maybe we should clean that up :D

PR #1665, opened 2021-08-17
Improve drawing style save and load

Gorgious56:
This one deserves a little bit of backstory :)

While looking for places where `bpy.context` was used instead of `context` when available, I stumbled upon the Drawing Style Save & Load mechanic (https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/blenderbim/blenderbim/bim/module/drawing/operator.py). Initialy I didn't understand how it all worked so I didn't touch anything in the module by fear of horribly breaking everything.

After wetting my feet elsewhere in the codebase, I decided to dig a bit further and I managed to piece everything together (not that it was particularly difficult, but it was a bit daunting at first). The different drawing properties are serialized when the user chooses to save a new drawing style, and then un-serialized when an existing drawing type is chosen to be made current. 

However the current implementation leads to the use of magic strings and code duplication. I thought the `context` refactor warranted a little optimization exercise and that it was a good candidate for the `enum.Enum` class.

How it works :

The Raster Style properties are defined in one and only one place, which will be accessed by whichever other module needs to.
The enum members will describe the property they relate to, and their value is a string used in the serialization process to get or set the value.

```
class RasterStyleProperty(enum.Enum):
    WORLD_COLOR = "bpy.data.worlds[0].color"
    RENDER_ENGINE = "scene.render.engine"
    RENDER_TRANSPARENT = "scene.render.film_transparent"
    etc.
```

Importing the enum in related modules :    
    `from blenderbim.bim.module.drawing.prop import RasterStyleProperty`

Setting default values "by hand" :
```
        drawing_style.raster_style = json.dumps(
            {
                RasterStyleProperty.WORLD_COLOR.value: (1, 1, 1),
                RasterStyleProperty.RENDER_ENGINE.value: "BLENDER_WORKBENCH",
                RasterStyleProperty.RENDER_TRANSPARENT.value: False,
            etc.
```

The nice thing is we can loop over the Enum members to initialize the properties with current values, instead of manually setting them :
```
        space = self.get_view_3d(context)
        scene = context.scene
        style = {}
        for prop in DrawingStyleProp:
            value = eval(prop.value)
            if not isinstance(value, str):
                try:
                    value = tuple(value)
                except TypeError:
                    pass
            style[prop.value] = value
```

And the deserialization is also straightforward, since the dictionary keys are the actual path to the property :

```
        scene = context.scene
        space = self.get_view_3d(context)
        style = json.loads(self.drawing_style.raster_style)
        for path, value in style.items():
            if isinstance(value, str):
                exec(f"{path} = '{value}'")
            else:
                exec(f"{path} = {value}")
```

The scalability is quite manageable, since adding a new raster style property is as easy as adding a new enum member, and specifiying its path.


Moult:
Makes sense!

Issue #1625, opened 2021-08-03
Migrate away from Travis CI


aothms:
Re https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing we had been been using their service quite generously, but our commit rate would make paying for the product quite expensive. We already have Github actions in place, we just need one for Pull Requests.

Krande:
If you haven't started on this already, I can give this a shot. What functionality specifically do you want for each pull request?

aothms:
That'd be really great! Most important is whether the code compiles. But
Travis did a bit more and ran some checks. Maybe take the Travis yml as a
starting point. Let me know if you have questions

Sent from a mobile device. Excuse my brevity. Kind regards, Thomas

Op ma 9 aug. 2021 09:23 schreef Kristoffer Andersen <
***@***.***>:

Krande:
Sounds good!

I guess we can either use the default ubuntu version of github action and install the prerequisites (which I believe is done for the [release.yml](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/.github/workflows/release.yml) workflow. 

OR we can create a docker image with all the necessary prerequisites pre-installed and do the compilation and tests inside the container. 

My gut feeling is that the latter option would be faster (depending on what we can do with caching each step). And personally I would love to have docker images with the latest version of ifcopenshell for local testing AND for compilation of daily versions.

Do we have a "latest" docker image for the pre-requisites of ifcopenshell hosted somewhere that I can pull into the github action workflow? If not maybe we should also include a github action workflow to build\update the pre-requisite docker images and maybe host them on github? 

let me know what you think! In the mean time I can start experimenting with building a docker image based on the `travis.yml` file.

aothms:
> use the default ubuntu version of github action and install the prerequisites 

> OR we can create a docker image 

I don't have a strong preference, but thanks to @plaes we have ccache enabled in the ci action which results in quite a massive speedup as most of the commits are python files anyway that don't incur any changes in compilation. https://github.com/IfcOpenShell/IfcOpenShell/commit/184a189f2a8dd4b881876c6fbeb1e5607db99185 That is potentially harder in Docker?

> Do we have a "latest" docker image for the pre-requisites 

Another issue we have is that I think ci action still depends on the old OCE package in Ubuntu. That's OCCT 6.9 or something which means a lot of new additions in OCCT are not available (related to booleans and layersets in particular).

Krande:
[...]

PR #1666, opened 2021-08-17
A proposal for CI workflow on PR and commits

Krande:
As promised in https://github.com/IfcOpenShell/IfcOpenShell/issues/1625

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,107 @@
+name: CD
+
+on: 
+  push:
+
+
+jobs:
+
+  activate:
+    runs-on: ubuntu-latest
+    if: |
+      github.repository == 'IfcOpenShell/IfcOpenShell' &&
+      !startsWith(github.event.head_commit.message, 'Release ') &&
+      !contains(github.event.head_commit.message, 'ci skip')
+    steps:
+    - run: echo ok go
+
+  build:
+    runs-on: ubuntu-20.04
+    needs: activate
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: |
+        sudo apt update
+        sudo apt-get install --no-install-recommends \
+        git cmake gcc g++ libboost1.67-all-dev python3-all-dev swig libpcre3-dev libxml2-dev \
+        liboce-foundation-dev liboce-modeling-dev liboce-ocaf-dev liboce-visualization-dev liboce-ocaf-lite-dev
+
+    -
+      name: ccache
+      uses: hendrikmuhs/ccache-action@v1
+
+    - 
+      name: Build ifcopenshell
+      run: |
+        mkdir build && cd build
+        cmake \
+         -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+         -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+         -DCMAKE_INSTALL_PREFIX=$PWD/install/ \
+         -DCMAKE_BUILD_TYPE=Release \
+         -DCMAKE_PREFIX_PATH=/usr \
+         -DCMAKE_SYSTEM_PREFIX_PATH=/usr \
+         -DBUILD_PACKAGE=On \
+         -DOCC_INCLUDE_DIR=/usr/include/oce \
+         -DOCC_LIBRARY_DIR=/usr/lib/x86_64-linux-gnu \
+         -DPYTHON_EXECUTABLE:FILEPATH=/usr/bin/python3 \
+         -DPYTHON_INCLUDE_DIR:PATH=/usr/include/python3.8 \
+         -DPYTHON_LIBRARY:FILEPATH=/usr/lib/x86_64-linux-gnu/libpython3.8.so \
+         -DCOLLADA_SUPPORT=Off \
+         -DLIBXML2_INCLUDE_DIR=/usr/include/libxml2 \
+         -DLIBXML2_LIBRARIES=/usr/lib/x86_64-linux-gnu/libxml2.so \
+         ../cmake
+        make -j $(nproc)
+        make install
+    - 
+      name: Package
+      run: |
+        make package
+      working-directory: build
+    - name: Upload 
+      uses: actions/upload-artifact@v2
+      with:
+        # Artifact name
+        name: ifcos-artifacts
+        # Directory containing files to upload
+        path: build/assets/Ifc*
+
+  deliver:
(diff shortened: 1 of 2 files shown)


Krande:
Ping @aothms (in case this slipped under your radar) :)

aothms:
Sorry yeah I was just waiting for the other PR. I reran and it failed on missing numpy which was now added to the same ids test. Maybe it needs to be made optional, but for now let's just add it to the deps here.

https://github.com/IfcOpenShell/IfcOpenShell/pull/1666/checks?check_run_id=3410459293#step:7:12

Maybe consolidate the `install xmlschema` section we have now into the `Install dependencies` and make it `pip install xmlschema numpy`?

Krande:
Okay, one step closer :) 

I see we are getting a `ModuleNotFoundError: No module named 'bcf'`. 

```
  File "../src/ifcopenshell-python/ifcopenshell/test_ids.py", line 2, in <module>
    import ids
  File "/home/runner/work/IfcOpenShell/IfcOpenShell/src/ifcopenshell-python/ifcopenshell/ids.py", line 30, in <module>
    from bcf.v2.bcfxml import BcfXml
ModuleNotFoundError: No module named 'bcf'
```

https://github.com/IfcOpenShell/IfcOpenShell/pull/1666/checks?check_run_id=3410778317#step:7:12

Do we have a plan to include this with ifcopenshell-python or should we make a separate python package for it?

Moult:
I don't think bcf should be included. It should perhaps be an optional dependency. I also wonder if this is the responsibility of the IDS library (i.e. the test parser / creator), or BIMTester (i.e. the test runner / reporter)? Or maybe BIMTester should be removed? Or IDS merged into BIMTester? Dunno. Many options :)

aothms:
@Moult There is a test case for IDS that emits BCF. So on the CI runner `bcf` definitely needs to be installed.

I had the same questions though. We had two parallel GSoC projects so there wasn't a lot of room to figure out what goes where in between of all the moving pieces. 

@Moult What are the plans for publishing bcf? For now should we just get a Github Action in place that pushes bcf to pypi? Should it be moved to a separate repo? I remember we discussed this earlier, but don't remember the conclusion. If it lives in the IfcOpenShell/IfcOpenShell repo I think it makes sense to import it as ifcopenshell.bcf and not bcf to prevent confusion. Also bcf already exists on pypi.

Krande:
I can make a CI for publishing bcf to pypi as bcf-py (it would still be imported as bcf when installed using pip). 
Or I can make a conda package out of it (we would just have to find a reasonable channel for it (-c ifcopenshell or -c conda-forge, etc..) and add that to the conda-build requirements). 

In any case it would need a setup.py (or .cfg) file containing information about the package. For now I can make a "nearly blank" setup file with author="ifcopenshell" and just the bare minimum to get it working. 

Your thoughts?

Moult:
Ah I didn't realise, yes true it needs to be installed then.

As for where it should live, I don't know. Out of the growing list at the bottom of [the readme](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/README.md#extra-tools), bcf, bsdd, and opencde (now incorrectly named foundationserver/bcfserver) do not have a strong dependency on ifcopenshell. bsdd has an optional dependency (which may grow), but bcf has no dependency. However, inversely blenderbim, ifcclash, and ifcbimtester depend on bcf (and potentially ifcsverchok and ifcfm in the future).

Another option is another repo also under the IfcOpenShell umbrella like ifcopenshell/bcf, ifcopenshell/bsdd, ifcopenshell/opencde, etc.

@aothms I guess it's merely a stylistic thing. You choose what you'd prefer, whether to move or not, and if so, where it should live. Maybe it doesn't need to move at all, and can be published as bcf-py (or bcf-client? not fussed) as @krande suggests.

aothms:
If we push to pypi I guess we need versions/tags. From the point of view if makes sense to move it out.

To me a stronger link between bcf and ifcopenshell-python is not unthinkable. Camera positions from placement or geometry. Something related to annotations. Screenshots. Etc. From that point of view we can also move bcf/ to ifcopenshell-python/ifcopenshell?

Krande:
Do we expect any major changes in the bcf library itself? And does bcf depend on anything but itself (and some general dependencies such as xmlschema and numpy)?

If both are no, then maybe it could make sense publishing this as a standalone package and add it to the runtime-requirements of ifcopenshell?

Personally I would prefer bcf as a lightweight standalone package for accessing cameras and annotations (which is something I would like to use in separate downstream applications that does not necessarily use ifcopenshell). The actual repository location for the bcf package is not a big issue to me. 

Anyways, I leave the final decision up to you @aothms and @Moult  :)

aothms:
I leave it to @Moult :)

Sent from a mobile device. Excuse my brevity. Kind regards, Thomas

Op di 24 aug. 2021 17:06 schreef Kristoffer Andersen <
***@***.***>:

Moult:
I'd say leave it where it is in `src/bcf` with the logic that:

 1. the ifcopenshell monorepo already has a bunch of loosely interdependant stuff - and so people already expect to find the suite of tools in this repo. If we split out bcf to another repo, to be consistent other things should be split out too.
 2. it's under active development. still to do is bcf xml version 2.0, bcf api version 2.1, a series of updates to support documents, more oauth methods, and maybe even a lightweight CLI/GUI. During this development, it will likely see correlated changes in ifcclash, bimtester, and blenderbim at least.
 3. Right now, there are no dependencies to ifcopenshell, which suggests that it may be separately distributed. I didn't see a need for repo-level versions/tags reading some guides on publishing to pypi, but I could be wrong. Happy for krande to go ahead and publish any way he sees fit :)

In the future, when there is a tighter integration with ifcopenshell and bcf, nothing is set in stone :) If a tighter integration emerges, then I would support merging with ifcopenshell-python. If not, it can stay separate. If ifcopenshell switches from a monorepo to split repos, then it can move out along with the rest.

Krande:
Sounds good. 

For the main CI in this PR I will just do a local install using `pip install .` before running the tests.

I will then add a separate CI for building "bcf-client" whenever it detects a change in src/bcf/*. This would also require a version bump in the `src/bcf/setup.cfg` for it to be allowed to be uploaded to pypi (I don't have much experience with automating versioning using ci).

BUT I would need to add (or use an existing) pypi token in the IfcOpenshell repo for securely uploading to pypi. Do we have any existing secrets env variables for pypi or should I add one myself (which I believe at the moment I do not have the privileges to do)?

PR #1668, opened 2021-08-18
BlenderBim : Improve BCF Project UX

Gorgious56:
Continuation of https://github.com/IfcOpenShell/IfcOpenShell/pull/1662

Summary :
- Store field props on the BcfProperties class instead of individual topics
- Clear fields when changing active topic
- Use dynamic enum to choose related topic

BSE_37

Further questions / Possible enhancements :
- If topic A is in topic B's `related_topics`, does topic B need to be in topic A's `related_topics` ? (Sorry I don't know much about BCF...)
- It's not immediately apparent that clicking on a related topic's name switches the current topic to it. There may be a way to make it more obvious ?
- Is it OK to allow several topics to share the same title ? Since we use the (unique) guid to save/retrieve them, I don't see anything that should prevent it.
- The display order of the topics is semi-random, it depends on the way topics are spitted out by `bcfxml.topics`. Usually when adding elements to a list, the user expects the new element to be placed either last or first in the list. It's currently not always the case.
- (Related) Is it worth adding operators to move topics up / down in the list ?
- (Related) Is it worth implementing a way to ensure new topics are placed last in the list or alternatively, automatically order them by their `date` property ?
- There is no way to remove a topic once it's been created. Is it by design ?


Moult:
1. Nope
2. Feel free :)
3. Yes, they may share the same title.
4. Fixing randomness would be good. Maybe in the underlying BCF lib we need a way to fetch a sorted list.
5. Don't think so.
6. See 4.
7. Just an omission, feel free to build :) This is the thing you need to call: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/bcf/bcf/v2/bcfxml.py#L341

By the way, not sure about a dynamic enum to choose related topic. There may be easily thousands of topics. Not sure how Blender handles a dropdown that large.

Gorgious56:
Thanks for the feedback ! :)
I'll starting working on that.

About the dynamic enum, your concerns are very valid, it will most likely hiccup blender for a second or two, but most importantly, dynamic enum display does not scale well with many elements. It's not possible to use a search bar either like with the UI List widget, which makes it useless.
However I think we'll be in trouble anyways because of the way topics are displayed in the top UI List, which is not very well suited either for thousands of items... Maybe a tree like structure would be more efficient, but afaict there is no clear hierarchy between BCF topics. Is there a way it is handled in other softs we could take inspiration from ? Maybe with the new [Spreadsheet Editor](https://docs.blender.org/manual/en/latest/editors/spreadsheet.html) capabilities we'll see an improvement on that front :)

In the meantime I can revert it back if you want.

Moult:
@Gorgious56 I have successfully used the UIList with thousands of items for scheduling and costing. I reckon it could be used similarly for BCF, but perhaps with pagination to limit it to X items (1k?) at a time. However it is quite rare so I think pagination is a very low priority. Just revert that one changes to the dropdown.

PR #1680, opened 2021-08-22
Minor iterator optimisation to prevent inverse check for associated materials

Moult:
You said `if (products->size() > 1) { return true } ` but it should be `== 1` right?


aothms:
Exactly. Thanks. Exciting... your first C++ contribution!

PR #1683, opened 2021-08-22
Various tweaks and fixes

Gorgious56:
I've done a sanity pass on the interface of the BlenderBim N-Panel and the drawing panel.
I grouped some unrelated small tweaks to the code, mainly :

- Use poll method instead of early return in operators when possible
- Small UI tweaks
- Add property accessors to get current active item in some collections properties, to simplify code in operators
- Mesh calculations in Qto can now be used in edit mode accurately with bmesh

As always I'm willing to discuss and tweak any of these changes if need be :)


Moult:
Amazing work!

Issue #1684, opened 2021-08-23
LENGTHUNIT conversion does not work on project libraries


Moult:
Related to #1247, I recently ran into this issue, caused by this assumption: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcparse/IfcParse.cpp#L2249

Naturally, converting length units from project libraries is super useful so that a project library in meters won't appear tiny in a project in millimeters :) and will also address some problems already found by @theoryshaw who uses imperial

Before I attempt to hack away some C++, I guess I'd like to confirm the approach. In theory there may be _multiple_ IfcProjectLibraries (see diagram in docs for IfcTaskType). So perhaps it should first check for an IfcProject, then if it is > IFC2x3 (do you need this check in C++?) then it'll find a IfcProjectLibrary (or, an IfcContext?) as a fallback. (note multiple IfcProjectLibrary seems to only occur when you have an IfcProject first - not in a standalone library, where it serves as a "root/entry" entity). Does this approach sound sane?

aothms:
```
try by_name(IfcContext)
catch by_name(IfcProject)
```

is the most succinct I can think of. I'm fine with any reasonable alternative.

Or if you want to be sure of the multiple libraries thing, then we need to traverse to the correct project but that seems prohibitively hard for an arbitrary IFC instance.

Moult:
Closing as fixed!

PR #1687, opened 2021-08-24
Detect units from project libraries, and convert nested double aggregates

Moult:
See #1684

I opted to check project first, then only do the try catch for project library because there are two scenarios:

 1. A regular IfcProject which may have zero or more IfcProjectLibraries. In this case, the check for IfcProject passes and we're done.
 2. An IfcProjectLibrary, which there is only one allowed. In this case, the check for IfcProject fails, and we do the try/catch for IfcContext.

In addition, I found that length unit conversion didn't seem to work for cartesianpointlist3d, which I use in tessellations. I hope my fix makes sense.

There seem to be multiple potentially more concise ways of writing the same thing, like:

```c++
for (std::vector<double>& v2 : v) {  for (double& elem : v2) {  elem *= conversion_factor; } }
```

... or ...

```c++
for (auto&& v2 : v) {  for (auto&& elem : v2) {  elem *= conversion_factor;  }  }
```

... but I opted to follow the coding style already present, but please correct me if you have a preference or if there is a pro/con to each approach.

### DIFF
--- a/src/ifcparse/IfcParse.cpp
+++ b/src/ifcparse/IfcParse.cpp
@@ -1747,6 +1747,18 @@ IfcUtil::IfcBaseClass* IfcFile::addEntity(IfcUtil::IfcBaseClass* entity, int id)
 					IfcWrite::IfcWriteArgument* copy = new IfcWrite::IfcWriteArgument();
 					copy->set(v);
 					we->setArgument(i, copy);
+				} else if (attr_type == IfcUtil::Argument_AGGREGATE_OF_AGGREGATE_OF_DOUBLE) {
+					std::vector<std::vector<double>> v = *attr;
+					for (std::vector<std::vector<double>>::iterator it = v.begin(); it != v.end(); ++it) {
+						std::vector<double>& v2 = (*it);
+						for (std::vector<double>::iterator jt = v2.begin(); jt != v2.end(); ++jt) {
+							(*jt) *= conversion_factor;
+						}
+					}
+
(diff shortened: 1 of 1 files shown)


aothms:
Wonderful!

> but please correct me if you have a preference or if there is a pro/con to each approach.

No it's just that these range based for and auto constructs are introduced in C++11 and the majority of the code is written before that was mainstream enough. They are equivalent.

How big do you see the issue that we only do this for length measure?

Moult:
Yes, I will have a go at implementing non-length measures, though it could be trickier, and although I didn't double check, upon a first read through at https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcparse/IfcParse.cpp#L2292-L2309 it seems as though the conversion based units only consider conversion from an IfcSIUnit, not any arbitrary IfcUnit.

I think it's an issue because project libraries will likely store quantity data which needs to be converted. However, I haven't actually used it yet, so it's not a high priority now, but sooner or later I will need it and I'm happy to have a go :)

aothms:
> it seems as though the conversion based units only consider conversion from an IfcSIUnit

Indeed, I think this may have been better in 0.5 but this has been rewritten in 0.6 using the latebound code which is a pain. This should be recursive I guess.

That said:

```
*unit->data().getArgument(
    unit->declaration().as_entity()->attribute_index("ConversionFactor")
);
```

Can be rewritten as `*unit->get("ConversionFactor")` these days.

Issue #1685, opened 2021-08-23
BlenderBim : Batch-adding Objects to a system and system inspection in the Object properties


Gorgious56:
Currently the operator used to assign an object to a system only does so to the active object. 
![image](https://user-images.githubusercontent.com/25156105/130433133-1769bf35-60a1-4591-877d-f8659ae9aad1.png)

It may be desirable to add a big number of objects to a system, which is cumbersome with this behaviour. Is it worth tweaking the operator to add every selected object or was it designed that way on purpose ?

Secondly, it may be nice to have a panel in the object's properties to see which systems it belongs to and similar interface to assign/unassign them. That way we a many/many interface.

I can work on it if you feel like it's relevant.

Gorgious56:
I think the same question applies to IFC Groups :
![image](https://user-images.githubusercontent.com/25156105/130440445-1f8726cb-9596-41dd-9653-91d4e864b2fb.png)

Moult:
Yes, definitely worth tweaking the operators for both, as well as adding a panel in the object properties. Go for it :)

Gorgious56:
Nice ! Given a blender object, what's the most graceful way to retrieve all the systems or groups it belongs to ? I'm a little lost on the ifcopenshell api.

Gorgious56:
During refactoring I arrived to a slightly different approach. I removed the possibility to assign objects to systems from the scene properties interface, and created a new panel in the object properties. I think this is more in line with the other types of data that can be linked, like constraints, documents...
So the scene panel is used only for adding / removing systems, changing their properties, and selecting all objects related to the system because that makes sense.

![image](https://user-images.githubusercontent.com/25156105/130467826-27e4d8ad-71ce-4667-99c4-7385ca6e4c1d.png)

And in the object properties, 

![image](https://user-images.githubusercontent.com/25156105/130468649-76be0cd6-2e5f-4d76-b004-5e41059d1b47.png)
-->
![image](https://user-images.githubusercontent.com/25156105/130468787-902ad0ce-6006-4af7-ab40-63a711ec67d3.png)

Very similar to how it works with the Constraints and Documents panels in the object properties, however there are small caveats :
The interface is updated relatively to the active object. (That's expected I think)
[...]

PR #1688, opened 2021-08-24
Improve groups and systems interface

Gorgious56:
Following https://github.com/IfcOpenShell/IfcOpenShell/issues/1685

This PR adds 2 panels in the object properties for accessing and assigning groups and systems.
It's no longer possible to assign systems or groups to objects from the panels in the scene properties.

I tried to stay consistent with the other data containers that work like it, like constraints and documents.

However I still have a bug to fix, that can be seen at the end of the gif :
BSE_39
When duplicating an object the interface does not get updated and it shows that the new object does not have any group or system associated with it. The groups and systems do get correctly assigned to the new copy though. The interface updates when trying to add another group to the new object.

I think a Data reloading might need to be done on object duplication for the interface to be up to date but I'm not sure how to do it.


Moult:
Looks great! By the way, systems and groups will become hierarchical in the future FYI. I just haven't got around to it yet.

Feel free to commit when you're ready, and I'll investigate the object duplication data reloading bug.

Also instead of the assigngrouptomany operator, how about https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/blenderbim/blenderbim/bim/module/root/operator.py#L81

Gorgious56:
Done & Done !
Also learnt how to squash merge commits :)

Thanks for the feedback.

PR #1696, opened 2021-08-25
Apply object operators to all selected objects

Gorgious56:
This change adds functionality to extend some operators to affect all selected objects rather than only the active one :

- Remove container
- Assign / Unassign Material
- Add Representation (silently passes on non-mesh objects)
- Assign / Unassign Constraint
- Assign / Unassign Document
- Small tweaks to attribute assignments which `data_type` wasn't updated when created
- Use poll method to `bim.add_classification` operator to avoid error if no classification is available

The PR is showing a lot more line additions and deletions than in reality mainly because I changed the order of some lines that didn't need retrieving on each iteration like `self.file = IfcStore.get_file()`.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/georeference/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/georeference/operator.py
@@ -49,14 +49,7 @@ class EnableEditingGeoreferencing(bpy.types.Operator):
             new.is_null = Data.projected_crs[attribute.name()] is None
             new.is_optional = attribute.optional()
             new.data_type = data_type
-            if data_type == "string":
-                new.string_value = "" if new.is_null else Data.projected_crs[attribute.name()]
-            elif data_type == "float":
-                new.float_value = 0.0 if new.is_null else Data.projected_crs[attribute.name()]
-            elif data_type == "integer":
-                new.int_value = 0 if new.is_null else Data.projected_crs[attribute.name()]
-            elif data_type == "boolean":
-                new.bool_value = False if new.is_null else Data.projected_crs[attribute.name()]
+            new.set_value(new.get_value_default() if new.is_null else Data.projected_crs[attribute.name()])
 
         props.is_map_unit_null = Data.projected_crs["MapUnit"] is None
         if not props.is_map_unit_null:
@@ -79,8 +72,7 @@ class EnableEditingGeoreferencing(bpy.types.Operator):
             new.is_null = Data.map_conversion[attribute.name()] is None
             new.is_optional = attribute.optional()
             # Enforce a string data type to prevent data loss in single-precision Blender props
-            new.data_type = "string"
-            new.string_value = "" if new.is_null else str(Data.map_conversion[attribute.name()])
+            new.set_value("" if new.is_null else str(Data.map_conversion[attribute.name()]))
 
         props.has_true_north = bool(Data.true_north)
         if Data.true_north:
--- a/src/blenderbim/blenderbim/bim/module/group/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/group/operator.py
@@ -130,7 +130,7 @@ class EnableEditingGroup(bpy.types.Operator):
             new.name = attribute.name()
             new.is_null = data[attribute.name()] is None
             new.is_optional = attribute.optional()
-            new.string_value = "" if new.is_null else data[attribute.name()]
+            new.set_value("" if new.is_null else data[attribute.name()])
         props.active_group_id = self.group
         return {"FINISHED"}
 
--- a/src/blenderbim/blenderbim/bim/module/layer/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/layer/operator.py
@@ -73,7 +73,7 @@ class EnableEditingLayer(bpy.types.Operator):
             new.name = attribute.name()
             new.is_null = data[attribute.name()] is None
             new.is_optional = attribute.optional()
-            new.string_value = "" if new.is_null else data[attribute.name()]
+            new.set_value("" if new.is_null else data[attribute.name()])
         props.active_layer_id = self.layer
         return {"FINISHED"}
 
--- a/src/blenderbim/blenderbim/bim/module/material/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/material/operator.py
@@ -124,21 +124,24 @@ class AssignMaterial(bpy.types.Operator):
         return IfcStore.execute_ifc_operator(self, context)
 
     def _execute(self, context):
-        obj = bpy.data.objects.get(self.obj) if self.obj else context.active_object
-        material_type = self.material_type or obj.BIMObjectMaterialProperties.material_type
         self.file = IfcStore.get_file()
-        element = self.file.by_id(obj.BIMObjectProperties.ifc_definition_id)
-        ifcopenshell.api.run(
-            "material.assign_material",
-            self.file,
-            **{
-                "product": element,
-                "type": material_type,
-                "material": self.file.by_id(int(obj.BIMObjectMaterialProperties.material)),
-            },
-        )
-        Data.load(IfcStore.get_file())
-        Data.load(IfcStore.get_file(), obj.BIMObjectProperties.ifc_definition_id)
+        objs = [bpy.data.objects.get(self.obj)] if self.obj else context.selected_objects
+        for obj in objs:            
+            if not obj.BIMObjectProperties.ifc_definition_id:
+                continue
+            material_type = self.material_type or obj.BIMObjectMaterialProperties.material_type
+            element = self.file.by_id(obj.BIMObjectProperties.ifc_definition_id)
+            ifcopenshell.api.run(
+                "material.assign_material",
+                self.file,
+                **{
+                    "product": element,
+                    "type": material_type,
+                    "material": self.file.by_id(int(obj.BIMObjectMaterialProperties.material)),
+                },
+            )
+            Data.load(IfcStore.get_file())
+            Data.load(IfcStore.get_file(), obj.BIMObjectProperties.ifc_definition_id)
         return {"FINISHED"}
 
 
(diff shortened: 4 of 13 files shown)


Moult:
Can you please revert all changes to `src/blenderbim/blenderbim/bim/module/geometry/operator.py `? Geometry operations are pretty special and complex and only a few should occur in bulk.

src/blenderbim/blenderbim/bim/module/georeference/operator.py
```diff
@@ -49,14 +49,7 @@ def execute(self, context):
...
             new.data_type = data_type
-            if data_type == "string":
-                new.string_value = "" if new.is_null else Data.projected_crs[attribute.name()]
-            elif data_type == "float":
-                new.float_value = 0.0 if new.is_null else Data.projected_crs[attribute.name()]
-            elif data_type == "integer":
-                new.int_value = 0 if new.is_null else Data.projected_crs[attribute.name()]
-            elif data_type == "boolean":
-                new.bool_value = False if new.is_null else Data.projected_crs[attribute.name()]
+            new.set_value(new.get_value_default() if new.is_null else Data.projected_crs[attribute.name()])
```
Moult, line 52: I'd rather revert this change and rewrite it using the `import_attributes` helper. However, you will need a callback to handle the units. In addition, the units code needs to be rewritten to pull from available named units (now that we have a new units system) instead of the hardcoded bunch that we have right now. Would you be interested in tackling this? It would be a good intro to the IFC data side of things.

src/blenderbim/blenderbim/bim/module/georeference/operator.py
```diff
@@ -79,8 +72,7 @@ def execute(self, context):
             new.is_null = Data.map_conversion[attribute.name()] is None
             new.is_optional = attribute.optional()
             # Enforce a string data type to prevent data loss in single-precision Blender props
-            new.data_type = "string"
-            new.string_value = "" if new.is_null else str(Data.map_conversion[attribute.name()])
+            new.set_value("" if new.is_null else str(Data.map_conversion[attribute.name()]))
```
Moult, line 75: See above comment - I would prefer to use `import_attributes`. Also triple check that the data type of string is enforced even though the detected data types are floats, since it is vital to prevent data loss.

src/blenderbim/blenderbim/bim/module/group/operator.py
```diff
@@ -130,7 +130,7 @@ def execute(self, context):
             new.name = attribute.name()
             new.is_null = data[attribute.name()] is None
             new.is_optional = attribute.optional()
-            new.string_value = "" if new.is_null else data[attribute.name()]
+            new.set_value("" if new.is_null else data[attribute.name()])
```
Moult, line 133: Same, use `import_attributes` - nothing special here, it can work without any special callbacks.

src/blenderbim/blenderbim/bim/module/layer/operator.py
```diff
@@ -73,7 +73,7 @@ def execute(self, context):
             new.name = attribute.name()
             new.is_null = data[attribute.name()] is None
             new.is_optional = attribute.optional()
-            new.string_value = "" if new.is_null else data[attribute.name()]
+            new.set_value("" if new.is_null else data[attribute.name()])
```
Moult, line 76: Use `import_attributes`

src/blenderbim/blenderbim/bim/module/material/operator.py
```diff
@@ -124,21 +124,26 @@ def execute(self, context):
...
+                "material.assign_material",
+                self.file,
+                **{
+                    "product": self.file.by_id(obj_id),
+                    "type": material_type,
+                    "material": material,
+                },
+            )
+            Data.load(self.file)
+            Data.load(self.file, obj_id)
```
Moult, line 146: Revert - materials, like geometry is complex. Materials can actually be used to parametrically define geometry (e.g. if a material has layers such as for a wall or slab, it auto updates prism-like geometries of objects. I'd rather not make this change and approach it slowly and carefully.

src/blenderbim/blenderbim/bim/module/material/operator.py
```diff
@@ -152,13 +157,16 @@ def execute(self, context):
...
-        )
+        objs = [bpy.data.objects.get(self.obj)] if self.obj else context.selected_objects
+        for obj in objs:                       
+            if not obj.BIMObjectProperties.ifc_definition_id:
+                continue
+            ifcopenshell.api.run(
+                "material.unassign_material",
+                self.file,
+                **{"product": self.file.by_id(obj.BIMObjectProperties.ifc_definition_id)},
+            )
```
Moult, line 169: Same - please revert, sorry :(

src/blenderbim/blenderbim/bim/module/material/operator.py
```diff
@@ -436,7 +444,7 @@ def execute(self, context):
                 new = props.material_set_attributes.add()
                 new.name = attribute.name()
                 new.is_null = material_set_data[attribute.name()] is None
-                new.string_value = "" if new.is_null else material_set_data[attribute.name()]
+                new.set_value("" if new.is_null else material_set_data[attribute.name()])
```
Moult, line 447: This can use `import_attributes`

src/blenderbim/blenderbim/bim/module/material/operator.py
```diff
@@ -498,26 +506,25 @@ def _execute(self, context):
...
+            Data.load(self.file, obj_id)
             bpy.ops.bim.disable_editing_assigned_material(obj=obj.name)
             return {"FINISHED"}
 
         material_set = self.file.by_id(self.material_set)
 
-        attributes = {}
-        for attribute in props.material_set_attributes:
-            attributes[attribute.name] = None if attribute.is_null else attribute.string_value
+        attributes = {attribute.name: attribute.get_value() for attribute in props.material_set_attributes}
```
Moult, line 521: Use `export_attributes` helper

src/blenderbim/blenderbim/bim/module/material/operator.py
```diff
@@ -593,14 +600,7 @@ def load_set_item_attributes(self, material_set_item, material_set_item_data):
...
                 new.data_type = data_type
-                if data_type == "string":
-                    new.string_value = "" if new.is_null else material_set_item_data[attribute.name()]
-                elif data_type == "float":
-                    new.float_value = 0.0 if new.is_null else material_set_item_data[attribute.name()]
-                elif data_type == "integer":
-                    new.int_value = 0 if new.is_null else material_set_item_data[attribute.name()]
-                elif data_type == "boolean":
-                    new.bool_value = False if new.is_null else material_set_item_data[attribute.name()]
+                new.set_value(new.get_value_default() if new.is_null else material_set_item_data[attribute.name()])
```
Moult, line 603: `import_attributes` helper

src/blenderbim/blenderbim/bim/module/material/operator.py
```diff
@@ -621,18 +621,12 @@ def load_profile_attributes(self, material_set_item, material_set_item_data):
...
-                    new.int_value = 0 if new.is_null else profile_data[attribute.name()]
-                elif data_type == "boolean":
-                    new.bool_value = False if new.is_null else profile_data[attribute.name()]
-                elif data_type == "enum":
+                if data_type == "enum":
                     new.enum_items = json.dumps(ifcopenshell.util.attribute.get_enum_items(attribute))
                     if profile_data[attribute.name()]:
                         new.enum_value = profile_data[attribute.name()]
+                else:
+                    new.set_value(new.get_value_default() if new.is_null else profile_data[attribute.name()])
```
Moult, line 629: `import_attributes` helper, but check that it doesn't break anything. May require a callback. May not.

src/blenderbim/blenderbim/bim/module/system/operator.py
```diff
@@ -130,7 +130,7 @@ def execute(self, context):
             new.name = attribute.name()
             new.is_null = data[attribute.name()] is None
             new.is_optional = attribute.optional()
-            new.string_value = "" if new.is_null else data[attribute.name()]
+            new.set_value("" if new.is_null else data[attribute.name()])
```
Moult, line 133: `import_attributes` helper

src/blenderbim/blenderbim/bim/module/unit/operator.py
```diff
@@ -248,8 +248,7 @@ def import_attributes(self, name, prop, data, context):
             new.name = name
             new.is_null = data[name] is None
             new.is_optional = False
-            new.data_type = "string"
-            new.string_value = json.dumps([e for e in IfcStore.get_file().by_id(data["id"]).Dimensions])
+            new.set_value(json.dumps([e for e in IfcStore.get_file().by_id(data["id"]).Dimensions]))
```
Moult, line 251: Does this actually work? I think it won't. Note that this is a callback - we are explicitly overriding the data_type to string. If you delete it then it won't know what to set value.

Moult (changes requested):

Gorgious56:
I felt I was threading on thin ice when changing this file :) 
Changes undone !

src/blenderbim/blenderbim/bim/module/material/operator.py
Gorgious56: Haha no problem I would rather not break everything when trying to streamline things

src/blenderbim/blenderbim/bim/module/unit/operator.py
Gorgious56: Unless `json.dumps` doesn't always return a `str` it should get detected and the `data_type` assigned in `set_value` :

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/blenderbim/blenderbim/bim/prop.py#L203-L215

But no problem rather safe than sorry then :)

src/blenderbim/blenderbim/bim/module/georeference/operator.py
Gorgious56: Oh right `import_attributes` makes sense ! I understand why the callback is needed in other instances but here it looks like it isn't needed ?

Yeah I'd be fine tackling the unit linking refactor, feel free to point directions and I'll dig around :)

src/blenderbim/blenderbim/bim/module/layer/operator.py
Gorgious56: This throws an error apparently `import_attributes` doesn't handle `select` attributes. Can i modify it like so ?         
```
if isinstance(data_type, tuple) or data_type in ("entity", "select"):
    callback(attribute.name(), None, data) if callback else None
    continue
```
instead of 
```
if isinstance(data_type, tuple) or data_type == "entity":
    callback(attribute.name(), None, data) if callback else None
    continue
```

src/blenderbim/blenderbim/bim/module/material/operator.py
Gorgious56: This one looks like it is initialized differently but when testing with regular `import_attributes` I didn't see any difference

src/blenderbim/blenderbim/bim/module/material/operator.py
Gorgious56: In the end I prefer not risking breaking things I don't have the data to test against :)

Issue #1497, opened 2021-05-27
BlenderBIM is getting bigger and bigger, what about using it's own repo?


Amoki:
BlenderBIM is becoming a very nice software, with many collaborators and functionalities. This implies a lot of PR, issues, discussions, and releases. 
I'm not a BlenderBIM user, just an IfcOpenShell user and it's becoming hard to follow IfcOpenShell evolution because GitHub can't allow me to filter notifications by tag. If I want to follow IfcOpenShell, I must follow BlenderBIM. I have the same issue if I want to read the latest commits to see bug fixes and improvements or check the latest release.
A simple button name change could imply an issue, a commit, a PR, a release, and 3 GitHub notifications.

If I understand well, BlenderBIM comes with its own IfcOpenShell build embedded and doesn't depend on IfcOpenShell source code directly.
Many people have write permission to the entire repo and are asked not to touch anything outside the BlenderBIM directory, but GitHub can't restrict them.

Using the IfcOpenShell repo is convenient for small projects. It gives a space with helpful people, simplifies hosting, upgrades management, and visibility. When the project gets bigger, these benefits become less significant. BlenderBIM has its own website, releases, and maintainers.

What do you think about moving the folder `src/blenderbim` to its own repository?

Moult:
The BlenderBIM Add-on source code is decoupled from the C++, but most of the functionality is part of the IfcOpenShell Python module, especially the API module which is how everything in the BlenderBIM Add-on gets executed - the Blender-related code is a GUI and UX layer syncing, primarily. In addition, the code for IfcDiff, IfcPatch, IfcP6 (+MSProject), etc are all Blender agnostic, yet is closely synced to a UI within Blender. In short, the BlenderBIM Add-on code is only half Blender - the other half is agnostic, and outside the `src/blenderbim` directory.

I understand the noise when following :( I personally wouldn't want to split the repos, though, with the primary reason being that development really is split halfway (the ifcopenshell.api namespace is the biggest aspect).

Curious on what others think.

aothms:
I think this is something debated a lot in the monorepo vs. submodules discussions. From what I understand most companies these days gravitate towards monorepos. But I must admit I don't know how notifications are handled in that sense. I also don't really use notifications personally. Tagging would probably be pretty inconsistent anyway so I'm not sure that is a real solution to combat the noise.

[...]

PR #1701, opened 2021-08-27
Add basic BCF pypi upload CI action.

Krande:
For now I propose a simple trigger using a commit tag "[bcf release]". I think this is just as well since we do not (yet) have an automatic versioning for the bcf package

### DIFF
--- a/.github/workflows/ci-bcf.yml
+++ b/.github/workflows/ci-bcf.yml
@@ -22,6 +22,7 @@ jobs:
           python-version: '3.x'
       - name: Build package
         run: |
+          cd src/bcf
           pip install build
           python -m build
       - name: Publish package
@@ -29,6 +30,7 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.PYPI_TOKEN }}
+          packages_dir: src/bcf/dist
 
 
 


aothms:
Great. Thanks @Krande. For others ending up here by google. The pypi project URL is

https://pypi.org/project/bcf-client/

Some of the stats look a bit misleading perhaps as it is not related to BCF (but that's life I guess)

afbeelding

I noticed that there is also https://pypi.org/project/bcf-api-xml/ maybe these efforts can be consolidated? I think they also use IfcOpenShell. Is it @Amoki?

Amoki:
Hi!
We just started a discussion with @Moult about it: https://github.com/bimdata/BCF-API-XML-translator/issues/3

I think we can merge our effort :)

Similar to the https://github.com/IfcOpenShell/IfcOpenShell/issues/1497 discussion, I think some projects will be easier to maintain if they are on their own repo. It simplifies the CI, issue management, and dependencies.
Some big projects use mono-repos, but they use complex tools to help them manage it. The bcf code has no dependency on IfcOpenShell and vice-versa, I see no reason to use a mono-repo for this project.

PR #1708, opened 2021-08-30
Notify if file already contains library element

Gorgious56:
Proposal to display a checkmark instead of the "append library element" operator if said element already exists inside the current file.
I thought it would be nice for the user to visually know if the element they're trying to append already exists in their file.
I used the same logic as the one in the ifcopenshell api to know if the element already exists https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py

BSE_42

I thought about using another icon like the link one image but that would imply a 2-way-relationship which is not the case here.
Could have detrimental effect on performance if hundreds of objects are displayed at once in the list though...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,4 +78,6 @@ jobs:
         run: |
           cd test
           sudo /usr/bin/python tests.py
-          sudo /usr/bin/python ../src/ifcopenshell-python/ifcopenshell/test_ids.py
+          cd ../src/ifcopenshell-python
+          sudo /usr/bin/python -m pip install pytest
+          make test
--- a/src/blenderbim/blenderbim/bim/module/project/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/project/operator.py
@@ -204,11 +204,13 @@ class ChangeLibraryElement(bpy.types.Operator):
 
     def execute(self, context):
         self.props = context.scene.BIMProjectProperties
+        self.file = IfcStore.get_file()
+        self.library_file = IfcStore.library_file
         ifc_classes = set()
         self.props.active_library_element = self.element_name
         crumb = self.props.library_breadcrumb.add()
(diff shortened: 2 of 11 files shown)


Moult:
This is alright, except it shouldn't be in the ui file - it should be in the operator file, and the append operator should update an `is_appended` property.

Gorgious56:
That makes sense. I originally used `ui.py` because the `poll` operator doesn't have access to the element `id` and can't know if it's already appended,
Although I don't really know on which object to update the `is_appended` property ?

Moult:
@Gorgious56 I'm suggesting a new `is_appended` property on https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/blenderbim/blenderbim/bim/module/project/prop.py#L35-L38 and you can write the detection code here: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/blenderbim/blenderbim/bim/module/project/operator.py#L233-L241 and also update it before this operator finishes https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/blenderbim/blenderbim/bim/module/project/operator.py#L371

Gorgious56:
A lot of file changes were added in the merge process to get the library code up to date, but the actual files I changed are in project/operator, project/prop and project/ui.

However I have a problem with unnamed profile def types, since the code relies on element name to check if the element is already appended, it fails to retrieve it since the elements don't have a unique name identifier. 
https://github.com/Gorgious56/IfcOpenShell/blob/display-when-library-asset-is-already-imported/src/blenderbim/blenderbim/bim/module/project/operator.py#L247-L248
Either it shows all IfcProfileDef as appended when I appended only one of them, or it doesn't ever display the checkmark. @Moult Is there a way to circumvent that limitation ?

PR #1715, opened 2021-08-31
Parametric object gizmo scale follows scene unit

Gorgious56:
This (I hope) fixes https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/blenderbim/blenderbim/bim/module/drawing/gizmos.py#L473

Previously the parametric gizmo that's used to interactively change a parametric object's `IfcExtrudedAreaSolid/Depth` parameter would only be visually correct when using Millimeters as a unit.

image

Now it detects the unit at runtime and scales the gizmo depending on it.
Example using imperial system and feets :
BSE_45

This PR also fixes an error when setting project units to MILES or THOU (thousandth of an inch) by adding entries in the `conversion_si` dictionary and 2 cases when assigning units.

There still is a problem with the gizmo in that the teal guide lines don't scale with the units and I don't really know how to achieve it. Gizmo documentation is pretty sparse online... It's problematic when using really small units like micrometers.

image

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/drawing/gizmos.py
+++ b/src/blenderbim/blenderbim/bim/module/drawing/gizmos.py
@@ -548,5 +548,5 @@ class ExtrusionWidget(types.GizmoGroup):
             elif length_unit == "INCHES":
                 scale_value /= si_conversions["inch"]
             elif length_unit == "THOU":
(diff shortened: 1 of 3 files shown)


Moult:
Only minor comment to use "thou" instead of "thousanths of an inch" other than that feel free to squash and merge yourself :)

Issue #1674, opened 2021-08-20
Optimising geometry iterator initialisation


Moult:
Typically with models coming from Tekla, I am finding that geometry iteration is extremely slow. I've dealt with hundreds of megabytes of IFCs from Revit, but Tekla seems to be able to pack a lot more into smaller files, though I'm not yet sure how. My computer will struggle to import an 80MB Tekla file, but will happily churn through a 300MB Revit file.

I'd like to investigate ways to optimise the iterator in this scenario, otherwise Tekla files become a bit unworkable.

I've only started code-diving, but this is my first time looking at this aspect, so I could be totally off the mark.

It seems as though the `iterator.initialize()` is already slow. It seems to spend a huge amount of time in [collect](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom/IfcGeomIteratorImplementation.h#L394) even before multiprocessing begins.

A few random sample files, I've also highlighted a count of by_type("IfcRepresentation") that has a non zero "OfProductRepresentation" or "RepresentationMap" in case it is useful:

Source | Stage | STEP Lines | IfcRepresentations | % | OfProductRepresentation | RepresentationMap
--- | --- | --- | --- | --- | --- | ---
Revit | Consultant | 569435 | 7647 | 1.34 | 6903 | 744
Revit | Trade | 1911426 | 48038 | 2.51 | 45883 | 2155
Tekla | Steel | 1184489 | 308607 | 26.05 | 308603 | 4

The first one initialises in 1.2 seconds. The second in 165 seconds (not linear, it seems). The third I gave up waiting.

Note 99% of the IfcRepresentations in the Revit files are Body. However, Tekla exports a Body, Axis, and CoG for everything. Therefore, out of 308603, only 106568 are body.

Thoughts? I will continue to look, but thought I'd post what I had started looking at.

Edit: happy to provide some sample files if it helps :)

Moult:
If I add `if ((*jt)->ContextIdentifier() == "Body") { ... }` [here](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom/IfcGeomIteratorImplementation.h#L345) it cuts it down to 106568 and it processes in 318 seconds, which seems a bit more reasonable.

Moult:
[...]

Issue #1290, opened 2021-02-03
Need to understand how INCLUDE CURVES works


Moult:
I'm a little confused by the behaviour of INCLUDE_CURVES and hope that maybe once I understand it I can implement in a consistent manner how 2D objects are loaded into the scene.

Below are all the scenarios I can think of. A few things which I'm confused by:

 1. Isn't it extra work for the iterator to process it twice in Scenario A? Shouldn't it pick one and stick to it?
 2. In Scenario B how does it decide which representation to process? What if there were more than one 2D representation?
 3. I guess scenario C has a similar priority but for non 2D shapes?

In the past, I generally disabled INCLUDE_CURVES by default because I noticed some elements that had 3D bodies would be processed as 2D curves instead. But this was a long time ago and maybe I wasn't using the iterator, or maybe there is something I haven't discovered yet with the iterator (is the order in which it processes it twice consistent?). I guess what I'm saying is that I'd love to enable it by default so that annotations always appear (#1142) but I need to be a bit more confident that it won't bork 3D representations so I guess I'm asking is there something I haven't considered?

```
import ifcopenshell
import ifcopenshell.geom

f = ifcopenshell.open("/home/dion/Downloads/JudgesReception.ifc")
element = f.by_id(653)
body_3d = f.by_id(644)
axis_2d = f.by_id(633)

# *****************************************************************************

# INCLUDE_CURVES = True
# Method: Geometry iterator
# Element: Has 3D body + 2D axis
# Result: First 2D axis is created, then 3D body is created

print("Scenario A")
settings = ifcopenshell.geom.settings()
settings.set(settings.INCLUDE_CURVES, True)
iterator = ifcopenshell.geom.iterator(settings, f, include=[element])
iterator.initialize()
while True:
    shape = iterator.get()
    print(shape.guid)
    print(shape.geometry.edges)
    print(shape.geometry.faces)
    if not iterator.next():
        break

# *****************************************************************************

# INCLUDE_CURVES = True
# Method: Create shape on element
# Element: Has 3D body + 2D axis
# Result: Only 2D axis is created

print("Scenario B")
settings = ifcopenshell.geom.settings()
settings.set(settings.INCLUDE_CURVES, True)
shape = ifcopenshell.geom.create_shape(settings, element)
print(shape.guid)
print(shape.geometry.edges)
print(shape.geometry.faces)

# *****************************************************************************

# INCLUDE_CURVES = False
# Method: Create shape on element
[...]

Issue #771, opened 2020-01-22
Can geom iterator or create_shape understand representation context?


Moult:
Is it possible to have the geom iterator or a create shape return only the shape from a representation that is part of a specific context/subcontext/target view, say Model/Body/MODEL_VIEW, or Model/Clearance/MODEL_VIEW, etc? Or perhaps when it processes the shapes, I can tell which is which?

With this ability, I can make the import script distinguish between contexts and build features to let users isolate contexts.

Moult:
To clarify on this issue, I note that it is possible to do create_shape on a particular item in a representation, this allows me to achieve this.

Similar to this issue, is it possible to use the geom iterator to pull out individual representation items? For example, if item 1 has style A, and item 2 has style B, then when I iterate, I want to be able to distinguish which item has which style. I understand that the geometry iterator returns a shape which has `shape.geometry.materials`, but I am actually parsing materials myself to handle situations like material layers, material sets, and distinguishing between styles assigned to objects and styles assigned to materials (which can be represented in Blender).

I'd like to just use the geometry iterator, not create_shape, since it supports multicore ideally.

aothms:
You can use create_shape on an individual representation item. But indeed within the returned mesh, the provenance to individual items is not retrievable. We could add another array for that similar to the material_ids. What do you propose?

The context name is appended to the mesh id. Often as e.g. `-body`. And indeed as you say you can call create_shape() with the representation and track their provenance to context based on the inverse attribute on the context or forward attribute on the rep.

Overall this is something I hear more. On the one hand the ease of use of the iterator on the other hand more granular and detailed return types. But it's not trivial due to the product level modifications that can occur, such as openings, layers and materials.

Moult:
If another two properties were added, e.g. `shape.geometry.representations` (which would hold things like `Model/Body/MODEL_VIEW` (note that Body alone is not enough, since all three are required for the user to apply filtering in the viewer), and then `shape.geometry.representation_ids` that would resolve the issue.

I don't mind not using iterator, but the three showstoppers for create shape currently are:

 1. Unable to filter elements (sometimes, like a partial import, I only need to iterate through a subset)
 2. I have to manually work out the local -> global coordinate conversion when placing an object, which since IfcOpenShell doesn't expose it in Python, I need to redo it, and I've made mistakes, and doubt I've caught all my mistakes :)
 3. Multicore - super, super, important when I need to import 1GB of models every week.

Moult:
[...]

PR #1741, opened 2021-09-22
Filter by context id, not using curves or solids. See #1674.

Moult:
Work in progress. Needs discussion. Note: this PR is incomplete - do not merge, if you want to test it, you need to comment out https://github.com/IfcOpenShell/IfcOpenShell/blob/636d80093652db449168757dae6d1bb2dba788e6/src/ifcgeom/IfcGeomIteratorImplementation.h#L325-L340 to stop the old behaviour.

Assuming you have it commented out, this implementation now allows:

 1. If no context IDs are set, all representations are iterated through, courtesy of https://github.com/IfcOpenShell/IfcOpenShell/blob/636d80093652db449168757dae6d1bb2dba788e6/src/ifcgeom/IfcGeomIteratorImplementation.h#L365-L368
 2. If you set one or more context IDs that have representations, it only processes those, this allows you to filter out just things like axes, footprint, box (assuming we build box support), etc very easily. This solves #771.
 3. Because it only processes the context you want, in 99% of the cases you only want bodies, then fall back to other contexts, this results in significant speedups. This contributes to #1674.
 4. Because you specify exactly the context ID, it allows the user to have zero ambiguity (compared to include_curves/solids) about what is being processed and in what order. This solves #1290.

Sample code:

```python
s = ifcopenshell.geom.settings()
s.set_context_ids([123,234])
```

Here's a screenshot of me pulling out just annotations (but not axis) from the test drawing file.

2021-09-22-103158_562x316_scrot

Questions:

 1. To expose `set_context_ids` I had to use `std::vector<int>`, then cast to set later. Is this OK?
 2. This implementation makes `INCLUDE_CURVES` and `EXCLUDE_SOLIDS_AND_SURFACES` obsolete. Is there any value in retaining them? One value I can think of is that it an easy way for users to say I want 2D and/or 3D. With the new approach, users need to be more specific ... 3D - what 3D? Body? 2D? What 2D? Axis? Footprint? Annotation? Target view? Scale? All? etc. This means the user (e.g. ifcopenshell-python users, and IfcConvert too) needs to do more work (select contexts) and filter on their end. Due to the uncertainty in this, I have not yet purged these options yet and not touched tools like IfcConvert.
 3. What should be the default behaviour if no representations are found? Right now:
    - Nothing specified -> Everything iterated (expected)
    - Contexts with representations specified -> Only those representations iterated (expected)
    - Contexts with no representation specified -> Everything iterated (? - I would expect nothing to iterate and initialisation to return false? If you agree, happy to make this change.)
 4. I know you asked "Something that also works with strings would be nice for the symmetry with IfcConvert." but I didn't fully understand. Is what I did OK?

### DIFF
--- a/src/ifcgeom/IfcGeomIteratorImplementation.h
+++ b/src/ifcgeom/IfcGeomIteratorImplementation.h
@@ -497,11 +497,25 @@ namespace IfcGeom {
 		void addRepresentationsFromContextIds() {
 			for (auto context_id : settings.context_ids()) {
 				IfcSchema::IfcGeometricRepresentationContext* context = ifc_file->instance_by_id(context_id)->as<IfcSchema::IfcGeometricRepresentationContext>();
+
+				if (!context) {
+					Logger::Error("Failed to process context ID " + std::to_string(context_id));
+					continue;
+				}
+
 				representations->push(context->RepresentationsInContext());
 
 				try {
-					if (context->Precision() && *context->Precision() < lowest_precision_encountered) {
-						lowest_precision_encountered = *context->Precision();
+					double precision;
+
+					if (context->as<IfcSchema::IfcGeometricRepresentationSubContext>()) {
+						precision = *context->as<IfcSchema::IfcGeometricRepresentationSubContext>()->ParentContext()->Precision();
+					} else {
+						precision = *context->Precision();
+					}
+
+					if (precision && precision < lowest_precision_encountered) {
+						lowest_precision_encountered = precision;
 						any_precision_encountered = true;
 					}
 				} catch (const std::exception& e) {
--- a/src/ifcgeom/IfcGeomIteratorImplementation.h
+++ b/src/ifcgeom/IfcGeomIteratorImplementation.h
@@ -342,6 +342,11 @@ namespace IfcGeom {
 				// WR31: The parent context shall not be another geometric representation sub context. 
 			}
 
+			for (auto context_id : settings.context_ids()) {
+				IfcSchema::IfcGeometricRepresentationContext* context = ifc_file->instance_by_id(context_id)->as<IfcSchema::IfcGeometricRepresentationContext>();
+				representations->push(context->RepresentationsInContext());
+			}
+
 			if (any_precision_encountered) {
 				// Some arbitrary factor that has proven to work better for the models in the set of test files.
 				lowest_precision_encountered *= kernel.getValue(IfcGeom::Kernel::GV_PRECISION_FACTOR);
--- a/src/ifcgeom/IfcGeomIteratorImplementation.h
+++ b/src/ifcgeom/IfcGeomIteratorImplementation.h
@@ -205,8 +205,10 @@ namespace IfcGeom {
 		IfcSchema::IfcProduct::list::ptr ifcproducts;
 		IfcSchema::IfcProduct::list::it ifcproduct_iterator;
 
-
         IfcSchema::IfcRepresentation::list::ptr ok_mapped_representations;
+
+		double lowest_precision_encountered;
+		bool any_precision_encountered;
 
 		int done;
 		int total;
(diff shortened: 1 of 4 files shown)


aothms:
> To expose set_context_ids I had to use std::vector<int>, then cast to set later. Is this OK?

Yes that's what I frequently do as well.

> INCLUDE_CURVES and EXCLUDE_SOLIDS_AND_SURFACES ... Is there any value in retaining them?

Well for the time being I think we should. It's basically a different use cases. When you're in IfcConvert I think you typically just want to get the data out and later perhaps filter on context. But still it might be the case that your viewer only supports 2d or 3d elements. We don't have the guarantee that there is a correspondence between Contexts and Dimensionality of the geometries.

Also it's a matter of convenience. When you don't have python available you don't really know what to expect and can't feed it a set of identifiers from the model.

> What should be the default behaviour if no representations are found? ... Contexts with no representation specified -> Everything iterated (? - I would expect nothing to iterate and initialisation to return false? If you agree, happy to make this change.)

Yes let's try and fix this along the way. I would say initialisation should still return `true`, but maybe we can provide more information in other methods what happened (like number of representations going to be processed)

> Something that also works with strings

We can add it later to lookup the numeric identifiers from ContextIdentifier/ContextType on the command line.

Moult:
Cheers, the latest commit should address all comments. It's a bit hard to see in the diff (because I changed indentation of the existing code - try viewing as a split instead of a unified diff, and it might be easier to read) but basically this is the logic:

```
if (settings.context_ids().size != 0) {
   ... add representations from context_id contexts ...
} else {
   ... do existing behaviour. I didn't change any code.
}
```

So if someone doesn't use context_ids, then behaviour is exactly the same as it is prior to this PR.

Since INCLUDE_CURVES et friends are no longer being removed, this means that anybody wishing to use context_ids has to be aware that they might need to enable INCLUDE_CURVES (e.g. if they select a Plan context).

I think it's ready for merging now, I can't think of anything else that needs changing.

Btw I apologise for doing a big if-else statement, let me see if I can refactor it safely into another function. I'm not too confident yet in this.

Moult:
OK I've refactored it a bit. Hope I didn't misunderstand C++ scopes and that it makes sense. Let me know if there is anything else, if not, happy for merging :D

src/ifcgeom/IfcGeomIteratorImplementation.h
```diff
@@ -590,6 +494,112 @@ namespace IfcGeom {
         const gp_XYZ& bounds_max() const { return bounds_max_; }
 
 	private:
+		void addRepresentationsFromContextIds() {
+			for (auto context_id : settings.context_ids()) {
+				IfcSchema::IfcGeometricRepresentationContext* context = ifc_file->instance_by_id(context_id)->as<IfcSchema::IfcGeometricRepresentationContext>();
```
aothms, line 499: `as<>()` returns 0 pointer when it isn't the correct type. You might want do a `if (context)` afterwards to test for this and either upgrade to an exception (which can be caught in Python as it is propagated by swig) or use Logger::Error and just skip this id for ex.

src/ifcgeom/IfcGeomIteratorImplementation.h
```diff
@@ -590,6 +494,112 @@ namespace IfcGeom {
...
+				representations->push(context->RepresentationsInContext());
+
+				try {
+					if (context->Precision() && *context->Precision() < lowest_precision_encountered) {
```
aothms, line 503: `Precision` is a derived attribute in case of sub contexts. In the other 'branch' we have guaranteed we aren't iterating over subcontexts for precision, but here I think there is a good chance it is a subcontext (?). Unforunately we have to 'implement' the derived attribute manually so `if (context->as<>(IfcGeometricRepresentationSubContext)) use context->as<>(IfcGeometricRepresentationSubContext)->ParentContext()->Precision()`

Moult:
Fixes (I think) pushed :)

aothms:
Yeah thanks, looks good to me. I didn't really test it, but I don't see much to worry. Since it's 0.7 it's not that risky. Thanks! I'm impressed by the speed with which you familiarized yourself with the C++ bits :)

Issue #1343, opened 2021-02-25
Box representations are not supported in IfcOpenShell


ecke101:
When I click this button, I get this error.

```
location: <unknown location>:-1
Error: Python: Traceback (most recent call last):
  File "C:\Users\erik\AppData\Roaming\Blender Foundation\Blender\2.92\scripts\addons\blenderbim\bim\module\geometry\operator.py", line 174, in execute
    self.pull_mesh_from_ifc()
  File "C:\Users\erik\AppData\Roaming\Blender Foundation\Blender\2.92\scripts\addons\blenderbim\bim\module\geometry\operator.py", line 191, in pull_mesh_from_ifc
    shape = ifcopenshell.geom.create_shape(settings, self.file.by_id(self.ifc_definition_id))
  File "C:\Users\erik\AppData\Roaming\Blender Foundation\Blender\2.92\scripts\addons\blenderbim\libs\site\packages\ifcopenshell\geom\main.py", line 193, in create_shape
    ifcopenshell_wrapper.create_shape(settings, inst.wrapped_data, repr.wrapped_data if repr is not None else None),
  File "C:\Users\erik\AppData\Roaming\Blender Foundation\Blender\2.92\scripts\addons\blenderbim\libs\site\packages\ifcopenshell\ifcopenshell_wrapper.py", line 634, in create_shape
    return _ifcopenshell_wrapper.create_shape(settings, instance, representation)
RuntimeError: Failed to process representation item

location: <unknown location>:-1
```

![bild](https://user-images.githubusercontent.com/249736/109205323-b92de000-77a6-11eb-89eb-9f61f72292f9.png)

Moult:
Known issue that Box representations are not supported in IfcOpenShell. See #85. Keeping this bug open as #85 wasn't exactly about box representations.

Moult:
Also ping @aothms

Moult:
Closing as fixed.

PR #1743, opened 2021-09-22
Implement IfcBoundingBox.

Moult:
Not too familiar with the OpenCascade API so maybe the implementation is overkill for a simple translation, but I think it works - here are some boxes:

2021-09-22-205650_866x518_scrot


Moult:
Just leaving a comment to link this to #1343

aothms:
It doesn't make a huge difference but did you see this constructor prototype that takes a gp_Pnt and three floats? https://dev.opencascade.org/doc/refman/html/class_b_rep_prim_a_p_i___make_box.html#a5f474dab6d876527977b391c3f685320

Moult:
Thanks! Makes a big difference to me :)

aothms:
Perfect

Issue #491, opened 2018-10-29
"Unable to find find keyword in schema"


Stinkfist0:
Note the duplicate "find". While at it, it would be nice to report the actual keyword which was not recognized.

aothms:
Ow oops. Thanks. Maybe if we're at it also create some different exception subclasses.

Moult:
I looked at the code for this. This string is in _many_ places in `ifcparse/`, which makes me suspect some ^C^V'ing :)

Do you mind if I just do a mass replace? I also notice the error message is quite generic whereas sometimes the "keyword" could be an enum, but not necessarily. Maybe when reporting the keyword, we should make the error message more specific.

If this sounds good to you I can start preparing a PR.

Stinkfist0:
These can be found from the buildingSmart website, however, I'm not completely sure which of these should be used now: http://www.buildingsmart-tech.org/mvd/IFC4Add2TC1/

aothms:
@Moult were you planning to commit a fix for this?

Moult:
I guess this can be closed :)

PR #1744, opened 2021-09-22
Improve error message when keyword not found. See #491.

Moult:
Better late than never. I didn't test it though, and I only added `+ s` for one of the error messages because I couldn't see how to do it for the other two.


aothms:
Haha yeah wonder how long these have been in there.

aothms:
We need to manually rerun the code gen afterwards though.

Moult:
Yes I avoided rerunning the codegen as you mentioned you also might want to exclude those and make it part of the build system.

aothms:
https://github.com/IfcOpenShell/IfcOpenShell/commit/fe533077f47c1ada9a334cabb79dea9a631f779a

aothms:
Let's save the build system changes for later

PR #1746, opened 2021-09-23
Fetch pytest-blender dependencies

Gorgious56:
This aims to automatically retrieve `pytest-blender` dependencies

I'm pretty sure I put that in the wrong place so please correct me if I did a horrible thing :)

### DIFF
--- a/src/blenderbim/Makefile
+++ b/src/blenderbim/Makefile
@@ -440,9 +440,6 @@ endif
 	cd dist/working/olca-ipc-0.0.10/ && cp -r olca ../../blenderbim/libs/site/packages/
 	rm -rf dist/working
 
-	# Required by pytest-blender
-	blender -p setup_pytest.py
-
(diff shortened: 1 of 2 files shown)


Moult:
Close ;)

```
.PHONY: dev
dev:
	blender -p setup_pytest.py
```

Gorgious56:
Ok the shroud around all of this is slowly dissipating ! Should be good now :)

Now that I think about it, I don't know if it's desirable to force update pip, maybe some users want to stick to a particular version ?
https://github.com/IfcOpenShell/IfcOpenShell/blob/dabb3b1cdeb3c75001060b421e2d06ea7d5eb11f/src/blenderbim/setup_pytest.py#L26

Moult:
Your call :) Feel free to merge when you feel like it! Looks good here.

Gorgious56:
OK I'll leave it for now and wait for an angry developer's feedback if something goes wrong ^^
Thanks for the pointers !

Issue #1711, opened 2021-08-31
Refactoring for mature native authoring, TDD, and scalability of the codebase.


Moult:
As a self-professed TDD enthusiast who embarrassingly didn't write a single test, I've run out of excuses :)

During the first year of the BlenderBIM Add-on, it was a monolithic import / export function. This was untestable and due to my lack of IFC knowledge at the time I was not clever enough to modularise it and all the IFC entities seemed rather spaghetti-like.

With the best of intentions, at the beginning of this year the add-on was rebuilt from scratch designed to be testable. Yet, I still did not write tests. Unfortunately, the Blender UI is not fun to test, and the IfcOpenShell API is basically a database / repository layer, which requires data to test. Writing database tests are [not fun](https://stackoverflow.com/questions/145131/whats-the-best-strategy-for-unit-testing-database-driven-applications). Essentially I was lazy and now I'm paying the price. However, I think that the IfcOpenShell API has matured enough that we can now use the API itself to write tests, which'll make test writing more fun.

Here's an unscientific roll through the last few bugfixes and categorising them by testable layer based on the test pyramid. Note that it's slightly less pyramid shaped because @Gorgious56 went crazy on the operator layer and focused on those problems in particular. Not all of these bugs could've been caught from tests, but they could've benefited from tests. All of these fixes have brought us closer to native authoring.

Bug in Blender UI:
bcda3d8e ("Minor fix", 2021-08-29)
52f4b564 ("Minor fix. It seems as though StructRNA of type Scene may be removed during undo/redo, which results in handlers failing.", 2021-08-22)

Bug in Blender operators:
0ed38a41 ("Fix #1707. Fix bug where project library browser could self reference.", 2021-08-31)
dd0a3d04 ("Minor fix.", 2021-08-27)
3c235208 ("Fix #1679. Again.", 2021-08-22)
8178dd33 ("Fix #1679", 2021-08-22)
616f2841 ("Minor fix", 2021-08-21)
6c6d4c04 ("Fix #1672. Bug where style attributes couldn't be edited.", 2021-08-20)
11e36bf7 ("Minor fix", 2021-08-20)
29cda050 ("Fix error when adding opening with None object", 2021-08-19)
95f2d939 ("Fix error when filling opening with None object", 2021-08-19)

Bug in API:
4accf306 ("Add transaction support for adding elements. See #1709.", 2021-08-31)
[...]

PR #1754, opened 2021-09-24
Improve filling operators and tests

Gorgious56:
The purpose of this PR is threefold :
- Prevent possible errors when certains combinations of operators and objects are not legal
- Write tests for add_filling and remove_filling operators
- Improve filling setup UX by displaying a list of all fillings of an opening :

image

Although I'm not sure this is the best way to retrieve an object from id :

https://github.com/Gorgious56/IfcOpenShell/blob/5ac533224e426c125e91089a9fa864b33215d853/src/blenderbim/blenderbim/bim/module/void/ui.py#L95


Moult:
Are you sure all tests are covered in this PR? What about adding the invalid filling types?

IfcStore.get_element() might be what you're looking for.

Gorgious56:
Yeah I should have pushed as a draft first, I wanted to know if you were ok with how I did it. I'll start working on the other tests then :)

Gorgious56:
@Moult I've added most of the test cases I think.
I'm having trouble with a couple of lines which I added in https://github.com/IfcOpenShell/IfcOpenShell/blob/7ead2c2a0b2e6a0fded5fe6816e0c15d3c5011e9/src/blenderbim/blenderbim/bim/operator.py#L675-L676
The tests seem to hiccup and fail when I uncomment these lines and I don't understand why. They're used to remove an opening's fillings relationships when the user deletes the opening object using the delete operator.
So this test fails https://github.com/IfcOpenShell/IfcOpenShell/blob/7ead2c2a0b2e6a0fded5fe6816e0c15d3c5011e9/src/blenderbim/test/bim/module/void/test_operator.py#L343-L360

Gorgious56:
https://github.com/IfcOpenShell/IfcOpenShell/issues/1711 for reference

Moult:
Doing a James Bond

Gorgious56:
Wut ?? ><

Moult:
Commit first ask questions later :D

Gorgious56:
Haha I like that !
Does https://github.com/IfcOpenShell/IfcOpenShell/commit/24eac8ccdd174d3f5ade62795b088709266d5fbc mean we can start refactoring the tests using this syntax ?

Moult:
Should be fixed. And yes, feel free to start refactoring the tests. One thing however yet to decide is how to organise them. Right now I have a single `test_feature.py` which triggers all tests in the feature/ subdir. This is probably a bad idea, as we want two abilities:

 1. Run only a single scenario, or a subset of a few scenarios and exclude everything else, very handy for debugging
 2. Run only scenarios pertaining to a module.

Probably an easy solution, just a convention needs to be established of how to structure the test files. Happy for you to go ahead and go for it :)

Moult:
So the reasons the tests hiccuped (I assume you mean crashed or segfaulted) is because indeed the tests are doing their job :)

In short, if you `delete_opening_element(element)` then you do `for rel in element.HasFillings`, the element is already deleted, so element is now an empty pointer which crashes both Blender and the tests.

Two solutions:

 1. Remove fillings first, then delete the opening element (which I did)
 2. Just delete the opening element anyway, which should purge fillings in the process.

Unfortunately I can't quite choose the more simple second solution just yet because right now we must call everything through operators and the operator expects both an opening and a thing it is voiding. In the future, as #1711 slowly trickles through the system, it will be much easier to make these types of changes. Also, complex scenarios like this in theory would be described much more elegantly using domain language (I assume when you wrote this code, you might've needed to read the logic in the operators quite a few times to get a feel for what was happening, which is not a very good experience).

Moult:
The good news is we have increasingly more (albeit slow) high level tests :) Therefore in the future when we want to do some serious gut-ripping, clever parametric whizbang features, etc, we can do so without fear. By the way, you might want try writing tests first, then writing the code - it's a fun trick to learn, but super powerful when you get the hang of it!

PR #1765, opened 2021-09-30
Use pytest-bdd syntax for void tests

Gorgious56:
It was pretty straightforward thanks to your show&tell of this morning (or afternoon down under) but I just wanted your input if you're okay with it @Moult on two things :

Some sentences can appear with the Given and When keywords eg `Given I add a cube` and `When I add a cube` is it okay to stack the decorators that way ?

https://github.com/Gorgious56/IfcOpenShell/blob/f660fbdebbd60f3e572a03d508565adc9af709b0/src/blenderbim/test/bim/test_feature.py#L45-L46

Also it seems the overriden delete operator doesn't get called anymore so the tests that rely on deleting voids and fillings as a callback don't work anymore. Any idea why ?

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/operator.py#L647-L689


Moult:
Stacking is fine :)

The tests seem to pass on my end - can't replicate the failing tests here?

Gorgious56:
Hmm yeah then there's something fishy on my end, will investigate ! :) Thansk for the feedback

Issue #1766, opened 2021-10-01
Should removing an element ensure cardinality is correct in IfcOpenShell core?


Moult:
```python
>>> import ifcopenshell
>>> ifcopenshell.version
'v0.7.0-7155f7ce'

>>> f = ifcopenshell.file()
>>> p = f.createIfcPerson()
>>> r = f.createIfcActorRole()
>>> p.Roles = [r]
>>> f.remove(r)
>>> p.Roles
()
```

The cardinality of `IfcPerson.Roles` is `? L[1:?]`. Should the remove() call be intelligent enough to preserve the correctness of the cardinality? I.e. should it instead make p.Roles = None?

I can understand if this is too complex (e.g. what is the cardinality is `L[1:?]` instead?). I have been writing code to handle these scenarios in the IfcOpenShell API, but it would be good to know whether or not it belongs upstream in C++ or downstream in Python, or if a general solution is even possible.

aothms:
> should it instead make p.Roles = None

At least given that we have the full schema present, it's possible. For this case specifically it's just near [0] `checking instance_list->size()` and `related_instance->declaration().as_entity()->all_attribute()[i]->is_optional()` and `->type_of_attribute()` etc. That would be a bit ad hoc, ideally it'd be a bit more general and configurable, but then the complexity skyrockets out of control.

[0] https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcparse/IfcParse.cpp#L2035

> e.g. what is the cardinality is L[1:?] instead?

Throw an exception? Also that is not impossible. Just as easy basically, just need to make sure it doesn't leave us in an incorrect state. Maybe some sort of "strict mode" users can enable to enforce integrity?

I don't know, I don't have strong opinions on this. But in general we should help people write files as valid as possible.

Moult:
Cheers. PR submitted. Indeed I cannot think of a general solution, but hopefully the PR will make cases more valid in a lot more situations. I'll keep a close eye on the remove API code and see if any other good strategies emerge. This at least will allow me to remove about half of the special handling code I have in the API.

Moult:
In the future, a strict mode would definitely be something to consider. For now, I don't think it's a big issue, as after all as we create objects cardinality is not enforced as well. BTW tests now document this behaviour.

PR #1769, opened 2021-10-03
Upon removal, empty aggregates prioritise null if possible.

Moult:
See #1766.

### DIFF
--- a/src/ifcopenshell-python/test/test_file.py
+++ b/src/ifcopenshell-python/test/test_file.py
@@ -203,6 +203,39 @@ class TestFile(test.bootstrap.IFC4):
         self.file.remove(element)
         assert len(list(self.file)) == 0
 
+    def test_removing_an_element_and_its_references_are_cleared(self):
+        history = self.file.createIfcOwnerHistory()
+        element = self.file.createIfcWall(GlobalId="global_id", OwnerHistory=history)
+        self.file.remove(history)
+        assert element.OwnerHistory is None
+
+    def test_removing_an_element_and_its_references_are_cleared_regardless_of_cardinality(self):
+        wall = self.file.createIfcWall(GlobalId="global_id")
+        rel = self.file.createIfcRelAggregates(RelatingObject=wall)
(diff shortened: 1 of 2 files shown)


aothms:
I see you don't check whether the lower bound of the list is indeed 1. I guess there could also be `OPTIONAL LIST [0:?]` (although the semantic differences is unclear)? Also it's only for aggregates of instances, not of other datatypes.

I'll add a note about these things, but let's not delay merging the PR. Checking the lower bound is a bit difficult because the attribute is not necessarily directly an aggregate. It can also be a named type for example IfcCompoundPlaneAngleMeasure or SELECT with IfcPropertySetDefinitionSet or something else equally crazy.

A complicating factor is that the attribute types on the C++ side are just enumerations. Ideally instead of Argument_AGGREGATE_OF_AGGREGATE_OF_ENTITY_INSTANCE we whould have some sort of class definition AGGREGATE(AGGREGATE(ENTITY_INSTANCE)) so that we could apply this logic more generally.

aothms:
> Also it's only for aggregates of instances, not of other datatypes.

Duh, this is the file.remove() function, that doesn't apply to other datatypes.

Moult:
Cheers :) yes I regard null and empty list to be semantically equivalent so I didn't bother checking the lower bound.

Agree about the third paragraph, but I didn't understand your second paragraph? How does select affect this code?

aothms:
> How does select affect this code?

Only if you want to check the lower bound. The aggregate type definition can be quite a few indirections away. Through various named type definitions or selects (actually maybe not selects, maybe there would be another entity instance always in between then). It's not important.

aothms:
> semantically equivalent

Maybe also something we can take another look at for 4.3. Although I can't really judge how important this is. You also see this with booleans for example. Where you have (a) `optional boolean` (b) `logical` and even (c) `optional logical`. (a) and (b) are I'd say equivalent unless you want to debate the semantic difference between "omitted" and "unknown". (c) I'm really not sure why one would need 4-state booleans.

PR #1771, opened 2021-10-04
IFCCSV - added new functionality and adjusted UI

vulevukusej:
Adjusted the UI slightly and added a few new useful operators:

1. Load Template Button - Gives the user the ability to load from previously saved templates.  (See 3)
2. Clear All Button - Clears all Attributes in the list
3. Create Template Button - Saves the currently defined list of attributes and saves to a JSON file.  Useful when using the tool for different export/import scenarios.

UI - Added a couple of spacers between UI elements


Moult:
Great work! So sorry for not replying earlier. Merging :)

Moult:
Great work! So sorry for not replying earlier. Merging :)

PR #1772, opened 2021-10-04
Pset Templates tests

Gorgious56:
Adding feature scenarios for pset and property templates.
In summary :
- Didn't find how to test the undo / redo functionality since it seems it's disabled in the test configuration
- Added a sentence to load the default pset templates file
- Didn't find a way to get a property template id from the ifc file, so I hardcoded it to 2. I tried `{ifc}.by_type('IfcSimplePropertyTemplate')[0].id()` but that wasn't it.

I also added an if statement before purging the file to prevent info message saying there is no object to purge cluttering console if the default file contains no object.

### DIFF
--- a/src/blenderbim/pytest.ini
+++ b/src/blenderbim/pytest.ini
@@ -3,3 +3,4 @@ markers =
     context
     owner
     void
+    pset_template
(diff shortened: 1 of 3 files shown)


Moult:
So sorry I missed this. Yes indeed undo redo probably cannot be tested generically across everything. The reason `{ifc}.by_type...` didn't work is because `{ifc}` returns the active IFC file, which is not the same as the active external pset template file. It's OK to hardcode it as 2 in this simple scenario. In the future, if the tests get more complex, we can introduce a new variable like `{pset_ifc}` which refers to that external file. Feel free to merge :)

PR #1777, opened 2021-10-04
Drop obsolete pythons

aothms:


aothms:
#1380 but then on 0.7

Issue #1776, opened 2021-10-04
Do we still want to maintain python 2.7 on 0.7.0?


Amoki:
It complexifies the build, many tools (BCF, bsdd) are python3 only.
I'm currently updating `build-all.py`, I can submit a PR for this file.

EDIT: Same question for python < 3.6 which are already unmaintained. The end of life of python 3.6 is scheduled for 2021-12-23. (source: https://devguide.python.org/#status-of-python-branches)

aothms:
Hi @Amoki I just merged https://github.com/IfcOpenShell/IfcOpenShell/pull/1777 it kept 2.7 and 3.5 though. For 0.7 I'm fine with everything. Let's just eliminate up to and including 3.6. At the very least it'll reduce the build time significantly for the IfcOpenBot builds. I think there is also some old remnants regarding PyUnicode PyBytes that can be discarded when no longer supporting <= 3.2

Amoki:
OK, I'm removing python <= 3.5.
I'm also replacing `%` and `.format()` with f-strings and improving configuration with some more env variables.

aothms:
That'd be great. There is some weird stuff in there, e.g get_os() calling `uname` instead of using the native python `platform` module. 

A lot of try except instead of os.environ.get(key, **default**)

In the long run I'd also be very interested in the ability to run it on Windows.

No pressure though, just listing some of my own concerns for the future.

Amoki:
I'm not using Windows, I won't be able to help on that but `os.getenv()` and stuff like that will be included :)

Is svgfill required to build IfcOpenShell or I can make it optional?

OpenCOLLADA is a required dependency on IfcGeom in targets and I'd like to make it optional. Should it be enabled or disabled by default?

aothms:
> Is svgfill required to build IfcOpenShell or I can make it optional?

I think it should be optional, but currently it isn't in the code. I can work on that.

> OpenCOLLADA is a required dependency on IfcGeom in targets and I'd like to make it optional. Should it be enabled or disabled by default?

What do you mean by default? You mean through env vars? I guess if you want that kind of control from the outside we should look at argparse or something?

Amoki:
> I think it should be optional, but currently it isn't in the code. I can work on that.

Could be awesome. svgfill seems to have its own dependencies and my first attempt to build it failed.

> What do you mean by default? You mean through env vars? I guess if you want that kind of control from the outside we should look at argparse or something?

[...]

PR #1781, opened 2021-10-06
Feature/improve build all

Amoki:
Related to https://github.com/IfcOpenShell/IfcOpenShell/issues/1776

I've added python 3.10 released 2 days ago, I didn't try it.

It works on my machine and on my servers, but it needs more testing.


aothms:
Many thanks! I only have one comment see above

aothms:
Can you see it now? Guess I missed a button.

PR #1798, opened 2021-10-14
BlenderBIM: added dropdown in addon preferences to toggle module visibility

vulevukusej:
I added a dropdown-menu to the addon preferences that allows the user to select from predefined roles.  Depending on the role chosen, panels will be either visible/hidden in the "PROPERTIES" window.

changes

### DIFF
--- a/src/blenderbim/blenderbim/bim/helper.py
+++ b/src/blenderbim/blenderbim/bim/helper.py
@@ -100,7 +100,7 @@ def export_attributes(props, callback=None):
 
 def create_module_property_group(modules):
     def change_state(self, context):
-        module_specific_classes = modules[self.module_name].classes
+        module_specific_classes = list(filter(lambda cls: issubclass(cls, bpy.types.Panel), modules[self.module_name].classes))
         try:
             for cls in reversed(module_specific_classes):
                 bpy.utils.unregister_class(cls)


Moult:
I think the add-on preferences is a good spot. Typically you'd set it once and leave it for a good while I assume.

vulevukusej:
Ok so I finally got around to doing this.  I went with the simple approach, as it doesn't seem too overwhelming when the list is split into three columns.  If you´re happy with that, I´ll update the code, otherwise let me know and I´ll spice it up a little ;)
update
:

Moult:
Looks great! Let me know when the code is ready for merging and I will merge it :)

Gorgious56:
This is awesome ! If I may suggest a little addition while you're polishing the UI, it would be even better if the BoolProperties displayed a Capitalized name instead of the variable name. You can do it easily by adding a `name` parameter to the properties. eg `aggregate: BoolProperty(default=True, name="Aggregate")`

Moult:
Skimming through the code on my phone I'd prefer if a whole list of bool properties were not hardcoded. I would prefer if a collection property were used instead, with the name referring to the module name, and an is visible bool property. The list of modules may be inferred.

Moult:
On second thoughts, many panels each doing a get function on a collection which has 50 items a few times every second might be a bad idea.

If a slick solution can be found go for it but don't let that stop you from committing :)

Gorgious56:
I think the lightest computationally wise is to dynamically unregister and re-register the panels that way we don't even have to use the `poll` method. It needs to be done in a smart way though but we could add some kind of flag to each parent panel class and fetch all panel classes introduced by the add-on and then use an operator with as a parameter the name of the panel class, that will dynamically unregister it

vulevukusej:
Alright guys thanks for the feedback.  I'll give @Gorgious56 idea with the dynamic register/unregister a go 👍

Moult:
Agreed I like the idea of simply unregistering the panel.

vulevukusej:
Ok, so after a couple of dodgy pull requests (still learning git! :) ), I've updated the code to reflect the latest ( as of writing this comment) version of BlenderBIM, and also added the dynamic register/unregister of modules functionality.

update2

Gorgious56:
> 
> 
> I've made an absolute mess of the commits - Should I just close this pull request and create a clean one without a million commits, or can you guys sort through the garbage I've created?

I wouldn't sweat over it, we can just squash the PR and merge into the main branch and all the steps will be hidden from everyone ^^

src/blenderbim/blenderbim/bim/__init__.py
```diff
@@ -67,8 +67,38 @@
...
+        return {'FINISHED'}
+
+    opclass = type(
+        f"BIM_OT_Toggle_Module_{nameUpperCase}",
+        (bpy.types.Operator, ),
+        {
+        "bl_idname": idname, 
+        "bl_label": nameUpperCase, 
+        "execute": func,
+        "classes": modules[name].classes
```
Gorgious56, line 98: I think it's clever to parse all classes in specific modules, but I think we're going to run into problems when some modules will try to comunicate with other modules. Maybe we can restrict classes to only those of type `bpy.types.Panel` ? That way only UI elements are unregistered and operators can still be called by other modules ?

vulevukusej: That`s a great idea, I was worrying about that too.  I´ll change it in the next commit.

src/blenderbim/blenderbim/bim/__init__.py
```diff
@@ -67,8 +67,38 @@
...
+
+    #generate operators that toggle module state
+    idname = f"bim.toggle_module_{name}"
+    
+    def func(self, context):
+        try:
+            for cls in self.classes:
+                bpy.utils.unregister_class(cls)
+                #self.is_active = False
+        except:
```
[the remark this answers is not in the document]
Gorgious56: Also FYI the inverse exception (trying to register a class twice) throws a `ValueError` :)

Issue #1403, opened 2021-03-26
IFC files with point cloud data are not displayed in Blender


cemunds:
When loading a point cloud .ply file into Blender and exporting to IFC, I can see that the point cloud is transformed into a IfcBuildingElementProxy and a project structure with building and site is created in the current Blender session and the points are written to the IFC file as IfcCartesianPointList3D. However, when trying to load the exported IFC file in a new session, project, building, site, etc. are loaded correctly, but the point cloud does not show up neither in the 3D view nor in the outliner.

Am I overlooking something or is the import of point cloud data from IFC files not supported yet?

Moult:
Import of point cloud data is not yet supported. This is currently not a high priority, as storing it is still a bit impractical in SPF format. When the .hdf5 format stabilises, this will become a higher priority :)

cemunds:
Ok, thank you for the info. I am working on generating IFC models from point clouds and I was hoping that for objects which are too occluded to reconstruct a mesh from them, I could use the point cloud itself as a representation of the object. Guess I'll have to come up with another solution then, at least for the time being.

Moult:
@cemunds would you be interested in helping with the development of the .hdf5 format?

LaurensJN:
@Moult I'm taking a look at the PointCloud IfcShapeRepresentation type now. 

- I created the self.create_point_cloud_representation() in https://github.com/LaurensJN/IfcOpenShell/commit/9164747e5860303457baecae1c9c3b5423fa8b6c. How should IFC2x3 be handled in this case, as there is no PointCloud shape representation there? 
- In the BlenderBIM geometry UI & operators I added a 'Update Mesh as Point Cloud' operator that uses the previous method. See https://github.com/LaurensJN/IfcOpenShell/commit/164170b966fb917069d7e8b177284e6e85596987
- I'm not sure where to start for importing the PointCloud ShapeRepresentation type (BlenderBIM won't import a PointCloud type). Should that be somewhere along the lines here? I don't exactly get what's going on: https://github.com/IfcOpenShell/IfcOpenShell/blob/5a52f3aeea4b331ce8a991a8c681f668e180f63f/src/blenderbim/blenderbim/bim/import_ifc.py#L1267-L1291

Also, should there be a check when converting the mesh representation to a point cloud, if there are any faces / edges? Seems not so foolproof to just have the convert button without any checks.

Moult:
[...]

PR #1804, opened 2021-10-17
BlenderBIM: import IFC PointCloud Representation

LaurensJN:
See issue #1403. 

There are a few things I am unsure about:
(1) naming of the _representation_ id for the geometry of the obejct.  Should this some  representation specific like "IfcTopologyRepresentation" in create_structural_point_connections?
https://github.com/LaurensJN/IfcOpenShell/blob/9479112de361c221d9fd157b674328a23ea10a3f/src/blenderbim/blenderbim/bim/import_ifc.py#L716
(2) Where should create_pointclouds function be invoked? If after the create_products(), only the bounding box will be added. 
https://github.com/LaurensJN/IfcOpenShell/blob/9479112de361c221d9fd157b674328a23ea10a3f/src/blenderbim/blenderbim/bim/import_ifc.py#L591-L593


Moult:
> (2) Where should create_pointclouds function be invoked? If after the create_products(), only the bounding box will be added.
https://github.com/LaurensJN/IfcOpenShell/blob/9479112de361c221d9fd157b674328a23ea10a3f/src/blenderbim/blenderbim/bim/import_ifc.py#L591-L593

Is there a sample file you can provide? Off the top of my head, I think the create_products should focus on body representations where possible. I would've put the `create_pointclouds` function after `create_curve_products`, but I could be missing a detail.

src/blenderbim/blenderbim/bim/import_ifc.py
```diff
@@ -684,6 +686,62 @@ def create_structural_point_connections(self):
...
+            placement_matrix = ifcopenshell.util.placement.get_local_placement(product.ObjectPlacement)
+            vertex_list = []
+            context = None
+            representation = None
+            for item in items:
+                if item.is_a("IfcCartesianPointList"):
+                    vertices = [mathutils.Vector(list(coordinates)) * self.unit_scale for coordinates in item.CoordList]
+                    vertex_list.extend(vertices)
+                elif item.is_a("IfcCartesianPoint"):
+                    vertex = mathutils.Vector(list(item.Coordinates)) * self.unit_scale
```
Moult, line 710: List comprehension here might make this faster?

src/blenderbim/blenderbim/bim/import_ifc.py
```diff
@@ -684,6 +686,62 @@ def create_structural_point_connections(self):
...
+                elif self.file.schema == "IFC2X3" and representation.RepresentationType == 'GeometricSet':
+                    items.extend(representation.Items)
+
+            placement_matrix = ifcopenshell.util.placement.get_local_placement(product.ObjectPlacement)
+            vertex_list = []
+            context = None
+            representation = None
+            for item in items:
+                if item.is_a("IfcCartesianPointList"):
+                    vertices = [mathutils.Vector(list(coordinates)) * self.unit_scale for coordinates in item.CoordList]
```
Moult, line 707: List comprehension directly doing vertex_list.append() might be faster?

src/blenderbim/blenderbim/bim/import_ifc.py
```diff
@@ -684,6 +686,62 @@ def create_structural_point_connections(self):
...
+            if not self.is_pointcloud(product):
+                continue
+            items = []
+            for representation in product.Representation.Representations:
+                if representation.RepresentationType == 'PointCloud':
+                    items.extend(representation.Items)
+                elif self.file.schema == "IFC2X3" and representation.RepresentationType == 'GeometricSet':
+                    items.extend(representation.Items)
+
+            placement_matrix = ifcopenshell.util.placement.get_local_placement(product.ObjectPlacement)
```
Moult, line 701: In theory it's more complex than this, for example if a mapped representation is used, it may apply another coordinate transformation to the coordinates themselves. However, I'm happy to turn a blind eye to this because I find it oh-so-messy-and-confusing.

src/blenderbim/blenderbim/bim/import_ifc.py
```diff
@@ -684,6 +686,62 @@ def create_structural_point_connections(self):
...
+        return result
+
+    def is_pointcloud(self, product):
+        if not hasattr(product, 'Representation') or not hasattr(product.Representation, 'Representations'):
+            return False
+
+        for representation in product.Representation.Representations:
+            if representation.RepresentationType == 'PointCloud':
+                return True
+            elif self.file.schema == "IFC2X3" and representation.RepresentationType == 'GeometricSet':
```
Moult, line 740: In 2X3 we may need to check the type of the items to ensure that it is storing points, as opposed to curves / surfaces.

src/blenderbim/blenderbim/bim/import_ifc.py
```diff
@@ -684,6 +686,62 @@ def create_structural_point_connections(self):
...
+            self.link_element(product, obj)
+            result.add(product)
+
+        return result
+
+    def is_pointcloud(self, product):
+        if not hasattr(product, 'Representation') or not hasattr(product.Representation, 'Representations'):
+            return False
+
+        for representation in product.Representation.Representations:
```
Moult, line 737: If RepresentationType = 'MappedRepresentation' then need to recurse.

src/blenderbim/blenderbim/bim/import_ifc.py
```diff
@@ -684,6 +686,62 @@ def create_structural_point_connections(self):
...
+            mesh.from_pydata(vertex_list, [], [])
+
+            obj = bpy.data.objects.new("{}/{}".format(product.is_a(), product.Name), mesh)
+            obj.matrix_world = self.apply_blender_offset_to_matrix_world(obj, placement_matrix)
+            self.link_element(product, obj)
+            result.add(product)
+
+        return result
+
+    def is_pointcloud(self, product):
```
Moult, line 733: In theory, we may want to target product types too. In this case you need to check the RepresentationMaps attributes. The rest of the code can be reused.

src/blenderbim/blenderbim/bim/import_ifc.py
```diff
@@ -684,6 +686,62 @@ def create_structural_point_connections(self):
             obj.matrix_world = self.apply_blender_offset_to_matrix_world(obj, placement_matrix)
             self.link_element(product, obj)
 
+    def create_pointclouds(self, products):
+        result = set()
+        for product in products:
+            if not self.is_pointcloud(product):
+                continue
+            items = []
+            for representation in product.Representation.Representations:
```
Moult, line 695: If a product type, representationmaps needs to be checked instead.

src/blenderbim/blenderbim/bim/import_ifc.py
```diff
@@ -684,6 +686,62 @@ def create_structural_point_connections(self):
             obj.matrix_world = self.apply_blender_offset_to_matrix_world(obj, placement_matrix)
             self.link_element(product, obj)
 
+    def create_pointclouds(self, products):
```
Moult, line 689: Can I be a pain and recommend a different approach?

```python
def create_point_clouds(self, products):
    for product in products:
        representation = self.get_pointcloud_representation(product)
        if representation:
            self.create_point_cloud(product, representation)

def create_point_cloud(self, product, representation):
    # since we already have the representation, we no longer need to duplicate the code for checking representation type https://github.com/IfcOpenShell/IfcOpenShell/pull/1804/files#diff-84a51bb6e3dd5dcf1060381cfa319cd64e6f4b819e5208387d0e74d73dcf77a5R696-R699
    # also, we just loop through items directly, no need to create another list for items so this variable disappears https://github.com/IfcOpenShell/IfcOpenShell/pull/1804/files#diff-84a51bb6e3dd5dcf1060381cfa319cd64e6f4b819e5208387d0e74d73dcf77a5R694
    # also https://github.com/IfcOpenShell/IfcOpenShell/pull/1804/files#diff-84a51bb6e3dd5dcf1060381cfa319cd64e6f4b819e5208387d0e74d73dcf77a5R713-R717 can be deleted and just use mesh_name = f"{representation.ContextOfItems.id()}/{representation.id()}"
    # If you self.file.traverse each representation item, this code can be deleted too https://github.com/IfcOpenShell/IfcOpenShell/pull/1804/files#diff-84a51bb6e3dd5dcf1060381cfa319cd64e6f4b819e5208387d0e74d73dcf77a5R719-R720
    # Rest of the code is the same as what you've written
```

PR #1805, opened 2021-10-18
Added operator to add psets to all selected objects.  

vulevukusej:
See discussion: https://github.com/IfcOpenShell/IfcOpenShell/discussions/1691


vulevukusej:
Awesome!  :)

PR #1833, opened 2021-10-27
Refactor Classification Module (see issue #1711)

maxfb87:
Refactor the Classification Module, following the n.1 of this comment https://github.com/IfcOpenShell/IfcOpenShell/issues/1711#issuecomment-946347979

I have copied what has done in refactoring Context Module and Spatial Module.

The code functions ok but i'm not sure i did the correct things. These are my dubts:
1. i added only "classification" and "references" in the ClassificationData.data structure. Is it better to structure in ".data" also other variables like library_classifications, library_file, products, etc...?

2. there is the purge method but it's unused (as i know). Is it correct to delete it?

3. if i made errore please let me know


Moult:
Sure sounds good :) I'll of course code review and clean up after merging. By the way, I might delay merging this since I hope to do a release tomorrow and I want to avoid a last minute big change in a part of the code which doesn't yet have test coverage.

Perhaps create a new tracking issue with a checklist of all the modules and go through one by one and see which modules have `ui.py` which still speak to a `Data` class in `ifcopenshell.api`. That way we can methodically go through one by one and figure out the easy ones. Would you mind helping with that?

maxfb87:
Ok @Moult ! No problem to delay merging, i can't work on it in next days so i'll do that next week (at least).

I'll do also the tracking issue.

maxfb87:
I decided to close this pull request because there are old commits already merged and i think it's not correct.
l'll create another draft PR and an issue regarding past comments.

PR #1841, opened 2021-10-29
BlenderBIM: Import Resources for XER files

HassanEmam:
I have implemented the parsing of resources from xer file and creating their corresponding IFC elements.
The parsing process seems to go fairly quick however the creation of ifc elements take a bit of time.
Perhaps we can investigate future optimisation and performance improvements
On the other hand I found a class which was duplicate in the sequence operators of BlenderBIM namely "ImportP6XER" and I have removed the duplicate

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/sequence/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/sequence/operator.py
@@ -1097,48 +1097,6 @@ class ImportP6(bpy.types.Operator, ImportHelper):
         print("Import finished in {:.2f} seconds".format(time.time() - start))
         return {"FINISHED"}
 
-class ImportP6XER(bpy.types.Operator, ImportHelper):
-    bl_idname = "import_p6xer.bim"
-    bl_label = "Import P6 XER"
-    bl_options = {"REGISTER", "UNDO"}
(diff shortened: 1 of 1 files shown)


Moult:
Cheers! Still room for more optimisation but it'll do for now.

As for a wizard, have you checked out the new IFC feature where when you load an IFC project you can choose "Is Advanced" in settings and it'll bring up a partial IFC loader, which will preload a portion of the file and let you use filters to choose what to load? This will be similar in UI to that, although the system might be implemented differently. The code begins here: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/project/operator.py#L542-L562 (see is_advanced attribute) which then if enabled calls this "load project elements" https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/project/operator.py#L577

See if that gives you enough of a code hint to get started, but if not no worries if you build the library side of things (i.e. functions in IFC4D which allow you to do a "partial import" and "hierarchy extraction" so that the user gets a list of things they can filter by, I can build the Blender side of it - and then you can perhaps continue from there! :)

PR #1842, opened 2021-10-29
Resource Parsing for P6XML

HassanEmam:
I have implemented the ability to parse resources to the IFC4D module for the Primavera XML files.


Moult:
Thanks so much! A few minor comments:

 - would be nice if ScheduleIfcGenerator could use kwargs or a settings object/dict to prevent excessive positional args which decreases code readability
 - @aothms is right that sooner or later we will create a CLI for this - perhaps in a separate `cli.py` file which will provide simple argparse + flags to convert p6xml, p6xer, mspxml, etc. One interface file for all the converters (and later on, gantt chart generators, CPM engines, etc).

aothms:
> use kwargs or a settings object

> is right that sooner or later we will create a CLI for this

Related to that. I quite like the fact that in draw.py we have the settings as a dataclass and can generate also the cli command line parameters from that. I didn't give it a lot of thought and don't want to impose this on the rest of the code base but maybe it's better than a dict? (should be trivial also to go from dataclass to dict and back in the code).

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/draw.py#L303

HassanEmam:
I have create a CLI as a separate file with takes three arguments (IFC file, Schedule File and Schedule Type)
This should be ready soon and I will share the code in the next PR

Moult:
Ah actually instead of a `cli.py` perhaps a `__main__.py` might be better, kinda like how it is implemented in `ifcpatch` and `ifcclash`. That way it can be also easily triggered as a Python module.

HassanEmam:
Yes that is a good idea; I have changed the file and that is how it should run now.
However, we might need to rethink about the return values as currently it does not return anything.
Or should it be adding the schedule data to the ifc file provided.
image

Moult:
it should also allow an `-o` output arg I reckon. Either it adds it to the `-i` input IFC, or if none provided, it creates a fresh IFC and outputs it in `-o`.

I don't think it needs to return anything, it's a CLI after all. If someone uses it as a library perhaps that's another story.

PR #1847, opened 2021-10-31
Fix projection of origin onto camera plane. See #1153

Moult:
So I think I worked it out, but apologies if the C++ looks like an absolute mess so tips would really help.

I guess firstly I think the overall logic is that the projection is drawn using the correct rotation of the camera, but not the right UV (2D) offset. To get the right 2D offset, the 0,0,0 is originally at the top left of the SVG, so we need to 1. Move it to the center (origin) of the camera (by using size->first and second), then 2. Move it again, using the 2D projection of the 3D 0,0,0 on the camera plane.

The first step seems correct, but the 2D projection seems to be where the bug was. The existing logic uses stuff like Extrema_ExtPElS that I don't fully understand, so I implemented it using GeomAPI_ProjectPointOnSurf and getting the UV coords, which might be overkill for a simple perpendicular projection, but I guess it works.

So although it now seems to work with any camera I throw at it in Blender, I left some of your existing code there because I was afraid to delete it. In particular, I wanted the original camera location before its location was changed in this code block:

```
			Extrema_ExtPElS ext;
			ext.Perform(gp::Origin(), *pln, 1.e-5);
			auto P0 = pln->Location();
			pln->SetLocation(ext.Point(1).Value());
```

Hope it makes sense and I didn't make things confusing.

### DIFF
--- a/src/serializers/SvgSerializer.cpp
+++ b/src/serializers/SvgSerializer.cpp
@@ -77,6 +77,7 @@
 #include <HLRBRep_PolyHLRToShape.hxx>
 
 #include <Extrema_ExtPElS.hxx>
+#include <GeomAPI_ProjectPointOnSurf.hxx>
 
 #include "../ifcparse/IfcGlobalId.h"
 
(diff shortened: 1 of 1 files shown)


aothms:
Awesome :)

Does it also work if reuse the calculation from Extrema_ExtPElS? Like `ext.Point(1).Parameter(pu, pv)` because they should be doing the same thing. Just Extrema_ExtPElS algebraicly and only for a plane.

Moult:
You're absolutely right. That's much smarter and more efficient. I'll need to read up on what Extrema_ExtPElS actually does :) I think it's good for merging now.

aothms:
Yeah naming in OCCT is interesting to say the least ;)

But thanks for fighting your way through it. You're solution is spot on. In a way I was doing the same thing, just clumsy (3d difference and then transform to plane coords) and apparently wrong (maybe as you said because I was using the mutated plane).

Thanks a lot!

PR #1851, opened 2021-11-02
Refactor Classification Module (see issue #1711) only ui.py

maxfb87:
So, following the past PR #1833, i refactored ui.py of classification module.
Operator.py and Prop.py use ClassificationsData and ClassificationReferencesData, but they need to be improved, due to the comment written in the past PR.


src/blenderbim/blenderbim/bim/module/classification/operator.py
```diff
@@ -54,9 +55,9 @@ def _execute(self, context):
         ifcopenshell.api.run(
             "classification.add_classification",
             IfcStore.get_file(),
-            **{"classification": Data.library_file.by_id(int(props.available_classifications))},
+            **{"classification": ClassificationsData.data["library_file"].by_id(int(props.available_classifications))},
         )
-        Data.load(IfcStore.get_file())
+        ClassificationsData.load()
```
Moult, line 60: The operator should not call the load() function. Instead, the operators should inherit from an `Operator` class like so: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/owner/operator.py#L27-L31 - then you can also delete call def execute().

src/blenderbim/blenderbim/bim/module/classification/data.py
```diff
@@ -0,0 +1,106 @@
...
+class ClassificationReferencesData:
+    is_loaded = False
+    data = {
+        "products" : {},
+        "references" : {},
+        "library_references" : {},  #never called
+    }
+
+    @classmethod
+    def load(cls,  product_id=None):
```
Moult, line 78: Signature must only be def load(cls), no other arguments allowed. This is because we cannot rely on any intelligence on the code calling the UI to be reloaded. The product is always the active object.

Moult:
I apologise that refactoring the classification module is harder than it seems. Would you mind trying another module? I'll see what I can do about merging and cleaning this up in the future.

maxfb87:
@Moult ok, i'm already working on refactor another simpler module.
I did this PR anyway because i thought that the code could be a useful start for the complete refactoring.
If you think it's not useful at all, feel free to delete the PR, for me it's ok.

PR #1852, opened 2021-11-03
Fix projection boolean for drawing generation. See #1153.

Moult:
I think I fixed it but I was afraid to delete the existing code since I didn't quite understand the original logic. It seems as though the half space created for the boolean cut was invalid (in particular, I don't under the logic of `f` and that causes `auto s = mhs.Solid();` to throw an exception). I didn't understand why these xs, ys, and cut_z were used to create a face when the projection plane was already available (and furthermore the half space reference point uses projection plane anyway) so I simply used the projection plane directly.

Happy to clean up this PR with some hints of what the existing code does :)

### DIFF
--- a/src/serializers/SvgSerializer.cpp
+++ b/src/serializers/SvgSerializer.cpp
@@ -989,7 +989,7 @@ void SvgSerializer::write(const geometry_data& data) {
 							BRepBuilderAPI_MakePolygon mp(points[0], points[1], points[2], points[3], true);
 							auto w = mp.Wire();
 							BRepBuilderAPI_MakeFace mf(w);
-							auto f = mf.Face();
+							TopoDS_Face f = BRepBuilderAPI_MakeFace(projection_plane);
 							gp_Pnt ref = projection_plane.Position().Location().XYZ() + projection_plane.Position().Direction().XYZ();
 							BRepPrimAPI_MakeHalfSpace mhs(f, ref);
 							auto s = mhs.Solid();


aothms:
Several years back we had this gentleman's agreement with occt never to use infinite geometry. A face constructed from a plane is infinite. I don't know to what extent infinite geometry is still an issue in occt, but I prefer to fix the original logic just in case if there is an issue with that.

There is the assumption still that we're dealing with a horizontal plane, that obviously needs to be fixed. So the xs,ys,zs that we get from the bounding box should be projected onto the plane. And then creating a polygon from that.

I'm a bit puzzled about why the halfspace would fail though. I think there is something rather fundamentally going wrong probably specifically with the handling of non-solid cases that we get from shells out of blender bim. Causing the boolean to fail here (maybe because of degerate bbox) and causing the boolean cube thing to only imprint the edges.

Moult:
Not at a computer but I remember checking the xyz of the face and the values didn't make sense to me. Can you replicate perhaps with the parametric column file?

Moult:
Here is a script to replicate the issue. Also, here are the coordinates of the 4 points used to make the face [here](https://github.com/IfcOpenShell/IfcOpenShell/pull/1852/files#diff-946e17589e48b004f872276c583713b5a82259b296010fcd636d14f8c548f03bR989). As you can see, the points are 2D - does this make sense?

Point 0 "gp_Pnt": [0.273015, -0.849975, inf]
Point 1 "gp_Pnt": [2.34802, -0.849975, inf]
Point 2 "gp_Pnt": [2.34802, 1.30003, inf]
Point 3 "gp_Pnt": [0.273015, 1.30003, inf]
Projection location "gp_Pnt": [0.399742, -0.443951, 0.957883]

Test file can be found [here](https://github.com/IfcOpenShell/IfcOpenShell/issues/1153#issuecomment-958599587)

```python
import time
import ifcopenshell
import ifcopenshell.geom

def yield_from_iterator(it):
    if it.initialize():
        while True:
            yield it.get()
            if not it.next():
                break

print("START")
f = ifcopenshell.open("/home/dion/parametric-column-drawing.ifc")

# Setup serialiser
cam = f.by_guid("0mgWeqh1v8khdcw_bhbOXR")
svg_settings = ifcopenshell.geom.settings(APPLY_DEFAULT_MATERIALS=True, DISABLE_TRIANGULATION=True, INCLUDE_CURVES=True)
buffer = ifcopenshell.geom.serializers.buffer()
sr = ifcopenshell.geom.serializers.svg(buffer, svg_settings)
sr.setFile(f)
sr.setElevationRefGuid(cam.GlobalId)
sr.setWithoutStoreys(True)
sr.setPolygonal(True)
sr.setUseNamespace(True)
sr.setAlwaysProject(True)
sr.setScale(1/100)
sr.setAutoElevation(False)
sr.setAutoSection(False)
sr.setPrintSpaceNames(False)
sr.setPrintSpaceAreas(False)
sr.setDrawDoorArcs(False)
sr.setNoCSS(True)

# Setup cache
cache_settings = ifcopenshell.geom.settings(APPLY_DEFAULT_MATERIALS=True, DISABLE_TRIANGULATION=True)
cache = ifcopenshell.geom.serializers.hdf5("cache.h5", cache_settings)

geom_settings = ifcopenshell.geom.settings(APPLY_DEFAULT_MATERIALS=True, DISABLE_TRIANGULATION=True)
geom_settings.set_context_ids([11])
it = ifcopenshell.geom.iterator(geom_settings, f, include=f.by_type("IfcColumn"))
it.set_cache(cache)
for elem in yield_from_iterator(it):
    sr.write(elem)

# Final pass just to make sure the camera is included
# I guess I could group this in the body pass, but for now it makes it easier to comment out code
geom_settings = ifcopenshell.geom.settings(APPLY_DEFAULT_MATERIALS=True, DISABLE_TRIANGULATION=True)
it = ifcopenshell.geom.iterator(geom_settings, f, include=[cam])
for elem in yield_from_iterator(it):
    sr.write(elem)

sr.finalize()
with open("test.svg", "w") as svg:
    results = buffer.get_value()
    # Hack so we can see stuff
    results = results.replace("<defs>", "<defs><style>.cut { fill: black; stroke: black; stroke-linecap: 'round'; stroke-width: 0.35; } .projection { fill: white; stroke: black; stroke-linecap: 'round'; stroke-width: 0.25; }</style>")
    svg.write(results)
```

aothms:
I added 

    sr.setSubtractionSettings(ifcopenshell.ifcopenshell_wrapper.ALWAYS)

and changed CSS to

    results = results.replace("<defs>", "<defs><style>path { stroke: black; fill: gray; fill-rule: evenodd; stroke-linecap: 'round'; stroke-width: 0.35; } .projection path { fill: none; }</style>")

Then result looks good to me :)

afbeelding

with https://github.com/IfcOpenShell/IfcOpenShell/commit/c203b78d343c15bb5f5100f0752e27f540b4b221

Like I said, for legacy reasons (or maybe in general, not sure) I prefer to keep using finite operands over the infinite approach in this PR.

Thanks for pointing at the issue though, the cut_z thing was obviously a remnant from when we only cut at floor plans.

Moult:
Thanks so much and glad I could help narrow the issue :)

PR #1859, opened 2021-11-07
Integrate ifccityjson in BlenderBIM

LaurensJN:
I created a small UI for ifccityjson, because that's easier to use than CLI. 
@Moult could you maybe take a look at the Makefile, if it's correct? Because CJIO has a couple of dependencies I use a venv to make sure everything is installed. Is this okay? Or is it better to use the Wheel/Source file: [https://pypi.org/project/cjio/#files](https://pypi.org/project/cjio/#files)?
Also, I'd like to have the option to import the created IFC file directly in Blender, but I couldn't find out how to do that. Is there a one-line import where you only need to specify the path?

### DIFF
--- a/src/blenderbim/blenderbim/bim/__init__.py
+++ b/src/blenderbim/blenderbim/bim/__init__.py
@@ -43,6 +43,7 @@ modules = {
     "sequence": None,
     "group": None,
     "system": None,
+    "brick": None,
     "structural": None,
     "boundary": None,
     "profile": None,
--- a/src/blenderbim/blenderbim/bim/__init__.py
+++ b/src/blenderbim/blenderbim/bim/__init__.py
@@ -63,10 +63,10 @@ modules = {
     "bimtester": None,
     "diff": None,
     "patch": None,
+    "gis": None,
     "covetool": None,
     "augin": None,
     "debug": None,
-    "gis": None,
 }
 
 
--- a/src/blenderbim/blenderbim/bim/__init__.py
+++ b/src/blenderbim/blenderbim/bim/__init__.py
@@ -43,7 +43,6 @@ modules = {
     "sequence": None,
     "group": None,
     "system": None,
-    "brick": None,
     "structural": None,
     "boundary": None,
     "profile": None,
@@ -66,6 +65,7 @@ modules = {
     "covetool": None,
     "augin": None,
     "debug": None,
+    "gis": None,
 }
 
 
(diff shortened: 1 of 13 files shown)


src/blenderbim/blenderbim/bim/__init__.py
```diff
@@ -66,6 +66,7 @@
     "covetool": None,
     "augin": None,
     "debug": None,
+    "gis": None,
```
Moult, line 69: Perhaps put this above covetool as order determines which UI panel is shown first.

src/blenderbim/blenderbim/bim/module/gis/prop.py
```diff
@@ -0,0 +1,17 @@
...
+
+class BIM_PT_ifccityjson_properties(PropertyGroup):
+    def get_lods(self, context):
+        return [(item.name, "LOD" + item.name, "Level of Detail " + item.name) for item in self.lods]
+
+    input: StringProperty(name="cj_input", default="", subtype='FILE_PATH')
+    output: StringProperty(name="ifc_output", default="", subtype='FILE_PATH')
+    name: StringProperty(name="identifier", default="")
+    split_lod: BoolProperty(name="split_lod", default=True)
+    lods: CollectionProperty(name="lods", type=BIM_PT_existing_lods)
```
Moult, line 15: Use the StrProperty (see where used elsewhere) instead of creating new property group

src/blenderbim/blenderbim/bim/module/gis/prop.py
```diff
@@ -0,0 +1,17 @@
+from bpy.types import PropertyGroup
+from bpy.props import StringProperty, BoolProperty, EnumProperty, CollectionProperty
+
+class BIM_PT_existing_lods(PropertyGroup):
+    name: StringProperty(name="name")
+
+class BIM_PT_ifccityjson_properties(PropertyGroup):
```
Moult, line 7: Can follow similar naming convention? BIMCityJsonProperties

Moult:
For a one line import this is what you're looking for :) https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/test/bim/feature/project.feature#L44

LaurensJN:
Thank you for the review! I changed everything like you said. If you think the Makefile is okay, I think we can merge?

Moult:
Thanks so much :) Great work!

Issue #1676, opened 2021-08-20
BlenderBIM : IfcRelSpaceBoundaries capabilities


CyrilWaechter:
It would be great to enhance IfcRelSpaceBoundary capabilities inside BlenderBIM Add-on with following:

- [ ] Display geometry
- [ ] Display attributes
- [ ] Modify geometry
- [ ] Modify attributes
- [ ] Apply consistent colors based on related building element similarly to [ladybug gbXML viewer](https://www.ladybug.tools/spider/gbxml-viewer/r12/gv-app/gv-app.html)
- [ ] Build test suite to prevent common mistakes or warn about common issues coming from other authoring tools (bimtester report ?)
- [ ] Tools to fix common mistakes from software vendors (ifcpatch + tools to help enhancing attributes data)

It would allow to use BlenderBIM Add-on as:

- Software agnostic IFC energy model auditing tool
- Basic IFC energy model creation tool
- IFC energy model improvement tool

Moult:
Have you got some sample files? This will help dev.

CyrilWaechter:
Yes multiple sample files have been produced during BIMxBEM project: [IfcTestFiles](https://github.com/ENAC-CNPA/BIMxBEM/tree/master/IfcTestFiles)
There is also Aalseth Lane Residence somewhere else.

brunopostle:
@CyrilWaechter since each ifcrelspaceboundary has to be attached to exactly one ifcspace, these external surfaces of external walls don't seem to be supported in IFC (unless we create an ifcspace representing the entire outside world, but with no Representation)

CyrilWaechter:
@brunopostle Hmmm, interesting. We need in some way identify what is `EXTERNAL`, `EXTERNAL_EARTH`, `EXTERNAL_WATER` and `EXTERNAL_FIRE`. Having external spaces might simplify this process and some others. The same way you don't represent your ground with infinite depth external spaces could be defined inside an extended bounding box of the project.

External surfaces of external walls are not in the IFC specifications anyway so nothing prevent you to create another kind of IfcRelSpaceBoundary (not 1st level or 2nd level) to represent it. Thermal breaks are also not defined in IFC so you also need to use an hacky way to store it but this is another story.

aothms:
[...]

PR #1864, opened 2021-11-08
[BlenderBIM] Tools to load IfcRelSpaceBoundary

CyrilWaechter:
# First commit 
* Add a tool at scene level to load all boundaries
* Add a tool at space level to load ones from a specific space or a specific one.
# Second commit
* Add a tool at scene level to colour boundaries based on their related building element

First step to solve #1676


Moult:
Thanks so much and so sorry for not working on boundaries :(

PR #1865, opened 2021-11-08
Improve support for Powerproject import

ecke101:
This is tested with swedish PowerProject 14.0 file


Moult:
Thank you so much :)

PR #1869, opened 2021-11-10
ifcopenshell python element utils, add get layer method

berndhahnebach:
This is what I have been using in the company in BIMTester and what we use in FreeCAD on import of IFC to get the layer of a element.

bernd

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/element.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/element.py
@@ -99,6 +99,24 @@ def get_material(element, should_skip_usage=False):
         return get_material(relating_type, should_skip_usage)
 
 
+def get_layer(element):
+    the_layer = None
+    if (
+        hasattr(element, "Representation")
+        and hasattr(element.Representation, "Representations")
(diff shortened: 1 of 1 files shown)


Moult:
I don't think this'll work with type representations nor mapped representations. Can I instead recommend using a traverse and check for layer assignment on shape representation? (I'm happy to deliberately forego representation item since that can be super expensive).

berndhahnebach:
> 
> ... Can I instead recommend using a traverse and check for layer assignment on shape representation? ... 

How would this work?

aothms:
Here it is in C++. https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcgeom_schema_agnostic/Kernel.cpp#L264

>forego representation item 

With the new handling of inverses it shouldn't be as slow as in 0.6

berndhahnebach:
What could I do to get this or some better code to directly get the layer get merged?

cheers bernd

Moult:
Should be resolved now :) Hope it makes sense @berndhahnebach !

Issue #1866, opened 2021-11-09
Iterator progress always returns 50 with multiprocessing


Moult:
I can't remember if this was borked a long time ago or just recently, but it seems as though calling `it.progress()` doesn't make any sense because the _progress variable is updated during it.initialize().

One solution is to expose the tasks_.size() so that after initialize finishes, i can get_task_size() or something and then create my own progress bar?

Sample code:

```python
import ifcopenshell
import ifcopenshell.geom

def yield_from_iterator(it):
    while True:
        yield it.get()
        if not it.next():
            break

print("START")

f = ifcopenshell.open("/home/dion/Downloads/JudgesReception.ifc")

geom_settings = ifcopenshell.geom.settings()
it = ifcopenshell.geom.iterator(geom_settings, f, 2) # multicore = 2
it.initialize() # progress all happens here, opaque to python?
# Proposed solution: task_size = it.get_task_size()
for elem in yield_from_iterator(it):
    # Proposed solution: progress = count / task_size
    print(it.progress()) # always says 50
```

Moult:
If it helps, I seem to recall this working in v0.6.0.

aothms:
> I seem to recall this working in v0.6.0.

Probably not, in order to maximize code reuse between single and multi threaded and at that time C++ didn't have nice coroutines, currently the main thread blocks in initialize() until all processing is done. I think this has been reported elsewhere. I don't have time to finish this up today, but I'll start a PR to fix this. Feel free to take over :)

This now crashes on calling get() because there isn't a result ready yet. So get() should call future<>::wait(). And process concurrently should probably get some sort of std::list<future<void>> to populate.

Moult:
Confirming as fixed with PR :) Thanks so much!

PR #1873, opened 2021-11-11
async iterator init

aothms:
#1866

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -904,10 +904,6 @@ int main(int argc, char** argv) {
 		Logger::Notice("Using " + std::to_string(num_threads) + " threads");
 	}
 
-	if (!quiet && num_threads > 1) {
-		Logger::Status("Creating geometry...");
-	}
-
 	if (vmap.count("log-file")) {
 		Logger::SetOutput(quiet ? nullptr : &cout_, &log_fs);
 	} else {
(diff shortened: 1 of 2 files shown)


aothms:
Ugh that was a nasty one. I'm really bad a multi threading in C++, but it's actually a slight improvement now, because at least I could eliminate the progress printing functions in iterator, and use the normal one in IfcConvert.

C++ has nice futures and promises, but the stupid thing is that we still have the requirement that we only return tasks that have succeeded, this requires a lot of waiting and because we don't know in advance how many tasks succeed we also can't use vanilla features that well. So in the end I used a synchronized list and a loop in the main thread to catch the elements added to the list. It looks dirty, but maybe it isn't so bad. It does mean there is yet another thread active (not sure if it matters), so there is the N worker threads, a thread to manage those, and a thread to respond to get() and next() calls.

@Moult can you help test a bit and review? Thanks.


~~~
import ifcopenshell
import ifcopenshell.geom

f = ifcopenshell.open("../nibs/duplex/Duplex_MEP_20110907_optimized.ifc")
it = ifcopenshell.geom.iterator(ifcopenshell.geom.settings(), f, num_threads=8)

assert set(it.progress() for _ in ifcopenshell.geom.consume_iterator(it)) == set(range(101))

print("Yay")
~~~

aothms:
@sanderboer hope you're well, since this is your original work also, I thought I'd CC you just in case you want to have a look and offer feedback. It's still your original threadpool, but started using std::async() so that the main thread can watch a std::list of results grow.

aothms:
If no comments, I'm just going to merge this before it goes stale.

Moult:
So sorry, I've been swamped. I will test it :)

PR #1877, opened 2021-11-15
Raise default section height from storey from 1.0 to 1.2

brunopostle:
When creating SVG plans, IfcConvert defaults to sectioning the plan at 1m above storey level.

A kitchen counter height of 0.9m typically results in kitchen window cills higher than 1m, so these windows don't appear in generated plans. This pull request raises the default section height from 1.0m to 1.2m.


aothms:
Thanks

PR #1878, opened 2021-11-16
Added option to generate selection sets

vulevukusej:
I added two new operators to the search ui that allow the user to generate selection sets à la bimcollab-zoom or similar.  I used the  in-built collections system as it meant no additional code was necessary.  Here's how it looks in action:
selectionsets

Some things I want to add at some point in the future:
- Multi-Level selection sets
- Selection sets based on numerical values, grouped according to user specification. For example:  Volume, grouped in intervals of 10m3

Performance isn't great atm - Generating the 2nd pset in the example above took about ~10seconds.  Ideas on how to improve?

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/search/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/search/operator.py
@@ -381,3 +381,106 @@ class ActivateIfcTypeFilter(bpy.types.Operator):
         row = self.layout.row(align=True)
         row.operator("bim.toggle_filter_selection", text="Select All").action = "SELECT"
         row.operator("bim.toggle_filter_selection", text="Deselect All").action = "DESELECT"
+
+
+class CreateIfcClassSelSet(bpy.types.Operator):
+    "Click to generate a selection set for all ifc classes"
+    bl_idname = "bim.create_ifcclass_ss"
+    bl_label = "Create IfcClass Selection Set"
+
+    def execute(self, context):
+        if "Selection Sets" in bpy.data.collections:
+            ss = bpy.data.collections["Selection Sets"]  
+        else:
+            ss = bpy.data.collections.new("Selection Sets")
+            ss.color_tag = "COLOR_04"
+
+        if "Ifc Classes" in bpy.data.collections:
+            ifc_classes = bpy.data.collections["Ifc Classes"]
+        else:
+            ifc_classes = bpy.data.collections.new("Ifc Classes")
+            ifc_classes.color_tag = "COLOR_04"
+            ss.children.link(ifc_classes)
+            
+        objects = context.selectable_objects
+        
+        for obj in objects:
+            ifc_element = tool.Ifc.get_entity(obj)
+            if not ifc_element:
+                continue     
+            
+            ifc_type = ifc_element.is_a()
+            if ifc_type in bpy.data.collections:
+                ifc_class = bpy.data.collections[ifc_type]
+            else:
+                ifc_class = bpy.data.collections.new(f"{ifc_type}")
+                ifc_class.color_tag = "COLOR_04"
+                ifc_classes.children.link(ifc_class)
+            
+            if obj.name not in bpy.data.collections[ifc_type].objects.keys():
+                ifc_class.objects.link(obj)
+            
+            try:#Try to link the collection to the scene, do nothing if already linked
+                bpy.context.scene.collection.children.link(ss)
+            except:
+                pass
+                
+        return {"FINISHED"}        
+    
+   
+class CreatePropSelSet(bpy.types.Operator):
+    "Click to generate a selection set for all unique property values"
+    bl_idname = "bim.create_prop_ss"
+    bl_label = "Create Property Selection Set"
+    
+
+    def execute(self, context):
+        pset = context.scene.BIMSearchProperties.search_attribute_name
+        prop = context.scene.BIMSearchProperties.search_attribute_value
+        
+        if "Selection Sets" in bpy.data.collections:
+            ss = bpy.data.collections["Selection Sets"]  
+        else:
+            ss = bpy.data.collections.new("Selection Sets")
+            ss.color_tag = "COLOR_04"
+
+        if f"{pset}.{prop}" in bpy.data.collections:
+            property = bpy.data.collections[f"{pset}.{prop}"]
+        else:
+            property = bpy.data.collections.new(f"{pset}.{prop}")
+            property.color_tag = "COLOR_04"
+            ss.children.link(property)
+            
+        objects = context.selectable_objects
+        
+        for obj in objects:
+            ifc_element = tool.Ifc.get_entity(obj)
+            if not ifc_element:
+                continue     
+            
+            psets = ifcopenshell.util.element.get_psets(ifc_element)
+            
+            if pset not in psets:
+                continue
+            else:
+                if prop not in psets[pset]:
+                    continue
+                else:
+                    propval = psets[pset][prop]
+                    
+                if propval in bpy.data.collections:
+                    propval_col = bpy.data.collections[propval]
+                else:
+                    propval_col = bpy.data.collections.new(f"{propval}")
+                    propval_col.color_tag = "COLOR_04"
+                    property.children.link(propval_col)
+            
+            if obj.name not in bpy.data.collections[propval].objects.keys():
+                propval_col.objects.link(obj)    
+                
+            try:#Try to link the collection to the scene, do nothing if already linked
+                bpy.context.scene.collection.children.link(ss)
+            except:
+                pass
+                
+        return {"FINISHED"}  
--- a/src/blenderbim/blenderbim/bim/module/search/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/search/operator.py
@@ -420,10 +420,10 @@ class CreateIfcClassSelSet(bpy.types.Operator):
             if obj.name not in bpy.data.collections[ifc_type].objects.keys():
                 ifc_class.objects.link(obj)
             
-            try:#Try to link the collection to the scene, do nothing if already linked
-                bpy.context.scene.collection.children.link(ss)
-            except:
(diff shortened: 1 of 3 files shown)


src/blenderbim/blenderbim/bim/module/search/operator.py
```diff
@@ -381,3 +381,106 @@ def draw(self, context):
...
+            
+        objects = context.selectable_objects
+        
+        for obj in objects:
+            ifc_element = tool.Ifc.get_entity(obj)
+            if not ifc_element:
+                continue     
+            
+            ifc_type = ifc_element.is_a()
+            if ifc_type in bpy.data.collections:
```
Gorgious56, line 413: I don't know for sure that this will have a beneficial effect on performance, but since we can have thousands of objects and thousands of collections, I assume this lookup can be optimised using a set for instance :
```
ifc_classes_set = set()
for obj in objects:
    ifc_element = tool.Ifc.get_entity(obj)
    if not ifc_element:
        continue     
    ifc_classes_set.add(ifc_element.is_a())
for ifc_class in ifc_classes_set:
    ifc_class_col = bpy.data.collections.get(ifc_type)
    if ifc_class_col is None:
        ifc_class_col = bpy.data.collections.new(f"{ifc_type}")
        ifc_class_col .color_tag = "COLOR_04"
        ifc_classes.children.link(ifc_class_col)
for obj in objects: 
    # Again, avoiding costly lookups ? Might have to test this since try/except can generate overhead too
    try:
        ifc_class.objects.link(obj)
    except RuntimeError:  # Will throw if object already in collection
        pass

File: src/blenderbim/blenderbim/bim/module/search/operator.py [hunk 2]
```diff
@@ -381,3 +381,106 @@ def draw(self, context):
...
+                ifc_class = bpy.data.collections[ifc_type]
```
Gorgious56, line 414: Might be interesting to add a prefix or suffix to avoid name collisions, it may be safe to assume that users will be able to create collections named after ifc classes in other contexts.

src/blenderbim/blenderbim/bim/module/search/operator.py
```diff
@@ -381,3 +381,106 @@ def draw(self, context):
...
+                ifc_class = bpy.data.collections.new(f"{ifc_type}")
+                ifc_class.color_tag = "COLOR_04"
+                ifc_classes.children.link(ifc_class)
+            
+            if obj.name not in bpy.data.collections[ifc_type].objects.keys():
+                ifc_class.objects.link(obj)
+            
+        try:#Try to link the collection to the scene, do nothing if already linked
+            bpy.context.scene.collection.children.link(ss)
+        except:
```
Gorgious56, line 425: `except RuntimeError:` :)

src/blenderbim/blenderbim/bim/module/search/operator.py
```diff
@@ -381,3 +381,106 @@ def draw(self, context):
...
+    "Click to generate a selection set for all unique property values"
+    bl_idname = "bim.create_prop_ss"
+    bl_label = "Create Property Selection Set"
+    
+
+    def execute(self, context):
+        pset = context.scene.BIMSearchProperties.search_attribute_name
+        prop = context.scene.BIMSearchProperties.search_attribute_value
+        
+        if "Selection Sets" in bpy.data.collections:
```
Gorgious56, line 441: I think it's more pythonic and it may be more performant to use `get`
```
ss = bpy.data.collections.get("Selection Sets")  # Returns None if collection is not found
if ss is None:
    ss = bpy.data.collections.new("Selection Sets")
    ss.color_tag = "COLOR_04"
```

src/blenderbim/blenderbim/bim/module/search/operator.py
```diff
@@ -381,3 +381,106 @@ def draw(self, context):
...
+            ss = bpy.data.collections["Selection Sets"]  
+        else:
+            ss = bpy.data.collections.new("Selection Sets")
+            ss.color_tag = "COLOR_04"
+
+        if f"{pset}.{prop}" in bpy.data.collections:
```
Gorgious56, line 447: Same as https://github.com/IfcOpenShell/IfcOpenShell/pull/1878/files#r750483459

src/blenderbim/blenderbim/bim/module/search/operator.py
```diff
@@ -381,3 +381,106 @@ def draw(self, context):
...
+            
+            if pset not in psets:
+                continue
+            else:
+                if prop not in psets[pset]:
+                    continue
+                else:
+                    propval = psets[pset][prop]
+                    
+                if propval in bpy.data.collections:
```
Gorgious56, line 471: Same as https://github.com/IfcOpenShell/IfcOpenShell/pull/1878/files#r750483459

src/blenderbim/blenderbim/bim/module/search/operator.py
```diff
@@ -381,3 +381,106 @@ def draw(self, context):
...
+                    propval_col = bpy.data.collections[propval]
+                else:
+                    propval_col = bpy.data.collections.new(f"{propval}")
+                    propval_col.color_tag = "COLOR_04"
+                    property.children.link(propval_col)
+            
+            if obj.name not in bpy.data.collections[propval].objects.keys():
```
Gorgious56, line 478: try/except _might_ be faster with large datasets

src/blenderbim/blenderbim/bim/module/search/operator.py
```diff
@@ -381,3 +381,106 @@ def draw(self, context):
...
+                propval_col.objects.link(obj)    
+                
+        try:#Try to link the collection to the scene, do nothing if already linked
+            bpy.context.scene.collection.children.link(ss)
+        except:
+            pass
+                
+        return {"FINISHED"}
```
Gorgious56, line 486: Traditionnaly files require a newline as a last character, this is why there is this warning icon :) Easy to miss but if you're using a linter it should fix it automatically
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

Gorgious56:
I've gone ahead and added some comments regarding possible improvements, feel free to follow or discard :)
One gotcha with collections is name collisions and user interaction, if you want to be sure everything works even if the user moves the collections or renames everything, you could use a [CollectionProperty](https://docs.blender.org/api/current/bpy.props.html?highlight=collectionproperty#bpy.props.CollectionProperty) 

(example)
```
class CollectionHolder(bpy.types.PropertyGroup):
    collection: bpy.props.PointerProperty(type=bpy.types.Collection)
    
bpy.utils.register_class(CollectionHolder)

bpy.types.Scene.collections = bpy.props.CollectionProperty(type=CollectionHolder)

new_col_holder = bpy.context.scene.collections.add()
new_col_holder.collection = bpy.data.collections.new("My New Collection")

bpy.context.scene.collection.children.link(new_col_holder.collection)
```

Then you can rename, move, etc your collections, since it's a pointer property the reference will stay. Also you don't have to re-fetch everything each time you execute the operator.

Moult:
Without reading the code, a huge speedup I discovered when writing the importer was to first link objects into collections, but only at the very end, after all objects were categorised, to link the collection into the view layer. Apparently if you update collections whilst its in the view layer it triggers a bunch of stuff.

For example: https://github.com/IfcOpenShell/IfcOpenShell/blob/fae607f8ce69496e91c400a6c1abd523fe0c52db/src/blenderbim/blenderbim/bim/import_ifc.py#L1043

vulevukusej:
> This is great ! I'll look at the commits and throw ideas for performance if you're open to it :)

Nice! You've given me some homework to do :)

vulevukusej:
> Without reading the code, a huge speedup I discovered when writing the importer was to first link objects into collections, but only at the very end, after all objects were categorised, to link the collection into the view layer. Apparently if you update collections whilst its in the view layer it triggers a bunch of stuff.
> 
> For example:
> 
> https://github.com/IfcOpenShell/IfcOpenShell/blob/fae607f8ce69496e91c400a6c1abd523fe0c52db/src/blenderbim/blenderbim/bim/import_ifc.py#L1043

I see - sounds like that might help.

src/blenderbim/blenderbim/bim/module/search/operator.py
vulevukusej: thanks for that, I had no idea, my linter was set to none in vscode.  I just turned it on and BAM!, so many warnings 😅

Gorgious56:
Yeah this is a very good point, changing anything in a view layer's `layer_collection` system triggers the re-calculation of every object of every collection that's included in the view layer (modifiers, constraints, drivers, etc.). That includes linking a new collection to a view layer.

vulevukusej:
@Gorgious56 @Moult -  Since we may have existing collections in the view layer, do you also recommend removing the selection sets from the view layer at the start of the code, and then re-adding them at the end?

Moult:
Have you considered not using collections at all? How about a generic grouping mechanism that can be converted to and from an IFC group? It'll be faster and won't break export  syncing.

Moult:
I should probably now mention I hope to see the search module totally oberhauled to support creating a sort of query builder with and and or support, and saving search queries, generically grouping or colouring query results, and round tripping all this in IFC groups.

vulevukusej:
> I should probably now mention I hope to see the search module totally oberhauled to support creating a sort of query builder with and and or support, and saving search queries, generically grouping or colouring query results, and round tripping all this in IFC groups.

:) That all sounds rather neat, perhaps we should scrap this pull request then?

Moult:
How about we schedule a group call to design it together?

vulevukusej:
> How about we schedule a group call to design it together?

Yeah i'm up for that.  Admittedly I only just now looked into IfcGroup and I realized that it is precisely what is needed for this usecase.

vulevukusej:
Closed -  To be integrated into a different PR with a more suitable design concept.

PR #1884, opened 2021-11-18
Fix build-all when use current python version on ubuntu

Amoki:


Amoki:
@aothms what do you think about this PR?

aothms:
Sorry, for the delay, I didn't really get the chance to digest this.

When I'm personally on ubuntu, I always install in `sys.getusersitepackages()` which is writable for the current user, it typically doesn't exist yet but can be created, and it is included in sys.path (I assume, since it can be imported afterwards). It is specific to the current user though.

My impression is that these changes would require the build-all.py script to be ran as root, which we should probably try to avoid, we don't want to later have one of those famous `rm -rf /` bugs.

Amoki:
There is no changes when using a virtualenv, it is a fix for installations without a virtualenv. Within a virtualenv `python_info["purelib"]` is a path in it and always exists. On Ubuntu (and Debian), `site_packages` is replaced by `dist_packages` so lib installed through `apt` don't conflict with libs installed by `pip`. When used without a virtualenv with the distrib's python (what we do with docker), it triggers a crash. When using the distrib python without a venv, root permissions are already needed.

aothms:
Coincidently I'm using your code from the previous PR now for the first time and at least on my end it doesn't seem to configure correctly.

You're passing the site-packages (or dist-packages with this PR) to the `python_library` variable in the script.  But as far as I can tell CMake expects that to be the path to `libpythonX.Y.so`.  See below.

Where to install the module is actually managed in CMake https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcwrap/CMakeLists.txt#L71-L81

~~~
import os, sysconfig

# You're passing this as python_library
sysconfig.get_paths()["purelib"]

# I would expect this as python_library
os.path.join(
    sysconfig.get_config_var('LIBDIR'), 
    sysconfig.get_config_var('multiarchsubdir').replace("/", ""), 
    sysconfig.get_config_var("INSTSONAME")
)
~~~

aothms:
superseded by #1901

PR #1896, opened 2021-11-26
bulk property editor

vulevukusej:
Here's a first draft of the bulk property editor.  Still to do: incorporate the method introduced in commit (https://github.com/IfcOpenShell/IfcOpenShell/commit/a1817e8a61f849c83f71384a8b03f836814b3763)


src/blenderbim/blenderbim/bim/module/pset/__init__.py
```diff
@@ -29,12 +29,19 @@
     operator.GuessQuantity,
     operator.RemovePset,
     operator.TogglePsetExpansion,
+    operator.AddPropertyToEdit,
```
Gorgious56, line 32: Maybe we should start using Blender's recommended operator names ? `BIM_OT_snake_case_op_name`

src/blenderbim/blenderbim/bim/module/pset/operator.py
```diff
@@ -374,3 +374,147 @@ def _execute(self, context):
...
+
+class AddPropertyToEdit(bpy.types.Operator):
+    bl_label = "Add new row"
+    bl_idname = "bim.add_property_to_edit"
+    bl_options = {"REGISTER", "UNDO"}
+    type: bpy.props.StringProperty()
+
+    def execute(self, context):
+        if self.type == "PropertiesToRename":
+            props = context.scene.PropertiesToRename
```
Gorgious56, line 387: Since `type` string value is the same as the scene's custom prop name, maybe doing `getattr(context.scene, self.type).add()` is simpler ? It also makes operators easier to extend if somehow you want to add other modes later. Matter of preference :)

src/blenderbim/blenderbim/bim/module/pset/operator.py
```diff
@@ -374,3 +374,147 @@ def _execute(self, context):
...
+
+    def execute(self, context):
+        return IfcStore.execute_ifc_operator(self, context)
+
+    def _execute(self, context):
+        self.file = IfcStore.get_file()
+        selected_objects = context.selected_objects
+        props = context.scene.PropertiesToAddOrEdit     
+
+        for object in selected_objects:
```
Gorgious56, line 480: It's good practice to avoid overriding python builtins (`object`,  `type`, `min`, `max`, etc.) to save headaches down the line : https://docs.python.org/3/library/functions.html , that's why you usually see `for obj in context.select_objects` for instance. If you're using a python linter the builtins are usually colored differently so you can differentiate them

src/blenderbim/blenderbim/bim/module/pset/prop.py
```diff
@@ -173,3 +173,26 @@ class WorkSchedulePsetProperties(PropertyGroup):
...
+
+class PropertiesToRename(PropertyGroup):
+    pset_name: StringProperty(name="Pset")
+    existing_property_name: StringProperty(name="Existing Property Name")
+    new_property_name: StringProperty(name="New Property Name")
+    
+    
+class PropertiesToAddOrEdit(PropertyGroup):
+    pset_name: StringProperty(name="Pset")
+    property_name: StringProperty(name="Property")
```
Gorgious56, line 186: Maybe it would be simpler to re-use the [Attribute](https://github.com/IfcOpenShell/IfcOpenShell/blob/3accc1dbd6510d176455e8f163fdbfe157762bdc/src/blenderbim/blenderbim/bim/prop.py#L160-L213) property group ?

src/blenderbim/blenderbim/bim/module/pset/ui.py
```diff
@@ -375,3 +375,92 @@ def draw(self, context):
...
+    def draw(self, context):
+        layout = self.layout
+        props = context.scene.PropertiesToAddOrEdit 
+        
+        if props:
+            for index, property in enumerate(props):
+                    row = layout.row()
+                    row.prop(property, "pset_name", text="")
+                    row.prop(property, "property_name", text="")
+                    if property.value_type == "String":
```
Gorgious56, line 444: Similar to previous comment : Maybe worthwhile to explore the [Attribute](https://github.com/IfcOpenShell/IfcOpenShell/blob/3accc1dbd6510d176455e8f163fdbfe157762bdc/src/blenderbim/blenderbim/bim/prop.py#L160-L213) property group : you can then just do `row.prop(property, property.get_value_name(), text="")` and keeping the `value_type` enum to dynamically change it afterwards with an `update` method to change the attribute `data_type`. https://docs.blender.org/api/current/bpy.props.html#update-example

vulevukusej:
@Gorgious56  - some great tips as always ;)

src/blenderbim/blenderbim/bim/module/pset/__init__.py
vulevukusej: Agreed.

src/blenderbim/blenderbim/bim/module/pset/operator.py
vulevukusej: Agreed

src/blenderbim/blenderbim/bim/module/pset/ui.py
vulevukusej: Yeah, sounds good.

src/blenderbim/blenderbim/bim/module/pset/prop.py
vulevukusej: Sure - i don't mind trying that.

src/blenderbim/blenderbim/bim/module/pset/operator.py
vulevukusej: Agreed.

Moult:
Let me know when this is ready for merging :) Love the code review!

vulevukusej:
Review comments from @Gorgious56 were incorporated.  This commit relies on Issue #1909 being resolved.

src/blenderbim/blenderbim/bim/module/pset/operator.py
```diff
@@ -374,3 +373,114 @@ def _execute(self, context):
...
+
+class BIM_OT_clear_list(bpy.types.Operator):
+    bl_label = "Clear list of properties"
+    bl_idname = "bim.clear_list"
+    bl_options = {"REGISTER", "UNDO"}
+    option: bpy.props.StringProperty()
+
+    def execute(self, context):
+        getattr(context.scene, self.option).clear()
+        return {"FINISHED"}
```
vulevukusej, line 409: I could probably combine the following 3 operator: (BIM_OT_add_property_to_edit, BIM_OT_remove_property_to_edit, BIM_OT_clear_list) into a single operator and just add another attribute "action" that expects either "add, delete or clear".

vulevukusej:
So I added another little feature just for fun :) - This is something I use often when cleaning up ifc files I receive.  
image

vulevukusej:
Following changes were made:

1. The dropdown for data types is now an enumeration of primary_measure_type: image
2. I moved the list of primary_measure_type to blenderbim.bim.prop so that I can re-use it in the pset and pset_template modules.  I also made it a constant - not sure if this is ok with you?
3. The new value is now explicitly passed to "pset.edit_pset" - See lines 471-486 in operator.py.

@Moult @Gorgious56  -  If tested this and I'm happy with the functionality.  Is there anything you'd like me to adjust?

src/blenderbim/blenderbim/bim/module/pset/operator.py
Gorgious56: You can do that but in my experience it tends to backfire and makes for more complicated operators when you have to extend them down the line :)

Gorgious56:
I didn't test it but I think code wise it's pretty clean. :) I can't say much about the functionality though since I didn't use it but it looks useful for managing large data sets !

vulevukusej:
@Moult did you get a chance to look at this mate?

Issue #1897, opened 2021-11-26
Segfault on IfcConvert 0.7.0


femesq:
I'm having segfault converting this one: http://openifcmodel.cs.auckland.ac.nz/Model/Details/348, and would be happy to try to try/help fix it.

Also saw this one https://github.com/IfcOpenShell/IfcOpenShell/issues/1895, so I tried to run singled-threaded, with no success...

It happens with both `dae`:
```
# IfcConvert -v --threads 1 20200820IFC4_Convenience_store_Renga_4.1.ifc 20200820IFC4_Convenience_store_Renga_4.1.dae
IfcOpenShell IfcConvert - (OCC 7.5.3)
Scanning file...
Done scanning file
Parsing input file took 1 second
Creating geometry...
[################                                  ]Segmentation fault
```
... and `glb` output formats:
```
# IfcConvert -v --threads 1 20200820IFC4_Convenience_store_Renga_4.1.ifc 20200820IFC4_Convenience_store_Renga_4.1.glb
IfcOpenShell IfcConvert - (OCC 7.5.3)
Scanning file...
Done scanning file
Parsing input file took 2 seconds
Creating geometry...
[################                                  ]Segmentation fault
```

Is there any troubleshooting steps guide that I can follow to identify the possible causes?

PS.:
I merged from `0.7.0` branch yesterday, so my commit is this one: https://github.com/IfcOpenShell/IfcOpenShell/commit/d7b27c0620a6beea4725eb4c6d1f79a1f8d19297

aothms:
Thanks. It was nesting that wasn't correctly handled in this case.

> Is there any troubleshooting steps guide that I can follow to identify the possible causes?

In general what is always beneficial is running IfcConvert with -vv to get  detailed output on what it is currently doing.

And running in valgrind/gdb to get a stacktrace.

femesq:
Thanks @aothms!

Can confirm it is fixed after rebuilding with current commit (https://github.com/IfcOpenShell/IfcOpenShell/commit/c5e6c3e2c7cfb980f41737b53e5dd6b4438bc644).

Also made a very tiny PR for making the `-vv` option visible as command-line parameter..

PR #1898, opened 2021-11-26
Added information for debugging level parameter

femesq:
Registering `-vv` option for IfcConvert debugging level (https://github.com/IfcOpenShell/IfcOpenShell/issues/1897#issuecomment-979971951).


aothms:
Thanks! Good idea.

PR #1899, opened 2021-11-27
Allow to manually update boundary connection geometry

CyrilWaechter:
Following have been created in this purpose:
* usecase in ifcopenshell.api
* blenderbim operator
* button in blenderbim manually trigger it


CyrilWaechter:
Related to #1676

PR #1901, opened 2021-11-28
Build all improvements 2

aothms:
See #1884


aothms:
@Amoki thoughts?

aothms:
Perhaps `multiarchsubdir` should be checked whether it is non-empty/exists?

Amoki:
It also fixes my problem!
LGTM for mee

aothms:
thanks for the feedback

Issue #1904, opened 2021-11-30
Optimise get inverse length check


Moult:
Upon further investigation I believe this is a separate issue to #1807 :)

I've noticed that authoring is slower with large files. One major culprit seems to be due to calling get_inverse() in the update_owner_history API usecase. This usecase is indirectly called by many other usecases, so performance improvements here will see performance improvements across the rest of the API.

In short, very often in the API usecases (I count at least 23 usecases with a quick grep out of 250) I am interested in whether the number of inverses of an entity are 0, 1, or >1. I'm not interested in exactly what they are, just the quantity (and only >1, not the exact number). This check is useful in deciding whether the usecase should edit the object directly, or whether it is referenced elsewhere and the usecase should make a copy of it and edit the copy. The check also helps me decide whether or not the usecase should remove a reference to it, or if there are no more references and the usecase should do a deep remove.

When I am only interested in 0, 1, or >1, I think the full get_inverse() call is needlessly expensive. Especially when checking inverses of "popular" elements. For example, in a 275mb file I'm working on, an owner history has >100,000 inverses.

I don't know if there is a cleaner solution, but as an attempt, I'm proposing a new function called `get_inverse_cardinality` which returns either 0, 1, or 2 (to represent >1).

Here's an example of running it 10 times (e.g. if I do a bulk edit function in Blender on 10 objects):

```
>>> timeit.timeit('x = f.get_inverse(o)', setup='import blenderbim.tool as tool; f = tool.Ifc.get(); o = f.by_type("IfcWall")[0].OwnerHistory', number=10)
15.806816399000127

>>> timeit.timeit('x = f.get_inverse_cardinality(o)', setup='import blenderbim.tool as tool; f = tool.Ifc.get(); o = f.by_type("IfcWall")[0].OwnerHistory', number=10)
2.8671999643847812e-05
```

I've submitted a PR for the proposed code (which I have no idea if it is correct or not) and hopefully this function is useful in general.

PR #1905, opened 2021-11-30
Experimental get_inverse_cardinality proposal. See #1904.

Moult:


Moult:
Commenting to link PR to #1904.

aothms:
Makes total sense, but is also a bit unfortunate that this is necessary. I have a quick request.

Could you see if similar results are obtained if you just implement it in SWIG while reusing the existing function.

~~~diff
diff --git a/src/ifcwrap/IfcParseWrapper.i b/src/ifcwrap/IfcParseWrapper.i
index 93fc9514..4dbf56f0 100644
--- a/src/ifcwrap/IfcParseWrapper.i
+++ b/src/ifcwrap/IfcParseWrapper.i
@@ -85,10 +85,15 @@ static IfcUtil::ArgumentType helper_fn_attribute_type(const IfcUtil::IfcBaseClas
        IfcUtil::IfcBaseClass* by_guid(const std::string& guid) {
                return $self->instance_by_guid(guid);
        }
+
        aggregate_of_instance::ptr get_inverse(IfcUtil::IfcBaseClass* e) {
                return $self->getInverse(e->data().id(), 0, -1);
        }

+       size_t get_inverse_cardinality(IfcUtil::IfcBaseClass* e) {
+               return IfcParse_IfcFile_get_inverse($self, e)->size();
+       }
+
        void write(const std::string& fn) {
                std::ofstream f(IfcUtil::path::from_utf8(fn).c_str());
                f << (*$self);

~~~

There's two reasons I ask this. It could be that the bottleneck you experienced is solely in Python (creating all the entity_instance instances and all). It could also be that given this code, your compiler will generate an optimal version similar to what you wrote by hand (although it's not a given, considering all the shared_ptr and vector stuff that goes on behind the scenes, heap stuff doesn't always get optimized away very well, but they are smart, smarter than us).

If there's a still a large performance gap between my proposal and yours, I'll happily merge yours.

aothms:
Ok, I now see there is another difference in that you short-circuit the compilation once you reach n > 1. I'm not sure this is wise, or at least we should rename the function in this case, because given the function name people might expect an accurate number.

If you really want to squeeze performance out till the last drop an alternative is that you create an additional mapping `instance_id -> num_inverses`. But that's a bit of code to maintain (not huge effort).

A precise num_inverses might be beneficial for things outside of your aim, such as graph analysis.

src/ifcparse/IfcParse.cpp
```diff
@@ -2387,6 +2387,25 @@ aggregate_of_instance::ptr IfcFile::getInverse(int instance_id, const IfcParse::
...
 }
 
+
+int IfcFile::getInverseCardinality(int instance_id) {
+	auto lower = byref_excl.lower_bound({ instance_id,-1,-1 });
+	auto upper = byref_excl.upper_bound({ instance_id, std::numeric_limits<int>::max(), std::numeric_limits<int>::max() });
+
+	int total_references = 0;
+	for (auto it = lower; it != upper; it++) {
+		for (auto& i : it->second) {
```
aothms, line 2397: besides the more general comment. the inner loop can be replaced with `total_references += it->second->size()`.

Moult:
Yes indeed I shortcutted it because I thought that's where the performance bottleneck was. What function name might you recommend?

I tested out the SWIG implementation and got this:

```
>>> timeit.timeit('x = f.get_inverse(o)', setup='import blenderbim.tool as tool; f = tool.Ifc.get(); o = f.by_type("IfcWall")[0].OwnerHistory', number=10)
3.2867554590011423

>>> timeit.timeit('x = f.get_inverse_cardinality(o)', setup='import blenderbim.tool as tool; f = tool.Ifc.get(); o = f.by_type("IfcWall")[0].OwnerHistory', number=10)
0.07737603299756302
```

Why this time the first one took 3.2 seconds and not 15s like it did in the past, I have no idea. This was a fresh Blender session, and perhaps the old one wasn't? In any case, the SWIG implementation is still faster, but interestingly still magnitudes slower than the one where I return from the loop early. How fast is fast enough? Not sure. Recently, I've been doing a lot of cleanup scripts on files, each around 200-600mb, editing almost every single element in each file, so the speed boost means a lot. Editing a thousand elements in a file which is quite realistic, means an extra 7.73 unnecessary seconds. But maybe my usecase is too niche to warrant this unfortunate function? Or maybe it isn't niche - especially as this bonus can be used throughout the API and who knows what people will use the API for in the future?

Your call on how to proceed :) I like the speed boost, but that could be just selfish.

Moult:
Aah I think this time it took 3.2 seconds because in the past when I was debugging in C++ I might've had some std::cout statements (sorry, still learning). So, 3.2s is the more accurate number, not 15s.

aothms:
yes writing to cout is very slow also in python

0.077/ 2.87e-05 is a 2.5e3 difference. That's hard to ignore. Let's keep your function then, but I don't think there is a reason to short circuit (especially if you follow my inline code comment suggestion). 

I just realized btw, that in the case of `byref_excl` the 2nd and 3rd element of the tuple (key of the map) is actually never read (it's always lowerbound/upperbound with 2nd and 3d [-1, int_max]). So for that map we can go back to just a key of int, which means that in your inverse cardinalities function there is no need for any loop or lower/upper bound, you can then simply go back to `byref_excl[instance_id].size()`.

Moult:
I had to read your third paragraph a few times - I'll try it out tomorrow (about to go AFK) and modify this PR. Should I also add a new function to get the size in swig? Perhaps it is useful in the future? Like `get_total_inverses()` which returns size() and `get_inverse_cardinality` (or a better name, which I can't think of right now) which does the 0, 1, 2 thing?

Also, I saw this line of code and I don't understand it - https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcparse/IfcParse.cpp#L2262 - where is `by_ref_cached_` used? I see it cleared and erased, but never used?

aothms:
I would only implement the size and do the >1 check all the way in your python code. It won't be slower (especially if we don't need to do lower/upper bound anymore)

> a few times

Sorry. 

> where is by_ref_cached_ used?

Good question. We had it in 0.6 when computing inverses was much slower, but it appears not be read anymore. It should be removed. https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcparse/IfcParse.cpp#L2089

Moult:
I think I did what you had in mind and it seems to work:

```
>>> timeit.timeit('x = f.get_inverse(o)', setup='import blenderbim.tool as tool; f = tool.Ifc.get(); o = f.by_type("IfcWall")[0].OwnerHistory', number=10)
3.2222955859906506

>>> timeit.timeit('x = f.get_total_inverses(o)', setup='import blenderbim.tool as tool; f = tool.Ifc.get(); o = f.by_type("IfcWall")[0].OwnerHistory', number=10)
2.543299342505634e-05
```

Please don't trust any of my code and I feel a bit uneasy about some of these changes and don't know if I messed it up. For example this line seems as though it used to use the 2nd and 3rd elements of the tuple: https://github.com/IfcOpenShell/IfcOpenShell/pull/1905/files#diff-b040b6c3201487d58cce8bdf084d9299d961719c7f722ca4802855a2f3daa818L1005

Fingers crossed!

Moult:
Hmm, the IfcOpenShell test cases segfault. I have no idea then what I'm doing wrong :)

src/ifcparse/IfcParse.cpp
```diff
@@ -2124,15 +2110,8 @@ void IfcFile::process_deletion_() {
...
 					}
 					{
-						auto lower = byref_excl.lower_bound({ name,-1,-1 });
-						auto upper = byref_excl.upper_bound({ name, std::numeric_limits<int>::max(), std::numeric_limits<int>::max() });
-
-						for (auto byref_it = lower; byref_it != upper; ++byref_it) {
-							auto& ids = byref_it->second;
-							ids.erase(std::remove(ids.begin(), ids.end(), id), ids.end());
-						}
+						byref_excl.erase(name);
```
aothms, line 2113: The issue is here. What's being erased previously wasn't `name`, but rather `id` being erased from the references of `name`. Try this:

						auto byref_it = byref_excl.find(name);
						if (byref_it != byref_excl.end()) {
							auto& ids = byref_it->second;
							ids.erase(std::remove(ids.begin(), ids.end(), id), ids.end());
						}

Moult: You're right! Thanks and fixed :)

Moult:
PR should be good to go!

Issue #1807, opened 2021-10-19
Optimising bulk element copying


Moult:
This is a bit of a difficult one to describe, but I'll try anyway as I'm a bit stumped and I need help :) This is similar to #1376 but instead it's about copy_deep.

I've been struggling to optimise copying a bunch of objects. In the Blender environment, a copy process does two things:

 1. Run the copy class usecase: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/api/root/copy_class.py this copies most direct and indirect attributes, but importantly, it does _not_ copy geometry.
 2. Run the add representation usecase: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/api/geometry/add_representation.py - in my tests, I found that it was faster to add a new representation from scratch from Blender as opposed to doing a copy deep of the previous representation. This baffles me, but it seems to be true.

Here's a snippet profiling the copy operation of 32 column objects (24 verts each) with 2 property sets full of properties. It takes **15.75 seconds** in total. This is a  painfully long time. Note that this occurs only in an existing medium sized file from Tekla which already has ~200,000 lines of STEP. In a fresh file, on a simple cube (8 verts) with no psets, it takes only 0.4 seconds.

(I cull everything with a cumtime < 1)

```
finished in 15.743012428283691
Tue Oct 19 18:47:55 2021    blender.prof

         21883612 function calls (19515989 primitive calls) in 15.752 seconds

   Ordered by: cumulative time
   List reduced from 436 to 50 due to restriction <50>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   15.752   15.752 {built-in method builtins.exec}
        1    0.000    0.000   15.752   15.752 <string>:1(<module>)
      2/1    0.000    0.000   15.752   15.752 /home/dion/drive/blender/blender-2.93/2.93/scripts/modules/bpy/ops.py:115(__call__)
      2/1    0.001    0.000   15.743   15.743 {built-in method _bpy.ops.call}
        1    0.000    0.000   15.743   15.743 /home/dion/.config/blender/2.93/scripts/addons/blenderbim/bim/module/geometry/operator.py:388(execute)
[...]

PR #1912, opened 2021-12-02
#1807 method mapping proposal

aothms:


aothms:
> E           KeyError: 'IFC4.IfcIdentifier'

Ah of course it only works for entities :pensive: I'll take a 2nd look

PR #1924, opened 2021-12-07
Extend ifc debug panel

Gorgious56:
New operator to select meshes by percentile. eg with a percentile of 90 if the densest mesh has 10000 polygons, all meshes with more than 9000 polygons will be selected. I found this helps figuring out the few occasionnal culprits that make export times very long. It does not take evaluated modifiers into account but it could with a little bit of tweaking. 

Could be extended to for instance instead select like the 10 most dense meshes ? Food for thought.

BSE_69

Console output :
```
Selected all Meshes with more than 445 polygons
Selected all Meshes with more than 140 polygons
Selected all Meshes with more than 75 polygons
Selected all Meshes with more than 10 polygons
Selected all Meshes with more than 0 polygons
```

Add tooltips to most operators.

Prevent bug if creating shape from non geometric ifc element.

Print number of failures when testing all shapes.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/debug/__init__.py
+++ b/src/blenderbim/blenderbim/bim/module/debug/__init__.py
@@ -28,6 +28,7 @@ classes = (
     operator.CreateAllShapes,
     operator.CreateShapeFromStepId,
     operator.SelectHighPolygonMeshes,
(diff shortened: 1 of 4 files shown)


Moult:
Beautiful feature and very useful! Also useful for analysing slow imports and bloated exports from other software.

PR #1927, opened 2021-12-08
fix mpfr checking problem

Ttofuuu:
https://github.com/BrianGladman/mpir.git 's mpfr will lead to a merge problem and make IFCOS fail to build depends on Windows. Change the link to anothms will fix this problem.


aothms:
Thanks, the PR included a commit that reverts your work, but I cherry-picked your first commit.

PR #1887, opened 2021-11-19
OCC7.5 to OCC7.6 -> depracated Poly_Triangulation Nodes()

bimworld:

PR #1930, opened 2021-12-10
OCC7.5 to OCC7.6 -> depracated Poly_Triangulation Nodes()

aothms:
#1887 applied to v0.7.0


Moult:
By the way I recently discovered that you can edit the merge base branch on PRs, not sure if that is applicable here but thought I'd mention it :)

aothms:
@Moult oh wow thanks that, that's hard to discover. yes that would have saved me a couple of clicks today...

aothms:
Thanks @e3dEF I moved your commit to 0.7 where we now have occt 7.3 from debian package manager and it works flawlessly.

PR #1936, opened 2021-12-10
Fix #1717 : Disable task editing if task was deleted as a result of deleting its parent

### LINKED ISSUE #1717: BlenderBim WorkSchedule : Removing a parent task while editing a children doesn't disable task edition
Here are two gifs of the problem :
BSE_46
BSE_47

Removing a parent task from a Work Schedule deletes all its children. However if one of its children is being edited at the same time, the "editing state" is never disabled, leaving the UI List in a broken state since we need to access the now-deleted task to disable it.

I don't know how to fix this since I understand there can be complicated hierarchical relationships between tasks.
So either maybe :

- Prevent removing parent tasks while a children is being edited
- Force-disable the "editing state" when the currently-edited task has been removed
- Display the "Disable Editing" button at all times, not only when the currently edited task is selected

Depending on how easy it is to fetch hierarchical relationships between tasks using the API,

Gorgious56:
Following   #1717
I elected to disable editing if the edited task has been removed, although maybe there are some more intricate dependencies that make it not ideal ?

**Basically :**

Store currently edited task before removing a task.
Check after removing the task if the previsously edited task still exists, if not, disable editing it to prevent softlock.

@Moult while testing this I noticed the behaviour is a bit weird when adding a new task to the list, usually when adding an element in a list (at least in Blender) this element becomes active, here it does not. Do you reckon it would be interesting to change the behaviour to select the element that's just been added in the list ? I can work on it if needed.

Also I'm wondering if you think it's doable to write unit tests for this, since I remember you saying testing UI is hard.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/sequence/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/sequence/operator.py
@@ -523,14 +523,13 @@ class RemoveTask(bpy.types.Operator):
         Data.load(self.file)
         bpy.ops.bim.enable_editing_tasks(work_schedule=props.active_work_schedule_id)
 
-        for i, task in enumerate(context.scene.BIMTaskTreeProperties.tasks):
-            if (
-                task.ifc_definition_id == props.active_task_id
-                and self.file.by_id(props.active_task_id) == active_task_ifc
-            ):  # Task was not deleted
-                break
-            if i == len(context.scene.BIMTaskTreeProperties.tasks) - 1:  # Task was deleted
-                bpy.ops.bim.disable_editing_task()
+        if not any(
+            task
+            for task in context.scene.BIMTaskTreeProperties.tasks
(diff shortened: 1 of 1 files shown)


Moult:
> Also I'm wondering if you think it's doable to write unit tests for this, since I remember you saying testing UI is hard.

Yes, unit tests can be written here since it is not actually UI code, it's just explicit state changing. If interested, see how unit tests are written for the rest of the core modules.

> @Moult while testing this I noticed the behaviour is a bit weird when adding a new task to the list, usually when adding an element in a list (at least in Blender) this element becomes active, here it does not. Do you reckon it would be interesting to change the behaviour to select the element that's just been added in the list ? I can work on it if needed.

Happy either way. Perhaps ask @myoualid who is a domain specialist in 4D and knows what the desirable behaviour should be.

Doesn't stop merging this code :) Go go go!

PR #1942, opened 2021-12-13
Bug Fix: Correct HeaderFile

molitoris:
The `HeaderFile` class of bcf.v3 has a incorrect attribute. Rename `file_name` to `filename` (without underline).

Otherwise `f.filename` in `writer_header()` of bcf.v3.bcxml.py raises an AttributeError.


Moult:
Thanks so much!

Issue #1941, opened 2021-12-13
Umlauts (ÄÜÖß) in dynamic enumerations


vulevukusej:
Using umlauts in dynamically generated enums results in the symbols not being represented correctly.  There appears to be a workaround according to [https://blender.stackexchange.com/questions/216230/is-there-a-workaround-for-the-known-bug-in-dynamic-enumproperty](https://blender.stackexchange.com/questions/216230/is-there-a-workaround-for-the-known-bug-in-dynamic-enumproperty)

Thanks for the link @Gorgious56.  When I have time I'll create a pull request to address this.

Moult:
So this is now fixed and can be closed?

vulevukusej:
Yup fixed

PR #1945, opened 2021-12-16
fix umlauts not showing up correctly in dynamic enums #1941

vulevukusej:
I've started working on this, but I've hit a small roadblock. The umlauts now show correctly, but the dropdown is actually showing the options from the previous enumeration. Once selected, the enum_value is correct
umlaut bug
?

Edit - no matter what dropdown is selected, the same options are always shown..

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/pset/ui.py
+++ b/src/blenderbim/blenderbim/bim/module/pset/ui.py
@@ -19,6 +19,7 @@
 from bpy.types import Panel
 from blenderbim.bim.ifc import IfcStore
 from blenderbim.bim.helper import draw_attribute
+from blenderbim.bim.prop import getAttributeEnumValues
 from blenderbim.bim.module.pset.data import (
     ObjectPsetsData,
     ObjectQtosData,
@@ -95,6 +96,7 @@ def draw_psetqto_ui(context, pset_id, pset, props, layout, obj_type):
 
 def draw_psetqto_editable_ui(box, props, prop):
     row = box.row(align=True)
+    getAttributeEnumValues(prop, "")
     draw_attribute(prop, row, copy_operator="bim.copy_property_to_selection")
     if (
         "length" in prop.name.lower()
--- a/src/blenderbim/blenderbim/bim/module/pset/ui.py
+++ b/src/blenderbim/blenderbim/bim/module/pset/ui.py
@@ -19,7 +19,6 @@
 from bpy.types import Panel
 from blenderbim.bim.ifc import IfcStore
 from blenderbim.bim.helper import draw_attribute
-from blenderbim.bim.prop import getAttributeEnumValues
 from blenderbim.bim.module.pset.data import (
     ObjectPsetsData,
     ObjectQtosData,
@@ -96,7 +95,6 @@ def draw_psetqto_ui(context, pset_id, pset, props, layout, obj_type):
 
 def draw_psetqto_editable_ui(box, props, prop):
     row = box.row(align=True)
-    getAttributeEnumValues(prop, "")
     draw_attribute(prop, row, copy_operator="bim.copy_property_to_selection")
     if (
         "length" in prop.name.lower()
--- a/src/blenderbim/blenderbim/bim/prop.py
+++ b/src/blenderbim/blenderbim/bim/prop.py
@@ -24,6 +24,7 @@ import ifcopenshell
 import ifcopenshell.util.pset
 import blenderbim.bim.handler
 from blenderbim.bim.ifc import IfcStore
+from collections import defaultdict
 from bpy.types import PropertyGroup
 from bpy.props import (
     PointerProperty,
@@ -80,12 +81,37 @@ def update_is_visible(self, context):
                 pass
 
 
-def getAttributeEnumValues(self, context):
+def InternStr(s):
+    if not hasattr(InternStr, "StringCache"):  # Another way to define a function attribute
+        InternStr.StringCache = defaultdict(str)
+    InternStr.StringCache[s] = s
+    return InternStr.StringCache[s]
+
+InternStr.StringCache = {}
+
+
+def getAttributeEnumValues(prop, context):
     # Support weird buildingSMART dictionary mappings which behave like enums
-    data = json.loads(self.enum_items)
+    items = []
+    data = json.loads(prop.enum_items)
+    
     if isinstance(data, dict):
-        getAttributeEnumValues.enum_values = [(str(k), v, "") for k, v in data.items()]
-    getAttributeEnumValues.enum_values = [(e, e, "") for e in data]
+        for k, v in data.items():
+            for e in data:
+                items.append((
+                InternStr(k),
+                InternStr(v),
+                "",
+        ))
+    else:
+        for e in data:
+            items.append((
+            InternStr(e),
+            InternStr(e),
+            "",
+        ))
+            
+    return items
 
 
 def update_schema_dir(self, context):
@@ -146,10 +172,7 @@ class Attribute(PropertyGroup):
     is_null: BoolProperty(name="Is Null")
     is_optional: BoolProperty(name="Is Optional")
     enum_items: StringProperty(name="Value")
-    enum_value: EnumProperty(
-        items=lambda self, context: getAttributeEnumValues.enum_values, 
-        name="Value", 
-        update=updateAttributeValue)
+    enum_value: EnumProperty(items=getAttributeEnumValues, name="Value", update=updateAttributeValue)
 
     def get_value(self):
         if self.is_null:
--- a/src/blenderbim/blenderbim/bim/prop.py
+++ b/src/blenderbim/blenderbim/bim/prop.py
@@ -84,8 +84,8 @@ def getAttributeEnumValues(self, context):
     # Support weird buildingSMART dictionary mappings which behave like enums
     data = json.loads(self.enum_items)
     if isinstance(data, dict):
-        return [(str(k), v, "") for k, v in data.items()]
-    return [(e, e, "") for e in data]
+        getAttributeEnumValues.enum_values = [(str(k), v, "") for k, v in data.items()]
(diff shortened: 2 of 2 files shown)


vulevukusej:
stepping through the code as it evaluates doesn't reveal that there's a discrepancy, so I'm stumped.  Any ideas?

umlaut bug 2

vulevukusej:
seems i'm getting another error in the background as well
```
location: <unknown location>:-1
  File "C:\Users\vpaji\AppData\Roaming\Blender Foundation\Blender\3.0\scripts\addons\blenderbim\bim\module\pset\ui.py", line 150, in draw
    draw_psetqto_ui(context, pset["id"], pset, props, self.layout, "Object")
  File "C:\Users\vpaji\AppData\Roaming\Blender Foundation\Blender\3.0\scripts\addons\blenderbim\bim\module\pset\ui.py", line 79, in draw_psetqto_ui
    draw_psetqto_editable_ui(box, props, prop)
  File "C:\Users\vpaji\AppData\Roaming\Blender Foundation\Blender\3.0\scripts\addons\blenderbim\bim\module\pset\ui.py", line 100, in draw_psetqto_editable_ui
    getAttributeEnumValues(prop, "")
  File "C:\Users\vpaji\AppData\Roaming\Blender Foundation\Blender\3.0\scripts\addons\blenderbim\bim\prop.py", line 85, in getAttributeEnumValues
    data = json.loads(self.enum_items)
  File "c:\Users\vpaji\Downloads\blender-3.0.0-windows-x64\3.0\python\lib\json\__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "c:\Users\vpaji\Downloads\blender-3.0.0-windows-x64\3.0\python\lib\json\decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "c:\Users\vpaji\Downloads\blender-3.0.0-windows-x64\3.0\python\lib\json\decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Gorgious56:
I'm starting to think we should be dynamically creating `EnumProperty`s on attributes instead of shoehorning this feature using the `json` module and string properties. This would natively support the enum items and the currently selected one and could be passed seamlessly to operators. I'm sure this would prevent some of these headaches ^^

Btw I've never explored VScode debugging tools too much, would you mind explaining your workflow to get these values ?

Gorgious56:
> seems i'm getting another error in the background as well
> 
> ```
> location: <unknown location>:-1
> Python: Traceback (most recent call last):
>   File "C:\Users\vpaji\AppData\Roaming\Blender Foundation\Blender\3.0\scripts\addons\blenderbim\bim\module\pset\ui.py", line 150, in draw
>     draw_psetqto_ui(context, pset["id"], pset, props, self.layout, "Object")
>   File "C:\Users\vpaji\AppData\Roaming\Blender Foundation\Blender\3.0\scripts\addons\blenderbim\bim\module\pset\ui.py", line 79, in draw_psetqto_ui
>     draw_psetqto_editable_ui(box, props, prop)
>   File "C:\Users\vpaji\AppData\Roaming\Blender Foundation\Blender\3.0\scripts\addons\blenderbim\bim\module\pset\ui.py", line 100, in draw_psetqto_editable_ui
>     getAttributeEnumValues(prop, "")
>   File "C:\Users\vpaji\AppData\Roaming\Blender Foundation\Blender\3.0\scripts\addons\blenderbim\bim\prop.py", line 85, in getAttributeEnumValues
>     data = json.loads(self.enum_items)
>   File "c:\Users\vpaji\Downloads\blender-3.0.0-windows-x64\3.0\python\lib\json\__init__.py", line 346, in loads
>     return _default_decoder.decode(s)
>   File "c:\Users\vpaji\Downloads\blender-3.0.0-windows-x64\3.0\python\lib\json\decoder.py", line 337, in decode
>     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
>   File "c:\Users\vpaji\Downloads\blender-3.0.0-windows-x64\3.0\python\lib\json\decoder.py", line 355, in raw_decode
>     raise JSONDecodeError("Expecting value", s, err.value) from None
> json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
> ```

Any chance to debug what's the content of `self.enum_items` when this error is thrown ?

vulevukusej:
> Btw I've never explored VScode debugging tools too much, would you mind explaining your workflow to get these values ?

I set a breakpoint just before the enum is drawn:
image

vulevukusej:
@Gorgious56 - you were onto something, when the error is through self.enum_items is empty..  Let me check my ifc file again..
image

vulevukusej:
I played around a little with the getAttributeEnumValues function to avoid the jsondecodeerror:
```
def getAttributeEnumValues(self, context):
    # Support weird buildingSMART dictionary mappings which behave like enums
    if self.enum_items:
        data = json.loads(self.enum_items)
    
        if isinstance(data, dict):
            getAttributeEnumValues.enum_values = [(str(k), v, "") for k, v in data.items()]
        getAttributeEnumValues.enum_values = [(e, e, "") for e in data]
        #     return [(str(k), v, "") for k, v in data.items()]
        # return [(e, e, "") for e in data]
    else:
        getAttributeEnumValues.enum_values = []
```

Now the dropdowns look good, at least initially and until you click on them and see that they're all empty.  Debugging at the point at which the enum is drawn shows that everything _seems_ to be correct.
all empty
debug

vulevukusej:
@Moult @Gorgious56 - I'm at my wits end, I can't seem to find a solution for this.

Gorgious56:
I don't know if it's related but I think `is_null` should not be `True` on these attributes. Or maybe to think of it `is_optional` should not be `True` ?

Is there any error printed in the console when you try to expand an enum prop and it shows no item ?

Could you commit your latest changes so I can investigate a bit ?

Gorgious56:
#1941 for reference

vulevukusej:
> I don't know if it's related but I think `is_null` should not be `True` on these attributes. Or maybe to think of it `is_optional` should not be `True` ?
> 
> Is there any error printed in the console when you try to expand an enum prop and it shows no item ?
> 
> Could you commit your latest changes so I can investigate a bit ?

I haven't made any new commits because everything I tried hasn't worked.  I don't understand what I'm doing differently to [](https://blender.stackexchange.com/questions/216230/is-there-a-workaround-for-the-known-bug-in-dynamic-enumproperty) .  Also, the debug looks fine just before drawing..

Gorgious56:
> I haven't made any new commits because everything I tried hasn't worked. I don't understand what I'm doing differently to https://blender.stackexchange.com/questions/216230/is-there-a-workaround-for-the-known-bug-in-dynamic-enumproperty . Also, the debug looks fine just before drawing..

Well it is different from the problem in the question because here you can have an indefinite number of `EnumProperty` to draw instead of only one, but on the other you only have one and only one `getAttributeEnumValues.enum_values`. So if you update it before drawing each attribute, it will be overwritten by the next one to be drawn.

Maybe it's worth exploring the [second answer there](https://blender.stackexchange.com/a/245145/86891) with a dictionary cache. However it's a recipe for memoy leak, we'd have to make sure it's cleared every time we update the property sets panel.

vulevukusej:
@Gorgious56 yeah ok, i'll try that next

vulevukusej:
@Gorgious56 - so that second answer worked.. 😲.   I'll commit the changes after I test it some more but, wow, the solution was sitting there the entire time..  Now, you get +10 bonus points if you can explain to me **_why_** this worked, and the other method didn't.

Gorgious56:
I tried to in the first paragraph of my last comment but I have to say it's a bit complicated to grasp ^^. And I have to admit I'm not even sure of my explanation. So lets say you have N different `EnumProperty`. They all have their own `items` method which is used by the API to populate the enumeration items. So we have N `items` methods. But they all reference (point to) only 1 method, which is `getAttributeEnumValues`. 

Then Blender reads the `draw` code : For each one of the N `EnumProperty`, there is a call to an individual `items` method, which fetches the possible enum items for that particular `EnumProperty` and links it `getAttributeEnumValues.enum_values` to it. 

But I think the actual evaluation happens not when the code is being read, but when you try to access the layout using the interface (ie when you move your mose over the panel for instance). At this point `items` has been evaluated N times, but every time it has been overwritten by the next `EnumProperty` to be drawn. That still does not explain why it is empty, I would assume it would take the values of the last `EnumProperty` to be evaluated.

When using a dictionary, you make sure there is no possibility to overwrite enum items since they are uniquely identified and accessed with a key. However as stated in the answer the elements are not automatically garbage collected and they will stay there "forever" (until you quit Blender or reload a save) unless explicitely deleted.

But the method is similar, and instead of using a `global` variable you could for instance : (not tested but I assume it would work)

def InternStr(s):
    if s not in InternStr.StringCache:
        InternStr.StringCache[s] = s
    return InternStr.StringCache[s]

InternStr.StringCache = {}

Also is looks like a good candidate for defaultdict :

from collections import defaultdict

```
def InternStr(s):
    if not hasattr(InternStr, "StringCache"):  # Another way to define a function attribute
        InternStr.StringCache = defaultdict(str)
    InternStr.StringCache[s] = s
    return InternStr.StringCache[s]
```

vulevukusej:
Ok, I think I get the gist of it.  Let me try to get rid of the global variable, test it some more and then I'll commit the changes :).  In the meantime, here is your reward: 🥇🥇🥇 :)

vulevukusej:
oh oh, so I was testing again and I realized I missed this:  as you can see some of the enums are glitching when they have umlauts, whereas most are ok.  This kind of "sometimes works, sometimes doesn't" behaviour gives me headaches.
new prob
.

Moult:
Sooo just coming back from a 3 week AFK period and reading this thread makes me smile :D 

A thousand points to everyone!

Issue #1943, opened 2021-12-14
Integers in enumerations cause an error


vulevukusej:
So I'm not sure if this is an error in the way I'm defining the .ifc file, or if its an issue with BB.  When I attempt to load this template file and then add the pset to an object, I get the following error message:
`File "C:\Users\vpaji\AppData\Roaming\Blender Foundation\Blender\3.0\scripts\addons\blenderbim\bim\prop.py", line 83, in getAttributeEnumValues
TypeError: EnumProperty(...): expected a tuple containing (identifier, name, description) and optionally an icon name and unique number`

Here is the .ifc file I was importing:
```
ISO-10303-21;
HEADER;
FILE_DESCRIPTION((),'2;1');
FILE_NAME('Pset_Custom.ifc','2020-01-01T00:00:00',(),(),'Sample','Sample',$);
FILE_SCHEMA(('IFC4'));
ENDSEC;

DATA;
#1= IFCPROPERTYSETTEMPLATE('05N2Dhepb9ouTCSmWBh22D',$,'Custom_Pset','An example custom pset.',.PSET_TYPEDRIVENOVERRIDE.,'IfcObject',(#2,#3,#4));
#2= IFCSIMPLEPROPERTYTEMPLATE('37EpTDq4v96gTZh196BBXQ',$,'NumericProperty','An example numeric property',.P_SINGLEVALUE.,'IfcCountMeasure','',$,$,$,$,.READWRITE.);
#3= IFCSIMPLEPROPERTYTEMPLATE('22H7CzDvz1XfVHKkGWoYUK',$,'StringProperty','An example string property',.P_SINGLEVALUE.,'IfcLabel','',$,$,$,$,.READWRITE.);
#4= IFCSIMPLEPROPERTYTEMPLATE('0p00ORU2nCnPA8UyQg2ZY$',$,'porr_Ganzzahl_E05','Impraegnierung',.P_ENUMERATEDVALUE.,'IfcInteger',$,#5,$,$,$,.READWRITE.);
#5= IFCPROPERTYENUMERATION('Impraegnierung',(IfcInteger(1), IfcInteger(2), IfcInteger(3), IfcInteger(4), IfcInteger(5), IfcInteger(6), IfcInteger(0)),$);
ENDSEC;

END-ISO-10303-21;
```

vulevukusej:
@Gorgious56 - Is this something you've also encountered?  Perhaps it can also be fixed with the same method as in #1941

[...]

PR #1946, opened 2021-12-16
Fix integers in enumerations #1943

vulevukusej:


Gorgious56:
#1943 just for reference

src/blenderbim/blenderbim/bim/prop.py
```diff
@@ -147,11 +148,17 @@ class Attribute(PropertyGroup):
...
-
+    enum_data_type: StringProperty(name="Enum Data Type")
+    
     def get_value(self):
         if self.is_null:
             return None
-        return getattr(self, str(self.get_value_name()), None)
+        if self.data_type == "enum":
+            type_map = blenderbim.bim.schema.ifc.type_map
+            type_fn = {'integer':int, 'string':str, 'float':float, 'bool': bool}[type_map[self.enum_data_type]]
```
Gorgious56, line 158: This is ugly but not unprecedented https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/patch/operator.py#L108-L113

Gorgious56:
It makes sense to me. 

Have you been able to test it and see if there is no loss of information when exporting / importing psets with non-str enum values ?

src/blenderbim/blenderbim/bim/prop.py
Gorgious56: Well I wrote it and I'm a bit ashamed ^^ so maybe we can add this lookup dictionary as a constant somewhere in the code ?

vulevukusej:
> 

importing and exporting works fine - i've been using this commit productively for the past few days

src/blenderbim/blenderbim/bim/prop.py
vulevukusej: haha - yeah it ain't pretty

PR #1955, opened 2021-12-31
Updated README.md for OCCT/IfcOpenShell7

ssg3d:
After #1930 the 0.7 branch isn't compatible anymore with OCCT version < 7.2. That includes the liboce that's found in debian (derivatives), which is OCCT 6.9 or even less. Updated the Linux section of README.md to explain the dependency on OCCT and include instructions to build on Unix with OCCT (from the OS repository as well). Also included changes that became necessary when we moved to v0.7.0 from v0.6.0 (like recursive clone and cmake additions).

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,9 @@ Dependencies
 Building IfcOpenShell
 ---------------------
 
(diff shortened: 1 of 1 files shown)


aothms:
Perfect, many thanks!

PR #1963, opened 2022-01-04
Update README.md

garylzimmer:
Updated based on my personal findings. Can provide documentation if necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Prerequisites
 -------------
 * Git
 * CMake (2.6 or newer)
-* Windows: [Visual Studio] 2008 or newer with C++ toolset (or [Visual C++ Build Tools]) or [MSYS2] + MinGW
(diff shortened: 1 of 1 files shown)


Moult:
Instead of linking to the latest build which will always change, it might be better to describe the process to find it (i.e. look for last commit with messages by IfcOpenBot) :) Otherwise, very good idea because this question is asked a lot!

garylzimmer:
@Moult I think your comment belongs here:
https://github.com/IfcOpenShell/IfcOpenShell/pull/1964
But to respond to your suggestion about the process to find it:
image

The slow pace of full update binaries releases I think wouldn't break the link too often and the directions would have to be pretty good because I'm still learning how to navigate github, and I'm not even sure if the link I posted is the latest!

Moult:
Hehehe yes, @aothms is the best to make the call on this. Lately builds have stagnated a bit, but we used to have a new build once or twice a month.

aothms:
Yeah I think the bot also deserved her Christmas holidays. When I have a little more time I'll enable her to run autonomously again.

Thanks for the text @garylzimmer  :)

PR #1964, opened 2022-01-04
Update README.md

garylzimmer:
Added quick link to the prebuilt binaries

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,13 @@ For more information, see
 * [http://ifcopenshell.org](http://ifcopenshell.org)  
 * [http://academy.ifcopenshell.org](http://academy.ifcopenshell.org)
 
+
+
+## Latest Pre-built Binaries
+[See this post](https://github.com/IfcOpenBot/IfcOpenShell/commit/b4ce5be86d293246dd84f59891e2d5d7f2240284#comments)
+
+# To Build Yourself
+
 [![Build Status](https://travis-ci.org/IfcOpenShell/IfcOpenShell.svg?branch=v0.6.0)](https://travis-ci.org/IfcOpenShell/IfcOpenShell)
 
 Prerequisites


aothms:
From @Moult 

> Instead of linking to the latest build which will always change, it might be better to describe the process to find it (i.e. look for last commit with messages by IfcOpenBot) :) Otherwise, very good idea because this question is asked a lot!

aothms:
And additionally, this is a build from the v0.6 branch if I'm not mistaken, that we'd be merging into the 0.7 docs. That will cause additional confusion.

I agree sth needs to be done awaiting the new website but I'm not sure what. The build bot code is fully available here https://github.com/IfcOpenShell/IfcOpenBot/blob/master/upload.py it uploads to S3 and posts the github comment (L93).

I'm not sure how we can get the README to display dynamic content without continuously triggering commits. The only thing I can think of is yet another repo with a SVG or sth that is edited by the build script. And that SVG is inlined into the main github repo here. It'd be a bit like all these badges you see in the README. I'm not sure if we can get it to be interactive with links (last time I checked the SVG needs to be inlined into the HTML to have interactive links, not references as an image) but perhaps there are ways around that with foreignObject in SVG again.

A static link I think will do more harm than good in the long run. Are you ok with taking Moult's approach or are you in any way interested in investigating the SVG approach? We actually need something like that for the new website anyway (just HTML or JSON).. So that would be super helpful! https://community.osarch.org/discussion/comment/8302/#Comment_8302

garylzimmer:
@aothms I'm not familiar with sth or SVG, but here's my suggestion:
image
Not sure how much work would be involved in automating the actual reposting of that shortened link as a commit to the README, but if you are considering redoing the website, maybe it would be easier to automate the updating of the link on the website itself rather than the github.

fbpyr:
how about linking to the dynamic latest release provided by github instead?: 
https://github.com/IfcOpenShell/IfcOpenShell/releases/latest

Moult:
Closing as no longer relevant with the new install instructions e.g. https://blenderbim.org/docs-python/ifcopenshell-python/installation.html though the fundamental issue with grabbing the latest bot build is still a manual responsibility (but I do it anyway for the BlenderBIM Add-on so I'm happy to continue doing it until we figure out a nice automated way).

Issue #1956, opened 2022-01-03
BlenderBIM: Suggestion: Support for IFC2X3 for 4D/5D capabilities


garylzimmer:
After feedback from my issue here
https://github.com/IfcOpenShell/IfcOpenShell/issues/1954

I decided to break this into it's own issue to be tracked separately.

Since proper support might not be possible due to a lack of 4D/5D parameters in the IFC2X3 spec, a first goal could be to have an option in the importing function to auto-run the upgrade recipe to migrate to IFC4 format.

garylzimmer:
Also, if there was a way to see what version IFC spec the file is using before committing to the importing, that would also be helpful

Moult:
Sorry, would you mind adding your suggestion for IFC schema data prior to importing in another issue? Just trying to stem the flood of reports :)

aothms:
> Also, if there was a way to see what version IFC spec the file is using before committing to the importing, that would also be helpful

~~~
>>> import ifcopenshell
>>> f = ifcopenshell.file()
>>> f.schema
'IFC4'
~~~

garylzimmer:
@aothms Thanks for that! I'm still exploring the ifcOpenShell scripts, they seem very useful! Does blender provide a way to incorporate that kind of functionality into the import dialog box?

Moult:
Prior to building the interface for this, you can use @aothms 's snippet, but you might want to do `f = ifcopenshell.open('/path/to/file.ifc')` instead.

The thing about putting this in the UI is that it may result in reading the file, which we don't want since a 300mb file can easily be quite a while to read. Perhaps to implement this efficiently we need to first grab the first N bytes to cover the header, then just parse the SPF header and nothing else... but that will break for other serialisations. Not sure.

garylzimmer:
@Moult Sorry about blowing up your inbox, I just got so excited by the possibilities of an Open Source AEC project that I could contribute to! I'm not sure what else you are looking for as far as the schema data. From your own site:

> The classes available for you to use are different for each IFC version. In general, IFC4 has many more classes, and the upcoming IFC5 will include even more for infrastructure and landscape projects. It is not advisable to use IFC2x3, for the simple reason that it lacks many useful IFC classes.
This tool below lets you type what type of object you have, and it will try and find the right IFC class for you. It will recommend IFC4 classes. We do not encourage the use of IFC2x3 files.

[...]

PR #1978, opened 2022-01-10
Display IFC schema in file selector right hand side panel

Gorgious56:
This commit is a proof of concept for displaying selected ifc information right in the file selector popup window.

Following https://github.com/IfcOpenShell/IfcOpenShell/issues/1956

Right now it works with `bpy.ops.bim.select_ifc_file` but could easily be extended to all the other file selector operators :

image

Pretend these are ifc files, I've hidden the names for privacy :

BSE_80

### DIFF
--- a/src/blenderbim/blenderbim/bim/operator.py
+++ b/src/blenderbim/blenderbim/bim/operator.py
@@ -22,6 +22,7 @@ import json
 import webbrowser
 import ifcopenshell
 import blenderbim.bim.handler
+from pathlib import Path
 from . import schema
 from blenderbim.bim.ifc import IfcStore
 from mathutils import Vector, Matrix, Euler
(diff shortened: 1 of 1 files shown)


Gorgious56:
And ping @garylzimmer since the PR was motivated by https://github.com/IfcOpenShell/IfcOpenShell/issues/1956#issuecomment-1004355726

Moult:
This looks awesome! We just need to be conscious that this trick only works for IFC-SPF. Other IFC formats will not be supported yet.

Gorgious56:
I'm not that well versed on other ifc files header formatting. Do you know where I could find ressources to cover more file types ?

Moult:
I actually don't know myself :) By the way would you mind implementing this for the import ifc / load project dialog?

Moult:
BTW I tested this out on larger files and no issues at all. Everything is instant. By the way, it failed on a few files, who have a space like this:

```
FILE_SCHEMA (('IFC2X3'));
```

Not sure how lenient you want to make the code since this is just string parsing but if you feel motivated here's the STEP header specification https://www.steptools.com/stds/step/IS_final_p21e3.html#clause-8

Gorgious56:
Yeah I considered using regex for that but then I remembered why everyone hates it ^^

I'll look into it and into extending it to other operators, thanks for the feedback !

PR #1979, opened 2022-01-10
UI Simplification with Panel Groups

vulevukusej:
Here's a concept demo of the ui simplification using panel groups.  The grouping is shown at the bottom after the gif:
ui rethink

image


Gorgious56:
Looks nice !

I don't work enough with IFC to really know how it might affect different workflows (especially the amount of increased *clicking* versus the amount of decreased *scrolling* :) )

I would drop the "Useful" in "Useful Tools", because by definition a tool is useful ! Maybe we can use "Utilities" or "Convenience" ? Idk, it might be worth checking in other softs how these are labeled.

And maybe we could brainstorm for a dedicated sub-panel name instead of "Misc" which is not very evocative for the layman.

Also the change would really shine I think if we could drop the "IFC" prefix from the sub-panels, reducing visual noise.

Have you tested how it fares with the work you did in the addon preferences to dynamically disable panels ?

Note if you want to enforce an order in subpanels, add `bl_order = integer_number` in the class attributes where lower number means higher in hierarchy (or the converse, can't remembler). Otherwise panels are ordered in the order of registration which might get messed up by the feature I talked about just earlier.

Cheers !

vulevukusej:
Thanks for the feedback.  I plan on using this layout for a couple of days to see how it compares to the long list.  For the last couple of hours though, I've felt it easier to get to the tool I want, but the experience will likely vary for other users.  

I fully agree with you regarding the group names and IFC Prefixes, this first commit was meant to be an initial POC, ideally I'd like to get some feedback from more users on the naming before it's finalized.

I haven't tested the preferences yet, but my feeling is that if we nail this right, we might not even need them anymore?

I`m aware of `bl_order` but I didn't really see a need to order the panels just yet, I'm open to suggestions on what this order should be though.

Moult:
It looks like a pretty sensible grouping. The best way to know whether it's the right move or not is to just commit it and we can test it until someone complains :)

I'm sure we all have preferences on panel naming but that's an easy thing to change so I'm not going to make any comments at all :)

Let me know when you're got some code ready to merge and let's go go go!

Moult:
BTW am I missing something? Where is `ui.BIM_PT_project_and_template_setup` et all defined?

vulevukusej:
Good point, I must've somehow left that file out!  I've committed the changes and even changed useful tools to utilities to keep @Gorgious56 happy ;).

Gorgious56:
> I`m aware of `bl_order` but I didn't really see a need to order the panels just yet, I'm open to suggestions on what this order should be though.

I don't have a preference either, but I believe the problem is the order will change when / if we dynamically unregister / register the panels which I think is the method used in the preferences panel filtering feature. 

Anyways, I don't think it's a big deal and as @Moult said we should worry about it when someone actually complains about it :)

Otherwise looks good to me !

Cheers

Gorgious56 (approved):

PR #1980, opened 2022-01-11
Show ifc specs in file selector

Gorgious56:
Following #1978

BSE_82

This is implemented on `bim.select_ifc_file`, `bim.select_library_file` and `bim.load_project`.

It will only work with a very fragile string parsing system so files coming from other vendors or other ifc formatting will not show anything. It should not throw any error either way.


Moult:
Looks sexy! Awesome work and thanks!

PR #1986, opened 2022-01-13
adjusted UI based on feedback received

vulevukusej:


vulevukusej:
- Changed "IFC Collaboration and Data Exchange" to "IFC Collaboration
- Changed "IFC Object Attributes, Properties and Relationships" to "IFC Object Metadata"
- Added poll methods to the panel groups to declutter the UI when no IFC project loaded.  Some panels were left behind since they can be used without a project (Quality Control for ex.)
- Removed "Default_Closed" from some of the panels
- Added new IFC Services and nested IFC Systems under this group.  IFC Services can then be the parent panel for future MEP related panels
- Rearranged some panels that were nested under the wrong group.

Issue #1982, opened 2022-01-12
Dynamic tooltips for attribute and property descriptions


Moult:
Here's a fun challenge if someone is up to it. It would be nice to have tooltips which annotate various things in IFC. There's two challenges to this: 1) getting the descriptions to use for tooltips, and 2) doing dynamic tooltips in Blender.

For the second, @Gorgious56 has done some sleuthing and this reregistration seems to be the way to go: https://community.osarch.org/discussion/comment/9204/#Comment_9204

For getting the tooltips, I wrote this little script a while back which scrapes the downloaded HTML docs from technical.buildingsmart.org. It's annoying we have to do this, but buildingSMART doesn't give these descriptions in any other parsable way.

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/extract.py

I also produced https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/schema/entity_descriptions.json and https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/schema/enum_descriptions.json somehow from somewhere.

Property descriptions can be extracted from pset template definitions.

vulevukusej:
This does sound like fun - If nobody makes a start on this before Feb I'll gladly look into it!

aothms:
> but buildingSMART doesn't give these descriptions in any other parsable way.

(a) Best would be to use the translation files https://github.com/buildingSMART/IFC4.3.x-output/blob/master/IfcActorResource.pot#L17-L18 That's what they are intended for after all :) and you'll even get i18n tooltips once the translation team gets going.

(b) Alternatively there is markdown https://github.com/buildingSMART/IFC4.3.x-development/blob/master/docs/schemas/shared/IfcSharedBldgElements/Entities/IfcWall.md

You can use this code to parse the attribute definitions according to the heading convention https://github.com/buildingSMART/IFC4.3.x-development/blob/master/code/md.py

It's a different version of the schema, but I think it's better to use this than scraping the HTML.

(c) If you insist on having the correct schema I would look at the IFC doc repo which has tags/branches for all current schemas https://github.com/buildingSMART/IFC/blob/Ifc4.0.2.2/Sections/Shared%20element%20data%20schemas/Schemas/IfcSharedBldgElements/Entities/IfcWall/Documentation.md

Moult:
[...]

PR #1990, opened 2022-01-17
Dynamic Attribute Info

Gorgious56:
Following #1982

This PR aims to provide user with a "newbie mode" and pepper info buttons in the interface (for now, only on dynamically generated attributes) in order to display more wordy information on the related item or element. The aim is to prevent having to constantly go online and read the docs to find out about such or such item.

Example in action :

Added a toggle in the addon preferences (ON by default for testing but maybe we should set it to OFF by default ?)

image

This add an info popup button next to attributes :

BSE_84

I still need to find a way to fetch ifc class documentation url + usecase and other relevant informations (help appreciated 😁 )


Moult:
Oh wow this is super cool. @vulevukusej did you want to try write a tooltip extractor? (or someone else like to volunteer?) If not, I'd be happy to write the extractor as it'd be awesome to see this in the next stable release!

Moult:
In the meantime, I can think of three locations for this to be a thing:

 - In the IFC Class panel's IFC Class dropdown, teaching people about different IFC classes (e.g. IfcWall is a such and such).
 - Anywhere attributes are drawn
 - Anywhere properties are drawn (Object Property Sets, Quantity Sets, material property sets, etc, see `draw_psetqto_ui` in `module/pset/ui.py`)

Classes and attributes will be extracted into a doc that we'll ship as a JSON dataset. However, the properties are different, as it can be dynamically fetched from property templates (which may be customised by users). So just need to keep in mind that they are slightly different.

vulevukusej:
@Moult - Unfortunately, I don't think I'll have any time for BB in the next 2 weeks or so..

@Gorgious56 -  This looks so damn good :).   Btw, after reading the suggestion [here ](https://github.com/IfcOpenShell/IfcOpenShell/issues/1982#issuecomment-1010795249) I started exploring the gettext module in python and it seemed like that was the way to go.  I ran out of time before I could dig deeper, but if you haven't already gotten too far ahead by Feb, I'd be happy to help in any way I can.

Gorgious56:
A bit of progress : (Added description field)

BES_59

Still most of the work is to be done there https://github.com/Gorgious56/IfcOpenShell/blob/5d6331626b7e63bfdc74a0636fd4b69afdc6fd8b/src/blenderbim/blenderbim/bim/helper.py#L103-L117 but the UI is shaping up !

Gorgious56:
Another batch of info display tests, for non-ifc classes. A little hard to find exactly what to say in these info popups but the system is here.

Still figuring out the exact way this should fall into place. 
@Moult do you think it's worth it to create a new submodule for documentation inside the add-on, so as not to pollute too much the other areas ? 
I think ideally it would be a module totally abstracted from the other parts of the code that would dynamically add info popups to the layout where necessary.

BSE_87

vulevukusej:
@Gorgious56 How did you get along with this, anything I can try and help with?

Gorgious56:
@vulevukusej erm unfortunately I did not advance in this area much.

I remember the thing I had trouble with was querying relevant information about the property type in the docs. Maybe we can get the info from somewhere in the ifcopenshell database without having to query from the internet ?

Moult:
@Gorgious56 indeed yes another module for doc-enriching helpers / code would be a great idea!

I believe for IFC4 the descriptions may be programmatically extracted (@aothms can confirm this) from this repository:

https://github.com/buildingSMART/IFC/tree/Ifc4.0.2.1

For example, in [this subfolder](https://github.com/buildingSMART/IFC/tree/Ifc4.0.2.1/Sections) all IFC entities are stored somewhere in here (in a folder called `Entities`). It's a little difficult to find if you're not used to their categorisation system, but here's an example for [IfcProject](https://github.com/buildingSMART/IFC/tree/Ifc4.0.2.1/Sections/Core%20data%20schemas/Schemas/IfcKernel/Entities/IfcProject). In this case, say the first sentence of the `Documentation.md` can be scraped. If the entity has attributes, then documentation for the attributes are stored in `DocEntity.xml`. In the case of IfcProject, it doesn't have any, but it's inherited parent class of IfcContext does, and so [here is an example](https://github.com/buildingSMART/IFC/blob/Ifc4.0.2.1/Sections/Core%20data%20schemas/Schemas/IfcKernel/Entities/IfcContext/DocEntity.xml).

Properties are stored alphabetically in [this dir](https://github.com/buildingSMART/IFC/tree/Ifc4.0.2.1/Properties) and quantities are [here](https://github.com/buildingSMART/IFC/tree/Ifc4.0.2.1/Quantities). Similarly they have a `Documentation.md` which can be scraped like [this example](https://github.com/buildingSMART/IFC/blob/Ifc4.0.2.1/Quantities/f/FinishCeilingHeight_1vYXYSNGLCGBg4RxwFeBW6/Documentation.md).

A script that loops through dirs, scrapes docs from the DocEntity.xml and Documentation.md file, should be sufficient I believe. Perhaps the processed results can be stored in a JSON file we can reference in the future.

Does that help? A similar exercise would need to be done for other schema versions but is that what you need to continue? If it is complex I'm also happy to write the scraping code.

aothms:
I have also been tasked to create 2x3 and 4 versions of the UML + markdown environment we have for 4.3. Maybe it's wise to wait for that and have a unified approach to deal with IFC documentation content.

Moult:
In that case, maybe the code can be written based on IFC4.3's docs (which in most cases is a superset of the former versions)?

aothms:
In that case. There is a little helper class here https://github.com/buildingSMART/IFC4.3.x-development/blob/master/code/md.py#L75 that can get you the md file as a dictionary of attribute -> definition.

The license of the code is a bit unclear still, but my plan is to separate out code from content (also for the purpose of providing 2x3 and 4 versions). Content will be CC BY NC ND (or other) and code can be LGPL if you ask me.

Moult:
The code is MIT licensed so we can probably reuse that code :) https://github.com/buildingSMART/IFC4.3.x-development/issues/241

Moult:
@Gorgious56 bump :) there is now a dummy database forming in https://github.com/IfcOpenShell/IfcOpenShell/tree/v0.7.0/src/ifcopenshell-python/ifcopenshell/util/schema worked on by @Andrej730

Gorgious56:
Woot Nice !  I'll look into it.

Moult:
@Gorgious56 ifcopenshell.util.doc functions are now available courtesy of @Andrej730 :) https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/util/doc.py#L83-L97

Gorgious56:
Alright I think for now this is a good compromise. I removed the props tied to attributes and elected for a dynamic fetching of data and using the enum description which kind of neatly wraps the sentences around and doesn't need to jump through hoops to display data to the user. 

For now only a few fields are supported (ifc class, ifc product, ifc predefined type) but I can see this being extended further with a bit of tweaking. I tried to tidy a bit but I'm not sure everything is where it's at, sorry if I made a mess. Don't hesitate to point me towards resources to make it right if necessary :)

Would you mind sharing your thoughts ? @Moult @theoryshaw @cvillagrasa @vulevukusej and anyone interested ?

I think this PR can be merged as is and see what users think of it. I can add more functionality afterwards.

Animation

vulevukusej:
@Gorgious56 This is really awesome :)!  I can't wait to try it out, I'll let you know if i find any bugs.

theoryshaw:
error when trying to enable BB...
at this commit.

```

  File "C:\Program Files\Blender Foundation\Blender 3.2\3.2\scripts\modules\addon_utils.py", line 335, in enable
    mod = __import__(module_name)
  File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\3.2\scripts\addons\blenderbim\__init__.py", line 39, in <module>
    import blenderbim.bim
  File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\3.2\scripts\addons\blenderbim\bim\__init__.py", line 21, in <module>
    from . import handler, ui, prop, operator, helper
  File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\3.2\scripts\addons\blenderbim\bim\handler.py", line 23, in <module>
    import blenderbim.tool as tool
  File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\3.2\scripts\addons\blenderbim\tool\__init__.py", line 21, in <module>
    from blenderbim.tool.boundary import Boundary
  File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\3.2\scripts\addons\blenderbim\tool\boundary.py", line 23, in <module>
    from blenderbim.bim.module.geometry.helper import Helper
  File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\3.2\scripts\addons\blenderbim\bim\module\geometry\__init__.py", line 20, in <module>
    from . import ui, prop, operator
  File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\3.2\scripts\addons\blenderbim\bim\module\geometry\ui.py", line 22, in <module>
    from blenderbim.bim.helper import prop_with_search
  File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\3.2\scripts\addons\blenderbim\bim\helper.py", line 26, in <module>
    from ifcopenshell.util.doc import get_entity_doc, get_attribute_doc, get_property_set_doc, get_property_doc
  File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\3.2\scripts\addons\blenderbim\libs\site\packages\ifcopenshell\util\doc.py", line 26, in <module>
    from markdown import markdown
ModuleNotFoundError: No module named 'markdown'

```

Moult:
Yep in the middle of fixing that.

Issue #1997, opened 2022-01-21
Improvements to the section cutaway shader groups


Gorgious56:
Opening this so I can track my progress.

Related : https://github.com/IfcOpenShell/IfcOpenShell/issues/1995

Target :
- Section shader must accept an input object, which will be used to make the geometry over or under its local Z axis transparent
- Backfaces must be drawn in another color defined in `context.scene.BIMProperties.section_plane_colour`
- Section shaders must stack

Improvements :
- Get rid of unneeded nodes
- Setup node positions so it's more convenient to tweak the node group in the shader editor (currently they are all stacked at 0, 0)

![BSE_85](https://user-images.githubusercontent.com/25156105/150489959-48435647-fb9f-4bc3-b2ad-307758e9ed68.gif)

Gorgious56:
Okay this issue can be closed as the PR is ready for evaluation or directly merging if nobody objects. 

Reading through the initial code I finally understood what the person who wrote it wanted to do. Fortunately I found a solution that is a bit neater (IMO ^^).

Issue #1995, opened 2022-01-21
Section Cutaways don't work with Blender 3.0


amurrayc:
When I apply a section cutaway in 3.0 everything disappears in material preview or rendered mode in the viewport.  In 2.93.5 it works fine.

I'm running Linux (NixOS) and see the same effect on a desktop with a Vega 56 GPU and a laptop with Intel integrated graphics.

Moult:
Thanks for catching this bug. Would you mind testing to see if the fix does anything for Blender 2.93 as well?

I think I need a shader guru ( @Gorgious56 ) to fully understand the implications of this.

amurrayc:
Yes, that appears to fix it in 3.0 and 2.93.5 still works.  That was quick!

Gorgious56:
@Moult I have to say I don't fully understand the intent behind this input value node, is it meant to be an offset to the clipping plane ? I would try and get rid of it and set the compare node value to 0 instead : https://github.com/IfcOpenShell/IfcOpenShell/blob/595c72e08dfb9dded86e2b7a26831ebeb180ec12/src/blenderbim/blenderbim/bim/operator.py#L166    `compare.inputs[1].default_value = 0`

But you'll get the opposite clip plane direction as the one you have now, so you'd have to invert it :


    invert = group.nodes.new(type="ShaderNodeInvert")

and then 


    group.links.new(compare.outputs[0], invert.inputs[1])
    group.links.new(invert.outputs[0], group_output.inputs[""])

Moult:
@Gorgious56 the absolute truth is I stole the shader from some site online and maybe I understood how it worked then, but I certainly don't anymore. Have you tried looking at the setup graphically in Blender's node editor? Also note that these shaders are designed to stack, so something that could seem redundant could be stubbing in a potential input node for stacking perhaps? (Try adding multiple section planes).

Gorgious56:
Alright, I can definitely help put some order into this then. 
Could you confirm the desired behavior of this shader, as I understood from looking at it in the node editor :
Clip all the shading of the object if it's below the origin of a target object along its local Z axis, and apply a white shader on backfacing polygons ?

![BSE_85](https://user-images.githubusercontent.com/25156105/150489959-48435647-fb9f-4bc3-b2ad-307758e9ed68.gif)

Moult:
[...]

PR #2000, opened 2022-01-23
Improve section cutaway shader

Gorgious56:
Following https://github.com/IfcOpenShell/IfcOpenShell/issues/1997 and #1995

The Section Cutaway features are :

- Section shader must accept an input object, which will be used to make the geometry under its local Z axis transparent

- Backfaces must be drawn in another color defined in context.scene.BIMProperties.section_plane_colour

- Section shaders must stack

Summary of the PR :

- Optimize node network (removed ~8 nodes)

- Make node network "human readable" by tweaking the node positions
Before :
image
After :
image

- Dynamically update the backface color of the cutaway from the interface
BES_56

- Do not forcefully delete objects that are not a section cutway object from the "Remove Temporary Section" button

- Rearrange nodes neatly when deleting section objects
BES_57

- Add tooltips to both operators

- Rename operator classes to adhere to blender guidelines

Possible improvements :

A lot of the code relies on fetching nodes and objects by name, which will throw errors if they are modified. It is unfortunately one of the problems of the node system. I think it's fair to assume that if the user is bold enough to modify things in the shader editor, they should be able to un-break things afterwards :p

### DIFF
--- a/src/blenderbim/blenderbim/bim/__init__.py
+++ b/src/blenderbim/blenderbim/bim/__init__.py
@@ -81,7 +81,7 @@ classes = [
     operator.SelectSchemaDir,
     operator.SelectIfcFile,
     operator.OpenUpstream,
-    operator.AddSectionPlane,
+    operator.BIM_OT_add_section_plane,
     operator.RemoveSectionPlane,
     operator.ReloadIfcFile,
     operator.AddIfcFile,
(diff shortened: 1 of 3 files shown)


Moult:
Happy bug #2000 !

Amazing set of improvements. Let's merge it all!

Gorgious56:
@Moult Woot didn't notice that ! Champagne for everybody :) 🍾 🍾

PR #2003, opened 2022-01-25
Fix #2002. Fix bug where sometimes shape IDs wouldn't be pure integers.

### LINKED ISSUE #2002: When would shape.id have a comma in it?
I assumed that if I took shape.id (or shape.geometry.id, depending on how I call `create_shape`) and did `shape.id.split('-')[0]` I would always get an integer of the STEP ID of the representation (or representation item, profile, etc). It looks that way in the code: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom/IfcGeomFunctions.cpp#L1837

However, @CyrilWaechter brought up a situation where if I have an IfcCurveBoundedPlane (a representation item) and do create_shape on that, then the returned shape.id is `5,889` that is, it has a comma in it.

What portion of the C++ would cause this? Should it be modified so that it is consistent with L1837 in the link above?

Moult:
I grepped for lexical_cast and to_string in ifcgeom, ifcwrap and ifcparse directories and couldn't find any other users of it that would be worth replacing (for example, they were used in things like logs, where a thousands separator is fine).

I don't fully understand the incantation you've provided but I hope I copied your code into the right spot :)

I did test it though and it works :)

### DIFF
--- a/src/ifcwrap/IfcGeomWrapper.i
+++ b/src/ifcwrap/IfcGeomWrapper.i
@@ -328,6 +328,14 @@ struct ShapeRTTI : public boost::static_visitor<PyObject*>
 };
 
 %{
+	template <typename T>
+	std::string to_locale_invariant_string(const T& t) {
+		std::ostringstream oss;
+		oss.imbue(std::locale::classic());
+		oss << t;
+		return oss.str();
+	}
+
 	template <typename Schema>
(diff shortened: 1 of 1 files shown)


aothms:
> incantation 

It's just that there is this default locale in C++ inherited from the OS that to_string and lexical_cast use. "Imbue"-ing the c classic locale before the stream operation overrides this. The default c locale is without thousands sep and point for decimal sep.

aothms:
Thanks for the commit

PR #2015, opened 2022-02-01
Fix build-deps.cmd for Visual Studio 2019

Henri-J-Norden:
Incorporated some fixes from #1993, so the project compiles with VS2019. This is still compatible with _only_ VS2019, since the mpfr directory is hard-coded to `build.vs19\`

_(Previous pull request #2013 closed due to repo restructuring)_


aothms:
Thanks. Much appreciated. It's really a pain to keep the build scripts up to date so any help here is massively appreciated.

I also find HDF5 takes quite long, I have the suspicion it's mostly building and linking all the tests (execution for most of them is quick). Perhaps this can be eliminated by invoking the build using cmake directly instead of using ctest. That way we can also reuse the generator selection and project building from the main build script.

Issue #2006, opened 2022-01-27
Module visibility doesn't quite work with grouped panels


Moult:
In the add-on settings, if I change the module visibility mode, the "Group" panels are still there even though they have no children

vulevukusej:
I'm admittedly a little torn on this one.  Ever since creating the panel groups I havent gone back to the ui preferences again, it just didnt seem necessary, even though a lot of the panels aren't useful to me specifically.  

I guess my concern is that, yes we are offering customization to each individual user, but at the cost of usage consistency.  It's the kind of thing I think will hurt curious users, who can't understand why their BB doesn't look anything like what they see in videos, forums etc.   An example of this would be iTWO, which has literally hundreds of menus and options that can be toggled on and off in abstracted settings menus -> It's ugly.

Anyway, maybe I just woke up and I'm getting too philosophical 😁 - Let me know your thoughts @Moult , also what do you think @Gorgious56, you seem to have a nice sense of aesthetic/ui?

Gorgious56:
I like the current implementation of the module visibility. I think, to solve the actual problem raised by this issue you would just need to parse in the panel `poll` method whether any of the childs of the parent panel are registered and not draw it if not.

I am personally not the biggest fan of the panel design of Blender so I'm biased against it, but we're stuck with it so... 

The million dollar question is how to offer the most functionality with the least visual pollution and the least resistance path to the desired tool, feature or property. But I believe that it's not the UI layout that really matters, but visual contrast, mnemonic devices and the brain plasticity and muscle memory will do the rest in the long run.

Other areas we can explore :

We have a lot of real estate in the 3D view header

![image](https://user-images.githubusercontent.com/25156105/151324425-fc7e94e1-0a96-40f2-8de4-78f38a7a9df3.png)

Also, I really like the right click context menu in other CAD softwares :

![image](https://user-images.githubusercontent.com/25156105/151324574-1ffcd068-07d8-41dc-86b1-291b938d05d1.png)

It's already very crowded in some modes but I think we can leverage it for some tools. also true for right-clicking in the outliner.

Also we can explore the [gpu module](https://docs.blender.org/api/current/gpu.html) to draw things directly in the viewport.

Your concern about new users getting confused while following tutorials is very valid, it's also a big problem of Blender as a whole since it changes so much between each version and the customization possibilities make it possible to have an unrecognizable interface. To top if off, the preferences layout of Blender is also confusing, and most people don't even know you can customize addons that way.

[...]

PR #2030, opened 2022-02-08
auto-hide panel groups when children not active

vulevukusej:


vulevukusej:
So the baby allowed me half an hour to update the code and commit :-).  Note - this still doesn't adress the quirk mentioned here: https://github.com/IfcOpenShell/IfcOpenShell/issues/2006#issuecomment-1024022868

PR #2042, opened 2022-02-12
Fixed build-deps.cmd failure for VS2017

ssg3d:
Script was trying to pick up vs19 instead of vs15 even for builds on vs15, fixed that by adding a check for vs15.


aothms:
Thanks!

PR #2086, opened 2022-03-11
Compute wall net quantities

Gorgious56:
This aims to fix the wall quantities computation not taking into account modifiers (and by extension openings).

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/model/wall.py
+++ b/src/blenderbim/blenderbim/bim/module/model/wall.py
@@ -866,33 +866,24 @@ def calculate_quantities(usecase_path, ifc_file, settings):
     width = obj.dimensions[1] / unit_scale
     height = obj.dimensions[2] / unit_scale
 
-    if product.HasOpenings:
-        # TODO: calculate gross / net
-        gross_footprint_area = 0
-        net_footprint_area = 0
-        gross_side_area = 0
-        net_side_area = 0
-        gross_volume = 0
-        net_volume = 0
-    else:
-        bm_gross = bmesh.new()
-        bm_gross.from_mesh(obj.data)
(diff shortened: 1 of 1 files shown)


Moult:
This is good except that gross vs net isn't as simple as "are modifiers applied". IfcOpenShell by default computes the net geometry. During editing, users selectively convert these baked net geometry into gross geometry + boolean modifiers if they choose to do so.

It's still better than giving a zero everywhere, so merging :)

Gorgious56:
@Moult Oh okay, then if I understand correctly, when the openings have been applied in Blender, the gross calculations will be off. Is it possible to get back the gross quantities from ifcopenshell or the ifc contents ? Should we be using ifcopenshell directly then to compute the quantities ?

Moult:
> when the openings have been applied in Blender, the gross calculations will be off. 

Correct.

> Is it possible to get back the gross quantities from ifcopenshell or the ifc contents ? Should we be using ifcopenshell directly then to compute the quantities ?

Yes, we can either recreate the geometry using IfcOpenShell, calculate using IfcOpenShell directly, or do some checks in Blender whether or not there are openings and make an informed decision. I'm also not sure how fast it is, or exactly what code it should be to calculate directly in IfcOpenShell unfortunately, hence the TODO.

aothms:
I think it was already possible with some low level trickery but I just mapped them to top level properties:

~~~
>>> import ifcopenshell
>>> import ifcopenshell.geom
>>> f = ifcopenshell.open("acad2010_walls.ifc")
>>> it = ifcopenshell.geom.iterator(ifcopenshell.geom.settings(), f)
>>> it.initialize()
True
>>> elem = it.get_native()
>>> elem.surface_area
10.220161925180994
>>> elem.volume
0.9425983811263172
~~~

PR #2093, opened 2022-03-16
Modification of triangulated face set creation

ilkinkavi:
This is a small modification to the IfcTriangulatedFaceSet creation. In the case of geometry having a UV Map, IfcTriangulatedFaceSet and IfcIndexedTriangleTextureMap values should match for interoperability reasons. This code block rewrites the IfcCartesianPointList3D and assigns every triangle their own cartesian points instead of reusing one cartesian point for different triangles.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/geometry/add_representation.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/geometry/add_representation.py
@@ -543,6 +543,20 @@ class Usecase:
         else:
             items = [self.file.createIfcTriangulatedFaceSet(coordinates, None, None, i) for i in ifc_raw_items if i]
 
+#This assigns every triangle their own Cartesian Points.          
+        tempCoord_IndexList=[]
+        finalCoord_IndexList=[]
+        if self.settings["should_generate_uvs"]:
+            for x in (coord_index):
+                tempCoord_IndexList = x
+                for v in range(len(tempCoord_IndexList)):
+                    if not tempCoord_IndexList:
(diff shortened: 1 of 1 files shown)


aothms:
Many thanks. I'm not very familiar with the code, so @Moult has the final say in this.

src/ifcopenshell-python/ifcopenshell/api/geometry/add_representation.py
```diff
@@ -543,6 +543,20 @@ def create_triangulated_face_set(self):
         else:
             items = [self.file.createIfcTriangulatedFaceSet(coordinates, None, None, i) for i in ifc_raw_items if i]
 
+#This assigns every triangle their own Cartesian Points.          
+        tempCoord_IndexList=[]
+        finalCoord_IndexList=[]
+        if self.settings["should_generate_uvs"]:
+            for x in (coord_index):
```
aothms, line 550: One problem at least that I see is that you depend on `coord_index` which is a scoped to a for loop, but use it outside of the loop. That's for me a clear indication, that the code will likely have issues in case of multiple elements in `ifc_raw_items`. Where your code will only inspect the last element, but affect `items[0]`. I would propose to merge your code with the block starting on line `534`.

Also please run `black` (`python -m black add_representation.py`) so that the code formatting is consistent with the rest of the code base.

aothms (changes requested):

theoryshaw:
Hi @ilkinkavi 

Issue #2078, opened 2022-03-04
Blenderbim import failure with ReflectanceMethod = FLAT


brunopostle:
This Surface Style definition fails to import into Blenderbim, it is ok if the ReflectanceMethod is set to PLASTIC, but not if it is FLAT:

```
ISO-10303-21;
HEADER;
FILE_DESCRIPTION(('ViewDefinition[DesignTransferView]'),'2;1');
FILE_NAME('/dev/null','2022-03-04T20:48:32+00:00',(),(),'IfcOpenShell 0.7.0','IfcOpenShell 0.7.0','Nobody');
FILE_SCHEMA(('IFC4'));
ENDSEC;
DATA;
#1=IFCPROJECT('3xhJn6etX0A9BtOyOiYa07',$,'Test',$,$,$,$,$,#5);
#2=IFCSIUNIT(*,.LENGTHUNIT.,$,.METRE.);
#3=IFCSIUNIT(*,.AREAUNIT.,$,.SQUARE_METRE.);
#4=IFCSIUNIT(*,.VOLUMEUNIT.,$,.CUBIC_METRE.);
#5=IFCUNITASSIGNMENT((#3,#4,#2));
#6=IFCSURFACESTYLE('A FLAT style',.BOTH.,(#7));
#7=IFCSURFACESTYLERENDERING(#8,0.8,$,$,$,$,$,$,.FLAT.);
#8=IFCCOLOURRGB('Bluesy',0.2,0.2,1.);
ENDSEC;
END-ISO-10303-21;
```

Blenderbim error:

```
Python: Traceback (most recent call last):
  File "/usr/share/blender/3.0/scripts/addons/blenderbim/bim/module/project/operator.py", line 499, in execute
    bpy.ops.bim.load_project_elements()
  File "/usr/share/blender/3.0/scripts/modules/bpy/ops.py", line 132, in __call__
    ret = _op_call(self.idname_py(), None, kw)
RuntimeError: Error: Python: Traceback (most recent call last):
  File "/usr/share/blender/3.0/scripts/addons/blenderbim/bim/module/project/operator.py", line 554, in execute
    ifc_importer.execute()
  File "/usr/share/blender/3.0/scripts/addons/blenderbim/bim/import_ifc.py", line 222, in execute
    self.create_styles()
  File "/usr/share/blender/3.0/scripts/addons/blenderbim/bim/import_ifc.py", line 1291, in create_styles
    self.create_style(style)
[...]

PR #2100, opened 2022-03-22
Removed apparently unused and buggy lines in BlenderBIM's IFC importer

yorikvanhavre:
These two lines cause an error when importing https://gitlab.com/openingdesign/101_W_33rd_St/-/blob/main/Models%20and%20CAD/IFC/101_W_33rd_St_Details_D.ifc
Removing them seems safe, they seem buggy anyway (IfcOpenShell.geom.settings don't have a 'materials' member) and materials are still correctly created, but I'm creating this PR mostly to ping @Moult to have a better look.

### DIFF
--- a/src/blenderbim/blenderbim/bim/import_ifc.py
+++ b/src/blenderbim/blenderbim/bim/import_ifc.py
@@ -1356,12 +1356,12 @@ class IfcImporter:
         elif surface_style.ReflectanceMethod == "FLAT":
             blender_material.use_nodes = True
 
-            output = {n.type: n for n in self.settings["material"].node_tree.nodes}.get("OUTPUT_MATERIAL", None)
+            #output = {n.type: n for n in self.settings["material"].node_tree.nodes}.get("OUTPUT_MATERIAL", None)
             bsdf = blender_material.node_tree.nodes["Principled BSDF"]
 
             mix = blender_material.node_tree.nodes.new(type="ShaderNodeMixShader")
             mix.location = bsdf.location
-            blender_material.node_tree.links.new(lightpath.outputs[0], output.inputs["Surface"])
+            #blender_material.node_tree.links.new(lightpath.outputs[0], output.inputs["Surface"])
 
             blender_material.node_tree.nodes.remove(bsdf)
 


Moult:
Thanks @yorikvanhavre ! This is actually a duplicate of https://github.com/IfcOpenShell/IfcOpenShell/issues/2078

Actually removing them isn't the right thing to do - it's actually an implementation of IFC shaders. IFC supports three different types of shaders which is fully compatible with X3D and glTF - it supports FLAT (unlit shadless), PHONG (old-school biased rendering, and PHYSICAL (modern PBR rendering).

It seems as though there were a few typos in my FLAT implementation :)

Here's the result if you're curious. On the left you can see the shader tree. You can see it is actually compatible with glTF in Blender here: https://docs.blender.org/manual/en/dev/addons/import_export/scene_gltf2.html#exporting-a-shadeless-unlit-material

2022-03-23-201816_1436x689_scrot

By the way I was very excited to see the use of IfcAnnotation! I think with a very little bit of work we can actually make annotations roundtrip between Blender and FreeCAD! All that is needed is to tag them with the appropriate IFC object type (e.g. ObjectType=DIMENSION) and associate it with a drawing object in a group, or within a drawing (e.g. IfcDocumentInformation) and then it will come through. I manually changed a couple and here's the result:

2022-03-23-202722_851x761_scrot

yorikvanhavre:
I understand fro the [doc](https://standards.buildingsmart.org/IFC/DEV/IFC4_2/FINAL/HTML/) ObjectType can be anything, but there are  ¨recommended" values to be used... We might want to already define a few on our own...

So what we need to implement in FreeCAD (I think) (sorry I'm hijacking this issue :) )

* Support [Draft Labels](https://wiki.freecadweb.org/Draft_Label)
* Set ObjectType when exporting an IfcAnnotation. In FreeCAD I suppose we would have:
   * Dimensions
   * Linework
   * Texts
   * Areas/surfaces (faces and hatches)
   * Smart labels
   * Undefined (for all the rest I cannot think of)
* For "usermade" labels, (a line + a text), find a way to export them as one single IfcAnnotation
* Add IfcRelAssignsToProduct when exporting smart labels and dimensions (if associated to an object)
* "associate it with a drawing object in a group, or within a drawing (e.g. IfcDocumentInformation)" @Moult care to explain a bit more how that works? What is a drawing object? So far FreeCAD doesn't export anything related to documents/drawings, but might not be too hard...

PR #2118, opened 2022-03-31
Update selector.py

bbrangeo:
Change l 188

value = filter_rule.children[2].children[0] in ["TRUE", "true"]


aothms:
Thanks. From the grammar I see:

BOOLEAN: "TRUE" | "FALSE" | "true" | "false"| "True" | "False"

So probably add 'True' or `filter_rule.children[2].children[0].lower() == 'true'`?

bbrangeo:
Yes thanks

bbrangeo:
filter_rule.children[2].children[0].lower() == 'true' is perfect

bbrangeo:
Thanks

PR #2129, opened 2022-04-07
Minor fixes to BlenderBIM's IFC importer

yorikvanhavre:
* Allow IfcAnnotations which have no related IfcGroup
* Added a couple of IfcAnnotation.ObjectType types exported by FreeCAD

### DIFF
--- a/src/blenderbim/blenderbim/bim/import_ifc.py
+++ b/src/blenderbim/blenderbim/bim/import_ifc.py
@@ -1510,8 +1510,10 @@ class IfcImporter:
         elif element.is_a("IfcStructuralConnection"):
             return self.structural_connection_collection.objects.link(obj)
         elif element.is_a("IfcAnnotation") and self.is_drawing_annotation(element):
-            group = [r for r in element.HasAssignments if r.is_a("IfcRelAssignsToGroup")][0].RelatingGroup
-            return self.collections[group.GlobalId].objects.link(obj)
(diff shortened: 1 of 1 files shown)


Moult:
```
            if group: # annotations could exist that are not in an IfcGroup
```

Is this change hunk required? I don't see the purpose of it. Have you got a file to test where it is needed?

The `place_object_in_spatial_decomposition_collection` function decides where to place an object in the Blender collection tree. If it's an annotation in a drawing, then I place it in a particular Blender collection. Otherwise, it follows the rules of all other IFC elements (e.g. follows the spatial hierarchy or whatever).

yorikvanhavre:
This file doesn't open with the current code: https://gitlab.com/openingdesign/101_W_33rd_St/-/blob/main/Models%20and%20CAD/IFC/101_W_33rd_St_Details_D.ifc it fails because the "group" list is empty

theoryshaw:
Thanks @yorikvanhavre 
exact file version here: https://gitlab.com/openingdesign/101_W_33rd_St/-/blob/1288e92f34e5c6dd9e38a229a012b5488d90f327/Models%20and%20CAD/IFC/101_W_33rd_St_Details_D.ifc

Moult:
I tested with the file and it works without that hunk. Are you sure that file demonstrates an issue?

Moult:
Bump @yorikvanhavre :)

yorikvanhavre:
Indeed everything seems working now... weird! Anyway, the best bugs are the bugs that fix themselves :)

PR #2132, opened 2022-04-08
PR daily builds of ifcopenshell and upload to conda

Krande:
Here is my suggestion for how we could create daily builds of ifcopenshell uploaded to our conda ifcopenshell channel using github actions. 

I think there is still some remaining work to be done to successfully run tests prior to upload to conda. But that should not be a big task (I just haven't had time to look properly into it yet).

Note! I had to make some edits to the CMakelists.txt file. This is something I did a while back to get it to compile at the time on macos and/or windows (if I recall correctly). It's primarily related to the zlib library and platform specific suffixes. If this change breaks other builds I am more than happy to find another solution. 

Let me know what you think,

Best Regards
Kristoffer


conda/conda_build_config.yaml
```diff
@@ -0,0 +1,2 @@
+CONDA_BUILD_SYSROOT:
+  - /Applications/Xcode_13.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk       # [osx]
```
aothms, line 2: This looks like it might break in the future with `macos-latest`.

.github/workflows/ci-daily-build.yml
```diff
@@ -0,0 +1,42 @@
...
+      fail-fast: false
+      matrix:
+        pyver: [
+          { name: py39, distver: '3.9' },
+          { name: py310, distver: '3.10'}
+        ]
+        platform: [
+          { name: Windows, distver: windows-latest, short: 'win-64' },
+          { name: Linux, distver: ubuntu-latest, short: 'linux-64' },
+          { name: macOS, distver: macos-latest, short: 'osx-64' }
```
aothms, line 28: Should we use slightly older versions of OSes so that there is a greater chance that the binaries can be ran on older systems. E.g on linux this mostly depends on the version of glibc.

aothms:
Great. Have a look at the comments.

> run tests prior to upload to conda

We have tests in the ci workflow. We should probably just make execution of this workflow dependent on the success of the ci.yml?

.github/workflows/ci-daily-build.yml
```diff
@@ -0,0 +1,42 @@
+name: ci-ifcopenshell-daily
+
+on:
+  push:
+    paths:
+      - src/**
+      - conda/**
+      - .github/workflows/ci-daily-build.yml
+    branches:
+      - pr-daily-builds
```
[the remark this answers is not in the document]
Krande: yes, when the other changes are given the "green light" I will remove this. Right now I keep it in place so that I can test the github actions pipeline on my fork

cmake/CMakeLists.txt
```diff
@@ -441,30 +441,34 @@ if(HDF5_SUPPORT)
...
+            set(zlib_post libz)
+            set(lib_ext dylib)
         else()
-        	set(lib_ext a)
+            # linux
+            set(zlib_post libz)
+            set(lib_ext so)
         endif()
 
-	if ("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
```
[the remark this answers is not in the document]
Krande: Yes, that must have been mistakenly removed by me. I will restore it

conda/conda_build_config.yaml
Krande: Yes, you are right. I can instead manually download specific MacOSX SDK's from https://github.com/phracker/MacOSX-SDKs (which seems to be a common practice elsewhere) to a specific location on the runner and use that instead.

.github/workflows/ci-daily-build.yml
Krande: Sure, that's fine by me. We could in principle test on multiple versions for each platform and just add an extra flag on the specific version of each platform we wish to upload to conda. Like so:

```
platform: [
          { name: Windows, distver: windows-latest, short: 'win-64', upload: false },
          { name: Windows, distver: windows-2019, short: 'win-64', upload: true },
          ...
]
```

That way we can test on all versions and still only compile the uploaded conda-packages for a specific platform version. If that makes any sense?

Krande:
> We have tests in the ci workflow. We should probably just make execution of this workflow dependent on the success of the ci.yml?

Hm, I see your point. But wouldn't we want to test the package on all 3 platforms before uploading it to conda? For linux only it would make sense, but our current ci-testing is currently only for linux right?

.github/workflows/ci-daily-build.yml
aothms: Great idea. Maybe it actually doens't matter so much. At least on Windows I think conda tries to find the appropriate compiler for the given python version. But I'm not sure how this works on nixes where there isn't a single authoritative python distribution and distro/compiler/glibc are more closely tied together.

aothms:
> our current ci-testing is currently only on linux?

But then maybe we should fix that? It's not a huge priority for me though, it's not so common these days anymore to find platform-specific bugs. Compilers have gotten a lot more compliant these days and the code bases and dependencies have matured.

cmake/CMakeLists.txt
```diff
@@ -441,30 +441,34 @@ if(HDF5_SUPPORT)
...
-        SET(HDF5_LIBRARIES 
-            "${HDF5_LIBRARY_DIR}/libhdf5_cpp${debug_postfix}.${lib_ext}"
-            "${HDF5_LIBRARY_DIR}/libhdf5${debug_postfix}.${lib_ext}"
-            "${HDF5_LIBRARY_DIR}/libz${zlib_post}${debug_postfix}.${lib_ext}"
-            "${HDF5_LIBRARY_DIR}/libsz${debug_postfix}.${lib_ext}"
-            "${HDF5_LIBRARY_DIR}/libaec${debug_postfix}.${lib_ext}"
+        SET(HDF5_LIBRARIES
+            "${HDF5_LIBRARY_DIR}/libhdf5_cpp.${lib_ext}"
+            "${HDF5_LIBRARY_DIR}/libhdf5.${lib_ext}"
+            "${HDF5_LIBRARY_DIR}/${zlib_post}.${lib_ext}"
```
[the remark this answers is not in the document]
Krande: I might have messed something up in the merge, but I do remember that I had to remove the permanent `libz` part in `libz${zlib_post}` and make it platform conditional so that it was able to find the right file on windows (which I believe is `zlib.lib` not `libz.lib`).

The `debug_postfix` is something relatively new in the v0.7.0 branch? So that could probably have been removed accidentally in my merge.

You are welcome to propose another solution. Or would you like me to propose something?

cmake/CMakeLists.txt
```diff
@@ -441,30 +441,34 @@ if(HDF5_SUPPORT)
...
-        	set(zlib_post lib)
+            # Windows
+            set(zlib_post zlib)
         	set(lib_ext lib)
+        elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+            # macOS
+            set(zlib_post libz)
+            set(lib_ext dylib)
         else()
-        	set(lib_ext a)
```
aothms, line 449: Hm, I now also see this change, this also needs to be conditional. Because we can't have the static builds reference dynamic libraries.

cmake/CMakeLists.txt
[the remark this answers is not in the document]
aothms: Yes that's is the common, sensible name, but unfortunately the HDF5 install (that we use in build-deps and build-all to obtain zlib) deviates from that. So we have to account for that case. 

I guess we should replace all this with proper:

    FIND_LIBRARY(NAMES z libz libz_debug PATHS ... NO_DEFAULT_PATH)

That way we can also remove all the `lib_ext` stuff.

But an alternative is to extend the ugly trick we currently use to detect a conda build and put different logic there

https://github.com/IfcOpenShell/IfcOpenShell/blob/19fea3684f330747093c3072f6af7e9a9dc98883/cmake/CMakeLists.txt#L170-L173

~~~
 Directory of C:\Users\tkrij\Documents\IfcOpenShell\v0.7.0\_deps-vs2017-x64-installed\HDF5-1.8.22-win64\lib

07/10/2021  09:52 PM    <DIR>          .
07/10/2021  09:52 PM    <DIR>          ..
07/10/2021  09:40 PM           492,348 hdf5.lib
07/10/2021  09:41 PM           385,702 hdf5_cpp.lib
07/10/2021  09:44 PM            26,826 hdf5_hl.lib
07/10/2021  09:47 PM            11,752 hdf5_hl_cpp.lib
07/10/2021  09:40 PM            31,636 hdf5_tools.lib
07/10/2021  09:37 PM           118,598 libaec.lib
07/10/2021  09:43 PM        13,694,350 libhdf5.lib
07/10/2021  09:35 PM             2,985 libhdf5.settings
07/10/2021  09:47 PM         4,612,856 libhdf5_cpp.lib
07/10/2021  09:44 PM           565,292 libhdf5_hl.lib
07/10/2021  09:47 PM            38,210 libhdf5_hl_cpp.lib
07/10/2021  09:44 PM         1,086,940 libhdf5_tools.lib
07/10/2021  09:37 PM            20,496 libsz.lib
07/10/2021  09:38 PM           327,580 libzlib.lib
                                       ^^^^^^^^^^^
07/10/2021  09:52 PM    <DIR>          pkgconfig
              14 File(s)     21,415,571 bytes
               3 Dir(s)   1,548,943,360 bytes free
~~~

cmake/CMakeLists.txt
aothms: > but do we need to add which paths it should look for zlib at or is it complete the way you suggested it

The `PATHS` should then be followed by the directories where to search.

I tend to add `NO_DEFAULT_PATH` so that cmake doens't get clever looking in other places.

For now, hack I think is fine, the cmake script needs a lot of work anyway.

Krande:
I just added a conditional flag to let the conda build process only happen on a successful completion of the ci.yml workflow. I haven't checked the results from my fork just yet, but I am hopeful it will work. 

BTW I did check to see if I got it working to test on all platforms. However, it looks like some of the test files either has special characters or their paths are too long so that it breaks in the process of preparing the test environment on the Github Actions runners for Windows.

I think everything else we discussed should be in order. So if everything else looks OK to you and my proposal does not break any of the other stuff we could do a squash and merge and test this out on the main v0.7.0 branch? (the conditional workflow stuff apparently doesn't work on other branches

If you agree give me a heads up first and I can remove the "temporary test stuff" inside the workflow yml file that I use for testing on my fork before squashing.

aothms:
Thanks for your continued efforts, when using the win build script I now get:

    LINK : fatal error LNK1181: cannot open input file 'C:\Users\Administrator\IfcOpenShell\_deps-vs2017-x64-installed\HDF5-1.8.22-win64\lib\zlib.lib' [C:\Users\Administrator\IfcOpenShell\_build-vs2017-x64\ifcwrap\_ifcopenshell_wrapper.vcxp
roj]

When using the nix build script I get

~~~
/usr/bin/ld: /home/ubuntu/IfcOpenShell/build/Linux/x86_64/install/hdf5-1.12.1/lib/libhdf5.a(H5Z.c.o): in function `H5Z__init_package':
/home/ubuntu/IfcOpenShell/build/Linux/x86_64/build/CMake-hdf5-1.12.1/hdf5-1.12.1/src/H5Z.c:107: undefined reference to `SZ_encoder_enabled'
/usr/bin/ld: /home/ubuntu/IfcOpenShell/build/Linux/x86_64/install/hdf5-1.12.1/lib/libhdf5.a(H5Zszip.c.o): in function `H5Z__filter_szip':
/home/ubuntu/IfcOpenShell/build/Linux/x86_64/build/CMake-hdf5-1.12.1/hdf5-1.12.1/src/H5Zszip.c:309: undefined reference to `SZ_BufftoBuffDecompress'
/usr/bin/ld: /home/ubuntu/IfcOpenShell/build/Linux/x86_64/build/CMake-hdf5-1.12.1/hdf5-1.12.1/src/H5Zszip.c:336: undefined reference to `SZ_BufftoBuffCompress'
~~~

So the removal of `libsz` and `libaec` and the truncation of `libz${zlib_post}` to `${zlib_post}` should still be made conditional. Or do you have other proposals?

Krande:
I agree. In my latest suggestion I now revert to the original v0.7.0 implementation of handling of zlib extension when it is not building a conda package. 

Let me know if this fixes things on your end?

Krande:
Ping @aothms :) Should we get a conda package build of ifcopenshell rolling? If you prefer I could (with some help) test the nix script myself.

aothms:
They're also working on it on the freecad side of things with conda-forge. https://github.com/conda-forge/ifcopenshell-feedstock/pull/20

It'd be good to finalize things nonetheless.

> If you prefer I could (with some help) test the nix script myself.

Sorry I lost track of the discussion a bit. It's just a matter of `cd nix; python3 build-all.py` :) Would be great if you can have a look.

Krande:
@aothms I tried running the nix build with a docker container and with github actions, but it seems it requires more memory than my wsl vm or github actions VM was willing to provide so it keeps crashing. 

I can take a stab at the windows compilation to ensure that it works okay. But if you could test the nix builds on your end? If they both pass then maybe we can finalize this PR? 

Once that's done we can also finalize https://github.com/conda-forge/ifcopenshell-feedstock/pull/20

aothms:
nix build completed successfully in my quick test 👍 

Don't forget to revert 22a552a9a

Krande:
Great! 

Will revert that in a moment. I tried running the windows compilation in the win/ directory, but I ran into an issue with the Boost library. Not sure if that was just a simple error in my dev setup. If you have a chance to test the windows compilation as well, I think we are ready for a merge after I revert the final changes

Krande:
Almost ready now. But for some reason it seems I am including changes to the submodules mvd and svgfill in my PR.. I am not familiar with submodules and PR's and how I should treat them. So if you spot something fishy, please let me know!

aothms:
Submodules are a pain. They don't automatically update when you switch branches and then later you inadvertently commit them as changes.

See my recent comment here on how to revert the changes https://github.com/IfcOpenShell/IfcOpenShell/pull/2249#issuecomment-1169604891

The windows build script seems to work as intended as well with your proposed changes :)

Krande:
Thanks, now the submodules are aligned (nevermind the strange commit message for the svg module. I was a bit quick and just used the same commit message that I used in the previous commit).

Then I think we are ready to merge?

aothms:
Great work, thanks for your perseverance. I leave on vacation tomorrow, feel free to coordinate with @looooo in case of any tagging or follow up work required. You have commit access right?

Krande:
Hm, I do not seem to have commit access. Either you can quickly add a tag now or I can do it if you give me commit access. Either way is fine for me

Krande:
Then I believe it's done. I've committed and added a tag here -> https://github.com/IfcOpenShell/IfcOpenShell/releases/tag/v0.7.0a4. 

Then I wish you all the best on your vacation! And then we can continue the work once you're back :)

Moult:
@Krande I've updated the install instructions https://blenderbim.org/docs-python/ifcopenshell-python/installation.html#conda - can you confirm the need to have -c conda-forge in the install command? Or let me know if the docs need updating in any way.

I also noticed in this PR that IFCXML, DAE, and GLTF support is disabled? Doesn't this mean that IfcOpenShell won't support .ifcxml, and ifcconvert won't support dae/gltf? Is that intentional?

Krande:
@Moult 

The conda forge commands you've added to the docs are correct. 

Regarding features in the conda daily build. I can't recall exactly why those features were turned off. Maybe I turned those off when adding/debugging occt==7.6 support. 

Nonetheless, I can test turning them ON again on my branch to ensure that including them don't break anything. 

I can start by setting these parameters:
```
-DIFCXML_SUPPORT:BOOL=ON \
-DGLTF_SUPPORT=On \
```

@aothms @Moult Do we have any preferences on which features we would like to be compiled in the conda daily builds?

Let me know if there are any other flags that needs to be switched on?

Moult:
My uneducated preferences would be for conda to match IfcOpenBot :) So that includes ifcxml, gltf and hdf5.

Krande:
Okay, HDF5 is already being compiled, and with the following two changes
```
-DIFCXML_SUPPORT:BOOL=ON \
-DGLTF_SUPPORT=On \
```
I believe the daily build should be aligned (feature-wise) with IfcOpenBot. All compilations seems to pass on my branch with the proposed changes, so I will push these changes to the main repo asap.

aothms:
I think

    -DCOLLADA_SUPPORT=Off

Would still be different.

I don't see an OpenCOLLADA project on conda-forge though (there isn't a big push for CPP only packages obviously). Strangely I do see https://anaconda.org/conda-forge/collada-dom

So we either need to

- Accept it and move on (we can convert from .glb anyway)
- Built OpenCOLLADA as part of build.sh|bat
- Fire up an OpenCOLLADA recipe

Being pragmatic I guess the first option doesn't sound so bad.

Krande:
Okay, that's good to know! I guess there are some dependencies on OpenCOLLADA in certain IfcConvert conversion options?

For me personally I have never directly used OpenCOLLADA, so would be perfectly fine with option 1 :)

PR #2135, opened 2022-04-12
Tweaks to IDS

dimven:
Hi @atomczak ,

I'm a colleague of Krzystof and together we used your IDS implementation on a large scale project. We had to make a few tweaks to make it work as desired and I wanted to share them with you. Our files were very large and had a lot of issues, so most of the focus for these changes revolved around reducing the size of the bcf reports to make them easier to manage by the users and to improve the overall performance wherever possible.

- we shifted the outer loop to go through the elements first and then through the specs, so that the elements are iterated only once
- we trimmed the description of the issues
- we fixed a bug where if both the instance and the type have a pset/property with the same name, they get overwritten and the report is wrong.
- we revised the bcf viewport creation, so that the offending elements show up in solibri correctly

Hopefully you'll find some of these changes useful too.

Regards,
Dimitar

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/ids.py
+++ b/src/ifcopenshell-python/ifcopenshell/ids.py
@@ -22,7 +22,7 @@ import logging
 import operator
 import os
 import numpy as np
-import datetime
+from datetime import date
 
 import ifcopenshell.util.element
 import ifcopenshell.util.placement
@@ -36,7 +36,7 @@ from xmlschema.validators import identities
 
 
 cwd = os.path.dirname(os.path.realpath(__file__))
-ids_schema = XMLSchema(os.path.join(cwd, "ids.xsd"))  # source: "http://standards.buildingsmart.org/IDS/ids_05.xsd"
+ids_schema = XMLSchema(os.path.join(cwd, "ids.xsd"))  # source: "http://standards.buildingsmart.org/IDS/ids_04.xsd"
 
 
 def error(msg):
@@ -48,29 +48,29 @@ class ids:
 
     def __init__(
         self,
-        title="Name",
-        copyright=None,
-        version=None,
+        ifcversion=None,
         description=None,
         author=None,
-        date=None,
+        copyright=None,
+        version=None,
+        creation_date=None,
         purpose=None,
         milestone=None,
     ):
         """Create an IDS object.
 
-        :param title: Name of the IDS file, defaults to None
-        :type title: str, required
-        :param copyright:, defaults to None
-        :type copyright: str, optional
-        :param version: IDS file version, defaults to None
-        :type version: float, optional
+        :param ifcversion: IFC schema version. If None, then schema independent. Options: '2.3.0.1'|'4.0.2.1'|'4.3.0.0'|None, defaults to None
+        :type ifcversion: str, optional
         :param description:, defaults to None
         :type description: str, optional
         :param author: Email of the IDS author, defaults to None
         :type author: str, optional
-        :param date: Date in 'yyyy-mm-dd' format, defaults to current date
-        :type date: str, optional
+        :param copyright:, defaults to None
+        :type copyright: str, optional
+        :param version: IDS file version, defaults to None
+        :type version: float, optional
+        :param creation_date: Date in 'yyyy-mm-dd' format, defaults to current date
+        :type creation_date: str, optional
         :param purpose:, defaults to None
         :type purpose: str, optional
         :param milestone:, defaults to None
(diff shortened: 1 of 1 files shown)


Moult:
Bump and ping @atomczak :)

atomczak:
Hi @dimven! Great! Sorry that I missed your PR, Gmail decided to skip that notification. I will review and test it as soon as I can!

src/ifcopenshell-python/ifcopenshell/ids.py
```diff
@@ -48,52 +48,53 @@ class ids:
...
+        :type creation_date: str, optional
         :param purpose:, defaults to None
         :type purpose: str, optional
         :param milestone:, defaults to None
         :type milestone: str, optional
         """
         self.specifications = []
         self.info = {}
-        if title:
-            self.info["title"] = title
```
Moult, line 82: Why is this deleted?

src/ifcopenshell-python/ifcopenshell/ids.py
```diff
@@ -270,8 +263,7 @@ def asdict(self):
         # if older python collections.OrderedDict()
         spec_dict = {
             "@name": self.name,
-            "@use": self.use,
-            "@ifcVersion": self.ifcVersion,
+            "@necessity": self.necessity,
```
Moult, line 266: I don't see this in either the IDS v4 or v5 XSDs?

src/ifcopenshell-python/ifcopenshell/ids.py
```diff
@@ -1276,52 +1165,42 @@ def emit(self, log_content):
...
         viewpoint = bcf.Viewpoint()
         viewpoint.perspective_camera = bcf.PerspectiveCamera()
-        ifc_elem = log_content.msg["ifc_element"]
-        # ifc_elem = ifc_file.by_guid(log_content.msg["guid"])
+
+        ifc_elem = data["ifc_element"]
         target_position = np.array(ifcopenshell.util.placement.get_local_placement(ifc_elem.ObjectPlacement))
         target_position = target_position[:, 3][0:3]
-        camera_position = target_position + np.array((5, 5, 5))
+        target_position /= 1000 # convert to meters
```
Moult, line 1182: This hardcoded unit conversion doesn't quite look right. Perhaps use something in ifcopenshell.util.unit to help convert?

Moult (commented):
Hey @dimven I had a quick look and a lot of the issues seems to be related to going back to IDS v4 instead of IDS v5. Is there a reason for using IDS v4?

dimven:
Hi @Moult 

We started looking into this a few months ago and I guess v4 was the latest back then. This was really not intended as a PR but more of a reference. If you find some of the changes useful, it would be great if you decide the implement them. Otherwise feel free to close this PR. We were trying to apply it on a large-scale project that was generating reports with thousands of issues and we were trying to review the bcfs mainly in Solibri.

One big problem was that the errant elements were not being reported at all and it was because the bcf were being saved with a perspective camera. Each software interprets the perspective cone differently and we were getting broken views with incredibly narrow or distant focal points that led to the exclusion of the elements. Once we switched things to simple orthographic cameras, everything started working as expected.

Another issue was the incredibly long and detailed titles of each issue. Not only was this generating huge reports, but it was also breaking the appearance in Solibri, with the text overflowing over the viewports. It would be great if the final titles were more succinct.

Finally, there were some issues in the processing logic that was misreporting or missing certain property requirements. That we had to work around to get the expected results.

aothms:
> This was really not intended as a PR but more of a reference

That'd really be a shame I think. I haven't followed all of the discussion, but this kind of first hand practical usage experience is really valuable to embed in the code. At the same time we have to make sure the code remains generally applicable. So for example an option to switch between long titles (suitable for text reports?) and short titles (for consumption in UI with limited screen estate?) seems meaningful. In python these kind of options are easily provided using e.g `argparse`. IDS is evolving rapidly so it'd really be a shame if our code bases diverge before IDS v1 is even public. Taking a little bit of time to generalize your insights seems worthwhile in the end and we are here to help with that.

> Finally, there were some issues in the processing logic that was misreporting or missing certain property requirements. That we had to work around to get the expected results.

This seems to be crucial feedback. Can you elaborate?

src/ifcopenshell-python/ifcopenshell/ids.py
atomczak: 'necessity' was in XSD for some time, but got replaced by 'use'. Now it will probably also be replaced by minOccur in IDS XSD.

Moult:
Closing, as I've reviewed the changes in this PR and I think performance wise the new IfcTester is pretty good and I'm running it on large models in decent time. There are also a number of formatting flexibility improvements with new support for TXT, JSON, HTML, ODS, BCF, and Console outputs. BCF has been updated to the latest BCF v2 library updates too.

Issue #1848, opened 2021-10-31
List of modules with and without its own data


maxfb87:
Following this closed PR (#1833), i write here the modules that have `ui.py` which still speak to a `Data class` in `ifcopenshell.api`, ones that have already been migrated to a new `blenderbim.bim.module.* data class` and modules that haven't a `ui` with `data class`

MODULES WITH ui that speaks from  `ifcopenshell.api data `class`

ALL DONE!

MODULES WITH `blenderbim.bim.module.* data class`
- aggregate
- attribute
- boundary
- constraint
- context
- cost
- classification
- group
- layer
- spatial
- geometry
- georeference
- material
- model
- owner
- profile
- pset
- pset_template
- resource
- root
- sequence
- structural
- spatial
- style
- system
- type
- unit
- void

MODULES WITHOUT `data class`
- augin
- bcf
- bimtester
- clash
- cobie
- covetool
- csv
- debug
- diff
- document
- drawing
- lca
- misc
- patch
- project
- qto
- search

Moult:
This is really excellent and it shows that we are much further along than originally thought. I've updated the list with difficulty ratings off the top of my head. Perhaps @Jesusbill can also help hack on some like the structural one.

Jesusbill:
@Moult yes I should start working on this, nice work @maxfb87

maxfb87:
Thanks @Jesusbill .
I would like to help with this refactoring , i need only to figure out how to handle the operator.py calls to data and after that i think i can help with that

Moult:
Cheers, here's the guidelines:

 1. One blenderbim.bim.module.foo.data.WidgetNameData class per widget. e.g. https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/context/data.py#L27
 2. Widget draw function always calls if not WidgetNameData.is_loaded: WidgetNameData.load(). This triggers loading if and only if the widget is visible. The is_loaded and load() is always the same with no variation allowed. e.g. https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/context/ui.py#L38-L39
[...]

PR #2142, opened 2022-04-15
Refactor georeference module

maxfb87:
Following issue #1848 (and all connected issues), i tried to refactor the georeference module.
I followed the developers guide, that i found very useful.
Please check it and tell me if i made mistakes or something could be better.

PS: I was unable to do one thing, look at https://github.com/maxfb87/IfcOpenShell/blob/a1df868bd264f658f839a058ab15a79173baacc6/src/blenderbim/blenderbim/tool/georeference.py#L57
I don't know how to use the .report method from the operator class. Any suggestion?

Thanks,
Massimo


src/blenderbim/blenderbim/bim/module/georeference/operator.py
```diff
@@ -17,345 +17,146 @@
...
-        context.scene.sun_pos_properties.north_offset = -radians(
-            ifcopenshell.util.geolocation.xaxis2angle(
-                float(context.scene.BIMGeoreferenceProperties.map_conversion.get("XAxisAbscissa").string_value),
-                float(context.scene.BIMGeoreferenceProperties.map_conversion.get("XAxisOrdinate").string_value),
-            )
-        )
-        return {"FINISHED"}
-
+    def _execute(self, context):
+        core.disable_editing_georeferencing(tool.Georeference)
```
Moult, line 66: This function doesn't exist?

src/blenderbim/blenderbim/bim/module/georeference/operator.py
```diff
@@ -17,345 +17,146 @@
...
 
-class SetIfcGridNorth(bpy.types.Operator):
+class SetIfcGridNorth(bpy.types.Operator, tool.Ifc.Operator):
     bl_idname = "bim.set_ifc_grid_north"
     bl_label = "Set IFC Grid North"
     bl_options = {"REGISTER", "UNDO"}
+    bl_description = "Set IFC grid north"
+    
+    def _execute(self, context):
+        core.set_ifc_grid_north()
```
Moult, line 75: The function signature is missing an argument?

src/blenderbim/blenderbim/bim/module/georeference/operator.py
```diff
@@ -17,345 +17,146 @@
...
-    bl_idname = "bim.set_blender_true_north"
-    bl_label = "Set Blender True North"
+class SetBlenderGridNorth(bpy.types.Operator, tool.Ifc.Operator):
+    bl_idname = "bim.set_blender_grid_north"
+    bl_label = "Set Blender Grid North"
     bl_options = {"REGISTER", "UNDO"}
+    bl_description = "Set Blender grid north"
+    
+    def _execute(self, context):
+        core.set_blender_grid_north()
```
Moult, line 84: Missing argument?

src/blenderbim/blenderbim/bim/module/georeference/operator.py
```diff
@@ -17,345 +17,146 @@
...
-        return {"FINISHED"}
-
+class GetCursorLocation(bpy.types.Operator, tool.Ifc.Operator):
+    bl_idname = "bim.get_cursor_location"
+    bl_label = "Get Cursor Location"
+    bl_options = {"REGISTER", "UNDO"}
+    bl_description = "Insert the current cursor coordinates"
+    
+    @classmethod
+    def poll(cls, context): #TODO is it right to use the poll method?
```
Moult, line 93: Good point, let's simplify this to just polling return tool.Ifc.get()

src/blenderbim/blenderbim/core/georeference.py
```diff
@@ -0,0 +1,99 @@
...
+    blenderbim.bim.helper.import_attributes(
+        "IfcProjectedCRS",
+        props.projected_crs,
+        GeoreferenceData.data["projected_crs"],
+        georeference.import_projected_crs_attributes,
+    )
+       
+    georeference.clear_map_conversion()
+        
+    blenderbim.bim.helper.import_attributes(
```
Moult, line 43: Don't call blenderbim.bim.helper from the core. The core should only ever call functions from the tool, and nothing else. This way the core can remain abstract and testable without launching Blender. This also forces all behaviour to be encapsulated in functions in a tool, which promotes code reuse.

So maybe something like georeference.import_map_conversion()

Then inside import_map_conversion use blenderbim.bim.helper.import_attributes2() which is simpler to use and you can then remove the need for GeoreferenceData.data["map_conversion"]

src/blenderbim/blenderbim/core/georeference.py
```diff
@@ -0,0 +1,99 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with BlenderBIM Add-on.  If not, see <http://www.gnu.org/licenses/>.
+
+import bpy
+import blenderbim.bim.helper
+
+from blenderbim.bim.module.georeference.data import GeoreferenceData
+
+from math import radians, degrees, atan, tan, cos, sin
```
Moult, line 24: Core should have no imports. All these details should be in a tool instead. See how other cores work :)

src/blenderbim/blenderbim/tool/georeference.py
```diff
@@ -0,0 +1,231 @@
...
+        if not prop.is_null and prop.data_type == "string":
+            # We store our floats as string to prevent single precision data loss
+            attributes[prop.name] = float(prop.string_value)
+            return True
+    
+    @classmethod
+    def set_ifc_grid_north(cls):
+        x_angle = -bpy.context.scene.sun_pos_properties.north_offset
+        bpy.context.scene.BIMGeoreferenceProperties.map_conversion.get("XAxisAbscissa").string_value = str(cos(x_angle))
+        bpy.context.scene.BIMGeoreferenceProperties.map_conversion.get("XAxisOrdinate").string_value = str(sin(x_angle))
```
Moult, line 113: So it sounds crazy why 3 lines of code should be split into three locations, but trust me on this one :) The core should be abstract, so perhaps something like:

def set_ifc_grid_north(cls, georeference):
    georeference.set_xaxis_vector_from_angle(georeference.get_x_angle_from_sun_position())

So the x_angle in the future could come from other sources like a Cycles sun object, these functions could be reused in another operator (e.g. auto generate shadow diagram operator), and so on. So it sounds overkill but it pays off :)

src/blenderbim/blenderbim/bim/module/georeference/operator.py
maxfb87: Yep, i forgot to write the function. I'll fix.

src/blenderbim/blenderbim/bim/module/georeference/operator.py
maxfb87: Yeeep, i forgot it. I'll fix.

src/blenderbim/blenderbim/bim/module/georeference/operator.py
maxfb87: So the following is ok?

@classmethod
def poll(cls, context):
    return file = tool.Ifc.get()

src/blenderbim/blenderbim/core/georeference.py
maxfb87: Ok, i understand. Just to be clear, have i to delete (in "core/georeference.py") also "import bpy" to get the props?

src/blenderbim/blenderbim/core/georeference.py
maxfb87: Ok, i'll try and i'll tell you if i can't make it.

src/blenderbim/blenderbim/tool/georeference.py
maxfb87: Ok, i follow you. I'll try and i'll tell you if i can't make it.

src/blenderbim/blenderbim/core/georeference.py
Moult: Yes, core/georeference.py should never import bpy. Think of any core code as pure Python functions that call tool functions. No knowledge of external libraries like bpy, ifcopenshell, etc. The only purpose is to:

 1. Allow code maintainers to read a short poem that describes what might be a complex operation in simple English
 2. Force code writers to encapsulate all concrete code in well named functions, that promotes maintainability and code reuse
 3. Through testing, quickly identify butterfly effects of changing code that impacts other parts of the system.
 4. Decouple operations from the Blender UI so that operators can be embedded in other operators without incurring the speed penalty from the Blender UI.

src/blenderbim/blenderbim/bim/module/georeference/operator.py
Moult: Yep

maxfb87: ok, l'll fix.

src/blenderbim/blenderbim/tool/georeference.py
maxfb87: Ok, i'll fix. I try to simplify also other methods in the same way (for example def set_blender_grid_north(georeference) in core)

src/blenderbim/blenderbim/core/georeference.py
maxfb87: Ok, i'll fix.

src/blenderbim/blenderbim/core/georeference.py
maxfb87: Ok, so i tried to remove imports from core and put same imports into tool.
The problem is that the import from GeoreferenceData creates a circular import exception.
So, to resolve, i saw what has been done in other module (Owner) and i saw that this problem is resolved avoiding the import GeoreferenceData and read ifc directly.
I'll fix in this way, is it correct?

src/blenderbim/blenderbim/tool/georeference.py
```diff
@@ -0,0 +1,231 @@
...
+            bpy.context.scene.BIMGeoreferenceProperties.true_north_ordinate = str(data[1])
+    
+    @classmethod
+    def get_true_north_props(cls, props):
+        true_north = None
+        if props.has_true_north:
+            try:
+                true_north = [float(props.true_north_abscissa), float(props.true_north_ordinate)]
+            except ValueError:
+                self.report({"ERROR"}, "True North Abscissa and Ordinate expect a number") #TODO use the correct report method
```
maxfb87, line 57: @Moult Here i need help: how can i report the error in a correct way?

src/blenderbim/blenderbim/core/georeference.py
Moult: Tool should not import GeoreferenceData, tool should access IFC directly. The GeoreferenceData (and any other Data classes) are only meant to be accessed by the UI. More reading: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/demo/data.py#L26-L30

Removing the relationships between Operators and Data was one of the goals of the refactor (which you are doing now), because if Operators and Data were related, then that would force you to know how to refresh all the correct portions of the UI for every operation, which is very difficult to know especially if operations can have side effects (e.g. when you delete an object, should an IfcTask in a schedule refresh its quantities? Etc)

src/blenderbim/blenderbim/tool/georeference.py
Moult: For now, just turn it into a print() statement to print to console. There are two possible solutions for this I haven't decided which would be better yet.

Option 1: make the core simply return errors and the operator does the reporting, typical client-server approach
Option 2: allow tools to be able to report errors by giving them access to the active operator during dependency injection by the operator

maxfb87: Ok, for now i only print the error.

My two cents on this: i think the solution 1 is better because it's more simple; also, i think tool as an instrument to create something so, in case of an error, it reports the error to another component (UI, operator) that can manage the error in its own correct way (print, popup, exception, ...)

Moult: I agree, please feel free to implement option 1 :)

maxfb87: Ok, i'll try to do something about the error report, but i'll do a different PR.

PR #2146, opened 2022-04-17
Fix several problems in `build-dep.cmd`

renjianqiang:
1. `IFCOS_INSTALL_PYTHON` is true by default, but the python is not installed after calling the `build-dep.cmd`. And i find the python related variables is empty because they are in the `IF` statement, so i move them out of the `IF` statement to fix the problem.

2. `build-dep.cmd` failed when the build type is `Debug`, due to the `mpir` and `mpfr` isn't copied to the right directory.


aothms:
Thank you for this. Looks good to me!

PR #2158, opened 2022-04-23
Enable option to build QtViewer using Qt5

dushyant-basson:
Edited cmake/CMakeLists.txt & src/qtviewer/CMakeLists.txt, to enable the option to be able to build QtViewer using Qt5. The path to Qt5 can be provided via **CMAKE_PREFIX_PATH** to **win/run-cmake.bat** file like:
`.\run-cmake.bat vs2019-x64 -DCMAKE_PREFIX_PATH="C:/Qt/5.15.2/msvc2019_64"`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,15 @@
 /_installed-vs*-x*/
 /build/
 
+# output directories
+/cmake/out/
+/docs/out/
+/src/examples/out/
+/src/ifcmax/out/
+/src/ifcwrap/out/
+/src/qtviewer/out/
+
+
 /win/BuildDepsCache*.txt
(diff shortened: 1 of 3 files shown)


aothms:
Awesome, curious what comes next :)

aothms:
I've set the option to OFF though for the time being so that CI/CD succeeds https://github.com/IfcOpenShell/IfcOpenShell/commit/c2cad06cfa5c76a8dac2043ba34d6c79bb5ff1d9

dushyant-basson:
Yes, it should be set to 'OFF' by default. I missed that.
Thanks.

PR #2165, opened 2022-04-26
Fix shared libraries exports.

aschvalberg:
Fix missing exports for building IfcGeom and IfcParse as shared libraries.


aothms:
Thanks!

For the files `src/ifcparse/Ifc____.h` please edit the template that generated the code too https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/express/templates.py#L20

PR #2171, opened 2022-04-28
Remove some memory leaks and add possibility to clear any resource allocated by registered schemas

aschvalberg:
It's possible to clear global resources with `IfcParse::clear_schemas();`
Any further request to re-populate the schemas through `IfcParse::schema_by_name()` or directly through `Ifc___::get_schema()` works as expected.

The typical code in IfcAdvanceHouse.cpp sample should now be free from memory leaks.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/express/schema_class.py
+++ b/src/ifcopenshell-python/ifcopenshell/express/schema_class.py
@@ -283,10 +283,22 @@ __attribute__((optnone))
 
         self.statements.extend(
             (
-                "const schema_definition& %s::get_schema() {" % schema_name_title,
+                "static std::unique_ptr<schema_definition> schema;",
                 "",
-                "    static const schema_definition* s = %(schema_name)s_populate_schema();" % locals(),
-                "    return *s;",
+                "void %s::clear_schema() {" % schema_name_title,
+                "    schema.reset();",
+                "}",
+                "",
+            )
+        )
+
+        self.statements.extend(
+            (
+                "const schema_definition& %s::get_schema() {" % schema_name_title,
+                "    if (!schema) {",
+                "        schema.reset(%(schema_name)s_populate_schema());" % locals(),
+                "    }",
+                "    return *schema;",
                 "}",
                 "",
                 "",
--- a/src/ifcparse/IfcParse.cpp
+++ b/src/ifcparse/IfcParse.cpp
@@ -1057,13 +1057,19 @@ std::string IfcEntityInstanceData::toString(bool upper) const {
 	return ss.str();
 }
 
-IfcEntityInstanceData::~IfcEntityInstanceData() {
+void IfcEntityInstanceData::clearArguments()
+{
 	if (attributes_ != NULL) {
 		for (size_t i = 0; i < getArgumentCount(); ++i) {
 			delete attributes_[i];
 		}
 		delete[] attributes_;
+		attributes_ = NULL;
 	}
+}
+
+IfcEntityInstanceData::~IfcEntityInstanceData() {
+	clearArguments();
 }
 
 unsigned IfcEntityInstanceData::set_id(boost::optional<unsigned> i) {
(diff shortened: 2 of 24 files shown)


src/ifcopenshell-python/ifcopenshell/express/schema_class.py
```diff
@@ -311,6 +315,7 @@ def can_be_instantiated(idx_name):
         self.statements[self.statements.index("{factory_placeholder}")] = (
             """
 class %(schema_name)s_instance_factory : public IfcParse::instance_factory {
+    virtual ~%(schema_name)s_instance_factory() { s = 0; }
```
aothms, line 318: I think I understand why you chose this approach, but it does seem a bit odd to have such side effects in a destructor. If we keep it, I think it deserves a quick comment on why and the implications.

Perhaps the line of control needs to be reversed. For example:
- Make `s` a `std::unique_ptr<schema_definition>`.
- Instead of having the destructor clear `s`, why not have a `clear_schema()` function in this namespace that does `s.reset(nullptr)`. 
- If `s` becomes global scope, maybe it also deserves a better name.

src/ifcparse/IfcParse.cpp
```diff
@@ -1407,6 +1413,7 @@ void IfcEntityInstanceData::setArgument(size_t i, Argument* a, IfcUtil::Argument
 	}
 
 	attributes_[i] = copy;
+	delete a;
```
aothms, line 1416: Is it wise to elegant to delete `a` here, since it is just passed to this function? I see how it might be the quickest solution, but probably not the cleanest. I'd say at least add a code comment on this behaviour.

src/ifcparse/IfcParse.cpp
```diff
@@ -2260,8 +2267,15 @@ IfcUtil::IfcBaseClass* IfcFile::instance_by_guid(const std::string& guid) {
 
 // FIXME: Test destructor to delete entity and arg allocations
 IfcFile::~IfcFile() {
-	for( entity_by_id_t::const_iterator it = byid.begin(); it != byid.end(); ++ it ) {
-		delete it->second;
+	std::set<IfcUtil::IfcBaseClass*> entities_to_delete;
+	for (const auto& pair : byid) {
+		entities_to_delete.insert(pair.second);
+	}
+	for (const auto& pair : entity_file_map) {
```
aothms, line 2274: I don't think this is necessary. I can't imagine a scenario where entity_file_map contains data that is not already contained in byid. Did you encounter such cases?

src/ifcparse/IfcSchema.cpp
```diff
@@ -149,3 +158,15 @@ std::vector<std::string> IfcParse::schema_names() {
 
 	return return_value;
 }
+
+void IfcParse::clear_schemas() {
+	// Populate vector with map values
```
aothms, line 163: Why the intermediate vector, why not a single loop and `delete pair.second`?

aothms (changes requested):

aothms:
Thanks for taking the time for this. Great effort!

I've made a couple of comments, and it seems the test segfault. Are you able to investigate that? The tests are written in Python, it's a bit of an additional complexity to debug.

src/ifcparse/IfcSchema.cpp
aschvalberg: currently, the destructor of schema_definition removes itself from the global `schemas`, so we need a local copy first to avoid concurrent iterating and erasing on `schemas`.

src/ifcparse/IfcParse.cpp
aschvalberg: For instance, the IfcPlaneAngleMeasure instance created in `IfcHierarchyHelper::addProject()` is a type_declaration and not an entity. It will be in entity_file_map but not in byid. (from IfcAdvancedHouse sample)

src/ifcparse/IfcParse.cpp
aschvalberg: you're right... all these arguments are created on the heap from client side. As a copy is made inside the function anyway, argument should be passed by const-ref instead of raw pointer. But that implies to change all call sites.

aschvalberg: I've investigated and found also why the python tests failed. The argument should not be deleted in the IfcEntiyInstanceData copy constructor (and it's the only case).
Instead of creating a copy and deleting the argument inside the function, I've added a parameter (default to false) to create a copy if needed (currently only in the IfcEntityInstanceData copy constructor) or directly set and use the argument pointer (the copy is not needed in this case).

aschvalberg:
@aothms do you think this PR still need some more work?
Thanks :-)

aothms:
Looks great. Many thanks again and sorry for the delay.

Issue #2168, opened 2022-04-28
Removing section cut object does not remove node group


schroef:
v0.09999 ?!?!?!)
bl2.93 Win

I was testing out this addon after looking for the cutawayshader. I noticed a link in the blendercommunity post and was happy to see this one works with Eevee as the other one is OSL and thus CPU only.

I was testing the addon only and noticed when removing the section cut object, the nodegroup is not removed from the materials.

tried it couple times and each time nodegroup stays in the scene. Is this perhaps im testing it outside of the BlenderBIM addon?

Moult:
v0.0.999999 means you are using an unstable release which is unversioned :) The 99999 convention comes from Gentoo package management.

Indeed the nodegroup isn't removed it seems. Ping @Gorgious56 who helped implement this, perhaps he knows a quick fix?

Gorgious56:
Indeed there is AFAICR no callback to ensure the shader is removed when the cutaway object is removed.
Perhaps something can be implemented using the msgbus system, or simply adding a button to clear the cutaway shader from the material.
IIRC you should first remove the section cutaway from the interface, and then you can delete the section cutaway object

Are you using the standalone script (https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/scripts/standalone_section_shader.py) or the feature inside the BlenderBIM addon ?

Moult:
Is it too onerous to loop through every material in the project and check that way? A little brute force but it only ever needs to run when there are no more section compare nodes in the node group?

Gorgious56:
I think the problem here is that when the object is removed by the user using X (delete) and not using the button in the addon panel, the addon is never notified of it.

I don't think it would be computationally expensive to parse every material that implements the cutaway shader.
Removing the node group from a material if there is no more section compare inside it can be implemented : when removing a section compare, check if this was the last one, run through every material, disconnect the node group where it makes sense. Although it won't fix the above problem.

There is also this problem that I'm sure you're aware of that an object can be modified / moved / deleted in a myriad of different ways in Blender, so it may not be possible to account for every one of them. For example deleting an object in the 3D viewport is not the same thing as deleting an object from the outliner.

Moult:
I think that's OK, two options I can think of:

 1. Let the user remove the section object, but whenever the user next presses add / remove section, a "clean up" is done in the node tree.
[...]

PR #2172, opened 2022-04-28
Refactor Standalone section shader

Gorgious56:
Follows https://github.com/IfcOpenShell/IfcOpenShell/issues/2168

There are a lot of line changes but mainly some tidying and uncoupling from Blender operators.

Code wise, I tried to remove as much logic as I could from the operators and created a static class for actual operations.

In short :

- New panel interface

image

- New "Refresh" button that cleans the node tree in case the user deleted cutaway empties without using the panel button. It also adds material overrides to objects that weren't part of the cutaway shader beforehand.

- New "Toggle" button that lets the user temporarily disable the cutaway shader while still keeping the empty objects and material override

- "Remove Section" extended to delete all selected section empties rather than only if it was the active object

- New "Remove all sections" button to remove all section objects and purge all data related to the cutter (materials, node groups)

I plan on extending it to the actual code in the BlenderBIM addon codebase, but I'd like feedback if there are things that need improving first.

i did notice some kind of bug with Eevee when removing section shaders where the material goes pink but I couldn't replicate it regularly. It is fixed by clicking on the "Refresh" button usually.

### DIFF
--- a/src/blenderbim/scripts/standalone_section_shader.py
+++ b/src/blenderbim/scripts/standalone_section_shader.py
@@ -286,19 +286,6 @@ class SectionCutawayManager:
                         obj.data.materials.append(material)
 
 
-class BIM_OT_section_plane_include_object(Operator):
-    """Add a temporary empty object as a section cutaway. Cull all geometry rendering below the empty's local Z axis"""
-
-    bl_idname = "bim.section_plane_add"
-    bl_label = "Add Temporary Section Cutaway"
-    bl_options = {"REGISTER", "UNDO"}
-
-    def execute(self, context):
-        SectionCutawayManager.add_section_plane(context)
-        SectionCutawayManager.clean()
-        return {"FINISHED"}
-
-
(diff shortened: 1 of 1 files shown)


Moult:
This looks great! The decoupled class is perfect. When merging into the add-on code base, can you add those classmethods and their signatures into core/tool.py and tool/drawing.py? This is because this sectioning code will likely be reused in other drawing related tasks (e.g. activate drawing).

Gorgious56:
Alright then, sure I'll do that ! Thanks for the feedback,

I'll see if I can get back into unit tests and write something that makes sense there too. :)

Gorgious56:
I'll merge the PR now since I don't really know when I'll have time to work on the BBim version... :)

schroef:
Are those animation options really useful? (see the small diamond behind first 3 options)

schroef:
Today noticed "selected only" doesn't work properly. Everything in my scene received the cutaway. I'll check again tomorrow

PR #2194, opened 2022-05-09
Add missing exports for ifcopenshell-python when we build shared libs.

aschvalberg:
when building IfcParse and IfcGeom as shared libs, some exports are missing in Serializers and libsvgfill libraries (also built as shared libs), and we cannot build ifcopenshell-python.

See also the other PR from sgvfill subrepo https://github.com/IfcOpenShell/svgfill/pull/1 which need to be merged first so I can change the hash of the subrepo here.

### DIFF
--- a/src/ifcwrap/IfcPython.i
+++ b/src/ifcwrap/IfcPython.i
@@ -72,6 +72,8 @@
 	}
 }
 
+%include "../serializers/serializers_api.h"
+
 // Include headers for the typemaps to function. This set of includes,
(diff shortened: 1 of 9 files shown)


aschvalberg:
The svgfill subrepo has been updated to the merged branch of https://github.com/IfcOpenShell/svgfill/pull/1 
This PR can be merged is everything else is OK.

aothms:
Looks great to me. Thanks again for your efforts.

PR #2197, opened 2022-05-11
[IfcConvert] Added IfcRelConnectsElements to xml export

christt105:


src/serializers/schema_dependent/XmlSerializer.cpp
```diff
@@ -665,6 +665,24 @@ void MAKE_TYPE_NAME(XmlSerializer)::finalize() {
...
+	IfcSchema::IfcRelConnectsElements::list::ptr pconnections = file->instances_by_type<IfcSchema::IfcRelConnectsElements>();
+	for (IfcSchema::IfcRelConnectsElements::list::it it = pconnections->begin(); it != pconnections->end(); ++it) {
+		IfcSchema::IfcRelConnectsElements* connection = *it;
+
+		ptree* nconnection = format_entity_instance(connection, connections);
+
+		ptree nrelatedElement;
+		ptree nrelatingElement;
+
+		nrelatedElement.put("<xmlattr>.id",    connection->RelatedElement()->GlobalId());
```
aothms, line 678: Did you consider using `format_entity_instance(..., as_link=true)` instead?

That would probably mean:

`<RelConnects><Related><IfcWall href=""></Related></RelConnects>`

instead of

`<RelConnects><Related type="IfcWall" id="..."></RelConnects>`

The first is more consistent maybe with the rest of the code base.

aothms: Or another option is to treat the relationship similar to how we handle links to property sets. So not a separate root `connections`, but rather provide the links within the definitions of the elements. That would be a bit harder to realize though in case of RealizingElements

christt105: Ok, didn't know about the ```as_link``` parameter. For now, I will go for the first option.

Moult:
It looks as though the change was made, so maybe this can be merged?

aothms:
I agree, sorry for the delay

PR #2200, opened 2022-05-15
blenderbim: rollback_link_element() check items exist before removing

brunopostle:
I'm using the `IfcStore.execute_ifc_operator()` wrapper in blenderbim to manage undo history. Note that I only use the wrapper once in each operation, but within that I'm building a lot of IFC elements, and reusing types and materials etc.. if already present. 

This works fine, but I only seem to be able to undo the last item in the history list, if I then try to undo again, then I'm getting these errors, the model is left at the previous state:

```
  File "/usr/share/blender/3.1/scripts/addons/blenderbim/bim/handler.py", line 172, in undo_post
    IfcStore.undo()
  File "/usr/share/blender/3.1/scripts/addons/blenderbim/bim/ifc.py", line 416, in undo
    transaction["rollback"](transaction["data"])
  File "/usr/share/blender/3.1/scripts/addons/blenderbim/bim/ifc.py", line 304, in rollback_link_element
    del IfcStore.id_map[data["id"]]
KeyError: 1273
```

The attached change works around the error by only deleting stuff from IfcStore that is deletable, so it seems mostly harmless. The result is that multiple undo works as expected, but I have no idea why the transaction data refers to elements that don't exist in the first place.


Moult:
I'm a little hesitant to merge this as although defensive code has its time and place, here it could be masking another error:

> but I have no idea why the transaction data refers to elements that don't exist in the first place.

Can you point me to your operator code so perhaps I might spot something?

brunopostle:
This is with the Homemaker add-on (either from git, or the most recent release), to reproduce:

- Create two or three default cubes

- For each cube in turn, run the Object -> Homemaker operator to create buildings

- Press Ctrl-z, the most recent building is removed

- Press Ctrl-z again, a building is removed in the viewport, but the IFC rollback fails, generating the error

There is a lot of IFC data generated each time this operator runs. I could try hacking out Homemaker functionality to see where it breaks, but not for a few days

Moult:
Sorry still haven't had a chance to look at this but does this commit change anything? https://github.com/IfcOpenShell/IfcOpenShell/commit/b5e27e4cedf2461b6f5580493f701e2205158a45

brunopostle:
> Sorry still haven't had a chance to look at this but does this commit change anything? [b5e27e4](https://github.com/IfcOpenShell/IfcOpenShell/commit/b5e27e4cedf2461b6f5580493f701e2205158a45)

Nope, I see the same error with a fresh build from git. It is still something on my todo list to try and reproduce it with a smaller transaction

brunopostle:
I created a minimal Blender add-on that reproduces the problem:
[test_undo.py.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/9034483/test_undo.py.txt)

This adds a TEST_UNDO item to the Blender Object menu. This function:

1. Populates IfcStore.file with a simple Ifc Project if it is None
2. Creates an IfcSlab inside the _execute() wrapper
3. Runs ifc_importer.execute() to update the blender GUI (though in this case there is nothing of interest to update, in Homemaker this refreshes the entire project)

How to test:
1. Select the default cube and run Object -> TEST_UNDO several times
2. Ctrl-Z undo
3. Ctrl-Z undo <- This generates the error

It is undoing more than one item from the undo stack that causes the problem, so this works ok:
1. Select the default cube and run Object -> TEST_UNDO several times
2. Ctrl-Z undo
3. Object -> TEST_UNDO
4. Ctrl-Z undo
5. Object -> TEST_UNDO
6. Ctrl-Z undo
7. Object -> TEST_UNDO
8. Ctrl-Z undo
9. Ctrl-Z undo <- This generates the error

Commenting-out the ifc_importer.execute() call makes the error go away, putting it inside the _execute() call doesn't help.

Moult:
The reason your minimal example fails is because it is calling IfcStore.link_element twice with the same ID, but different object (indirectly, via your call to the IFC importer).

The first time it says hey make ID 6 be this Blender slab.

When you run it again, you import the slab twice, so it says hey make ID 6 (again) this _new_ Blender slab, without first unlinking whatever you might've had previously.

The commit fixes that by checking if you already had an obj with that ID linked to it.

However even though this commit fixes that, your minimum example as it stands still leaves the file in an invalid state- there are now multiple Blender slab objects pointing to the same IFC element. This is invalid because the BlenderBIM Add-on assumes a max 1:1 relationship between 1 Blender obj and 1 IFC entity. If you have multiple objects, then Blender won't know which one you want to be the source of truth.

Moult:
Is this PR still relevant with the latest fixes?

brunopostle:
I still see the same problem, (in Homemaker) I can undo the most recent item in the undo stack, but then undoing the previous item in the stack fails:

```
  File "/usr/share/blender/3.2/scripts/addons/blenderbim/bim/handler.py", line 173, in undo_post
    IfcStore.undo()
  File "/usr/share/blender/3.2/scripts/addons/blenderbim/bim/ifc.py", line 464, in undo
    transaction["rollback"](transaction["data"])
  File "/usr/share/blender/3.2/scripts/addons/blenderbim/bim/ifc.py", line 326, in rollback_link_element
    del IfcStore.id_map[data["id"]]
KeyError: 4916

brunopostle:
@Moult in your previous comment you say that the problem is the minimal script isn't removing blender objects before re-running the import. The script does have a delete_collection() function to remove blender objects, this is commented-out because I didn't think it was relevant. This function is how I clear the blender scene in Homemaker.

brunopostle:
@moult with the patch reapplied I can create four buildings in Homemaker, undo three of them, create two more, undo, save etc.. and the ifc file validates ok.

Attached a version of the test script which deletes the blender objects before the import, this fails without the patch:

[test_undo.py.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/9905503/test_undo.py.txt)

brunopostle:
Closing, this bug seems to now be fixed (probably after changes to the undo system a couple of weeks ago, ie. 0a84018).

Moult:
Ahh a huge thanks for checking back into this one. Undo bugs are tricky nuggets.

Issue #2212, opened 2022-05-24
BCFv3 "get_comments" error


AToradaIBIM:
There's an error when you try to get the comments of a topic using the BCF util V3.

The error that is throwing is "KeyError: Topics"  and this is because in the line 649 is trying to parse the element "Topics" instead of "Topic".
Based on the XSD there's no "Topics" element so this must be a missclick.

Here's the code:

```
def get_comments(self, guid):
        comments = {}
        data = self._read_xml(os.path.join(guid, "markup.bcf"), "markup.xsd")
        if "Comments" not in data["Topics"]:
            return comments
```

It's located on [this file](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/bcf/src/bcf/v3/bcfxml.py).

aothms:
Not an active user of this part of the codebase, but I think you are correct. Can you send in a Pull Request (you can just use the edit button in the github web interface for that these days).

Moult:
Thanks so much! Indeed a typo.

PR #2215, opened 2022-05-25
Fix BCFv3 "get_comments" error #2212

AToradaIBIM:
**Fix for Issue** #2212 

The method get_comments wasn't working as expected because on line 649, it was trying to access the element "Topics" instead of "Topic", based on the XSD "Topics" doesn't exist, this fix is supposed to solve that by changing this missclick.


aothms:
Many thanks!

PR #2223, opened 2022-06-02
Update README travis badge URL

EstebanDugueperoux2:
Update README travis badge URL since travis-ci.org indicates to use travis-ci.com now.
One question, on github repo the default branch is v0.7.0 while on travis the default is v0.6.0.
Is it a divergence?


aothms:
Yes, we kind of stopped using Travis more or less after they changed their pricing policy for open source projects. We now primarily use Github actions https://github.com/IfcOpenShell/IfcOpenShell/tree/v0.7.0/.github/workflows Not sure if they also have badges or anything.

PR #2228, opened 2022-06-08
Minimal building

disednet:
I added some modifications for cmake files and I suggest using these changes for building the minimal version of IfcOpenShell that will be used as a converter to ifc.


aothms:
Ok, sure, if you think it helps :) Thanks

aothms:
Changed the default to OFF though for backwards compatibility

PR #2234, opened 2022-06-10
catch missing instances one by one. do not throw away whole relation.

otsoa:
F.e. a IfcRelContainedInSpatialStructure where one of the elements in a storey is missing we can still keep the rest


Moult:
This seems pretty benign. Ping @aothms ?

PR #2235, opened 2022-06-10
convert boundingbox geometry

otsoa:
At least IfcSpaces can have only a bounding box as their geometry


Moult:
I vaguely remember implementing bounding box conversions... is this PR still needed? Ping @aothms

PR #2236, opened 2022-06-10
Topic/ignore contexts setting

otsoa:
This can be useful when we want to get all the geometry out of files where the context information is invalid / missing.
This can also be used if we need need f.e. bounding box geometry for spaces which might not be under 'model' representation context.


Moult:
Ping @aothms

theoryshaw:
Hi @otsoa 

PR #2241, opened 2022-06-20
Add more subdirs for blenderbim build trigger

Krande:
Added more subdirectories to the github actions workflow to trigger builds whenever a change occurs in relevant directories.

@Moult : If you can take a look at the directories I've added and see if there are any other directories we should monitor for changes in order to trigger new blenderbim builds?


Moult:
This looks great! Thanks!

PR #2250, opened 2022-06-29
Search for ifc class when assigning ifc data to an object

Gorgious56:
This has a simple goal : Allowing searching for ifc class when assigning ifc data to an object. Here's a preview :

Animation

It adds a button on the right of the ifc class enum field. I am slightly annoyed by the fact the fields are not perfectly aligned and I will look for a way to fix it. In the meantime I consider this functional.

If there are less than 10 entries, they will be displayed individually, otherwise it wil display a regular, filtered enum field.

If wish Enum Props would implement builtin fuzzy search but AFAIK they don't so I think this is a good compromise.


Moult:
Would it be possible for the search to also search for the predefined types? Since sometimes the object is listed in the predefined type. e.g. if you search for "BASIN".

PR #2253, opened 2022-06-30
Add prop search to fields with a lot of enumerated values

Gorgious56:
This is a WIP following https://github.com/IfcOpenShell/IfcOpenShell/pull/2250.

Thanks to @vulevukusej for the heads up in the OSarch chat about `prop_search`.

Currently it supports some fields in the IFC class metadata panel but I'd like to extend it to other ones. In order to prevent duplicate code doing the same thing all around the codebase I'd like to build something that is extendable by all the existing data constructs.

Animation

It currently feels hacky and there are some errors so it's not mergeable in this state but I'll be working on it.

Cheers

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/brick/prop.py
+++ b/src/blenderbim/blenderbim/bim/module/brick/prop.py
@@ -69,3 +69,9 @@ class BIMBrickProperties(PropertyGroup):
     libraries: EnumProperty(name="Libraries", items=get_libraries)
     namespace: EnumProperty(name="Namespace", items=get_namespaces)
     brick_equipment_class: EnumProperty(name="Brick Equipment Class", items=get_brick_equipment_classes)
+    
+    getter_enum = {
+        "libraries": get_libraries,
+        "namespace": get_namespaces,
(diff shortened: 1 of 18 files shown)


Moult:
When it's ready to merge feel free to go right ahead and merge :)

Gorgious56:
Alright there are still some enums that can benefit from this, but I'm not used to all of them and sometimes it doesn't make too much sense but anyone please feel free to post them here so I can add them up later.

The basic functionality should linearly scale when adding new props. 

1- In your `PropertyGroup` define a `getter_enum` dictionary linking the enum prop name with the items getter function

I would have liked to use the [EnumProperty.enum_items](https://docs.blender.org/api/current/bpy.types.EnumProperty.html#bpy.types.EnumProperty.enum_items) method so it is perfectly automated but for some reason it returns an empty list in this case... It's a limitation of enum items according to the devs https://developer.blender.org/T86803

2- Use `blenderbim.bim.ui.prop_with_search(layout, prop, prop_name, **kwargs (like text, icon, etc.))` to display a looking glass icon next to the enum type that will open a popup on click.

Cheers

PR #2290, opened 2022-07-18
Ifc class suggestion

Gorgious56:
Following a discussion with @Moult and @vulevukusej  on osarch chat and the documentation on the website https://blenderbim.org/search-ifc-class.html.

I added the "regular" suggestions on top of the page but thought it would be pretty harmless to also add the predefined types since some of them can be really helpful eg TOILETPAN > IfcSanitaryTerminal.

However it does create a dissonance since in that case it only changes the ifc class and NOT the predefined type which I assume is confusing to the end user. I tried to solve it elegantly but failed and made a mess of the already too big for my taste code managing this whole system. So be aware of it, even though you search&select a predefined type linked to an ifc class, it won't currently change the actual predefined type.

Animation

I won't merge it right now since I'd like to get a bit of feedback for

1- Usability standpoint
2- Is it a no-no to select a predefined type in the search box and it not updating afterwards ? As I write this I'm leaning more on the yes so I'd like to revisit it with a clear-er head later and solve it.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/root/prop.py
+++ b/src/blenderbim/blenderbim/bim/module/root/prop.py
@@ -106,13 +106,6 @@ class BIMRootProperties(PropertyGroup):
     ifc_predefined_type: EnumProperty(items=getIfcPredefinedTypes, name="Predefined Type", default=None)
     ifc_userdefined_type: StringProperty(name="Userdefined Type")
 
-    getter_enum = {
-        "contexts": get_contexts,
-        "ifc_product": get_ifc_products,
-        "ifc_class": get_ifc_classes,
-        "ifc_predefined_type": getIfcPredefinedTypes,
-    }
-
     getter_enum_suggestions = {
         "ifc_class": get_ifc_classes_suggestions,
     }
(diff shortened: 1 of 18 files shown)


Moult:
I think this is a great addition and according to goal number 10 in the [developer guide](https://blenderbim.org/docs/devs/hello_world.html#software-architecture): "Perfect is the enemy of the good. Half broken is better than completely broken." I'd say go ahead and commit :)

src/blenderbim/blenderbim/bim/module/root/prop.py
```diff
@@ -88,6 +89,24 @@ def get_ifc_classes(self, context):
...
+    if file:
+        for ifc_class in get_ifc_classes(None, None):
+            ifc_class = ifc_class[0]
+            declaration = IfcStore.get_schema().declaration_by_name(ifc_class)
+            for attribute in declaration.attributes():
+                if attribute.name() == "PredefinedType":
+                    for e in attribute.type_of_attribute().declared_type().enumeration_items():
+                        if e in ("NOTDEFINED", "USERDEFINED",):
+                            continue
+                        suggestions[ifc_class].append(e)
```
Moult, line 106: Can all this code be refactored into IfcClassData?

Gorgious56: Sure that makes sense. Thanks for the feedback ! :)

PR #2294, opened 2022-07-18
UI for IFC Selector tool

vulevukusej:
@Gorgious56 would love your feedback on this

Here's an example of the UI:
IfcSelector

still to do:

- [x] - make the UI portable so that it can be called from anywhere in the UI
- [ ] - figure out how to make loading properties for the chosen pset quicker, there's a noticable hiccup when doing this
- [ ] - undo all of the changes the black vs-code formatter made to all the code 🤦‍♂️
- [ ] - write tests (yes, yes, I know, I should've written them first! ;) )
- [x] - fix the regex in selector.py that is causing an error in the tests
- [ ] - add selection to IfcGroup
- [x] - add option to save selections for later use


Gorgious56:
Wow this looks very powerful ! I really like the feature, although it feels like you're battling Blender's UI system and you'd need a wide screen to be able to fit everything in it :) But you can't do much about that, until we have a reliable way to build our own custom editors...

I think you may have a bit of trouble testing it since it's primarily UI based but you can test the query results.

Also, when it's finished I think adding a way to store queries as presets to re-use them later would be a nice addition !

Regarding the pset hiccup I'm sure the bottleneck can be identifed easily. Unless it's already been identified and we can't do much about it...

Regarding the black formatter if using Vscode you can extend the line length limit by following either methods described on this page https://dev.to/adamlombard/vscode-setting-line-lengths-in-the-black-python-code-formatter-1g62

Cheers

src/blenderbim/blenderbim/bim/module/search/prop.py
```diff
@@ -16,10 +16,13 @@
 # You should have received a copy of the GNU General Public License
 # along with BlenderBIM Add-on.  If not, see <http://www.gnu.org/licenses/>.
 
-import bpy
```
Gorgious56, line 19: I would not remove this line, sometimes bpy is highlighted as not used directly by the IDE but it may be used in an expression or in a test in the script

src/blenderbim/blenderbim/bim/module/search/operator.py
```diff
@@ -431,5 +460,102 @@ def draw(self, context):
...
+        row.operator("bim.toggle_filter_selection",
+                     text="Select All").action = "SELECT"
+        row.operator("bim.toggle_filter_selection",
+                     text="Deselect All").action = "DESELECT"
+        
+        
+class Reset3dView(bpy.types.Operator):
+    """Filter model elements based on selection"""
+    bl_idname = "bim.reset_3d_view"
+    bl_label = "Reset 3D View"
```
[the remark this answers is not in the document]
vulevukusej: Agreed!

src/blenderbim/blenderbim/bim/module/search/prop.py
vulevukusej: Ok good to know!

vulevukusej:
@theoryshaw - I'm not sure I understand, do you mean if you select an object and then click on the eyedropper, the ifctype would load?

src/blenderbim/blenderbim/bim/module/search/operator.py
```diff
@@ -433,3 +450,236 @@ def draw(self, context):
         row = self.layout.row(align=True)
         row.operator("bim.toggle_filter_selection", text="Select All").action = "SELECT"
         row.operator("bim.toggle_filter_selection", text="Deselect All").action = "DESELECT"
+
+
+class UnhideAllElements(bpy.types.Operator):
+    """Filter model elements based on selection"""
+
+    bl_idname = "bim.reset_3d_view"
+    bl_label = "Reset 3D View"
```
Gorgious56, line 459: I think you forgot to remove these :)

Gorgious56:
Ooh nice, does that mean you've implemented saving specific queries for later ?

vulevukusej:
> Ooh nice, does that mean you've implemented saving specific queries for later ?

@Gorgious56 yes i have ;)

vulevukusej:
So I think this is looking pretty decent now, I might commit the changes and then keep chipping away at optimizations and fringe test cases ;).

theoryshaw:
> @theoryshaw - I'm not sure I understand, do you mean if you select an object and then click on the eyedropper, the ifctype would load?

yes, conceptually... 
would order it 1)click on eyedropper 2)click on object and then value of that object is loaded. 

could work with other entities and values, not just ifctypes. 

This is especially helpful if you have hundreds of 'things'... and you don't want to look up the name of thing, but just want select/hide everything like 'that' thing, in the viewer.

Issue #2291, opened 2022-07-18
Selector - handling situations where attribute is None


vulevukusej:
In the following example, am I wrong to expect that all objects are matched?:

.IfcWall[PredefinedType != "non-existent predefined type"]'

Perhaps the following code can be updated to check whether comparison == "not"?
https://github.com/IfcOpenShell/IfcOpenShell/blob/d71ebb8d2519051a59122b49b18a8d505578fc18/src/ifcopenshell-python/ifcopenshell/util/selector.py#L192-L194

Moult:
Great suggestion! More than happy to see this implemented, but please write tests for it to help document the behaviour.

PR #2298, opened 2022-07-20
Selector: added tests to check non-standard characters and missing values

vulevukusej:
Since it's my first test I figure I'd better create a pull request and gather some feedback ;)

Also added a potential solution to #2291


Moult:
I'm happy to merge this but it seems this PR has started to grow with other code. Feel free to commit directly :)

PR #2312, opened 2022-07-25
Deprecated fixes

skinkie:
One annoying complition error, and some deprecation warnings.

### DIFF
--- a/src/ifcgeom/IfcGeom.cpp
+++ b/src/ifcgeom/IfcGeom.cpp
@@ -3458,6 +3458,7 @@ bool IfcGeom::Kernel::triangulate_wire(const std::vector<TopoDS_Wire>& wires, To
 	Handle_Poly_Triangulation tri = BRep_Tool::Triangulation(face, loc);
 
 	if (!tri.IsNull()) {
+
 #if OCC_VERSION_HEX < 0x70600
 		const Poly_Array1OfTriangle& triangles = tri->Triangles();
 		for (int i = 1; i <= triangles.Length(); ++i) {
--- a/src/ifcgeom/IfcGeom.cpp
+++ b/src/ifcgeom/IfcGeom.cpp
@@ -3458,11 +3458,18 @@ bool IfcGeom::Kernel::triangulate_wire(const std::vector<TopoDS_Wire>& wires, To
 	Handle_Poly_Triangulation tri = BRep_Tool::Triangulation(face, loc);
 
 	if (!tri.IsNull()) {
-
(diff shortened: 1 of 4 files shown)


aothms:
Thanks.

HCompCurve is still used on OCCT versions under 7.6, which is still the default on e.g most linux distros. So for the code to remain compatible the deleted include also needs to be conditional wrt OCCT version.

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcgeom/IfcTrimmedCurve.cpp#L58

See CI Failure:

https://github.com/IfcOpenShell/IfcOpenShell/runs/7495267968?check_suite_focus=true#step:5:660

skinkie:
@aothms how are these things checked within QA?

aothms:
We just have some CI scripts in github actions. Currently they compile against 7.5 as that's what the distros provide and compiling occt in the github action would be prohibitive in terms of runtime https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/.github/workflows/ci.yml#L44

skinkie:
This still gives me deprecation warnings, I am at 7.6.2. Please advise.

skinkie:
Uncovered a bug in the existing code while fixing this :-)

theoryshaw:
Hi @skinkie 

CyrilWaechter:
From my understanding this PR has been superseded. Thanks for your work and feel free to reopen if I missed something.
Additional AI analysis for info:
```
Thanks for this. Against the current active branch (v0.9.0) this change is superseded and can be closed:

- The equivalent fix already landed in e104e7486 ("OCCT: avoid using deprecated Triangles method"), which switched tri->Triangles() to tri->NbTriangles() + tri->Triangle(i) in opencascade_conversion_result.cpp and serialization/schema/serialization.cpp. The successor of IfcGeom.cpp::triangulate_wire (kernels/opencascade/wire_utils.cpp:221) uses the same non-deprecated API.
- All four files touched here no longer exist at these paths on v0.9.0 (the tree was restructured), and the PR is currently CONFLICTING/DIRTY. No Poly_Array1OfTriangle remains on v0.9.0.
- The HCompCurve concern raised earlier is moot: mapping/IfcTrimmedCurve.cpp on v0.9.0 no longer references it.

Suggest closing in favour of the landed upstream change, unless there's a specific OCCT < 7.6 path you still want covered — note v0.9.0 already uses OCC_VERSION_HEX guards elsewhere and requires OpenCASCADE ≥ 7.5.1.
```

PR #2315, opened 2022-07-25
Addition of Construction Type Browser

cvillagrasa:
This PR adds a panel operator which allows to preview the construction type of assets loaded from IFC libraries.

imagen

Further discussion here: https://community.osarch.org/discussion/1031/blenderbim-ifc-type-thumbnails


cvillagrasa:
Let me write a quick summary with some specific points in this PR for an easier review:

1. This PR makes use of Blender assets in order to make the preview possible. At my first iterations, upon a certain preview was needed within the UI, I was trying to quickly add an object to the scene, make an asset out of it and delete it, finally showing the preview with the `template_icon` _UILayout_ method. @Moult [suggested to use the Types collection](https://twitter.com/BlenderBIM/status/1539173658778046465) under IfcProject instead, since there are already objects created. However, I also found that depending on the construction type, those objects could be empty. Hence, I came back to the generate and delete strategy, which is what is used in this PR. Assets are memoized for performance, but maybe one first easy improvement after this working PR is to add a check of whether the object under Types is not an empty, when assetizing a certain construction type.
2. I found that the addition of _IfcCovering(s)_ was somewhat unreliable, with slab code being executed with `LayerSetDirection` properties having  its  `prop.NominalValue.wrappedValue` equal to `"AXIS2"`, and the same thing for wall code with `"AXIS3"`. Solved it temporarily with [this strange commit](https://github.com/IfcOpenShell/IfcOpenShell/pull/2315/commits/662cd6232c46217c4900ff3aebf032ecd2703f28), but certainly further revision of the creation of geometry from the model module is needed.
3. After some discussion of how to call existing `ifc_class` and `relating_type` from a UX perspective, as one would speak of Category/Family in Revit, Dion [suggested _Construction Class_ and _Construction Type_](https://twitter.com/BlenderBIM/status/1546443847114592257). Along those lines, I've changed many variable names from `ifc_class` to `constr_class` and from `relating_type` to `constr_type_id`, since the latter was usually saved as an id.
4. Added the utility function `layout_with_margins` [to the helper file](https://github.com/IfcOpenShell/IfcOpenShell/pull/2315/files#diff-c838550eb5ddc153a8b67777ba79248ed015b990e3e358cde96b3f712386ca41). We should possibly find a better location, as discussed with @Gorgious56 on the OSArch Live Chat.
5. If adding and deleting construction types on an empty project, it is relatively easy to unintentionally select and delete an _IfcSpatialElement_ in the process. Added [this quick fix](https://github.com/IfcOpenShell/IfcOpenShell/pull/2315/commits/319f63c86179ec5dc2ee68cfdcf88b2b39ee6461) to avoid an error in that particular scenario.
6. I set it up so that the _Browse Construction Types_ button is only available from the tool header region, as in the screenshot on the first comment. Both sidebar and workspace settings just show the selected combination of _construction class/type_.
7. There is a minimal _Help Panel_ added, both in the upper left corner of the tool header region (also available when there is no _IfcProjectLibrary_ added and the _Browse Construction Types_ button is not available), and within the upper right corner of the browser operator panel itself, once loaded. Some work will be needed in the future to integrate its contents with the _BlenderBIM_ documentation. Potentially extensible to any other part in the bb add-on.
8. The browser includes two modes, depending on whether the upper left boolean _Preview All Construction Types_ is toggled or not. This is a result of two different lines of feedback: those wanting several previews together (all of the construction types for one construction class) and those wanting only one with a cascading dropdown (for performance or practical reasons). With the toggle, everyone should be happy :)
9. Three tests [have been added](https://github.com/IfcOpenShell/IfcOpenShell/pull/2315/files#diff-b49e53ab515808f5615f4b23895631b972d7731d03199f9b5013e3148a6d0898), regarding the preview of assets of both display modes available in the browser, as well as the addition of the selected construction type. I have tested them directly with `pytest`, and not with `make test-bim`, because for some reason I can't get the latter to work in my environment. It is worth noting that the preview tests don't get to assure that a successful preview has been generated. This is due to the nature of the `asset_generate_preview()` function, which needs the OpenGL renderer from the UI and cannot work in headless mode, as done with tests (thanks @Gorgious56 for the hint). I left a comment on the Blender Chat regarding this issue, but so far I haven't received any idea. I guess that if a certain `preview` with a certain `icon_id` equal to 0 is generated in headless mode, it should be ok, even if in UI mode, the actual `icon_id` will be a relevant number, and not 0.
10. Lastly, I see that a couple of minor unintended commits have been included in the PR ([this](https://github.com/IfcOpenShell/IfcOpenShell/pull/2315/commits/89023df4e588322d61e5f5f0bb08dcab782707f0) and [this](https://github.com/IfcOpenShell/IfcOpenShell/pull/2315/files#diff-81880bc509277ff79bc75f608d25b1afd8ddea448f7bf6042f17b25ad7c27743)).

cvillagrasa:
Point 10 is solved. It seems that I got distracted and the first time I tried to merge the latest commits to my branch, I did it from the `master` branch, instead of `v0.7.0`... anyway, I've reverted it in https://github.com/IfcOpenShell/IfcOpenShell/pull/2315/commits/58150a296d43f892a2acb29f90cbab0833432518

Moult:
"This branch cannot be rebased due to conflicts" <-- can you resolve this and/or create a new PR and then perhaps we can go through a pair programming session together to go through your 10 points in your summary? I think we can solve some of these together :)

Moult:
Sorry can you also change the variables you renamed to constr/construction back to the original IFC names? I strongly believe that the code should always use the correct, exact IFC terminology. If you want to use the term "Construction" in the UI to help users, sure, but the internal code should remain true to the schema.

There also seem to be a few suspicious naming changes like add_type_instance has been changed to add_constr_type which to me reads like two different concepts (e.g. the first name is clear you are adding an instance of the type, whereas the second seems you are adding the type itself).

Would it be possible to make these name changes (i know I know pedantic but I really think it's important) prior to a more detailed code review session together?

cvillagrasa:
Ok @Moult ! thanks for the feedback, I've reverted the variable naming back. Happy to go through the pair programming session when you have availability.

cvillagrasa:
Just for reference, I list some of the points we have treated in the screenshare:

1. Remove the SELECT button
2. Rethink the way the width of the panel is assigned. Less logic is better.
3. Remove title labels from the construction type browser.
4. Separate UI and product code (I believe this is already achieved by your commit).
5. Remove UNDO from operators that don't need it.
7. Don't close panels with another operator. Remove the hacky function.

Thanks @Moult !

Issue #1812, opened 2021-10-20
Proposed improvement to remove_deep


Moult:
Here is a simple subgraph shown three times. The first with letters for explanation purposes, the second using the existing remove_deep functionality, and the third using a newly proposed replacement for remove_deep.

Given a subgraph starting with element `A`, where `A` has _no inverses_ (otherwise, unbatching fails in my proposal), imagine we want to perform a deep removal. The existing remove_deep algorithm traverses the subgraph backwards (`F`, then `E`, etc) each time checking there are any inverses outside the subgraph. If all inverses are part of the subgraph, it can be deleted. Deleted elements are marked as `X`.

This creates an issue, such as where element `C` only has semantic meaning if `D` exists. For example, if `C` is an extruded area solid, and `D` is a profile def. Therefore, if `C` is referenced elsewhere, it is preserved, but it doesn't help much because `D` is deleted.

I propose a different approach, to instead remove_deep traversing the subgraph forwards instead of backwards. If the element has inverses outside the subgraph, the entire branch is no longer considered for deletion. This should also in theory lead to faster remove deeps, as well as preserving the assumption that an element only makes sense together with all of its subelements.

Am I making a mistake? Is this logic sound?

```
A──►B──►C──►D──►E──►F
        ▲       ▲
    G───┘   H───┘

X──►X──►0──►X──►0──►X
        ▲       ▲
    0───┘   0───┘

X──►X──►0──►0──►0──►0
        ▲       ▲
    0───┘   0───┘
```

Here is sample code for a make-believe scenario of the graph above. Currently, batch and unbatch throws an error but I don't fully understand the rules of when I can and can't batch, so hopefully that can be sorted out later if the premise is sound.

```python
import ifcopenshell
ifc = ifcopenshell.file()

A = ifc.createIfcProductDefinitionShape()
B = ifc.createIfcShapeRepresentation()
C = ifc.createIfcExtrudedAreaSolid()
D = ifc.createIfcArbitraryClosedProfileDef()
E = ifc.createIfcIndexedPolyCurve()
F = ifc.createIfcCartesianPointList3D()

G = ifc.createIfcShapeRepresentation()
H = ifc.createIfcArbitraryClosedProfileDef()

A.Representations = [B]
B.Items = [C]
C.SweptArea = D
D.OuterCurve = E
E.Points = F

G.Items = [C]
H.OuterCurve = E

print('A', A)
print('B', B)
print('C', C)
print('D', D)
[...]

PR #2316, opened 2022-07-26
Transfer psets and style representations when appending a material

### LINKED ISSUE #2311: Blenderbim: appending Materials from Project Libraries doesn't transfer psets or styles
I tried to fix this by changing api/project/append_asset.py `append_material()`:

```
#return self.file.add(self.settings["element"])
self.whitelisted_inverse_attributes = {"IfcMaterial": ["HasProperties", "HasRepresentation"]}
return self.add_element(self.settings["element"])
```

This does copy psets and styles, but it also creates duplicate Geometric Representation Contexts for every Styled Item, and I can't figure-out how to avoid this :(

brunopostle:
This seems to fix #2311, if it looks ok I'll see about writing tests

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py
@@ -51,7 +51,25 @@ class Usecase:
     def append_material(self):
         if [e for e in self.file.by_type("IfcMaterial") if e.Name == self.settings["element"].Name]:
             return
-        return self.file.add(self.settings["element"])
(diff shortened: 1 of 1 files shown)


Moult:
BTW can you test this using remove_deep2() instead of remove_deep? In general remove_deep2() should always be used, but I was just afraid to replace existing uses of it without test coverage: https://github.com/IfcOpenShell/IfcOpenShell/issues/1812

brunopostle:
remove_deep2() seems to work too. It may take me a while to figure out how to write the tests.

PR #2330, opened 2022-08-02
add logic to build blenderbim-nightly chocolatey package

fbpyr:
@Moult  as discussed adding the logic to build the blenderbim-nightly chocolatey package. 
will PR the github action changes separately: 
most likely as a scheduled action before midnight, so that it builds only once per day if there 
are published releases for that day, in order to cause minimum traffic over at 
🍫  [chocolatey](https://community.chocolatey.org/) 🙂


fbpyr:
@Moult  apparently two tests fail, but looking at their logs they do not seem to be related 
to the pushed code, which is fully contained in choco directory. 🤔

PR #2355, opened 2022-08-19
[BCF] New BCF XML handler with better performances

### LINKED ISSUE #2321: BCF - proposal to manipulate DOM instead of file I/O per operation to improve performance
I there a reason why the `BcfXml` class directly writes to/reads from disk instead of having an in-memory parser and writer?

In one of my tests I had to create many (~46k) topics - not something that we will (hopefully) encounter in production, but it still... - and I have very long running times (748seconds),  60% of it spent `open` ing files for writing the files.

I realized that the various methods of `BcfXml` are completely reliant on I/O operations, and I wonder why is it the case...

Wouldn't it be better if all the operations were made on the `minidom.Document` objects (or even better, using something like [`xsdata`](https://xsdata.readthedocs.io/en/latest/) to generate the dataclasses from the BCF XSD and using them), then dump the resulting XMLs into StringIO objects and feed them to the `ZipFile` object?

I will work on a I/O-less version the next week for my project; if I'll got it right, I'll be happy to provide a PR, if it's of any interest!

sanzoghenzo:
BREAKING CHANGE: incompatible methods and objects with previous version.

closes #2321


sanzoghenzo:
This is by no means a finished product, but it's in a good state to show you what I accomplished so far.

I only created a quick roundtrip test (serialize and deserialize) without complex data and files in it.

I used tox to be able to run tests on multiple python versions (but I actually tested it only on 3.10).
I also added isort to the `make qa` target to keep imports sorted, and added a `make typecheck` target that runs mypy.

I still have to profile this, since I changed a good amount of code (to enable on-demand loading) and I don't have my test files (I left them at work...)

Moult:
This is awesome! Let me know when you think it is ready to merge.

Since this is such a large change, you'd know best how you'd like it to be reviewed. At a glance a lot are improving the type hinting, and the lack of tests (I see two?) means I'm not entirely sure how drastic changes are from a user perspective :) Let me know how you'd like to proceed or we can just merge first ask questions later :)

The new QA stuff is excellent - a huge thank you! XSD generated data classes are also very smart.

sanzoghenzo:
I updated the readme to give a sense of what's changed; I used properties where I could to get rid of the `get_*` methods.

One big change I did is to store the topics (wrapped inside `TopicHandler` classes) into a list instead of a dictionary with the guids as keys.
If you think this is too much of a breaking change, I can revert it back to a dictionary; note that the value of that dictionary would now be the wrapper class and not the topic itself.
This wrapper class handles the access to the comments and viewpoints of that topic, so there's no need to get the guid of the topic and use it in the (deprecated) `BcfXml.get_*(guid)` methods.

the `BcfXml.edit_*()` methods are deprecated and they don't do anything. One needs to only `save()` when ready to commit the changes.

Moult:
That all sounds good. If the topics are stored as a list, how would you quickly retrieve a topic via a guid (e.g. via a related topic, or if you just wanted to fetch one)?

sanzoghenzo:
> That all sounds good. If the topics are stored as a list, how would you quickly retrieve a topic via a guid (e.g. via a related topic, or if you just wanted to fetch one)?

For now, I left the `get_topic(guid)` method that does a search in the list using `next`; do you think the access by guid is done frequently? in that case, it sure is better to revert to a dictionary!

Moult:
It's probably not a huge deal since BCF XML is unlikely to hold lots of topics, until I heard about your monster dataset :)

That said if we treat BCF like the AEC equivalent of a bugtracker, I think fetching by ID is important, for example in BCF API that's what they do: https://github.com/BuildingSMART/BCF-API#323-get-topic-service

There was also probably some intention previously to try and mirror the function signatures between BCF XML and BCF API, but I'm also not too fussed either way.

sanzoghenzo:
Hi there,
What is the minimal python version that I should support?
I generated the dataclasses from the xsd files using the `slots`  and `kw_only` parameters, that are available only starting from python version 3.10.
Blender uses python 3.10 already, so it's not a problem;
given that this is a breaking change, I would stick to 3.10 only to make my life easier (and to have the memory improvement offered by `slots`), but I'm open to suggestions here.

As per BCF-API: I think it would be better to create a new package autogenerated by [openapi-python-client](https://github.com/openapi-generators/openapi-python-client) instead of reinventing the wheel - but we'll discuss it later 😉

Moult:
@sanzoghenzo I'm personally OK with only supporting Python 3.10, this is alpha software so I don't have a strong attachment to older versions.

aothms:
Following along with great interest.

+1 on 3.10 only if it makes lives easier. I think the core features of IfcOpenShell are expected to work on all non-eol python versions, but for bcf specifically I'm fine with 3.10 only, especially since the aim of this PR is better performance, so the dataclasses + slots seem like a good approach.

Moult:
Bump :)

Moult:
Out of curiosity, what would be the performance if you download a BCF of say 5,000 issues (e.g. an export from your CDE to migrate to another platform), including all image attachments, and then only want to edit one, and save it out? Will the saving out procedure be very slow, or is it smart enough to just change the XML of the one that was changed, and all other assets (e.g. images) are also untouched?

sanzoghenzo:
Hi, sorry for the radio silence, I haven't got the chance to continue the development yet.

I mainly focused on the creation of a new bcf file, as this was the feature needed by my team; 
I started the on-demand loading part, but as of now it is only at the "top level" and pretty useless (as soon as you access the topics, they are all loaded), need to pass the zipfile to the TopicHandler class and handle it.
Right now the save overwrites the zip file (and I still have to test if it works, since now the file stays open - thanks for the question, now I know what unit tests I need to implement 😉)  so it will take a long time to save only one change. Do you know any smart way to keep track of the changes?

Moult:
Maybe the dumb approach of leaving it up to the user is fine rather than keeping is_dirty flags and so on? E.g. the user may either choose to explicitly save a single topic/viewpoint/etc or save_all.

sanzoghenzo:
> Maybe the dumb approach of leaving it up to the user is fine rather than keeping is_dirty flags and so on? E.g. the user may either choose to explicitly save a single topic/viewpoint/etc or save_all.

OK, I can do that - there's already a `save` method in each wrapper class I implemented (called by BcfXml and the parent classes); right now it has the zip file as parameter, but this will change with the other adjustments I need to take care of.

The only thing that is not handled by this is when you delete an object, for that I need to keep track of it somehow...

Moult:
I reckon when this library was first created I was very short sighted and only thought about the case where a large collection of BCFs already existed and the user was editing BCFs one by one via an interface. Hence the I/O heavy operations. Now I realise that both scenarios: individual editing from a large data set, and mass creation of a large dataset are both equally important usecases.

sanzoghenzo:
I took a look at this, and unfortunately the current implementation of ZipFile is not able to delete or overwrite a single file, so one still needs to recreate the entire archive.
As of now, this PR does that, but adds the overhead of a round-trip parsing and serialization of the XML files. I'll see if I can find a way to short-circuit this without adding too much complexity.

Moult:
@sanzoghenzo why don't you just unzip like the current implementation? Speed?

sanzoghenzo:
Well, the point of my PR was to avoid as much I/O as possible, so I went all in with the "file in memory" route.
Also, I saw that most of the cloud drive (where my colleagues tend to/ have to work on), add significant overhead to the I/O operations, so this should help to reduce the wait time (at the expense of more memory consumed, of course)

This is where I am with the latest commit:

- the zip file is opened in read-only mode and closed on class disposal, on context manager exit and on save (there's the `keep_open` option to reload the saved file and continue the editing)
- files are read and parsed in-memory, and it will be done on-demand (for now topics and other elements are loaded all at once while accessing the list, but I swear I'll fix this somehow)
- the save method serializes all the objects to in-memory XML files, zips them to an in-memory zip file and then writes it to disk - this will call the reading and parsing of objects that were not already loaded, instead of directly copying the file objects.

Moult:
Bump :)

sanzoghenzo:
Sorry for the delay... I'm struggling at implementing the "smart save" feature while keeping the code simple.
I'll try once more this sunday, I hope to find an elegant way to solve this!

sanzoghenzo:
maybe I got the on-demand loading and short-circuit saving all wrong: I started with the XML files, and they're pretty much done, but I should have started with the "heavier" part, the viewpoints snapshots and other files/bytes object to realize that I need to always load the XMLs that list those resources, or else I miss what I need to load and save :sweat_smile: 

hopefully on the next days I can get a hold on some BCF files for some more thorough testing!

sanzoghenzo:
Hi guys,
I'm making some progress here!
I tested the code against a real-world example (BCF v2.1; I anonymized it and added to the tests folder to be able to use it in the unit tests).
Files are loaded and saved correctly, at least topics/markups, viewports and and snapshots.

I think this is ready for review!
Well... I don't have examples with BIM snippets and documents in them, I can use some help from you!

Moult:
> Well... I don't have examples with BIM snippets and documents in them, I can use some help from you!

Maybe https://github.com/buildingSMART/BCF-XML/tree/release_3_0/Test%20Cases/v2.1/Markup/MaximumInformation ?

sanzoghenzo:
> > Well... I don't have examples with BIM snippets and documents in them, I can use some help from you!
> 
> Maybe https://github.com/buildingSMART/BCF-XML/tree/release_3_0/Test%20Cases/v2.1/Markup/MaximumInformation ?

Oh boy, I totally missed that repo! I didn't find anything on prevoius searches 😅 I'll use those example to extend the test cases ASAP

Moult:
BTW there were some errors in those BCF test cases that I had fixed, but I can't remember which branch they fixed it on (there's one branch for each BCF version). So if you run into funny behaviour, check the branches :) There was also one point in time where the unzipped version had more updated data than the zipped but I hope they've fixed that now.

sanzoghenzo:
Tests for v2.1 are up and running!

Too bad MaximumInformation example is not available for v3.

Moult:
Fantastic! Do you think it's ready for merge? I also see there is a conflict, is it possible to fix that conflict?

If it is ready for merge, I'd advise waiting until this weekend, since there is a release this weekend, and if we merged now, then the BCF GUI in the BlenderBIM Add-on would break. I'd propose delaying the merge until this weekend, then I/you/someone can rebuild the Blender BCF GUI to use the new library (I haven't looked in detail, so I've got no idea if this is just a drop-in replacement or how drastic it'd be to rewire).

sanzoghenzo:
There are breaking changes, so it's better to pospone the merge.
Also the new features of bcf v3 are not tested, I'll see if there are some other sample file that can help with this.

I also would like to rebase and squash everything in a single commit to avoid polluting the git log, and with that resolve the conflicts.

Moult:
OK, the release is now over so we can discuss merging this. Once we merge, we instantly break the BCF functionality. In an ideal world, we can rebuild the BCF functionality to minimise the feature breakage downtime.


But the world is not ideal, so I'll merge it tomorrow so at least I can make sure the build scripts still work and module initialisation works, and then deal with the rest of the issues whenever we have time :)

Issue #2332, opened 2022-08-03
IFCEXTRUDEDAREASOLID wrong Position


ninichki:
Hello,

In this case some IfcCovering has a wrong place. Here with ID: 3J2XeDAq14tANzHxSwtyxf
![image](https://user-images.githubusercontent.com/1086380/182605225-0fff1a7b-2779-4511-99d1-c8fde39d5a8c.png)

Solibri is showing it on the right position.

It's a model from Allplan.

I tested it with the latest v0.7.0. And OCCT 7.6.3

kind regards,

Christian
[Domus_ohne Sockelleiste.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/9251166/Domus_ohne.Sockelleiste.zip)

Moult:
In the attached IFC, your geometry is invalid because the extruded area is set to 0. If however I set that to 0.1 arbitrarily, then it seems to work as expected, tested with the latest IfcOpenBot release on the BlenderBIM Add-on:

![2022-08-22-163922_891x553_scrot](https://user-images.githubusercontent.com/88302/185855366-92f6ba21-bfa4-470c-be85-a8de7f04f0f9.png)

Can you try and recreate your issue with the latest version from https://github.com/IfcOpenBot/IfcOpenShell/commit/dc67287d4b083cabae39ee4f92fa93b6f74cae11#commitcomment-78799114 ? If not it seems to be fixed and we can close this.

Moult:
Closing as fixed :)

PR #2357, opened 2022-08-22
Set Position of extrusions with height = 0

ninichki:
Siehe Issue #2332


Moult:
Isn't this just invalid geometry? The data type is a positive length measure, so I'm not entirely sure why it should be supported.

It's @aothms ' decision though :)

ninichki:
Yes, Allplan from Nemetscheck does it this way... :-/

aothms:
It is indeed invalid, but at the same time the PR is a clear improvement over the current state of the code. For some reason negative or zero length extrusions are fairly common. Great job on figuring things out while I was travelling.

PR #2377, opened 2022-09-03
From append methode to use numpy and ifcpatch bug

Martin15135215:


Moult:
Just holding off merging this since it also has the ifcpatch fix, but I'd like to replicate this first with a sample model and see if there is a better fix.

Issue #2375, opened 2022-09-03
[IfcPatch] Bug in recipe RecycleNonrottedElements


Martin15135215:
https://github.com/IfcOpenShell/IfcOpenShell/blob/af4cb22756ffd44ebb6eeaf5d6a62d22c75973fb/src/ifcpatch/ifcpatch/recipes/RecycleNonRootedElements.py#L34

Change this `if element.is_a('IfcRoot'):` to this `if element.is_a('IfcRoot') or element.is_a('IfcPropertySingleValue'):`
#2377 

When I use the recipe, suddenly all my objects had the same property. But with the new code this did not happen to me.

Moult:
@Martin15135215 can you give a sample file for me to replicate this? In theory IfcPropertySingleValue should not be excluded here.

Martin15135215:
Here is the jupyter notebook to replicate the problem: https://github.com/Martin15135215/git_ifc_test/blob/main/psets_test.ipynb

the code from class ifc is from [here ](https://github.com/IfcOpenShell/IfcOpenShell/blob/ffa3aae132301cd32b216b315be1f9f208ef1cb7/src/blenderbim/scripts/obj2ifc-meshlab.py#L90)

this script creates test_1.ifc and test_2.ifc and there you should see the anomaly

test_1.ifc:
![image](https://user-images.githubusercontent.com/110968398/188306348-0678cc72-9b1c-4c6d-8851-39a4804cef71.png)

test_2.ifc:
![image](https://user-images.githubusercontent.com/110968398/188306366-58e258b0-09b9-49cd-b9e4-44ebbc743d73.png)

Can you replicate this problem with the jupyter notebook?

Moult:
Excellent catch! I had no idea that IfcLabel('foo') and IfcLabel('bar') would produce the same hash. It's been fixed now.

aothms:
> I had no idea that IfcLabel('foo') and IfcLabel('bar') would produce the same hash

Can you explain?

Moult:
```
>>> import ifcopenshell
>>> f = ifcopenshell.file()
>>> x = f.createIfcPropertySingleValue(NominalValue=f.createIfcLabel('foo'))
>>> y = f.createIfcPropertySingleValue(NominalValue=f.createIfcLabel('foo'))
>>> z = f.createIfcPropertySingleValue(NominalValue=f.createIfcLabel('bar'))
>>> hash(tuple(x))
5892372526342220303

[...]

PR #2378, opened 2022-09-04
Compute meaningful hash and eq on simple type instances

aothms:
See #2375

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/entity_instance.py
+++ b/src/ifcopenshell-python/ifcopenshell/entity_instance.py
@@ -234,10 +234,23 @@ class entity_instance(object):
     def __eq__(self, other):
         if not isinstance(self, type(other)):
             return False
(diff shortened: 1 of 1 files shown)


aothms:
New behaviour

~~~
import ifcopenshell
f = ifcopenshell.file()
x = f.createIfcPropertySingleValue(NominalValue=f.createIfcLabel('foo'))
y = f.createIfcPropertySingleValue(NominalValue=f.createIfcLabel('foo'))
z = f.createIfcPropertySingleValue(NominalValue=f.createIfcLabel('bar'))
# For hash of instance *values*, x and y are equal
print(*map(hash, map(tuple, (x,y,z))))
# For hash of instances, all three are different
print(*map(hash, (x,y,z)))
~~~

Moult:
Looks great and this is a much better solution than my downstream attempt! Go ahead and merge and I'll revert my commit to the patch.

PR #2380, opened 2022-09-04
Don't lock grids upon import and creation

theoryshaw:
Propose to not lock grids upon import or creation, as often times grids move a lot during the design phase.


Moult:
Someone on OSArch requested this feature. I don't know who is right or wrong :)

Note that it's easy to lock/unlock in bulk, so which should be the default? Right now less people use it for design, and more use it for viewing. Also note that it depends on the type of design, like (most) resi will frequently have moving grids, whereas retail, commercial, or health projects won't.

For those reasons (majority of users and majority of design typologies) I'm leaning towards rejecting this PR.

theoryshaw:
No worries.

PR #2389, opened 2022-09-07
Allowing float values to work in IFC smart tags

theoryshaw:


theoryshaw:
see new and improved PR: https://github.com/IfcOpenShell/IfcOpenShell/pull/2431

PR #2405, opened 2022-09-12
Add additional quantity calculations to QtoCalculator

vulevukusej:
Goal of this PR is to augment the QtoCalculator class with functions that mimic the qto abilities of bim calculation software.

For each new function there should be clear docstrings as well as a conceptual sketch of what the formula is calculating (For reference see documents below).  Please give feedback on the maths and code as I update this draft PR.  

Formulas covered so far:

- [x] get_linear_length - already covered by @Moult
- [x] get_volume - already covered by @Moult 
- [x] get_length - already covered by @Moult 
- [x] get_width - already covered by @Moult 
- [x] get_height - already covered by @Moult 
- [x] get_perimeter - already covered by @Moult 
- [x] get_net_footprint_area - already covered by @Moult 
- [x] get_net_top_footprint_area - adaptation of @Moults code by @vulevukusej 
- [x] get_side_area - already covered by @Moult 
- [x] get_area - already covered by @Moult , will be renamed to get_total_surface_area
- [x] get_projected_area - (Includes toggle for choosing net or gross) @vulevukusej
- [x] get_net_side_area - @vulevukusej and @Gorgious56 
- [x] get_gross_side_area - @vulevukusej 
- [x] get_net_top_area - @vulevukusej 
- [x] get_gross_top_area - includes a method for calculating opening methods @vulevukusej 
- [x] GrossFootPrintArea - achieved using get_gross_projected_area @vulevukusej 
- [x] get_side_opening_area - includes filter for opening size @vulevukusej 
- [x] get_opening_type - determines whether opening is a "RECESS" or "OPENING", since sometimes this is undefined. @vulevukusej 
- [x] get_OOB_object - returns the oriented bounding box  @vulevukusej 
- [x] get_AABB_object - returns the axis-aligned bounding box @vulevukusej 
- [x] get_bisected_obj - returns the object bisected by two planes @vulevukusej 
- [x] get_total_contact_area - Returns the total contact area of the object with other objects. @vulevukusej 
- [x] get_touching_objects - Returns a list of objects that are touching the object. @vulevukusej 
- [x] get_contact_area : Returns the contact area between two objects. @vulevukusej 
- [x] get_intersection_between_polygons: Returns the intersection between two polygons. @vulevukusej 
- [x] Width/Height at given height/width - achievable through get_bisected_obj
- [x] Area between two intersection planes - achievable through get_bisected_obj
- [x] Volume between two intersection planes- achievable through get_bisected_obj
- [x] Cross-Sectional Area at intersection - - achievable through get_bisected_obj
- [x] Perimeter Base/IntersectionPlane/Top  - achievable through get_bisected_obj

Considerations to keep in mind:
- Modifiers for formulas, such that users can fine-tune the calculation according to their needs or that of a standard (SMM7, VOB, ÖNORM etc.)
- I think it would be nice if the script could generate a calculation output that allows the user to inspect the calculation. This could be an output using matplotlib for example.
- In cases where no IfcOpeningElements are defined, perhaps we need an algorithm for defining opening meshes such that norm calculations can be performed correctly?  I know iTWO users use this feature regularly @TZwielehner 

The following files are from the iTWO Documentation - In particular, "Element Planning - Geometric Parameters" will probably serve us more than well enough. It's not exactly a concise document, but I commend the original designers who developed such a comprehensive feature.
[Element.Planning.-.Geometric.Parameters.pdf](https://github.com/IfcOpenShell/IfcOpenShell/files/9553940/Element.Planning.-.Geometric.Parameters.pdf)
[Element.Planning.-.Overview.Geometric.Parameter_EN.pdf](https://github.com/IfcOpenShell/IfcOpenShell/files/9553941/Element.Planning.-.Overview.Geometric.Parameter_EN.pdf)


vulevukusej:
> Cool initiative! There's an iTWO document with supported quantities, we could use that as a guidance

Yeah I was planning on using that ;)

src/blenderbim/blenderbim/bim/module/pset/qto_calculator.py
```diff
@@ -175,3 +178,63 @@ def get_volume(self, o, vg_index=None):
 
                 volume += v1.dot(v2.cross(v3)) / 6.0
         return volume
+
+def angle_between_vectors(v1, v2):
+    return np.arccos(np.clip(np.dot(v1, v2), -1.0, 1.0))
+
+def calculate_net_lateral_area(object, angle_a=45, angle_b=135):
```
Gorgious56, line 185: `object` overrides the builtin type, you can use `obj` instead, it's widely recognized in blender addons

Gorgious56:
I've tried rather unsuccessfully to leverage `foreach_get` here instead of looping over all indices. I've only attained a marginal speedup (~5%, bigger as the mesh gets denser) because of the `angle_between_vectors` part which takes a LOT of time to process and I couldn't figure how to vectorize. Here's the code for reference :

```
import bpy
import numpy as np
import math

from time import time

def angle_between_vectors(v1, v2):
    return np.arccos(np.clip(np.dot(v1, v2), -1.0, 1.0))


def calculate_net_lateral_area_np(object, angle_a=45, angle_b=135):
    z_axis = (0, 0, 1)
    area = 0
    
    mesh = object.data
    areas = np.zeros(shape=len(mesh.polygons), dtype=float)
    bad_areas = np.zeros(shape=len(mesh.polygons), dtype=bool)
    normals = np.zeros(shape=3 * len(mesh.polygons), dtype=float)
    mesh.polygons.foreach_get("normal", normals)
    normals.shape = (len(mesh.polygons), 3)  # Reshape to get a list of 3D vectors
    
    angles_to_z_axis = np.zeros(shape=len(mesh.polygons), dtype=float)
    for i, normal in enumerate(normals):
        angles_to_z_axis[i] = math.degrees(np.arccos(np.clip(np.dot(z_axis, normal), -1.0, 1.0)))
    
    bad_areas = angles_to_z_axis < angle_a
    bad_areas += angles_to_z_axis > angle_b
    
    mesh.polygons.foreach_get("area", areas)
    areas *= 1 - bad_areas
    
    return sum(areas)

def calculate_net_lateral_area(object, angle_a=45, angle_b=135):
    z_axis = (0, 0, 1)
    area = 0
    polygons = object.data.polygons

    for polygon in polygons:
        normal_vector = (polygon.normal.x, polygon.normal.y, polygon.normal.z)
        angle_to_z_axis = math.degrees(angle_between_vectors(z_axis, normal_vector))

        if angle_to_z_axis < angle_a or angle_to_z_axis > angle_b:
            continue
        area += polygon.area
    return area

start = time()
print(calculate_net_lateral_area_np(bpy.data.objects["Suzanne"]))
print(time() - start)

start = time()
print(calculate_net_lateral_area(bpy.data.objects["Suzanne"]))
print(time() - start)

```

Cheers

vulevukusej:
@Gorgious56 wow, appreciate the contribution mate!  Yeah I was researching ways to go about this but went eith what seemed the simplest - tbh i hadn't logged the time to complete though..

maxfb87:
@vulevukusej very good initiative! I played around a bit with _calculate_net_top_area_ and _calculate net_lateral_area_ and i think that they don't calculate the area in a way that we want.
Look at this picture
image
For example, _calculate_net_lateral_area_ calculates also the area of the inside faces (red pen) and i think that we want only external faces (blue pen).
What do you think?

vulevukusej:
@maxfb87 interesting question because it highlights the need for either more functions or modifiers to the function.  In your example, if the object were a concrete column to be poured on-site, then you would want both the red and blue areas.  If however you were calculating cladding area, or painting area, you'd probably only need the blue area.  Perhaps a modifier for this function could toggle whether polygon normals facing inwards should be ignored/considered when calculating area.

vulevukusej:
[Element Planning - Overview Geometric Parameter_EN.pdf](https://github.com/IfcOpenShell/IfcOpenShell/files/9553870/Element.Planning.-.Overview.Geometric.Parameter_EN.pdf)
[Element Planning - Geometric Parameters.pdf](https://github.com/IfcOpenShell/IfcOpenShell/files/9553871/Element.Planning.-.Geometric.Parameters.pdf)

Moult:
Btw tool.Cad now has tool.Cad.is_a(value, x) which allows you to check whether a value is == x within a tolerance. Might be something you want to start using?

vulevukusej:
> Btw tool.Cad now has tool.Cad.is_a(value, x) which allows you to check whether a value is == x within a tolerance. Might be something you want to start using?

bloody hell, didn't even notice this module - thanks for pointing me in this direction! :-)

PR #2407, opened 2022-09-13
oops, correct version of wrapper that quotes special characters

brunopostle:
Correct version of the wrapper script that quotes spaces and special characters in filenames


Moult:
Maybe should just give direct commit rights? Ping @aothms

brunopostle:
> Maybe should just give direct commit rights? Ping @aothms

But, but, I've only just figured-out how to create clean branches and pull-requests ;)

PR #2412, opened 2022-09-15
update macOS SDK versions for github workflows

dirkolbrich:
This PR updates the macOS SDKs used for the conda-daily workflow:
- macos-10.15 workflow updates to use the macOS 10.15 SDK instead of macOS 10.13 SDK
- macos-11 workflow changed to use the correct macOS 11.3 SDK


dirkolbrich:
Nevermind, I was mistaken about the use of the macOS SDK. The use of macOS SDK 10.13 was correct, as it is needed for the Conda compiler tools -> [Anaconda compiler tools with macOS SDK](https://conda.io/projects/conda-build/en/latest/resources/compiler-tools.html#macos-sdk)

PR #2440, opened 2022-09-25
Update howto advanced mode docs

xpnerd:
Added description to better reflect how import in advanced mode works and where the options can be found.
Added description to elaborate the location of **Unload Project**. Sadly, could not upload `.png`

import-advanced-mode


Moult:
Apologies, the reason I didn't merge this is because it doesn't match the style of images used elsewhere in the docs.

PR #2441, opened 2022-09-25
Add calculate all qtos button

maxfb87:
This PR introduces n.2 new buttons in qtos n-panel for:
- add pset_qto to selected object
- calculate all quantities in defined pset_qto with "guess" method

It works ok, the only thing i think can be improved is the speed because it seems a bit slow; so @Moult could you help me please? :-)

Also, followed the core/tool structure and i wrote only one test for tool; after the review i'll write more.

Thanks,
Massimo

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/qto/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/qto/operator.py
@@ -151,6 +151,20 @@ class QuantifyObjects(bpy.types.Operator):
             PsetData.load(self.file, obj.BIMObjectProperties.ifc_definition_id)
         return {"FINISHED"}
 
+class AssignPsetQto(bpy.types.Operator):
+    bl_idname = "bim.assign_pset_qto"
+    bl_label = "Assign Pset Qto"
+    bl_options = {"REGISTER", "UNDO"}
+    bl_description = "Assign pset qto to selected object"
+
+    @classmethod
+    def poll(cls, context):
+        return context.active_object
+
+    def execute(self, context):
+        core.assign_pset_qto(tool.Qto, selected_objects = context.selected_objects)
+        return {"FINISHED"}       
+
 class CalculateAllQtos(bpy.types.Operator):
     bl_idname = "bim.calculate_all_qtos"
     bl_label = "Calculate All Qtos"
--- a/src/blenderbim/blenderbim/bim/module/qto/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/qto/operator.py
@@ -159,11 +159,11 @@ class AssignPsetQto(bpy.types.Operator):
 
     @classmethod
     def poll(cls, context):
-        return context.active_object
+        return context.active_object and context.selected_objects
 
     def execute(self, context):
         core.assign_pset_qto(tool.Qto, selected_objects = context.selected_objects)
-        return {"FINISHED"}       
+        return {"FINISHED"}
 
 class CalculateAllQtos(bpy.types.Operator):
     bl_idname = "bim.calculate_all_qtos"
@@ -173,7 +173,7 @@ class CalculateAllQtos(bpy.types.Operator):
 
     @classmethod
     def poll(cls, context):
-        return context.active_object
+        return context.active_object and context.selected_objects
 
     def execute(self, context):
         core.calculate_all_qtos(tool.Qto, selected_objects = context.selected_objects)
--- a/src/blenderbim/blenderbim/bim/module/qto/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/qto/operator.py
@@ -150,3 +150,17 @@ class QuantifyObjects(bpy.types.Operator):
             ifcopenshell.api.run("pset.edit_qto", self.file, qto=qto, properties={props.prop_name: result})
             PsetData.load(self.file, obj.BIMObjectProperties.ifc_definition_id)
         return {"FINISHED"}
+
+class CalculateAllQtos(bpy.types.Operator):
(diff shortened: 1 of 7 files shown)


Moult:
Nice! Have you tested to find out exactly what lines of code are the slow bits?

src/blenderbim/blenderbim/bim/module/qto/operator.py
```diff
@@ -150,3 +150,31 @@ def _execute(self, context):
...
+    bl_idname = "bim.assign_pset_qto"
+    bl_label = "Assign Pset Qto"
+    bl_options = {"REGISTER", "UNDO"}
+    bl_description = "Assign pset qto to selected object"
+
+    @classmethod
+    def poll(cls, context):
+        return context.active_object and context.selected_objects
+
+    def execute(self, context):
```
Moult, line 164: Shouldn't this use the new _execute() function? Otherwise undo won't work. Same goes for the other operator.

src/blenderbim/blenderbim/core/qto.py
```diff
@@ -21,3 +21,22 @@ def calculate_circle_radius(qto, obj=None):
     result = qto.get_radius_of_selected_vertices(obj)
     qto.set_qto_result(result)
     return result
+
+def assign_pset_qto(qto, selected_objects):
+    for object in selected_objects:
+        qto.set_active_object(object)
```
Moult, line 27: Setting active object is slow. I don't think it is necessary to do this. Can you recode it so it doesn't need to do this?

src/blenderbim/blenderbim/tool/qto.py
```diff
@@ -35,3 +40,113 @@ def get_radius_of_selected_vertices(cls, obj):
...
+        entity = tool.Ifc.get_entity(object)
+        ifc_object_type = entity.get_info()['type']
+        applicable_pset_names = pset_qto.get_applicable_names(ifc_object_type)
+        return applicable_pset_names
+
+    @classmethod
+    def edit_qto(cls, object, calculated_quantities):
+        file = tool.Ifc.get()
+        pset_qto_id = cls.get_pset_qto_id(object)
+        pset_qto_name = cls.get_pset_qto_name(object)
```
Moult, line 85: This looks overly complex. What's going on here?

src/blenderbim/blenderbim/core/qto.py
```diff
@@ -21,3 +21,22 @@ def calculate_circle_radius(qto, obj=None):
     result = qto.get_radius_of_selected_vertices(obj)
     qto.set_qto_result(result)
     return result
+
+def assign_pset_qto(qto, selected_objects):
+    for object in selected_objects:
```
Moult, line 26: Don't use "object" as a variable name in Python

src/blenderbim/blenderbim/tool/qto.py
```diff
@@ -35,3 +40,113 @@ def get_radius_of_selected_vertices(cls, obj):
...
+
+    @classmethod
+    def get_new_quantity(cls, object, quantity_name, alternative_prop_names):
+        calculator = QtoCalculator()
+        new_quantity = calculator.guess_quantity(quantity_name, alternative_prop_names, object)
+        return new_quantity
+
+    @classmethod
+    def get_non_calculated_value(cls):
+        return 0
```
Moult, line 116: Remove this function

src/blenderbim/blenderbim/core/qto.py
```diff
@@ -21,3 +21,22 @@ def calculate_circle_radius(qto, obj=None):
...
+        qto.set_active_object(object)
+        qto.assign_pset_qto_to_selected_object(object)
```
Moult, line 28: I don't understand this core function. All it does is assign a qto. Why isn't this "built-in" into the existing add_qto function?

src/blenderbim/blenderbim/bim/module/qto/operator.py
maxfb87: Yep, done! I'll do a PR with this and other fixes.

src/blenderbim/blenderbim/core/qto.py
maxfb87: Ok, i deleted it. The only side effect is that the pset panel (in Object Property panel) doesn't refresh so it seems that the button doesn't function but it actually works correctly.
Will this problem be resolved refactoring pset module?

src/blenderbim/blenderbim/core/qto.py
maxfb87: Ook! I'll do a PR with this and other fixes.

src/blenderbim/blenderbim/tool/qto.py
maxfb87: Ok... i added it just to highlight the possibility that "guess quantity" (or in the future "calculate quantity") doesn't actually calculate the value...

src/blenderbim/blenderbim/tool/qto.py
maxfb87: Maybe is it better if the name of this function was "_assign_calculated_quantities_to_object_"?
Because it actually doesn't only edit the qto but it also get the pset_qto assigned to object in order to assign quantities to correct pset..
Am i missing something? :-)

src/blenderbim/blenderbim/core/qto.py
maxfb87: Is it better if the name of this function was "_assign_predefined_pset_qto_"?
Because this function also gets the predefined pset_qto based on the ifc class assigned to the object..
Am i missing something? :-)

PR #2451, opened 2022-09-29
incorporating ElevationOfRefHeight in Section_Level annotation

theoryshaw:
to address: https://community.osarch.org/discussion/1152/blenderbim-how-to-change-0-ground-reference#latest


Moult:
Note to self to rebuild this using the georeference calculation of map conversion OrthogonalHeight + Z.

PR #2460, opened 2022-09-30
Construction Types Browser: Solution to most of the current issues

cvillagrasa:
Hopefully flickering is gone, it sort of works with 3.2, things are previewed on the first popup and there are no insane object creations.

The mouse position when displaying multiple types still needs some fine-tuning.


Moult:
I get the impression the hacks are piling up. Should we take a fresh stab at simpler code?

cvillagrasa:
Indeed, hacks are piling up. They're all related with async-ish behaviour between C and Python, in the ways that Blender allow for now...

That said, we can prioritize user-friendliness over hacks that users aren't even aware of, or less features with no hacks. And we know where we both stand there 😂

Let me say that the hacks are not soo bad in my view, maybe there is a better place in the codebase and they can be moved around. Also, we have to believe that as time passes by, more and more internals from the C codebase will be accessible to Python add-ons, allowing for these hacks to be removed (`is_job_running` is ultranew in 3.3 for instance).

PR #2461, opened 2022-09-30
fix: daily builds of conda ifcopenshell

Krande:
Ran into a lot of `build not compiled for windows` errors on windows when testing the latest ifcopenshell build. So by copying the package-dependencies settings from the conda-forge meta.yaml I was successful in at least fixing this when compiling locally on my windows machine. 

Also, pinning boost-cpp to <=1.74 to at least allow windows builds of ifcopenshell to be created while we figure out how to support boost-cpp >=1.78


aothms:
Awesome, thanks for your continued support to these.

Issue #1880, opened 2021-11-16
Unused build dependencies


Amoki:
Many dependencies are defined here: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/nix/build-all.py#L222 but `yacc` never seems to be used. Is it still relevant? Are there more dependencies that may be removed?
cc @aothms

aothms:
I don't even remember why yacc is there. At first glance I don't see others that can be eliminated.

Moult:
@Amoki could you help remove yacc for some closure on this bug? :)

PR #2466, opened 2022-10-03
Remove yacc in build

Amoki:
Fixes https://github.com/IfcOpenShell/IfcOpenShell/issues/1880

I only tried to build (with success), but I didn't test much

### DIFF
--- a/nix/build-all.py
+++ b/nix/build-all.py
@@ -23,7 +23,7 @@
 # This script builds IfcOpenShell and its dependencies                        #
 #                                                                             #
 # Prerequisites for this script to function correctly:                        #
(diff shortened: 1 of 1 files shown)


Moult:
Thanks so much for returning to a 1 year old bug!

aothms:
I also don't remember exactly why it was there. It already as referenced in the first commit of the build script 18f8d7b3a3. My guess would that occt required them during build. They're probably still in use for parsing STEP, but I guess the generated C sources are backed into the repo https://github.com/Open-Cascade-SAS/OCCT/blob/master/CMakeLists.txt#L829-L841

Issue #1040, opened 2020-10-20
Plans to include Python convenience API for shape analysis?


Moult:
Is there a plan to include convenience functions (i.e. to avoid Python OCC) for common tasks when a shape is created? e.g. volume / area calcs, check intersection, centroid?

aothms:
I think most of this exists:

check intersection: https://github.com/IfcOpenShell/IfcOpenShell/pull/130 (uses the OCCT in C++ IfcOpenShell)

volume / area calcs: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom/IfcGeomRepresentation.h#L80 `calc_area()` etc. not sure if it's accessible from Python also uses OCCT in C++ IfcOpenShell

Can you create a complete overview of desired functionality?

Moult:
Fantastic! I had no idea that existed! I'll record them in the OSArch Wiki to help expose it to users and will definitely make use of them!

I was having issues with fcl in IFC clash reporting clashes on objects that merely have faces that are coplanar, and was wondering if functions existed in IfcOpenShell that would make my life easier. Is there a function that deals with mesh-mesh intersection with consideration of possible co-planarness?

I was also planning to start writing MicroMVDs for geometric requirements, and so those functions would be useful. I didn't really have a list of desired functionality, but wanted to explore what was already created :)

Moult:
Are there checks for manifold / closed geometry? This would be useful to verify whether we can later on recalculate volumes for costing / quantity take-off.

Moult:
Ping @fbpyr you were searching for this I believe.

aothms:
> Are there functions for checking for manifold / closed geometry? This would be useful to verify whether we can later on recalculate volumes for costing / quantity take-off.

Sorry, I missed that, calculate_volume() does actually a check like that, but it's not very informative. Also it can give partial results when there are multiple representation items. How can we improve that?

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcgeom/IfcGeomRepresentation.cpp#L217

Moult:
Just bumping this, since it recently cropped up again in [conversation](https://community.osarch.org/discussion/comment/12249#Comment_12249) it would be nice to have:

 - derive_volume()
 - derive_area()
 - derive_centroid()
 - derive_bbox()

Moult:
[...]

PR #2487, opened 2022-10-08
Add gross volume calculation

maxfb87:
Added gross volume calculation.
The procedure basically creates a new ifc object without voids of the original one and calculate that volume, so the function works only if the object has voids.
Not sure if it is the correct method or the best method, but it functions and i think that it could be considered native ifc.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/pset/qto_calculator.py
+++ b/src/blenderbim/blenderbim/bim/module/pset/qto_calculator.py
@@ -208,16 +208,18 @@ class QtoCalculator:
                 return False
         return True
 
-    def get_volume(self, o):
+    def get_net_volume(self, o):
         o_mesh = bmesh.new()
         o_mesh.from_mesh(o.data)
         return o_mesh.calc_volume()
 
-<<<<<<< HEAD
-=======
     def get_gross_volume(self, o):
         ifc_model = tool.Ifc.get()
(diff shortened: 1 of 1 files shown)


Moult:
Why are you creating a new IFC entity? Can't you just use the old entity? If you want to calculate the body without openings, just disable opening subtractions: https://blenderbim.org/docs-python/ifcopenshell/geometry_settings.html#disable-opening-subtractions as one of the iterator settings.

Also don't use the from_pydata, it's much slower and a better method is to use foreach_set (see example in import_ifc.py).

maxfb87:
@Moult done and done, thanks for the suggestions ... :-) btw, what do you think about this approach? Is it correct?

Moult:
That should work! Is it necessary to create the Blender object? Can you just create a mesh?

BTW this should be heavily simplified if/when @aothms or someone helps expose the already pre-calculated values from IOS e.g. https://blenderbim.org/docs-python/ifcopenshell/geometry_settings.html#validate-quantities or https://github.com/IfcOpenShell/IfcOpenShell/issues/1040

maxfb87:
@Moult yep, i've removed the object, so now bmesh is used to calculate the volume.
I used the object because, in this way, i could use the self.net_volume() function so, if changes the net_volume calculation, changes also the gross_volume.

Issue #2418, opened 2022-09-17
BBIM issue loading materials


myoualid:
Scenario: Add material
    Given an empty IFC project
    And I add a cube
    And the object "Cube" is selected
    And I set "scene.BIMRootProperties.ifc_class" to "IfcWall"
    And I add a material
    When I press "bim.add_material(obj='Material')"
    And I save my file.
    And I close my file.
    When I load my file,
    **Unfortunately** : the object material panel doesn't load the assigned material
    
Scenario2: Add material
    Given an empty IFC project
    And I add a cube
    And the object "Cube" is selected
    And I set "scene.BIMRootProperties.ifc_class" to "IfcWall"
    And I add a material
    When I press "bim.add_material(obj='Material')"
    When I press"bim.add_style"
    And I save my file.
    And I close my file.
    When I load my file,
    **Then happy days**

theoryshaw:
I think you may have to apply the IfcMaterial via 'IFC Object Material' in the Object Properties...
https://www.dropbox.com/s/e3gtpve0ql6xtg0/2022-09-21_14-58-19_Blender_blender.mp4?dl=0

Moult:
Agreed, it doesn't seem like a bug to me, adding a material doesn't actually assign a material. @myoualid can you screenshare?

theoryshaw:
> can we move the object material panel to the Material's tab rather than the Object Properties? It makes a bit more sense that bit of info in the same area.

agree. 
might address it: https://github.com/IfcOpenShell/IfcOpenShell/pull/2492/commits/556dc709704d6998a1660c2cdfd7a6fc9b180562

Moult:
Reverting since this causes breakages when dealing with non-geometric parametric materials like wall type empties, slab type empties, etc, so I guess it wasn't a good idea to move it to the materials tab.

PR #2492, opened 2022-10-09
Relocate the bim pt object material panel in the material context

theoryshaw:
relative to the following idea...

"can we move the object material panel to the Material's tab rather than the Object Properties? It makes a bit more sense that bit of info in the same area."

https://github.com/IfcOpenShell/IfcOpenShell/issues/2418#issuecomment-1268380605


Moult:
I've been avoiding this since the Blender materials panel actually doesn't correlate very well to the concept of an IFC material. It only represents a single material, style, or material with style attached. I can't think of a UX to make it represent material sets, profiles, constituent settings, and layers. E.g. what happens if you remove the active material from the object? The material should exist, so is the Object Material panel independent of the rest of the data in the materials tab? Also, there's a lot of Blender boilerplate with shaders that pushes our IFC panels to the bottom which isn't very nice either.

The ideal scenario is to invent our own IFC Materials tab, but I don't think that's currently possible in Blender.

Moult:
What if, what if... what if we used the Texture Tab instead? Would that be a crazy idea?

Gorgious56:
FWIW I _think_ the panel may be displayed in both contexts with `bl_options = {"INSTANCED"}`

https://docs.blender.org/api/current/bpy.types.Panel.html#bpy.types.Panel.bl_options

Moult:
I caved and switched it, and at least it seems to appear at the top of the screen for me.

PR #2493, opened 2022-10-10
fix get sha256sum directly from release download

fbpyr:
With the recently added `-alpha` tag choco builds seem to update now, 
but there still seemed to have been a glitch with the sha256sum. 
Now I get the sha256sum directly from the realese download. 
This also conveniently removes the need to compile a release in this action, 
which should speed up things. 🙂

### DIFF
--- a/.github/workflows/ci-blenderbim-choco.yml
+++ b/.github/workflows/ci-blenderbim-choco.yml
@@ -8,7 +8,7 @@ on:
     #         │  │ │ ┌───────────── month (1 - 12 or JAN-DEC)
     #         │  │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
     #         *  * * * *
-    - cron: "55 23 * * *" # 5min before utc midnight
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-blenderbim-choco.yml
```diff
@@ -8,7 +8,7 @@ on:
     #         │  │ │ ┌───────────── month (1 - 12 or JAN-DEC)
     #         │  │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
     #         *  * * * *
-    - cron: "55 23 * * *" # 5min before utc midnight
+    - cron: "56 23 * * *" # 4min before utc midnight
```
Moult, line 11: :)

fbpyr: Heheh, yes. 🙂  
Idea is to have the window very small, where someone could publish 
another release before midnight, which would result in a diverging hash for choco.

PR #2503, opened 2022-10-16
searching for an IfcMaterial in IfcMaterialDefinitions

theoryshaw:
A start to trying to address: https://community.osarch.org/discussion/comment/13024/#Comment_13024

This code works now, but could probably be refactored.  I tried to take a stab creating another function to simplify things, but couldn't figure it out. 

Just wanted to send what i have so far, for comment.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/selector.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/selector.py
@@ -188,7 +188,7 @@ class Selector:
             elif token_type == "NULL":
                 value = None
         for element in elements:
-            element_value = cls.get_element_value(element, key)
+            element_value = cls.get_element_value(element, key, value)
             if element_value is None and value is not None and "not" not in comparison:
                 continue
             if comparison and cls.filter_element(element, element_value, comparison, value):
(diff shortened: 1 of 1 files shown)


Moult:
Am I reading this right, that you are proposing to filter based on a single material within a set of materials, they'll specify something like `.IfcWall[material.Name="Foo"]`? And so long as there is a single matching material in the set, the filter succeeds?

So the current constraints are that they can only filter materials, not the layers / profiles / constituents themselves? (maybe `.IfcWall[material.item.Name="Foo"]` to strictly test set items (e.g. layers)?

If the interpretation is correct, I think this is a great addition. It was definitely a missing capability, and the approach used here mirrors IDS. I'll be happy to write some tests.

theoryshaw:
Cool!....yes, you got the intent.

Moult:
Cool, any more changes on your end or do you think it's ready for merging?

theoryshaw:
yes, i think, but do you think it would be a good idea to create a helper function here, so these dont' repeat?

I tried to do it, but couldn't figure it out... mind taking a stab at it?  or give me a clue?

image

aothms:
One helper function could be using `ifc_file.traverse()`, like something like: `[inst for inst in ifc_file.traverse(material_definition) if inst.is_a('IfcMaterial')]`

theoryshaw:
ha... that's cleans stuff up, if i'm understanding it correctly.  Will try to tackle.

theoryshaw:
Thanks @aothms, that's a lot tighter!

PR #2512, opened 2022-10-18
Little improve about code example doc and geometry setting doc

maxfb87:


aothms:
Very cool, thanks:

1)

I think for completeness it would be good to have the simple:

```py
f = ifcopenshell.open(...)
g = ifcopenshell.file(schema=f.schema)
g.add(f.by_type(...)[0])
```

In there and explain that it does copy over recursively, factor in length unit conversion if both files `f` and `g` have project length unit defined. But does not make any other attempts at resulting in a valid file.

2)

I'd prefer we use the terminology _entity instance_. Express (schema language of IFC) has chosen somewhat unfortunate to use the term _entity_ for a _class_. If we use _entity_ for an _instance_ it will confuse people coming from the express side of things.

maxfb87:
@aothms thanks for suggestions. The number 2 i think i fixed, but the number 1 i'm not sure i understood correctly...can you check please?
Also, i've better specified how to use the arguments of the functions...

aothms:
Sorry I was indeed a bit brief. My point was that it'd be good for users to know that you don't need ifcopenshell.api to copy from file to file. That `file.add(other_file[x])` also takes care of quite a few things. But it's fine, I'll update it a bit later :)

PR #2513, opened 2022-10-19
Add IsGroupedBy in function get_decomposition

bbrangeo:
selector example : "@ .IfcZone & .IfcSpace"


Moult:
Hey @bbrangeo ! Thanks for your contribution, but can you please change it to a new function instead of adding to get_decomposition? This is because get_decomposition handles the spatial decomposition hierarchy, and IfcGroup is not part of this hierarchy.

Moult:
One minor suggestion, can the function be renamed to `get_grouped`, so that in the future we can have `get_group` which goes from child element back to the group?

aothms:
Or `get_grouped_by` for being in line with the attribute names in the schema.

Moult:
Yeah it's a bit of strange name isn't it.

PR #2515, opened 2022-10-19
Add function grouped_by in Selector

bbrangeo:


Moult:
It would be awesome if you could also add a test to test_selector.py :)

Moult:
Oh and same for test_element.py for your new is_grouped_by function.

PR #2522, opened 2022-10-22
'X angle' rotates ifcslab/ifcramp in space.

theoryshaw:
a proposal to rotate the ifcslab/ifcramp in space, as well.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/model/slab.py
+++ b/src/blenderbim/blenderbim/bim/module/model/slab.py
@@ -176,7 +176,8 @@ class DumbSlabGenerator:
             obj.matrix_world = matrix_world
             bpy.context.view_layer.update()
             self.collection.objects.link(obj)
-            obj.rotation_euler[0] = self.x_angle
+            rotation = mathutils.Matrix.Rotation(self.x_angle, 4, 'X')
+            obj.matrix_world = rotation
 
         element = blenderbim.core.root.assign_class(
             tool.Ifc,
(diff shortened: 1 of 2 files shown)


theoryshaw:
2022-10-22_15-44-49_Blender_blender

Moult:
The solution isn't quite right. This sets the rotation euler which means 1) I think this means it's out of sync with the matrix_world which may be used by subsequent functions, kinda similar to why you should not edit the "location" attribute in Blender Python and 2) it loses any existing rotation the user might have done.

The correct approach is to multiply the matrix_world matrix by a Rotation matrix. Would you like to try?

theoryshaw:
sure...will try to grok it.

theoryshaw:
Like that?

Moult:
Almost, you've set the matrix world where instead you should multiply it. Try obj.matrix_world @= rotation (the @ symbol is matrix multiplication)

theoryshaw:
hmm... did this...
`obj.matrix_world @= rotation`
but created this behavior.
it did work, when it was just `obj.matrix_world = rotation`
learned me some matrix rotation here: https://www.youtube.com/watch?v=770YOBfOFlQ

2022-10-22_21-39-14_Blender_blender

Moult:
Ah yes, you're right its more intuitive to set the rotation outright.

Have you tested with walls?

theoryshaw:
i think this affects everything, except ifcwalls since i put it in the else statement...

https://github.com/IfcOpenShell/IfcOpenShell/blob/4f585fd1dc16e2e9fcb12a0b13fae9d74ef0095d/src/blenderbim/blenderbim/bim/module/model/wall.py#L273-L282

Moult:
Nice! (sorry I was checking the code on a phone :) )

Issue #2523, opened 2022-10-22
keeping thickness the same, even after 'x angle' is used.


theoryshaw:
for IfcSlabType & IfcRampType that uses IfcMaterialLayerSet, when the 'x angle' is used, it would seem that all these dimensions should be equal.. that is, not change.

![image](https://user-images.githubusercontent.com/507113/197361527-f520e78c-867b-426c-a22f-007f11727cde.png)

Moult:
Yep this is definitely a bug and is actually invalid to the "standardcase" behaviour described in IFC, just gotta do one more step in the math... :)

theoryshaw:
can i try to fix this?

Moult:
Yes please!

theoryshaw:
took at stab: https://github.com/IfcOpenShell/IfcOpenShell/pull/2525

Moult:
Whilst you're chipping away, you might also want to have a play with maintaining the perpendicular wall height for sloped walls. Right now the height represents the extrusion vector's magnitude, which is also not quite right. This would involve checking the code in two places, one for the operator which does ChangeExtrusionXAngle (which should maintain the height) and ChangeExtrusionDepth (which should interpret the depth as the perpendicular depth instead of the non-perpendicualr depth which it does now)

PR #2525, opened 2022-10-23
Maintaining the MaterialLayer thickness when X_Angle is used

theoryshaw:
@Moult is this a correct approach to maintain the MaterialLayer thickness when X_angle is used?

As discussed here: https://github.com/IfcOpenShell/IfcOpenShell/issues/2523


theoryshaw:
...if so, i can take a stab at ifcslabs, etc.

Moult:
Yes, looks correct! (I can't do sin/cos/etc in my head but it should look something like what you did - maybe test with positive and negative angles ... my math is really terrible so I end up doing a lot of manual experimentation and testing with matrixes and trig!)

Issue #2533, opened 2022-10-24
Connect/implement calculating quantities functions


maxfb87:
This is not a issue but a feature request about quantities calculations.

### The Goal
The goal is to connect the already implemented calculating quantities functions with the quantity property.
If there isn't the calculating function or the existing function is not suited, it's necessary to write a new calculating function.
All the Qto Psets and the Quantity Properties have been extracted from IFC4x3 schema and reported into the following .xlsx attached files:
- "qtos_all_quantities.xlsx" has all quantities (file: [qtos_all_quantities.xlsx](https://github.com/IfcOpenShell/IfcOpenShell/files/9853211/qtos_all_quantities.xlsx))
- "qtos_all_quantities_sorted" shows all quantities, grouped into qto pset  (file: [qtos_all_quantities_sorted.xlsx](https://github.com/IfcOpenShell/IfcOpenShell/files/9853213/qtos_all_quantities_sorted.xlsx))
- "qtos_all_quantities_counted" shows how many quantities there are (file: [qtos_all_quantities_counted.xlsx](https://github.com/IfcOpenShell/IfcOpenShell/files/9853212/qtos_all_quantities_counted.xlsx))

### How to help
Only two files are involved, both stored in "/src/blendebim/blenderbim/bim/module/pset"
- "qto_calculator.py" that has all calculating functions (for example get_height that returns the height of an object)
- "calc_quantity_function_mapper" that has a dictionary called _mapper_ that connect the "qto_name.quantity_name" with the function (for example, "Qto_WallBaseQuantities.Height" with get_height)

So, to help, it's enough to work on the mapper to connect the quantity with the proper calculating function.
There are 346 quantities, so help is welcome :-)

Of course, many quantities are repeated so, presumably, it will be used the same function (for example, there are 36 "Lenght" and presumably "get_legth" is a suitable calculating function).

After the mapper dictionary has been modified, to check if the function works correctly, open blendebim->object property panel->IFC object quantity set -> edit the qto and there will be a new icon next to the quantity
![capture](https://user-images.githubusercontent.com/79401028/197573152-d834a5d7-8546-4c6e-84c8-0c5d8bd1b239.jpg)

Cheers, Massimo

Moult:
What's the best way to help / see progress on the number of mapped quantities? Is it possible to either share a spreadsheet, or dump it all out to the calc_quantity_function_mapper.py file with "None" values for the ones which are not yet mapped?

[...]

PR #2534, opened 2022-10-24
Start quantity calculator implementation (related to #2533)

maxfb87:
This is how i think to connect/implement every quantity property calculating function.
I created a new github issue (#2533) that explains the workflow and this is the first commit.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/pset/calc_quantity_function_mapper.py
+++ b/src/blenderbim/blenderbim/bim/module/pset/calc_quantity_function_mapper.py
@@ -1,0 +1,10 @@
+mapper = {
+    "Qto_WallBaseQuantities" : {
+        "Length" : "get_length",
+        "Width" : "get_width",
+        "Height" : "get_height",
+    },
+    "Qto_SlabBaseQuantities" : {
+        "Depth" : "get_height",
+    },
+}
--- a/src/blenderbim/blenderbim/bim/module/pset/ui.py
+++ b/src/blenderbim/blenderbim/bim/module/pset/ui.py
@@ -19,6 +19,8 @@
 from bpy.types import Panel
 from blenderbim.bim.ifc import IfcStore
 from blenderbim.bim.helper import prop_with_search
+from blenderbim.bim.module.pset.calc_quantity_function_mapper import mapper
+import bpy
 from blenderbim.bim.module.pset.data import (
     ObjectPsetsData,
     ObjectQtosData,
@@ -139,6 +141,11 @@ def draw_psetqto_ui(context, pset_id, pset, props, layout, obj_type):
 def draw_psetqto_editable_ui(box, props, prop):
     row = box.row(align=True)
     draw_property(prop, row, copy_operator="bim.copy_property_to_selection")
+    obj = bpy.context.active_object
+    if obj.PsetProperties.active_pset_name in mapper.keys():
+        if prop.name in mapper[obj.PsetProperties.active_pset_name]:
+            op = row.operator("bim.calculate_quantity", icon="MOD_EDGESPLIT", text="")
+            op.prop = prop.name
     if (
         "length" in prop.name.lower()
         or "width" in prop.name.lower()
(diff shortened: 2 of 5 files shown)


src/blenderbim/blenderbim/bim/module/pset/ui.py
```diff
@@ -139,6 +141,11 @@ def draw_psetqto_ui(context, pset_id, pset, props, layout, obj_type):
 def draw_psetqto_editable_ui(box, props, prop):
     row = box.row(align=True)
     draw_property(prop, row, copy_operator="bim.copy_property_to_selection")
+    obj = bpy.context.active_object
```
Moult, line 144: Don't check this ins the draw function, since checking what is going to be a pretty big dictionary on every draw call will be expensive. Instead modify this to precalculate it in the data/import function (I haven't checked exactly where).

Also note that not all properties will have an active object (e.g. task, cost, or resource properties) so check the type of property before doing this.

src/blenderbim/blenderbim/bim/module/pset/calc_quantity_function_mapper.py
```diff
@@ -0,0 +1,10 @@
+mapper = {
```
Moult, line 1: Can put this in a helper.py instead?

maxfb87: I don't exactly get it, do you mean rename the file into "helper.py"? :)
I put the mapper dict in a separate file so it's easier to help

src/blenderbim/blenderbim/bim/module/pset/ui.py
maxfb87: @Moult  Yeah, i forgot about the data thing, sorry.
Is it better now?

src/blenderbim/blenderbim/bim/module/pset/ui.py
```diff
@@ -139,6 +140,11 @@ def draw_psetqto_ui(context, pset_id, pset, props, layout, obj_type):
 def draw_psetqto_editable_ui(box, props, prop):
     row = box.row(align=True)
     draw_property(prop, row, copy_operator="bim.copy_property_to_selection")
+    qto_name = ObjectQtosData.data['qtos'][0]['Name']
```
Moult, line 143: The [0] looks suspicious to me. Why are you fetching that index?

src/blenderbim/blenderbim/bim/module/pset/ui.py
```diff
@@ -139,6 +140,11 @@ def draw_psetqto_ui(context, pset_id, pset, props, layout, obj_type):
...
+    if qto_name in mapper.keys():
```
Moult, line 144: This if statement should be evaluated in a data class, not in the draw function.

src/blenderbim/blenderbim/bim/module/pset/calc_quantity_function_mapper.py
Moult: ah ok, sure then :) never mind

src/blenderbim/blenderbim/bim/module/pset/ui.py
maxfb87: :-) just because it is a list with all qtos assigned to the element, but an element can only has one pset_qto afaik...so the zero

src/blenderbim/blenderbim/bim/module/pset/ui.py
maxfb87: Ook, so i have to create a variable (list, dict or something) that reads the mapper dict, right?
But the "if statement" that checks the data remains in the ui.py..is it right?

Moult: Correct, in the data class or operator import function this should be evaluated as a Boolean. Then the draw function only needs to check a boolean for speed :)

Moult:
Ive merged this since the intention is there but tomorrow when I find some time I'll look in more detail on the things which look suspicious.

Great job!

maxfb87:
Ook, so i wait a bit to update something, feel free to contact me on matrix btw.
Thanks, Massimo

PR #2544, opened 2022-10-28
Blenderbim : Implement tooltips and online docs in the interface

Gorgious56:
I'll merge all my commits related to tooltips and online docs in this PR.

Question : 
- Does it make sense to implement tooltips for property sets ? Currently only its properties are described, and it seems there is no description in the IFC schema for actual top level property sets, but I find it odd.


Gorgious56:
@Moult Nice!

New question : Is there a way to extract attribute enum items descriptions ? I can fetch the predefined types with no problem. 

But for example I'd like to get the descriptions from IfcWindow.Attributes.PartitioningType attribute. In [the other resource there is one](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/schema/enum_descriptions.json#L1855) so I wonder if it's possible to fetch it in the new doc system ?

Gorgious56:
Here's snippets of latest commits :

Animation

I'd like to populate the tooltips dynamically instead of having to right click > hover but that requires a ugly hack to un_register then re_register the attributes which I don't think I can do easily.

Animation2

Moult:
Yes it is possible, will add it to the list for @Andrej730 :)

I dug back in the Git history and discovered I used HTML scraping https://github.com/IfcOpenShell/IfcOpenShell/commit/d2fcf07b371900d445b52f68102b20ac18b9a5c5#diff-af6af938c66c38eec49564b490810124736f0b50d335e1e960cdc4f546b4204f no idea if it's helpful anymore :) Maybe marginally helpful for IFC2X3 where the XMLs seem to be missing some docs.

Moult:
That looks cool! Agree better than an unregister/register hack. What if instead of hovering you can click it and it opens up the description in a popup window? That popup window can also include a button to launch the website?

Gorgious56:
Alright ! Here goes. I didn't find a way to neatly wrap paragraphs in an operator popup window... so I left the operator dynamic description as a fallback :

Animation2

I realized the attribute doc doesn't return the attribute definitions that are defined in parent classes. For example here in IfcWindow I get OverallHeight, OverallWidth, PredefinedType, PartitioningType, UserDefinedPartitioningType but I don't get ObjectType which is an IfcObject attribute, and GlobalId, OwnerHistory, Name, Description which are IfcRoot attributes. Would it be possible to modify the API to be able to retrieve the parent class definitions ? 

eg with an optional attribute to `get_attribute_doc(entity, attribute, recursive=True)` or something like that ?

Gorgious56:
Unit types also don't have a description :
image

Moult:
> Would it be possible to modify the API to be able to retrieve the parent class definitions ? Yep, leave the function signature as is, I'll add it to the list for @Andrej730 :)

Andrej730:
> @Gorgious56 good idea. IFC does actually have pset definitions. I'll assign the extraction task to @Andrej730 :)
> 
> @Andrej730 here is an example markdown to be extracted for IFC2X3 and IFC4: https://github.com/buildingSMART/IFC/blob/Ifc4.0.2.1/Sections/Domain%20specific%20data%20schemas/Schemas/IfcBuildingControlsDomain/PropertySets/Pset_ActuatorTypeCommon/Documentation.md :)

Added in 0b5ccd88a52bb9f5715a29e99dfcb1bef6425278
Example command for getting pset description: `get_property_set_doc("IFC4", "Pset_ZoneCommon")['description']`

Andrej730:
> I realized the attribute doc doesn't return the attribute definitions that are defined in parent classes. For example here in IfcWindow I get OverallHeight, OverallWidth, PredefinedType, PartitioningType, UserDefinedPartitioningType but I don't get ObjectType which is an IfcObject attribute, and GlobalId, OwnerHistory, Name, Description which are IfcRoot attributes. Would it be possible to modify the API to be able to retrieve the parent class definitions ?
> 
> eg with an optional attribute to `get_attribute_doc(entity, attribute, recursive=True)` or something like that ?

@Gorgious56 Hello!) Added it in 937f39c898690736f29268c056fb4786116a99eb, example: `get_attribute_doc("IFC4", "IfcWindow", "OwnerHistory", recursive=True)`

Moult:
Oh, @Andrej730 I was going to suggest using IfcOpenShell to query the parent, which would then make it easier to support runtime loaded schemas in the future.

```
>>> import ifcopenshell
>>> s = ifcopenshell.ifcopenshell_wrapper.schema_by_name('IFC4')
>>> d = s.declaration_by_name('IfcWall')
>>> d.supertype()
<entity IfcBuildingElement>

>>> d.supertype().name()
'IfcBuildingElement'

>>> d.supertype().supertype().name()
'IfcElement'
```

Andrej730:
> Oh, @Andrej730 I was going to suggest using IfcOpenShell to query the parent, which would then make it easier to support runtime loaded schemas in the future rather than hardcoding into JSON.
> 
> Also, recursive should definitely be true by default, I can't imagine any case where you wouldn't want it to be true unless you were doing some weird custom parsing.

Fixed in 9a25db6584ce61c57ca0ed5dfdcc57a4ef0b0e74 and 17ccd0adf7d9719dcd7312c4db9eaaaa4e571560

Moult:
@Gorgious56 I discovered today with @theoryshaw that https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/__init__.py#L162 doesn't work on Blender <= 3.2, is it possible to verify or still support these versions? (I'm also OK with disabling the feature for older versions).

Andrej730:
> Unit types also don't have a description : image

Added in d9f8aa4e7b1af0a47f983442fe8490f2cc447b3c, now it can be accessed with API function `get_type_doc(version, ifc_type)`

Gorgious56:
This is great thank you !! @Andrej730 I'll start playing with the new toys :)

I was wondering @Moult what you think of https://github.com/Gorgious56/IfcOpenShell/blob/d9b1c26bf0ef5ce7360b467916eaa1626b133952/src/blenderbim/blenderbim/bim/prop.py#L109-L131 these functions, at the beginning I felt it would reduce the amount of duplicate code but now I think it's adding another layer of abstraction / complexity to the docs api. 

Are you in favor of removing them and calling the ifcopenshell.util.doc function directly ?

Gorgious56:
I can't test https://github.com/IfcOpenShell/IfcOpenShell/commit/d9f8aa4e7b1af0a47f983442fe8490f2cc447b3c because for some reason the pset templates don't load in a starting project. Any idea why ?

```
op.py", line 48, in getPsetTemplateFiles
    return PsetTemplatesData.data["pset_template_files"]
KeyError: 'pset_template_files'
```

And I have a new batch of missing descriptions :) (if they exist in the schema)

image

image

Moult:
> Are you in favor of removing them and calling the ifcopenshell.util.doc function directly ?

Yes.

> for some reason the pset templates don't load in a starting project. Any idea why ?

Does `os.path.join(bpy.context.scene.BIMProperties.data_dir, "pset")` exist?

Gorgious56:
I've had to merge the main branch which sent the lines count through the roof. I'll see what I can do to fix it.

Moult:
AWESOME work! Super keen to see what users think of this!

PR #2552, opened 2022-10-30
add BCF reporter to IDS

atomczak:
plus add instructions and fix error when the reason is not a list.

### DIFF
--- a/src/ifctester/ifctester/reporter.py
+++ b/src/ifctester/ifctester/reporter.py
@@ -204,63 +204,58 @@ class Html(Json):
                 return outfile.write(pystache.render(file.read(), self.results))
 
 
-class BcfHandler(logging.StreamHandler):
-    """Logging handler for creation of BCF report files.
+class Bcf(Reporter):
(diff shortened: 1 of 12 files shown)


src/ifctester/ifctester/reporter.py
```diff
@@ -289,18 +284,17 @@ def emit(self, log_content):
...
-        self.bcf.add_viewpoint(topic, viewpoint)
-
-    def flush(self):
-        """Saves the BCF report to file. Triggered at the end of the validation process."""
-        if not self.filepath:
-            self.filepath = os.getcwd() + r"\IDS_report.bcf"
-        if not (self.filepath.endswith(".bcf") or self.filepath.endswith(".bcfzip")):
-            self.filepath = self.filepath + r"\IDS_report.bcf"
-        self.bcf.save_project(self.filepath)
+        # add single pixel png file as a snapshot, as viewpoints without snapshots won't open in viewers
```
Moult, line 293: Gotta start filing bugs with those viewers then

PR #2555, opened 2022-10-31
Openings should be either polygonal or extrusion based

Gorgious56:
Following the task in the project board, I've added a dropdown in the tool interface to select which type of void the user wants to apply.

Here's a short video explaing the workflow

capture

It seems all types works except for `IfcExtrudedAreaSolid/IfcMaterialProfileSetUsage` which throws an error. 

```
  File "c:\Program Files (x86)\Steam\steamapps\common\Blender\3.3\python\lib\site-packages\ifcopenshell\api\geometry\add_representation.py", line 69, in execute
    return self.create_model_representation()
  File "c:\Program Files (x86)\Steam\steamapps\common\Blender\3.3\python\lib\site-packages\ifcopenshell\api\geometry\add_representation.py", line 122, in create_model_representation
    return self.create_variable_representation()
  File "c:\Program Files (x86)\Steam\steamapps\common\Blender\3.3\python\lib\site-packages\ifcopenshell\api\geometry\add_representation.py", line 239, in create_variable_representation
    return self.create_material_profile_set_extrusion_representation()
  File "c:\Program Files (x86)\Steam\steamapps\common\Blender\3.3\python\lib\site-packages\ifcopenshell\api\geometry\add_representation.py", line 523, in create_material_profile_set_extrusion_representation
    profile_set = self.settings["profile_set_usage"].ForProfileSet
AttributeError: 'NoneType' object has no attribute 'ForProfileSet'
```

I don't really know how to test if everything is in order in the ifc file in the end though.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/model/opening.py
+++ b/src/blenderbim/blenderbim/bim/module/model/opening.py
@@ -367,6 +367,8 @@ class AddPotentialOpening(Operator, AddObjectHelper):
     circle_radius: FloatProperty(name="Circle Radius", min=0, default=0.5)
     extrusion_resolution: IntProperty(name="Resolution", min=1, default=4)
     extrusion_depth: FloatProperty(name="Extrusion Depth", default=1)
+    rectangle_dim_x: FloatProperty(name="X Dimension", default=1, min=0)
+    rectangle_dim_y: FloatProperty(name="Y Dimension", default=1, min=0)
 
     def draw_settings(context, layout, tool):
         row = self.layout.row()
@@ -386,6 +388,10 @@ class AddPotentialOpening(Operator, AddObjectHelper):
         elif self.representation == "IfcExtrudedAreaSolid/IfcCircleProfileDef":
             layout.prop(self, "extrusion_resolution")
             layout.prop(self, "circle_radius")
+            layout.prop(self, "extrusion_depth")
+        elif self.representation == "IfcExtrudedAreaSolid/IfcRectangleProfileDef":
+            layout.prop(self, "rectangle_dim_x")
+            layout.prop(self, "rectangle_dim_y")
             layout.prop(self, "extrusion_depth")
 
     def invoke(self, context, event):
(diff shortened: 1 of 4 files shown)


Gorgious56:
I guess the next step is to ask the user for the type before clicking on "Add Void" and instead of adding a cube, adding the relevant representation ? ie a cylinder for circle extrusion, etc ?

Moult:
Indeed it will never work for "IfcExtrudedAreaSolid/IfcMaterialProfileSetUsage" because that is a parametric type based profile, and openings don't have types, so all good :)

> I guess the next step is to ask the user for the type before clicking on "Add Void" and instead of adding a cube, adding the relevant representation ? ie a cylinder for circle extrusion, etc ?

Yep :) Extrusions are always in the +Z, so that's the convention that needs to be followed.

Gorgious56:
Quick update 
capture
capture

And a messy torus void: 
image

Test in BIMvision
image



@Moult I was thinking about the best way to implement this. Should profiles be tied to the new Profile editor you created so that the user can select an existing profile in a list ? for now I just added hardcoded profile parameters but I'm sure it can be tied more tightly with existing systems. I also don't really know how to handle arbitrary extrusions and text literals ?

Moult:
> Should profiles be tied to the new Profile editor you created

Nah, managed profiles are necessary for standardised type-based modeling, like structural steel, mullion, skirting, molding, curtain wall transoms, etc. Openings / penetrations and booleans are the one thing that don't need any managed profiles.

Openings are almost, almost always a rectangle or cylinder, nothing else ... because that's what you can build. Booleans are either those simple shapes, or flat polyline planes that are bespoke to an area. Anything more complex is an arbitrary mesh because an architect is being creative.

> how to handle arbitrary extrusions and text literals

An arbitrary extrusion is just a mesh. The user can edit it using ops.bim.enable_editing_extrusion_profile and ops.bim.edit_extrusion_profile I suspect. A text literal is... unrelated to openings and shouldn't be supported I guess?

Issue #2361, opened 2022-08-24
Conda CI daily build not succeeding often


aothms:
See https://github.com/IfcOpenShell/IfcOpenShell/actions/workflows/ci-daily-build.yml

with 2/x runs succeeded. Always macos-11-py39?

Maybe due to some conda conflict around lark:

```
ClobberWarning: This transaction has incompatible packages due to a shared path.
  packages: conda-forge/noarch::lark-parser-0.12.0-pyhd8ed1ab_0, conda-forge/noarch::lark-1.1.2-pyhd8ed1ab_0
  path: 'lib/python3.9/site-packages/lark/__pycache__/visitors.cpython-39.pyc'
```

@Krande any ideas?

Krande:
I have noticed it as well. 

Hm, it's difficult to say. The ClobberWarning I believe appears on most (if not all) macOS runtime versions and python versions

![image](https://user-images.githubusercontent.com/9642232/186596949-a94e15c5-4a78-48ef-91bd-707aac750560.png)

![image](https://user-images.githubusercontent.com/9642232/186597011-4c0d21a2-c4ca-44c3-b46e-61ca9d43912d.png)

![image](https://user-images.githubusercontent.com/9642232/186597268-05911641-6335-418c-87a7-e9036a6b04b4.png)

The issue itself seems to be that the test of checking if python can do `import ifcopenshell` results in a `ModuleNotFoundError: No module named 'ifcopenshell'` error. 

I am not sure it is related, but I have observed lately that some Linux agents are randomly killed by lack of VM resources. 

So to eliminate the possibility of this being a result of random VM resource issues I have re-started the last failing macOS-11-py39 job. If the issue prevails, then it is something strange with this particular combination of python and runtime.

I don't really know what the `ClobberWarning` tells us, but as you pointed out, the error refers to the `lark` package. 

I guess if the issue is related to VM resources we might consider reducing the frequency (not on every successful ci.yml run) or perhaps limit to only 1 runtime version? I have noticed that other github action pipelines are often queued due to the resources demanded by the 12 compilations we are doing in parallel. 

Another thing I've noticed slightly related would be if we want to substitute OSX version v10.15 with v12 of macOS. I believe we're seeing some deprecation notices for both osx 10.15 and ubuntu 18.04

![image](https://user-images.githubusercontent.com/9642232/186601263-5010c4e3-df4e-4ad7-a8bb-fae781f9fb47.png)

[...]

PR #2558, opened 2022-11-01
fix: add conditional #include <windows.h> for certain files breaking …

Krande:
…with boost v1.78+ based on discussion in https://github.com/IfcOpenShell/IfcOpenShell/issues/2361

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -26,10 +26,6 @@
  *                                                                              *
  ********************************************************************************/
 
-#ifdef _MSC_VER
-#include <windows.h>
-#endif
-
 #include "../serializers/ColladaSerializer.h"
 #include "../serializers/GltfSerializer.h"
 #include "../serializers/HdfSerializer.h"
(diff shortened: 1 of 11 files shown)


aothms:
Thanks :)

I think the only include of boost/regex is here 

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcgeom_schema_agnostic/IfcGeomFilter.h#L31

Did you also try sliding it in right before that in that header file?

Krande:
I think I tried that first. Although I did try a lot of different things at first when I was experimenting. So I can make another check just to be sure. It would be nice to isolate the hack to a single location :)

Krande:
Oh, I probably did not try the `BOOST_REGEX_NO_W32`. It seems adding only that to the line before including regex solves everything :) 

I'll let the local test finish before updating my PR

Krande:
@aothms I've made modifications now so that the only change is adding `BOOST_REGEX_NO_W32`. It worked on my local test, but I haven't tested it on a github actions workflow, yet, so we'll have to see how it performs after merge. But if you're happy, then I can do a "squash and merge" :)

aothms:
Thanks a lot @Krande

Issue #2545, opened 2022-10-28
IfcMaterialLayer Priority : Enforce range [0...100] ?


Gorgious56:
According to the docs, unless I'm mistaken the "Priority" attribute value of [IfcMaterialLayer](https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/schema/ifcmaterialresource/lexical/ifcmateriallayer.htm) should be in the enclosed range [0...100].

There is currently nothing to prevent users to set the attribute outside of this bound. I don't know if :
- This shouldn't be enforced anywhere (in this case this issue can be closed)
- This shouldn't be enforced but we add a warning popup next to the field to warn users the value is not valid
- This should be enforced by a callback to the field in Blenderbim after the user validates the value
![image](https://user-images.githubusercontent.com/25156105/198582224-8d82e255-1222-496e-a65d-b5a54905c6b5.png)
- This should be enforced in Blenderbim when the user validates the material attributes editing state
![image](https://user-images.githubusercontent.com/25156105/198582482-a1ed13bd-27d8-47b3-8e93-28513eb0f33e.png)
- This should be enforced by the ifcopenshell api when the validation occurs. In this case maybe values < 0 should be clamped to 0 and values > 100 should be clamped to 100 ?

It could be part of a broader project for other attributes that may need some kind of state validation. I would be happy to contribute if need be.

Moult:
There aren't too many of these. I think the easiest thing to do is to do it in the API. Would you like to have a shot?

Gorgious56:
Sure, I'll dig around but if you have some pointers for where to begin I'm all ears :)

Moult:
Just clamp it here I guess:

 - https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/api/material/add_layer.py
 - https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/api/material/edit_layer.py

aothms:
How about this: we build an IDS spec for these kind of "soft" constraints in the schema. That way users can exercise some control over which of these constraints they want to honor. As a first immediate result it would allow us to do validation on the in-memory or exported IFC model. In a later stage we would be able to infer the correct blenderbim controls based on the IDS. All in a vendor neutral way; the entire community can benefit.

[...]

PR #2562, opened 2022-11-02
Numerical Attribute validation callback

Gorgious56:
This is a proof of concept following #2545, to enforce constraints directly in the interface.

The [IfcMaterialLayer](https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/schema/ifcmaterialresource/lexical/ifcmateriallayer.htm) docs indicate its Priority attribute shall be an integer value between 0 and 100.

Animation2

How it works :
- Add two attributes to the Attribute class (for the minimum and maximum constraints)
- When creating the attribute, fetch from a custom dictionary if the attribute has specific numerical constraints
- If it has, change the two attributes to reflect it. Otherwise keep the min an max attributes to very low and very high values so they're never fired
- Use a callback when the integer or float value is changed

I can see this being updated also with values that should alway be > 0 like these ones ? 
image
 Although as far as I understand these are not supposed to be filled in by hand but computed automatically so maybe this doesn't apply here.

I'm of course open to any criticism regarding improvements.

### DIFF
--- a/src/blenderbim/blenderbim/bim/helper.py
+++ b/src/blenderbim/blenderbim/bim/helper.py
@@ -109,6 +109,17 @@ def import_attribute(attribute, props, data, callback=None):
         if data[new.name]:
             new.enum_value = data[new.name]
     add_attribute_description(new)
+    add_attribute_min_max(new)
+
+
+ATTRIBUTE_MIN_MAX_CONSTRAINTS = {"IfcMaterialLayer": {"Priority": {"value_min": 0, "value_max": 100}}}
+
+
+def add_attribute_min_max(attribute_blender):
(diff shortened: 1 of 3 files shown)


Moult:
Only a minor comment that separating has_max (boolt) and max_value (float) is probably more explicit but that's OK.

Gorgious56:
@Moult do you mean that you rather have a flag on the attribute that we enable only if there is actually a constraint on its max or min value ? 

Or that you'd rather have a max_value_float and max_value_int and separate callbacks ? I can make the change if necessary

Moult:
> @Moult do you mean that you rather have a flag on the attribute that we enable only if there is actually a constraint on its max or min value ?

Yep - that way we don't rely on magic large/small numbers

Issue #43, opened 2016-03-12
Embed more detailed version information (git commit id) in binaries using build scripts


aothms:
Now that there are build scripts for all major operating systems. Run a `git describe` and capture the output in a preprocessor directive somewhere in order to have more detailed version information embedded in binaries.

Moult:
This is done, isn't it? ifcopenshell.version reports the git-sha.

PR #2576, opened 2022-11-07
IfcSverchok - GSoC 2022

mdjska:
Final PR for issue #43 (https://github.com/opencax/GSoC/issues/43) as part of GSoC 2022.


Moult:
BTW can you please add the copyright header to all files? Make sure it's 2022 and feel free to put your name/email as the copyright holder.

src/ifcsverchok/__init__.py
```diff
@@ -84,7 +96,131 @@ def make_node_list():
...
+    node_group: bpy.props.StringProperty(default="")
+    force_mode: bpy.props.BoolProperty(default=False)
+
+    @classmethod
+    def poll(cls, context):
+        return any("IFC" in n for n in context.space_data.edit_tree.nodes.keys())
+
+    def ensure_hirarchy(self, file):
+        elements_in_buildings = []
+        if not 0 <= 0 < len(file.by_type("IfcBuilding")):
```
aothms, line 145: This is an interesting if statement. Is that `if len(file.by_type("IfcBuilding")) == 0`?

src/ifcsverchok/__init__.py
```diff
@@ -84,7 +96,131 @@ def make_node_list():
...
+            for building in file.by_type("IfcBuilding"):
+                elements = ifcopenshell.util.element.get_decomposition(building)
+                elements_in_buildings.extend(elements)
+
+        for spatial in (file.by_type("IfcSpatialElement") or file.by_type("IfcSpatialStructureElement")):
+            if (not (spatial.is_a("IfcSite") or spatial.is_a("IfcBuilding")) and (spatial not in elements_in_buildings)):
+                elements = ifcopenshell.util.element.get_decomposition(spatial)
+                ifcopenshell.api.run("aggregate.assign_object", file, product=spatial, relating_object=file.by_type("IfcBuilding")[0])
 
+        elements_in_buildings_after = []
```
aothms, line 158: This variable appears not to be read.

In general, `set`s would be more appropriate for this kind of things. You don't depend on the order of the list and sets are much more efficient to check for containment. They don't have duplicates and allow for easy removal of elements also.

src/ifcsverchok/__init__.py
```diff
@@ -84,7 +96,131 @@ def make_node_list():
...
+    """Update current Sverchok node tree"""
+    bl_idname = "ifc.sverchok_update_current"
+    bl_label = "Update current node tree"
+    bl_options = {'REGISTER', 'UNDO', 'INTERNAL'}
+
+    node_group: bpy.props.StringProperty(default="")
+    force_mode: bpy.props.BoolProperty(default=False)
+
+    def execute(self, context):
+        print("#"*10, "Update current node tree")
```
aothms, line 115: For now it's fine to center around buildings. But for the future it'd be good to also allow other infra-related spatial structure elements, such as IfcBridge. Maybe make a comment regarding that.

src/ifcsverchok/__init__.py
```diff
@@ -84,7 +96,131 @@ def make_node_list():
...
+        self.file = file
+        return
+
+    def execute(self, context):
+        self.file = SvIfcStore.file
+        if not self.file:
+            raise Exception("No IFC file in SvIfcStore.")
+        _, ext = splitext(self.filepath)
+        if not ext:
+            raise Exception("Bad path. Provide a path to a file.")
```
aothms, line 190: Strictly speaking filenames do not need to have an extension. Did you meant checking something like `not os.path.isdir()`?

src/ifcsverchok/nodes/ifc/add_spatial_element.py
```diff
@@ -0,0 +1,149 @@
...
+        self.inputs.new("SvStringsSocket", "Elements").prop_name = "Elements"
+        self.outputs.new("SvStringsSocket", "Entities")
+        self.node_dict[hash(self)] = {}
+    
+    def draw_buttons(self, context, layout):
+        layout.operator("node.sv_ifc_tooltip", text="", icon="QUESTION", emboss=False).tooltip = "Ifc entity by type."
+
+    def process(self):
+        self.sv_input_names = [i.name for i in self.inputs]
+        if hash(self) not in self.node_dict:
```
aothms, line 33: Why `hash(self)` and not just `self`? This generates collisions. dicts are are hash maps, so you don't need to manually hash the key. Internally in a hash map the hash is used to determine the "bucket" the value will go into, but there will still be actual equality testing within the bucket. As the hash is not necessarily unique.

src/ifcsverchok/__init__.py
mdjska: in this case yes. It was a way to check if index exists without throwing an IndexError

src/ifcsverchok/__init__.py
mdjska: Yea, I know files don't need extensions, but since there are no other checks of overwriting/validity, I made the assumption that if there is an extension the written path was conscious. Since os.path.isdir() of an 'invalid' path will return false. But I wrote this for the write_file node, where the user can actually give raw input paths. The file picker pop-up window in the panel, makes the whole operation much safer, so maybe it can be deleted/changed.

src/ifcsverchok/nodes/ifc/add_spatial_element.py
mdjska: In short, hash(self) was the only number/ID that I could find was, and stayed, unique for each node instance. Sverchok uses the `node_id` attribute internally, which is also based on hash(self) (https://github.com/nortikin/sverchok/issues/2445). And I don't remember exactly why right now, but I had some issues with that attribute either not staying unique or changing during the life time of the node, but didn't have the same issues with using hash(self) directly. I would have to test it again.
I'm not sure how you would do it with just `self`? The issue with not having an internal node_dict was, that two nodes of the same type (Eg. two "IFC By Type" nodes) will not make new class instances, but both use the same one, so stored information would leak from one node to another. If you have an idea on how this can be handled differently, I'd like to change it. Not sure if the problem mentioned [here](https://github.com/nortikin/sverchok/issues/3564#issuecomment-695770526) also applies?

aothms: That approach there would indeed not stay unique as it explicitly `xor`s in the current time. I'm also not sure how authoritative that post is. You can see explained here what the relationship is between id() [what internally would be used when you use `self`] and hash() for types that don't override __hash__ https://stackoverflow.com/a/33572401 Python may recycle ids when they are disposed so that's why they factor in the current time. But the problem remains. I still think `self` here is better than `hash(self)`. If you want a numeric ID, the proper way would probably be a globally defined auto incrementing counter https://stackoverflow.com/a/1045724

Anyway this is a bit of a rabbit hole. I'm fine if we keep it like this. Just wanted to provide some bg info. I'll mark this as resolved.

Moult:
Sorry just checking in after some time overseas, is this ready for merging?

mdjska:
> Sorry just checking in after some time overseas, is this ready for merging?

Sorry for the delay, I have resolved the review, so it should be ready for you to merge now.

nortikin:
https://github.com/nortikin/sverchok/commit/8192dcdd6de89908e2541432f688949a13ca252a registration problems?

Moult:
I merged this using command line copy pasted from the Github step by step instructions, which worked, but somehow there now seem to be duplicate commits for the most recently N commits, which makes absolutely no sense to me.

Moult:
I seemed to manage to get into this situation: https://stackoverflow.com/questions/9264314/git-commits-are-duplicated-in-the-same-branch-after-doing-a-rebase through a rebase, though how I ended in the situation might've been different from the OP in that SO question.

The end problem was identical though, so I _think_ I fixed it. I really hope I fixed it.

nortikin:
> I seemed to manage to get into this situation: https://stackoverflow.com/questions/9264314/git-commits-are-duplicated-in-the-same-branch-after-doing-a-rebase through a rebase, though how I ended in the situation might've been different from the OP in that SO question.
> 
> The end problem was identical though, so I _think_ I fixed it. I really hope I fixed it.

image

nortikin:
``` bash
git checkout v0.7.0
  File "~/Downloads/INSTALL/blender-3.3.1-linux-x64/3.3/scripts/modules/addon_utils.py", line 333, in enable
    mod = __import__(module_name)
  File "~/.config/blender/3.3/scripts/addons/ifcsverchok/__init__.py", line 33, in <module>
    from sverchok.core import make_node_list
ImportError: cannot import name 'make_node_list' from 'sverchok.core' (~/.config/blender/3.3/scripts/addons/sverchok/core/__init__.py)
```

nortikin:
we have not make node list definition

nortikin:
sverchok.menu rewrited also, generally follow https://github.com/nortikin/sverchok/commit/8192dcdd6de89908e2541432f688949a13ca252a#diff-6e982446d45d5ffa23213eb0cb66991fb02283df19b49da3bc8057ce1ab5094a

nortikin:
it breaks menues, but makes it easier to maintain after with yaml file

Moult:
@nortikin master being 5k commits behind v0.7.0 is expected, as we don't use the master branch. Someday in the future we probably should revisit the branching model that we use (perhaps for v0.8.0?) but what you're seeing is expected :)

Thanks for the hint about how nodes are now defined, @mdjska will check whether or not she needs to update anything :)

PR #2618, opened 2022-11-25
Added ReloadSelectedIfcFile operator

Andrej730:
Added ReloadSelectedIfcFile operator and added it to IFC Quality Control -> IFC Debug, it's useful when you're debugging and doing a lot of iterations on the same file.

Submitting the change as PR because not sure maybe there is already some way to reload the file without selecting it each time and I'm just missing it.


Moult:
Great idea! Thanks so much!

PR #2622, opened 2022-11-27
Feature/spreadsheet writer

C-Claus:
Added spreadsheet writer module.


src/blenderbim/blenderbim/bim/module/spreadsheet_writer/operator.py
```diff
@@ -0,0 +1,652 @@
+import os
+import sys
+import time
+import site
+import collections
+import subprocess
+
+site.addsitedir(os.path.join(os.path.dirname(os.path.realpath(__file__)), "libs", "site", "packages"))
```
Gorgious56, line 8: I'm not sure this is mandatory anymore

src/blenderbim/blenderbim/bim/module/spreadsheet_writer/operator.py
```diff
@@ -0,0 +1,652 @@
...
+
+import blenderbim.bim.import_ifc
+from blenderbim.bim.ifc import IfcStore
+import blenderbim.tool as tool
+import ifcopenshell
+
+import openpyxl
+from openpyxl import load_workbook
+import pandas as pd
+import xlsxwriter
```
Gorgious56, line 23: These few lines will throw an error on startup if the user doesn't have these packages installed in their environment. Is the plan to distribute openpyxl and xlsxwriter with blenderbim ? (inside site-packages)

src/blenderbim/blenderbim/bim/module/spreadsheet_writer/operator.py
```diff
@@ -0,0 +1,652 @@
...
+        height = 'Height'
+        area = 'Area' 
+        netarea = 'NetArea'
+        netsidearea = 'NetSideArea'
+        volume = 'Volume'
+        netvolume = 'NetVolume'
+        perimeter = 'Perimeter'
+
+        for product in products:
+            ##################################################################
```
Gorgious56, line 123: Instead of using comments as delimiters it is better practice to decompose bigger functions into multiple smaller ones. A good length to aim for in a function is 7 - 10 lines top.

src/blenderbim/blenderbim/bim/module/spreadsheet_writer/operator.py
```diff
@@ -0,0 +1,652 @@
...
+            if blenderbim_spreadsheet_properties.my_loadbearing:     
+                ifc_dictionary[loadbearing].append(self.get_common_properties(context, ifcproduct=product, property_name=loadbearing)[0])
+                 
+            if blenderbim_spreadsheet_properties.my_firerating:    
+                ifc_dictionary[fire_rating].append(self.get_common_properties(context, ifcproduct=product, property_name=fire_rating)[0])
+               
+            ##################################################################
+            ##################### Base Quantities ############################
+            ##################################################################
+            if blenderbim_spreadsheet_properties.my_length:
```
Gorgious56, line 165: Since the code is reused a lot it may be condensed a bit, for instance :
```
for quantity_name in (length, width, height):  # etc.
    if getattr(blenderbim_spreadsheet_properties, f"my_{quantity_name.lower()}"):
        ifc_dictionary[quantity_name].append(self.get_quantities(context, ifcproduct=product, quantity_name=quantity_name)[0])
```

src/blenderbim/blenderbim/bim/module/spreadsheet_writer/operator.py
```diff
@@ -0,0 +1,652 @@
...
+        print("Unhide all")
+        
+        for obj in bpy.data.objects:
+            obj.hide_viewport = False 
+        
+        return {'FINISHED'}  
+ 
+ 
+class MyCollectionActions(bpy.types.Operator):
+    bl_idname = "my.collection_actions"
```
Gorgious56, line 638: Since this will be merged in a rather extensive codebase, it's a good idea to try and reduce the chance of having name collisions. You can rename this to something more specific eg. "bim.spreadsheet_collection_action"

Gorgious56:
This is awesome, I can't wait to see this implemented for all to test out ! 

I left a few comments from skimming through the files, of course these are just comments, mostly to make the code more in line with the general vibe of the existing codebase. 

Feel free to discard, no hard feelings ;)

src/blenderbim/blenderbim/bim/__init__.py
```diff
@@ -70,6 +70,7 @@
     "covetool": None,
     "augin": None,
     "debug": None,
+    "spreadsheet_writer" None,
```
Gorgious56, line 73: This will likely throw an error, I think you forgot a `:` in there ;)

src/blenderbim/blenderbim/bim/module/spreadsheet_writer/operator.py
C-Claus: Good point, I forgot to remove it.

src/blenderbim/blenderbim/bim/module/spreadsheet_writer/operator.py
C-Claus: I thought about completely removing the xlsx functionality, since this is not really an open-source and free solution.

src/blenderbim/blenderbim/bim/module/spreadsheet_writer/operator.py
```diff
@@ -0,0 +1,652 @@
...
+        attrs = e.attrs
+        rows = int(attrs.get("table:number-rows-repeated", 1))
+        if "table:visibility" in attrs.keys():  # If the key is here, we can assume it's hidden (or can we ?)
+            #for row_idx in range(rows):
+            #    yield row + row_idx
+            yield from range(row, row + rows)
+        row += rows
+
+        
+
```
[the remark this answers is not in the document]
C-Claus: I used VS Code with the Blender extensions, but I think I accidenlty made changes in notepad ++

Moult:
Sorry it took so long to take a look at this. There seem to be a lot of hardcoded properties. This is something we try to avoid to allow for easy support of multiple schemas. Is it possible to either refactor this, or perhaps even merge the functionality into IfcCSV? I would expect the feature to be very similar?

C-Claus:
Thanks for replying,  I agree. I think it should completeley be refactored from scratch. I wrote this already more than a year ago and I've learned a lot since then.  The functionality is very similar to IFCCSV with the note a filtering from a spreadsheet is possible. I am also having doubt if I should keep the xsls file support and only focus on the .ods file as output to keep it a more open source-solution.

Moult:
Very well, let's leave this open until we add ODS supports to IFCCSV, so that at least we have a slightly similar feature set so users don't miss out on this useful function.

C-Claus:
For my understanding, BlenderBIM comes with the pandas and odf modules installed? so technically I could develop the spreadsheet writer ootb and integrate it with ifccsv?

PR #2627, opened 2022-12-01
Mapping quantities (see #2533)

maxfb87:


src/blenderbim/blenderbim/bim/module/pset/qto_calculator.py
```diff
@@ -1063,6 +1081,16 @@ def get_object_main_axis(self, o):
         else:
             return "x"
 
+    def is_opening_horizontal(self, o):
+        x = (Vector(o.bound_box[4]) - Vector(o.bound_box[0])).length
+        y = (Vector(o.bound_box[3]) - Vector(o.bound_box[0])).length
+        z = (Vector(o.bound_box[1]) - Vector(o.bound_box[0])).length
+
+        if z < x and z < y:
```
Amoki, line 1089:
```python
return z < x and z < y
```
😇

maxfb87: What? :-)

aothms: @Amoki means `if a: return True; else: return False` is the same as `return a`.

maxfb87: Ahh now i see, thanks! I looked at it in a hurry and i didn't see the difference!

PR #2642, opened 2022-12-09
Remove unecessary line which make build failing

CyrilWaechter:


CyrilWaechter:
From my understanding. Specifying python version is not necessary as there is nothing run with python in steps.

For a fix rather than a complete removal, see https://github.com/actions/setup-python/issues/555

PR #2646, opened 2022-12-12
action container image: bump python version

fbpyr:
bumping python version, so it works with our ubuntu latest image and `actions/setup-python@v2`

@Moult I noticed the action error and thus choco upgrades were not coming in anymore. 
the above should fix it. 🙂


fbpyr:
for context to the issue: https://github.com/actions/setup-python/issues/555

PR #2662, opened 2022-12-25
Feature: express rules

aothms:
🎅 happy holidays everybody 🎅

This pull request gives us preliminary support in Python for evaluating rules, functions and derived attributes defined in the express schema.

Following our elaborate testing strategy. The behaviour and semantics of the generated rule code has been tested on 104 specifically crafted minimal .ifc files.

It is implemented as a code generation process taking the express ast and rewrite it to python. This works surprisingly well up to some extent:

### query expressions

E.g query expressions are appropriately mapped to python list comprehensions.

```
HasAdvancedFaces : SIZEOF(QUERY(Afs <* SELF\IfcManifoldSolidBrep.Outer.CfsFaces |
  (NOT ('IFC4X3_TC1.IFCADVANCEDFACE' IN TYPEOF(Afs)))
)) = 0;
```

```python
assert (sizeof([afs for afs in self.Outer.CfsFaces if not 'ifc4x3_tc1.ifcadvancedface' in typeof(afs)])) == 0
```

### aggregate literals

List literals are also fairly equivalent:

```
V : IfcDirection
  := IfcRepresentationItem() || IfcGeometricRepresentationItem () || IfcDirection([1.,0.]); 
```

```python
v = IfcDirection(DirectionRatios=[1.,0.])
```

Although the repetition clause needed to be represented using a different syntax:

```
Res  := IfcRepresentationItem() || IfcGeometricRepresentationItem () || IfcDirection([0.0:Ndim]);
```

```python
res = IfcDirection(DirectionRatios=([0.0] * ndim))
```

(notice how the complex entity construction is resolved to a simple construction using named parameters by the schema knowledge embedded into the code generation process)

### static vs dynamic typing

Another complexity was in static typing. The typing on locals was mostly eliminated, but when assigned an aggregate literal we need to assign the correct type:

```
Names : SET OF IfcIdentifier := [];
```

```python
names = express_set([])
```

### inverse roles

Also ifcopenshell provided great fundamentals. The `usedin()` function was easily implemented using `ifcopenshell.file.get_inverse(with_attribute_indices=True)` to easily filter on the roles an entity participates in.

## Complexities

At the same time there were quite a few complexities regarding semantics that made it somewhat harder to use python as the base language for implementing an express rule verifier.

### One based indexing

This was resolved by postfixing every index_qualifier with `- 1`. We didn't simply subtract one during the code generation process because in rules and function indexing can also happen based on variable names (e.g `i` in a loop).

```
MinutesInRange : ABS(SELF[2]) < 60;
```

```python
assert (abs(self[2 - 1])) < 60
```

### Out of bound indexing

Even more difficult was the fact that out of bound indexing in express simply returns indeterminate, whereas in Python an exception is raised. This is actually hard to address because it affects the behaviour of builtin python classes and we also don't want to dramatically affect the behaviour of core IfcOpenShell. Therefore only a workaround is presented when coupled with the exists() function, to defer evaluation of the arguments by wrapping them in a lambda and catching any potential index error in the implementation of the exists() function.

```
WR35 : (
  EXISTS(SELF\IfcPropertySetDefinition.DefinesType[1])) 
  AND 
  ('IFC4X3_TC1.IFCDOORTYPE' IN TYPEOF(SELF\IfcPropertySetDefinition.DefinesType[1])
);
```

```
assert (exists(lambda: self.DefinesType[1 - 1])) and ('ifc4x3_tc1.ifcdoortype' in (typeof(self.DefinesType[1 - 1])))
```

The fact that boolean operators short-circuit (i.e the second component of the `and`-clause is not evaluated when the first component returns False means that no exception in python will be thrown

### Case insensitivity

Case insensitivity is implemented simply by converting all variable cases to lowercase, which is not a fully general solution, but in working state for the rules in the schemas.

# NB: Breaking change!

Note that this PR results into one breaking code change. Derived attributes are now supported, this is more of an addition: attribute names that previously didn't **exist** now do. However, express also allows redeclaring inherited attributes as derived (the `*` in .ifc spf files). In that case previously getattr returned None in IfcOpenShell, but now returns the properly derived attribute, e.g:

```
>>> import ifcopenshell
>>> f = ifcopenshell.file()
>>> f.createIfcSIUnit(UnitType='LENGTHUNIT', Name='METRE') 
#1=IfcSIUnit(*,.LENGTHUNIT.,$,.METRE.)
>>> f[1].Dimensions
#0=IfcDimensionalExponents(1,0,0,0,0,0,0)
```

🎅 happy holidays everybody 🎅


Moult:
This is great! Is there anything special which needs to be done or is calling ifcopenshell.validate() like before still sufficient?

aothms:
There is a new keyword argument to that function to enable. Some testing
would be very helpful.

Sent from a mobile device. Excuse my brevity. Kind regards, Thomas

Op do 5 jan. 2023 08:40 schreef Dion Moult ***@***.***>:

Moult:
Is there a way to make pytest an optional dependency? The only function used is _pytest.assertion.rewrite. I'd prefer not to have to bundle pytest with the BlenderBIM Add-on and still let users easily validate models.

aothms:
Without pytest's assertion rewriting, the only thing you'd see is simply
"assertionerror". With rewriting it's also not perfect, but at least quite
improved. It will be replaced by sth custom at some point. Feel free to
make it optional.

Sent from a mobile device. Excuse my brevity. Kind regards, Thomas

Op do 5 jan. 2023 10:55 schreef Dion Moult ***@***.***>:

PR #2672, opened 2023-01-03
Delete unuseful functions and improve speed

maxfb87:
Using PsetProperties.qto_name for retrieve all available object pset_qto, performance is better and the code is clearer.

### DIFF
--- a/src/blenderbim/blenderbim/core/qto.py
+++ b/src/blenderbim/blenderbim/core/qto.py
@@ -24,10 +24,12 @@ def calculate_circle_radius(qto, obj=None):
 
 def assign_pset_qto(qto, selected_objects):
     for obj in selected_objects:
+        qto.set_active_object(obj)
         qto.assign_pset_qto_to_selected_object(obj)
 
 def calculate_all_qtos(qto, selected_objects):
     for obj in selected_objects:
+        qto.set_active_object(obj)
         if not qto.get_pset_qto_object_ifc_info(obj):
             print(f"There is no pset qto instance associated to object {obj.name}")
             continue
--- a/src/blenderbim/blenderbim/core/qto.py
+++ b/src/blenderbim/blenderbim/core/qto.py
@@ -24,7 +24,7 @@ def calculate_circle_radius(qto, obj=None):
 
 def assign_pset_qto(qto, selected_objects):
     for obj in selected_objects:
-        qto.assign_pset_qto_to_selected_object(obj)
+        qto.assign_base_qto_to_object(obj)
 
 def calculate_all_qtos(qto, selected_objects):
     for obj in selected_objects:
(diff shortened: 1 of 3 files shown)


Moult:
I don't fully understand this sorry, why are you doing `qto.set_active_object(obj)`?

maxfb87:
Hi @Moult , i'm happy you asked me because i had questions regarding this.
There are two aspects:
- before this commit, i didn't use set active object but now i use it because if i want to use obj.PsetProperties.qto_name i have to set the active object.
I want to use obj.PsetProperties.qto_name because it is faster than the tool function i used before this commit
- i created this tool function made by only one line of code because qto.py, in core folder, can't has import and set active object needs import bpy. Have i understood it right?

Moult:
What information are you trying to retrieve? `obj.PsetProperties.qto_name` can hold any arbitrary qto_name, not necessarily the one you're after. Perhaps there is a better way? Can you describe exactly what logic you're trying to implement in pseudocode / english?

I'd prefer not to set the active object since that effectively changes the Blender scene which could lead to very slow operations.

Moult:
Yeah, better not to use qto_name because it may not hold Qto_WindowBaseQuantities, it may hold something like My_Custom_Qto. `psets = blenderbim.bim.schema.ifc.psetqto.get_applicable(ifc_class, qto_only=True)` is the proper way to get it. Is that function too slow for you? I'm surprised because it has `lru_cache()` which should make it cache results and there aren't that many IFC classes out there.

maxfb87:
Ok @Moult , i'll go with that!
Nono, this function should be fast, i was referring to my old function that was slow :-)

And what about second point above? Is it correct?

Moult:
Yes, the core should only have calls to tools, and should never do any logic directly. This allows the core to be testable agnostic of Blender and serve only to keep track of function resource and domino effects.

maxfb87:
@Moult thanks for explanation. I removed the the active_object function and cleaned around...

Moult:
Two more comments, can you rename from `assign_pset_qto_to_selected_object` to just `assign_base_qto_to_object` since that's what it's doing, it's not doing psets, and it's also not doing selected objects. Also, `get_applicable_names(ifc_class, qto_only=True)[0]` is assuming that Qto_FooBaseQuantities is the 0th index in that list, it may not be if the user has defined custom qtos. So it needs to be changed to something like [q for q in get_applicable_names(ifc_class, qto_only=True) if "Qto_" in q and "Base" in q] I think. Currently get_applicable_names does not let you isolate bS official qtos, maybe we need to add that as a feature. Otherwise feel free to merge :)

maxfb87:
Ook @Moult ... but i don't understand, can `blenderbim.bim.schema.ifc.psetqto.get_applicable_names(ifc_class, qto_only=True)` return more than one value? Doesn't it get data from the official schema? And the schema supposed to specify only one official Qto...no?

Moult:
It gets any applicable that are in your list of property set templates https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/util/pset.py#L39-L41 by default only the bS ones are loaded, but the user might have more.

maxfb87:
@Moult ah ok... so what about if i use directly
`ifcopenshell.util.pset.PsetQto('IFC4').get_applicable_names(ifc_class, qto_only=True)[0]`
instead of
`blenderbim.bim.schema.ifc.psetqto.get_applicable(ifc_class, qto_only=True)[0]` ?
This should be always only one value and takes the value directly from the schema (that is what i wanted to do). What do you think?

Moult:
I'm not convinced, would that create a new instance of the PsetQto class each time? What would that do to lru_cache?

Moult:
Cheers. lru_cache is a Python thing where repeated calls with the same arguments will return the same result, so it makes expensive functions less expensive :)

PR #2680, opened 2023-01-08
Fix #2631. Add support for flange edge radius and flange slope for I-shape

### LINKED ISSUE #2631: FlangeEdgeRadius and FlangeSlope for IfcIShapeProfileDef
Looks like flange rounding and sloping is not working as expected...

https://community.osarch.org/discussion/1272/flangeedgeradius-and-flangeslope-for-ifcishapeprofiledef

Moult:


Moult:
I forgot to update the `dy1 < ALMOST_ZERO || dy2 < ALMOST_ZERO` check, so now I think it should be good to go :)

src/ifcgeom/IfcIShapeProfileDef.cpp
```diff
@@ -65,8 +93,25 @@ bool IfcGeom::Kernel::convert(const IfcSchema::IfcIShapeProfileDef* l, TopoDS_Sh
 		IfcGeom::Kernel::convert(l->Position(), trsf2d);
 	}
 
-	double coords[24] = {-x1,-y, x1,-y, x1,-y+dy1, d1,-y+dy1, d1,y-dy2, x2,y-dy2, x2,y, -x2,y, -x2,y-dy2, -d1,y-dy2, -d1,-y+dy1, -x1,-y+dy1};
-	int fillets[4] = {3,4,9,10};
-	double radii[4] = {f1,f2,f2,f1};
-	return util::profile_helper(12,coords,(doFillet1||doFillet2) ? 4 : 0,fillets,radii,trsf2d,face);
+	double coords[24] = {
```
aothms, line 96: I had to quintuple check. I made the following diagram.

afbeelding

But I think `-y` should always be accompanied with `+ft1` and `y` should always be accompanied with `-ft2`. `-y` is the bottom flange and should always be paired with bottom flange thickness `ft1`.

src/ifcgeom/IfcIShapeProfileDef.cpp
```diff
@@ -65,8 +93,25 @@ bool IfcGeom::Kernel::convert(const IfcSchema::IfcIShapeProfileDef* l, TopoDS_Sh
...
+		x2,y,
+		-x2,y,
+		-x2,y-ft2+dy2,
+		-d1,y-ft2-dy1,
+		-d1,-y+ft2+dy1,
+		-x1,-y+ft2-dy2
+	};
+	int fillets[8] = {2,3,4,5,8,9,10,11};
+	double radii[8] = {fe1,f1,f2,fe2,fe2,f2,f1,fe1};
+#ifdef SCHEMA_IfcIShapeProfileDef_HAS_FlangeEdgeRadius
```
aothms, line 112: It's a minor thing, but I'd prefer to have `doFlangeEdgeRadiu`s globally available and defaulting to false. That way this last ifdef is not necessary. That's the same approach we currently have with `doFillet2`.

aothms (changes requested):
Thanks for this, see comments :)

src/ifcgeom/IfcIShapeProfileDef.cpp
Moult: Are you sure? `-y` is half the overall depth, and so e.g. the first coordinate `-x1,-y,` should represent the bottom left coordinate, I don't see why there should be `+ft1`?

2023-01-09-083711_833x676_scrot

aothms: Ok, sorry, "always accompanied" unnecessarily confusing, what I meant, but didn't clearly articulate -y likely should never incur a +ft2 only +ft1.

```
       vvv
 x1,-y+ft1-dy2,
 d1,-y+ft1+dy1,
...
-d1,-y+ft2+dy1,
-x1,-y+ft2-dy2
       ^^^
```

Moult:
Cool should be all fixed now. By the way, the flange thickness +-dy1 is an assumption. I could not find any clear diagram explaining how to interpret the flange thickness. This is the current implementation, where the flange thickness is actually measured at the very center, see image:

2023-01-09-102757_756x509_scrot

This is probably wrong, but I have no idea what is right? Maybe @Jesusbill might know as a structural engineer?

Jesusbill:
@Moult pretty sure it refers to the thickness at the edge of the flange and not at the center, like the parameter `t` in this [link](http://www.b2bmetal.eu/ipn-inp-beams-european-standard-universal-steel-i-beams-ipn-section--flange-slope-14--properties-dimensions-specifications-din-1025-1-1995)

Moult:
Yes, that also makes sense to me. I've adjusted to follow that definition. Kinda weird how for U Shape profiles IFC calls the middle point the flange thickness.

Moult:
Bump?

aothms:
Yes I had the same question in mind. For U-shape it's also not explicitly mentioned in the docs where the thickness is measured, but just somewhat unclearly drawn in the middle in the diagram. Should we try and clarify this in the spec also?

PR #2694, opened 2023-01-14
fix: ifcclash bcfxml export

sanzoghenzo:
this will hopefully fix the issue reported on [OSArch forum](https://community.osarch.org/discussion/comment/14064/#Comment_14064).

I only unit-tested the new methods in the bcf library, I don't know how to unit-test ifcclash.

I've noticed that the logic used to create the camera direction and up vectors is a bit convoluted, and the rototranslation matrices are useless since we already calculate the vectors, so I commented out that part; but I may be missing something here, a real world scenario would uncover issues on that.

### DIFF
--- a/src/bcf/requirements-dev.txt
+++ b/src/bcf/requirements-dev.txt
@@ -3,5 +3,5 @@ mypy
 pylint
 isort
 xsdata
-tox
+tox==3.27.1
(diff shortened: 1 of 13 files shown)


Moult:
Great work! Thanks!

PR #2696, opened 2023-01-14
Fix little bug

maxfb87:
Fix little bug @myoualid


myoualid:
My bad, it seems I double named an operator as "bim.calculate_all_qtos" in the pset module, and it may have ovverriden the operator from the qto module. 
commit [b41878e](https://github.com/IfcOpenShell/IfcOpenShell/commit/b41878e960e499a6b2118180588be766b5555f56) should fix it

PR #2699, opened 2023-01-15
adjust attribute type comparison character and characterization

c4rlosdias:
Ifc Selector does not work in comparisons with numbers and boolean properties, I propose to mischaracterize the attribute type in string and leave this attribution up to the user, using quotation marks in the case of string attributes. in addition, I included 'f.comparison' in the query because it is always using the equals comparator.


Moult:
Only one line of code changed, is that sufficient? Shouldn't more code also change?

c4rlosdias:
it was enough.  maybe in the future do the automatic identification of the type of the property to insert the quotes automatically.  but at the moment, this modification is enough, as long as the user inserts the quotes when dealing with a string. Thank you

Issue #2690, opened 2023-01-12
Run IfcConvert on sample files in CI/CD


aothms:
Previously we had this in travis.yml

```
  - |
    (for i in *.ifc; do \
    echo $i | tee -a log; \
    timeout 1m /usr/local/bin/IfcConvert -yv "$i" "$i.dae" --validate >> log 2>&1; \
    echo $i $? >> statuses; \
    done) || true
  - echo Failed
  - grep -v 0$ statuses
  - grep -v 0$ statuses | wc -l
  - echo Succeeded
  - grep 0$ statuses
  - grep 0$ statuses | wc -l
```

it wasn't perfect in the sense that we wouldn't get notified when there were regressions, but still it's a nice metric to have.

Li1506:
Any documentation on how IfcConvert build process works?

aothms:
IfcConvert is built and installed in this step https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/.github/workflows/ci.yml#L61-L87

PR #2739, opened 2023-02-05
add IfcConvert testing sample files step to ci.yml

Li1506:
#2690

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,20 @@ jobs:
           sudo make -j $(nproc)
           sudo make install
 
+      - name: Run IfcConvert on Sample files
+        run: |
+          (find test/input src/blenderbim/test/files -name '*.ifc' | while read i; do \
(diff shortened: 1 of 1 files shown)


aothms:
Thanks. It currently says:

> No files found matching pattern '*.ifc'

Probably the pattern needs to be adapted to `test/input/*.ifc` or you use `find . -name '*.ifc'` to recursively search.

https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/4096233236/jobs/7064367618#step:6:22

Moult:
FYI there are blenderbim specific test files in src/blenderbim/test/files :)

aothms:
Thanks this is great. Seems there are quite a few files that are failing now. Would be cool if in the future this is extended to upload the data as a github actions artefact https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts and maybe catch more information from the logs. Just a not for future reference :)

Issue #2756, opened 2023-02-10
Error trying to reassign the ifc class to an object


maxfb87:
I get an error when trying to reassign the ifc class to an object.
To recreate the bug, simply start a new fresh file, add an ifc project, add a cube, assign an ifc class to it (IfcWall for example) and after that try to reassign the ifc class to the cube in blender UI.
This is the error i get:

Traceback (most recent call last):
  File "/home/max/.config/blender/3.3/scripts/addons/blenderbim/bim/module/root/operator.py", line 91, in execute
    return IfcStore.execute_ifc_operator(self, context)
  File "/home/max/.config/blender/3.3/scripts/addons/blenderbim/bim/ifc.py", line 425, in execute_ifc_operator
    result = getattr(operator, "_execute")(context)
  File "/home/max/.config/blender/3.3/scripts/addons/blenderbim/bim/module/root/operator.py", line 100, in _execute
    product = ifcopenshell.api.run(
  File "/home/max/.config/blender/3.3/scripts/addons/blenderbim/libs/site/packages/ifcopenshell/api/__init__.py", line 66, in run
    result = usecase_class(ifc_file, **settings).execute()
  File "/home/max/.config/blender/3.3/scripts/addons/blenderbim/libs/site/packages/ifcopenshell/api/root/reassign_class.py", line 65, in execute
    element = ifcopenshell.util.schema.reassign_class(
  File "/home/max/.config/blender/3.3/scripts/addons/blenderbim/libs/site/packages/ifcopenshell/util/schema.py", line 81, in reassign_class
    inverse_pairs = ifc_file.get_inverse(element, allow_duplicate=True, with_attribute_indices=True)
TypeError: file.get_inverse() got an unexpected keyword argument 'with_attribute_indices'
Error: Python: Traceback (most recent call last):
  File "/home/max/.config/blender/3.3/scripts/addons/blenderbim/bim/module/root/operator.py", line 91, in execute
    return IfcStore.execute_ifc_operator(self, context)
  File "/home/max/.config/blender/3.3/scripts/addons/blenderbim/bim/ifc.py", line 425, in execute_ifc_operator
    result = getattr(operator, "_execute")(context)
  File "/home/max/.config/blender/3.3/scripts/addons/blenderbim/bim/module/root/operator.py", line 100, in _execute
    product = ifcopenshell.api.run(
[...]

PR #2757, opened 2023-02-10
Fix bug #2756

maxfb87:
Maybe i fixed this bug #2756 , but @Moult  can you review to check?


Moult:
Your fix works but hardcodes the solutions to just predefined types whereas in theory this bug could occur for any incompatible enumeration. I've committed a more generic fix :)

maxfb87:
Yeah thanks, this helped my also to better understand the feauture testing and the schema util functions :-)

PR #2854, opened 2023-03-11
add a ci workflow that catches and restarts failed conda builds due to 143…

Krande:
… errors. And minor cleanup of unused variables in conda meta.yaml

### DIFF
--- a/.github/workflows/ci-restart-failed-daily-conda-builds.yml
+++ b/.github/workflows/ci-restart-failed-daily-conda-builds.yml
@@ -2,9 +2,12 @@ name: Restart failed daily conda builds
 
 on:
   workflow_run:
-    workflows: [ci-ifcopenshell-conda-daily]
+    workflows: [ ci-ifcopenshell-conda-daily ]
     types:
       - completed
+  push: # Remove this before merging
+    branches:
+    - add-conda-daily-restart-workflow
 
 env:
   REPO_OWNER: IfcOpenShell
@@ -15,19 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event.workflow_run.conclusion == 'failure' }}
     steps:
-    - name: Check for failed runs
-      id: check-failed-runs
-      run: |
-        echo "Checking for failed runs..."
-        # Replace "my-workflow" with the name of your workflow
-        failed_runs=$(curl -s https://api.github.com/repos/${{ env.REPO_OWNER }}/REPO/actions/runs?status=completed&conclusion=failure&workflow_id=${{ env.MY_WORKFLOW }} | jq '. | map(select(.status.code == 143))')
-        echo "Found $(echo $failed_runs | jq length) failed runs"
-    - name: Restart failed runs
-      if: steps.check-failed-runs.outputs.failed_runs != null
-      run: |
-        echo "Restarting failed runs..."
-        for run_id in $(echo $failed_runs | jq -r '.[].id'); do
-          echo "Restarting run $run_id..."
-          curl -s -X POST https://api.github.com/repos/${{ env.REPO_OWNER }}/REPO/actions/runs/$run_id/rerun \
-            -H "Accept: application/vnd.github+json"
-        done
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
+      - name: Check for failed runs
+        id: check-failed-runs
+        run: |
+          python check_repo_ci_jobs.py
--- a/.github/workflows/ci-restart-failed-daily-conda-builds.yml
+++ b/.github/workflows/ci-restart-failed-daily-conda-builds.yml
@@ -5,12 +5,9 @@ on:
     workflows: [ ci-ifcopenshell-conda-daily ]
     types:
       - completed
-  push: # Remove this before merging
-    branches:
-    - add-conda-daily-restart-workflow
 
 env:
-  REPO_OWNER: IfcOpenShell
+  REPO_OWNER: IfcOpenShell/IfcOpenShell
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-restart-failed-daily-conda-builds.yml
```diff
@@ -0,0 +1,33 @@
...
+  restart-failed-runs:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    steps:
+    - name: Check for failed runs
+      id: check-failed-runs
+      run: |
+        echo "Checking for failed runs..."
+        # Replace "my-workflow" with the name of your workflow
+        failed_runs=$(curl -s https://api.github.com/repos/${{ env.REPO_OWNER }}/REPO/actions/runs?status=completed&conclusion=failure&workflow_id=${{ env.MY_WORKFLOW }} | jq '. | map(select(.status.code == 143))')
```
aothms, line 23: Is `/REPO` correct here?

.github/workflows/ci-restart-failed-daily-conda-builds.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  restart-failed-runs:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    steps:
+    - name: Check for failed runs
+      id: check-failed-runs
+      run: |
+        echo "Checking for failed runs..."
+        # Replace "my-workflow" with the name of your workflow
```
aothms, line 22: Remove I guess

aothms (commented):
Is this somehow safe against infinite loops?

.github/workflows/ci-restart-failed-daily-conda-builds.yml
Krande: I think so. It at least mangaes to find the failed jobs on my [test repo](https://github.com/Krande/condapackaging/actions/workflows/ci-restart-failed.yml) for conda packaging repo

.github/workflows/ci-restart-failed-daily-conda-builds.yml
Krande: Regarding your comment on recursiveness. You are right. If the ci job for some reason continues to produce the same reoccurring 143 error this will run forever. 

How about I comment out the "Restart failed runs" and let it run the "check-failed-runs" only for a while to ensure that it's safe before we use it properly? That way this ci only runs once and it's more in "monitoring mode" only for a while to see that it reaches the right conclusions in the daily jobs.

To be honest I haven't really gotten to test this ci workflow properly yet since the runs on my test repo never (or rarely) manages to run into the 143 error :( But the daily ci conda build seems to always run into arbitrary 143 errors. So I thought it might be a good idea to test this out here? 

While we run this in a "monitoring mode" I can find a way to stop any recursiveness by ensuring that any ci build is only retried once?

Krande:
Regarding your comment on recursiveness. You are right. If the ci job for some reason continues to produce the same reoccurring 143 error this will run forever.

How about I comment out the "Restart failed runs" and let it run the "check-failed-runs" only for a while to ensure that it's safe before we use it properly? That way this ci only runs once and it's more in "monitoring mode" only for a while to see that it reaches the right conclusions in the daily jobs.

To be honest I haven't really gotten to test this ci workflow properly yet since the runs on my test repo never (or rarely) manages to run into the 143 error :( But the daily ci conda build seems to always run into arbitrary 143 errors. So I thought it might be a good idea to test this out here?

While we run this in a "monitoring mode" I can find a way to stop any recursiveness by ensuring that any ci build is only retried once?

Krande:
Yeah, so I see that I might have been a bit premature in creating this PR. 

So I just converted the code to a python script and although I am not quite done, it seems to be a bit closer to something that is both easier to read and also should be easier to control. I also see that I would like to re-run just the failing matrix builds not the entire ci-run. So I will try to add those changes as well in the "restart" step also.

However, we might consider moving this particular script to another repo (like say the IfcOpenBot) to not put too much ci-stuff in the main repo? I see in my printed logs that the 2 endpoints that query the logs and tries to restart needs admin access. This can solved in 2 ways, 1) run it from this repo or 2) we include a bearer token in the header in a script from a different repo.

```
{'message': 'Must have admin rights to Repository.', 'documentation_url': 'https://docs.github.com/rest/reference/actions#download-workflow-run-attempt-logs'}
restarting job 4389411890
{'message': 'Must have admin rights to Repository.', 'documentation_url': 'https://docs.github.com/rest/reference/actions#re-run-a-workflow'}
```

I am fine if you'd prefer we close this PR and redo it in a different repo or if you'd like me to test this more in-depth a bit more before redoing this? If not, I'll finalize this tomorrow and we can test it out in the coming days.

Let me know what you think :)

Krande:
Hey @aothms! In my latest [commit](https://github.com/IfcOpenShell/IfcOpenShell/pull/2854/commits/2e111badb90cd9d68e1562f7341adcae12337222) I've made sure that the workflow has been tested so that it prevents infinite loops. It is also possible to test this locally :) 


This does however, not necessarily address the root issue of why the jobs are failing in the first place. But this will at least give us a second chance in shutdowns due to VM's running out of resource.

I have added two flags it should look out for in the logs from the individual runs:

* `fatal error C1060: compiler is out of heap space` (on windows)
* `Error: The operation was canceled.` (on linux)

I've also made sure that it only re-runs the failed jobs, not the entire workflow.

aothms:
Ok, thanks for your continued effort on this. I will merge this and let's see how it goes!

Issue #2758, opened 2023-02-11
Weight qto is not calculated based off material MassDensity


arunkvijay:
Hi,
I am trying to define the `MassDensity` in `PSet_MaterialCommon` and assign the same to `IfcMaterial` defined in my ifc file. My goal is to get the GrossWeight and NetWeight Properties through, say for ex., `Qto_ColumnBaseQuantities`.
Currently, I am able to assign density of the material, however, while populating the Ifc Object Quantity Sets, the weights are not calculated as is shown as 0.0

vulevukusej:
Hey @arunkvijay I don't think weight is calculated yet, right @maxfb87 ?

maxfb87:
Yes, weight is not calculated afaik.
At the moment, weight is just a property, independent from density and volume.
I"m still thinking about how to create custom properties defined by a rule (for example weight = density x volume)

vulevukusej:
Shouldn't be too difficult though, you'd just to check that the units are compatible or convert accordingly.

maxfb87:
There is something i don't know how to approach... for example, have i to define a new pset with the custom properties(something like "DerivedPset")? Also, i would define every derived property with 3 columns: name, rule and value but i don't know if this is correct, maybe there is something better... i don't know, i'm still confused :-/

arunkvijay:
I was referring to this [post](https://forums.buildingsmart.org/t/confusing-documentation-on-ifcmaterialproperties/2253/3) by Moult on a similiar issue. Here the `IfcMaterialProperties` from `Pset_MaterialCommon` do appear to be linked to `ifcMaterial` and then to the element through `IfcRelAssociatesMaterial`

As an example, the [Qto_SlabBaseQuantities](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/Qto_SlabBaseQuantities.htm) has 'IfcQuantityWeight` defined as 
> It is normally derived from the physical properties of the element under the specific measure rules given by a method of measurement

similar to any other `IfcQuantityLength` or `IfcQuantityArea` or `IfcQuantityVolume`

I am also confused on this and not sure if my understanding is correct too.

maxfb87:
So, i'm not an expert but i tried to better understand the issue.
@arunkvijay: like what is written in the post you linked, a material can have properties, like an object can have properties.
[...]

PR #2859, opened 2023-03-12
Now qto calculator calculates also gross weight and net weight (see #2758)

maxfb87:


src/blenderbim/blenderbim/bim/module/pset/qto_calculator.py
```diff
@@ -472,6 +472,74 @@ def get_obj_decompositions(self, obj):
...
+        if material.is_a('IfcMaterialProfileSetUsage'):
+            material_profiles = material.ForProfileSet.MaterialProfiles
+            if len(material_profiles) == 1:
+                obj_mass_density = self.get_single_material_mass_density(material_profiles[0].Material)
+                return obj_mass_density
+            else:
+                return
+
+
+    def get_single_material_mass_density(self, material):
```
Moult, line 534: ifcopenshell.util.element.get_pset?

maxfb87: ok... look at #2861
Is it better?

Issue #2853, opened 2023-03-10
IfcSverchok - No module named 'sverchok.utils.logging' 


andrsbtrg:
Hello, I tried installing IfcSverchok and I had this message when trying to activate the addin.

I have BlenderBIM and Sverchok installed.

![Screenshot from 2023-03-10 15-14-31](https://user-images.githubusercontent.com/63083862/224441727-b1f0d36d-eb9e-484d-8530-2c0a4eac4815.png)

I tried commenting out and this line:  https://github.com/IfcOpenShell/IfcOpenShell/blob/55036653dfdda0b58aae6f0cc608094de49a20c2/src/ifcsverchok/__init__.py#L32

`from sverchok.utils.logging import info, debug`

and the addin started.

I am running Fedora 37 and Blender 3.4

Moult:
This is a great first issue to debug by figuring out where is the logging used, and why it used to exist, and was it removed or renamed, and what happened to it. Would you like to try @andrsbtrg ? You may need to investigate in the upstream Sverchok codebase which moves pretty fast and check out their history.

andrsbtrg:
@Moult for sure! I will take a try on it.

PR #2860, opened 2023-03-13
Fix: ifc sverchok - no module named 'sverchok.utils.logging' #2853

andrsbtrg:
- This pull request updates the logger style in IfcSverchok to follow the latest logger style for Sverchok 1.2: https://github.com/nortikin/sverchok/pull/4891 

Related issue: https://github.com/IfcOpenShell/IfcOpenShell/issues/2853 would not let the add-on be enabled.

- Another issue related was disabling the add-on would throw a RuntimeError due to trying to unregister a class that may not be registered: 
image

A way around this for now was trying to catch the exception and continue Unregistering the rest of the classes.


Moult:
Thanks for investigating! It seems as though they say:

> For Sverchok extensions first option is especially useful since during reloading event sverchok can reload after an extension. It has sense to add its own sub logger logging.getLogger('sverchok.megapolis')

So should we do `sverchok.ifc` ?

Could you investigate why the unregister doesn't work? I'd prefer not to catch these types of errors which are potentially quite severe errors which represent packaging failures.

andrsbtrg:
Makes sense to register the logger as sverchok.ifc

I've investigated further and the SvIfcTooltip class is not registering correctlly, therefore when we try to unregister it returns this error:

`04:06:11.046 [ERROR] sverchok.ifc nodes/ifc/api.py:114 - Could not unregister class: unregister_class(...):, missing bl_rna attribute from 'RNAMeta' instance (may not be registered)`

I can confirm, when I attempt to not unregister that particular class it doesn't throw the error.

There is a class with the same name in api_WIP.py. The error be because it did already 'unregister' that class from the api_WIP module before. 

We could load only one of the api* modules or name the `SvIfcTooltip` to `SvIfcTooltipWIP`

Moult:
Thanks for the fix! I think we should ping @mdjska the original author for what to do with `api_WIP.py` - is it still needed or can it be deleted?

PR #2865, opened 2023-03-13
Add IFC4 support for functions of get_connected_to and get_connected_from in util.system

chunchk:
Due to the limitations of the original code base, support for IFC4 versions had been lacking. To address this, we have updated the relevant functions to better align with modern requirements. This new commit incorporates changes that represent a different approach to the problem.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/system.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/system.py
@@ -83,31 +83,34 @@ def get_connected_port(port):
 
 
 def get_connected_to(element):
-    from blenderbim.bim.ifc import IfcStore
-    ifcSchema = IfcStore.get_file().schema
     results = []
-    if ifcSchema == "IFC4":
-        for port in ifcopenshell.util.system.get_ports(element):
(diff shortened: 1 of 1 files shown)


Moult:
If I'm reading this right, doesn't that mean that it'll now break for IFC2X3?

chunchk:
I thought you want it to be IFC4 version only
but I think I can make it compatible with IFC2X3 as well

chunchk:
@Moult 
Will adding a condition be better? or is there any other way to get the IFC schema?

Moult:
I guess you can check the schema using `element.wrapped_data.file.schema`, then indeed use an if statement.

chunchk:
> 

Got it~
And I found something interesting, maybe I can further optimise it

aothms:
There's also this ugly approach to get schema name from instance (which doesn't depend on file), we should really streamline this. https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/entity_instance.py#L143

PR #2866, opened 2023-03-13
adding additional contexts

theoryshaw:


Moult:
Is this right? Shouldn't it be Model not Plan?

theoryshaw:
Not sure... see this convo
https://community.osarch.org/discussion/comment/14875/#Comment_14875

Moult:
Looking at this again, I don't think this is valid. They should use Model/Annotation/SECTION_VIEW or ELEVATION_VIEW to superimpose annotations on top of the body. Alternatively, Model/Body/SECTION_VIEW or ELEVATION_VIEW will replace the body.

theoryshaw:
The following contexts were already in the default model, when they got the error.
image

error: "No annotation context for drawing"

Adding the following contexts removed the error.
Not sure what's the right way, just wanted to share the current functionality. 

image

Moult:
Yeah they probably tried to add a

Moult:
You're absolutely right, I missed the original intention of the error report. This fix is still not appropriate, but thanks for pointing me in the right direction. The proper solution is to allow annotations to use 3D coordinates in non plan drawings.

Issue #2868, opened 2023-03-15
Python API little bug in material.edit_layer


cvillagrasa:
I was taking a look at the API and noticed the following:

According to the code example [in the material.edit layer documentation](https://blenderbim.org/docs-python/autoapi/ifcopenshell/api/material/edit_layer/index.html), the typical line:
`ifcopenshell.api.run("material.edit_layer", model, layer=layer, attributes={"LayerThickness": 13})`

doesn't include a `material=layer.Material` argument, which makes `layer.Material = None` after that call [(link)](https://github.com/IfcOpenShell/IfcOpenShell/blob/cce0867f6365288aa2a0c1963ceee610f96b14d3/src/ifcopenshell-python/ifcopenshell/api/material/edit_layer.py#L66). On the other hand, it wouldn't seem necessary to call `material.edit_layer` if a thickness argument could be passed to `material.add_layer`. That's also in line with the behaviour of `material.add_profile` accepting the profile, without the need to call `material.edit_profile` at first creation.

Am I missing something or it would be a good thing to add more arguments to `material.add_layer`? Perhaps even IsVentilated and so on.

Moult:
Indeed it would better to only edit the material if the material is specified.

As for more arguments to add_layer, the general principle I followed was that "adding" would only include the bare minimum required for valid IFC data (and in some exceptions some really useful defaults like a Name). The rest would be relegated to an edit call. There's no real strong logic behind this, other than making it easier to start coding APIs as this is a learning journey, and having some arbitrary rules on where logic should live.

PRs welcome :)

PR #2869, opened 2023-03-15
Api material fix

cvillagrasa:
Fix to https://github.com/IfcOpenShell/IfcOpenShell/issues/2868

For the moment I've just added everything to material.add_layer, including optional parameters. @Moult do you think it's necessary to make two calls (add and edit) when knowing everything beforehand?

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/material/add_layer.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/material/add_layer.py
@@ -18,7 +18,8 @@
 
 
 class Usecase:
-    def __init__(self, file, layer_set=None, material=None):
+    def __init__(self, file, layer_set=None, material=None, thickness=1., ventilated=None, name=None, description=None,
+                 category=None, priority=None):
         """Adds a new layer to a layer set
 
         A layer represents a portion of material within a layered build up,
@@ -39,6 +40,20 @@ class Usecase:
         :type layer_set: ifcopenshell.entity_instance.entity_instance
         :param material: The IfcMaterial that the layer is made out of.
         :type material: ifcopenshell.entity_instance.entity_instance
+        :param thickness: The thickness of the material layer.
+        :type thickness: float
+        :param ventilated: Indication of whether the material layer represents an air layer (or cavity).
+        :type ventilated: bool
+        :param name: The name by which the material layer is known.
+        :type name: str
+        :param description: Definition of the material layer in more descriptive terms than given by attributes Name or Category.
+        :type description: str
+        :param category: Category of the material layer, e.g. the role it has in the layer set it belongs to
+            (such as 'load bearing', 'thermal insulation' etc.).
+        :type category: str
+        :param priority: The relative priority of the layer, expressed as normalised integer range [0..100].
+            Controls how layers intersect in connections and corners of building elements.
+        :type priority: int
         :return: The newly created IfcMaterialLayer
         :rtype: ifcopenshell.entity_instance.entity_instance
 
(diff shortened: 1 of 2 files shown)


Moult:
Isn't it easier just to call edit within the add usecase? And just use the same kwargs that way the signature is the same? This code also hardcodes the attribute names which makes it harder to handle different schemas.

cvillagrasa:
Mmm ok, so if I understand correctly, your logic conceives three (or at least the first two) argument types for the Python API:

- "Important" args, which are hardcoded into the signature (layer_set or layer and material).
- "Non-important" optional args, which are packed into an `attributes` dictionary, with its keys being called exactly as the corresponding IFC attribute.
- The ones which could fall in between, like `LayerThickness`. Since it's not optional, it could be hardcoded if considered from the first group; else it should be passed into `attributes`.

I don't know, there's something inside me that doesn't feel super comfortable with hardcoding a thickness of 1 on creation and then calling `material.edit_layer`.

There's also something that moves me towards treating every argument the same. Otherwise, the user is overloaded with having to know which arguments should be consulted from the bbim docs and which ones from the bSI docs.

In any case, I agree that hardcoding every parameter as in the current PR status is not ideal. Is it too crazy to go the other way around and have a generic signature like `file, IfcMaterialLayerSet, **kwargs`?

And then document the "important" arguments in the bbim docs (as they already are, but `Material` instead of `material`, stepping away from pythonic naming conventions in favour of IFC coherence). And maybe add a little note referring to the fact that of course several schemas exist, and the user should look up the corresponding `IfcMaterialLayer` bSI reference for the chosen one, which will always hold the truth about which attributes apply.

Then, the typical add call would be:
`ifcopenshell.api.run("material.add_layer", file, layer_set, Material=mat, LayerThickness=13)`

Or maybe a way to make it more pythonic would be to make use of the `attributes` dict:
`ifcopenshell.api.run("material.add_layer", file, layer_set, attributes={"Material": mat, "LayerThickness": 13})`

That way, it is crystal clear that the first two positional arguments refer to the creation of an `IfcMaterialLayer` within a given `IfcMaterialLayerSet` within a given file, and then the `attributes` dict refers 100% to the properties of that newly created layer. 

Am I being too revolutionary with such proposals for the existing codebase? 😂 These are just my thoughts written down, no problem with hardcoding only the thickness or keeping it into attributes, if you feel it's the better option.

cvillagrasa:
Interesting discussion! Maybe it's worth to open a dedicated issue and try to have as much feedback as possible. This PR could be merged with just the one-liner bug fix commit for now.

Firstly, I feel it's necessary to define what the API is and what it is intended for. If I was asked, I'd say that it's a means to use _ifcopenshell_ at a higher level than with the Python bindings of the C++ core. For instance, to avoid having to deal with _IfcRelationships_ and so on. But there are more differences than fewer lines of code, right? I recall something about the potential of using it as a REST API or something along those lines, and hence why the modules are being called by string. I always wonder if it's not possible to also **use dot notation** in order not to lose Intellisense. Ideally, I'd love to see some agreed alias in the Data Science way like `import ifcopenshell.api.run as ifc`, and then use it as `ifc.root.create_entity(file, ifc_class="IfcProject")`.

Referring now to the series of principles, for nº 1, 3, 4, 5 and 8, I'm thinking of a typical user who wants to make use of the API. Let's say they have some random knowledge of Python and IFC... but they're not a superexpert, so they start to write their script side by side with the API docs for reference. If there are mandatory args (let's call it **option A**), they can have a look at the docstring in the docs, understand it, populate all those args and voilà, they have a valid API call, fair enough. However, say they keep working on their script and now they find that they want to define some non-mandatory attribute, for instance the priority of a certain layer within a layerset. Are we expecting that they look for the bSI documentation of their schema version, find `IfcMaterialLayer`, browse its attributes until locating their exact spelling and explanation of their value, to then go back to the script and add it into the dict for the optional list of args? Maybe too low-level for a high level API. There's also some room for confusion on whether the mandatory args are exactly the non optional ones from the bSI docs.

So, **option B** could be to hardcode all the arguments into the signature, with the optionals defaulting to `None`, pythonic naming and a docstring with every possible attribute. Of course, this approach potentially creates the need to design three different APIs: for IFC 2x3, 4 and 4x3... as well as three doc websites, which would be a nightmare. Even with this amount of drawbacks, I don't think it's an absurd option at all, given the fact that the schema versions are written in stone as ISO standards and won't change... but yes, it's definitely the option which would require the most work.

Lastly, **option C** would consist of receiving all of the IFC args within the dict, regardless of whether they are mandatory or optional. In this scenario, we lose Intellisense (but we would only fully have it in option B), and some rules would need to be followed so as to generate a useful documentation and take advantage of bSI website, if properly referenced. For instance, keeping with the example, `material.add_layer` would have an `attributes` argument, annotated as a dictionary containing the attributes of the `IfcMaterialLayer` class (with link to bSI, at least IFC4 and IFC4x3). The mandatory _Material_ and _LayerThickness_ would be listed and explained, and a general note "for a complete list of attributes, consult bSI docs" would be added. In case there's some optional but highly useful in practice (principle 8), they could also be listed and explained within the docstring.

A variant C2 is to remove the dict, and pack every argument into `**kwargs`, like `matplotlib` does, but I like the fact that in base option C, the attributes dict seems to get along better with IFC's PascalCase, and they're grouped as what they are, instead of being at the same level of basic arguments like `file`.

For all these reasons, I vote for option C, then option B (definitely not being me who does all the hard work 😂), and lastly, option A.

cvillagrasa:
With respect to the rest of the points:

2) Agree. In addition, maybe it could be complemented with raising Exceptions from inside the Usecases when any unexpected state is reached.
6) Agree
7) With my current IFC knowledge, I know there's a global `IfcUnitAssignment` defining project units, and then an optional _Unit_ attribute in most `IfcProperty` subclasses, which overrides global units if exists. Are there more API calls which could accept custom units other than `pset.edit_pset`? With https://github.com/IfcOpenShell/IfcOpenShell/pull/2872, we're accepting both options, which I feel is a win-win. Where do you foresee that _is_si_ toggle?
9) Superagree
10) Sounds great

PR #2872, opened 2023-03-16
Allow for custom IfcUnits when editing psets

cvillagrasa:
I was wondering how to define reasonable global units in IfcUnitAssignment, while overwriting certain psets with other units. For instance, `Pset_MaterialMechanical -> YoungModulus` in _GPa_.

With this little proposal, previous code remains compatible, while it's also possible to:

`gpa = file.create_entity("IfcSIUnit", Dimensions=None, UnitType="PRESSUREUNIT", Prefix="GIGA", Name="PASCAL")`
`ifcopenshell.api.run("pset.edit_pset", file, pset=pset, properties={"YoungModulus": {"NominalValue": 27, "Unit": gpa}})`

@Moult @vulevukusej what do you think?

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
@@ -189,6 +189,7 @@ class Usecase:
         if prop.Name not in self.settings["properties"]:
             return
         value = self.settings["properties"][prop.Name]
+        unit, value = self.unpack_unit_value(value)
         if isinstance(value, list):
             sel_vals = []
             for val in value:
@@ -210,12 +211,15 @@ class Usecase:
             elif isinstance(value, ifcopenshell.entity_instance):
                 prop.EnumerationReference.EnumerationValues = value.EnumerationReference.EnumerationValues
                 prop.EnumerationValues = value.EnumerationValues
+        if unit:
+            prop.Unit = unit
         del self.settings["properties"][prop.Name]
 
     def update_existing_property(self, prop):
         if prop.Name not in self.settings["properties"]:
             return
         value = self.settings["properties"][prop.Name]
+        unit, value = self.unpack_unit_value(value)
         if value is None:
             if self.settings["should_purge"]:
                 del self.settings["properties"][prop.Name]
@@ -230,6 +234,8 @@ class Usecase:
             )
             value = self.cast_value_to_primary_measure_type(value, primary_measure_type)
             prop.NominalValue = self.file.create_entity(primary_measure_type, value)
+        if unit:
+            prop.Unit = unit
         del self.settings["properties"][prop.Name]
 
     def add_new_properties(self):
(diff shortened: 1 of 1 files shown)


cvillagrasa:
BTW I have no clue what's happening [in this line](https://github.com/IfcOpenShell/IfcOpenShell/blob/c9f52d2f8e29425ae51e4fd1accda5106ab8ae0c/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py#L248), since value is an ifcopenshell entity.

aothms:
> BTW I have no clue what's happening [in this line](https://github.com/IfcOpenShell/IfcOpenShell/blob/c9f52d2f8e29425ae51e4fd1accda5106ab8ae0c/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py#L248), since value is an ifcopenshell entity.

(

This is called (dictionary) unpacking https://python-reference.readthedocs.io/en/latest/docs/operators/dict_unpack.html

It's meaningless in this case and could be written as `Name=name, NominalValue=value`

But would allow things like control flow affect the keyword arguments you pass to a function:

```python
args = {'A': 1}
if something:
  args['B'] = 2
do_something(**args)
```

)

cvillagrasa:
Yeah sure! I know about dict unpacking. I mean that in that unpacking, `value` is an `ifcopenshell.entity_instance`, as per line 240 above, when it looks like it should be a plain value like a float or a string.

aothms:
Ah sorry, then, no you can't.

```
>>> import ifcopenshell
>>> f = ifcopenshell.file()
>>> f.createIfcPropertySingleValue(NominalValue=1)
  File "C:\Users\tkrij\miniconda3\envs\ifopsh07\lib\site-packages\ifcopenshell\file.py", line 285, in create_entity
    e[idx] = arg
  File "C:\Users\tkrij\miniconda3\envs\ifopsh07\lib\site-packages\ifcopenshell\entity_instance.py", line 240, in __setitem__
    self.method_list[idx](
  File "C:\Users\tkrij\miniconda3\envs\ifopsh07\lib\site-packages\ifcopenshell\ifcopenshell_wrapper.py", line 4923, in setArgumentAsEntityInstance
    return _ifcopenshell_wrapper.entity_instance_setArgumentAsEntityInstance(self, i, v)
TypeError: in method 'entity_instance_setArgumentAsEntityInstance', argument 3 of type 'IfcUtil::IfcBaseClass *'
>>> f.createIfcPropertySingleValue(NominalValue=f.createIfcInteger(1))
#1=IfcPropertySingleValue($,$,IfcInteger(1),$)
```

The reason is that IfcPropertySingleValue.NominalValue is a select type (IfcValue). In ifcopenshell select types are always handled through entity instances (even if they are only selected values of simple types).

cvillagrasa:
Ahh that was it, then. Thanks @aothms !

Moult:
This is great! Would you be able to add test cases for it please?

cvillagrasa:
Ok! I will add tests during this week

PR #2907, opened 2023-03-26
blenderbim: add more tool tips as part of issue 1646

rileywong311:
I added more tooltips to operators in the "IFC Project Info" and "IFC Clash Sets" panels. 
- I put the `bl_descriptions` variable under `bl_options` when possible. This is how other operators with tooltips have been formatted already.
- "Add Clash Source" (`bim.add_clash_source`) and "Export IFC" (`export_ifc.bim`) are tricky because they are used multiple times with different logic, so I wrote a generic tooltip that applied to both.


Gorgious56:
Re your second point @rileywong311  , operators can have a static description with `bl_description` but they can also have a dynamic tooltip with `description`: https://docs.blender.org/api/current/bpy.types.Operator.html#bpy.types.Operator.description

You can access the operator properties within the context of this method.

eg

```

class SimpleOperator(bpy.types.Operator):
    bl_idname = "object.simple_operator"
    bl_label = "Simple Object Operator"
    prop: bpy.props.StringProperty(default="My default arg")
    
    @classmethod
    def description(cls, context, properties):
        if properties.prop == "My default arg":
            return "My default tooltip"
        else:
            return "My other tooltip"

    def execute(self, context):
        return {'FINISHED'}
```

rileywong311:
@Gorgious56 I'll try that out!

Issue #2547, opened 2022-10-28
BlenderBIM: Temporal Section backface material


TZwielehner:
It is possible to assign a backface material in the shader nodes. That is very helpful to indicate element materials in sections/floorplans. Would be great if we could decide whether to override the cut section with one color or keep the shader settings.

Moult:
The temporary section tool works by using a backface shader, but things don't always look very good with it (e.g. non-manifold entourage, or overlapping / intersecting meshes). I can't wait until vanilla Blender supports a sketchup style viewport edge rendering.

TZwielehner:
I also use a backface shader but by material. Can the default color be overridden?

Moult:
The temporary section tool lets you define the colour yes, but it does not group by material.

Gorgious56:
Hello, I have a feeling this is possible but I don't exactly understand what you're asking for. 

Would you mind adding a diagram or a screenshot from your interface showing what you're after ?

TZwielehner:
AFK ATM, but just a quick explanation: when I'm in a drawing view I see the backface of my e.g. walls. To better distinguish between the wall types I made the backface of my brick walls red, the concrete walls green, insulation coverings pink etc.
In 3D view, when using the temporary section shader, everything is cut in red. So I lose that additional information (I want to try technical hatching, too, but have to get deeper into shading for this. Colors are easier).
Screenshots will follow.

Gorgious56:
@Moult I'm not that familiar with sketchup, would you have an example of the edge rendering style you're talking about in your first comment ?

Moult:
@Gorgious56 in SketchUp a thick line is drawn at the section cut like this: https://visualizingarchitecture.com/wp-content/uploads/2012/01/Section_architecture_2-1024x601.jpg - however in Blender if you section cut via clipping plane or alt-b, there is no decoration to that exact line being cut, so if you view a regular wall in a plan view, you never actually see the outline of the wall, you only see the backface of the bottom of the wall. If that backface happens to be modeled badly and intersecting with the slab or z-fighting occuring, then the wall might even be invisible in a plan view.

TZwielehner:
![image](https://user-images.githubusercontent.com/9949323/199460489-f2df7883-e240-444e-900c-0d497bbd6e90.png)
![image](https://user-images.githubusercontent.com/9949323/199460526-f85f99ed-638d-4a26-a3a0-877b29294430.png)

[...]

PR #2941, opened 2023-04-05
Section cutaway plane cut line decoration

Gorgious56:
Re the discussion on #2547

This is a proposal to display a thick black line at the section limit on the mesh, when using the temporary section cutaway tool.

Here it is in action

Animation3

The PR adds a new parameter to tweak the line width in the output properties. It can be set to 0 to display like it was before. The value is in Blender units, but due to how it's setup I think it generates a line double the width of what is set in the field. (absolute on signed distance)

image

Pros : Realtime, does not create any new geometry

Cons : 

Doesn't work when viewing the section orthogonally, 

image

Doesn't work well for faces parallel or next to parallel to the section plane (it creates black surfaces instead of a thick line since the effect is distance base)

image

Edit : Also it doesn't work with more than one section object. I think I have a solution if it's really a problem but it requires a bit more tinkering with the node group.

### DIFF
--- a/src/blenderbim/blenderbim/bim/operator.py
+++ b/src/blenderbim/blenderbim/bim/operator.py
@@ -210,12 +210,12 @@ class BIM_OT_add_section_plane(bpy.types.Operator):
         return bpy.data.node_groups.get("Section Override")
 
     def create_section_compare_node(self):
-        # Add a new socket to group input / output https://blender.stackexchange.com/a/5446/86891
         group = bpy.data.node_groups.new("Section Compare", type="ShaderNodeTree")
         group.inputs.new("NodeSocketFloat", "Value")
         group.inputs["Value"].default_value = 1.0  # Mandatory multiplier for the last node group
         group.inputs.new("NodeSocketVector", "Vector")
         group.outputs.new("NodeSocketFloat", "Value")
+        group.outputs.new("NodeSocketFloat", "Line Decorator")
         group_input = group.nodes.new(type="NodeGroupInput")
         group_input.location = 0, 50
 
@@ -226,6 +226,12 @@ class BIM_OT_add_section_plane(bpy.types.Operator):
         greater.operation = "GREATER_THAN"
         greater.inputs[1].default_value = 0
         greater.location = 400, 0
+
+        compare = group.nodes.new(type="ShaderNodeMath")
+        compare.operation = "COMPARE"
+        compare.inputs[1].default_value = 0
+        compare.inputs[2].default_value = 0.04
+        compare.location = 400, -200
 
         multiply = group.nodes.new(type="ShaderNodeMath")
         multiply.operation = "MULTIPLY"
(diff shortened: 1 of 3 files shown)


Moult:
This is AWESOME! I have been looking for functionality like this in Blender for a very, very long time! I'll need to investigate the incantation you used to make it happen :)

Gorgious56:
@Moult hehe I only added a couple of nodes actually :)

A compare math node to get the distance to the section plane

image

And a mix shader to display the black line when the distance threshold is met

image

schroef:
I see this implementation has issue when multiple section cuts are in the scene. The section cuts have loop from on to the other. But the line decoration is only added later and thus looks weird

issue line decorator-multiple section cuts pn

Gorgious56:
Yes the black lines features only supports one section cut as far as I remember. Theoretically I think it could be extended to support multiple cuts. In the meantime you can disable it by setting the line width to 0

schroef:
Yeah, I got it sort of working. But the issue is that I need to add many extra nodes to get all the cuts also working on all the line decorations. If I use a compare node and then use the section compare the t works. But the issue afterwards is when I try to clean all sections or remove a single section.
Getting the corrects nodes and removing it and the reconnecting all the proper nodes to each other again is a pain. Currently ingave a setup where the line decorator is added after the section compare. The line don't follow the shape properly.

I'm going to test what happens if I add the line to the material inside the section compare group. I think this makes cleaning and removing each section easier. My current setup uses a lot of extra code just for line decoratie. It's a mess.

I've also got the section shader working on collections which actually have a material. The current version doesn't do that. It only works on objects without a material.

schroef:
Okay did a test today again and i got it to work. Though i made the setup manually. I think this can be done using code as well. Atleast in theorie. It also looks cleaner than my first try and is easier to clean afterwards. Where this new approach only needs to remove the compare node and text coor node. But the other would have cycle over many more nodes.

This was my first try. Lines get cut properly per section. But making this work by code is a nightmare!
fix for crossing lines

This is my second approach. the line decorator is moved inside the compare node. There it is mixed the the material and transparent node. Then because i simply add the sections one after the other. The line decorator still work nicely. Plus cleaning each section is easy. Its the same as the original, where the tex coord and compare node are removed.

Example with 3 sections. Each section is placed after the other and connects in tandem
line-decorator-inside-section-compare-1

This is how i setup the compare node inside
line-decorator-inside-section-compare-0

Now need to find the proper method of laying out the nodes. So it looks organized. Issue is the spaghetti going all over the place

Gorgious56:
Ooh this looks nice ! IMO node spaghetti is not a big deal. In the first
iteration of the section tool all nodes were placed at 0,0,0. Most users
won't dig into the node tree and those that will should know how to
navigate around a messy node tree. Where we don't want spaghetti is in the
code though ^^

schroef:
Well a bit of tidying up and it looks clean now. Also got the deletion work now of single sections and reconnecting everything. Was a bit of a hustle since the last and the first in the section compare line are connected to different nodes. But with some if elif and else options, i got that sorted out now.

PR #2957, opened 2023-04-08
Add dynamic tool tip for parametrized operator "Export IFC"

rileywong311:
- reworked the `bl_description` of the "Export IFC" operator to change based on the class variable `should_save_as` which is dynamically set in the IFC project panel `UI.py` file. 
- there doesn't seem to be a standard for where this method is put, so I put it at the bottom of all the class's definitions.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/project/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/project/operator.py
@@ -890,13 +890,12 @@ class ExportIFC(bpy.types.Operator):
             bpy.ops.wm.save_mainfile(filepath=bpy.data.filepath)
         blenderbim.bim.handler.purge_module_data()
         return {"FINISHED"}
-    
+
     @classmethod
     def description(cls, context, properties):
         if properties.should_save_as:
             return "Export the IFC project to a selected file"
-        else:
-            return "Export the IFC project to this file"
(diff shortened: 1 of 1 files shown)


Gorgious56:
Nice use of dynamic tooltip ! If I may suggest a change, I would just invert the boolean logic here, it looks more "pythonic" to me :

```
        if properties.should_save_as:
            return "Export the IFC project to a selected file"
        else:
            return "Export the IFC project to this file"
            
```
When tested in an `if` statement, objects will be automatically cast to the boolean value, there is usually no need to test `if a is True:`, since it's equivalent to `if a:` and can be used for all sorts of objects. For instance you don't have to test `if len(my_list) == 0` to test an empty list, you can just use `if my_list:`. Please disregard if you already knew this, just wanted to elaborate in case it'd be useful to you.

Also you've got a couple spaces mixed in here at line 822 :)

Cheers

rileywong311:
I've updated the branch!

I guess one thing I noticed though is that where every other form of this description method in BlenderBIM is used, it loads a class attribute `description` or an empty string. For example, 

```
    @classmethod
    def description(cls, context, operator):
        return operator.description or ""
```

but I'm not sure if I should do that format because the pre-existing class attribute `should_save_as` provides all the necessary information already.

PR #2958, opened 2023-04-08
Length, depth and x angle properties now update when the field is updated

Gorgious56:
I find it mildly annoying to have to click the Refresh button next to the properties for the object geometry to be updated. 

This PR removes these buttons and adds a callbacks to the properties so the geometry is updated every time the fields are updated.

Pros : realtime feedback

Cons : may lag user computer when scrubing the fields with a lot of items selected. I think this is not really an issue since the purpose of these fields is to input a precise value so I think it's unlikely the users would scrub them in the case they have multiple elements selected. It's not that bad either for simple shapes.

Animation (1)

I also set the minimal values to 0 since negative lengths don't really make sense (unless they do in the ifc schema ?) and the geometry creator doesn't seem to like it either.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/model/prop.py
+++ b/src/blenderbim/blenderbim/bim/module/model/prop.py
@@ -202,7 +202,7 @@ class BIMModelProperties(PropertyGroup):
     constr_browser_state: bpy.props.PointerProperty(type=ConstrBrowserState)
     extrusion_depth: bpy.props.FloatProperty(
         min=0,
-        default=42.0, 
+        default=42.0,
         update=update_extrusion_depth,
     )
     cardinal_point: bpy.props.EnumProperty(
@@ -233,7 +233,7 @@ class BIMModelProperties(PropertyGroup):
     )
     length: bpy.props.FloatProperty(
         soft_min=0,
-        default=42.0, 
+        default=42.0,
         update=lambda self, context: bpy.ops.bim.change_layer_length(length=self.length),
     )
     openings: bpy.props.CollectionProperty(type=ObjProperty)
@@ -243,7 +243,7 @@ class BIMModelProperties(PropertyGroup):
     rl1: bpy.props.FloatProperty(name="RL", default=1)  # Used for things like walls, doors, flooring, skirting, etc
     rl2: bpy.props.FloatProperty(name="RL", default=1)  # Used for things like windows, other hosted furniture
     x_angle: bpy.props.FloatProperty(
-        name="X Angle", 
+        name="X Angle",
         default=0,
         update=lambda self, context: bpy.ops.bim.change_extrusion_x_angle(x_angle=self.x_angle),
     )
--- a/src/blenderbim/blenderbim/bim/module/model/prop.py
+++ b/src/blenderbim/blenderbim/bim/module/model/prop.py
@@ -145,6 +145,15 @@ def update_preview_multiple(self, context):
             update_relating_type(self, context)
 
 
+def update_extrusion_depth(self, context):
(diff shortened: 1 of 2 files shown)


Moult:
From memory I believe the reason this isn't the case is because it actually edits IFC data. I have to check whether or not these edit in-place or if they create and delete in the IFC (which would then start incrementing the step ID unnaturally).

Gorgious56:
I also contemplated displaying the ° for `x_angle` with `subtype="ANGLE"`, to convey they're degrees and not radians, and it might make the subsequent calculations more straightforward.

image

 Since after looking it up it seems this property is systematically converted to radians afterwards in the code.
eg 

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/model/wall.py#L251

 So to be clear the displayed value would be in degrees, but the actual sotred value would be in radians.

It would mean changing a bit more than a few lines to accomodate for the change, though.

Gorgious56:
> From memory I believe the reason this isn't the case is because it actually edits IFC data. I have to check whether or not these edit in-place or if they create and delete in the IFC (which would then start incrementing the step ID unnaturally).

Oh right then this makes sense. A bit unfortunate, but understandable for the sake of having a sane bookkeeping.

Moult:
Yes I suspect many elements will be created / removed in IFC, and this may also have an impact on the IFC undo step. I haven't had time yet (rushing for this upcoming release) to check these suspicions.

theoryshaw:
Could we do something like the following?
Yes, this is a chatGPT inquiry. :)
Seems to work.

```
import bpy

# Store the last entered value to compare changes
class MyProperties(bpy.types.PropertyGroup):
    last_value: bpy.props.StringProperty()

# Operator to execute on "Enter"
class MyOperator(bpy.types.Operator):
    bl_idname = "object.my_operator"
    bl_label = "Operator on Enter"

    def execute(self, context):
        input_value = context.scene.my_input_property
        self.report({'INFO'}, f"Operator triggered with input: {input_value}")
        return {'FINISHED'}

# Callback function to detect "Enter" on value change
def update_on_enter(self, context):
    current_value = context.scene.my_input_property
    props = context.scene.my_props
    
    # Check if the value has truly been "entered"
    if current_value != props.last_value:
        props.last_value = current_value  # Update the last value
        bpy.ops.object.my_operator()  # Run the operator

class MyPanel(bpy.types.Panel):
    bl_label = "Enter to Confirm"
    bl_idname = "OBJECT_PT_my_panel"
    bl_space_type = 'VIEW_3D'
    bl_region_type = 'UI'
    bl_category = 'My Addon'

    def draw(self, context):
        layout = self.layout
        layout.prop(context.scene, "my_input_property", text="Enter Value")

def register():
    bpy.utils.register_class(MyProperties)
    bpy.utils.register_class(MyOperator)
    bpy.utils.register_class(MyPanel)
    
    bpy.types.Scene.my_input_property = bpy.props.StringProperty(
        name="My Input",
        update=update_on_enter
    )
    bpy.types.Scene.my_props = bpy.props.PointerProperty(type=MyProperties)

def unregister():
    bpy.utils.unregister_class(MyProperties)
    bpy.utils.unregister_class(MyOperator)
    bpy.utils.unregister_class(MyPanel)
    
    del bpy.types.Scene.my_input_property
    del bpy.types.Scene.my_props

if __name__ == "__main__":
    register()
```

theoryshaw:
If we can only get something like this to work, with removing the scrub bar, I would vote to remove the scrub bar.
We don't really get real time use of the scrub bar, anyway.

Issue #2825, opened 2023-03-02
BlenderBIM - UI Visibility Preferences do not reapply on launch.


oeykmiih:
Steps to reproduce the bug:

1. in BlenderBIM Addon Preferences change UI visibility.
2. Restart Blender
3. the Preferences reset! :(

Expected Result: BlenderBIM visibility preferences saved over different sessions.
Actual Result: BlenderBIM visibility preferences resets on every session.

![230212_201820](https://user-images.githubusercontent.com/69322448/222447427-6f0b4097-33b6-4363-a77c-3dd71f63961b.gif)

Moult:
@vulevukusej maybe one for you?

vulevukusej:
I won't have time for at least a couple of months unfortunately.

maxfb87:
@oeykmiih i think that this is not really a bug... you can save the visibility status with File -> Defaults -> Save Startup File .
In this way, every time you start a new fresh file, the visibility configuration is what you have saved.
On the other hand, if you are working on a file, the session is stored in the .blend file.

Your video btw shows an error on the console: i tried to investigate and it seems that the error is that a not visible module is unregistered 2 times (when the it becomes unvisible and on the blender exit) ...

Gorgious56:
As @maxfb87 said this is not a bug since the system was designed to store the parameters on the active Scene, so the visibility states are stored per file. Or more exactly, per scene. At least theoretically since there is no callback to actually change the enabled tabs to reflect the current scene properties on scene change. Also, the BlenderBIM addon is not designed to support multiple scenes per file as far as I know (like most addons, since we can't efficently store settings per file, we have to use the current scene, but that is another debate...). 

It is arguably misleading to have this feature accessible from the addon preferences which are usually where blender-wide settings are displayed. Maybe the solution is to move the visibility popup into the BIM scenes properties panels, that way it is clear these are settings that belong to the current scene.

maxfb87:
@Gorgious56 it is a good suggestion but honestly i don't imagine where it could be moved...

Have you seen the related bug about already unregistered module? I have done also a PR to solve it but i don't know if it correct. Have you got any ideas?

Moult:
Agreed, this is not actually a bug.

PR #2962, opened 2023-04-09
Fix minor bug

maxfb87:
Related to #2825, have been created others panels or operators that haven't the is_regitered method so there was a new error in blender console.
So i found a better and cleaner method to solve the problem also in case of, in the future, there would be new similar problems.


maxfb87:
@Andrej730 i have seen your commit here https://github.com/IfcOpenShell/IfcOpenShell/commit/e71ee988bf45510a009d0c46ead7aa0a2f4e9d63

Maybe this is better?
The problem here was that if a module is not visible, it is unregistered two times causing the error...

Andrej730:
Sure, it will work.
Btw any ideas why did the original error with unregistered modules occured? I thought we import all modules in `bim/__init__.py` so it shouldn't happen. Currently it still might occur for classes based on Gizmo, GizmoGroup, Macro if they were not registered for some reason.

maxfb87:
Yes, i think that the original error was found as a side effect of #2825.
If you look at the video shared, you'll see that there was an error in console line.
I tried to solve it in a previous PR, but i think that something can be done.

Basically the issue is when a module is not visible: this module is unregisted too so, on exit, blender tries to unregister a module that is already unregistered.
From this perspective, i think that the method ".is_registered" is suited to check if a module is already unregistered or not but not every cls has the method ".is_registered".
So i did a sort of workaround a few weeks ago but yesterday a saw the new bug and i did this PR that, i think, it's more general and cleaner.

Andrej730:
kk, merged it, just was curious to pin down why classes appear to be unregistered.
thank you for the contribution.

PR #2973, opened 2023-04-12
add pandas to makefile

C-Claus:
Is it possible to add pandas to ```../../blenderbim/libs/site/packages```? 
Seems like an extremly useful python module, I am planning to incorporate blenderbim spreadsheet to blenderbim. It needs this module.


C-Claus:
Next to pandas, I think the openpyxl module would be also useful?

Moult:
I think you're right, sooner or later pandas would be a huge help in doing spreadsheet and cost related things.

But are you sure this link is correct? It seems to point to a source distribution (pandas-2.0.0.tar.gz) https://pypi.org/project/pandas/#files

BTW a better approach would be to use pip sort of like this https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/Makefile#L220 and specify the platform flag https://stackoverflow.com/questions/45980037/how-to-download-pip-packages-for-a-different-operating-system

Moult:
I also agree about openpyxl. IfcCSV currently uses xlsxwriter but I think that's a bad idea (it only writes) and recently I've been using openpyxl more and find it much more capable.

C-Claus:
@Moult  Thanks for your reply, I have no experience with makefiles. I will look into into and try to find the correct way of adding it. _I just copied a piece of code from the thing next to it with no real clue what is going on...😅_

> I've been using openpyxl more and find it much more capable.

I've came to the same conclusion when writing BlenderBIM spreadsheet.

C-Claus:
After reading this post, https://community.osarch.org/discussion/comment/16243/#Comment_16243

I guess this PR can be closed?

It says pandas is now included.

Moult:
Ah thanks for reminding me about this. Pandas isn't included in BBIM, but if you have Pandas you can use it via a library in IfcCSV as described here https://blenderbim.org/docs-python/ifccsv.html#source-installation

I still can't approve this pull request as it isn't quite right since:

 1. It includes Brick.ttl, which we shouldn't store in the repo
 2. It doesn't handle multiple platforms, ideally we should do this via pip install with a platform argument

Would you like to make those changes? Alternatively I can just commit. Is pandas needed for openpyxl?

PR #3020, opened 2023-04-18
fix EditBcfTopic and EditBcfTopicName

htlcnn:
@Moult 
fix EditBcfTopic and EditBcfTopicName
`bcfxml.topics[blender_topic.name]` returns `TopicHandler` instead of `Topic`. Shoudl be `topic = TopicHandler.topic`.


Moult:
Thanks so much @htlcnn !

PR #3032, opened 2023-04-19
Converting quantities to project units to fix #3025

### LINKED ISSUE #3025: If an imperial file, have .bim.calculate_all_quantities() calculate imperial values
If an imperial file, have `bim.calculate_all_quantities()` calculate imperial values

Currently if calculates metric values.

Andrej730:
Added conversion to project units at the end of calculations - currently all the length/area/volume units are calculated in meters and they need to be converted if the project is using some other unit or prefix.

Please take a look - I haven't worked on this part of code before.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/pset/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/pset/operator.py
@@ -381,11 +381,19 @@ class CalculateQuantity(bpy.types.Operator):
         self.qto_calculator = QtoCalculator()
         obj = context.active_object
         prop = obj.PsetProperties.properties.get(self.prop)
-        prop.metadata.float_value = self.calculate_quantity(obj, context)
+        quantity = self.calculate_quantity(obj, context)
+
+        if quantity is None:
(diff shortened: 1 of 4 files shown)


Moult:
At a glance you're using the right functions. Have you tested it? If so, feel free to merge. Would be nice to write one or two tests for this.

theoryshaw:
seems to work on my side!

image

maxfb87:
@Andrej730 ok, i think it is ok!
Maybe it is also useful to compare it with  the solution already available in /blenderbim/bim/module/pset/operator.py
in lines 374-408 but maybe this is better because it reads directly from the ifc file...

Also, maybe, for doing better tests, the function "convert_to_project_units" should be a tool function?

Just thinking loud btw, for me it can be merged :-)

maxfb87 (approved):

Andrej730:
> Maybe it is also useful to compare it with the solution already available in /blenderbim/bim/module/pset/operator.py
Missed that one! I'll connect them all to the same tool method.

maxfb87 (approved):

maxfb87 (approved):

PR #3056, opened 2023-04-22
Ifcgit module

brunoperdigao:
**WIP**
This PR aims to integrate [Ifc-Git](https://github.com/brunoperdigao/ifc-git) from @brunopostle as a BlenderBIM Module.
In order to test it, please be aware of the additional installations still needed as mentiond [here](https://community.osarch.org/discussion/comment/14894/#Comment_14894)
There are a few steps to take before merging, as discussed [here](https://github.com/brunopostle/ifc-git/issues/7).


brunopostle:
Notes on dependencies:

1. This requires [GitPython](https://gitpython.readthedocs.io/) which is pure python and which will make the blenderbim installer about 200kB larger.
2. It also needs the `git` binary to actually work. Currently the IfcGit panel just says 'Git is not installed' if git is not in the `PATH`. For now this will have to be an additional download. There is a 100% python Git implementation called [dulwich](https://github.com/jelmer/dulwich), maybe we can port the module to use this at some point.
3. Merging requires `ifcmerge`. The merge button is greyed-out if ifcmerge is not in the `PATH`. Merging is not core functionality so I think this is ok for now.

brunopostle:
The module in this pull request isn't working at all, you need the attached patch. Though using it requires more clicking the refresh button than in the original add-on.
[ifcgit.patch.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/11302322/ifcgit.patch.txt)

brunoperdigao:
Sorry, I tested in the wrong Blender instance and was too optmistic about the results. Thanks for the patch

src/blenderbim/blenderbim/bim/__init__.py
```diff
@@ -71,6 +71,7 @@
     "covetool": None,
     "augin": None,
     "debug": None,
+    "ifcgit": None,
```
Moult, line 74: The module load order determines the panel draw order so maybe put it above "debug" at least :)

brunopostle: The IFC Git panel gets added as a sub-panel of IFC Project Info, so it appears a bit further up the Scene Properties anyway

Issue #3059, opened 2023-04-24
IfcTester request - Report "shall not be" for prohibited facets 


mdjska:
A small request; it would be nice if when a facet is prohibited (maxOccurs=0) the reporter printed the facet as "The XX shall not be YY" instead of "The XX shall be YY".

Moult:
Great idea! Would you like to have a go at this?

mdjska:
I tried, but it's a bit of a hack

PR #3062, opened 2023-04-24
Add negation to reports for prohibited facets

mdjska:
https://github.com/IfcOpenShell/IfcOpenShell/issues/3059, Reports "The XX shall not be YY" instead of "The XX shall be YY" if a facet is prohibited (maxOccurs=0)


Moult:
Holding off merging this since I'm not too happy with the approach used. I'd prefer explicit templates instead of string replacement.

Moult:
Implemented https://github.com/IfcOpenShell/IfcOpenShell/commit/fab8b5b702cf5f4b4a6b1deca66d24f6c1297630

Issue #3076, opened 2023-04-30
Bug in getting unit value of IfcConversionBasedUnit 


shoheikuni:
Hi. I found a bug in `IfcFile::getUnit(const std::string& unit_type)` in IfcParse.cpp.

The function checks if each unit in the input IFC file is IfcConversionBasedUnit or IfcSIUnit and returns the unit's value in SI, but the process for IfcConversionBasedUnit has a mistake.

Here is the current code in v0.7.0:
https://github.com/IfcOpenShell/IfcOpenShell/blob/83084aeb7b118e7ebb184f31019d7842dce3a252/src/ifcparse/IfcParse.cpp#L2593-L2615

Attribute "ValueComponent" is used for variable `vlc` and again "ValueComponent" is used for `unc`, but the latter should be "UnitComponent".
Line 2603 should be corrected as follows:
```diff
 	IfcUtil::IfcBaseClass* unc = *mu->data().getArgument( 
- 		mu->declaration().as_entity()->attribute_index("ValueComponent") 
+ 		mu->declaration().as_entity()->attribute_index("UnitComponent") 
 	); 
```

aothms:
Thank you, you're right. We have rewritten this fairly recently to make use of latebound entity definitions so that it works independent of schema version, but apparently in that process this mistake was introduced. When writing latebound code the compiler can't help you anymore with these kind of errors...

Could you submit your suggestion as a Pull Request here on Github? Thanks!

shoheikuni:
OK. I've made a PR.

PR #3077, opened 2023-04-30
Fix wrong attribute name usage in IfcFile::getUnit

shoheikuni:
This PR is for #3076.
In IfcFile::getUnit in IfcParser.cpp, a wrong attribute "ValueComponent" was used to get the value of IfcConversionBasedUnit.
I fixed it to "UnitComponent".


aothms:
Thanks again :)

Issue #2983, opened 2023-04-14
Tab an object to `bim.enable_editing_extrusion_profile()`


theoryshaw:
Would be boss if you could double click an object to `bim.enable_editing_extrusion_profile()` and once in edit mode, double click in empty space to exit out of edit mode.

theoryshaw:
...and/or use 'tab' to enter `bim.enable_editing_extrusion_profile()`.

Moult:
I like the Tab idea. That would free up Shift-E to be used for something else.

Gorgious56:
I don't think double click fits well with the usual Blender workflow but "tabbing" in and out of edit mode would be way more straightforward.

Would just need to think about a way to inform newbies that they're modifying the profile and not the resulting geometry, to dissipate confusion. Maybe adapting the info message that's currently written when trying to edit a non-mesh object. something like "Entering edition of profile xyz"

Gorgious56:
Should be fixed with #3142

PR #3142, opened 2023-05-12
Edit profile based representations with TAB

Gorgious56:
Following #2983

This allows users to TAB into edit mode when one profiled-based (SweptSolid) representation is selected.
I didn't know how to check if the profile editor was active so I added a little helper in blenderbim.tool.Profile to check if the ProfileDecorator is installed. I guess it's a weak way to check if edition is going on but it seems reliable enough.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/geometry/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/geometry/operator.py
@@ -748,6 +748,18 @@ class OverrideModeSetEdit(bpy.types.Operator):
             if not representation:
                 continue
 
+            if len(context.selected_objects) == 1:
+                if tool.Pset.get_element_pset(element, "BBIM_Railing"):
+                    # Needs to run BEFORE is_meshlike check
+                    bpy.ops.bim.enable_editing_railing_path()
+                    obj.data.BIMMeshProperties.mesh_checksum = tool.Geometry.get_mesh_checksum(obj.data)
+                    return {"FINISHED"}
+                elif len(context.selected_objects) == 1 and tool.Geometry.is_profile_based(representation):
+                    # Only one profile can be edited at a time
+                    bpy.ops.bim.enable_editing_extrusion_profile()
+                    obj.data.BIMMeshProperties.mesh_checksum = tool.Geometry.get_mesh_checksum(obj.data)
+                    return {"FINISHED"}
+
             if tool.Geometry.is_meshlike(representation):
                 if getattr(element, "HasOpenings", None):
                     # Mesh elements with openings must disable openings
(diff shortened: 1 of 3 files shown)


Gorgious56:
I guess next step is to remove the SHIFT + E shortcut which is now superfluous ? 

I would keep this one though, since it lets user go into local front mode and quit edition without saving the changes.
image

Gorgious56:
This fails for railing objects. How can I easily know if the active object is a railing object ? https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/model/workspace.py#L209-L211 seems a bit overkill

Moult:
Sorry I'm late catching up on these PRs.

All profile meshes (or other magic meshes) have their mesh type set e.g. `mesh.BIMMeshProperties.subshape_type = "PROFILE"` - this _should_ correlate with a corresponding decorator. Does that help?

Agreed that Shift-E should be removed once Tab is implemented. Go ahead and merge when you think it's ready :)

theoryshaw:
@Gorgious56 This is great, of course. 
It doesn't seem to work on ifcslab instances, however.  I might be missing something.

Gorgious56:
I had a problem with slab not regenarating but I think I fixed it in a recent commit. Are you using the latest version ? Can you share a stripped down ifc file with the problematic slab ?

theoryshaw:
It works with just starting a new demo file, and adding a slab instance...
[untitled.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/11713122/untitled.ifc.txt)

Gorgious56:
I think I found the problem but I hope there will be no unintended consequences.

Plae-Blue-Dot:
Just tried the new update tab on horizontal elements, it's amazing thankyou @Gorgious56

Plae-Blue-Dot:
Is it possible to extend this functionality to edit wall axis aswell?

normal wall:
image

Curve 2d / axis representation:
image

Gorgious56:
I don't know about allowing to edit wall axis with TAB because there is a dedicated editing workflow specifically for walls with the Extend / Butt / Merge / etc. shortcuts. I don't know if editing the wall axis can be integrated well with these features. Especially because as per the schema I think the wall axis must stay along the local X axis which is hard to enforce if we let users mess with the vertices. @Moult what do you think ?

Moult:
Agreed. It will take a bit more implementation to do properly. @Plae-Blue-Dot if you think its worth it feel free to create a bug to track it.

Plae-Blue-Dot:
I think it is a nice to have, even if it is restrained in the x direction, absolutely not essential as we do have a workflow around walls
I made a issue here:

https://github.com/IfcOpenShell/IfcOpenShell/issues/3297

I think there is something natural about pressing tab and being able to edit if it all follows the same language, people are going to press tab to see what can be edited that way, but absolutely not important right now.

Issue #3098, opened 2023-05-03
[Error] IfcGradientCurve Not Implemented


civilx64:
Per recent discussion with @aothms I have spun back up on IFC 4x3 implementation with an initial focus on `IfcAlignment` geometry parsing.

Testing with [UT_AWC_1.ifc](https://github.com/bSI-RailwayRoom/IFC-Rail-Sample-Files/blob/main/1_Alignment%20with%20Cant%20(AWC)/UT_AWC_1/IFC%20reference%20files/RC4/UT_AWC_1.ifc) and using the representation only (not segment parameters) I get the following error:

```
RuntimeError: [Error] [2023-04-30 04:20:35] Not implemented
#511=IfcCurveSegment(.CONTSAMEGRADIENTSAMECURVATURE.,#517,IfcNonNegativeLengthMeasure(0.),IfcParameterValue(61.6732236801535),#520)
[Error] [2023-04-30 04:20:35] Failed to convert:
#510=IfcGradientCurve((#511,#524,#536,#549,#561,#574,#586,#599,#611,#624,#636,#649,#661,#674,#686,#699,#711,#724,#736,#749),.F.,#204,$)
```

I suspect that this means exactly what it says: `get_shape` is not implemented for `IfcGradientCurve`.  If so then I will proceed with a python implementation that can be ported to C++ once working and tested for correctness.

aothms:
Welcome back. Your suspicion is fully correct. For myself a long time back I basically broke down the work required in:

- transition curves / spirals
- horizontal+vertical (IfcGradientCurve)
- cant (IfcSegmentedReferenceCurve)
- sweeps - we can't rely on opencascade because of cant (occt only has tracking along surface normal (cfr IfcSurfaceCurveSweptAreaSolid) and efficiency (feeding a segmented 3d polyline from hor+ver would not perform).
- linear placement

RickBrice:
How can I help? I'm generating slab on beam superstructure models and the geometry doesn't display in BlenderBim. My guess is the need for alignment and linear placement support.
@civilx64 
@aothms

aothms:
So where we currently stand:

We have wonderful work by @civilx64 to implement support for the business logic side of alignment in python using numpy. https://github.com/IfcOpenShell/IfcOpenShell/pull/3148

Basically the ball is in my court now to translate this to C++ so that it becomes generally available in all of ifcopenshell.

[...]

PR #3148, opened 2023-05-13
IfcAlignment implementation

civilx64:
Addresses #842 and #1470.  @aothms this is a refactoring of the functors and piecewise work you did in #1480 but appears to be working correctly for generic test data that I generated on my end.  That data is in the series of `C3D-UTxx` files referenced in the code which are not in the repository or test data yet but can be added to the sample files submodule if that is the most logical place for them to live.  I believe I was sym-linking into `tests/fixtures` locally during development.

With `alignments.py` as-is you should be able to verify the first four UT0 synthetic test cases from the [IFC Rail Sample Files](https://github.com/bSI-RailwayRoom/IFC-Rail-Sample-Files) repository.  It should also work for the two `UT_AWC1` test cases as well, but for horizontal parameters only.

Also I should note this is only utilizing the parameter definitions of the alignment, not the representation.  IOW this does not yet address #3098. I wanted to hand this current work over as MVP to study how you port this into a new `ifcgeom/IfcAlignment.cpp`  as that may have some bearing on the path forward.  For example the clothoid implementation as currently coded in python uses a Taylor series approximation but other schools of though e.g. Open Infra Platform seem to tend towards numerical integration.


aothms:
Amazing, most of this is still way over my head in terms of the math, but I hope to catch up.

I really appreciate the systematic approach comparing to the excel sheet. If you don't mind I'll rewrite it to a an actual parametrized pytest test case like we have in various other places, https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/test/test_validate.py#L27-L31 That way it will be ran automatically here on Github Actions. I would also add that repo then as a submodule in the existing submodule.

The biggest question I have is on the point you brought up as well on semantic-level parameters vs representation-level geometries. Conceptually on the C++ side, the representation-level geometries is easier to handle, because it would be identical to how geometrical elements are currently handled. Also, ultimately, linear placement will be built around the composite curve representation of the alignment, not the semantic counterpart.

I think the equivalence between the two is described to some extent here https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/concepts/Product_Shape/Product_Geometric_Representation/Alignment_Geometry/content.html#Horizontal-Alignment 

(I remember from discussions with other vendors that when implementing off of the geometric side of things that less integration or taylor approximation is necessary. But I didn't discover what the mechanism is behind this.)

What do you see as the ideal end-result regarding semantic-level parameters vs representation-level geometries?

Is it an option to programmatically rewrite one into the other and implement only one of the two? This appears what Peter has done here IFC-Rail-Unit-Test-Reference-Code/tree/master/EnrichIFC4x3/EnrichIFC4x3/business2geometry. But we only need a less complete solution. E.g we don't need to generate the actual placements for the https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcCurveSegment.htm to make it fully independently positioned segments, but rather just translate the predefined types and parameters to the equivalent geometric entities so that I can implement them in the mapping and placements will be handled generically.

Also, I think I'll need to input on how to handle additional dimensions (up to cant). Eventually we will need to implement this within the framework of the v0.8.0 (See https://github.com/IfcOpenShell/IfcOpenShell/issues/2977) geometry mapping idiom. In v0.8.0 we two layers of geometry implementation. First, we translate IFC-specific geometries to generally understood concepts. For example here we translate an i-shape profile to a loop https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/ifcgeom/mapping/IfcIShapeProfileDef.cpp#L26. Later-on in the process we can handle a loop in either CGAL or OpenCASCADE. If we just focus on horizontal alignment I can make sense of this, but for vertical and especially cant (which doesn't have a proper concept yet in our mapping ontology) I have some more doubts. Is a stateless conversion sufficient, i.e a mapping of `IfcCurveSegment` -> `loop`, `IfcCompositeCurve` -> `loop` (2d), `IfcGradientCurve` -> `loop` (3d), `IfcSegmentedReferenceCurve` -> something new 6d? Or should this not be implemented as stateless conversion functions, but rather some stateful incremental process is needed? I.e how is does the output evolve as the components horizontal+vertical+cant are evaluated? When we have a composite curve for horizontal and vertical how do we fold it into a single 3d curve? Same for when we evaluate placements and/or sweeps.

Sorry for this wall of text, I would love to get started on this, but to summarize I need some more input first. Maybe we should collaboratively sketch some sort of high level approach with datatypes for these three workflows (segmented evaluation of alignment axis, linear placement, sweeps) so that I can better plan what we're working towards and how I can fit this in into the v0.8 C++ architecture.

civilx64:
> If you don't mind I'll rewrite it to a an actual parametrized pytest test case

Absolutely! I was going to do this on my end but was having issues running tests due to circular import of ifcopenshell.

> What do you see as the ideal end-result regarding semantic-level parameters vs representation-level geometries?

As a infrastructure professional I am much more interested in the semantic-level parameters due to how the business logic relates to design requirements e.g. minimum radius for a given design speed.  Also in LandXML there is no concept of geometric representation for alignments - only semantic parameters.  My understanding of the spec is that a valid IFC exchange can contain representation only, semantic parameters only, or both.

> Is it an option to programmatically rewrite one into the other and implement only one of the two?

I believe so - interpret the semantic parameters into the representation equivalents and then code geometry for the representation entities.

> Also, I think I'll need to input on how to handle additional dimensions (up to cant). 

Understood.  I will move forward with implementing vertical alignment.  IMO the 3D alignment axis workflow should ignore cant for the time being as that does not become important until (possibly important) linear placement or (definitely important) sweeps are addressed.

> Eventually we will need to implement this within the framework of the v0.8.0

I had suspected this might be the case.  I will re-base off of v0.8.0 moving forward and do some study of the code per your links provided.

> Maybe we should collaboratively sketch some sort of high level approach with datatypes for these three workflows

For alignment axis I envision a 6d array of `[distance along alignment, x, y, z, bearing, radius]` where the geometry would be generated as a Bspline through the x,y,z coordinates as control points.  Linear placement would map to a 4x4 transformation matrix and sweeps would generate a manifold solid of tesselated faceset or brep of bsplines.  I'm happy to discuss further via whatever medium would be best.  Maybe as a megathread on OSArch wiki?

civilx64:
Thanks for the wiki page!  I reviewed the dot graph and will make a few revisions that hopefully will clarify the data model. I also realized that cant is addressed in an interesting way so I think our alignment points array will end up as 8d to accommodate IfcSegmentedReferenceCurve with two different elevations at the basis curve 2D location. 

Polyline is fine IMO. I was thinking back to original discussions about maybe using BSplines for smoother interpolation between points.

aothms:
Sounds great. Do you have upload rights for images to the wiki? Otherwise I'll ask @DuncanLithgow or @Moult (?) to grant me image upload rights on the wiki.

> Polyline is fine IMO. I was thinking back to original discussions about maybe using BSplines for smoother interpolation between points.

I guess it depends on the granularity with which we sample the distance along. It would sure be interesting to compare both approaches, but I don't think we can rely on opencascade's existing algorithms for the sweeps, at least not with cant (for fixed vertical I think we can), so we'd have to do our own nurbs-based sweep with cant, which sounds complex.

civilx64:
After studying the graphic multiple times and reviewing multiple examples with horizontal, vertical, and cant I think it's fine to leave as-is.  I worked a bit on some of the other horizontal transition curves (BLOSS, COSINE, etc.) but now will focus on the Alignment With Cant example files, focusing on mapping the business logic to the representation items.

aothms:
That's great to hear!

We just had a session going over the ISO comments. One of the important issues that came up (once we get to linear placement) is that if you look at IfcPointByDistanceExpression [0] you'll see that the distance along the curve can be either a length measure or a parameter value, but parameter values are not defined for all types of spirals. I don't know yet in which direction this will be resolved, but just wanted to put this out here as it might be relevant for our "api design".

https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcPointByDistanceExpression.htm#8.9.3.49.3-Attributes

civilx64:
@aothms I have implemented vertical and cant alignment calculation in this latest update.  I ended up using `numpy.piecewise()` for delegating the calculation of points by distance based on which segment they belong to.

I have not yet addressed any mappings from representation to business logic.  There is also additional work to be done on some of the various transition types, but I am approaching this like an MVP and focusing on the most common element types: horizontal lines, arcs, and clothoids; vertical lines, circles, and parabolas, and cant constant & linear transition segments.

As-is, the work to date supports a 3D swept component, e.g. `IfcRoadPart` following the generated horizontal and vertical alignment geometry.  Linear placement is up next, as it will be similar to the sweep in terms of coding as OCC transforms.

aothms:
Amazing progress, wonderful news.

>  I ended up using numpy.piecewise() 

I like that a lot. Conceptually I think an alignment is a piecewise defined function.

I don't think there is something equivalent in Eigen directly. But it only took some seconds to have ai generated some imperative variant of it. So I think we can copy that exact approach and maybe find a way later to vectorize it.

```C++
#include <Eigen/Dense>
#include <iostream>
#include <vector>
#include <functional>

using namespace Eigen;

struct LambdaDomain {
    std::function<VectorXd(const VectorXd&)> lambda;
    double domainStart;
    double domainEnd;
};

int main() {
    // Define the vector of lambdas with their respective domains
    std::vector<LambdaDomain> piecewiseFunctions = {
        {[](const VectorXd& x) { return x.array() + 1.0; }, -std::numeric_limits<double>::infinity(), 0.0},
        {[](const VectorXd& x) { return x.array() - 1.0; }, 0.0, std::numeric_limits<double>::infinity()}
    };

    // Create an Eigen vector
    VectorXd x(3);
    x << -1.0, 0.0, 1.0;

    // Apply the piecewise function to each element of the vector
    VectorXd result(x.size());

    for (int i = 0; i < x.size(); i++) {
        double currentValue = x(i);
        for (const auto& piecewiseFunc : piecewiseFunctions) {
            if (currentValue >= piecewiseFunc.domainStart && currentValue < piecewiseFunc.domainEnd) {
                result(i) = piecewiseFunc.lambda(x)(i);
                break;
            }
        }
    }

    // Print the result
    std::cout << "Result: " << result.transpose() << std::endl;

    return 0;
}
```

I will in the end probably not use OCCT for building the sweeps in C++ because of the overhead (and because I've had quite some issues in the past with robustness of complex sweeps in OCCT). So if it's easier for your to say in numpy (instead of gp_Trsf) then feel free. Otherwise `BRepBuilderAPI_MakePolygon` is a bit easier to generate a wire from points. If it works on your end though there is no reason to change as far as I am concerned.

civilx64:
I will plan to leave as-is for now.  I consider these python implementations as prototypes that are fair game for refactoring when we start porting to C++.

civilx64:
Closing this PR given that C++ alignment geometry calculation has been incorporated into v0.8.0 - shout out to @RickBrice !

PR #3166, opened 2023-05-17
Feat: USD serializer

D4ve-R:
### Description
This pull request adds a new serializer called `USDSerializer` to the IFCOpenShell project. The `USDSerializer` enables serialization of IFC (Industry Foundation Classes) files to the USD (Universal Scene Description) file format. This contribution enhances the capabilities of IFCOpenShell by providing support for the USD format, which is widely used in the computer graphics industry.

### Changes Made
- Added the `USDSerializer` class, which uses Pixar's USD library to serialize IFC files to the USD format.
- Integrated the `USDSerializer` into the existing IFCOpenShell project structure (e.g. IfcConvert).
- Changes in CMakeLists.txt for building the Project with USD support enabled.

### HowTo
- [Install Pixar's USD Library](https://github.com/PixarAnimationStudios/USD#getting-and-building-the-code)
- Add options for building with USD support to cmake:
```
cmake ../cmake \
   ...
   -DUSD_SUPPORT=On \
   -DUSD_INCLUDE_DIR=/path/to/usd/include/ \
   -DUSD_LIBRARY_DIR=/path/to/usd/lib/
```

### Testing Done
- Built IfcOpenshell successful with USD support on macOS 13.3.1(Intel) & Ubuntu 22.04.
- Build IfcOpenshell successful without USD support.
- Converted 3 different Ifc files to USD and validated results visually in Blender & NV Omniverse.
[AC20-FZK-Haus.ifc](https://www.ifcwiki.org/images/e/e3/AC20-FZK-Haus.ifc)
Bildschirmfoto 2023-05-17 um 16 14 43


src/serializers/USDSerializer.cpp
```diff
@@ -0,0 +1,166 @@
...
+  	auto usd_mesh = pxr::UsdGeomMesh::Define(stage_, pxr::SdfPath("/World/" + name));
+
+  	usd_mesh.AddTranslateOp().Set(pxr::GfVec3d(m[9], m[10], m[11]));
+
+  	pxr::VtVec3fArray points;
+  	for(std::size_t i = 0; i < verts.size(); i+=3) {
+  	  	points.push_back(pxr::GfVec3f(static_cast<float>(verts[i] * m[0] + verts[i+1] * m[3] + verts[i+2] * m[6]),
+  	                                  static_cast<float>(verts[i] * m[1] + verts[i+1] * m[4] + verts[i+2] * m[7]),
+  	                                  static_cast<float>(verts[i] * m[2] + verts[i+1] * m[5] + verts[i+2] * m[8])));
+  	}
```
aothms, line 135: Any particular reason why you choose to use a translation, but factor in the rotation into the vertex coords? That way you cannot reuse meshes across different instances (assuming usd has a concept like that).

`o->transformation().matrix().data()` comes from an https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcAxis2Placement3D.htm so it's guaranteed to be only rotation + translation, no scaling or shear.

Otherwise, an other option is to use USE_WORLD_COORDS in the settings. That way you can just ignore the matrix altogether and have ifcopenshell factor it into the coords for you.

aothms: (Objects in IFC sharing the same representation will be sequential in the iterator output and have the same `o->geometry().id()`)

src/serializers/USDSerializer.h
```diff
@@ -0,0 +1,84 @@
...
+	void write(const IfcGeom::BRepElement*) {}
+	void finalize();
+	bool isTesselated() const { return true; }
+	void setUnitNameAndMagnitude(const std::string&, float) {}
+	void setFile(IfcParse::IfcFile*) {}
+};
+
+#endif
+
+#endif
```
aothms, line 84: Appears like the src/svgfill submodule is updated in this pr? Submodules are a pain to work with I know. Could you cd into the directory and do `git checkout 4d2aa7 && cd .. && git add svgfill && git commit -m 'Update submodule' && git push` to revert this? 

afbeelding

aothms (changes requested):
Thanks for this amazing effort. See some comments inline. Are you by any chance interested in also updating the build scripts? https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/nix/build-all.py https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/win/build-deps.cmd That way this feature would be generally available in the automated builds we publish. If not, I'll try to sit true it. Thanks again :)

src/serializers/USDSerializer.h
aothms: Oh, or looking at the github actions log, maybe it's a commit you authored but didn't push?

D4ve-R: I reverted the changes in the submodule as you said. Not sure how and why this happend

src/serializers/USDSerializer.cpp
D4ve-R: You are right! I'm just getting to know USD and it's concepts. I know there is support for referencing etc, just have to figure out how it works

D4ve-R:
@aothms No problem! I will take a look a the build scripts and let you know if/when i tackle this.

D4ve-R:
@aothms, i tried updating the build script for *nix system but failed.
It should be pretty straight forward, by using the [build script in the USD Library](https://github.com/PixarAnimationStudios/USD/blob/b53573ea2a6b29bc4a6b129f604bbb342c35df5c/build_scripts/build_usd.py), which installs tbb & boost.
I don't know if this will conflict with another version of boost or checks for an existing one.
But maybe this approach is not so nice for automated builds, since there are flags for minimal builds, see 6c1218d3521b4d3eb3e71ab98e3109461362deb0

aothms:
I gave it a go, but also gave up. There is some work on making USD and the oneapi TBB version compatible [0], but even that is still really an ancient TBB version without CMake support (which I would really like for integration into the build script). Compiling USD with a recent oneTBB gives all sorts of gnarly compilation issues which doesn't seem like something we should be fighting with (I did fix a couple: [1])

It's really a shame as I wanted this to be generally available for people not doing their own compilation, but it seems like we have to give it some time for things to modernize on the USD side.

I'll merge this as is. Thank you for your great effort!

[0] https://github.com/PixarAnimationStudios/USD/pull/2466
[1] https://github.com/aothms/USD/commit/6f08b4deffe261289aa2e36b8a6a99f894edb8ae

aothms:
For reference these are the local changes I made when trying to get some things to work


```diff
diff --git a/nix/build-all.py b/nix/build-all.py
index 30a131bf7..0a0f2c019 100644
--- a/nix/build-all.py
+++ b/nix/build-all.py
@@ -203,7 +203,8 @@ dependency_tree = {
     'hdf5': (),
     'cgal': (),
     'freetype': (),
-    'usd': ('boost', 'oneTBB')
+    'usd': ('boost', 'tbb'),
+    'tbb': ()
 }

 def v(dep):
@@ -322,7 +323,11 @@ def run_cmake(arg1, cmake_args, cmake_dir=None, cwd=None):
     if "wasm" in flags:
         wasm.append("emcmake")

-    run([*wasm, "cmake", P, *cmake_args, f"-DCMAKE_BUILD_TYPE={BUILD_CFG}"], cwd=cwd)
+    build_type = []
+    if not any(arg.startswith("-DCMAKE_BUILD_TYPE") for arg in cmake_args):
+        build_type.append(f"-DCMAKE_BUILD_TYPE={BUILD_CFG}")
+
+    run([*wasm, "cmake", P, *cmake_args], cwd=cwd)


 def git_clone_or_pull_repository(clone_url, target_dir, revision=None):
@@ -745,12 +752,12 @@ if "cgal" in targets:
         revision=f"v{CGAL_VERSION}"
     )

-if "usd" in targets:
+if "tbb" in targets:
     build_dependency(
         name=f"oneTBB-{TBB_VERSION}",
         mode="cmake",
         build_tool_args=[
-            f"-DCMAKE_INSTALL_PREFIX={DEPS_DIR}/install/tbb-{TBB_VERSION}",
+            f"-DCMAKE_INSTALL_PREFIX={DEPS_DIR}/install/oneTBB-{TBB_VERSION}",
             f"-DTBB_TEST=OFF"
         ],
         download_url="https://github.com/oneapi-src/oneTBB",
@@ -759,25 +766,27 @@ if "usd" in targets:
         revision=f"v{TBB_VERSION}"
     )

+if "usd" in targets:
     build_dependency(
         name=f"usd-{USD_VERSION}",
         mode="cmake",
         build_tool_args=[
             f"-DCMAKE_INSTALL_PREFIX={DEPS_DIR}/install/usd-{USD_VERSION}",
             f"-DBOOST_ROOT={DEPS_DIR}/install/boost-{BOOST_VERSION}",
-            f"-DTBB_ROOT_DIR={DEPS_DIR}/install/tbb-{TBB_VERSION}",
+            f"-DTBB_ROOT_DIR={DEPS_DIR}/install/oneTBB-{TBB_VERSION}",
             f"-DPXR_ENABLE_PYTHON_SUPPORT=FALSE",
             f"-DPXR_ENABLE_GL_SUPPORT=FALSE",
             f"-DPXR_BUILD_IMAGING=FALSE",
             f"-DPXR_BUILD_TUTORIALS=FALSE",
             f"-DPXR_BUILD_EXAMPLES=FALSE",
             f"-DPXR_BUILD_USD_TOOLS=FALSE",
-            f"-DPXR_BUILD_TESTS=FALSE"
+            f"-DPXR_BUILD_TESTS=FALSE",
+            f"-DCMAKE_BUILD_TYPE={'Debug' if BUILD_CFG == 'Debug' else 'Release'}"
         ],
-        download_url="https://github.com/PixarAnimationStudios/USD",
+        download_url="https://github.com/aothms/USD",
         download_name="USD",
         download_tool=download_tool_git,
-        revision=f"v{USD_VERSION}"
+        revision=f"brechtvl_onetbb"
     )

 cecho("Building IfcOpenShell:", GREEN)
```

PR #3213, opened 2023-05-29
fix error on load an existing drawing layout created on a different os

maxfb87:
If a layout was created on windows and i try to load it on linux i get this error
image
This fix prevents this error

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/drawing/sheeter.py
+++ b/src/blenderbim/blenderbim/bim/module/drawing/sheeter.py
@@ -28,6 +28,7 @@ import blenderbim.tool as tool
 import ifcopenshell.util.geolocation
 from xml.dom import minidom
 from mathutils import Vector
+from sys import platform
 
 VIEW_TITLE_OFFSET_Y = 5
 DEFAULT_POSITION = Vector((30, 30))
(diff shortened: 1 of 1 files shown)


Moult:
I think this has already been fixed.

maxfb87:
Ah ok, i'll try and i'll write here, thanks

Moult:
More specifically now drawings should always use / separators which works on all OSes. However if you have an older file prior to that update, it may contain \ separators and you need to update it manually.

maxfb87:
@Moult i think that what you said is right for the drawings but sheets (or layouts) still have the bug... in fact, i tried to open a .ifc file on linux, created on windows, and the error still exists (latest BBim version in both os).
If you want to try, i attached the files created on windows
[checksheets.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/11971558/checksheets.zip)

maxfb87:
I noticed that the sheets (called also layouts) are stored in a different way then drawings... the drawings are stored like every other external document but sheets are not and sheets get the path from a w3c property...

Moult:
Ah thanks you're right sheets have an additional check to store them as paths relative to the IFC model, so hopefully this is now actually fixed :)

Issue #3055, opened 2023-04-22
Drawing references files have problems between Windows and Linux


maxfb87:
If i press the button "create drawing" with a drawing created on windows (and i'm on Linux) i get the following error
![image](https://user-images.githubusercontent.com/79401028/233796441-517243b9-5c2f-49b8-823e-a84973c00633.png)
If i look here:
![image](https://user-images.githubusercontent.com/79401028/233796480-601b2ab7-b7b4-477e-9860-6bcf099e0297.png)
i see that the backslash is stored with windows system.
In fact, if i modify the path manually, i solve the error.

The file is the Clothing Boutique ifc https://gitlab.com/openingdesign/clothing_boutique

Moult:
From what I can tell, there are two approaches: prior to storage, or after storage. Prior to storage means we store all paths using /, not \. After storage means we convert any \ to / on non-Windows.

The benefit of prior to storage is that / generally works on any platform, and \ is a valid character on Linux file paths, so it accommodates more cases.

The benefit of after storage is that it accommodates users typing things in using \ and also maybe other IFC software storing document URIs with those characters.

Right now, I'm going for prior to storage.

PR #3214, opened 2023-05-29
New patch recipe

maxfb87:
Basically i had problems creating drawings on linux from a file created on windows because of the different path separator.
So i created an issue #3055 that was solved, but the fix works for new created files, existing ones show the same error.
So here is a simple patch to fix it :-)


Moult:
This is useful, but I'm debating whether or not to bundle this as an IfcPatch given that this is just a temporary issue and we're alpha software. This should either be hosted separately (after all in a couple months time nobody should be experiencing this issue anymore) or if it common, this might be better placed here https://github.com/IfcOpenShell/IfcOpenShell/tree/v0.7.0/src/blenderbim/scripts

maxfb87:
@maxfb87 yeah, i wasn't sure too... I forgot about the script folder, i'll move the file there, thanks!

maxfb87:
@Moult just moved the file and rewrited it as a script

Issue #3167, opened 2023-05-17
Can't import ifcopenshell.geom.app


byildiz:
Hi everyone,

I can't import the package `ifcopenshell.geom.app` and get the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/burakyildiz/.pyenv/versions/ifc/lib/python3.9/site-packages/ifcopenshell/geom/app.py", line 87, in <module>
    from .. import get_supertype
ImportError: cannot import name 'get_supertype' from 'ifcopenshell' (/Users/burakyildiz/.pyenv/versions/ifc/lib/python3.9/site-packages/ifcopenshell/__init__.py)
```

I have a macbook with apple silicon and following setup:
Python version is 3.9.
IfcOpenShell version is 0.7.0.230418.
I installed IfcOpenShell with pip.

I think the problem stems from IfcOpenShell version because when I print IfcOpenShell version I expect `0.7.0.230418` but get `(HEAD detached at 476ab506d)-476ab506d` and it is a weird version. So the following line tries to import `get_supertype` because if check is true. I think `get_supertype` is no more exist in IfcOpenShell version `0.7.0.230418`.
https://github.com/IfcOpenShell/IfcOpenShell/blob/ae9372c3d3c0a1389744403b75719f947cb8f62c/src/ifcopenshell-python/ifcopenshell/geom/app.py#L87

Do you have any idea why IfcOpenShell version is related to git commit hash instead of its actual version?

aothms:
Are you interested in properly reworking this? Should be pretty easy to resolve, you can use the following example to look up the supertype in more recent versions of ifcopenshell:

```
>>> ifcopenshell.ifcopenshell_wrapper.schema_by_name('IFC4').declaration_by_name('IfcWall').supertype().name()
'IfcBuildingElement'
```

> (HEAD detached at 476ab506d)-476ab506d

That is indeed a weird version. How did you install ifcopenshell?

byildiz:
Hi @aothms,

[...]

PR #3215, opened 2023-05-29
fix ifcopenshell.version

byildiz:
This small change fixes ifcopenshell.version issue discussed in #3167.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -897,7 +897,7 @@ find_package (Git)
 if (GIT_FOUND)
 	message("git found: ${GIT_EXECUTABLE} with version ${GIT_VERSION_STRING}")
 	execute_process(
-		COMMAND ${GIT_EXECUTABLE} branch --contains HEAD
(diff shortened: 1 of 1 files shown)


Moult:
Nice! That's weird, so you need to add `-a` because ... HEAD is only on a remote tracking branch? I'd assume naively that once you clone HEAD must always be on a local branch but oh well.

byildiz:
I agree that's weird. I didn't try but I think the problem appears only with some old versions of git. On my mac I tried without `-a` and it lists v0.7.0 branch:

```
$ git branch --contains HEAD
* (HEAD detached at 0cd3cd9e0)
  v0.7.0
```

and with `-a`:

```
$ git branch -a --contains HEAD
* (HEAD detached at 0cd3cd9e0)
  v0.7.0
  remotes/origin/HEAD -> origin/v0.7.0
  remotes/origin/v0.7.0
```

I think the solution is not an elegant one but solves the problem for now.

aothms:
Thank you!

PR #3222, opened 2023-05-31
New 2 buttons on bim tool to create spaces from cursor position and f…

maxfb87:
…rom selected walls. Add also shortcuts (CTRL+G and CTRL+S)


Moult:
Aaah so sorry about this, but actually I think it's better to have a separate "Space Tool" which focuses on spaces, such as :

 1. Generating / updating spaces
 2. Authoring space boundaries
 3. Zoning
 4. Spatial referencing and containment checks (essential for FM)
 5. Non-hierarchical spatial zones (different to regular zones) creation
 6. Managing partial composition and storey elevations

There seem to be enough functions that it warrants its own tool since the BIM Tool is getting a bit crowded.

I've asked @Andrej730 to make a start on this new Space Tool, and one of the firs things is he'll move the "Generate space" as you've done to this tool (along with the existing boundary functions).

maxfb87:
@Moult ah ok, so many new good functions! :-)
I also wanted to add a "toggle visualize button" that switches spaces from solid visualization to wireframe visualization.
Nothing special, but what about if i add this toggle button temporary  below the existing 2 space buttons as a reminder until the new space tool will be created?

Moult:
Go ahead :) Agreed that the space tool should have a few ways to change visibilities.

Issue #3236, opened 2023-06-02
Restructure edit_pset() for ListValue and more


aothms:
Let me rephrase to make sure I got it right, since I'm new to ifc :)

If I understand your idea correctly, we end up with 3 paths:
1. If `value` is an `IfcProperty` or its subclass, then we just append it to the properties
2. If `value` is an `IfcValue`, then we create `IfcPropertySingleValue`, since `NominalValue` must be `IfcValue`
3. Something is wrong and we raise an error?

If that is your idea, then I like it more, and it sounds easy enough to do even for me heh.

_Originally posted by @Booplicate in https://github.com/IfcOpenShell/IfcOpenShell/discussions/3228#discussioncomment-6070927_

aothms:
Just created an issue out of it for more visibility.

I think there are still a two missing "branches". The whole idea of this property template concept is that we can uplift pure python data types to the appropriate measure semantics in IFC.

4. value is a pure python type (str, number). We lookup the appropriate data type from the template so that it is a proper IfcValue-selected type. E.g `float` -> IfcAreaMeasure
5. (4b) We do not have a template and we just assign a default sensible type. E.g `str` -> IfcLabel, `float` -> IfcReal
6. value is a list, tuple. We lookup the appropriate data type and see if it is a IfcPropertyEnumeratedValue. We check whether the supplied values are valid. Etc.
7. value is a list, tuple. There is no template or the template is a PropertyListValue. We create a propertylistvalue.

CC @Moult @vulevukusej @cvillagrasa can you help structure this logic into a complete and correct branch?

Booplicate:
@aothms during my research I found that while the default template has `TemplateType` set to `P_LISTVALUE`, its `PrimaryMeasureType` is not set to `IfcLabel`. How will we assign a class in this case?

```
[...]

Issue #3187, opened 2023-05-23
express `typeof()` does not includes select referencing the type


aothms:
> for each select data type that the current name appears in as a member of the select list, do the following:
>
> i) add the name of the select data type to the list;

Issue #2693, opened 2023-01-13
Create a static version of the IfcOpenShell hl Python API


aothms:
We have a wonderful high level API, kudos to everybody involved.

We were also all in agreement with the `ifcopenshell.api.run("module.function", ...args)` kind of approach. By centrally invoking all API functionality through a central function we eliminate boiler-plate within the individual API functions and we are in really good shape to run API calls e.g collaboratively on a central server!

However, I think the python ecosystem is more and more evolving towards static type decorations. Our approach makes the required arguments of a call completely opaque at the call site so we can't benefit from the static analysis built into IDEs these days. And, tbh, we've seen quite a couple of test cases failing lately due to mismatching in arguments passed to API functions.

Here's a plan. Why don't we at the cmake install step [0] generate [1] [2] static equivalents of all the API usecases?

Thoughts?

#### `api/aggregate/assign_object.py`

```python
class Usecase:
    def __init__(self, file, product=None, relating_object=None):
        ...
```

#### `api/aggregate/__init__.py`

```python
# existing content
# ...

# appended at install
def assign_object(file : ifcopenshell.file, product : ifcopenshell.entity_instance = None, relating_object : ifcopenshell.entity_instance = None) -> ifcopenshell.entity_instance:
    # [doc_string]
    ifcopenshell.api.run("aggregate.assign_object", file, product=product, relating_object=relating_object)
```

[0] https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcwrap/CMakeLists.txt#L95

[1] I think that would be `install(SCRIPT` or `install(CODE`, but maybe `add_custom_command(<target> POST_BUILD)` or `add_custom_target()` + `add_dependency()` would also work?

[2] For python parsing is it `ast.parse()` or is there sth more convenient specific-purpose we can use?

sanzoghenzo:
Late to the party, as I totally disagree with the api/usecase approach... And [I'm not the only one](https://www.youtube.com/watch?v=o9pEzgHorH0) - A class with only an init and a single method has no reason to exist. And building it "just in case" is not useful, especially to new library users that have to figure out the magic behind `ifcopenshell.api.run`. But anyway...

It seems to me that it would be better to do the opposite: extract the api code to typed, standalone functions and call them from the `Usecase`s.
That way the `importlib`+`getattr(getattr(getattr()))` is only executed by the programs that really need it.

aothms:
[...]

Issue #182, opened 2017-01-17
Error in build-all.py


Gochip:
The build-all.py uses PCRE_VERSION=8.38 which is not found. I believe it should use 8.40.

aothms:
Apparently the download location we used only retains the last two versions. With the release of 8.40 last week the 8.38 release was removed. I changed the download location to sourceforge, which didn't have 8.40 yet, but which will retain the older builds so that we do not need keep updating the build script for every version change. Thanks for the report.

PR #3237, opened 2023-06-02
Attempt to improve pset support

Booplicate:
As per #3236/#3228

1. Adds support to directly passed in property entities
2. Adds support for tuples
3. Adds support for `P_LISTVALUE`

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
@@ -180,6 +180,19 @@ class Usecase:
         """
         return prop.Name in self.settings["properties"]
 
+    def _try_purge(self, prop) -> bool:
+        """
+        Tried to remove the property
+        if successful, returns True, otherwise False
+        NOTE: Assumes the prop exists
+        """
+        if not self.settings["should_purge"]:
+            return False
+
+        del self.settings["properties"][prop.Name]
+        self.file.remove(prop)
+        return True
+
     # TODO - Add support for changing property types?
     #   For example - IfcPropertyEnumeratedValue to
     # IfcPropertySingleValue.  Or maybe the user should
@@ -197,7 +210,7 @@ class Usecase:
 
     def update_existing_enum(self, prop):
         """
-        NOTE: assumes the prop exists
+        NOTE: Assumes the prop exists
         """
         value = self.settings["properties"][prop.Name]
         unit, value = self.unpack_unit_value(value)
@@ -217,9 +230,7 @@ class Usecase:
             and value.is_a("IfcPropertyEnumeratedValue")
         ):
             if not value.EnumerationReference.EnumerationValues:
-                if self.settings["should_purge"]:
-                    del self.settings["properties"][prop.Name]
-                    self.file.remove(prop)
+                if self._try_purge(prop):
                     return
                 prop.EnumerationReference.EnumerationValues = None
                 prop.EnumerationValues = None
@@ -234,14 +245,12 @@ class Usecase:
 
     def update_existing_property(self, prop):
         """
-        NOTE: assumes the prop exists
+        NOTE: Assumes the prop exists
         """
         value = self.settings["properties"][prop.Name]
         unit, value = self.unpack_unit_value(value)
         if value is None:
-            if self.settings["should_purge"]:
-                del self.settings["properties"][prop.Name]
-                self.file.remove(prop)
+            if self._try_purge(prop):
                 return
             prop.NominalValue = None
         elif isinstance(value, ifcopenshell.entity_instance):
--- a/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
@@ -174,9 +174,9 @@ class Usecase:
             self.psetqto = ifcopenshell.util.pset.get_template(self.file.schema)
             self.pset_template = self.psetqto.get_by_name(self.settings["pset"].Name)
 
-    def _has_property(self, prop) -> bool:
+    def _should_update_prop(self, prop) -> bool:
         """
-        Checks if the property already exists
+        Checks if the given property should be changed
         """
         return prop.Name in self.settings["properties"]
 
@@ -199,7 +199,7 @@ class Usecase:
     # just delete the property first? - vulevukusej
     def update_existing_properties(self):
         for prop in self.get_properties():
-            if not self._has_property(prop):
+            if not self._should_update_prop(prop):
                 continue
 
             if prop.is_a("IfcPropertyEnumeratedValue"):
--- a/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
@@ -244,22 +244,47 @@ class Usecase:
                 continue
             unit, value = self.unpack_unit_value(value)
             if isinstance(value, ifcopenshell.entity_instance):
-                if value.is_a(True) == "IFC4.IfcPropertyEnumeratedValue":
+                if value.is_a("IfcProperty"):
                     properties.append(value)
-                    continue
-                else:
+
+                elif not value.is_entity():
                     args = {"Name": name, "NominalValue": value}
                     if unit:
                         args["Unit"] = unit
                     properties.append(
                         self.file.create_entity("IfcPropertySingleValue", **args)
                     )
+
+                else:
+                    raise TypeError(f"'{value}' is not a valid property value for '{name}'")
+
             # TODO-The following "elif" is temporary code, will need to refactor at some point - vulevukusej
-            elif isinstance(value, list):
+            elif isinstance(value, (tuple, list)):
                 if not value:
                     continue
                 for pset_template in self.pset_template.HasPropertyTemplates:
-                    if pset_template.Name == name:
+                    if pset_template.Name != name:
+                        continue
+
+                    if pset_template.TemplateType == "P_LISTVALUE":
+                        # TODO: What class to pick if none was provided? We get an error otherwise:
+                        # ValueError: invalid null reference in method 'new_IfcBaseClass', argument 2 of type 'std::string const &'
+                        ifc_class = getattr(pset_template, "PrimaryMeasureType", None)
+                        if ifc_class is None:
+                            ifc_class = "IfcLabel"
+                        properties.append(
+                                self.file.create_entity(
+                                "IfcPropertyListValue",
+                                Name=name,
+                                ListValues=[
+                                    self.file.create_entity(ifc_class, v)
+                                    for v in value
+                                ]
+                            )
+                        )
+                        break
+
+                    elif pset_template.TemplateType == "P_ENUMERATEDVALUE":
                         prop_enum = self.file.create_entity(
                             "IFCPROPERTYENUMERATION",
                             Name=name,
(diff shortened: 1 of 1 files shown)


src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
cvillagrasa, line 1: #L189 and #L218 -> Move check to #182

#L193 -> Also include tuples when updating existing enums

#L203 -> Maybe check move the `elif isinstance(value, ifcopenshell.entity_instance)` in #L210 one step outwards to the #L202 else, and add `and value.is_a('IfcPropertyEnumeratedValue')`? And leave #L210 as else.

#L177 -> Agree with Vukas comment of allowing for overwritting an IfcPropertyEnumeratedValue with an IfcPropertySingleValue and viceversa. Perhaps it's a good opportunity to rewrite the logic in #L183 and update_existing_enum & update_existing_property. Also foresee the edge case when updating with an empty/None object.

#L204 and #L223 beware DRY on the if purge block, refactor as a `purge_property()` method.

#L185 -> explicitly check `if prop.is_a('IfcPropertySingleValue')`, rename update_existing_property -> update_existing_prop_single_value and include new update_existing_prop_list. Foresee also bounded properties and any other to be implemented IfcProperty.

#L245 unpack_unit_value before the previous `if value is None` to prevent the edge case when passing units but NominalValue=None. This one is on me 😅.

#L333 -> At the end of `get_properties()`, I'd add a general else issuing a warning to let the user know the passed pset object looks odd (easily can save someone some time if a string, a step ID or whatever was mistakely passed instead of a pset entity).

src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
```diff
@@ -244,22 +244,46 @@ def add_new_properties(self):
...
+                elif not value.is_entity():
                     args = {"Name": name, "NominalValue": value}
                     if unit:
                         args["Unit"] = unit
                     properties.append(
                         self.file.create_entity("IfcPropertySingleValue", **args)
                     )
+
+                else:
+                    raise ValueError(f"'{value}' is not a valid property value for '{name}'")
```
cvillagrasa, line 259: Maybe include .is_a() for this edge case TypeError? -> `f"A {value.is_a()} cannot be assigned to the property set {name}"`

src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
```diff
@@ -244,22 +244,46 @@ def add_new_properties(self):
...
-            elif isinstance(value, list):
+            elif isinstance(value, (tuple, list)):
                 if not value:
                     continue
                 for pset_template in self.pset_template.HasPropertyTemplates:
-                    if pset_template.Name == name:
+                    if pset_template.Name != name:
+                        continue
+
+                    if pset_template.TemplateType == "P_LISTVALUE":
```
cvillagrasa, line 269: Support for custom units also here.

src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
```diff
@@ -244,22 +244,46 @@ def add_new_properties(self):
                 continue
             unit, value = self.unpack_unit_value(value)
             if isinstance(value, ifcopenshell.entity_instance):
-                if value.is_a(True) == "IFC4.IfcPropertyEnumeratedValue":
+                if value.is_a("IfcProperty"):
                     properties.append(value)
-                    continue
-                else:
+
+                elif not value.is_entity():
```
cvillagrasa, line 250: @aothms I'm a bit lost in this point. From my ignorance, isn't it possible to do something like value.is_a('IfcValue') for an IfcInteger, IfcLabel, etc.? or this doesn't work for the Resource Layer? I see it throws False. What does `is_entity()` return on an entity_instance?

src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
```diff
@@ -244,22 +244,46 @@ def add_new_properties(self):
...
                     if unit:
                         args["Unit"] = unit
                     properties.append(
                         self.file.create_entity("IfcPropertySingleValue", **args)
                     )
+
+                else:
+                    raise ValueError(f"'{value}' is not a valid property value for '{name}'")
+
             # TODO-The following "elif" is temporary code, will need to refactor at some point - vulevukusej
```
cvillagrasa, line 261: After such a refactor, I think we can safely remove this comment.

src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
```diff
@@ -275,7 +299,12 @@ def add_new_properties(self):
                             EnumerationReference=prop_enum,
                         )
                         properties.append(prop_enum_value)
-                        continue
+                        break
+
+                    raise NotImplementedError(f"Template type '{pset_template.TemplateType}' is not supported yet")
+
+                # TODO: what if no template was found/matched?
```
cvillagrasa, line 306: I vote for foreseeing all possible cases (bounded properties, etc.) and throwing a NotImplementedError while they're not implemented.

cvillagrasa (commented):
I made some general inline comments.

src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
Booplicate: Sure, but I believe `TypeError` is used for wrong Python type used, in this case the type is valid - `entity_instance`, but its ifc class is not. So I lean towards `ValueError` more.

src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
Booplicate: It is unclear what's I'm doing here, I don't know of a better way to check if it's a primitive data type.

`value.is_a("IfcValue")` doesn't work, perhaps because `IfcValue` doesn't actually exist, it's some kind of abstract enumeration/union that holds other abstract enumeration/union that in the end hold actual ifc classes like `IfcLabel`.

`value.is_entity()` indeed works, for entity it returns `True`, for primitives it returns `False`:
https://github.com/IfcOpenShell/IfcOpenShell/blob/032f52d434e13059fd24c12436ec71a645a5108c/src/ifcopenshell-python/ifcopenshell/entity_instance.py#L342-L352

If there's a more clear way of doing this, let me know.

aothms: > value.is_a("IfcValue") doesn't work

If we agree that it *should* work, then I don't mind adding support for this. It's related to #3187

Booplicate: I think that'd be useful, but I don't know if it'd case any overhead.

src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
Booplicate: > #L177 -> Agree with Vukas comment of allowing for overwritting an IfcPropertyEnumeratedValue with an IfcPropertySingleValue and viceversa. Perhaps it's a good opportunity to rewrite the logic in #L183 and update_existing_enum & update_existing_property. Also foresee the edge case when updating with an empty/None object.

This looks complex to me, not sure if I should touch it yet, it might be better to handle separately.

>  include new update_existing_prop_list

Same here, it'd be a big refactoring, but if I figure out a way to implement that method, I will add it.

> #L245 unpack_unit_value before the previous `if value is None` to prevent the edge case when passing units but NominalValue=None.

I'm not sure if I understood you correctly, but I see that it's possible to try to access `None` as a `dict`, I added a sanity check to `unpack_unit_value`. If that's not what you meant, I will undo the change.

Moult:
Many thanks for this! This is desperately needed. Have you considered writing test cases for this?

cvillagrasa:
Hey @Booplicate sorry for the delay. What @Moult suggests is that some tests concerning the newly added functionalities could be added to [the correspondent file](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/test/api/pset/test_edit_pset.py) (perhaps you could add the test cases in which the prop value is a list/tuple, as you posed them in [#3228](https://github.com/IfcOpenShell/IfcOpenShell/discussions/3228)).

Then, with respect to the current version of the PR, I'd still make the following comments:
- (My own view) When dealing with methods of the `Usecase` class, which by design are internal to the API, I wouldn't care so much about docstrings. People looking at those methods won't be beginner users of the library, so I'd just put all the attention in naming variables in an understandable way, hence avoiding the need for the rest of the library to match highly detailed docstrings in internal methods.
- I see that you've introduced some type hints. I'm very in favour of them! but afaik there's currently not any coordinated approach to progressively add them to the code base (and possibly integrate `mypy` checks with the rest of the testing). @aothms @Moult what's your view on this? it's somewhat related to [#2693](https://github.com/IfcOpenShell/IfcOpenShell/issues/2693), but more general, maybe this deserves an issue of its own. I believe that it would be super awesome to be able to type hint something like `ifcopenshell.entity_instance['IfcProduct']`, and then `mypy` to accept an 'IfcWall', but complain of an 'IfcWallType', for instance. Or `ifcopenshell.entity_instance['IfcCartesianPoint', 2]` for a 2D point. I have no idea how hard it is to implement this, although if I were to judge from the timings of `numpy` and `pandas`, it definitely doesn't look easy. Maybe for now it doesn't hurt to start adding type hints of primitive types here and there? That could help to shorten some variable names thanks to the Huffman Coding principle (I'm particularly thinking of removing `should_` prefixes by having a bool type hint, for instance).
- With respect to the `unpack_unit_value`, I'd revert that commit and leave it as before, with one single return statement (I find it more readable that way). What I was saying in the first round of  comments was to swap these lines (very minor detail):

```python
# existing
if value is None:
	continue
unit, value = self.unpack_unit_value(value)

#proposed
unit, value = self.unpack_unit_value(value)
if value is None:
	continue
```


Summarizing, the overall code of `add_new_properties` is looking something like this on this PR (numbers are those of [#3236](https://github.com/IfcOpenShell/IfcOpenShell/issues/3236)):

```python
def add_new_properties(self):
	for name, value in self.settings['properties'].items():
		if value is None: continue
		unpack_unit_value()
		if isinstance(value, ifcopenshell.entity_instance):
			if value.is_a("IfcProperty"): ...  # 1
			elif value.is_a("IfcValue"): ...  # 2, needs some help from @aothms 
			else: raise ValueError  # 3
		elif isinstance(value, (tuple, list)): ...  # 6, 7
		else: ... # create IfcPropertySingleValue using get_primary_measure_type and cast_value_to_primary_measure_type
```

I see we're doing things in two different ways at the same time. I guess we could refactor `get_primary_measure_type` into a `get_prop_template`, plus the rest of the code, and then reuse that `get_prop_template` into the (tuple, list) case. Similarly, the `cast_value_to_primary_measure_type` can be extended to include enumerations and aggregations. And I'd respect the philosophy of having these properly named methods outside of the main logic, keeping it light, which increases readability.


Lastly, maybe for the (tuple, list) / enumeration case, we're currently missing a check of whether the provided value types match the template.

Otherwise, all the rest looks correct to me! ;)

aothms:
@cvillagrasa let's take the typing discussion out #3306 I did some experiments to be able to overlay some typing annotations, but it's really quite hard to decorate it properly.

Also the test cases are failing for this PR. @Booplicate are you still available to look into this further?

aothms:
Merged as part of #3321

Booplicate:
Sorry, didn't have time to look into this.
I actually wasn't sure how to write a test for this, thank you for handling it and fixing the code ❤️

PR #3270, opened 2023-06-07
Parsing IFC files with IFCRelAssociatesMaterial but no IFCMaterial

chetan-set:
There are IFC authoring tools presently in circulation that do not enforce the mandatary parameter of IFCMaterial in IFCRelAssociatesMaterial.

If an IFC file has an IFCRELASSOCIATESMATERIAL but did not supply an IFCMATERIAL, IfcConvert will crash. 

This merge request prevents the crash, catches the exception and prints an error on the terminal.

The discussion thread relating to this MR can be found here: https://github.com/IfcOpenShell/IfcOpenShell/discussions/3257


aothms:
Looks good to me. Thanks for taking the time.

PR #3286, opened 2023-06-10
Update ifccityjson

balazsdukai:
A small update to the ifccityjson script, so that it works with:

- ifcOpenShell 0.7.0
- CityJSON v.1.1
- cjio 0.8


Moult:
Thanks so much!

PR #3307, opened 2023-06-14
Package for ifccityjson

balazsdukai:
Restructure the script into a package, so that it can be embedded in other applications, such as https://github.com/3DGI/cityjson2ifc


aothms:
Looks good to me, thanks Balázs

PR #3315, opened 2023-06-15
Update README.md

Vasile-Corjan:
I have added the parameters "minOccurs" and "maxOccurs"


aothms:
Thanks but according to the latest discussions. maxOccurs=1 is not valid. Could you change it to `unbounded`? The valid configurations are:

https://github.com/buildingSMART/IDS/blob/7abd7ce4631ba3b24a4e030d3b06e6763f9b7165/Documentation/developer-guide.md?plain=1#L68-L72

Vasile-Corjan (commented):
I have changed the value of maxOccurs to "unbounded".

Issue #3253, opened 2023-06-04
Door missing after duplication


theoryshaw:
Duplicating a door, and adding the void to the door, and then duplicating the wall. :)

A video's easier...

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/0165c2ea-5f45-4e4f-9ba9-e1f38e92fc3a

Issue #3068, opened 2023-04-27
[BlenderBIM] - Idea - Duplicating IfcElementAssembly


brunoperdigao:
I created two operators to deal with IfcElementAssembly. The first one copies the element and all its parts in order to create a copy of the whole assembly. The second updates changes made in the original assembly to the other copies. The GIF below shows an example:
![GIF 26-04-2023 13-21-30](https://user-images.githubusercontent.com/57102715/234859484-93443086-782b-4f28-a087-57c7de76dd3d.gif)

I'd like to make sure that it makes sense to implement this feature before creating a PR. The code is still messy, and I'd have to organize a few things before submitting.

If so, a few topics for future development:
1. Where would this go in the UI? The copying operator works like a shift+d and can be used with a keymap.
2. Other operators should be created for selecting the parent objects and to break the link with the parent.
3. This uses most of the same logic as IfcArray, so #3063  is an issue for this too.
4. This could be extended to implement IfcArray for IfcElementAssembly? This is a whole new topic, but I think the only reason why the array is not possible in assembly elements is because the array UI is stored inside blender's modifiers panel, and blender hides them when an empty object is selected.

brunoperdigao:
related: https://community.osarch.org/discussion/comment/15536/#Comment_15536

brunoperdigao:
Hi @Moult,

I'm getting back to this after a while, and I have this branch working: https://github.com/brunoperdigao/IfcOpenShell/tree/duplicate-assembly2
Here's a demo;
![GIF 19-06-2023 17-20-33](https://github.com/IfcOpenShell/IfcOpenShell/assets/57102715/e6d306a2-c5a9-44b1-bb19-efa8cf47868b)


However, I realized that this branch was outdated, and I started working in a new branch that I synced with the main development yesterday. I'm getting this strange behavior where it successfully creates the aggregate but doesn't create a new Blender collection for the new objects:
![GIF 19-06-2023 17-17-35](https://github.com/IfcOpenShell/IfcOpenShell/assets/57102715/7b11218a-597d-4f76-bb95-ad174d5568e4)




I can't figure out what change is causing this issue. Any ideas? Here is the new branch: https://github.com/brunoperdigao/IfcOpenShell/tree/duplicate-aggregate

theoryshaw:
I have no clue, but could it be related to this commit?
[...]

PR #3328, opened 2023-06-20
Duplicate aggregate

### LINKED ISSUE #3186: With aggregates, ifcopeningelements not showing in BB drawings.
We're using aggregates on [this](https://gitlab.com/openingdesign/highland_haven/-/blob/ed715b78194e319b79b778b1d30b4a4bd972766f/Models_and_CAD/Highland_Haven.ifc) project. 

As you can see from the following, the ifcopeningelement doesn't show in the drawing. 

Is there a relative easy way to correct this?

FLOOR PLAN - 1ST FLOOR

2023-05-22_19-13-38_Blender_blender

brunoperdigao:
This PR implements the idea presented in #3068 for duplication of aggregates. It also gives the possibility to refresh the copies if anything is changed in the original aggregates.

To duplicate, you have to select only the `IfcElementAssembly`. Right now, this operator is triggered by `crtl + shift + d`. It also works with sub-aggregates.

https://github.com/IfcOpenShell/IfcOpenShell/assets/57102715/ac112ab3-b30b-456d-af0a-9b0cb1f76e5d

After changing something in the original aggregate, select the `IfcElementAssembly`you want to refresh, press `F3` and type `refresh_aggregate`:

https://github.com/IfcOpenShell/IfcOpenShell/assets/57102715/356f8018-4126-4caa-bb3d-af4051955763

Also works with sub-aggregates copies that are detached:

https://github.com/IfcOpenShell/IfcOpenShell/assets/57102715/bff29454-6520-45d9-8607-34be2ca498d0

The current version might fix #3186 . There is still a problem with openings placement, but I think it's related to duplication in general (related to #3253 ). Nevertheless, this can be solved by refreshing the copy.

https://github.com/IfcOpenShell/IfcOpenShell/assets/57102715/a11c16bc-1061-4945-8cce-bb9c65174ec4

**Future improvements**
This PR is the basic implementation, and it's somewhat ready to use. I'm committed to work on the following improvements moving forward:
- Regarding the code, the operators are placed in the geometry module, because it was easier to start with the `OverrideDuplicate` as reference. Maybe this could evolve and go to the aggregate module.
- Create an ui for refresh operator, maybe here:
image

- Create an operator to refresh all instances
- Create an operator to select the base `IfcElementAssembly` of that group
- Create an operator to select the original instance of that aggregate
- Create an operator to unlink the aggregate from the original instance

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/geometry/__init__.py
+++ b/src/blenderbim/blenderbim/bim/module/geometry/__init__.py
@@ -74,7 +74,7 @@ def register():
         kmi = km.keymap_items.new("bim.override_object_join", "J", "PRESS", ctrl=True)
         kmi = km.keymap_items.new("bim.override_object_duplicate_move_macro", "D", "PRESS", shift=True)
         kmi = km.keymap_items.new("bim.override_object_duplicate_move_linked_macro", "D", "PRESS", alt=True)
-        kmi = km.keymap_items.new("bim.override_object_duplicate_move_aggregate_macro", "F", "PRESS", ctrl=True)
+        kmi = km.keymap_items.new("bim.override_object_duplicate_move_aggregate_macro", "D", "PRESS", ctrl=True, shift=True)
         kmi = km.keymap_items.new("bim.override_paste_buffer", "V", "PRESS", ctrl=True)
         kmi = km.keymap_items.new("bim.override_mode_set_edit", "TAB", "PRESS")
         kmi = km.keymap_items.new("bim.override_object_delete", "X", "PRESS")
--- a/src/blenderbim/blenderbim/bim/module/geometry/__init__.py
+++ b/src/blenderbim/blenderbim/bim/module/geometry/__init__.py
@@ -29,6 +29,9 @@ classes = (
     operator.OverrideDuplicateMoveLinked,
     operator.OverrideDuplicateMoveLinkedMacro,
     operator.OverrideDuplicateMoveMacro,
+    operator.OverrideDuplicateMoveAggregate,
+    operator.OverrideDuplicateMoveAggregateMacro,
+    # operator.RefreshAggregate,
     operator.OverrideJoin,
     operator.OverrideModeSetEdit,
     operator.OverrideModeSetObject,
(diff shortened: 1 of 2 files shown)


brunopostle:
Some comments:

I think there is a single 'parent' aggregate, and the 'children' are all duplicates of this. The children each have a epset indicating the guid of the parent, and the parent has an epset listing all the children. Is this right?

My experience of inkscape is that it has a similar parent/child system, and it it very hard to remember which is the parent since they all look the same. Sketchup components (for example) don't have this heirarchy, or at least it is hidden from the user - you can edit any component and they all behave as 'parent'.

Storing the relationship in two places requires that both entries are kept in sync, which could be fragile.

Could an IfcGroup be used to collect these duplicate aggregates? ie. all the sibling agregates can be found by checking if the current aggregate belongs to an IfcGroup and listing its children.

brunopostle:
Maybe there is a 'make siblings the same as selected' operator. This would need to check all the siblings and only update them if they differ.

brunoperdigao:
How would you suggest that I can check if they differ? I couldn't think of a way to do it and why I opted for a different approach

brunopostle:
Checking would be hard as it needs to recurse, compare attributes, and check indirect stuff like materials and psets. Though however you do this tool, doing a minimal partial update would be preferable to discarding everything and recreating

brunoperdigao:
So, here are my thoughts to move forward with this:
Create a simple "Duplicate Aggregate" that would not create a link between objects. Would work like a `shift+d` in Blender. I think this can be already useful in some cases and I can simplify the code in this PR in order to make it easier to be merged, so the syncing issues can be solved later.

The other operator would be "Duplicate Linked Aggregate". About this, I'm just thinking out loud, an option to simplify things and make it more efficient would be to restrict its use to work only if all the elements in the aggregate are types. That way, most of the syncing problems would be addressed. The "refresh" operator would just have to check the matrix for location and rotation, and other things like profile shapes for slabs, length and height for walls... Makes sense?

brunopostle:
I think you have to try it and see, some types have representations, but others have profilesets or layersets and the representation is in the instance. But I think these are equivalent, a simple representation can be referenced in multiple instances, so your duplicate-linked operator doesn't need to copy any representations, it just copies the element and references the old representation, it can maybe do the same with placements, just use the original placement.

theoryshaw:
Can this be pulled now? Is useful to me, in current state.

Moult:
Thank you so much! Sorry for the delay in merging.

I am always a little hesitant to merge changes related to parts of the system that I don't have confidence in. I know for sure there are some aspects which I personally haven't tested much and are missing a ton of unit tests. Dealing with aggregates in practical scenarios is one of them. I hope I didn't make the wrong choice!

theoryshaw:
Through use, we'll find them. :)

brunoperdigao:
@Moult I understand. Can I help with testing? I'll take a look at the docs to see if I can understand how to do it

Moult:
@brunoperdigao yes please! The first thing to do is make sure you can run tests: https://blenderbim.org/docs/devs/running_tests.html

The most important test for this type of things are BIM Tests (third bullet point on that page). Here are the existing basic aggregate tests: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/test/bim/feature/aggregate.feature

However your tests should be part of the geometry feature, see how we already have tests for duplicating various types of objects in different scenarios: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/test/bim/feature/geometry.feature#L282-L422 you should add tests related to duplicating aggregates.

The definition of those sentences are here https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/test/bim/test_feature.py you will probably need to add some of your own e.g. https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/test/bim/test_feature.py#L199-L202

PR #3359, opened 2023-06-29
New drawing performance enhancement: SVG prefiltering

aothms:
We have long discussions about performance in #1153. One of the things we wanted to try was to use the fact that often in arch drawings you have large faces obscuring a bunch of detailed objects. E.g a higher level floor slab obscuring all the furniture below. This is relatively easy to implement by taking a bunch of these large faces and see if they fully contain in 2d the bounding box of detailed elements. If so, don't feed those detailed elements to the OCCT HLR algo.

Limitations:

- View direction must be orthogonal to large face (can be enhanced in the future by projecting slanted large faces).
- All tolerances and constants currently hard coded and not relative (e.g the a large face is now area > 2m, but we could also for ex take the largest n procent).
- Co-planar faces could be merged prior to participating in occlusion detection
- There is no filtering among the large faces, e.g large faces could be obscuring other large faces (e.g. layered slabs).
- There is no 2d acceleration in occlusion detection, so many large faces next to each other might be problematic
- Code quality is a mess. I had a good stretch of 9h time due to a long train ride, but my time is depleted

Nevertheless first test results are rather promising:

From `2 minutes 36 seconds` to `17 seconds` (4 storeys) for duplex a:

    duplex_a.ifc -y duplex_a.svg --section-height-from-storeys [--svg-prefilter] --svg-project --bounds 1024x1024 --exclude entities IfcOpeningElement

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -414,6 +414,8 @@ int main(int argc, char** argv) {
 			"Stores name and guid in a separate namespace as opposed to data-name, data-guid")
 		("svg-poly",
 			"Uses the polygonal algorithm for hidden line rendering")
+		("svg-prefilter",
+			"Prefilter faces and shapes before feeding to HLR algorithm")
 		("svg-write-poly",
 			"Approximate every curve as polygonal in SVG output")
 		("svg-project",
@@ -1082,6 +1084,7 @@ int main(int argc, char** argv) {
 		}
 		static_cast<SvgSerializer*>(serializer.get())->setUseNamespace(vmap.count("svg-xmlns") > 0);
 		static_cast<SvgSerializer*>(serializer.get())->setUseHlrPoly(vmap.count("svg-poly") > 0);
+		static_cast<SvgSerializer*>(serializer.get())->setUsePrefiltering(vmap.count("svg-prefilter") > 0);
 		static_cast<SvgSerializer*>(serializer.get())->setPolygonal(vmap.count("svg-write-poly") > 0);
 		static_cast<SvgSerializer*>(serializer.get())->setAlwaysProject(vmap.count("svg-project") > 0);
 		static_cast<SvgSerializer*>(serializer.get())->setWithoutStoreys(vmap.count("svg-without-storeys") > 0);
--- a/src/ifcopenshell-python/ifcopenshell/draw.py
+++ b/src/ifcopenshell-python/ifcopenshell/draw.py
@@ -59,6 +59,7 @@ class draw_settings:
     cells: bool = True
     merge_cells: bool = False
     include_projection: bool = True
+    prefilter: bool = True
 
 
 def main(settings, files, iterators=None, merge_projection=True, progress_function=DO_NOTHING):
@@ -130,6 +131,8 @@ def main(settings, files, iterators=None, merge_projection=True, progress_functi
     sr.setNoCSS(not settings.css)
     if settings.subtract_before_hlr:
         sr.setSubtractionSettings(W.ALWAYS)
+
+    sr.setUsePrefiltering(settings.prefilter)
 
     try:
         sh = ["none", "full", "left"].index(settings.storey_heights)
--- a/src/ifcwrap/IfcGeomWrapper.i
+++ b/src/ifcwrap/IfcGeomWrapper.i
@@ -636,6 +636,10 @@ struct ShapeRTTI : public boost::static_visitor<PyObject*>
 	}
 %}
 
+%ignore hlr_writer;
+%ignore hlr_calc;
+%ignore occt_join;
+%ignore prefiltered_hlr;
 %ignore svgfill::svg_to_line_segments;
 %ignore svgfill::line_segments_to_polygons;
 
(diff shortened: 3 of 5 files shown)


Moult:
> I had a good stretch of 9h time due to a long train ride, but my time is depleted

Alrighty giddy up everyone we've got to buy @aothms more train tickets let's go go go!

> From 2 minutes 36 seconds to 17 seconds (4 storeys) for duplex a:

This is super exciting! As soon as a bot build is up I'll bump and I wonder if others producing drawings see improvements too! (ping @theoryshaw @ChubbyQuark @Plae-Blue-Dot )

ChubbyQuark:
Sounds great, I've got some chinky files to test on. I'll keep an eye out for the latest version and run some tests.

ChubbyQuark:
So I've tested on a large building (shell and core only), the drawing appears to get created but the SVG is empty. Works ok on an existing small file I have. No errors in the info panel in Blender

This is the drawing I'm trying to create:

image

Plae-Blue-Dot:
That is 9,17x faster than before! 

I'll try it out with some hefty files once the new build is up! Amazing! I'm in favour of keeping aothms permanently on a train hahha

theoryshaw:
Cool. 

I pulled the latest and tried to print `OPTION 2 - FLOOR PLAN` drawing from the following files, and it spins for awhile, but crashes.  I was able to print this in previous commits. 

files: [Restaurant_Livingston.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/11910861/Restaurant_Livingston.zip)
open Blender file first.

theoryshaw:
It crashes with 3.5, but works with 3.4. 
:)

Moult:
@theoryshaw @ChubbyQuark it isn't available yet, we'll need to wait for the bot build to be ready.

theoryshaw:
sounds good, is this the one? https://github.com/IfcOpenShell/IfcOpenShell/releases/tag/blenderbim-230630

aothms:
https://github.com/IfcOpenBot/IfcOpenShell/commit/fc50bdd3a0354cf332ad846f0a336a9bbfd4a769

aothms:
Edit: keep in mind that it still needs to be enabled in the serializer https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/draw.py#L135

theoryshaw:
I've installed 0.0.230702, is it safe to say this release contains the changes, or do I need to enable something?

Moult:
Not yet, I'll get on this tomorrow and let you know.

Moult:
So I've just bumped the IfcOpenShell build and enabled prefiltering! If anybody wants to test before / after they can toggle this line in the code, restart Blender, and try again: https://github.com/IfcOpenShell/IfcOpenShell/blob/7ab01e6ac0a50504901cca9a749eaf18bde2f625/src/blenderbim/blenderbim/bim/module/drawing/operator.py#L796

I didn't do much tests, but I did run the Schependomlaam model and the TXG model. I think one thing to be clear is that this optimisation only really has an impact if you have detailed meshes being obscured by a large flat thing (i.e. slab). This means that it doesn't really apply to the Schependomlaam or TXG models in plan view because they don't really have that type of scenario so unfortunately I didn't see really any change.

However if I create a elevation of the TXG model and set the camera clip end to 30m (i.e. the entire building) this starts to kick in and basically the drawing is almost twice as fast than before (then again the TXG facade isn't that complex).

2023-07-05-222648_1387x425_scrot


Strangely enough I did notice that when prefiltering was enabled, the linework seemed more correct, compared to disabled (not sure if this is a regression or not, didn't do extensive testing). For example this is disabled:

2023-07-05-222613_851x820_scrot

... vs enabled:

2023-07-05-222622_862x879_scrot

And this is disabled:

2023-07-05-222633_872x708_scrot

... vs enabled:

2023-07-05-222639_917x684_scrot

aothms:
> Strangely enough I did notice that when prefiltering was enabled, the linework seemed more correct, compared to disabled (not sure if this is a regression or not, didn't do extensive testing)

There is also some filtering on excluding faces that point away from the camera, so you'll likely get some less duplicate lines. Do you see this with both shapely and svgfill?

But yes, the main improvement you'll see in cases where there is furniture or other detailed smaller elements.

Moult:
I did a bit more testing and I can confirm there's been some regression when prefiltering is disabled. This is the result using build 8f41ae0 before prefiltering existed (with shapely):

2023-07-06-191252_840x867_scrot

And this is the result using fc50bdd with prefiltering _disabled_ (with shapely):

2023-07-06-191244_836x809_scrot

Note that the output with prefiltering enabled looks identical to the older build 8f41ae0 as expected.

 - 8f41ae0 (no prefiltering, yes shapely) took 18.7 seconds
 - fc50bdd (yes prefiltering, yes shapely) took 17.7 seconds :) but this is not a very fair test since there isn't really any obscured dense linework in this situation.

aothms:
Argh I think it's this https://github.com/IfcOpenShell/IfcOpenShell/commit/2ee0dcaf642897575b20f773f324c2f311af5767

It would be good though to check if this is due to OCCT HLR or shapely. Because in the end all we would have here is some duplicated elements. If duplicated elements cause line work to disappear we better know where, when and why.

Moult:
I don't think it's shapely related since the missing linework occurs even when shapely isn't involved.

aothms:
Ok, So it then that when you have duplicated elements that they filter out each other in the HLR? Or is it only in this specific case when we have multiple instances of exactly the same OCCT geometry, which you would normally not encounter (unless mapped representations?). Well let's keep an eye on it.

PR #3369, opened 2023-07-02
Update operator.py (IFC to CSV Export)

c4rlosdias:
missing output file format in arguments


Moult:
Thank you!

Moult:
Would you be interested in helping write extra format support?

c4rlosdias:
yes of course.  xlsx and ods, right? perhaps the export button should have something like "export" instead of "export csv".

Moult:
Indeed, it would be nice to have a dropdown to select either CSV, XLSX, or ODS, then depending on the format it might show extra options (e.g. delimiter for CSV), then the export button will use whatever format you choose.

For now import only works with CSV.

c4rlosdias:
ok I will work on it

PR #3377, opened 2023-07-04
If ObjectPlacement is missing, then coordinates are 0,0,0

ageeye:
When I want to convert a simple IFC file from FreeCAD to a COBie, I get an error message that the ObjectPlacement is missing. Currently I just set xyz = ["0", "0", "0"]. Whether this complies with the specification, I don't know. Maybe "n/a" would be better than "0". Feedback is welcome.

### DIFF
--- a/src/ifccobie/cobie.py
+++ b/src/ifccobie/cobie.py
@@ -572,15 +572,19 @@ class IfcCobieParser:
         )
         for coordinate in coordinates:
             coordinate_name = "{}/{}".format(coordinate.is_a(), self.get_object_name(coordinate))
+            xyz = ["0", "0", "0"]
+            if coordinate.ObjectPlacement:
+                for axis in range(3):
+                    xyz[axis] = coordinate.ObjectPlacement.RelativePlacement.Location.Coordinates[axis]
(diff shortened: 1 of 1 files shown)


Moult:
Wow, this is a blast from the past when I was first learning IfcOpenShell and none of the utilities and API functions existed. This is also before I did a lot of work with Bill East on facility management and understood it better. In short, I'd like to delete IfcCOBie and replace it with IfcFM, which can do COBie and more, including custom FM deliverables a lot better. I'd recommend not really using IfcCOBie, even though it tries (tried?) to be as close to the old spec as possible (especially as nobody can actually deliver it with proprietary software, even if they claim they can).

For now I've just fixed that one thing, but if you're doing some FM related stuff would you be interested in collaborating and switching to IfcFM? Happy to go over a call if this is all confusing jargon :)

ageeye:
First, thanks for the quick response. I will test your bugfix later. Now, I am taking my first steps with IfcOpenShell. I just need this as a proof of concept. How intensive I will get in remains to be seen. We will see how intensively I get involved.

Moult:
No worries, though FYI the IfcCOBie is a very old utility :)

Issue #3352, opened 2023-06-26
Mathutils not in build depenencies


OrionSehn:
I do not think mathutils is in your current build dependencies for a pip install, and I had to go install the Microsoft Visual C++ Build Tools.
https://pypi.org/project/mathutils/

```
Traceback (most recent call last):
  File "c:\Users\OTSehn\Projects\Cable-Tray-Opt\src\IFCConvert.py", line 15, in <module>
    import ifcopenshell.util.shape_builder
  File "C:\Users\OTSehn\Projects\Cable-Tray-Opt\env\lib\site-packages\ifcopenshell\util\shape_builder.py", line 23, in <module>
    from mathutils import Vector, Matrix
ModuleNotFoundError: No module named 'mathutils'
```
I updated to the latest version on pip (released may.23 https://pypi.org/project/ifcopenshell/0.7.0.230523/) 
I'm not sure if I'm using a wrong version, or if these tools have yet to have been pushed into the build? 

we should offer it as an "extra" (in pip you can offer "extra" flags that pull in more deps)

OrionSehn:
mathutils is used by the ifcopenshell.util.shape_builder.ShapeBuilder module

Moult:
Agreed, it isn't yet in the deps. Would you like to help submit a PR to include it as an extra?

OrionSehn:
Sure I don't mind! I'll add it to my backlog

PR #3378, opened 2023-07-04
Proposed fix to #3352

OrionSehn:
Added optional dependency to pyproject.toml, as a proposed fix to #3352 

Please look this over as this is my first commit to this repo, and I'm not familar with your deployment pipelines yet.

Thank you


Moult:
Looks great! Thanks!

Issue #3332, opened 2023-06-21
If an array modifier is removed gives the user the option to keep all the instances in the scene, or remove them all together. 


theoryshaw:
Would be nice if an array modifier is removed that it gives the user the option to keep all the instances in the scene, or remove them all together. 
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/86c812b4-a585-422b-bfe7-1a8c5b7ff9d0)

brunoperdigao:
I can try that if no one is working on it. This is relatively easy if you have only one array, but gets a little complicated with multiple arrays combination.

Moult:
Is this now resolved?

theoryshaw:
feels good to me. :)

PR #3397, opened 2023-07-06
Proposed implementation of #3332

brunoperdigao:
Allows the users to keep objects when removing the array. In case of  an object that has multiple arrays, it's hard to keep track of the objects that should be removed from the array. For now, to make things simple, this PR only allows removing and array when the user is removing the last array of an object.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/model/array.py
+++ b/src/blenderbim/blenderbim/bim/module/model/array.py
@@ -142,13 +142,9 @@ class RemoveArray(bpy.types.Operator, tool.Ifc.Operator):
     bl_label = "Remove Array"
     bl_options = {"REGISTER", "UNDO"}
     item: bpy.props.IntProperty()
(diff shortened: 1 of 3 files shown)


src/blenderbim/blenderbim/bim/module/model/array.py
```diff
@@ -142,8 +142,13 @@ class RemoveArray(bpy.types.Operator, tool.Ifc.Operator):
     bl_label = "Remove Array"
     bl_options = {"REGISTER", "UNDO"}
     item: bpy.props.IntProperty()
+    total_items: bpy.props.IntProperty()
```
Moult, line 145: Don't pass total_items into the operator. It should work it out itself.

brunoperdigao: I made a new commit. Let me know if it solves the issue.

PR #3413, opened 2023-07-10
Add operator to copy railing modifier parameters

Gorgious56:
I've had a stab at adding an operator to copy Ifc railing modifier parameters from active to selected, however I have trouble actually updating the mesh in the viewport in the final step. Maybe I'm doing it wrong ?

Animation3

todo : 
- Update target railing meshes
- Ensure operator only operates on ifc railings so it doesn't throw all kinds of errors
- Restore selection state after execution


Moult:
Assigning to @Andrej730 since he's familiar with how this part of the code works. Happy to merge and for @Andrej730 to finish that last step ;)

Andrej730:
`update_railing_modifier_bmesh()` is mainly used during the phase when you're editing the railing therefore it's using data from `BIMRailingProperties`. So to make it work you can just update props with `target_props.set_props_kwargs_from_ifc_data(railing_data)` before calling `update_railing_modifier_bmesh(context)`.


https://github.com/IfcOpenShell/IfcOpenShell/assets/9417531/6cba72a4-c841-45cc-9600-d2db2007ec58

PR #3457, opened 2023-07-16
Ifc Systems Navigator (WIP)

chuncahk:
It is still in WIP. **Please do not merge until it is ready**.

I am facing some problems I don't know how to solve, it would be very appreciated if there is a helping hand.
1. I don't know how to integrate the sorting function correctly.
2. Is there any method to register multiple sub-panels?

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/system/__init__.py
+++ b/src/blenderbim/blenderbim/bim/module/system/__init__.py
@@ -45,6 +45,11 @@ classes = (
     ui.BIM_PT_port,
     ui.BIM_UL_systems,
     ui.BIM_UL_object_systems,
+    operator.SelectSystemProductsByGuid,
+    operator.GetIfcSystemInfoByGuid,
+    operator.IfcSystemEditingPanel,
+    ui.BIM_PT_systems_navigator_main_panel,
+    ui.BIM_PT_systems_navigator_sub_panel
 )
 
 
--- a/src/blenderbim/blenderbim/bim/module/system/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/system/operator.py
@@ -204,3 +204,107 @@ class SetFlowDirection(bpy.types.Operator, Operator):
         core.set_flow_direction(
             tool.Ifc, tool.System, port=tool.Ifc.get_entity(context.active_object), direction=self.direction
         )
+
+
+class SelectSystemProductsByGuid(bpy.types.Operator, Operator):
+    """Select All IFC Products in the IFC System"""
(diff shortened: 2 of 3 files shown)


Andrej730:
I don't think creating panels dynamically is a good approach here, I think all panels should be registered when you start the addon and then update their UI as IFC data changes.

1) Sorting panels or ...?
2) register_class?

chuncahk:
The main panel is registered when the addon starts.....but all the subpanels are registered as per the number of IfcSystem in the file. 
The script was running fine when I test it in Blender, but I don't know how to implement merge this into the main branch....
1. What the script doing is extracting all the name of IfcSystems in the file and sort them using def get_ifc_systems_in_current_file()...but I am not sure where should I place it
2. I am not sure how should I register all the sub panels....

Gorgious56:
Maybe sub-panels is not the right approach here, you could study how dynamic collections are done in the addon, for example the pset elements. 

image

rows are added dynamically to the main panel layout and then you can click on the pen icon to enable edition. 
You can also use UIList for 1d or tree-like structure

image

chuncahk:
> 

Thanks for the insight, I agree that the current approach may not be the best.
Dynamic collections seems possible, but I want all the system types can be collapsed so it won't be too long in the UI.

Moult:
The UI Lists can handle tree structures. It's implemented in quite a few spots - like in schedules, or the spatial manager.

chuncahk:
> 

Then using UIList sounds more promising
Is it possible to add buttons inside the UIList?

Andrej730:
@chuncahk sure

image

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/profile/ui.py#L115-L123

chuncahk:
> @chuncahk sure
> 
> image
> 
> https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/module/profile/ui.py#L115-L123

Thank you. I'll see what I can do

chuncahk:
Hi guys, I need a little more help please.
I am taking spatial manager as reference to build a tree structures UIList.
but the list goes like this....did I do something wrong...? Is there a simpler example for tree structure UIList?
Thanks

image

Andrej730:
> it is basically faking indent instead of a real tree structure uilist

@chuncahk yes, it's just adding indent using "BLANK" icons depending on hierarchy level.

image

chuncahk:
I need a second thought about the approach base on the new version of Bonsai

Issue #3466, opened 2023-07-18
Error when creating IfcDistributionElementType type


theoryshaw:
Error when creating type...

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/b3c237bf-fd13-4545-b5ff-241c82e8c23c

Andrej730:
Same errors happens also for 'IfcDistributionElementType', 'IfcFurnishingElementType' and 'IfcBuiltElement' (in IFC4x3)
I think we had the similar issue with IfcCivilElementType and it was fixed by manually adding it to the mapping 3ce010d, related to https://github.com/buildingSMART/IFC4.3.x-development/issues/583. Can fix it the same way.

Also noted that IfcTypeProduct is also not present in the entity to type mapping, don't know if there is any issue with that.

PR #3475, opened 2023-07-19
Fix index error in product.py

BIMvoice:
This pull request resolves an index error in `product.py` (see #3466 for details). The error occurred because the code was trying to access the first element of an empty list. I added a check to ensure the list is not empty before trying to access its elements.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/model/product.py
+++ b/src/blenderbim/blenderbim/bim/module/model/product.py
@@ -19,7 +19,6 @@
 import bpy
 import mathutils
 from functools import reduce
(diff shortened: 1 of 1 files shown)


Moult:
Sorry but this fix may not be appropriate. Have you got a file that exhibits the problem so I can check?

BIMvoice:
This happens when creating a new project with Demo Library and trying to assign that class to a cube. It's not any specific file, you can try with a new one to see the error like Ryan shows in his gif.

Moult:
Closing this as the fix isn't appropriate.

PR #3520, opened 2023-07-28
add ids html report template percentage bars per section

ay-ex:
adding html report template percentage bars to each section. 
@Moult I guess there is a more elegant solution for the percentage bar regarding the css (even without js) 
so I am sure it can be improved. but imho this already helps a lot when scrolling through large reports:

image


Moult:
Whoah whoah there I just checked the code... HTML has a progress element!

ay-ex:
Thank you, it saw that it does. 
But could not get it to work nicely (consistent style with existing elements) with the existing formatting.

PR #3522, opened 2023-07-28
Add new api function that allows to create arbitrary profiles with voids

maxfb87:


Moult:
Although I've merged this, have you seen the relatively recent shape_builder utility? It's far more granular than the API. Do you think it's worthwhile to keep these API functions or should they be deprecated and removed in favour of the shape_builder?

maxfb87:
Ok thanks, i'll take a look to shape builder and i'll tell you

maxfb87:
Ok, i tried to use profile builder to create profiles like the ones that i created with the new api and indeed shape builder is simpler and it can be used for more usecases.
However, i would keep the new api just because it's already done maybe it could be useful for somebody, as well as the other profile api functions.

PR #3523, opened 2023-07-30
Imporve texture support for IfcTessellatedFaceSet

### LINKED ISSUE #2021: Getting UV coordinates of representations
Right now, it seems there is no UV support such as for IfcIndexedTriangleTextureMap in the geom iterator / create_shape. This means that we'd either bypass the iterator or kinda reverse engineer UV coords afterwards (which probably wouldn't be super successful - one potentially welded vertex can correlate to multiple UV coordinates, equivalent to the number of polygons it is part of).

Ideally, the support would need to be built into C++. But I can also understand this would be tricky.

Here are some C++ locations I think would make sense to start looking, but I have no idea if I know enough to succeed here.

 - This triangulation function should capture uvs similar to how it captures material_ids https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcgeom/IfcGeomRepresentation.h#L139
 - Maybe a new UVs parameter here? https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcgeom/IfcRepresentationShapeItem.h#L29
 - Make sure UVs propagate through mapped items by bubbling the uvs here: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcgeom/IfcGeomShapes.cpp#L840
 - `create_brep_for_representation_and_product` to be updated to include UVs. Is this the right spot to actually grab the UV coordinates to begin with? https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcgeom/IfcGeomFunctions.cpp#L2157

wolfboyyang:
Temporary and partial solution to fix #2021 

In my prototyping project, I'm trying to improve the texture support for IFC model to make it looks more realistic in VR/AR application. I start with [Understanding how textures and shaders work in IFC](https://community.osarch.org/discussion/877/understanding-how-textures-and-shaders-work-in-ifc) and the work of İlkin Kavi. In this PR, I'm trying to read the uv coordinates ([IfcIndexedTriangleTextureMap](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcIndexedTriangleTextureMap.htm)/[IfcIndexedPolygonalTextureMap](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcIndexedPolygonalTextureMap.htm)) from ifc file and apply to the cresponding faces for [IfcTriangulatedFaceSet](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcTriangulatedFaceSet.htm)/[IfcPolygonalFaceSet](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcPolygonalFaceSet.htm).

Main steps:
1. remap the vertex index of ifc model to blender mesh
2. find the cresponding face (after remap) in ifc model for blender mesh face
3. remap the texture coordinate to the blender face loop
4. apply uv to blender mesh via face loop

Issues:
1. ifcOpenShell seems triangulated IfcPolygonalFaceSet and the clean mesh step is too late for load the mesh material as I choose to load the texture map before append material to the mesh.
2. set blend_method "BLEND" better only when transparency is set in ifc file as the Alpha Blend mode looks bit strange for tessellation-with-image-texture sample.
3. Image Texture Extension method support for Extend (Clip/Mirror) as [Image Texure Node](https://docs.blender.org/manual/en/latest/modeling/geometry_nodes/texture/image.html) to make it looks properly for the sample IfcIndexedTriangleTextureMap and IfcIndexedPolygonalTextureMap.

Test files: [test-tessellation-with-image-texture](https://github.com/IfcOpenShell/IfcOpenShell/files/12016060/test-tessellation-with-image-texture.zip)
Referrence: [tessellation-with-image-texture](https://github.com/buildingSMART/Sample-Test-Files/blob/master/IFC%204.0/BuildingSMARTSpec/tessellation-with-image-texture.ifc)

test_tessellation-with-image-texture_IfcIndexedPolygonalTextureMap_4x3.ifc
image
tessellation-with-image-texture.ifc
image


Moult:
Is there a way to implement this without bpy.ops.mesh.tris_convert_to_quads()? It sounds like a pretty big assumption to assume that the UVs are only applicable to quads and not any other ngon.

wolfboyyang:
> Is there a way to implement this without bpy.ops.mesh.tris_convert_to_quads()? It sounds like a pretty big assumption to assume that the UVs are only applicable to quads and not any other ngon.

Thanks for your suggestion. I was stuck with this problem for some time when I noticed the triangulation issue of IfcPolygonalFaceSet. I removed the tris_convert_to_quads from my code and by fitting the triangle in to polygon which defined in ifc file, a mix of triangles and quads (and more n-polygon) in IfcPolygonalFaceSet should load properly now.

With this test file [test_quads_tris_mix.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/12208721/test_quads_tris_mix.zip), which just divide the top quad in to two triangles while other 5 face still in quads.

test_tessellation-with-image-texture_4x3_mix_quads_tris.ifc
image

test_tessellation-with-image-texture_4x3_mix_quads_tris.ifc with should_clean_mesh = False
image

Moult:
This is incredible work, I will be sure to test it out and I'm sure many users will find this incredibly useful! A huge thank you!

Baruchware:
@wolfboyyang hello ser, I am working on something similar, could I contact you privately?

theoryshaw:
This is definitely cool.  Is this only on import, so far?

wolfboyyang:
What’s your issue? It would be better to discuss with the community as I’m
only a newbie in BIM although I don’t mind you contacting me directly.

theoryshaw:
Was just wondering if there's a way to format the shader nodes anew in Blender, and then export this out into the IFC?

wolfboyyang:
I started my work from this discussion
https://community.osarch.org/discussion/877/understanding-how-textures-and-shaders-work-in-ifc
which may also give you some idea if you hadn’t read it.

and in another discussion Moult said “As a related issue to texture
support, BlenderBIM32 <https://blenderbim.org/> has now implemented
improved support for externally defined materials. You can link in a
“material passport”, which defines shaders for multiple rendering engines.
This allows you to link in not just textures, but highly complex material
shaders for multiple uses, such as traditional visualisation, real time
game engine views, and light simulation. Please note that this is purely a
link to external shader standards, not the built-in IFC texture support,
but it is highly useful nonetheless and therefore may be relevant to this
discussion.”
(
https://forums.buildingsmart.org/t/ifc-problem-about-elements-with-texture/1157/20
)

As IFC is designed as an exchange format, it is less useful to include some
specific shader node configuration, even it may work for your case.

Issue #3547, opened 2023-08-05
Add_profile_representation and clippings


maxfb87:
I think that there is a problem with "add_profile_representation" api and clippings.
In order to recreate the problem, try to create a new project with demo library and add a new column on 0,0,0 with the length 5m
After that, run this script

import blenderbim.tool as tool
import numpy as np
import ifcopenshell

model = tool.Ifc.get()
obj = C.active_object
entity = tool.Ifc.get_entity(obj)
representation = entity.Representation[2][1]
matrix = np.eye(4)
matrix = ifcopenshell.util.placement.rotation(45,"X") @ matrix
matrix[:,3][0:3] = (0, 0, 3)
matrix = matrix.tolist()
ifcopenshell.api.run("geometry.add_boolean", model, representation = representation, type = "IfcHalfSpaceSolid", matrix = matrix)

Basically i want to apply a simple clipping plane on the column.
After that, try to regen the object: i get the same error as reported [here](https://community.osarch.org/discussion/comment/16767/#Comment_16767)

I think that the problem is in the "add_profile_representation" api.
I have created a possible solution and, if you want, i'll do a PR but, before that, i wanted to be sure that this is actually a bug...
Thanks,
Massimo

Moult:
Yes, it's a bug :) PR very welcome!

Moult:
I ended up doing this slightly simpler and using the ifcopenshell.util.element.copy which preserves the operation type instead of hardcoding DIFFERENCE and also fixing the same bug in the wall representation function.

PR #3548, opened 2023-08-05
Fix bug #3547

maxfb87:
I tried to resolve the bug #3547 and actually this works but i'm not sure if this solution covers all cases...
Also, the comment [here](https://github.com/IfcOpenShell/IfcOpenShell/blob/79584dba13947442f15ed37d290b67821492c98c/src/ifcopenshell-python/ifcopenshell/api/geometry/add_profile_representation.py#L32)  looks suspicious ...


Moult:
Your suspicions about the comment are correct. Basically clippings should be able to be specified in two ways: either by an existing IFC entity, or by a regular Python dictionary with parameters about operand type and matrix. Currently, we only support the latter. In your fix, you've made it only support the former. Can you make it support both via an isinstance(foo, dict) or isinstance(foo, ifcopenshell.entity_instance) check?

maxfb87:
All right, i've done it and thanks for the explanation :-)

PR #3553, opened 2023-08-07
Update create_2pt_wall.py

arunkvijay:
included unit scale for use cases other than si unit


Moult:
Should there be an is_si kwarg to control this?

arunkvijay:
Hi Moult, 
I have included is_si kwarg to select between si unit and other units.
Also, I have changed the dtype of p1 and p2 array to float incase they are entered as integer values (which was causing an error while calculating unit vector `v`)

Issue #280, opened 2017-10-24
update viewer app to pyqt5


thorade:
The latest Anaconda comes with pyqt5 by default,
and to my understanding pythonocc prefers pyqt5 over pyqt4.

It is not possible to install pyqt4 and pyqt5 in parallel, one has to use virtual environments.

The viewer app currently uses pyqt4, and it would be awesome if it could be updated to pyqt5.
https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcopenshell-python/ifcopenshell/geom/app.py
I do not know how easy or difficult that would be...

aothms:
I agree. @jf--- has also brought this up in the past.

>  I do not know how easy or difficult that would be...

I can't imagine it to be too hard. But it will be a while until I will find the time. If you want to take the lead on this I can pitch in where needed.

thorade:
Would it have to support pyqt4 and pyqt5, or is it ok to supprt pyqt5 ONLY?

aothms:
> is it ok to supprt pyqt5 ONLY

I'd be fine with that. There's limited resources in a project like this so effort needs to be spent wisely.

http://blog.qt.io/blog/2014/11/27/qt-4-8-x-support-to-be-extended-for-another-year/

thorade:
Here is a list of changes from pyqt4 to pyqt5:
http://pyqt.sourceforge.net/Docs/PyQt5/pyqt4_differences.html

Out of that list, I so far found two changes that are relevant:
 - QtGui has been split into QtGui and QtWidgets
    http://pyqt.sourceforge.net/Docs/PyQt5/pyqt4_differences.html#qtgui-module
    Whenever something is no longer found in QtGui, it has probably been moved to QtWidgets, so this is easy to fix.
 - pyqt5 no longer supports old-style signals:
    http://pyqt.sourceforge.net/Docs/PyQt5/pyqt4_differences.html#old-style-signals-and-slots
    This will be more work to fix.

aothms:
Many thanks for looking into this.

> pyqt4 supports old-style and new-style signals and slots, pyqt5 only supports new-style:
> http://pyqt.sourceforge.net/Docs/PyQt4/new_style_signals_slots.html
> http://pyqt.sourceforge.net/Docs/PyQt5/pyqt4_differences.html#old-style-signals-and-slots
> This will be more work to change.

[...]

PR #3574, opened 2023-08-11
Updated to use Qt6; Created initial structure of the viewer with Open…

dushyant-basson:
Updated CMake to use Qt6 and include the new files in the executable. The MainWindow now loads a basic container with  QOpenGLWidget set as the central widget. Open file dialog works. 
ParseIfcFile - shall contain code to parse the IFC and extract geometry & data.
IfcViewerWidget - shall deal with rendering the parsed geometry.


Moult:
Nice! I think this addresses #280 - a very old, old bug :)

src/qtviewer/ParseIfcFile.cpp
```diff
@@ -0,0 +1,46 @@
+#include "ParseIfcFile.h"
+
+#include "../ifcparse/Ifc2x3.h"
+#define IfcSchema Ifc2x3
+
```
aothms, line 5:
```suggestion

```

Including a schema in the C++ code is not necessary. It only increases compile time. You're not referencing the Ifc2x3 namespace. (You shouldn't also, because it would make the code specific to a single schema.)

src/qtviewer/ParseIfcFile.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "../ifcgeom/IfcGeom.h"
```
aothms, line 6:
```suggestion

```

This ought to be better documented in IfcOpenShell, but basically stay away from the ../ifcgeom/ folder. Only use `../ifcgeom_schema_agnostic`. That way the IfcSchema define is not necessary and the viewer can open any supported schema.

aothms (changes requested):

aothms:
Thanks @dushyant-basson it's exciting to see this progressing!

PR #3615, opened 2023-08-20
Fix boolean clipping

maxfb87:
I tried to fix the problem created after commit https://github.com/IfcOpenShell/IfcOpenShell/commit/d26a2580ea5a165c0e26b9f2290a0644edcb30b8 and also i made two tool tests.
Not sure btw if the fix is correct...

### DIFF
--- a/src/blenderbim/blenderbim/tool/model.py
+++ b/src/blenderbim/blenderbim/tool/model.py
@@ -489,10 +489,8 @@ class Model(blenderbim.core.tool.Model):
         items = list(body.Items)
         while items:
             item = items.pop()
-            if item.is_a() == "IfcBooleanResult":
+            if item.is_a() == "IfcBooleanResult" or item.is_a() == "IfcBooleanClippingResult":
                 booleans.append(item)
-                items.append(item.FirstOperand)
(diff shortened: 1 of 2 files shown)


Moult:
This is quite a terrible explanation, and we really need a better solution:

BBIM sees two types of booleans. Automatic booleans (e.g. you connect two walls together, and so an automatic boolean clipping is added, or you join wall to roof and the roof clips the wall). And manual boolens (e.g. you manually create a shape or clipping plane and apply that boolean (could be difference, intersection, etc).

The idea is that lets say you extend a wall. You want the automatic boolean to be updated (or removed) based on how you extend that wall. However, the manually added (explicit) booleans should remain.

Right now, we (very badly) assume that if it's an IfcBooleanClippingResult, it's automatic / parametric, and if it's just an IfcBooleanResult, it's manual / explicit. This allows us to isolate the automatic ones and update them, and preserve the manual ones.

This is not a long term solution. Especially because clipping results restrict us to clipping planes. There may be an alternative solution in storing explicit booleans in a pset (ID based, also terrible), or ifc constraints / ifc reference abuse, or something else entirely. We definitely should track this as a bug so it doesn't grow too nasty and bite us later.

PR #3638, opened 2023-08-25
Give each accessor its own buffer view to increase the memory and rendering efficiency of glb files

chetan-set:
Currently, generated glb files in IfcOpenShell will have two buffer views regardless of how many accessors there are.  Creating a buffer view for each accessor makes the rendering and memory effeciency better by allowing the renderer to load and unload vertices, indices and textures as needed. This commit creates a parity between accessors and buffer views.


src/serializers/GltfSerializer.cpp
```diff
@@ -121,17 +121,28 @@ const uint32_t component_type<int>::value = CT_UNSIGNED_INT;
...
 
-	accessor["bufferView"] = N == 1 ? 0 : 1;
-	accessor["byteOffset"] = (size_t)ofs.tellp();
+	accessor["bufferView"] = bufferViewId;
+	accessor["byteOffset"] = 0;
 	accessor["componentType"] = component_type<typename It::value_type>::value;
 	accessor["count"] = num;
 
+	if (N == 1) {
+		j["bufferViews"].push_back({ {"buffer", 0}, {"byteOffset", (size_t)ofs.tellp()}, { "byteLength", num *  4}, {"target", 34963} });
```
aothms, line 138: Could you create a C++ constants for the glTF spec values you use https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_bufferview_target

Like we have here https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/serializers/GltfSerializer.cpp#L32

src/serializers/GltfSerializer.cpp
```diff
@@ -121,17 +121,28 @@ const uint32_t component_type<int>::value = CT_UNSIGNED_INT;
...
+	}
+	else {
```
aothms, line 140:
```suggestion
	} else {
```

We don't have a proper C++ style guide yet. But this is more inline.

src/serializers/GltfSerializer.cpp
```diff
@@ -325,8 +336,12 @@ void GltfSerializer::finalize() {
 	json scene_0;
 	scene_0["nodes"] = node_array_;
 	json_["scenes"].push_back(scene_0);
-	json_["bufferViews"].push_back({ {"buffer", 0}, { "byteLength", indices_length } });
-	json_["bufferViews"].push_back({ {"buffer", 0}, {"byteStride", 12}, { "byteOffset", indices_length },  { "byteLength", binary_length - indices_length } });
+
+	for (auto &n : json_["bufferViews"]) {
+		if (n.contains("byteStride"))
+			n["byteOffset"] = (int)n["byteOffset"] + indices_length;
+	}
```
aothms, line 343:
```suggestion
		if (n.contains("byteStride")) {
			n["byteOffset"] = (int)n["byteOffset"] + indices_length;
		}
	}
```

We don't have a proper C++ style guide yet. But this is more inline.

src/serializers/GltfSerializer.cpp
```diff
@@ -325,8 +336,12 @@ void GltfSerializer::finalize() {
 	json scene_0;
 	scene_0["nodes"] = node_array_;
 	json_["scenes"].push_back(scene_0);
-	json_["bufferViews"].push_back({ {"buffer", 0}, { "byteLength", indices_length } });
-	json_["bufferViews"].push_back({ {"buffer", 0}, {"byteStride", 12}, { "byteOffset", indices_length },  { "byteLength", binary_length - indices_length } });
+
+	for (auto &n : json_["bufferViews"]) {
```
aothms, line 340: Could you add a comment to why this is necessary for the next reader. When the accessors are first written you don't know the size of the indices yet I guess...

src/serializers/GltfSerializer.cpp
```diff
@@ -121,17 +121,28 @@ const uint32_t component_type<int>::value = CT_UNSIGNED_INT;
 template <>
 const uint32_t component_type<float>::value = CT_FLOAT;
 
+static int bufferViewId = 0;
```
aothms, line 124: For this we need to find another solution though. It's possible in Python to have multiple gltf serializers open at once. They can't use the same static counter. It seems every call to write_accessor() is going to increment bufferViewid with 1 so I would recommend moving the counter to a local variable at the callsite of write_accessor and passing it as a function parameter (didn't check if possible, but something like that).

aothms (changes requested):

aothms:
Thanks @chetan-set I trust you on the usefulness of this. See review remarks.

src/serializers/GltfSerializer.cpp
chetan-set: ok

src/serializers/GltfSerializer.cpp
chetan-set: ok

src/serializers/GltfSerializer.cpp
chetan-set: ok

src/serializers/GltfSerializer.cpp
chetan-set: ok

src/serializers/GltfSerializer.cpp
chetan-set: ok sure. I have changed bufferViewId into a member variable and I now send it as a parameter to write_accessor.

chetan-set:
Thank you for your feedback @aothms. I have addressed your comments and updated my pull request.

aothms:
Awesome, thanks for taking the time @chetan-set really appreciated :)

PR #3699, opened 2023-09-03
Fix compiler warnings

dirkolbrich:
fix for the following compiler warnings using llvm clang 16.0.2:
```shell
/IfcOpenShell/src/ifcgeom/../ifcgeom_schema_agnostic/../ifcgeom_schema_agnostic/IfcGeomRepresentation.h:162:7: warning: field 'styles_' will be initialized after field '_item_ids' [-Wreorder-ctor]
                                , styles_(styles)
                                  ^~~~~~~~~~~~~~~
                                  _item_ids(item_ids)
```

```shell
/IfcOpenShell/src/ifcgeom/../ifcgeom_schema_agnostic/../ifcparse/IfcFile.h:275:31: warning: 'const' type qualifier on return type has no effect [-Wignored-qualifiers]
        const IfcParse::declaration* const ifcroot_type() const { return ifcroot_type_; }
                                     ^~~~~~
```

```shell
/IfcOpenShell/src/ifcgeom/IfcGeom.cpp:1087:7: warning: variable 'edge_count' set but not used [-Wunused-but-set-variable]
                int edge_count = 0;
```

### DIFF
--- a/src/ifcgeom/IfcGeom.cpp
+++ b/src/ifcgeom/IfcGeom.cpp
@@ -1084,11 +1084,9 @@ bool IfcGeom::Kernel::convert_layerset(const IfcSchema::IfcProduct* product, std
 
 		TopExp_Explorer exp(axis_shape, TopAbs_EDGE);
 		TopoDS_Edge axis_edge;
-		int edge_count = 0;
 
 		if (exp.More()) {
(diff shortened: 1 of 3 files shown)


aothms:
Thanks!

PR #3708, opened 2023-09-04
cmake: reorder build options

dirkolbrich:
- reorder build option into logical groups
- set conditions for minimal build in one place

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -36,35 +36,50 @@ endif()
 set(EXTRA_VERSION "-alpha.3")
 
 option(MINIMAL_BUILD "The build is to make a minimal version of IFC converter from OCCT into IFC." OFF)
-option(COLLADA_SUPPORT "Build IfcConvert with COLLADA support (requires OpenCOLLADA)." ON)
-option(IFCXML_SUPPORT "Build IfcParse with ifcXML support (requires libxml2)." ON)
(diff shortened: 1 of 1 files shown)


cmake/CMakeLists.txt
```diff
@@ -36,35 +36,50 @@ endif()
...
-option(WASM_BUILD OFF)
+option(USERSPACE_PYTHON_PREFIX "Installs IfcPython for the current user only instead of system-wide." OFF)
+option(ADD_COMMIT_SHA "Add commit sha and branch in version number, warning results in many rebuilds, requires git" OFF)
 
-if(${BUILD_CONVERT})
-    option(GLTF_SUPPORT "Build IfcConvert with glTF support (requires json.hpp)." OFF)
-    option(USD_SUPPORT "Build IfcConvert with USD support (requires pixar's USD library)." OFF)
+if(MINIMAL_BUILD)
+    message(STATUS "Setting options for minimal build")
+    set(BUILD_CONVERT OFF)
```
aothms, line 69: I think MINIMAL_BUILD did not exclude building IfcConvert.

aothms (changes requested):

cmake/CMakeLists.txt
dirkolbrich: Yes, you are right. It check that BUILD_CONVERT was set if GLTF_SUPPORT was set in line 173. Will delete this line.

PR #3739, opened 2023-09-12
Viewer loads and displays IFC geometry; Application icons implemented

dushyant-basson:
The viewer now loads and displays the IFC geometry in the OpenGL container. Implemented OpenSceneGraph for OpenGL rendering in the Qt widget. Added application icons.


src/qtviewer/IfcViewerWidget.cpp
```diff
@@ -1,27 +1,170 @@
...
+    m_viewer->setCameraManipulator(m_mouseHandler);
+}
+
+void IfcViewerWidget::prepareSceneWithGeometry(osg::ref_ptr<osg::Geode> geode)
+{
+    // Apply a rotation to the model
+    // MatrixTransform allows applying transformations (translation, rotation, scaling) to its children
+    osg::ref_ptr<osg::MatrixTransform> rotationTransform = new osg::MatrixTransform;
+    rotationTransform->setMatrix(osg::Matrix::rotate(osg::DegreesToRadians(45.0), osg::Vec3(0.0, 0.0, 1.0)));
+    rotationTransform->addChild(geode);
```
aothms, line 166: I would propose to solve this in the camera orientation as opposed to rotating the model. For example, if we later add a X,Y,Z axis indicator it wouldn't be in sync with the model when rotated. Maybe use something like `ComputeBoundsVisitor` to get a bounding box of the scene and `viewer->getCamera()->setViewMatrixAsLookAt()`. Take the bounding box center as the lookat target, center + radius * some_vector as the lookat eye position.

src/qtviewer/ParseIfcFile.cpp
```diff
@@ -21,22 +31,145 @@ void ParseIfcFile::Parse(const std::string& filePath)
...
 
     do {
         //const IfcGeom::BRepElement* bRepElem = it->get_native();
         const IfcGeom::TriangulationElement* triElem = static_cast<const IfcGeom::TriangulationElement*>(it->get());
-        outputMsg(triElem->type() + ": " + triElem->name());
+        MessageLogger::log(triElem->type() + ": " + triElem->name());
 
         const boost::shared_ptr<IfcGeom::Representation::Triangulation>& triElemGeom = triElem->geometry_pointer();
 
+        const std::vector<int>& elemFaces = triElemGeom->faces();
```
aothms, line 46: `triElemGeom->material_ids()` will give you a mapping of triangle (a 3-length subarray of faces) to an index in `triElemGeom->materials()`. So if you want to split by material. You can do something like:

```C++
std::vector<std::vector<int>> grouped_faces;
auto fit = triElemGeom->faces().begin();
for (auto& mid : triElemGeom->material_ids()) {
    while (grouped_faces.size() <= mid) {
        grouped_faces.emplace_back();
    }
    for (int i = 0; i < 3; ++i) {
        grouped_faces[mid].push_back(fit++);
    }
}
```

src/qtviewer/ParseIfcFile.cpp
```diff
@@ -1,18 +1,28 @@
...
-
-void ParseIfcFile::Parse(const std::string& filePath)
-{
+bool ParseIfcFile::Parse(
+    const std::string& filePath,
+    std::vector<osg::ref_ptr<osg::Geometry>>& geometries
+) {
     IfcParse::IfcFile file(filePath);
 
     IfcGeom::IteratorSettings settings;
```
aothms, line 28:
```suggestion
    IfcGeom::IteratorSettings settings;
    settings.set(IfcGeom::IteratorSettings::APPLY_DEFAULT_MATERIALS, true);
```

I would recommend this. If the IFC model does not have materials for some elements, ifcopenshell will then generate defaults. It's one less case to deal with below.

src/qtviewer/ParseIfcFile.cpp
```diff
@@ -21,22 +31,145 @@ void ParseIfcFile::Parse(const std::string& filePath)
...
-        outputMsg(triElem->type() + ": " + triElem->name());
+        MessageLogger::log(triElem->type() + ": " + triElem->name());
 
         const boost::shared_ptr<IfcGeom::Representation::Triangulation>& triElemGeom = triElem->geometry_pointer();
 
+        const std::vector<int>& elemFaces = triElemGeom->faces();
+        const std::vector<double>& elemVertices = triElemGeom->verts();
+        const std::vector<double>& elemNormals = triElemGeom->normals();
+
+        std::vector<Triangle> triangles = this->createTriangles(elemFaces, elemVertices, elemNormals);
```
aothms, line 50: This seems a bit inefficent to first compose everything into triangles and later on in the for loop on line 56 decompose everything again into arrays of vertices and normals. It also likely creates copies of the vertices, where previously a vertex could be referenced by multiple triangles.

src/qtviewer/ParseIfcFile.cpp
```diff
@@ -21,22 +31,145 @@ void ParseIfcFile::Parse(const std::string& filePath)
...
-        outputMsg("Error: Iterator failed to initialize! Aborting.");
+        MessageLogger::log("Error: Iterator failed to initialize! Aborting.");
         delete it;
-        return;
+        return false;
     }
 
     do {
         //const IfcGeom::BRepElement* bRepElem = it->get_native();
         const IfcGeom::TriangulationElement* triElem = static_cast<const IfcGeom::TriangulationElement*>(it->get());
```
aothms, line 41: I don't see you use. `triElem->transformation().matrix().data()`. This is the 4x3 matrix that would position the element in world coordinates. You can use this to populate an Matrixd and MatrixTransform.

What's more: if you look at triElem->geometry().id() you might see that the id is equal to the previous iteration. In that case you don't need to create a new osg::Geometry at all, but reuse the last one, but with a a different ElementTransform parent.

aothms (commented):
This is amazing progress! I'm very eager to try this out and connect it to the daily build system. Since these commits cover a bit more of the ifcopenshell API I have a bit more comments and questions.

src/qtviewer/IfcViewerWidget.cpp
dushyant-basson: Absolute agree. Removed rotation transformation and applied camera orientation upon loading the IFC.

src/qtviewer/ParseIfcFile.cpp
dushyant-basson: Thanks. Updated by mapping faces to material IDs, so now an element can display multiple materials applied to parts (like a layered composite wall).

src/qtviewer/ParseIfcFile.cpp
dushyant-basson: Thanks, this is useful and indeed helps avoid additional checks and manual application of default material.
Code updated.

src/qtviewer/ParseIfcFile.cpp
dushyant-basson: I agree, cycling through the triangles again could lead to an overhead in case of very complex geometries.
Implemented a revised flow.

src/qtviewer/ParseIfcFile.cpp
dushyant-basson: I tried it for different IFCs and found that the matrix data was the same for all the elements:
```
1.000000  0.000000  0.000000  0.000000  
1.000000  0.000000  0.000000  0.000000  
1.000000  0.000000  0.000000  0.000000 
```
Not sure what I'm missing here.

The sequence of triElem->geometry().id() values is like this:
```
355-world-coords
14488-world-coords
15016-world-coords
17018-openings-17111-world-coords
17095-world-coords
17456-world-coords
18443-world-coords
18676-openings-18771-18839-world-coords
18757-world-coords
18825-world-coords
19187-world-coords
19492-world-coords
20360-world-coords
20576-world-coords
...
```
So they don't seem to be consecutive. Please explain what you meant to suggest, based on these outputs. Thanks.

src/qtviewer/ParseIfcFile.cpp
```diff
@@ -1,42 +1,169 @@
...
-{
+bool ParseIfcFile::Parse(
+    const std::string& filePath,
+    std::vector<osg::ref_ptr<osg::Geometry>>& geometries
+) {
     IfcParse::IfcFile file(filePath);
 
     IfcGeom::IteratorSettings settings;
     settings.set(IfcGeom::IteratorSettings::USE_WORLD_COORDS, true);
     settings.set(IfcGeom::IteratorSettings::WELD_VERTICES, false);
```
aothms, line 31:
```suggestion
    settings.set(IfcGeom::IteratorSettings::WELD_VERTICES, false);
```

Sorry, I missed this bit. Setting USE_WORLD_COORDS to true indeed causes the matrix to be the identity matrix.

```
1 0 0 0
0 1 0 0
0 0 1 0
```

src/qtviewer/ParseIfcFile.cpp
aothms: Yep, sorry, see https://github.com/IfcOpenShell/IfcOpenShell/pull/3739/files#r1355661970

Also it helps to print the matrix as

```
1.000000  0.000000  0.000000  
0.000000  1.000000  0.000000  
0.000000  0.000000  1.000000
0.000000  0.000000  0.000000
```` 

3 columns instead of 4. When passing the data to OpenGL you probably need to add a fourth column

```
0.000000
0.000000
0.000000
1.000000
```

So that you get a 4x4 square matrix.

(This 3x4 matrix is what opencascade does, so we copied it).

Comment (dushyant-basson) [reply src/qtviewer/ParseIfcFile.cpp:43]: Hi,
Setting USE_WORLD_COORDS to false does give the coordinates in the transformation matrix data. The triElem->geometry().id()s are still not consecutive. They seem to be in ascending order. For some IFCs they are just numbers while for some they are a string like:
```
32-material-89
34-material-89
35-material-99
37-material-89
38-material-99
```
In the current method, new osg::Geometry is required to create new geometry nodes for OSG. Please suggest how you were proposing to do it in a single osg::Geometry node. And would that affect the selection of elements in the model if they are all a single node?

aothms: You can try with Duplex_A https://portal.nibs.org/files/wl/?id=O9inoVWDytV4P3UTS7iIdqhVKcTyVlcs

What may also help is to convert the ifc to .dae and import it in Blender (just plain Blender, no need for BlenderBIM here) to have a look. The collada (.dae) serializer operates just on the same data.

**Indeed as you observed quite a few elements with just a numeric representation name**

afbeelding

**Others with openings or materials applied get that postfixed**

afbeelding

**But again others, typically things like windows and doors have that representation shared. Note the [`4`] next to the "representation-5994" box. When you edit that representation in Blender you can see that you're editing 4 element representations at once.**

afbeelding

---

Regarding OSG I'm not an expert, but I think you'd create multiple osg::MatrixTransform that basically represent the Object (or rather it's placement) and then assign osg::Geometry to these. In case of reused shapes you'd assign the same osg::Geometry to multiple osg::MatrixTransform.

dushyant-basson: Thanks for the reference model and the images. Btw how did you convert the IFC to Collada? I used Archicad to convert it to .dae and then imported it in Blender. All the nodes are numbered serially for me in a consistent format:
Screenshot 2023-12-19 at 7 02 54 PM

dushyant-basson: Updated code. _(Using transformation-matrix data; Turned off world-coords; Re-using geometry of repeating IDs)_

aothms:
Brilliant. No further comments. I will try and get this included in the automatic builds. Thanks again for all of this.

PR #3761, opened 2023-09-15
clang-format: add styles and reformat ifcparse files

dirkolbrich:
This is a first test setting with clang-format and a complete reformat of the `/src/ifcparse` folder except generated files.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -4,9 +4,7 @@ ColumnLimit: 0 # we do not enforce column lenght, please format responsibly
 IndentWidth: 4
 ---
 Language: Cpp
-AllowShortBlocksOnASingleLine: Never
 AllowShortEnumsOnASingleLine: false
-AllowShortIfStatementsOnASingleLine: Never
 AlwaysBreakTemplateDeclarations: Yes
 BinPackArguments: false
 BinPackParameters: false
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,25 @@
+# We'll use defaults from the LLVM style, but with 4 columns indentation.
+BasedOnStyle: LLVM
+IndentWidth: 4
+---
+Language: Cpp
+AllowShortBlocksOnASingleLine: Always
+AllowShortEnumsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: OnlyFirstIf
+AlwaysBreakTemplateDeclarations: Yes
+BinPackArguments: false
+BinPackParameters: false
+BraceWrapping:
+    SplitEmptyFunction: false
+    SplitEmptyRecord: false
+    SplitEmptyNamespace: false
+IncludeBlocks: Regroup
+InsertNewlineAtEOF: true
+PackConstructorInitializers: CurrentLine
+PenaltyReturnTypeOnItsOwnLine: 120
+PointerAlignment: Left
+ReflowComments: false # keep licenze comment blocks
+SortIncludes: CaseInsensitive
+SpacesInLineCommentPrefix:
+    Maximum: 1
+TabWidth: 4
--- a/.clang-format
+++ b/.clang-format
@@ -1,11 +1,12 @@
-# We'll use defaults from the LLVM style, but with 4 columns indentation.
+# We'll use defaults from the LLVM style with modifications
 BasedOnStyle: LLVM
+ColumnLimit: 0 # we do not enforce column lenght, please format responsibly
 IndentWidth: 4
 ---
 Language: Cpp
-AllowShortBlocksOnASingleLine: Always
+AllowShortBlocksOnASingleLine: Never
 AllowShortEnumsOnASingleLine: false
-AllowShortIfStatementsOnASingleLine: OnlyFirstIf
+AllowShortIfStatementsOnASingleLine: Never
 AlwaysBreakTemplateDeclarations: Yes
 BinPackArguments: false
 BinPackParameters: false
@@ -14,9 +15,9 @@ BraceWrapping:
     SplitEmptyRecord: false
     SplitEmptyNamespace: false
 IncludeBlocks: Regroup
+InsertBraces: true
 InsertNewlineAtEOF: true
 PackConstructorInitializers: CurrentLine
-PenaltyReturnTypeOnItsOwnLine: 120
 PointerAlignment: Left
 ReflowComments: false # keep licenze comment blocks
 SortIncludes: CaseInsensitive
--- a/src/ifcparse/Argument.h
+++ b/src/ifcparse/Argument.h
@@ -20,59 +20,58 @@
 #ifndef ARGUMENT_H
 #define ARGUMENT_H
 
-#include <set>
-#include <string>
-#include <vector>
-#include <sstream>
-#include <algorithm>
-
-#include "ifc_parse_api.h"
 #include "ArgumentType.h"
+#include "ifc_parse_api.h"
 
-#include <boost/shared_ptr.hpp>
+#include <algorithm>
 #include <boost/dynamic_bitset.hpp>
 #include <boost/logic/tribool.hpp>
+#include <boost/shared_ptr.hpp>
+#include <set>
+#include <sstream>
+#include <string>
+#include <vector>
 
 class aggregate_of_instance;
 class aggregate_of_aggregate_of_instance;
 
 namespace IfcUtil {
-	class IfcBaseClass;
+class IfcBaseClass;
 
-    IFC_PARSE_API const char* ArgumentTypeToString(ArgumentType argument_type);
+IFC_PARSE_API const char* ArgumentTypeToString(ArgumentType argument_type);
 
-	/// Returns false when the string `s` contains character outside of {'0', '1'}
-	IFC_PARSE_API bool valid_binary_string(const std::string& s);
-}
+/// Returns false when the string `s` contains character outside of {'0', '1'}
+IFC_PARSE_API bool valid_binary_string(const std::string& s);
+} // namespace IfcUtil
 
 class IFC_PARSE_API Argument {
-public:
-	virtual operator int() const;
-	virtual operator bool() const;
-	virtual operator boost::logic::tribool() const;
-	virtual operator double() const;
-	virtual operator std::string() const;
-	virtual operator boost::dynamic_bitset<>() const;
-	virtual operator IfcUtil::IfcBaseClass*() const;
+  public:
+    virtual operator int() const;
+    virtual operator bool() const;
+    virtual operator boost::logic::tribool() const;
+    virtual operator double() const;
+    virtual operator std::string() const;
+    virtual operator boost::dynamic_bitset<>() const;
+    virtual operator IfcUtil::IfcBaseClass*() const;
 
-	virtual operator std::vector<int>() const;
-	virtual operator std::vector<double>() const;
-	virtual operator std::vector<std::string>() const;
-	virtual operator std::vector<boost::dynamic_bitset<> >() const;
-	virtual operator boost::shared_ptr<aggregate_of_instance>() const;
+    virtual operator std::vector<int>() const;
+    virtual operator std::vector<double>() const;
+    virtual operator std::vector<std::string>() const;
+    virtual operator std::vector<boost::dynamic_bitset<>>() const;
+    virtual operator boost::shared_ptr<aggregate_of_instance>() const;
 
-	virtual operator std::vector< std::vector<int> >() const;
-	virtual operator std::vector< std::vector<double> >() const;
-	virtual operator boost::shared_ptr<aggregate_of_aggregate_of_instance>() const;
+    virtual operator std::vector<std::vector<int>>() const;
+    virtual operator std::vector<std::vector<double>>() const;
+    virtual operator boost::shared_ptr<aggregate_of_aggregate_of_instance>() const;
 
-	virtual bool isNull() const = 0;
-	virtual unsigned int size() const = 0;
+    virtual bool isNull() const = 0;
+    virtual unsigned int size() const = 0;
 
-	virtual IfcUtil::ArgumentType type() const = 0;
-	virtual Argument* operator [] (unsigned int i) const = 0;
-	virtual std::string toString(bool upper=false) const = 0;
-	
-	virtual ~Argument() {};
+    virtual IfcUtil::ArgumentType type() const = 0;
+    virtual Argument* operator[](unsigned int i) const = 0;
+    virtual std::string toString(bool upper = false) const = 0;
+
+    virtual ~Argument(){};
 };
 
 #endif
--- a/src/ifcparse/ArgumentType.h
+++ b/src/ifcparse/ArgumentType.h
@@ -1,21 +1,21 @@
 /********************************************************************************
-*                                                                              *
-* This file is part of IfcOpenShell.                                           *
-*                                                                              *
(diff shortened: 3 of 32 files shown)


dirkolbrich:
Please look through the changes in formatting and comment, if this would work for the overall project and C++ code.

src/ifcparse/Argument.h
```diff
@@ -20,59 +20,59 @@
...
 
-	virtual operator std::vector<int>() const;
-	virtual operator std::vector<double>() const;
-	virtual operator std::vector<std::string>() const;
-	virtual operator std::vector<boost::dynamic_bitset<> >() const;
-	virtual operator boost::shared_ptr<aggregate_of_instance>() const;
+    virtual operator std::vector<std::vector<int>>() const;
+    virtual operator std::vector<std::vector<double>>() const;
+    virtual
+    operator boost::shared_ptr<aggregate_of_aggregate_of_instance>() const;
```
dirkolbrich, line 66: line break prop. due to statement line length

src/ifcparse/ArgumentType.h
```diff
@@ -17,44 +17,43 @@
 *                                                                              *
 ********************************************************************************/
 
-#include "../ifcparse/IfcSchema.h"
-
 #include "ifc_parse_api.h"
+#include "IfcSchema.h"
```
dirkolbrich, line 21: These need to go after the include guard.

src/ifcparse/IfcSIPrefix.cpp
```diff
@@ -20,154 +20,194 @@
...
+        return 1.e-9;
+    else if (v == "PICO")
+        return 1.e-12;
+    else if (v == "FEMTO")
+        return 1.e-15;
+    else if (v == "ATTO")
+        return 1.e-18;
+    else
+        return 1.;
 }
```
dirkolbrich, line 91: to prevent a break after every `else if ()`this block needs to be commented with `// clang-format off`. Probably needs to tweak the setting `AllowShortIfStatementsOnASingleLine`.

src/ifcparse/IfcSpfHeader.cpp
```diff
@@ -17,161 +17,167 @@
...
+#include "IfcFile.h"
+
+static const char* const ISO_10303_21 = "ISO-10303-21";
+static const char* const HEADER = "HEADER";
+static const char* const FILE_DESCRIPTION = "FILE_DESCRIPTION";
+static const char* const FILE_NAME = "FILE_NAME";
+static const char* const FILE_SCHEMA = "FILE_SCHEMA";
+static const char* const ENDSEC = "ENDSEC";
+static const char* const DATA = "DATA";
 // The following header entities are not normally encountered in IFC files and are not parsed.
```
dirkolbrich, line 31: needs to tweak `AlignConsecutiveAssignments` and its options

src/ifcparse/IfcWrite.h
```diff
@@ -28,148 +28,149 @@
...
+
+        // AGGREGATES OF AGGREGATES:
+        empty_aggregate_of_aggregate_t,
+        // An aggregate of an aggregate of ints. E.g. ((1, 2), (3))
+        std::vector<std::vector<int>>,
+        // An aggregate of an aggregate of floats. E.g. ((1., 2.3), (4.))
+        std::vector<std::vector<double>>,
+        // An aggregate of an aggregate of entities. E.g. ((#1, #2), (#3))
+        aggregate_of_aggregate_of_instance::ptr>
+        container;
```
dirkolbrich, line 118: here the closing right angle of `boost::Variant<>` should break to the next line to get a better readability. have not found the right setting yet.

src/ifcparse/Argument.h
aothms: `virtual operator boost::shared_ptr<aggregate_of_aggregate_of_instance>() const;` is +- 80 chars it seems. Can we increase that?

src/ifcparse/ArgumentType.h
aothms: agreed. not sure what happened here, maybe some merge artefact, or just a weird oversight.

src/ifcparse/IfcSIPrefix.cpp
aothms: For me it's not an issue. This kind of formatting/alignment for readability tends to be a pain to maintain anyway. For me it's not necessary to preserve it.

I wonder though. Shouldn't we require braces around every if-statement/for-loop/etc. body? It's so easy to introduce errors (granted indentation would help you, but still):

```C++
if (something)
   do_something();
   do_something_else();
```

src/ifcparse/IfcSpfHeader.cpp
aothms: Again, not a big issue for me. You already see in the previous state it got mangled between tab width and spaces. Besides https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignconsecutiveassignments does a lot of things I wouldn't consider to improve readability. Only in these cases where you're basically defining aliases it makes some sense to me, not in the general case of assignments. But I don't think clang-format will be able to meaningfully differentiate between those cases.

src/ifcparse/IfcWrite.h
aothms: This is really an exception though. I wouldn't spend too much time fighting with this.

.clang-format
```diff
@@ -0,0 +1,25 @@
+# We'll use defaults from the LLVM style, but with 4 columns indentation.
+BasedOnStyle: LLVM
+IndentWidth: 4
+---
+Language: Cpp
+AllowShortBlocksOnASingleLine: Always
+AllowShortEnumsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: OnlyFirstIf
```
aothms, line 8: I would prefer never. Just for consistency.

```suggestion
AllowShortBlocksOnASingleLine: Never
AllowShortEnumsOnASingleLine: false
AllowShortIfStatementsOnASingleLine: Never
```

src/ifcparse/IfcLogger.h
```diff
@@ -20,93 +20,124 @@
...
-	static void Error(const std::exception& exception, const IfcUtil::IfcBaseInterface* instance = 0) { Message(LOG_ERROR, exception, instance); }
-
-	static void Status(const std::string& message, bool new_line=true);
-
-	static void ProgressBar(int progress);
-	static std::string GetLog();
-	static void PrintPerformanceStats();
-	static void PrintPerformanceStatsOnElement(bool b) { print_perf_stats_on_element = b; }
+  public:
+    typedef enum {
```
aothms, line 38: I guess this doesn't need to be a typedef. Is there a rule for that. Not sure if we use it often.

aothms:
This is excellent. Thanks for highlighting some things. I'm really happy that it doesn't indent after namespaces. That's such a relief.

> removed unnecessary ../ifcparse/ path in include statements, if the included file is in the same folder

This was on purpose. To make it discoverable which module you're importing from. I must admit it's not incredibly useful in ifcopenshell. But I've seen codebases where it really helps. 

One thing that I naively hoped, but which doesn't seem possible to to reformat case and postfix. In general I prefer this convention:

```C++
class IFC_PARSE_API base_class : public virtual base_interface {
  private:
    uint32_t identity_;
    static std::atomic_uint32_t counter_;
```

So lowercase with underscore. Private/protected members get a postfix additional underscore.

It seems it can be checked, but not formatted:

https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html#readability-identifier-naming

dirkolbrich:
Yeah, this would be `ColumnLimit` with a sensible default of 120?
phonephone

aothms:
120 sounds good to me

src/ifcparse/Argument.h
dirkolbrich: LLVM and Google style use a `ColumnLimit` of 80. We could set it to 120, but there will always be a line which is just 1-2 characters too long. Or we could NOT introduce a limit for the beginning (setting it to 0) and bet on some form of self-control?

src/ifcparse/IfcSIPrefix.cpp
dirkolbrich: There is an option to `InsertBraces`, but clang advises on extra caution to set this https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces. I think the first format with this option could lead to falsely introduced braces, the following format runs should be fine.

src/ifcparse/IfcSpfHeader.cpp
dirkolbrich: Ok, we can leave this out for the start.

dirkolbrich:
> I'm really happy that it doesn't indent after namespaces. That's such a relief.

If you really insist that much, we could turn this on via [NamespaceIndentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#namespaceindentation) 😜

> In general I prefer this convention:
> 
> ```c++
> class IFC_PARSE_API base_class : public virtual base_interface {
>   private:
>     uint32_t identity_;
>     static std::atomic_uint32_t counter_;
> ```
> 
> Private/protected members get a postfix additional underscore.

No, there is no option to format that in. I will check if we can setup clang-tidy to give a warning during linting and recommend the preferred style. Usually the type of naming convention should be documented in a style guide like [Google does it](https://google.github.io/styleguide/cppguide.html#Naming) and enforced during code review.

Lower case with underscore for function or variable names is fine with me (despite I personally prefer camelCase). But generally here I disagree with the use of a trailing underscore for private members variable names. This tends to get visually lost, especially if the declaration or use is within a larger codeblock. Better to explicitly name a private member with a prefix like `m_private_member`.

For type names like Classes I personally prefer [PascalCase](https://en.wiktionary.org/wiki/Pascal_case) like [Google does it](https://google.github.io/styleguide/cppguide.html#Type_Names) as this differentiates a type from a function visually more clearly.

src/ifcparse/Argument.h
aothms: I would choose either 120 or 0. You decide.

src/ifcparse/IfcSIPrefix.cpp
aothms: I would say let's give it a try. I don't know which kind of situations could give issues. I'd imagine they would result into compilation errors due to changes in scope, but yeah if you're unlucky maybe it results into a subtle behaviour changes. I can only think macro-based loops such as BOOST_FOREACH that would depend on more context than clang-format has, but that would then be a case of too little braces. I will review carefully.

aothms:
> Better to explicitly name a private member with a prefix like m_private_member.

I really prefer a postfix due to typing speed with code completion. Any convention you're aware of that'd be a good compromise?

The lower_case vs camelCase thing I can adapt to in either way. I personally prefer lower case due to similarity with boost / stl / most of python, but my intuition says actually camelCase is preferred among most. I do want more consistency between C++ and Python classes (less renames in swig) so this might turn out to be backwards incompatible on the python side where we have the most use.

dirkolbrich:
Ok, will leave the naming conventions for variables and type definitions as is.

dirkolbrich:
Next format test with minor changes to the clang-format style. The most notable not endorsing column length. the formatting looks good to me.

dirkolbrich:
> >removed unnecessary ../ifcparse/ path in include statements, if the included file is in the same folder

> This was on purpose. To make it discoverable which module you're importing from. I must admit it's not incredibly useful in ifcopenshell. But I've seen codebases where it really helps.

This should not be a readability problem any more, as the formatting now groups include blocks with the same folder includes as the first block, see the file `IfcParse.h` for example:

```cpp
#ifndef IFCPARSE_H
#define IFCPARSE_H

#if defined(IFCOPENSHELL_BRANCH) && defined(IFCOPENSHELL_COMMIT)
#define IFCOPENSHELL_VERSION STRINGIFY(IFCOPENSHELL_BRANCH) "-" STRINGIFY(IFCOPENSHELL_COMMIT)
#else
#define IFCOPENSHELL_VERSION "0.8.0"
#endif

#include "aggregate_of_instance.h"
#include "Argument.h"
#include "ifc_parse_api.h"
#include "IfcBaseClass.h"
#include "IfcCharacterDecoder.h"
#include "IfcLogger.h"
#include "IfcSpfStream.h"
#include "macros.h"

#include <boost/dynamic_bitset.hpp>
#include <boost/shared_ptr.hpp>
#include <cstring>
#include <fstream>
#include <iostream>
#include <map>
#include <sstream>
#include <string>
#include <vector>

namespace IfcParse {
```

aothms:
It looks good to me as well. Very nice :) Shall we merge?

dirkolbrich:
Let it sit for a minute. Maybe I see some things during my Sunday coffee and cake session, like duplicate clang-format setting, which are already in the derived base style.

dirkolbrich:
I fixed some alignments of license blocks and shortened some exec line length. I think I'm good to merge.

dirkolbrich:
@aothms I think we can merge, if you agree.

dirkolbrich:
@aothms Can we move forward with this PR or are there any modifications needed?

dirkolbrich:
@aothms Here is my weekly reminder for this PR. Can we move forward with the integration of clang-format. Or are there any changes needed on this? I'm happy contribute.

Krande:
@aothms Can this be merged or will it interfere with https://github.com/IfcOpenShell/IfcOpenShell/pull/3826 or https://github.com/IfcOpenShell/IfcOpenShell/pull/3827 (I haven't checked it myself, but can do so if you want)? 

@dirkolbrich I can merge this if @aothms gives us the thumbs up :)

dirkolbrich:
There shouldn't be any conflicts, as PR #3826 and #3827 only touch `src/ifcgeom` code. This PR modifies `src/ifcparse` exclusively, but only in code style and formatting, no functional modifications.

aothms:
Sorry, for the delay. I've been flooded with other things. Thanks again!

aothms:
FYI I get this with my local MSVC 2017 build:

afbeelding

Which seems to be because of an outdated clang-format v6 that's shipped with msvc 2017

https://developercommunity.visualstudio.com/t/clang-format-incorrect-options/683472

The following changes seem to work with msvc 2017. But I must say I'm not impressed in general with the clang-format integration in msvc2017. This is not an issue though. For v0.8 I'll move to msvc 2019/2022 because anyway also Qt6 requires it, which we now depend on for the C++ ifc viewer.

I'm just leaving this here in case somebody stumbles upon this but for now I don't see this as an issue. It would be good to pin a clang-format version in the future though.

```diff
diff --git a/.clang-format b/.clang-format
index 40a0f424d..e0b190914 100644
--- a/.clang-format
+++ b/.clang-format
@@ -5,7 +5,7 @@ IndentWidth: 4
 ---
 Language: Cpp
 AllowShortEnumsOnASingleLine: false
-AlwaysBreakTemplateDeclarations: Yes
+AlwaysBreakTemplateDeclarations: true
 BinPackArguments: false
 BinPackParameters: false
 BraceWrapping:
@@ -14,11 +14,10 @@ BraceWrapping:
     SplitEmptyNamespace: false
 IncludeBlocks: Regroup
 InsertBraces: true
-InsertNewlineAtEOF: true
 PackConstructorInitializers: CurrentLine
 PointerAlignment: Left
 ReflowComments: false # keep licenze comment blocks
-SortIncludes: CaseInsensitive
+SortIncludes: true
 SpacesInLineCommentPrefix:
     Maximum: 1
 TabWidth: 4

```

PR #3777, opened 2023-09-20
Fix getting unit for IfcPropertyEnumeratedValue

andy-5:
This PR improves getting the unit of an `IfcPropertyEnumeratedValue`.

It first checks if the `EnumerationReference` has an unit assigned (this was ignored before), and only then falls back to using the type of the first enumerated value to get a default (project defined) unit.

Also, the use of the `measure_class` variable is refactored a bit, to avoid `UnboundLocalError` exceptions if none of the `if`/`elif` cases matches.


Moult:
Thanks so much!

PR #3805, opened 2023-09-26
Fixed most issues with IDS auditing

CBenghi:
Hello @Moult and @aothms,

I've made a few changes to the code to reduce the audit failures on the IDS test cases.

There has also been work on the side of the audit tool, that I still have to publish.

With this PR and those changes we are down to 7 issues in the auditing of passing scenarios. Once I'll have sorted those, we will have to determine if we want a special identification for fail cases that do not pass the auditing, if we want to have full QA on the IDS repository.

Thanks,
Claudio

### DIFF
--- a/src/ifctester/ifctester/facet.py
+++ b/src/ifctester/ifctester/facet.py
@@ -286,7 +286,7 @@ class Attribute(Facet):
 
 
 class Classification(Facet):
-    def __init__(self, value=None, system=None, uri=None, minOccurs=None, maxOccurs=None, instructions=None):
+    def __init__(self, value=None, system=None, uri=None, minOccurs=None, maxOccurs="unbounded", instructions=None):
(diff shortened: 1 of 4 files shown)


Moult:
Thanks so much! Would you like to have a meeting together to work through latest IDS changes?

CBenghi:
> Thanks so much! Would you like to have a meeting together to work through latest IDS changes?

Hi @Moult,
if might be useful, but my calendar is a bit packed these days.
Ok if we arrange a chat after the general call next week?

From the 4th of October to the 10th it's gonna be ok for me to work on Aussie time.

:)

Moult:
Sure sounds good give me a buzz. I'm happy to do the work, I just need someone who's been keeping up to date with all the changes to run the past me since i've been out of the loop for a while :(

PR #3823, opened 2023-09-28
Work in progress for alignments

RickBrice:
PR for [https://github.com/IfcOpenShell/IfcOpenShell/issues/3098]().

Primarily working in IfcCurveSegment.cpp to map horizontal alignment geometry.
Stubbed out framework for dealing with spirals.

Very uncertain if I'm dealing with placements correctly. IfcCurveSegment and some types of the ParentCurve (IfcCircle and IfcClothoid) have placements. Somehow these two placements need to work together to get the geometry in the right place.


aothms:
This is great. Thanks again!

> Very uncertain if I'm dealing with placements correctly. IfcCurveSegment and some types of the ParentCurve (IfcCircle and IfcClothoid) have placements. Somehow these two placements need to work together to get the geometry in the right place.

I'm quite sure it's correct. Conceptually think of it as a procedural definition, from the depth of the tree to the root representation item upwards. The various transformations should also be applied in that order.

PR #3826, opened 2023-09-29
Some minor tweaks and implements suggestions

RickBrice:
Sorry - not sure if I'm doing this PR thing correctly. After the PR merge, my fork is out of sync with upstream. The sync feature said it was going to delete my commits, but I didn't want to lose them so I just created this RP.

### DIFF
--- a/src/ifcgeom/mapping/IfcCurveSegment.cpp
+++ b/src/ifcgeom/mapping/IfcCurveSegment.cpp
@@ -40,11 +40,11 @@ namespace
 	double integrate(double a, double b, unsigned n, std::function<double(double)> fn)
 	{
 		double area = 0;
-		double h = (b - a) / (n + 1);
-		for (auto i = 0; i < n; i++)
+		double h = (b - a) / n;
+		for (auto i = 1; i <= n; i++)
 		{
-			auto x1 = a + h * i;
-			auto x2 = a + h * (i + 1);
+			auto x1 = a + h * (i - 1);
+			auto x2 = a + h * i;
 			auto f1 = fn(x1);
 			auto f2 = fn(x2);
 			area += h * (f1 + f2) / 2.0;
--- a/src/ifcgeom/mapping/IfcCurveSegment.cpp
+++ b/src/ifcgeom/mapping/IfcCurveSegment.cpp
@@ -33,6 +33,8 @@ typedef boost::mpl::vector<
 #ifdef SCHEMA_HAS_IfcClothoid
 	, IfcSchema::IfcClothoid
 #endif
+	, IfcSchema::IfcPolyline
+	, IfcSchema::IfcCircle
 > curve_seg_types;
 
 class curve_segment_evaluator {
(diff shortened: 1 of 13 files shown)


RickBrice:
There seems to be something wrong with the clothoid geometry when computing points on the curve using numeric integration. I did a comparison of the points using the series approximation and the numeric integration and the points where significantly different, especially away from (0,0). I don't have a solution yet.

src/ifcgeom/mapping/IfcCurveSegment.cpp
```diff
@@ -249,48 +206,29 @@ class curve_segment_evaluator {
...
-				auto xl = R * cos(angle);
-				auto yl = R * sin(angle);
+				auto x = R * cos(angle);
+				auto y = R * sin(angle);
 
 				// transform point into circle's coodinate system
-				auto x = xl * cos(theta) - yl * sin(theta) + Cx;
-				auto y = xl * sin(theta) + yl * cos(theta) + Cy;
-				return Eigen::Vector3d(x, y, 0.0);
+				auto result = transformation_matrix * Eigen::Vector4d(x, y, 0.0, 1.0);
```
RickBrice, line 228: @aothms is there a better way to apply the transformation matrix? When I multiply matrix4 with vectorxd, the results are not correct

src/ifcgeom/mapping/IfcCurveSegment.cpp
```diff
@@ -376,11 +316,59 @@ class curve_segment_evaluator {
...
+				return vec;
+				};
+
+		}
+	}
+
+	void operator()(IfcSchema::IfcPolynomialCurve* p) {
+
+		if (segment_type_ == ST_HORIZONTAL) {
+			auto coeffX = p->CoefficientsX();
```
RickBrice, line 345: @aothms horizontal not finished yet since polynomial is not typically. For vertical, profile typically use a parabolic curve. I've been using IfcPolynomialCurve assumption Cy1*u^2 + Cy2*u + Cy3. However, the ifc specification is silent regarding how the coefficients should be applied. It could be Cy1 + Cy2*u + Cy3*u^2. I'm also assuming the polynomial curve is in a plane (2D curve). 

Any thoughts or suggestions?

src/ifcgeom/mapping/IfcCurveSegment.cpp
aothms: `Eigen::Vector3d(x, y, 0.0).homogeneous();` can be used instead of Vector4D. For example when you have a contiguous matrix of XYZ values, homogenous can be used to colwise add the 1 symbollically without making a copy (using Eigen::map)  of the underlying data. I can't find anything on VectorXd in the docs, other than that there would be a runtime check instead of a compile time check on size. I have no problem with temporarily casting to Vector4d. I assume the compiler will help us out here with optimizations.

src/ifcgeom/mapping/IfcCurveSegment.cpp
aothms: Maybe @peterrdf do you have time for this quick question?

civilx64: Based on [this thread](https://forums.buildingsmart.org/t/ifcpolynomialcurve-clarification/4716) in bSI forums it appears that it should be the latter interpretation of coefficient order.  Or at least that is what GeometryGym appears to be using.  I'd be happy to test the implementation against the data that exists for synthetic test cases from the IfcRail project.  I would just need a compiled version of the python extension as my C++ skills are generally in the range of `Hello, world!`.  Or perhaps I could implement github actions on my fork for compiling? @aothms any recommendations here?  I have VS 2022 Community for Win but much more comfortable in WSL2 Ubuntu.

peterrdf: On Tuesday (10th of October) we have a meeting with somebody also getting different values for Clothoid. Let me check this first, there could be very well an issue in the examples also, and then come back here with a clean explanation for calculating the points of a clothoid based on some of the theoretical examples as we have available on GitHub (https://github.com/bSI-RailwayRoom/IFC-Rail-Unit-Test-Reference-Code/tree/master/alignment_testset).

Also let me then try to make a basic example based on IfcClothoid itself (i.e. without trimming).

RickBrice: @civilx64 here are some instructions I put together for building with vs2022. https://github.com/RickBrice/PGSuperIfcExtensions/blob/BIM_for_Bridges_TPF/devdocs/Dependencies.md

aothms:
I'm going to merge this in, so that we don't create conflicts. And so that I can try and see what is needed to connect to this from python.

RickBrice:
@aothms Thanks - I've been away from this for a couple weeks. I'll be digging back in this weekend. Trying to get IfcGradientCurve working for horizontal + vertical alignments.

aothms:
@RickBrice @civilx64 

I've made some small changes throughout to be able to get evaluated points in the python wrapper https://github.com/IfcOpenShell/IfcOpenShell/commit/661c8bc50f0a7226531a954e896bbe2cd971690c

```
>>> import ifcopenshell, ifcopenshell.geom
>>> f = ifcopenshell.open('linear-placement-of-signal.ifc')
>>> ifcopenshell.geom.create_shape(ifcopenshell.geom.settings(), f.by_type('IfcCurveSegment')[0]).verts
(452270.188250964, 4539403.94736217, 0.0, 452270.57072091993, 4539404.2694129925, 0.0, 452270.9531908758 ...
>>> ifcopenshell.geom.create_shape(ifcopenshell.geom.settings(), f.by_type('IfcGradientCurve')[0]).verts
```

Model is [0], but you need to change the schema from IFC4X3_ADD2 to IFC4X3_ADD1 and all IFCLENGTHMEASURE to IFCNONNEGATIVELENGTHMEASURE.

[0] https://github.com/buildingSMART/IFC4.3.x-sample-models/blob/main/models/alignment-geometries-and-linear-positioning/linear-placement-of-signal/linear-placement-of-signal.ifc

I'll start a v0.8 build so that you can use it without compilation

civilx64:
Thanks! I was finally able to get the solution to compile per @RickBrice 's instructions above.  I'm running the `win` install batch file now and all seems to be working as expected.  Hopefully I'm close to being able to fish for myself on compiling but the v0.8.0 build artifacts will be very welcome in the meantime.

aothms:
The windows build succeeded, but looks like we have a couple of compilation issues on gcc/clang that I will address. https://github.com/IfcOpenBot/IfcOpenShell/commit/661c8bc50f0a7226531a954e896bbe2cd971690c#comments

Issue #2878, opened 2023-03-17
Principles about ifcopenshell.api and signature design


cvillagrasa:
See discussion in https://github.com/IfcOpenShell/IfcOpenShell/pull/2869

@myoualid @vulevukusej @Gorgious56 and everyone interested, what do you think?

Moult:
Copy pasting discussion here for convenience:

I'd like to set a series of principles which will guide signature design. How does this sound?

 - Signatures should not be excessively long (i.e. require lots of kwargs).
 - Signatures should require enough mandatory attributes to prevent invalid IFC data from being created. These should be individual kwargs to help distinguish them.
 - Where there is a long optional list of arguments, it is acceptable to provide it as a dictionary which may be unpacked for convenience. Because they will be auto unpacked, the dict keys should match the capitalisation of IFC attribute names.
 - To be consistent, any kwargs that affect a particular IFC attribute should also match the capitalisation of IFC attribute names.
 - To distinguish kwargs that affect an IFC attribute vs a regular kwarg that affects other behaviour, other kwargs should use pythonic_underscore_names.
 - API functions should "read like English" in code where possible.
 - Where a length value may be specified should we always require project units? Or allow SI units? Or both with an "is_si" toggle?
 - Attributes which are technically optional but highly useful in practice should be promoted to non-dictionary kwargs (Name, Description, etc)
 - We should allow combined add / edit into a single API call for convenience. It's very annoying to add then have to edit immediately after.
 - APIs that deal with relationships (rels with assign / unassign) should allow for batching. So you should always be able to specify a list of objects if the relationship allows a list instead of separate calls for every single object. This will have a huge impact on optimisation for scripts and large cleanups.

What do you think? Naturally a lot of the current APIs fail to meet these requirements, but that's all part of the process and if we agree on a strong set of principles we can incrementally improve every function.

**response by @cvillagrasa **

Interesting discussion! Maybe it's worth to open a dedicated issue and try to have as much feedback as possible. This PR could be merged with just the one-liner bug fix commit for now.

[...]

PR #3830, opened 2023-09-30
Patch API calls to ifcopenshell.api

cvillagrasa:
This PR is related to https://github.com/IfcOpenShell/IfcOpenShell/issues/2693. It adds static type decorations without changing the underlying architecture of the Usecases and the centralized `run`. Or, more specifically, it grabs the signatures from the Usecases, so to take full advantage of the benefits of static typing in smart IDEs or Jupyter Notebooks, they'd also need to update its signature, as proposed in https://github.com/IfcOpenShell/IfcOpenShell/pull/3774 for the `geometry.add_wall_representation` case.

The proposal consists of the simple addition of `patch_api` method to IfcOpenShell files. For now, nothing happens if one doesn't explicitly call `file.patch_api()`, so it is safe to merge in case this strategy sounds good. If/when it is fully agreed, it can then be included to `file.__init__` and patch all new files by default at runtime. I guess it could also be added to the cmake in order to patch the file class at install time, by making it a classmethod, but I can't think of a straightforward way to reach the file instance in that case.

It is worth noting that there aren't as much API calls so as to add a very significant overhead. There's a glob over there to find the Usecase files, but a call to `patch_api` takes 100 ms on my system. I'd say it's acceptable.

With this design, one calls the API following the dot notation from a file:
patch_api

I find it very user friendly, even better than what I was proposing before of `ios.root.create_entity(file, "IfcWall")`. It more or less follows the notation of the existing `file.create_entity`, and the dot notation, grouping by API module, makes the namespace not super polluted, at least in my view.

What do you think of an approach like this, @aothms @Moult and everyone interested?

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/__init__.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/__init__.py
@@ -19,6 +19,14 @@
 """High level user-oriented IFC authoring capabilities"""
 
 import json
+import os
+from pathlib import Path
+import importlib
+import inspect
+from types import ModuleType
+from typing import Any, Optional, DefaultDict
+from dataclasses import dataclass, field
+from collections import defaultdict
 import numpy
 import importlib
 import ifcopenshell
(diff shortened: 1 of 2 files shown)


cvillagrasa:
Just tried it on a Jupyter environment and it works great! ;) maybe I should polish the docstring.

patch_api_01

patch_api_02

BTW, for this PR I was focused on Jupyter, now re-reading my message above... including also IDEs wasn't correct. So this PR only brings the benefits of static typing to Jupyter, with a live session, unfortunately not to IDEs.

aothms:
I'm fully inline with the goals, but I'm not sure if putting this on the file instance makes most sense. What if we later have API calls with multiple files (which I guess is already somewhat possible with templates and all).

Can't you monkey patch the ifcopenshell.api module itself?

cvillagrasa:
> Can't you monkey patch the ifcopenshell.api module itself?

Sure. Let me explain the reasoning behind the choice I made for this PR, but definitely your vote weights a lot. We can do `import ifcopenshell.api as ios` and then `ios.root.create_entity(file, "IfcWall")`. Do you like ios as an alias?

The reasoning behind: It was to follow the user-friendliness of numpy, pandas or any of those Data Science libs, which provide a ton of methods on instances. As for API calls on multiple file instances (are there any, yet? what's possible with templates?), they look fundamentally different to me, so no question that they should be placed under `ifcopenshell.api`. The same way that `np.vstack((arr1, arr2))` needs to come from numpy and `my_array.flatten()` can be called as a method of the ndarray instance. BTW, what happens with `file.create_entity`? should it be homogeneized with the rest of the API? 

I kind of like the semantics of "On this file, I'm performing this action, with these arguments" rather than "I'm performing this API action with these arguments, being the file one of them". But of course this is a very minor nuance, and the most important bit is to have hinting aids.

aothms:
> It was to follow the user-friendliness of numpy, pandas or any of those Data Science libs, which provide a ton of methods on instances

I always use the free functions actually..

```python
>>> import numpy as np
>>> arr = np.array([1,2,3])
>>> arr.sum()
6
>>> np.sum(arr)
6
```

> Do you like ios as an alias?

Not a fan, but I'm not a dictator either. I would probably write:

```python
from ifcopenshell import api as ifc
ifc.create_entity()
```

let's see what the other say

cvillagrasa:
> I always use the free functions actually..

Yeah, I guess it's a matter of preference. I moved it to ifcopenshell.api, with the name ifc. Note that patching it to api directly would collide with the actual Usecase modules. So now, the current state of this PR allows for the following:

```python
from ifcopenshell.api import ifc

file = ifc.project.create_file()
wall = ifc.root.create_entity(file, ifc_class="IfcWall")
```

My [first thought for the alias](https://github.com/IfcOpenShell/IfcOpenShell/issues/2693#issuecomment-1475287797) also was ifc, then I thought it might get confused with the file (since  some people call the file instances ifc), but the truth is even the file shouldn't be called file. Dion often calls them model, which in turn can collide with the model context... and it never ends. Choosing ios could help avoid some of these collisions, but maybe it's true the naming can feel foreign. So happy with ifc!

> but I'm not a dictator either

I'd totally name you BDFL! 😉

cvillagrasa:
Following up with https://github.com/IfcOpenShell/IfcOpenShell/pull/3774 and this PR, let me guys restate a bit the spirit of this set of PRs regarding ifcopenshell-python.

If in IFC, we have different releases / addendums / corrigendums, I believe that it should be possible to initialize a model of any schema definition, and yet have type hinting benefits according to that specific definition. Or what's the same, I find it very annoying as a user not to have those benefits, that I would expect in every modern Python library.

The classical example I use to illustrate this is `material.add_layer`. If we could pass it every IFC layer related arg for it to be correctly constituted, we could for instance use the kwarg `category="Insulation"`, so long as the version is >= IFC4. So category should appear in the signature of an IFC4 model, but not if it was IFC2x3. And here's the thing, not having this distinction (as it is now) makes it easy to develop, but the user has to then monkey-patch schema attributes with the corresponding bSI docs opened in the backgroung. But... having this distinction without any Python trick requires to have an ifcopenshell-python distribution per schema definition.

To be clear beyond this PR, I believe the way to go is precisely to have one ifcopenshell-python distribution per schema definition. Not by maintaing all of them, for sure, but by autogenerating schema-dependent modules. Then, there would be no need to ship them all together, and we could manage optional dependencies and do perhaps `pip install ifcopenshell[IFC4]` or `pip install ifcopenshell[IFC4X3]` or explicitly `pip install ifcopenshell[all]`, with maybe `pip install ifcopenshell` getting IFC4 and IFC4x3. Am I dreaming?

I wouldn't break the tight relationship to the schema on an API that makes things such as `add_layer`, referring to an IfcMaterialLayer. I'd leave that to higher level stuff built on top with something like `add_fancy_floor`.

And to come back to this PR (and #3774 and #3851), I've found these techniques to be a pragmatical way of having type hinting benefits in the short term, specially in notebook environments. So my idea was to experiment with a friendly API first, to then manufacture some CI/CD to produce the multischema API.

And that's my take! ;)

Moult:
> we could for instance use the kwarg category="Insulation", so long as the version is >= IFC4. So category should appear in the signature of an IFC4 model, but not if it was IFC2x3.

What if the category is always there as an arg, but silently ignored if it's IFC2X3?

I see two parts of this PR. The first is the difference between passing API calls as a string to run() which is totally undiscoverable with no autocompletion, no docstring fetching, and no signature previews in IDEs and interpreters. For this, 100% agree needs to be fixed ASAP.

The second aspect is the actual signature design. How much type hinting is done, how strict should they be at runtime, how much validation needs to be done, how they handle schema differences #3851 (manual definition? autogenerated at runtime? autogenerated at buildtime?). Note that even though BBIM has no support for Python <3.10, IOS is still distributed down to Python 3.6 so we need to make a call. I believe type hinting is 100% a good idea. Runtime checking and schema differences I'm uncertain how to handle. I'll answer more on #3851 and #2878. (wow, this is turning into a maze of issues and PRs, really gotta get some closure on this!)

Moult:
I'm closing this PR because after a lot of discussion and consideration, I don't think putting the API on the file object is a good idea. Reasons:

 - Currently, API functions expects a single file. In the future, it may not.
 - The API is big, and will grow in the future. I don't want to enforce it to be loaded. It should be possible to work with the core ifcopenshell at a low level with no api/utils.
 - Keeping the file lean also makes it easier to support wasm.
 - Patching it at runtime doesn't work for static analysis and in most IDEs. Generating code to support it on the file means work to support two ways to do the same thing with different syntactic sugar. Refactoring code statically to only work on the file enforces large dependencies the moment someone creates a file object.

PR #3850, opened 2023-10-05
Fixed model reference

### LINKED ISSUE #3837: Assign product usecase requires file reference.
Running the python example for assign product to task results in:

`Usecase.__init__() missing 1 required positional argument: 'file'`

I insert file object as the second argument and it works.
Is this the intended method or should I be instanciating something on my task/entitiy_instance?

https://github.com/IfcOpenShell/IfcOpenShell/blob/9461bf21291a11e82ae4713eb68a74f119e48437/src/ifcopenshell-python/ifcopenshell/api/sequence/assign_product.py#L64C1-L65C1

It should be:
```
# Let's construct that wall!
ifcopenshell.api.run("sequence.assign_product", model, relating_product=wall, related_object=task)
```

taylorhorwood:
fixes #3837


aothms:
Thanks!

PR #3853, opened 2023-10-06
Cleaning up early fetch cancellation

hilko-o:


aothms:
Thank you! Sorry for the trouble.

PR #3864, opened 2023-10-08
Check IfcDateTime according to ISO standards

johltn:


src/ifcopenshell-python/ifcopenshell/type_checker.py
```diff
@@ -0,0 +1,149 @@
...
+    %import common.WS
+    %ignore WS
+"""
+
+parser = Lark(grammar, start='time_value')
+
+
+class DatetimeTransformer(Transformer):
+    def year(self, items):
+        return ''.join(token.value for token in items)
```
aothms, line 29: How about something like this:

```suggestion
    def _integer(self, items):
        return int(''.join(token.value for token in items))

    year = _integer
```

and same as below.

I also added the `int()` I would convert to the appropriate type in the transformer already.

src/ifcopenshell-python/ifcopenshell/type_checker.py
```diff
@@ -0,0 +1,149 @@
...
+    time_value: year "-" month "-" day "T" hour ":" minute ":" second ["." millisecond]
+
+    hour: DIGIT DIGIT
+    minute: DIGIT DIGIT
+    second: DIGIT DIGIT
+    millisecond: DIGIT+
+
+    day: DIGIT DIGIT
+    month: DIGIT DIGIT
+    year: MINUS? DIGIT+
```
aothms, line 15: Wikipedia says: 

> ISO 8601 prescribes, as a minimum, a four-digit year

Should we add that to the grammar?

src/ifcopenshell-python/ifcopenshell/type_checker.py
```diff
@@ -0,0 +1,149 @@
...
+
+    MINUS: "-"
+    DIGIT: "0".."9"
+
+    %import common.WS
+    %ignore WS
+"""
+
+parser = Lark(grammar, start='time_value')
```
aothms, line 24: `start='time_value'` still needed? You already defined it in the grammar. Also suggest `datetime_value`.

src/ifcopenshell-python/ifcopenshell/type_checker.py
```diff
@@ -0,0 +1,149 @@
...
+def validate(d):
+    parsing = parse_datetime(d)
+    if parsing["status"] != -1:
+        if validate_datetime(parsing["msg"]) != 1:
+            print(validate_datetime(parsing["msg"]))
+    else:
+        print("syntax error")
+
+
+if __name__ == '__main__':
```
aothms, line 130: Please add these as a `def test_datetime()` in https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/test/test_validate.py

aothms: Also we should add some test cases of actual IFC file content to check whether the values of such types are correctly discovered. For those you can follow the same convention as here https://github.com/IfcOpenShell/IfcOpenShell/tree/v0.7.0/src/ifcopenshell-python/test/fixtures/validate

E.g create an IfcPropertyListValue with multiple IfcDateTime values. IfcDateTime is IfcSimpleValue is IfcValue.

aothms (changes requested):

src/ifcopenshell-python/ifcopenshell/type_checker.py
johltn: added https://github.com/IfcOpenShell/IfcOpenShell/pull/3864/commits/4a2f34a600a0326946042a6b6f2f302947943601

src/ifcopenshell-python/ifcopenshell/type_checker.py
johltn: changed in https://github.com/IfcOpenShell/IfcOpenShell/pull/3864/commits/5c5aaf16e072d9e96d7c2f99b0cff6945b897f3c

src/ifcopenshell-python/ifcopenshell/type_checker.py
johltn: fixed in https://github.com/IfcOpenShell/IfcOpenShell/pull/3864/commits/9021d126ef28ed3687eee3c6eed0d48c8ce4d42c

src/ifcopenshell-python/ifcopenshell/type_checker.py
```diff
@@ -0,0 +1,149 @@
...
+
+# Grammar to retrieve the following structure: YYYY-MM-DDThh:mm:ss
+grammar = """
+    start: time_value
+    time_value: year "-" month "-" day "T" hour ":" minute ":" second ["." millisecond]
+
+    hour: DIGIT DIGIT
+    minute: DIGIT DIGIT
+    second: DIGIT DIGIT
+    millisecond: DIGIT+
```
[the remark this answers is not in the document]
johltn: fixed in https://github.com/IfcOpenShell/IfcOpenShell/pull/3864/commits/0f75565236f2571671eeb184eb8a7f7ea9358842

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -365,6 +367,36 @@ def validate(f, logger, express_rules=False):
...
+                                    "For instance:\n    %s\n    %s\nInvalid attribute value %s for %s.%s",
+                                    inst,
+                                    annotate_inst_attr_pos(inst, i),
+                                    "syntax not conform to ISO",
+                                    entity,
+                                    attrs[i],
+                                )
+                                
+                except Exception as e:
+                    #todo: handle aggregations
```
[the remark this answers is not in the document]
johltn: added in https://github.com/IfcOpenShell/IfcOpenShell/pull/3864/commits/73834ea38467d99c174b47168ffbeac9cb6b765f

src/ifcopenshell-python/ifcopenshell/type_checker.py
johltn: test files created https://github.com/IfcOpenShell/IfcOpenShell/pull/3864/commits/d0abe8fa7afc77fc5e8bfd64c8aac91d56383e2f

aothms:
Fails when we run the tests https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/6607354944/job/17944635714#step:8:37

Probably needs to be a relative import. (relative) Python module import never fails to confuse me.

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -365,7 +412,61 @@ def validate(f, logger, express_rules=False):
...
+                try:
+                    type_of_attribute = attr.type_of_attribute()
+                    declared_type = type_of_attribute.declared_type()
+                    type_name = declared_type.name()
+
+                    if type_name == "IfcDateTime" and val:
+
+                        parsing = parse_datetime(val)
+
+                        if isinstance(parsing, DateErrorParsing().__class__):
```
aothms, line 424:
```suggestion
                        if isinstance(parsing, DateErrorParsing):
```

DateErrorParsing().__class__ == DateErrorParsing

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -365,7 +412,61 @@ def validate(f, logger, express_rules=False):
...
+                                )
+                        elif isinstance(validate_datetime(parsing), DateErrorContent().__class__):
+                            logger.error(
+                                    "For instance:\n    %s\n    %s\nInvalid attribute value %s for %s.%s",
+                                    inst,
+                                    annotate_inst_attr_pos(inst, i),
+                                    parsing.msg,
+                                    entity,
+                                    attrs[i],
+                                )
```
aothms, line 441: I think these two branches are equal?

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -271,6 +273,51 @@ def get_entity_attributes(schema, entity):
...
+            value.is_a(),
+            ifcopenshell.ifcopenshell_wrapper.entity,
+        ):
+            # top level entity instances will be checked on their own
+            pass
+        else:
+            # unpack the type instance
+            yield from check(value[0], value.is_a(), instance=instance)
+
+        if value.is_a() == "IfcDateTime" and value[0]:
```
aothms, line 314: This function still does the type check here, so can't be reused.

Rather yield the values and do the check outside of the function. That way the same function can also be used for checking uniqueness for example.

aothms, line 314: Is it safe to always use `value[0]`? We need to add some test files where the IfcDateTime is used as an actual attribute such as IfcLibraryInformation and not just a simple type in a select, such as IfcPropertyListValue.

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -365,7 +412,61 @@ def validate(f, logger, express_rules=False):
 
         if not has_invalid_value:
             for i, (attr, val, is_derived) in enumerate(zip(attrs, values, entity.derived())):
-
+                try:
```
aothms, line 415: Why is a try catch statement used here? Catching all subtypes of Exception silences a lot of other issues. Please rewrite to proper if statement.

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -365,7 +412,61 @@ def validate(f, logger, express_rules=False):
...
+                    type_of_attribute = attr.type_of_attribute()
+                    declared_type = type_of_attribute.declared_type()
+                    type_name = declared_type.name()
+
+                    if type_name == "IfcDateTime" and val:
```
aothms, line 420: I would implement validation of https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcDuration.htm in this PR as well. And really make sure that the code is as general and reusable as possible. So that validators are stored for example in a global mapping like:

```
type_validators = {
  'IfcDateTime: [parse_datetime, validate_datetime],
  'IfcDuration: [ ... ]
  ...
}
```

aothms (changes requested):

theoryshaw:
Hi @johltn 

PR #3888, opened 2023-10-14
Adding support for non ascii characters when reporting

isma3lMB:
Allowing non ascii content when reporting (e.g., accentuated French words in IDS Tilte, description ...)


aothms:
Thank you this makes sense. Probably most of the team here didn't notice this because utf-8 is the default on linux.

PR #3899, opened 2023-10-16
IfcGradientCurve geometry generation

RickBrice:
This PR fixes some problems with generating horizontal alignment geometry for circular curves and gets the basic approach for IfcGradientCurve with vertical profile working.


aothms:
This is great!

Issue #3770, opened 2023-09-19
When using Duplicate Aggregate it does not duplicate the wall connections. 


theoryshaw:
When using [Duplicate Aggregate](https://github.com/IfcOpenShell/IfcOpenShell/pull/3328) it does not duplicate the wall connections. 

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/c92ba887-bca2-4c6f-9e2c-002c6b812f79

brunoperdigao:
It seems that `blenderbim.core.root.copy_class()` copies the `IfcRelConnectsElements` but not the `IfcRelConnectsPathElements`. @Andrej730 could you point me to where I can find the source of this problem in the code?

brunoperdigao:
I just found that this is the intended behavior. As stated in IfcOpenShell documentation (https://blenderbim.org/docs-python/autoapi/ifcopenshell/api/root/copy_class/index.html?highlight=copy_class):
> Path connectivity is not copied, as there is no guarantee that the connections are still valid.

The IfcOpenShell api is beyond my knowledge right now 😕

Andrej730:
@brunoperdigao you're right, it does not but copying wall connections sounds very similar to how we copy some other relationships during duplication by storing them

https://github.com/IfcOpenShell/IfcOpenShell/blob/6c9e130ca831838b4e22361a606940a41c55031f/src/blenderbim/blenderbim/bim/module/geometry/operator.py#L755

and then recreating them

https://github.com/IfcOpenShell/IfcOpenShell/blob/6c9e130ca831838b4e22361a606940a41c55031f/src/blenderbim/blenderbim/bim/module/geometry/operator.py#L810

So, perhaps we can extend `tool.Root.get_decomposition_relationships` and `tool.Root.recreate_decompositions` to store connections too? Currently it works only for fills.

brunoperdigao:
@Andrej730 Thanks, I'll take a look at this.

brunoperdigao:
@theoryshaw This should be fixed after merging #3969.

theoryshaw:
Yes, this looks good to go!

Issue #2187, opened 2022-05-04
Improvement for duplicating an IfcElementAssembly


Jesusbill:
Currently duplicating an IfcElementAssembly clones only the "empty" object, while the aggregates must be cloned and attached after.
Should this include the aggregates by default, eventually, for faster copy-paste operations?
The main issue is the empty not being in a collection of its own, even without parts, that requires first to attach through the UI the first duplicated aggregated so as to be able to drag n' drop the rest.
The image shows the result from `Shift D`'ing all the contents of an ElementAssembly

![image](https://user-images.githubusercontent.com/6861582/166829848-bb0229b9-adac-4c33-9606-34397ed93f36.png)

PR #3902, opened 2023-10-17
Improvements on `OverrideDuplicateMove`

brunoperdigao:
This PR aims to improve IFC object duplication, by implementing a solution for #3770, #2187, #3789.

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/geometry/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/geometry/operator.py
@@ -858,17 +858,30 @@ class OverrideDuplicateMove(bpy.types.Operator):
 
     @staticmethod
     def recreate_assembly(old_to_new):
-        for old_element, new_element in old_to_new.items():
-            old_parent = ifcopenshell.util.element.get_aggregate(old_element)
-            if old_parent:
-                new_parent = old_to_new[old_parent]
+        for old, new in old_to_new.items():
+            old_aggregate = ifcopenshell.util.element.get_aggregate(old)
+            if old_aggregate:
+                new_aggregate = old_to_new[old_aggregate]
                 blenderbim.core.aggregate.assign_object(
                                             tool.Ifc,
                                             tool.Aggregate,
                                             tool.Collector,
-                                            relating_obj=tool.Ifc.get_object(new_parent[0]),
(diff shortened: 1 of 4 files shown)


src/blenderbim/blenderbim/bim/module/geometry/operator.py
```diff
@@ -846,6 +856,31 @@ def process_arrays(self, context):
 
         return arrays_to_create, array_children
 
+    @staticmethod
+    def recreate_assembly(old_to_new):
```
brunoperdigao, line 860: Should this be moved to `.../tool/root.py` ?

brunoperdigao:
@Moult I noticed that the BIM tests only account for the Blender side of things. I want to create a test to check wall connections after duplicating them, but I'm not finding a way to this using only Blender, without using IfcOpenShell. Are connections stored in Blender object properties?

theoryshaw:
@brunoperdigao this this pr, have this functionality? 

image

brunoperdigao:
No, this PR only changes things in the normal duplication (shift + d). I need this to be merged first in order to create the PR for the duplicate aggregate refactor. But once this is ready it should be easy to implement the rest.
For this PR it will be useful to test 3 things:
- Select all the objects of an aggregate and shift + d, it should duplicate it (but without any linking property)
- Select two objects that share connection and duplicate them. The connection in the new objects should be reconstructed like the original ones
- The same as the second but with walls that share a connection

theoryshaw:
Thanks!  I assuming it was this branch, correct? https://github.com/IfcOpenShell/IfcOpenShell/compare/v0.7.0...brunoperdigao:IfcOpenShell:refactor-duplicate-aggregate

brunoperdigao:
Yes, I think I'll create the PR already as a draft so you can give me some insights

Moult:
@brunoperdigao there's nothing wrong with checking IOS IFC data in the BIM tests, just that we haven't done so yet (simply because either it's already captured elsewhere or the test was only written as a smoke test not a comprehensive test). Go ahead and write new sentence definitions for it :)

theoryshaw:
> Yes, I think I'll create the PR already as a draft so you can give me some insights

sounds great. Have some ideas. :)

brunoperdigao:
@Moult I finished all the changes that I had planned for this. Let me know if you find there is anything else to improve.

PR #3923, opened 2023-10-22
When adding a context, check to see if it already exists first.

theoryshaw:


Moult:
Sorry, rejecting this because contexts may have different target scales (which currently isn't very well exposed)

PR #3933, opened 2023-10-24
v0.8.0 - fix clang-tidy warnings

dirkolbrich:
This PR is a working progress to start fixing clang-tidy warnings. Each warning type gets a dedicated commit across all files in `src/ifcparse`, except the generated schema files by `src/ifcopenshell-python/ifcopenshell/express/express_parser`. these should be fixed by refactoring the `express_parser`.

Some more rewrites within this PR:
- unify private member variable names to use trailing underscore

### DIFF
--- a/src/examples/IfcAlignment.cpp
+++ b/src/examples/IfcAlignment.cpp
@@ -25,12 +25,13 @@
 // Sections and page number for this document are cited in the code comments.
 
 // Disable warnings coming from IfcOpenShell
-#pragma warning(disable:4018 4267 4250 4984 4985)
+#pragma warning(disable : 4018 4267 4250 4984 4985)
 
-#include "../ifcparse/IfcHierarchyHelper.h"
 #include "../ifcparse/Ifc4x3_add2.h"
+#include "../ifcparse/IfcHierarchyHelper.h"
 
 #include <boost/math/constants/constants.hpp>
+#include <fstream>
 
 const double PI = boost::math::constants::pi<double>();
 double to_radian(double deg) { return PI * deg / 180; }
--- a/src/examples/IfcParseExamples.cpp
+++ b/src/examples/IfcParseExamples.cpp
@@ -20,8 +20,9 @@
 // TODO: Multiple schemas
 #define IfcSchema Ifc2x3
 
-#include "../ifcparse/IfcFile.h"
 #include "../ifcparse/Ifc2x3.h"
+#include "../ifcparse/IfcFile.h"
+#include "../ifcparse/IfcLogger.h"
 
 #if USE_VLD
 #include <vld.h>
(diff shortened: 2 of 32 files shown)


aothms:
> warning unused-includes

I didn't know that existed. I've used [include what you use](https://github.com/include-what-you-use/include-what-you-use) previously, but if you get these as direct compiler warning feedback nowadays that's great. Compile time of ifcopenshell is still orders of magnitude larger than what I hope it to be and maybe reducing the amount of includes is a step in the right direction. I've also contemplated a bit using an approach like sqlite that before building you do some sort of concatenation step and filter out the duplicate includes.

dirkolbrich:
Haven't done a lot in the C++ source lately. Maybe next year. 
Precompiled headers would be my first thing to try. Another thing would be C++23 modules. But as long as this standard version isn't even official and more widespread supported, it's just a guess.

aothms:
I'm also a bit hesitant with modules, but maybe it's not justified I don't know. Precompiled headers indeed sounds like a sane alternative to manual concatenation (https://sqlite.org/amalgamation.html) of the translation unit sources, but last time I tried I ran into some issues. It's not something I have a lot of experience with.

dirkolbrich:
Have you tried Ccache as another option? It saves some time especially on subsequent builds. Branch v0.8 has a setup for Ccache in the Cmake option.

dirkolbrich:
@aothms just a head-up: Is there anything blocking this PR from merging? I would like to move forward in improving the current ifcparse module and are happy to implement any changes you might have. ty

aothms:
Sorry, it's just that sometimes I'm burried in deadlines, and it's a bit tragic that still there isn't really a sense shared ownership of the C++ code parts. Which is why I'm even happier with your dedication to this.

There is definitely an interest in a more modern codebase. Very high on my list is seeing if we can leverage std::tuple as the storage instead of the virtual Argument classes in IfcEntityInstanceData. 

I even learned something new. I didn't know `auto*` was something.

```diff
-    for (auto attribute : attributes_) {
+    for (const auto* attribute : attributes_) {
```

I went through all the changes in a whitespace agnostic diff (`git diff -w a60e272eee613b68dd98581dfc8654a29199bf06 dirkolbrich/v0.8.0-fix-clang-warnings`) and it looks great to me. Thanks again for all your efforts!

dirkolbrich:
Thank you for merging. Sorry, I didn't wanted to come over too harshly. I understand your time constraints as I'm able to work on the code only sporadically as well. Someone's gonna pay the rent.

Apart from thinking about a refactored API for IfcParse (mentioned here [refactor IfcParse API](https://github.com/IfcOpenShell/IfcOpenShell/discussions/3787)), which need's a lot more thinking time on my side, I will try to further refactor IfcParse. Here is my roadmap for that:

- [x] introduce standard code formatting -> clang-format
- [x] introduce compile time checks -> clang-tidy
- introduce performance checks -> perf?
- separate large code files, use of .h and .cpp files, e.g. there is no IfcFile.cpp file but code definitions for IfcFile.h are located in IfcParse.cpp
- unify naming styles for file names, namespaces, classes, variables, members, methods etc., currently there is a wild mix of everything -> I started a Coding style guide file, will share it for discussion later
- separate the schema files, which are generated, into its own `/schemas` subfolder
- make the ifcparse library its own self containing cmake project, the main cmake file just uses add_subfolder()
- remove the Boost dependency where it is not needed anymore, prefer the standard library
- adding unit tests
- adding benchmark tests
- rewrite for more modern C++17 features
- maybe move to C++20?
- maybe move to C++23 and make the library a module?

PR #3936, opened 2023-10-25
Fixes runtime check for expected IfcPolynomialCurve coefficients and adds fallback check to IfcObjectPlacement

RickBrice:
I tweaked @aothms IfcObjectPlacement implementation to check for the presence of a fallback placement and compare it to the computed placement for IfcLinearPlacement.


aothms:
> check for the presence of a fallback placement 

Sorry... forgot to merge that back in.

PR #3943, opened 2023-10-26
Update unit.py - convert function, case issue

ceegartner:
Hi,
I noticed an issue and suggest a fix:
Unit names from <si_conversions> are all lowercase, whereas project units extracted from the IFC files with <get_project_unit(...).Name> are uppercase. So the convert function didn't seem to work properly. Hope this helps!

(it's my first time contributing to an open-source GitHub project, I hope I didn't mess anything up, please tell me if this is the case!)


aothms:
Thanks, looks good to me. Indeed IfcConversionBasedUnit has name as a simple string label, so it's good to normalize the case.

PR #3945, opened 2023-10-27
Implements framework for adjusting geometry based on start location next segment

RickBrice:
There are two things in this PR.
1. the implementation of piecewise_function::evaluate is moved from the header to the cpp to prevent massive rebuilds when this function is modified.
2. Implemented a framework to adjust the geometry of a segment based on the start point of the next segment. This will also be used for defining cant cross slopes that are defined by the placement information at the start of sequential segments. Implementing this framework was a bit involved and might be difficult for others to follow. The code is heavily commented to aid in understanding.


RickBrice:
The commit history looks crazy. I rebased my fork onto the current 0.8 but there are still conflicts. There are only 2 new commits.
Not enough of a github expert to know what to do. please advice.

aothms:
Thanks for taking the time to comment so extensively. It's a pleasure to read.

Initially we thought about applying the correction outside of the functor, which would probably remove the necessity for selectively disabling the correction to get the correction vector, and would not recurse all the way to the end. On the other hand we would likely not be able to as granularly control the lerping as we do now based on transition code (really nice), so I understand and support this choice.

Questions:
- When we recurse from start to end, how many segments do we expect in a somewhat pathological case. Would it trigger a stack overflow? We can also not disable correction easily for the next segment, because caching would cause the uncorrected segment to be retained (?).
- Your current definition of interpolate is not really the interpolation, but calculating the component to add to `a` in order to interpolate. Which is not a problem, but could confuse readers I guess. I would probably also change the function to a template, (see below) and then call it on matrix blocks (you never know what kind of SIMD instructions Eigen might end up using) `p.block<3,3>(0,0) = interpolate(u, end_this.block<3,3>(0,0), start_next.block<3,3>(0,0), length)`.
- After interpolation, I guess the chance is considerable that the matrix is no longer orthogonal and the columns are no longer normalized. Should we do something about that?
- Why do you only lerp matrix rows 0 and 1, and not the Z component?
- I also don't fully understand why `segment_geometry_adjuster` is in charge of applying the `inst->Placement()` as opposed to merely doing the adjustment (before or after applying the matrix). But maybe this has to do with this cant placement interpolation thing that I haven't really understood yet.

templated lerp

```C++
template <typename T>
auto interpolate = [](double u, const T& a, const T& b, double l) -> double {
    return l == 0.0 ? 0.0 : (u / l) * (b - a) ; 
};
```


Regarding the git workflow. The 'easiest' approach is to:

1. before work, do a `git fetch --all && git checkout -b feature_something_new origin/v0.8.0`; this creates a new feature branch from the tip of v0.8 as it exists on the remote
2. commit as desired
2. `git push origin` to push your feature branch to the remote
2. create a pull request using the web interface
2. after merging you can do `git checkout v0.8.0 && git pull`
3. but don't commit to v0.8 directly, create a new feature branch as in step 1.; because when you do your own local v0.8 will likely diverge from the remote v0.8 (because of committed suggestions, squashing, other work being merged in, etc.).

RickBrice:
> When we recurse from start to end, how many segments do we expect in a somewhat pathological case. Would it trigger a stack overflow? We can also not disable correction easily for the next segment, because caching would cause the uncorrected segment to be retained (?).

That's a great question - I have the same concern. One option is to loop from end to start - but this might be difficult for IfcCompositeCurve because the curve might not be for an alignment and the segments might be IfcCompositeCurveSegment instead of IfcCurveSegment. Looping in reverse order might work well for gradient and segmented reference curves.

I don't have experience with long roadway/highway projects or railways (just bridges) so I don't have a feel for how many twists and turns could be expected along an alignment. The number might multiply quickly if models start to cover board to board alignments across an entire country or continent (are those federated models?). @civilx64 might be able to provide some guidance.

For now, the current implementation works, I see the potential pitfall of stack overflow, I'll continue to seek a more robust solution.

> Your current definition of interpolate is not really the interpolation, but calculating the component to add to a in order to interpolate. Which is not a problem, but could confuse readers I guess. I would probably also change the function to a template, (see below) and then call it on matrix blocks (you never know what kind of SIMD instructions Eigen might end up using) p.block<3,3>(0,0) = interpolate(u, end_this.block<3,3>(0,0), start_next.block<3,3>(0,0), length).

Perhaps rename `interpolate` to `compute_adjustment`.

I'll make the change to the template function and call it as suggested with matrix blocks. My experience with Eigen is still limited so I appreciate you pointing out better ways of using the library.

> After interpolation, I guess the chance is considerable that the matrix is no longer orthogonal and the columns are no longer normalized. Should we do something about that?

Oh yes - I made myself a note to address that and, in my rush to get this done before the weekend started I forgot to implement.

Do you have a suggestion on the best way to normalize the column vectors with Eigen?

> Why do you only lerp matrix rows 0 and 1, and not the Z component?

My intention is to lerp matrix columns 0 and 1. You say rows. Is that a typo or are we misaligned on a fundamental concept?

In the context of alignments IfcCompositeCurve, IfcGradientCurve, and IfcSegmentedReferenceCurve, in a 2D (or pseudo 2D) plane. Composite curve is XY plan view plane. Gradient and segmented reference curves are in the Distance Along, Y plane.

> I also don't fully understand why segment_geometry_adjuster is in charge of applying the inst->Placement() as opposed to merely doing the adjustment (before or after applying the matrix). But maybe this has to do with this cant placement interpolation thing that I haven't really understood yet.

Near the end of `taxonomy::ptr mapping::map_impl(const IfcSchema::IfcCurveSegment* inst)` we created a lambda function for the piecewise curve that evaluates the eval_ function returned from curve_segment_evaluator.evalution_function() and then premultipled it with the IfcCurveSegment placement matrix. This lambda is then stored in the piecewise curve and called later. 

~~~
	auto& eval_fn = cse.evaluation_function();
	if(!eval_fn) throw std::runtime_error(inst->ParentCurve()->declaration().name() + " not implemented");
	auto fn = *eval_fn;
	auto length = fabs(cse.length());

	auto transformation_matrix = taxonomy::cast<taxonomy::matrix4>(map(inst->Placement()))->ccomponents();

	auto fn_transformed = [fn, transformation_matrix](double u)->Eigen::Matrix4d {
        Eigen::Matrix4d f = fn(u);
        Eigen::Matrix4d result = transformation_matrix * f;
        return result;
		};

	// @todo it might be suboptimal that we no longer have the spans now
	auto pwf = taxonomy::make<taxonomy::piecewise_function>();
	pwf->spans.push_back({ length, fn_transformed });
~~~

The matrix segment_geometry_adjuster applies the geometric corrections to must have the inst->Placement matrix applied to the return matrix of eval_ before geometric adjustments can be applied. The reason for this is illustrated in the sketch below. It is a simple composite curve with four segments (the last being a zero length segment). The ParentCurves are a circle and line with origins at (0,0). eval_ is the function for the ParentCurve. The end points of the ParentCurves don't align (and they don't need to). The CurveSegment placement must be applied to move the curve segments from the parent curve coordinate system to the curve segment coordinate system where the geometry adjustment can then be applied.

Applying inst-Placement() in segment_geometry_adjuster seemed to be necessary in order to properly apply any geometry corrections.

> edit: maybe it's also more efficient to calculate diff = start_next - end_this outside of apply_adjustments(), but it's a bit harder I guess with the virtual functions. feel free to ignore.

I think you are right - if I make segment_geometry_adjuster::set_segment_end_point virtual, subclasses can compute and cache information that is constant and used repetitively (like diff - start_next - end_this). I will investigate this option.

thumbnail_IMG_0209

RickBrice:
I just watched you BIM Voice interview from a year ago-very cool. 

Related to workflow. I've been working in a forked repository. Your recommendations seem to imply working directly from the IfcOpenShell repository on a branch from v0.8.0. Is this correct? Or should I continue to used a fork?

aothms:
> For now, the current implementation works, I see the potential pitfall of stack overflow, I'll continue to seek a more robust solution.

Fully agree. Let's not loose time on every detail. It's good to be aware that there is some kind of limit. 

> Do you have a suggestion on the best way to normalize the column vectors with Eigen?

From Wikipedia I found

https://en.wikipedia.org/wiki/Orthogonal_matrix#Nearest_orthogonal_matrix

which I guess could be

```C++
#include <unsupported/Eigen/MatrixFunctions>

auto ortho = M * ((M.transpose() * M)).sqrt().inverse();
```

(I don't know why for matrix sqrt we need to include `unsupported/Eigen`)

> Is that a typo 

For each column you seem to lerp cell 0 and 1, which I summarized as lerping rows, which is indeed confusing. But your explanation is clear. If I were to summarize, this is before composing the various 2d functions, into 3d(+cant). Makes sense, and I should have seen that I guess.

> Applying inst-Placement() in segment_geometry_adjuster seemed to be necessary in order to properly apply any geometry corrections.

Thanks. I see your point. Correction can only be obtained after applying the placements. I guess I would have done placement in the `curve_segment_evaluator` and only the correction in the adjuster, but practically there is no difference and I guess it even saves a couple of lines of code. So let's keep it like this.

aothms:
> Your recommendations seem to imply working directly from the IfcOpenShell repository on a branch from v0.8.0. Is this correct? Or should I continue to used a fork?

It doesn't matter. But it's a detail that I indeed omitted. When you push a new branch to your own fork, Github will still create a PR for you on the official IfcOpSh/IfOpSh repo. Because GH knows it's a fork.

RickBrice:
I've been thinking your comment about applying the correction outside of the functor. I think this is possible and will result in a much cleaner implementation. We wont give up any granularity control for the IfcTransitionCode.

A scratch implementation of map_impl(IfcSegmentCurve) looks like this:

~~~
taxonomy::ptr mapping::map_impl(const IfcSchema::IfcCurveSegment* inst) {
   // Find the next segment after inst
   const IfcSchema::IfcCurveSegment* next_inst = nullptr;
   auto composite_curves = inst->data().getInverse(&IfcSchema::IfcCompositeCurve::Class(), 0);
    if (composite_curves) {
       if (composite_curves->size() == 1) {
            auto segments = (*composite_curves->begin())->as<IfcSchema::IfcCompositeCurve>()->Segments();
            bool emit_next = false;
            for (auto& s : *segments) {
               if (emit_next) {
                next_inst = s->as<IfcSchema::IfcCurveSegment>();
                   break;
               }
               if (s == inst) {
                   emit_next = true;
               }
            }
       }
       else {
            Logger::Warning("IfcCurveSegment belongs to multiple IfcCompositeCurve instances. Cannot determine the next segment. Geometry adjustments will not be made.");
       }
    }

   bool is_horizontal = false;
   bool is_vertical = false;
   bool is_cant = false;

   {
      aggregate_of_instance::ptr segment_owners = inst->data().getInverse(&IfcSchema::IfcCompositeCurve::Class(), 0);
      if (segment_owners) {
         for (auto& cc : *segment_owners) {
            if (cc->as<IfcSchema::IfcSegmentedReferenceCurve>()) {
               is_cant = true;
            }
            else if (cc->as<IfcSchema::IfcGradientCurve>()) {
               is_vertical = true;
            }
            else {
               is_horizontal = true;
            }
         }
      }
   }

   if ((is_horizontal + is_vertical + is_cant) != 1) {
      // We have to choose the correct functor based on usage. We can't 
      // support multiple, because we don't know the caller at this point.
      return nullptr;
   }

   auto segment_type = is_horizontal ? ST_HORIZONTAL : is_vertical ? ST_VERTICAL : ST_CANT;

   curve_segment_evaluator cse(this, inst, next_inst, length_unit_, segment_type);
   boost::mpl::for_each<curve_seg_types, boost::type<boost::mpl::_>>(std::ref(cse));
   
   auto& eval_fn = cse.evaluation_function();
   if(!eval_fn) throw std::runtime_error(inst->ParentCurve()->declaration().name() + " not implemented");
   auto fn = *eval_fn;
   auto length = fabs(cse.length());

   auto inst_placement = taxonomy::cast<taxonomy::matrix4>(map(inst->Placement()))->ccomponents();
   auto next_placement = taxonomy::cast<taxonomy::matrix4>(map(next_inst->Placement()))->ccomponents();

   // Call abstract factory method to create a curve_segment_adjuster
   // Concrete types could be "do nothing", linear or something entirely new
   // Could also pass in inst and next_inst and determine curve_segment_adjuster
   // concrete type based on ParentCurve type (do nothing for IfcLine, linear adjustment for IfcSpiral)
   // The sky is the limit here
   std::shared_ptr<curve_segment_adjuster> segment_adjuster = create_segment_adjuster(inst_placement, next_placement); 

    auto fn_transformed = [fn, inst_placement,segment_adjuster](double u) -> Eigen::Matrix4d {
        Eigen::Matrix4d f = fn(u);
       Eigen::Matrix4d result = inst_placement * f;
        return segment_adjuster->apply_adjustment(u, result);
    };

   // @todo it might be suboptimal that we no longer have the spans now
   auto pwf = taxonomy::make<taxonomy::piecewise_function>();
   pwf->spans.push_back({length, fn_transformed});
   pwf->instance = inst;
   return pwf;
}
~~~

However, this will not solve the need to recursively traverse to the last segment.

>   auto next_placement = taxonomy::cast<taxonomy::matrix4>(map(next_inst->Placement()))->ccomponents();

This line will recurse into `mapping::map_impl(const IfcSchema::IfcCurveSegment* inst)` to get the placement next_inst which will in turn get the placement of it's next_inst, and so on until there is no next_inst.

The only solution to the recursion problem I can see is iterating over the IfcCompositeCurve.Segments in reverse order. For each _Segment i-1_, the Placement of _Segment i_ is already computed.


This may be a little tricky since IfcCompositeCurve.Segments are IfcSegment types which can be IfcCompositeCurveSegment or IfcCurveSegment. Given CT 4.1.7.1.1 and the subsections, I think we can assume that a mixture of IfcCompositeCurveSegment and IfcCurveSegment is not permitted. The ParentCurve of IfcCompositeCurveSegment is required to be a BoundedCurve per Formal proposition 8.9.3.22.4 whereas ParentCurve of IfcCurveSegment is any curve, even unbounded, and the curve bounds are defined by IfcCurveSegment.SegmentStart and .SegmentLength.

Given that, we could look at the type of the first segment in IfcCompositeCurve.Segments and determine the direction of iteration.

From `mapping::map_impl(const IfcSchema::IfcCompositeCurve* inst)`
~~~
...
 	for (auto& segment : *segments) {
		if (segment->as<IfcSchema::IfcCompositeCurveSegment>() && segment->as<IfcSchema::IfcCompositeCurveSegment>()->ParentCurve()->as<IfcSchema::IfcLine>()) {
			Logger::Notice("Infinite IfcLine used as ParentCurve of segment, treating as a segment", segment);
			double u0 = 0.0;
			double u1 = segment->as<IfcSchema::IfcCompositeCurveSegment>()->ParentCurve()->as<IfcSchema::IfcLine>()->Dir()->Magnitude() * length_unit_;
			if (u1 < conv_settings_.getValue(ConversionSettings::GV_PRECISION)) {
				Logger::Warning("Segment length below tolerance", segment);
			}
...
~~~

we could instead do something like

~~~
   if(segments->front()->as<IfcSchema::IfcCompositeCurveSegment>())
     Iterate forward
  else if (segments->font()->as<IfcSchema::IfcCurveSegment>())
     Iterate in reverse
~~~

@aothms what do you think?

civilx64:
> I don't have experience with long roadway/highway projects or railways (just bridges) so I don't have a feel for how many twists and turns could be expected along an alignment. The number might multiply quickly if models start to cover boarder to boarder alignments across an entire country or continent (are those federated models?). @civilx64 might be able to provide some guidance.

There are no hard and fast rules on what constitutes a "long" alignment but I'll provide some feedback based on 20+ years in industry.  For a typical design project, yes you would expect federated models and alignment lengths generally not more than 5 mi / 8 km in length and maybe up to 40 segments.

The longest alignment I can remember working with was 30 miles / 48 km with maybe 150 segments for preliminary design during the environmental phase.  Close behind that would be a regressed rail alignment of 20 miles / 32 km but with maybe 250-300 segments based on its urban context.

So for a high level order-of-magnitude summary I would say up to 50 segments in a single alignment is fairly common, with up to 500 segments in a single alignment starting to get pretty rare.  Also I just realizing I'm thinking purely in terms of horizontal segments but if you had multiple vertical alignments and a few cant alignments then the overall total number of segments could be quite higher, maybe as many as a few thousand total for an alignment.  Admittedly from review of the code I'm not 100% certain if we're talking about potential stack overflow of segments just for  `IfcSegmentedReferenceCurve` or for all three components including `IfcGradientCurve` and `IfcCompositeCurve`.

RickBrice:
@civilx64 Potential stack overflow for IfcCompositeCurve and its subclasses IfcGradientCurve and IfcSegmentedReferenceCurve.... but I think I have a solution as described above.

aothms:
> we could instead do something like
> 
> ```
>    if(segments->front()->as<IfcSchema::IfcCompositeCurveSegment>())
>      Iterate forward
>   else if (segments->font()->as<IfcSchema::IfcCurveSegment>())
>      Iterate in reverse
> ```
> 
> @aothms what do you think?

Let's do that.

There is still one little caveat in the sense that an individual IfcCurveSegment could also be the direct representation of an IfcAlignmentSegment while simultaneously being defined as a part in the composite curve. In the proposed solution, when evaluating the geometry for the *segment* we would still forwardly recurse through the composite, because locally in the mapping function we don't know that we're evaluating the segment purely in the context of that single segment.

I'm personally more interested in seeing the full picture first, before worrying about these kind of corner cases of stack overflow. One brute force way is also to count recursion dept in mapping->map() and return (uncached) nullptr if we exceed some value. I think that would also be a way to solve this (and other recursive problems that might wrongfully exist in IFC structures). (with 500 segments I think the chance of blowing up the stack is considerable though, due to the frame size of all these lambdas and the fact that there's quite some indirections in a single mapping call anyway).

I think the backwards approach makes sense computationally anyway, because that is how the corrections propagate.

I also agree on requiring uniform composite curve member type. I think that would be something to enforce as an express rule in some addendum. I've made a note.

RickBrice:
@aothms You are right - I'm not looking at the full picture and should be. 

> an individual IfcCurveSegment could also be the direct representation of an IfcAlignmentSegment while simultaneously being defined as a part in the composite curve.

The first thing we have in `mapping::map_impl(const IfcSchema::IfcCurveSegment* inst)` is code that gets the next segment. This needs to be improved to support CT 4.1.4.4.1.1 and CT 4.1.7.1.1.4 where the IfcCurveSegment is related directly to the IfcAlignmentSegment and the next IfcAlignmentSegment comes from the IfcRelNests.RelatedObjects. On the surface it seems like we can iterate over IfcCurveSegment's in either case - only the data structure differs but the concept is the same.

I've pick up some new IFCOS concepts over the last couple of weeks (like data().getInverse()). I'll attempt to derive a solution.

RickBrice:
IfcCurveSegment must be part of a composite curve based on the UsingCurves description. I think something like this rough code sketch covers all the cases.

~~~
mapping::map_impl(IfcCurveSegment* inst)
{
// delegates mapping to owning composite curve.
   curves = inst->data().getInverse(...)
   for (c : curves)
   {
      If c->as<IfcCompositeCurve>()
          map(c->as<IfcCompositeCurve>());
      Else if IfcGradientCurve
								map(gradient curve)
						else
								map(segmented reference curve)
   }

// look up mapping of this segment from cache
   Return cache_[inst->hash()];
}

mapping::map_impl(IfcCompositeCurve* inst)
{
    For each segment in reverse order
						{
						map_curve_segment(segment)
						add result to piecewise function 
						}
}

Similar map_impl overloads for gradient and segmented reference curves
~~~

`map_curve_segment` would do the actual mapping of a segment and cache the result using the implementation in the current map_impl(IfcCurveSegment*)

aothms:
> On the surface it seems like we can iterate over IfcCurveSegment's in either case - only the data structure differs but the concept is the same.

> IfcCurveSegment must be part of a composite curve based on the UsingCurves description. I think something like this rough code sketch covers all the cases.

Excellent find. These inverses can be so hard to discover. I wasn't aware of this particular one. So even if we have individual segment geometry, there needs to be a composite curve in the model. Then I don't think we need to actually change our approach. Except maybe for limiting recursion when retrieving individual segment geometry, but I'm sure we can patch this in somewhere. Implementation detail like this I can contribute, but I really depend on you for the domain specific geometry. So I think we should prioritize that.

RickBrice:
The one thing that is still kind of strange, from [CT 4.1.7.1.1.4](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/concepts/Product_Shape/Product_Geometric_Representation/Alignment_Geometry/Alignment_Geometry_-_Segments/content.html), is that a single IfcAlignmentSegment can have multiple IfcCurveSegment in its IfcShapeRepresentation.Items.

image

I guess it is valid from a geometric standpoint. We can deal with it since the IfcCurveSegment must be attached to an IfcCompositeCurve. But it doesn't seem logical. I would expect a 1-to-1 relationship between an IfcAlignmentSegment and an IfcCurveSegment, especially since the IfcAlignmentSegment.DesignParameters (cardinality = 1) seem to be applicable to a single segment. 

Consider IfcAlignmentHorizontalSegment. There is a single StartPoint, StartDireciton, StartRadius, SegmentLength, etc allowing the DesignParameter to reflect the properties of only a single segment. How does IfcAlignmentHorizontalSegment.DesignParameters define the business logic of segments defined as IfcShapeRepresentation.Items in a sequence of IfcCurveSegment for a linear, clothoid, arc, clothoid, linear arrangement?

image


Maybe there needs to be a rule that restricts IfcShapeRepresentation.Items to [0..1] for IfcAlignmentSegment?

civilx64:
> The one thing that is still kind of strange, from [CT 4.1.7.1.1.4](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/concepts/Product_Shape/Product_Geometric_Representation/Alignment_Geometry/Alignment_Geometry_-_Segments/content.html), is that a single IfcAlignmentSegment can have multiple IfcCurveSegment in its IfcShapeRepresentation.Items.

The cardinality of `IfcShapeRepresentation.Items` is [1:?] or more because that's what is defined in the express schema for the `IfcShapeRepresentation` type.  I agree that in practice for `IfcAlignmentSegment` it should be [1:1] similar to `IfcAlignmentSegment.DesignParameters`.

> Consider IfcAlignmentHorizontalSegment. There is a single StartPoint, StartDireciton, StartRadius, SegmentLength, etc allowing the DesignParameter to reflect the properties of only a single segment. How does IfcAlignmentHorizontalSegment.DesignParameters define the business logic of segments defined as IfcShapeRepresentation.Items in a sequence of IfcCurveSegment for a linear, clothoid, arc, clothoid, linear arrangement?

It doesn't. `IfcAlignmentHorizontalSegment.DesignParameters` can only describe a single segment, so you need 5 instances of `IfcAlignmentHorizontalSegment` to define linear, clothoid, arc, clothoid, linear.

> Maybe there needs to be a rule that restricts IfcShapeRepresentation.Items to [0..1] for IfcAlignmentSegment?

I'll defer to @aothms and others on whether it should be a `WHERE` rule in the schema or if it's better documented elsewhere, say maybe as an implementor's agreement.

aothms:
> > The one thing that is still kind of strange, from [CT 4.1.7.1.1.4](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/concepts/Product_Shape/Product_Geometric_Representation/Alignment_Geometry/Alignment_Geometry_-_Segments/content.html), is that a single IfcAlignmentSegment can have multiple IfcCurveSegment in its IfcShapeRepresentation.Items.
> 
> The cardinality of `IfcShapeRepresentation.Items` is [1:?] or more because that's what is defined in the express schema for the `IfcShapeRepresentation` type. I agree that in practice for `IfcAlignmentSegment` it should be [1:1] similar to `IfcAlignmentSegment.DesignParameters`.

Agree, don't read too much into this. This is simply a limitation of this templating language where we define a tree/graph of class instances (potentially constraining certain values [e.g =Axis]), but without the ability to constrain cardinalities.

I think this is a good rule to specify in Gherkin, because it's a simple constraint on cardinalities. I think there is some potential of integrating the gherkin rules and the templates at some point (because we have explicit applicability information we can potentially render those gherkin rules alongside the templates), which would be harder with the imperative python code Scott is authoring. https://github.com/buildingSMART/ifc-gherkin-rules/issues/105

The alternative is to implement it in express. We actually have something very similar for BoundingBox https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcShapeRepresentationTypes.htm But this results in a completely intelligible error report to the user.

RickBrice:
After all the discussion, I did not make any major changes. I'm staying the course. I'll address stack overflow issues if they become a real problem. I'm uncertain about the quality of the segment geometry adjustments. I think that (1) it probably isn't needed in most cases and (2) we can do better than linear once we encounter real cases and better understand the geometric issues. There is a framework in place for future improvements.

@aothms do you want to merge this PR?

I'll refocus on IfcSegmentedReferenceCurve and cant geometry.

aothms:
Fully agree. Thanks again and let me know of there's anything I can do to assist.

RickBrice:
@aothms I guess there was one last thing related to adjusting geometry that I was going to ask you.
Right now, I've got two objects for adjusting geometry (do nothing and linear). The choice of which one to use is basically hard coded.

Could a setting by added to IfcGeom::IteratorSettings that indicates which geometry adjustment strategy to use?

This would allow us to make runtime decisions or change a hard coded setting in a little driver program, rather than recompiling the entire set of libraries every time we want to experiment with a strategy.

aothms:
Sure thing. I will work on that, but probably use it as an opportunity to refactor things a bit, because it's quite painful now to go through boost program options, iteratorsettings, conversionsettings. None of this currently makes a lot of sense. Also the int in which we stuff all these settings as bit flags is overflowing. I'll get back to you on this.

RickBrice:
@aothms is there an easy way to visualize the alignment geometry? I've been dumping xy points to a file, importing them to excel and plotting them. It's a bit tedious.

aothms:
Well, things are tedious to begin with due to the slow compilation times. We should perhaps investigate precompiled headers and incremental builds, Or maybe even `cling` to speed up the development of these kind of things in C++. https://blog.llvm.org/posts/2020-11-30-interactive-cpp-with-cling/

I think some light python scripting will probably help to speed up this workflow a bit. Maybe listen to changes to the file and use matplotlib to render the points (and tangents). This shouldn't take too long and I can help. It's probably also a bit easier then to render the tangents and cant vectors when we have them.

The other alternative is to use the full IfcConvert pipeline and write to something like OBJ. We already have --edge-arrows to document the direction of edges. We could somehow try to extend it with options for cant. And then you'd use something like Blender to inspect the output. Advantage is that you have a full UI so you can rotate the view and inspect coordinates of points.

Lastly you could also investigate defining some global functions that help you with debugging and that you can invoke from the debugger in the immediates window.

aothms:
> Could a setting by added to IfcGeom::IteratorSettings that indicates which geometry adjustment strategy to use?

I've completed my work on the settings refactoring https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/ifcgeom/ConversionSettings.h#L388 Took me a whole lot more work then I imagined, but I think the result is reasonable:

- the overflowing bitfield is resolved
- we have type safe settings (previously we only had bools on the iterator side and floats on the geometry side)
- we also have a direct integration with boost program options to read them from the command line, so no more duplication there

Downside is massive recompilations in case of setting changes... Settings should probably be made more modular so that recompilation is also more selective. Or maybe a bigger part could be moved to the translation unit instead of the header file (but how? maybe more virtual instead of templates and a vector instead of tuple?).

So feel free to add a setting. Define the struct and add it to the tuple is enough. Then get some coffee while recompiling almost the entire code base :(

RickBrice:
@aothms I wanted to make my first attempt at settings easy so I opted for controlling the evaluation of piecewise curves. I wanted two options: Minimum Step Size (like 0.5 m, which is what we are doing now), or Maximum Number of Steps so I could reduce the number of points for testing.

I created a struct with the struct PiecewiseDiscretization : public SettingBase<PiecewiseDiscretization,std::pair<bool,double>>;

This was a complete fail because the pair isn't going to work for defaultvalue. I'll just do it as two settings.

So my question, when in `ifcopenshell::geometry::taxonomy::piecewise_function::evaluate()` how do I get access to the settings so I can setup evaluation loop?

BTW - I had an extra large coffee :)

aothms:
> This was a complete fail because the pair isn't going to work for defaultvalue. I'll just do it as two settings.

It also needs to be istream deserializable, so that boost program options can read it from the command line. You could provide the template, but I agree, it's probably easier to just separate it out over two settings.

> So my question, when in ifcopenshell::geometry::taxonomy::piecewise_function::evaluate() how do I get access to the settings so I can setup evaluation loop?

Good point. I think we just have to pass it as a function param:

```diff
diff --git a/src/ifcgeom/AbstractKernel.cpp b/src/ifcgeom/AbstractKernel.cpp
index 09f0cda1a..a6dae693f 100644
--- a/src/ifcgeom/AbstractKernel.cpp
+++ b/src/ifcgeom/AbstractKernel.cpp
@@ -77,7 +77,7 @@ bool ifcopenshell::geometry::kernels::AbstractKernel::convert_impl(const taxonom
 }
 
 bool ifcopenshell::geometry::kernels::AbstractKernel::convert_impl(const taxonomy::piecewise_function::ptr item, IfcGeom::ConversionResults& cs) {
-	auto expl = item->evaluate();
+	auto expl = item->evaluate(settings_);
 	expl->instance = item->instance;
 	return convert(expl, cs);
 }
diff --git a/src/ifcgeom/taxonomy.cpp b/src/ifcgeom/taxonomy.cpp
index 432d2d2c7..5039a6d02 100644
--- a/src/ifcgeom/taxonomy.cpp
+++ b/src/ifcgeom/taxonomy.cpp
@@ -453,7 +453,7 @@ ifcopenshell::geometry::taxonomy::solid::ptr ifcopenshell::geometry::create_box(
 	return solid;
 }
 
-ifcopenshell::geometry::taxonomy::item::ptr ifcopenshell::geometry::taxonomy::piecewise_function::evaluate() const {
+ifcopenshell::geometry::taxonomy::item::ptr ifcopenshell::geometry::taxonomy::piecewise_function::evaluate(const Settings& s) const {
 	// @todo configure resolution
 	double length = 0.0;
 	for (auto& s : spans)
@@ -466,7 +466,7 @@ ifcopenshell::geometry::taxonomy::item::ptr ifcopenshell::geometry::taxonomy::pi
    auto resolution = length / num_steps;
 	for (int i = 0; i <= num_steps; ++i) {
 		auto u = resolution * i;
-      Eigen::Matrix4d m = evaluate(u);
+      Eigen::Matrix4d m = evaluate(s, u);
       polygon.push_back(taxonomy::make<taxonomy::point3>(m.col(3)(0), m.col(3)(1), m.col(3)(2)));
 	}
 
@@ -480,7 +480,7 @@ ifcopenshell::geometry::taxonomy::item::ptr ifcopenshell::geometry::taxonomy::pi
 	return polygon_from_points(polygon);
 }
 
-Eigen::Matrix4d ifcopenshell::geometry::taxonomy::piecewise_function::evaluate(double u) const {
+Eigen::Matrix4d ifcopenshell::geometry::taxonomy::piecewise_function::evaluate(const Settings& s, double u) const {
    // @todo: rb optimize, assume monotonic evaluation and store last evaluated segment?
    for (auto& [length, fn] : spans) {
    if (u < length + 0.001) { // @todo: rb - need to use consistent tolerance
diff --git a/src/ifcgeom/taxonomy.h b/src/ifcgeom/taxonomy.h
index 8d1358351..811d873cf 100644
--- a/src/ifcgeom/taxonomy.h
+++ b/src/ifcgeom/taxonomy.h
@@ -2,6 +2,7 @@
 #define TAXONOMY_H
 
 #include "../ifcparse/IfcBaseClass.h"
+#include "../ifcgeom/ConversionSettings.h"
 
 #include <boost/variant.hpp>
 #include <boost/functional/hash.hpp>
@@ -128,7 +129,7 @@ typedef item const* ptr;
 			struct implicit_item : public item {
 				DECLARE_PTR(implicit_item)
 
-				virtual item::ptr evaluate() const = 0;
+				virtual item::ptr evaluate(const Settings& s) const = 0;
 			};
 
 			struct piecewise_function : public implicit_item {
@@ -149,9 +150,9 @@ typedef item const* ptr;
 					return boost::hash<decltype(v)>{}(v);
 				}
 
-				virtual item::ptr evaluate() const;
+				virtual item::ptr evaluate(const Settings& s) const;
 
-				Eigen::Matrix4d evaluate(double u) const;
+				Eigen::Matrix4d evaluate(const Settings& s, double u) const;
 			};
 
 #ifdef TAXONOMY_USE_SHARED_PTR
```

PR #3967, opened 2023-11-04
bf: pass-attr-count-ok.ifc failed

I-Sokolov:
These tests are marked as passed but violates where-rules:
1) IfcPerson.IdentifiablePerson
2) IfcWallStandardCase.WR1 (needs material)
3) IfcFacilityPart is abstract


aothms:
Thanks Igor, much appreciated!

PR #3968, opened 2023-11-05
Added classification to selector

theoryshaw:


theoryshaw:
I tried to take a stab at 'setting' the values, but got lost. :\

PR #3969, opened 2023-11-07
Refactor duplicate aggregate

brunoperdigao:
I'm creating a PR as a draft because there are a few things that are worth discussing. Also, this PR depends on #3902 being merged first. The original functionality was developed as a highly experimental feature and had a few code duplications and naming inconsistencies. A few notes about the changes:
- There is some refactoring in variable name consistency, and I also propose to change the main operator's name to "DuplicateMoveLinkedAggregate" because it better describes what it does and also differentiates from simply duplicating an aggregate with "shift+d".
- The duplication process is now called directly from `OverrideDuplicateMove.execute_ifc_duplicate_operator`, which removes code duplication.
- Instead of using Psets, all the objects are linked by sharing an `IfcGroup`. That way, every `IfcElementAssembly` share the same group and that's what the "RefreshLinkedAggregate" will look for when updating. This also made the code much simpler.
- With this new approach, you can select any aggregate and use the refresh operator, and all the other aggregates will be updated accordingly.
- Also, every object in the aggregate shares a group with its corresponding duplicate. That opens up the possibility for the refresh operator to compare each object and only update the ones that changed, which would improve the efficiency of this operator. I still have to figure out how to do this comparison.

In case you want to test this with an existing file, [this script](https://gist.github.com/brunoperdigao/13f609173388c1abd810427484f957c1) will update the linked aggregates to this new system.

There are a few things to add and improve later (like creating a UI), which will be worked on in future PRs.

Thanks to @brunopostle who pointed me to the idea of using IfcGroup and @theoryshaw for being brave enough to use this feature and providing a great file to test.

blender_TunSNI81D9


brunopostle:
I haven't tested (there are merge conflicts), do you have an example IFC file that shows the data structure?

brunopostle:
..so duplicated assemblies are placed in a Group with Name 'Linked Aggregate', and a each item in the assembly is connected to its duplicate in the other assemblies by placing it in another Group (also called 'Linked Aggregate'), this seems reasonably robust:

image

I'm not sure if the spec allows it, but is there an equivalent assembly that uses IfcRelNests instead of IfcRelAggregates? this would preserve the ordering of the elements within the assembly, and there wouldn't be a need for the second level of Groups.

This would be a similar user experience to Sketchup where a 'component' is a special kind of 'group' that can be instanced.

theoryshaw:
Awesome! 

dumb question.  Should all the following be merged into one PR? Especially if one merges relies on the merging of a previous one? 

image

theoryshaw:
There's probably a way to for me to merge all these PRs into one, on my local end, but my head would explode trying to figure it out. :)

theoryshaw:
I 'think' i got it. :)  

image

theoryshaw:
I got this error when duplicating a subaggregate.  Would assume subaggregates could be recomposed into other aggregates. 

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/7c892101-a904-4491-9d3c-40d8386f6189

theoryshaw:
Perhaps in this instance, if a subaggregate is duplicated it pushes that aggregate to the top most level.. that is, it's no longer nested in anything aggregate.

theoryshaw:
also got this error. 


https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/a16c17cb-b9a9-4465-9097-46f88c4a95dd

theoryshaw:
Even when there's no change, and I `bim.refresh_linked_aggregate` an error is thrown...

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/9ee4d35a-58c0-4968-a5bf-b8cbb0b951ad

brunoperdigao:
> each item in the assembly is connected to its duplicate in the other assemblies by placing it in another Group (also called 'Linked Aggregate'), this seems reasonably robust

Yes, right now the only group necessary for it to work is the one with the IfcElementAssembly. The reason I created a group for each element was to allow for future object comparison. I'll take a look at IfcRelNests.

brunoperdigao:
@theoryshaw I think that it might be something with your merge. I'm not getting these errors. There might be some merge conflicts between PRs 3902 and 3939. I splitted between different PRs because I thought the others were easier to approve. Could you try using this PR alone?

>Perhaps in this instance, if a subaggregate is duplicated it pushes that aggregate to the top most level.. that is, it's no longer nested in any aggregate.

It should work like that:
GIF 14-11-2023 23-29-01

theoryshaw:
> I think that it might be something with your merge. 
ha... very likely. ;)

Will just try this PR.

theoryshaw:
Cool, it works now.... yes, was my borked merging. :)

theoryshaw:
would be nice if `bim.refresh_linked_aggregate` would work on multiple selected aggregates, at one time. 
Currently it only refreshes all the aggregates of only 1 selected aggregate.  

Since it take a long time to refresh aggregates, it nice to run many at one time.

theoryshaw:
@brunoperdigao whoops... i didn't think it would add this commit automatically when i pushed. I thought it would prompt me for a pull request.
Anyways, this is just a quick tweak to accommodate updating multiple aggregates at one time.

theoryshaw:
Also, since it takes some time to refresh larger aggregates, it would be nice to have a way to refresh only certain ones. 

Maybe if you select more than 1 instance of a certain type, only those selected will be refreshed. 

So if you have 100 instances of  certain type, and select only 3 of them, or these 3 will be undated.

brunoperdigao:
> From a power user's perspective this all seems good to me. Possible to merge @Moult?

Not so fast :)
This PR depends on another (3902). So, maybe you could test that too?
I'm starting to work on a basic UI that I think can be ready for this PR.
And I also have to write some tests.


> Also, since it takes some time to refresh larger aggregates, it would be nice to have a way to refresh only certain ones. 
Maybe if you select more than 1 instance of a certain type, only those selected will be refreshed. 
So if you have 100 instances of certain type, and select only 3 of them, only these 3 will be updated.
...
And if only 1 instance of the type was selected, all 100 would get refreshed.

I can put this on this PR too

As for this request (https://github.com/IfcOpenShell/IfcOpenShell/issues/3068#issuecomment-1789118697) I kinda disagree. I think it's useful to 'shift+d' objects and get them inside the aggregate. My plan is to add an operator to "break the link" so you could select any object, and it would remove it from the linked aggregate. I think it's easier to exclude an object than to include a new one (which would require going to the aggregates panel, click the plus button and select the parent). What do you think?

Moult:
This is awesome stuff happening feel free to merge whenever you think its appropriate :)

theoryshaw:
> Not so fast :)
I get so excited. ;)

> This PR depends on another (3902). So, maybe you could test that too?
Yes.  Would be possible to merge that one into this one?  Will make it easier on my side, to test.  Since my merging experiments before were less than successful. :)

> As for this request (https://github.com/IfcOpenShell/IfcOpenShell/issues/3068#issuecomment-1789118697) I kinda disagree. I think it's useful to 'shift+d' objects and get them inside the aggregate. My plan is to add an operator to "break the link" so you could select any object, and it would remove it from the linked aggregate. I think it's easier to exclude an object than to include a new one (which would require going to the aggregates panel, click the plus button and select the parent). What do you think?

Yes, this make sense. I agree. 

...

> Not so fast :)
Would only say, that I think it's better to release early and often.  There will alway be more tests and bells and whistle to apply.  If its in the main branch users can more easily tease out the kinks, and provide feedback.  A typical user is not used to pull requests, merging and rebasing, etc.

brunoperdigao:
@theoryshaw I merged the two PR's into this and also implemented the multiple selection option. Let me know if you can catch any bug. If it's working, feel free to merge into main. I'll leave the UI stuff to another PR

brunoperdigao:
I also removed the IfcGroup from every element, and added a pset with an "Index" property, that should work to identify linked elements inside the aggregate.

theoryshaw:
Awesome, on the merge. 

I'm getting the following error. It probably has something to do with removing the [ifcgroups](https://github.com/IfcOpenShell/IfcOpenShell/pull/3969#issuecomment-1819851568). 
If so, is there a suggested approach, or script to run to clean a file up?

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/d9f7a853-d684-4ff2-8217-9a289a82f129

brunoperdigao:
Yeah, you're right. Please use [this file](https://github.com/IfcOpenShell/IfcOpenShell/files/13430537/aggregate-demo3.ifc.txt) instead.

theoryshaw:
What if someone was stupid, like me and already updated the Highland haven project, with the ifcgroup approach? :)

brunoperdigao:
@theoryshaw this might work: https://gist.github.com/brunoperdigao/cc7c263db59ba03c83d8c1f126c64217

theoryshaw:
> @theoryshaw this might work: https://gist.github.com/brunoperdigao/cc7c263db59ba03c83d8c1f126c64217

awesome, so far, so good. 

...

Currently if you hide an object, and then refresh the aggregate, the object go missing in the 'children' aggregates.  I feel this could cause some confusion.  Would assume hiding an object has no bearing on the refresh.  

https://www.dropbox.com/scl/fi/meugo0yyharhskedqpzmg/2023-11-21_19-46-17_KeyShowView_Carnac.mp4?rlkey=19tltegfhhythq2624yhezy3u&dl=0

theoryshaw:
> and also implemented the multiple selection option. 

This is awesome. 

Would it also be possible to implement [this](https://github.com/IfcOpenShell/IfcOpenShell/pull/3969#issuecomment-1813723834) suggestion as well?

brunoperdigao:
@theoryshaw Just pushed a new commit. Now the user have 3 options:
- Select only one instance: it will refresh all the linked aggregate related.
- Select multiple instances of the same group: it will only refresh the selected linked aggregates, based on the active object.
- Select one instance of different groups: all the linked aggregate of each group will be refreshed.

If you select more than one instance from different groups, it won't work. If it's not clear, I can make a video later.

theoryshaw:
@brunoperdigao Perfect. This is awesome.

theoryshaw:
Merge yet? :)

brunoperdigao:
Fine by me

theoryshaw:
@brunoperdigao has my vote for push privileges in the future.  If anyone wants to make the executive decision, in granting them.

scoop96:
Is this already in blenderbim 240108 ? or should I self build it if I want to use it?

brunoperdigao:
> Is this already in blenderbim 240108 ? or should I self build it if I want to use it?

It doesn't have a UI right now, so it's a little hidden. Basically, if you have an aggregate, you can `crtl + shift + d` to create a linked copy. Then, if you modify one of the copies, you can select it and update all the other by pressing F3 and typing `refresh_linked_aggregates`.
Hopefully I'll have a basic UI soon, and I'm planning to write some documentation about it.

scoop96:
Hi, sorry for spam, the script is not included in blenderBIM 24 as in F3 neither operators, so far two options I think?

Manually launch, or manually install the script  xD

brunoperdigao:
> Hi, sorry for spam, the script is not included in blenderBIM 24 as in F3 neither operators, so far two options I think?
> 
> Manually launch, or manually install the script xD

No problem. Do you have the option "Developer Extras" selected in Blender Preferences? I forgot to mention that.
image

PR #2219, opened 2022-05-31
docs: update Doxyfile, remove duplicate docs

dirkolbrich:
- specify ./output/doxygen as main path for output
- use relative paths for subfolder output

This allows to build the docs as local html output with minor adjustments to the Doxyfile.

PR #3979, opened 2023-11-08
bbim docs: update version info, reorder image folders, add subsections for user interface

dirkolbrich:
This is a start to add more in depth explanations to the BlenderBIM documentation as a written supplement to the numerous YouTube videos (especially by [@IfcArchitect](https://www.youtube.com/@IfcArchitect)).

For this I also would like to restructure the current grouping of the docs topics into a Quick start section, a more hands-on Workflow section and a classic documentation for the different functionality of BlenderBIM.

Getting started
  - Introduction to BIM
  - Installation
  - Exploring an IFC model
  - Creating an IFC model
  - Georeferencing
  - Dealing with large models
  - Tracking revisions with Git
  - Other add-ons

Workflows
  - Creating a wall and a wall type
  - Creating a window
  - ...

Tips & Tricks <-- (would this section be helpful?)

General usage
  - Interface
  - Modelling
    - Walls
    - Windows
  - Types
  - ...
   
Developers:
  - Installation
  - Hello, world!
  - ...

What do you think?


dirkolbrich:
And while we are on it, how about moving the BlenderBIM docs to the more general `/docs/blenderbim` folder under `root` as mentioned in this PR https://github.com/IfcOpenShell/IfcOpenShell/pull/2219#issuecomment-1737174892?

Moult:
Awesome work! Thanks so much!

> how about moving the BlenderBIM docs to the more general /docs/blenderbim folder 

I'm cool with that but I'll let @aothms make the call.

aothms:
> I'm cool with that but I'll let @aothms make the call.

Ok, from my side :)

PR #3982, opened 2023-11-09
Implements IfcSegmentedReferenceCurve and first cut at cant geometry

RickBrice:
There is still a lot to do, but this is a start. I'll keep working through the details, testing, and adding support for more IfcCurve subtypes.

### DIFF
--- a/src/ifcgeom/mapping/IfcSegmentedReferenceCurve.cpp
+++ b/src/ifcgeom/mapping/IfcSegmentedReferenceCurve.cpp
@@ -1,0 +1,84 @@
+/********************************************************************************
+ *                                                                              *
+ * This file is part of IfcOpenShell.                                           *
+ *                                                                              *
+ * IfcOpenShell is free software: you can redistribute it and/or modify         *
+ * it under the terms of the Lesser GNU General Public License as published by  *
+ * the Free Software Foundation, either version 3.0 of the License, or          *
+ * (at your option) any later version.                                          *
+ *                                                                              *
+ * IfcOpenShell is distributed in the hope that it will be useful,              *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 *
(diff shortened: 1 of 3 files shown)


src/ifcgeom/mapping/IfcSegmentedReferenceCurve.cpp
```diff
@@ -0,0 +1,84 @@
...
+		auto c = cant->evaluate(u);
+      c.col(3)(0) = 0;
+      std::swap(c.col(3)(1),c.col(3)(2));
+      Eigen::Matrix4d m;
+      m = xyz * c;
+      return m;
+	};
+
+	std::array<taxonomy::piecewise_function::ptr, 2> both = { gradient , cant };
+	// @todo: rb - this constrains the range of u to the minimum of gradient and cant
```
aothms, line 63: Maybe also a setting? I'll work on that part.

aothms (commented):
Conceptually I think the code is very nice. It's elegant how you were able to mix in this functionality by means of the adjuster.

Unfortunately I can't help with the open questions. Hopefully the IF meetings continue to be enlightening.

src/ifcgeom/mapping/IfcSegmentedReferenceCurve.cpp
RickBrice: I don't think a setting is the solution here. The same issue exists with IfcGradientCurve where the horizontal and vertical curves could differ in length. We discussed earlier that maybe u-values beyond the end of a curve would yield NAN results. In that case we want to set the span to the maximum length instead of minimum length.

The question of cases with cant and gradient curves being longer or shorter and beginning before or after the composite curve has been posted to the IF miro board. Perhaps it is best to wait for that discussion to happen before changing the current approach.

aothms: Ok, while I agree that NaNs are maybe not so useful to the end user, but yielding 0s for those parts, so no super elevation, or no elevation just coincident with the horizontal, I imagined as potentially useful.

PR #3998, opened 2023-11-11
Update reference_structure.py

VDobranov:
A mistype in docs: should be IfcRelReferencedInSpatialStructure instead of IfcRelContainedInSpatialStructure


aothms:
Thanks 👍

PR #4023, opened 2023-11-17
v0.8.0 Alignment Geometry - Bug fixes and more work in progress

RickBrice:


RickBrice:
This is still a work in progress. I struggle with cant and IfcSegmentedReferenceCurve, but Peter is nice enough to help me through the IF. Hopefully my pea sized brain will understand soon. :)

An interesting new addition in this PR is an example program that demonstrated how to build the business logic and geometry of a basic alignment in C++.

Using my Debug build of IfcConvert, I ran 

>  F:\IfcOpenShell\_build-vs2022-x64\Debug> .\IfcConvert.exe F:\IfcOpenShell\_build-vs2022-x64\examples\FHWA_Bridge_Geometry_Alignment_Example.ifc

and was hoping to get a .obj file to open in Blender. There was no .obj file output and the .mtl file is size 0 kb.

@aothms  Any ideas? I want to view the alignment data graphically - I don't really care how, I just want some assurance that everything I've done thus far works up the toolchain.

RickBrice:
@aothms @civilx64 I have hit a significant milestone in the alignment work. There is now support in the C++ geometry mapping for IfcCompositeCurve, IfcGradientCurve, IfcSegmentedReferenceCurve (sort of) with IfcLine, IfcCircle, and IfcClothoid for parent curves. IfcLinearPlacement is supported with IfcPointByDistanceExpression. This should cover a large percentage of practical cases. There is also support for IfcOffsetCurveByDistance.

What remains are the following:

- Testing, testing and more testing
- Improving the implementation of IfcSegmentedReferenceCurve
- Adding support for more parent curve types (IfcEllipse, the other spiral types, ???)
- Getting IfcOffsetCurveByDistance to work with IfcPolyline and IfcIndexPolyCurve as basis curves
- Generated geometry for IfcSectionedSolidHorizontal
- Creating business logic alignment from geometry, and visa-versa (maybe?)

If you have access to models with alignments, please send them my way for testing.

aothms:
> @aothms Any ideas? I want to view the alignment data graphically - I don't really care how, I just want some assurance that everything I've done thus far works up the toolchain.

Not sure whether this is still relevant.

To be frank this turned out a bigger puzzle then I had hoped. Stepping through the code is painful (compile time loops) and a lot of the datatypes are near opaque to the debugger (boost variant, occt). But I did find a couple of things.

The first item is something still very much broken in IfcConvert v0.8, but I can begin fixing now: the selection of whether to include elements other then volumetric objects and which representation contexts to include. We have the --plan option which can be specified in addition to --model to indicate you'd want to include "linestrip" (e.g alignment axis) geometry in the output. In v0.8 it's currently not possible to have both linework as well as volumetric objects. This used to be a bit of a fundamental thing because we haven't really properly decided how to map things like IfcProduct in the taxonomy. But it looks like we can circumvent that issue because we actually consider pairs of (Product.ObjectPlacement, Product.Representation[i]) now in the iterator. Some remnants in the code to hardcode and filter on body are now eliminated.

Then, getting to the actual evaluation of the geometry. I didn't have that model you referred to (the name looked familiar, but I couldn't find it). So I used linear-placeemnt-of-signal.ifc from the spec. There seem to be two cases:

a)

```
[Notice] [2023-11-21 14:23:34] {0t59UGHs9DHxT45UBjGnHV} Evaluated points: point3 0 5 0
point3 0 5 4.4098e-17
point3 0 5 8.8196e-17
point3 0 5 1.32294e-16
point3 0 5 1.76392e-16

[Error] [2023-11-21 14:23:34] {0t59UGHs9DHxT45UBjGnHV} Getting a triangulation element from model failed.
[Notice] [2023-11-21 14:23:34] {0t59UGHs9DHxT45UBjGnHV} No geometrical elements found or none succesfully converted
```

Small edges, OpenCASCADE cannot deal with this. So it gives the triangulation element failed error. It might actually work with `--kernel cgal`.

b)

Another evaluation returns all nans:

[Notice] [2023-11-21 14:27:48] {0o7VCJvavDdA9xYwz5oBRK} Evaluated points: point3 -nan(ind) -nan(ind) -nan(ind)
point3 -nan(ind) -nan(ind) -nan(ind)
point3 -nan(ind) -nan(ind) -nan(ind)
point3 -nan(ind) -nan(ind) -nan(ind)
point3 -nan(ind) -nan(ind) -nan(ind)

[Notice] [2023-11-21 14:31:42] {0o7VCJvavDdA9xYwz5oBRK} Adjusted edge end-point with distance -nan on:
[Notice] [2023-11-21 14:31:43] {0o7VCJvavDdA9xYwz5oBRK} No geometrical elements found or none succesfully converted

If you want to run convert only a specific element you can add `--include attribute GlobalID 0o7VCJvavDdA9xYwz5oBRK` to the **end** of your command line.

Run with `-yvv` to get all log messages and to not prompt to overwrite the output file

So the full command line (minus executable) I used is:

    linear-placement-of-signal.ifc -yvv linear-placement-of-signal.obj --plan --include attribute GlobalId 0o7VCJvavDdA9xYwz5oBRK

See 87f39b382 and b52c5ab08

I'm currently traveling for a workshop. Will likely not be able to continue this before Monday. Sorry.

aothms:
> I have hit a significant milestone in the alignment work

Congratulations! to the entire community actually.

RickBrice:
Thanks. I'll try your recommendations. This is the US Thanksgiving holiday so I won't be getting much done until Monday either.

RickBrice:
FHWA_Bridge_Geometry_Alignment_Example.ifc is generated by the new IfcAlignment example program in the examples folder.

civilx64:
Congrats and well done! 

 
> What remains are the following:
> 
> 
>     * Creating business logic alignment from geometry, and visa-versa (maybe?)

[EnrichIFC4x3](https://github.com/bSI-RailwayRoom/IFC-Rail-Unit-Test-Reference-Code/tree/master/EnrichIFC4x3) will create geometry from business logic, but not the other way around.  If nothing else there may be some inspiration for spiral implementation, e.g. [spiral.cpp](https://github.com/bSI-RailwayRoom/IFC-Rail-Unit-Test-Reference-Code/blob/master/EnrichIFC4x3/EnrichIFC4x3/business2geometry/spiral.cpp).

From the standpoint of IFCOS my thought is that some capability for business logic -> representation would be required to calculate the points along the alignment curves.  I don't see a compelling case for representation -> business logic other than maybe assigning it to an intern so that they can earn their stripes.

> If you have access to models with alignments, please send them my way for testing.

You may be aware of these already:

- [IFC-Rail-Sample-Files](https://github.com/bSI-RailwayRoom/IFC-Rail-Sample-Files)

I was working specifically with the AWC samples when troubleshooting the initial python implementation work.

I have real-world examples from industry we could potentially look at, but they are horizontal & vertical only and do not include any spirals.  Maybe that makes them not particularly compelling as part of a unit test suite.

PR #4032, opened 2023-11-21
[IfcConvert] IfcBoundingBox conversion exclusion

### LINKED ISSUE #2044: Exclusion of BoundingBoxes in IfcConvert
Hi,
I have a problem with IfcConvert 0.7.0 (OCC 7.5.3) on some models.

If an IfcObject has an IfcShapeRepresentation of type "BoundingBox" (in addition to another representation, of type "SweptSolid" for example): both representations will be taken into account in the result of the conversion.  The bounding box will completely hide the geometry of the resulting object.

I'm not sure if this is a bug (because both representations are present in these IFC files) but opening them on different BIM software gives me a result without the BoundingBox representation.

Currently, the only effective solution I could put in place is to remove the problematic representations before launching a conversion 

```
for bbox in [
        rep
        for rep in ifcwrap.ifc.by_type("IfcShapeRepresentation")
        if rep.RepresentationType == "BoundingBox"
    ]:

        ifc_file.remove(bbox)
    ifc_file.write(ifc_path)

```

Is there a "cleaner" way to ignore some IfcProducts (by filtering on their Representation.Representations) ?

christt105:
Hi, I am not sure if this is the best place to put the code. Also there is missing the command line to be able to convert the IfcBoundingBox, but I don't know how to retrieve the settings from the IfcRepresentation class.

The change will fix #2044 and all the duplicates.

### DIFF
--- a/src/ifcgeom/IfcRepresentation.cpp
+++ b/src/ifcgeom/IfcRepresentation.cpp
@@ -27,6 +27,12 @@ bool IfcGeom::Kernel::convert(const IfcSchema::IfcRepresentation* l, IfcRepresen
 	if ( items->size() ) {
 		for ( IfcSchema::IfcRepresentationItem::list::it it = items->begin(); it != items->end(); ++ it ) {
 			IfcSchema::IfcRepresentationItem* representation_item = *it;
+
+			if(/*TODO: settings.get(CONVERT_IFC_BOUNDING_BOXES) && */representation_item->data().type() == &IfcSchema::IfcBoundingBox::Class()) {
+				part_succes |= true;
+				continue;
+			}
+
 			if ( shape_type(representation_item) == ST_SHAPELIST ) {
 				part_succes |= convert_shapes(*it, shapes);
 			} else {


aothms:
@christt105 many thanks for this, after some deliberation I feel it's better to *not* handle this in the shape handling. Setting part_success to true is clever in that it doesn't result in a lot of error messages, but would end up keeping empty results (as the representation is supposedly processed successful) for the representation polluting the geometric output.

I've been prototyping an approach in IfcConvert.cpp to just delete them from the file, but haven't been able to find the time to finish up.

```C++
diff --git a/src/ifcconvert/IfcConvert.cpp b/src/ifcconvert/IfcConvert.cpp
index f73b13792..94ef23c29 100644
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -155,6 +155,7 @@ static std::basic_stringstream<path_t::value_type> log_stream;
 void write_log(bool);
 void fix_quantities(IfcParse::IfcFile&, bool, bool, bool);
 std::string format_duration(time_t start, time_t end);
+void remove_boundingboxes(IfcParse::IfcFile& f);

 /// @todo make the filters non-global
 IfcGeom::entity_filter entity_filter; // Entity filter is used always by default.
@@ -347,6 +348,9 @@ int main(int argc, char** argv) {
                        "Disables computation of normals. Saves time and file size and is useful "
                        "in instances where you're going to recompute normals for the exported "
                        "model in other modelling application in any case.")
+               ("keep-bounding-boxes",
+                       "Default is to removes IfcBoundingBox from model prior to converting geometry."
+                       "Setting this option disables that behaviour")
                ("deflection-tolerance", po::value<double>(&deflection_tolerance)->default_value(1e-3),
                        "Sets the deflection tolerance of the mesher, 1e-3 by default if not specified.")
                ("force-space-transparency", po::value<double>(&force_space_transparency),
@@ -1009,6 +1013,10 @@ int main(int argc, char** argv) {
         Logger::Notice(msg.str());
     }

+       if (!vmap.count("keep-bounding-boxes")) {
+               remove_boundingboxes(*ifc_file);
+       }
+
        IfcGeom::Iterator context_iterator(settings, ifc_file, filter_funcs, num_threads);

 #ifdef WITH_HDF5
@@ -1512,6 +1520,56 @@ namespace latebound_access {
        }
 }

+void remove_boundingboxes(IfcParse::IfcFile& f) {
+       auto delete_reversed = [&f](const aggregate_of_instance::ptr& insts) {
+               if (!insts) {
+                       return;
+               }
+               // Lists are traversed back to front as the list may be mutated when
+               // instances are removed from the grouping by type.
+               for (auto it = insts->end() - 1; it >= insts->begin(); --it) {
+                       IfcUtil::IfcBaseClass* const inst = *it;
+                       f.removeEntity(inst);
+               }
+       };
+
+       auto boxes = f.instances_by_type("IfcBoundingBox");
+       std::set<IfcUtil::IfcBaseEntity*> reps_to_delete;
+
+       if (boxes) {
+               for (auto& b : *boxes) {
+                       auto reps = f.getInverse(
+                               b->data().id(),
+                               f.schema()->declaration_by_name("IfcRepresentation"),
+                               -1
+                       );
+                       if (reps) {
+                               for (auto& r_ : *reps) {
+                                       auto r = r_->as<IfcUtil::IfcBaseEntity>();
+                                       auto items_val = r->get("Items");
+                                       bool to_remove = true;
+                                       if (!items_val->isNull()) {
+                                               try {
+                                                       aggregate_of_instance::ptr items = *items_val;
+                                                       to_remove = items->size() == 0;
+                                               } catch (IfcParse::IfcException&) {
+                                               }
+                                       }
+                                       if (to_remove) {
+                                               reps_to_delete.insert(r);
+                                       }
+                               }
+                       }
+               }
+       }
+
+       delete_reversed(boxes);
+
+       for (auto& r : reps_to_delete) {
+               f.removeEntity(r);
+       }
+}
+
 void fix_quantities(IfcParse::IfcFile& f, bool no_progress, bool quiet, bool stderr_progress) {
        {
                auto delete_reversed = [&f](const aggregate_of_instance::ptr& insts) {
```

Moult:
Isn't a simpler solution to loop through geometric contexts with an identifier of "Box" and exclude that from the list of considered contexts?

Moult:
Bump @aothms :)

aothms:
Hi, thanks for this. As discussed I went with another solution that was a little bit more involved to write, but at least allowed to keep all logic self contained in IfcConvert.cpp without touching the somewhat broken settings module. Hope you stay involved in IfcOpenShell development and look forward to any future PRs.

PR #4120, opened 2023-12-15
fix cmake regex for used blender python version

ay-ex:
@Moult : Upstream Blender repo now uses `set` instead of `SET` for declaring their used Python version - 
the proposed extended regex covers both cases and should fix the currently broken choco pipeline. 🙂


Andrej730:
that will work, thank you!)

Issue #4116, opened 2023-12-14
Potential compatibility issues with hdf5=1.14.2 and/or boost 1.82


Krande:
Hey,

when testing the latest conda releases for IfcOpenShell I accidentally came across issues with converting units using the IfcPatch `ConvertLengthUnit` (in my case from millimeter to meter).

After doing some digging I found the root cause being a change in dependencies between conda releases and not anything to do with the IfcPatch itself. 

From what I can tell the issue appears when moving to libboost 1.82 (from 1.78) and hdf5 1.14.2 (from 1.12.2).

I can't think the `hdf5` library is being used for this, so this might be a `boost` issue?

@aothms Can you think of any reason why this might be happening? If you can point me in the right direction, I could try to help :)

Here is the full dependency list:
[ifcva10_0.txt (working)](https://github.com/IfcOpenShell/IfcOpenShell/files/13674369/ifcva10_0.txt)
[ifcva10_1.txt (no working)](https://github.com/IfcOpenShell/IfcOpenShell/files/13674371/ifcva10_1.txt)

To recreate the issue here's what I did.

(note that the tests I am referring to now are done on windows, but I have tested this for linux and osx in a github actions, and it seems to produce similar results).

First I download the [beam-standard-case.ifc](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_1/FINAL/HTML/annex/annex-e/beam-standard-case.ifc).

Then I make a copy of the IfcPatch `ConvertLengthUnit.py` and call it `convert_length_unit_patch.py` and slightly modify it (see the full code below) to run a simple test that prints all IfcCartesianPoints as tuples from my terminal and place it next to my ifc file.

I started by testing the latest conda-forge release `v0.7.0.231127`

Open terminal and do

```
mamba create -n test1 numpy ifcopenshell=v0.7.0.231127
mamba activate test1
python convert_length_unit_patch.py
```

This prints a list of all the IfcCartesianPoints, where you'll notice that some points are converted to meters (ie. x,y,z coords position magnitudes are typically < 10) while most of the coordinates are still in millimeters (x,y,z coords > 1000)

I did this also for the conda daily build (from today) and the same issue appears. 

However, If I use an older version of the conda IfcOpenShell (`mamba create -n ifctest2 ifcopenshell=*0.7.0a6*`) it will print consistently only small values -> meters only. 

[...]

PR #4125, opened 2023-12-16
Proposal for updating IfcPatch ConvertLengthUnit.py for compatibility with boost>=1.82

Krande:
Based on the reported issue in #4116, here is a suggested change for the IfcPatch `ConvertLengthUnit.py` that ensures the intended behaviour is maintained despite changes in the boost>=1.82.

I also moved the core functionality from the patch itself to ifcopenshell-python -> `ifcopenshell.util.unit.convert_file_units(ifc_file, target_unit) -> ifcopenshell.file` so that users of the python library could also enjoy this functionality :)

Let me know what you think?

- closes: #4116


src/ifcopenshell-python/ifcopenshell/util/unit.py
```diff
@@ -624,3 +622,83 @@ def format_length(
...
+            cur_decl = cur_decl.declared_type()
+            if hasattr(cur_decl, "name") is False:
+                continue
+            if cur_decl.name() == "IfcLengthMeasure":
+                return cur_decl
+
+    return None
+
+
+def convert_file_units(ifc_file: ifcopenshell.file, target_units: str) -> ifcopenshell.file:
```
aothms, line 652: This probably applies to other APIs as well, but should we copy over the header values?

src/ifcopenshell-python/ifcopenshell/util/unit.py
```diff
@@ -624,3 +622,83 @@ def format_length(
...
+    prefix = "MILLI" if target_units == "MILLIMETERS" else None
+
+    file_patched = ifcopenshell.api.run("project.create_file", version=ifc_file.schema)
+    if ifc_file.schema == "IFC2X3":
+        user = file_patched.add(ifc_file.by_type("IfcProject")[0].OwnerHistory.OwningUser)
+        application = file_patched.add(ifc_file.by_type("IfcProject")[0].OwnerHistory.OwningApplication)
+        old_get_user = ifcopenshell.api.owner.settings.get_user
+        old_get_application = ifcopenshell.api.owner.settings.get_application
+        ifcopenshell.api.owner.settings.get_user = lambda ifc: user
+        ifcopenshell.api.owner.settings.get_application = lambda ifc: application
```
aothms, line 663: Is this still needed?

src/ifcopenshell-python/ifcopenshell/util/unit.py
```diff
@@ -624,3 +622,83 @@ def format_length(
...
+
+    # Copy all elements from the original file to the patched file
+    for el in ifc_file:
+        file_patched.add(el)
+
+    unit_assignment = ifcopenshell.util.unit.get_unit_assignment(file_patched)
+
+    old_length = [u for u in unit_assignment.Units if getattr(u, "UnitType", None) == "LENGTHUNIT"][0]
+    new_length = ifcopenshell.api.run("unit.add_si_unit", file_patched, unit_type="LENGTHUNIT", prefix=prefix)
```
aothms, line 672: It probably would be a small effort to generalize this beyond length unit, we can save that for later when there is a demand, but maybe make it more explicit in the name of the function.

src/ifcopenshell-python/ifcopenshell/util/unit.py
Krande: I just naively copied the IFC2x3 stuff from the patch not really knowing its intended purpose. 

I am fine with removing it

src/ifcopenshell-python/ifcopenshell/util/unit.py
aothms: Also we can only convert to SI, fine limitation as is, but maybe document.

src/ifcopenshell-python/ifcopenshell/util/unit.py
```diff
@@ -624,3 +622,83 @@ def format_length(
...
+                continue
+            if isinstance(val, tuple):
+                new_el = [ifcopenshell.util.unit.convert_unit(v, old_length, new_length) for v in val]
+                setattr(element, attr.name(), tuple(new_el))
+            else:
+                new_el = ifcopenshell.util.unit.convert_unit(val, old_length, new_length)
+                # set the new value
+                setattr(element, attr.name(), new_el)
+
+    file_patched.remove(old_length)
```
aothms, line 697: Does it make more sense to detect this in the loop at line 666 (𝅘𝅥𝅮  number of the beast 𝅘𝅥𝅮) and simply don't add it in the first place?

src/ifcopenshell-python/ifcopenshell/util/unit.py
Krande: Yeah, I think so too. For now I updated the name to make its current purpose more clear in https://github.com/IfcOpenShell/IfcOpenShell/pull/4125/commits/c9552fac29034dae1bec6faa9dcebe446d6b5a5c.

src/ifcopenshell-python/ifcopenshell/util/unit.py
Krande: Well, we could (for the musical reference alone). However, I kind of liked the simplicity of the "move all elements to new ifc, and do an in-place edit of the new file". 

But I don't really have a strong opinion on this, so please feel free to decide what you think is best :)

src/ifcopenshell-python/ifcopenshell/util/unit.py
Krande: Ah, that never occured to me. Do we have any code examples for copying the header values somewhere else?

src/ifcopenshell-python/ifcopenshell/util/unit.py
Krande: I just removed it in https://github.com/IfcOpenShell/IfcOpenShell/pull/4125/commits/f5500da0af1ad79a3350bd304bf7b03365b065ea

If anyone wants the IFC2x3 functionality we can add it to the patch instead. Or suggest using the conversion utility to convert from 2x3 to a more recent version :)

Krande:
@aothms We should probably include a test for the units conversion. I don't have time to add one today, but I might have some time to add one tomorrow :)

src/ifcopenshell-python/ifcopenshell/util/unit.py
Krande: In https://github.com/IfcOpenShell/IfcOpenShell/pull/4125/commits/9335c8a0c8c9cbc2fc9056cec79dd0bdd10c7350 I changed the "copy" method to a more direct copy of the entire file instead of going element by element.

`file_patched = ifcopenshell.file.from_string(ifc_file.wrapped_data.to_string())`

Which I believe will include header information

Krande:
@aothms I fixed the failing ci.yml and removed ci-py-only-yml (as it was identical to ci.yml just without the IfcConvert tests).

Now all we need to fix the failing tests

aothms:
Great 🎉 yeah that ci-py-only was there because at some point IfOpSh didn't build if IfcConvert wasn't configured to be built in CMake, but that probably was a one of event that didn't need a full ci script to run till eternity.

PR #4135, opened 2023-12-19
Linked aggregate preserve names

brunoperdigao:
This PR implements:
1. Custom incremental renaming for Linked Aggregates. The Element Assemblies will be rename as "_01" instead of blender default renaming which is ".001". This will help to preserve the name when the file is reloaded, as requested here by @theoryshaw: https://community.osarch.org/discussion/comment/17435/#Comment_17435.
2. Preserve object's original name when using the "Refresh Linked Aggregates".
3. A few tests.

There has been a small change on how the "BBIM_Linked_Aggregate" Index is implemented, so for files that already have Linked Aggregates, please use this script: https://gist.github.com/brunoperdigao/6b0f0d255a9d4eebd8faf9b8f778329d


theoryshaw:
Seems to work great. 
@Moult, @Andrej730 mind if we merge this one?

PR #4145, opened 2023-12-25
Added .clangd to .gitignore

dushyant-basson:
git-ignoring .clangd


aothms:
We already have this commit https://github.com/IfcOpenShell/IfcOpenShell/commit/34f038530ff0abab4e11ccae58e7f6df181bb39b

PR #4152, opened 2023-12-29
Added more comparisons in search

c4rlosdias:
Added more comparisons in selector.py and in BlenderBim's UI (filters):
Captura de tela 2023-12-28 214913
Added comparisons:

equal to : =
not equal to: !=
less than: <
greater than: >
less than or equal to: <=
greather than or equal to: >=


Moult:
Thanks so much!

PR #4154, opened 2023-12-30
add geometry processing with multiple representations example

sennhvi:
Illustrates an usage  example of  the third parameter of the create_shape method

### DIFF
--- a/src/ifcopenshell-python/docs/ifcopenshell-python/geometry_processing.rst
+++ b/src/ifcopenshell-python/docs/ifcopenshell-python/geometry_processing.rst
@@ -170,7 +170,7 @@ The element in your ifc file might look like this.
     #1618938=IFCSHAPEREPRESENTATION(#4,'Body','BRep',(#1617583));
     #1618939=IFCSHAPEREPRESENTATION(#4,'Body','BRep',(#1617630));
     #1618957=IFCPRODUCTDEFINITIONSHAPE($,$,(#1618937,#1618938,#1618939));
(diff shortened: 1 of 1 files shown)


src/ifcopenshell-python/docs/ifcopenshell-python/geometry_processing.rst
```diff
@@ -157,6 +157,42 @@ In these scenarios, a ``geometry`` is returned directly, equivalent to
     # Process a profile
     geometry = geom.create_shape(settings, ifc_file.by_type("IfcProfileDef")[0])
 
+
+Process individual element with multiple shape representations
+---------------------
+
+When a element contains multiple shape representations, you can use third parameter of ``create_shape()`` to process each shape representation.
```
aothms, line 164:
```suggestion
When an element contains multiple shape representations with the same identifier or when you want more explicit control over which representation is processed (e.g `Body` or `Tesselation`), you can use the third parameter of ``create_shape()`` to nominate a specific shape representation to be processed in the context of a product.
```

aothms:
Super, thanks! I've added a quick sentence on another case to use this function. Btw, I understand you problem in the issue better now. This kind of multiple Body representation was actually discussed here https://forums.buildingsmart.org/t/multiple-body-representations-for-a-single-ifcproductrepresentation/4405/10 It seems most people tend to agree multiple Body representations should be disallowed. A representation conveys a particular view (Body, Axis, ...) of the product, not a part (for that use Representation.Items or decomposition into building element parts).

sennhvi:
Using multiple body representation may not be practical, but I often encounter this when exporting ifc files using some software

PR #4160, opened 2024-01-02
Update geometry_processing.rst

sennhvi:
Add an usage example of the parameter include  for the Geometry Iterator


Moult:
Thanks so much!

PR #4169, opened 2024-01-03
fix surfacestyle loader bug

c-mellueh:
List can't add Tuple next line, so the oneliner also needs to output a Tuple


aothms:
@c-mellueh thanks for this, this looks sane to me (although @Andrej730 should do the review), but by deleting your forked repo we don't have access to the code anymore and can't merge this in. This appears to be a recent Github thing. Do you still have a local copy you can push again?

c-mellueh:
@aothms 
i allready recomitted the PR ([#4171](https://github.com/IfcOpenShell/IfcOpenShell/pull/4171))

aothms:
Oh sorry, thanks!

Issue #44, opened 2016-03-12
Cherry pick a subset of recent commits onto the 0.4.0 branch


aothms:
This in order to do a 0.4.0-final release.

Build scripts need to be merged as well as they will be used for the build.

Moult:
Can this be closed? :)

aothms:
God yes. It's a bit of an issue though that we never really do final releases of the branches.

Issue #30, opened 2016-02-15
Missing colors for IfcMappedItem-geometry


larsmoa:
I have a case where colors are missing for some geometry. I have not been able to pinpoint the issue, but I have been able to create a minimal example file that reproduces the issue. The file can be opened in e.g. Solibri Model Viewer and here the color is correct. IfcConvert outputs the default gray material - even though there is a keyword IFCCOLOUR which should have been picked up by IfcConvert.

I created an illustration of how the material is associated with the geometry in the failing example, and how it is associated in a working file, see [colormapping.pdf](https://github.com/IfcOpenShell/IfcOpenShell/files/130799/colormapping.pdf).

Note! The file is an exempt from a full model. I cannot share this file here, but if it will help I can send it to you by email.

[GreenOnlyColour.ifc.zip]
(https://github.com/IfcOpenShell/IfcOpenShell/files/130790/GreenOnlyColour.ifc.zip)

aothms:
Hi Lars, thanks for your detailed report. Glancing over the code [1], it does indeed seem this is omitted. More commonly styles would be applied somewhere in the `MappingSource.MappedRepresentation` of the `IfcMappedItem`, but indeed this is an oversight. I will update you on the progress. Would you object to the referenced file being added to the test files [2] to prevent regressions?

[1] https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcgeom/IfcGeomShapes.cpp#L509
[2] https://github.com/IfcOpenShell/IfcOpenShell/tree/master/test/input

larsmoa:
Thank you for the quick reply! Feel free to add the test file to the tests :)

aothms:
Thanks again. This was a trivial fix. Let me know if you encounter other issues.

larsmoa:
Wow - thanks, that was really fast! I’m happy to say that this improves color handling a lot. Before this file was all grey:


Again, thanks!

PR #4170, opened 2024-01-03
Skip IfcRelAggregates entities with missing RelatingObject

c-mellueh:
when importing a specific ifc file I was getting this error:
```bash
  File "C:\Users\xxx\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\blenderbim\bim\module\project\operator.py", line 734, in execute
    ifc_importer.execute()
  File "C:\Users\xxx\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\blenderbim\bim\import_ifc.py", line 240, in execute
    self.create_collections()
  File "C:\Users\xxx\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\blenderbim\bim\import_ifc.py", line 1515, in create_collections
    self.create_aggregate_and_nest_collections()
  File "C:\Users\xxx\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\blenderbim\bim\import_ifc.py", line 1565, in create_aggregate_and_nest_collections
    rel_aggregates = [
  File "C:\Users\xxx\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\blenderbim\bim\import_ifc.py", line 1568, in <listcomp>
    if r.RelatingObject.is_a("IfcElement") or r.RelatingObject.is_a("IfcElementType")
```
and the import stopped.

After further investigation i found out, that my ifc file is a bit faulty because some IfcRelAggregates entities are missing their RelatingObject.

Now the importer throws an **error message and skips** the corresponding entity instead of breaking entirely.

### DIFF
--- a/src/blenderbim/blenderbim/bim/import_ifc.py
+++ b/src/blenderbim/blenderbim/bim/import_ifc.py
@@ -1562,16 +1562,24 @@ class IfcImporter:
                 elif element.Nests:
                     rel_aggregates.add(element.Nests[0])
         else:
-            rel_aggregates = [
-                r
-                for r in self.file.by_type("IfcRelAggregates")
-                if r.RelatingObject.is_a("IfcElement") or r.RelatingObject.is_a("IfcElementType")
-            ] + [
-                r
-                for r in self.file.by_type("IfcRelNests")
-                if (r.RelatingObject.is_a("IfcElement") or r.RelatingObject.is_a("IfcElementType"))
-                and [e for e in r.RelatedObjects if not e.is_a("IfcPort")]
-            ]
+            def check_relation(r) -> bool:
+                return bool(r.RelatingObject.is_a("IfcElement") or r.RelatingObject.is_a("IfcElementType"))
+
+            rel_aggregates = list()
+
+            for relation in self.file.by_type("IfcRelAggregates"):
+                if relation.RelatingObject is None:
+                    print(f"Skipping '{relation.GlobalId}' because 'RelatingObject' needs to exist.")
+                    continue
+                if check_relation(relation):
+                    rel_aggregates.append(rel_aggregates)
+
+            for relation in self.file.by_type("IfcRelNests"):
+                if relation.RelatingObject is None:
+                    print(f"Skipping '{relation.GlobalId}' because 'RelatingObject' needs to exist.")
+                    continue
+                if check_relation(relation) and [e for e in relation.RelatedObjects if not e.is_a("IfcPort")]:
+                    rel_aggregates.append(rel_aggregates)
 
         if len(rel_aggregates) > 10000:
             # More than 10,000 collections makes Blender unhappy
--- a/src/blenderbim/blenderbim/bim/import_ifc.py
+++ b/src/blenderbim/blenderbim/bim/import_ifc.py
@@ -1572,14 +1572,13 @@ class IfcImporter:
                     print(f"Skipping '{relation.GlobalId}' because 'RelatingObject' needs to exist.")
                     continue
                 if check_relation(relation):
-                    rel_aggregates.append(rel_aggregates)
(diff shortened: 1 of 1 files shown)


Andrej730:
Hi! Thank you for the suggestion but `RelatingObject` for both [IfcRelAggregated](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcRelAggregates.htm) and [IfcRelNests](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcRelNests.htm) is not optional. So if it's missing it means that IFC is invalid and could produce any other unpredictable errors and skipping them may hide actual causes that produced invalid IFC in the first place.

image

Was the invalid .ifc file created using BlenderBIM or some other software? We handle this kind of exceptions in case we know that some other software is persistently producing invalid IFC in some way.

c-mellueh:
Hi the file is from [Desite MD Pro](https://thinkproject.com/products/desite-bim/)
As far as i understand this can allways happen in Desite because you can create empty Sites. If they were exported, they automatically create an IfcSite and IfcRelAggregates afterwards without checking if there are entities that use the relation

Andrej730:
> As far as i understand this can allways happen in Desite because you can create empty Sites. If they were exported, they automatically create an IfcSite and IfcRelAggregates afterwards without checking if there are entities that use the relation

But that sounds like a different case - since wouldn't that mean RelatingObject would be set to IfcSite and not to None?

c-mellueh:
@Andrej730 
yeah, you're right i didn't look to much into it, as far as i see it now, its the aggregation to the project that is missing (#&#8203;45):
image
idk why this is happening. 
But for me an error message and skipping of the aggregation is better than just breaking the import without explaining to the user.

Andrej730:
> yeah, you're right i didn't look to much into it, as far as i see it now, its the aggregation to the project that is missing (#​45):

I'm not familar with Desite MD Pro - any ideas what can be special about IfcSite `#44`? `#30` seems to be the same but has correct IfcRelAggregates to it.

> But for me an error message and skipping of the aggregation is better than just breaking the import without explaining to the user.

Agreed that it is better to inform the user but the problem is that skipping these errors can produce other unpredictable behaviour later on.

c-mellueh:
> I'm not familar with Desite MD Pro - any ideas what can be special about IfcSite #44?

Yes, i think Desites internal datastructure is the problem. Desite splits the IFC-File into a "Building Structure" and a "Project Structure":
image
They both resembles the aggregation structure in the IFC-Spec but are somewhat different to it. There seems to be a linkage problem between the TypeBsSite (Building Structure) IfcSite (Project Structure) and Desite gets confused. 
I'll contact the Desite support and hope they'll fix this bug.

Moult:
Sorry I know this is probably not what you want to hear but I don't think this issue should be caught / skipped. There are some IFC viewers out there which do skip this type of stuff but I've found through experience that it really does mask the problem and users get confused as to where the problem truly is. Given that BBIM is a native IFC editor it's quite important to make sure that we're as strict and valid as possible.

It's not possible to check for every single schema invalidity - that's what the validate feature is for (available in the debug panel, or now on https://validate.buildingsmart.org ) - in particular relaggregates is a very fundamental concept in IFC so at least for now in alpha I'm happy for the software to crash and burn. It might also be good to start introducing catchalls to improve the user experience like "your model didn't load, step 1) run this validator, step 2) file a bug, etc".

c-mellueh:
@Moult no problem i understand your reasoning. For me it's an easy fix. I hope Desite will correct there exporter so that people that aren't that deep into understanding the IFC-Specification and their restrictions can continue to work with their exports.

PR #4171, opened 2024-01-03
Fix Surface Style Loader Bug

c-mellueh:
List can't add Tuple next line, so the oneliner also needs to output a Tuple


Andrej730:
Nice catch, thank you! 👍

Issue #4184, opened 2024-01-09
BsDD baseURL


c-mellueh:
Hi,
why are we using "https://bs-dd-api-prototype.azurewebsites.net/" as baseurl and not "https://api.bsdd.buildingsmart.org/api/"?

aothms:
Old version. Should be updated I guess. Feel free to send a PR, but not entirely sure if the API is 100% the same on those end points.

c-mellueh:
ok. I checked the endpoints, they're not the same. I'll rework the bsDD and add the new calls + deprecation warnings to existing functions.

PR #4193, opened 2024-01-10
update BsDD to match new API

c-mellueh:
+ implements new api calls from #4188 
+ move Functions from operator to core/tool

see #4184

### DIFF
--- a/src/blenderbim/blenderbim/core/bsdd.py
+++ b/src/blenderbim/blenderbim/core/bsdd.py
@@ -1,0 +1,40 @@
+def load_bsdd(props, client, bsdd):
+    props.domains.clear()
+    if props.load_preview_domains:
+        dictionaries = bsdd.get_dictionaries(client)
+    else:
+        dictionaries = bsdd.get_dictionaries(client, "Active")
+    if not props.load_preview_domains:
+        dictionaries = filter(lambda d: d["status"] == "Active", dictionaries)
+    for dictionary in sorted(dictionaries, key=lambda d: d["name"]):
+        new = props.domains.add()
+        bsdd.fill_dictionary_prop(new, dictionary)
+
+
+def set_active_bsdd_dictionary(context, name, uri):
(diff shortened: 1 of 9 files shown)


src/blenderbim/blenderbim/core/bsdd.py
```diff
@@ -0,0 +1,40 @@
+def load_bsdd(props, client, bsdd):
+    props.domains.clear()
```
Moult, line 2: A minor note that things like this shouldn't be in core. E.g. `props.domains.clear()` references a concrete implementation of the Blender API. Same with stuff like `props.domains.add()` below. The core should be completely pure Python with no concretions.

c-mellueh: ah ok, so this should happen in tools/bsdd.py?
is there anything that shouldn't be passed to the tools Ccasses or to core and should instead happen in the operator function?

c-mellueh: is something like this allowed in core:
```python
props = context.scene.BIMBSDDProperties
```
or should the context be passed to the tool file?

Moult: Core should only call tool functions, nothing else. It should read like English and be a short poem of code that describes what is going on at a high level, something like:

```python
dictionaries = bsdd.get_dictionaries()
bsdd.import_dictionaries(dictionaries)
```

or

```python
if bsdd.should_load_previews():
    dictionaries = bsdd.get_all_dictionaries()
else:
    dictionaries = bsdd.get_active_dictionaries()
bsdd.import_dictionaries(dictionaries)
```

or whatever - it's up to you, but the important point is that it's 1) not concrete 2) reads like English. This is a very simple operator, so it doesn't matter too much, but as complexity grows, having a nice high level poem to refer to can be very important.

Forcing it to be abstract ensures that all concrete interactions with Blender or IFC happens in a tool. These can be isolated, tested, and reused.

PR #4201, opened 2024-01-11
Clean up core/bsdd.py

c-mellueh:
see discussion in  [#4193](https://github.com/IfcOpenShell/IfcOpenShell/pull/4193#discussion_r1448163915)
move function calls from core to tool and sort functions alphabetically


Moult:
Nice!

Issue #3434, opened 2023-07-13
Impl of IfcTriangulatedFaceSet does not take into account PnIndex attribute


aothms:
An additional indirection of indices (I doubt it's used regularly).

![afbeelding](https://github.com/IfcOpenShell/IfcOpenShell/assets/1096535/25cabaf6-386c-4628-96b1-22f9358d8e08)

PR #4251, opened 2024-01-25
Add docs export excel - read excel 

chuongmep:
Add Documentation totorial export - read metadata excel from IFC

### DIFF
--- a/src/ifcopenshell-python/docs/ifcexcel.rst
+++ b/src/ifcopenshell-python/docs/ifcexcel.rst
@@ -1,0 +1,88 @@
+IfcExcel
+======
+
+IfcExcel lets you view and edit IFC data using spreadsheets or tabular datasets, and review metadata easily.
+
+Library installation
+-------------------
+
+To install the library, you need some python library dependencies :
+
+```bash
+pip install ifcopenshell -U
+pip install openpyxl -U
+pip install pandas -U
+pip install lark -U
+```
+
+Export Excel
+-------------------
+
+Import library
+
+```python
+import ifcopenshell
+import pandas as pd
+import warnings
+warnings.filterwarnings('ignore')
+```
+
+Open IFC file
+
+```python
+file_path = r"model.ifc"
+ifc_file = ifcopenshell.open(file_path)
+```
+
+Get all Classes
+
+```python
+# get all classes
+classes = ifc_file.by_type("IfcProduct")
+# print all class name
+class_names = [class_name.is_a() for class_name in classes]
+class_names = list(set(class_names))
+class_names.sort()
+```
+
+Export to Excel
+
+```python
+file_name = "result.xlsx"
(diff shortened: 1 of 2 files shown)


chuongmep:
Hi @Moult, I reviewed csv, it is unconvinion to review a large model. When we export to excel, all parameter and class easier to review, analyst. This is why I'm want add more to support for user. 
Regard .

Moult:
I don't understand, what can this do that IfcCSV cannot do?

chuongmep:
@Moult example :  some value of parameter include character '',

Moult:
@chuongmep you can already do that, IfcCSV supports regex: https://blenderbim.org/docs-python/ifcopenshell-python/selector_syntax.html#getting-element-values

chuongmep:
Hi @Moult , thank you for your help !I can't deny the usefulness of csv and other formats, but I'm sure it depends on the use case as to how easily the data can be shared to the end user.
Thank you.

Moult:
IfcCSV is just the name - it supports reading/writing ODS, XLSX, and Pandas Dataframes, not just CSV by the way.

chuongmep:
Hi @Moult , I will take look into that. thank you. If have any issue I will reopen request.

chuongmep:
how you can export all data like my file excel @Moult  ?

Moult:
Right now a single workbook only contains one sheet. That sounds like a nice feature to add. Would you like to help build it?

Right now IfcCSV also requires you specify each column instead of automatic expansion of possible properties. See #3434. Would you like to help build it?

Moult:
I just realised it isn't really documented anywhere, so FYI the current general features of IfcCSV are:

 1. Custom filtering of elements to show on your spreadsheet
 2. Custom columns selecting data from attributes, psets, classifications, materials, locations, systems, and other relationships
 3. Querying commands to select Nth relationship, or count number of relationships
 4. Custom formatting like custom column titles, formatting rules like imperial number formatting for units, rounding numbers to nearest X, etc
 5. Grouping rows with grouping rules e.g. count totals, get average, sum each row, etc
 6. Sorting rules including multi column sort
 7. Custom values to show null, true/false, or empty strings
 8. Preserve formatting of pre-formatted spreadsheets
 9. CSV/XLS/ODS/PD support
 10. Both export and import data (to then edit the IFC based on the changes in the spreadsheet)
 11. Summary rows such as grand totals
 12. GUI integration in Blender
 13. Integration with SVG table generation so you can place schedules on drawings / sheets

chuongmep:
I don't think can do mutiple dataframs with csv format, may be we should seperate to new class ? 
I will colaborate when I have free time.

Moult:
Indeed, for CSV the behaviour would split out to multiple CSV files e.g. with a suffix. I'm closing this PR and tracking this in #4252.

chuongmep:
For sure, I will following with that part. Thank you.

PR #4263, opened 2024-01-28
Try improve clone opening

maxfb87:


theoryshaw:
Very cool!
It doesn't seem to work, however, on my side.  The walls don't seem to regen correctly after moving the window. 


https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/6ab3edbd-967d-49da-8c49-cb59a75f966f

maxfb87:
@theoryshaw i actually don't know why it doesn't work at your side... i tried on my side another time with the same steps as yours and it works...
I don't know what to say, maybe merge the PR and after that retry on both two sides...

theoryshaw:
This is weird, it worked 'once', as you can see.  Then it didn't work after that. 

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/a1fd63b9-82ae-442d-a85f-225ebd811c6d

maxfb87:
@theoryshaw i tried another time and it works also multiple times at my side...

theoryshaw:
hmm... what's your OS?
I'm using Windows. 
What do you think might be the cause of the difference?

theoryshaw:
Nevermind, i think i got it working. 

I don't remember doing this before, but maybe I tried to regen the window, and not the wall, since i noticed if you try to regen the window first, it doesn't allow you to regen the 2nd wall. 

...

I guess that would be a suggestion, to have this work when regening the window too. 

...

Also, is it possible if you regen the 1st wall, that the 2nd wall regens automatically?

maxfb87:
@theoryshaw good to know!
Yeah maybe it was the cause because maybe if you regen the window it creates a new placement so the cloned opening is not linked with the other opening anymore...

Atm the cloning feature is wip function, so i guess it's like a proof of concept, but i'll see if i can improve the feature like you have suggested...thanks for the feedback :-)

PR #4271, opened 2024-01-30
UI  for linked aggregate

brunoperdigao:
This PR creates a UI for linked aggregates, as a sub-panel of the aggregate panel. The UI shows:
- The number of instances of linked aggregates
- Operator to select all linked aggregates (This operator is part of this PR).
- Operator to select refresh all linked aggregates (This operator was already implemented).
- Operator to break the link to other aggregates (This operator is part of this PR).

There is a small addition to `Unassign Object From Aggregation` to also remove the "BBIM_Linked_Aggregate" Pset.

image


brunoperdigao:
@theoryshaw It would be great if you could test this out.

@Andrej730 I'd like your help, if possible. When I click the operator to remove the link, it doesn't instantly reload the information. I have to click another object and then come back to it to see the result in the UI. Is there a way I can make this better?

Andrej730:
@brunoperdigao sure! 
Long story short - for IFC operators (operators inheriting from the class below) need to use `_execute` method instead of `execute`. Example - https://github.com/IfcOpenShell/IfcOpenShell/blob/3d7e8d7c6c4502defce0b472ba6a99e649647ab2/src/blenderbim/blenderbim/bim/module/aggregate/operator.py#L44-L47

IFC Operator class:
```python
class Operator:
    def execute(self, context):
        blenderbim.bim.handler.refresh_ui_data()
        return {"FINISHED"}
```

The reason why UI didn't updated after operator was executed - `blenderbim.bim.handler.refresh_ui_data()` wasn't executed (it's the method that refreshes all `data.py` modules) and `AggregateData.data` kept the old data. In case if operator wasn't making changes to IFC and wouldn't be inheritting from IFC Operator then it would be fine to force update with `blenderbim.bim.handler.refresh_ui_data()` directly but IFC operators do this automatically.

Using `_execute` in case if IFC data changes is important not just for UI updates but since `IfcStore.execute_ifc_operator` part of IFC Operator helps to make sure undo/redo on IFC will work correctly.

brunoperdigao:
@Andrej730 Thanks for the explanation! It's working now

theoryshaw:
very cool... This seems to work great for me!

theoryshaw:
One comment, i guess. :)
Maybe when running `bim.break_link_to_other_aggregates` it removes the aggregate empty, and associated collection as well.

theoryshaw:
> One comment, i guess. :) Maybe when running `bim.break_link_to_other_aggregates` it removes the aggregate empty, and associated collection as well.

On 2nd thought, i won't do this... as most likely the workflow will be changing it into another aggregate link.

brunoperdigao:
@theoryshaw mind testing this too? https://github.com/IfcOpenShell/IfcOpenShell/pull/4135

When we get these two PRs merged, I consider the Linked Aggregates to be with it's basic functionality implemented.

brunoperdigao:
@theoryshaw The problem was that #4135 had conflicts with the base branch. I fixed that, it should be fine now. Let me know if it works.

theoryshaw:
@brunoperdigao seems to work great on my side!

theoryshaw:
@Moult, @Andrej730 mind if we merge this one?

Issue #4278, opened 2024-02-01
Upgrades to IfcClash to be more fit for purpose


Moult:
IfcClash has never progressed much beyond "it shows things which touch and if you're lucky it'll run" in my opinion. There's also a bunch of related open bugs:

 - #2430 (the most critical bug)
 - #3189 (trip for devs)
 - #3444 (also critical)
 - #4069 (usability)
 - #4128 (packaging has always been troublesome)
 - #4238 (packaging has always been troublesome)
 - #3938

I'd like to work on IfcClash prior to the next release cycle (March) to get it to a state that it can actually be practically used as an open source solution for clash detection. There are quite a few facets about this, starting with getting correct clash results. The types of clashes to detect are:

 - **Protrusions** - a shape is partially inside another shape
 - **Envelopes** - a shape is fully enclosed by another shape
 - **Piercings** - a shape completely skewers another shape (entering, and leaving through the other side)
 - **Touching** - a shape touches another shape, but does not protrude
 - **Encroaching** - a shape is within a specified clearance distance of another shape
 - **Duplication** - a shape is identical to another shape and overlapping in the same spot

The protrusion distance is defined the innermost vertex of shape A enveloped by another shape B, as measured to the nearest face of shape B. *

Note * (naturally there are a lot of details and "except when" to this)

A **Protruding** or **Enveloping** shape will have a negative protrusion distance. A **Touching** shape will have an protrusion distance of 0. An **Encroaching** shape will have a positive protrusion distance.

In the case of **Piercing**, there may be no vertex of shape A inside shape B. In this case, the protrusion distance is given as the maximum protrusion distance possible.

The maximum protrusion distance may be approximated as the minimum OBB dimension / 2.

Non-manifold shapes (trees? ground?) do not have a logical penetration distance. They need to be dealt with somehow.

The data to return for each clash shall be:

 - Shape A
 - Shape B
 - Protrusion distance
 - Coordinate(s?) of clash
 - Type of clash (from above)

Users should be able to define:

 - A set A of elements to be clashed against a set B of elements. (In contrast, right now during the broadphase IfcClash loops through set A elements one by one instead of batch).
 - Worst protrusion distance (to aid in early returns, we can terminate early once a clash is found that satisfies this)
 - A clashing mode, chosen from:
   - **Intersection** which detects when shapes intersect one another
   - **Clearance** which detects when shapes encroach past a buffer distance
[...]

Issue #4279, opened 2024-02-01
Improve viewing of large federated models inside Blender


Moult:
Blender is poor at showing scenes with many objects. This can be partially overcome through multiple blends and linking, but still falls short on massive projects. There was an effort to overcome this through object merging strategies. The advanced mode allows uses to choose a merge mode which merges objects by type, class, or material. This drops the object count drastically at the expense of much longer first-load times (merging objects in Blender is slow!) and complete loss of BIM metadata since there is no longer a 1:1 Blender Object <-> IfcElement map.

Solving the many objects issue in Blender is not happening super soon. Ideally there is a dedicated "viewing" app, but at the same time it's super nice to have as much as possible inside Blender so you can have portions of the model as "inactive" and portions as active.

If this is solved, the BlenderBIM Add-on can be used for much more model coordination tasks on huge projects where full model context is important.

So I'm thinking of building features that:

 - Allow loading models as pure streams of triangles and colours (no materials / styles) that can be split into N objects per X triangles or similar. This is purely for viewing, not for editing. The geom iterator needs to be modified for this since this is very slow on the Python side.
 - Allow saving out the geom tree (upgraded to a BVH tree) alongside the blend to be read later.
 - Instead of using Blender objects to break things down, we can raycast the pointer click (or 3D cursor at first), fetch object(s) from the (earlier saved) tree and let users see the properties of it case by case in the properties menu.
 - Allow saving out pure dumb properties (not fetched live from the IFC in memory which is slow) similar to the preprocessing done in https://github.com/IfcOpenShell/xeokit-metadata-ios - it's more than likely good enough for coordination tasks.
 - Doing some sort of viewport culling that changes objects to bounding box view when not in view. This significantly speeds up the Blender viewport.
 - Ideally automate the saving and loading of .blend files as right now it's a pain to manage them alongside IFCs.

marwiss:
It seems like this is a feature of blender.

I was wondering if there could be a feature request about this on https://projects.blender.org/blender/blender/issues.

I am thinking if there could be a "fast view render mode" in blender.

Moult:
[...]

PR #4282, opened 2024-02-01
Features in geometry tree tailored for clash detection. See #4278.

Moult:
Definitely do not merge :)

First attempt at implementation of intersection clash checks. See #4278.

This considers only protruding clashes. It does not consider touching nor encroaching clashes and does not consider protrusion direction in the protrusion distance.

Still a lot to move around and tweak so it's also not yet ready for a detailed review and you can see lots of random prints and probably horrifically incorrect usage of pointers and structs and what not but at least something is committed to a branch for now ...


Moult:
I think this is now ready for a discussion (but not for merging) :)

You'll notice this PR is mostly lines added, not removed because I didn't want to touch existing functionality.

Things not yet implemented:

 - Many-many filtered clash sets
 - Self clashing
 - Considering piercing clashes in "intersection" clash mode
 - Nicer returns rather than independent protrusion_distances etc
 - Clash grouping
 - Duplication clashing
 - Box check for clearance distance
 - No-return-early for clearance distance option
 - Contained within for collision checks
 - Multithreading!

aothms:
I see a couple of follow up tasks:

- move physx code to separate file (.cpp) maybe just use the whole lib as a proper dependency
- separate product geometries into solids/shells so that we can rely on bvh-based point containment test instead of non-deterministic'ish ray hit counting
- most unordered maps should probably be vectors because the key is simply a contiguous range of ints
- factor out the common bits of the clash functions
- I think the distrinction between add/add_triangulated shouldn't be made at insertion time but rather at tree initialization.
- filter out invalid tris before tree insertion instead of maintaining a map of bools
- as you said replace `triangleintersects.hpp` with physx equivalent
- tag open shells so that we know we can't rely on/do containment tests
- I would probably store the points in a vector and index into them for every triangle, as opposed to storing all three points duplicated, to save some mem

I don't think I'm going to touch any of the geometric predicates. It sounds rather thought through. In some cases I'd have different preferences (like I conceptually don't really like how the distinction between piercing and protruding happens based on loose edges/verts as opposed to topology, but don't see a quick way out of resolving that without a more proper triangle mesh datastructure)

I can work on these follow up tasks if you want, but then we need to agree on some "handover time" so that we're loosing time on conflicts.

Moult:
> filter out invalid tris before tree insertion instead of maintaining a map of bools

I'm not sure how to do this. The BVH tree is created from a triangle_set, and a triangle_set is created from a shape list, and a shape list is a list of TopoDS_Face, not triangles. I'm guessing somehow under the hood it reuses the triangulation data from incremental mesh, but I don't know how to get at the triangles until after the BVH is actually created.

Moult:
A small benchmark regarding the vertex indices indirection on a 40MB data set clashing all 6400 elements against each other using a intersection check (without early returns). I'm measuring this portion of the code:

```python
    for element in sorted(all_els, key=lambda e: e.Name):
        clashes = tree.clash_intersection(element, tolerance=0.002, check_all=True)
```

Test | Mem (excl 682MB for file.open()) | Time
--- | --- | ---
Before indirection | 1,473MB | 10.2s
After indirection | 1,112MB | 12.7s

Moult:
OK I think I've done what I can. The remaining unchecked boxes I either don't know how to do properly or probably need a bigger architectural decision (also probably worth thinking what signatures to use for many-many clashes). I'll be pens down for the rest of today if you want to write code :)

Moult:
Current benchmarks :) Notice the growing majority of time in opening, building the tree, and the crazy high memory use.

All times in seconds. Memory in MB. All elements are clashed against all other elements.

 - Open = time to ifcopenshell.open(...)
 - Tree = time to create tree, either (B) baseline or (T) with triangulation / elem BVH
 - Col = collision, allowing touching
 - Int = intersection, 2mm tolerance with (R) return early or (A) check all
 - Clr = clearance check of 100mm
 - Sel = tree.select(e), which is functionally equivalent to Col or Int (if you need protrusion distances - except that Sel distances aren't correct atm)
 - Mem = memory, where (B) represents a "baseline" of only ifcopenshell.open() and creating the current box-based UBTree, and (A) represents the peak memory usage after everything including triangulated tree building and clashing. Measured using `print(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1000)`

Note that tree.select(e, extend=x) is not measured (equivalent to Clr) because it's simply too slow (gave up measuring after 5 minutes).

Dataset | # Objs | Open | Tree(B) | Tree(T) | Col | Int(R) | Int(A) | Clr(R) | Sel | Mem(B) | Mem(A)
--- | --- | --- | --- |  --- | --- | --- | --- | --- | --- | --- | ---
40MB of Elec / Fire | 6,412 | 1.65 | 1.85 | 3.3 | 1.6 | 2.4 | 4 | 8.5 | 9.5 | 995 | 1,760
200MB or Elec / Fire / Hyd / Mech | 18,513 | 8.8 | 34.21 | 39.6 | 8.5 | 14.6 | 21 | 110 | 130 | 8,410 | 10,820

Moult:
The latest commit introduces many-many variants of the clash functions: `clash_intersection_many`, `clash_collision_many`, and `clash_clearance_many`. I don't think there's any reason to keep the old 1:N clash functions (e.g. `clash_intersection`) because the N:N versions are always faster... like significantly so. See the edit I've made to the benchmarks post to see the huge improvement in numbers :) (especially when the same objects are in both set A and B)

Moult:
I had a shot at implementing multithreading. I attempted to try to do something similar to https://github.com/IfcOpenShell/IfcOpenShell/blob/c66cf0a22b1b7f008cdef70888d5cadb5de93da5/src/ifcgeom/IfcGeomIteratorImplementation.h#L326-L362 but failed horribly. I ended up using a mutex with a very naive approach:

 1. After the box BVH clash, create a task queue of clashes
 2. Divide that queue by num_threads
 3. Use threads to do the work, use a mutex to lock, and merge the results into a final results vector to return

I've got no idea if there is a better way to do it, but I've just updated the results table again and I'm very impressed with the results. I think I've run out of tricks I can think of to further optimise the clash portion of the code. I reckon it's now time to move on to optimising opening / tree creation.

Moult:
Here is some messy Python code to help test - it reads from a dir of IFC models so its easy to increase or decrease the dataset:

```python
import os
import time
import resource
import multiprocessing
import ifcopenshell
import ifcopenshell.geom
import ifcopenshell.util.representation

def run():
    total_els = 0
    total_filesize = 0
    total_open_time = 0
    total_tree_time = 0
    sstart = time.time()
    start = time.time()
    tree = ifcopenshell.geom.tree()

    models = {}
    all_els = []

    model_dir = "/home/dion/Projects/llrepos/cqr/clashtest/"

    # whitelist = ["-EL-", "-FS-", "-ME-", "-HY-", "-AR-"]
    whitelist = ["-EL-", "-FS-"]
    # whitelist = ["-EL-", "-FS-", "-ME-", "-HY-"]
    for fn in os.listdir(model_dir):
        if not fn.endswith(".ifc"):
            continue
        is_ok = False
        for wl in whitelist:
            if wl in fn:
                is_ok = True
                break
        if is_ok:
            fn = os.path.join(model_dir, fn)
            total_filesize += os.path.getsize(fn) / 1024 ** 2
            print(fn)
            start_open = time.time()
            models[fn] = ifcopenshell.open(fn)
            total_open_time += time.time() - start_open
            start_tree = time.time()
            model = models[fn]
            # settings.set(settings.USE_WORLD_COORDS, True)
            settings.set_context_ids([ifcopenshell.util.representation.get_context(model, "Model", "Body", "MODEL_VIEW").id()])
            els = set(model.by_type("IfcElement"))
            els -= set(model.by_type("IfcFeatureElement"))
            els = list(els)
            all_els.extend(els)
            iterator = ifcopenshell.geom.iterator(settings, model, multiprocessing.cpu_count(), include=els)
            if iterator.initialize():
                while True:
                    # tree.add_element(iterator.get_native(), should_triangulate=False)
                    tree.add_element(iterator.get_native(), should_triangulate=True)
                    if not iterator.next():
                        break
            total_tree_time += time.time() - start_tree
            # break

    print('results')
    print('#' * 100)

    start = time.time()
    results = {}

    # MANY
    # clashes = tree.clash_collision_many(all_els, all_els, allow_touching=True)
    # print(len(clashes))

    clashes = tree.clash_intersection_many(all_els, all_els, tolerance=0.002, check_all=True)
    print(len(clashes))

    # clashes = tree.clash_clearance_many(all_els, all_els, clearance=0.1, check_all=False)
    # print(len(clashes))

    # for clash in clashes:
    #     print('start')
    #     print(clash.distance)
    #     print(clash.clash_type) # garbage data here
    #     print(clash.a) # segfault here
    #     print(clash.b)
    #     print(list(clash.p1))
    #     print(list(clash.p2))

    # # SELECT
    # for element in sorted(all_els, key=lambda e: e.Name):
    #     continue
    #     # clashes = tree.select(element, extend=0.1)
    #     clashes = tree.select(element)

    print("Filesize", total_filesize)
    print("Elements clashed", len(all_els))
    print("Total results", len(results.keys()))
    print('Open time', total_open_time)
    print('Tree time', total_tree_time)
    print('Clash time', time.time() - start)
    print('Total', time.time() - sstart)
    
run()
print(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1000)
```

Moult:
I briefly investigated the current HDF5 implementation and I think it's currently storing all vertices of all elements without any instancing. I think this is the current biggest cause of filesize bloat for the HDF5 files. From tests it ends up that the H5 file is larger than the IFC. Yikes! (I've already temporarily disabled the storage of the brep and everything apart from the verts). Is there a reason for this or have I misunderstood something?

Is it too expensive to store instanced (based on shape.geometry.id) geometry and have the HDF5 loader apply the transformation matrix?

Moult:
To start testing the viability of H5, I've written a small script to dump just triangles and colours into H5. A few of the big moves:

 - No materials, only colours.
 - If an object has no material or only one material, we don't store material_ids
 - Shapes are stored separately and referenced per element
 - Element data is stored as a H5 dataset, not a group (groups are heavy filesize-wise apparently due to metadata overhead)
 - Max gzip compression
 - Variable dtype selection based on quantity of data and precision requirements
 - GlobalIds stored as bits

You can visit https://myhdf5.hdfgroup.org/ and check out this sample: [test3.h5.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14258112/test3.h5.txt)

There don't seem to be too many knobs to adjust to optimise things further, but the results are pretty good! The filesize seems to be half the IFC, it loads in a fraction of the time (I wrote a test Blender importer),and the memory usage is tiny (which is the biggest showstopper towards users loading in large models). I suspect most of the loading time is still in the Blender object creation, so there is work to be done there (and potentially with parallel processing to prepare chunks of triangles it can be very fast).

IFC Filesize | H5 Filesize | IFC Load Time | H5 Load Time | IFC Mem | H5 Mem
--- | --- | --- | --- | --- | ---
10.7 | 4.5 | 1.35 | 0.27 | 441 | 39
59.7 | 19.1 | 7 | 2.05 | 1493 | 99

2024-02-13-193404_1506x960_scrot

Sample script to convert IFC to H5:

```python
import os
import uuid
import h5py
import time
import numpy as np
import multiprocessing
import ifcopenshell
import ifcopenshell.geom
import ifcopenshell.util.representation

def run():
    fn = "/home/dion/drive/ifcs/asdf.ifc"
    f5 = "/home/dion/test3.h5"
    h5 = h5py.File(f5, "w")
    shapes = h5.create_group("shapes")
    shape_groups = {}
    colours = {}
    element_shape_ids = []
    global_ids = []
    names = []
    matrices = []

    compression_opts = 9

    total_filesize = 0
    total_open_time = 0

    all_els = []

    # fn = "/home/dion/test.ifc"
    total_filesize += os.path.getsize(fn) / 1024 ** 2
    print(fn)
    start_open = time.time()
    model = ifcopenshell.open(fn)
    total_open_time += time.time() - start_open
    settings.set_context_ids([ifcopenshell.util.representation.get_context(model, "Model", "Body", "MODEL_VIEW").id()])
    els = set(model.by_type("IfcElement"))
    els |= set(model.by_type("IfcSite"))
    els -= set(model.by_type("IfcFeatureElement"))
    els = list(els)
    all_els.extend(els)
    iterator = ifcopenshell.geom.iterator(settings, model, multiprocessing.cpu_count(), include=els)
    if iterator.initialize():
        while True:
            shape = iterator.get()
            m = shape.transformation.matrix.data
            faces = shape.geometry.faces
            edges = shape.geometry.edges
            verts = shape.geometry.verts
            materials = shape.geometry.materials
            material_ids = shape.geometry.material_ids
            # print(faces)

            geom_id = shape.geometry.id

            group = shape_groups.get(geom_id, None)
            if not group:
                group = shapes.create_group(str(len(list(shape_groups.keys()))))
                # group = shapes.create_group(geom_id)
                shape_groups[geom_id] = group
                total_verts = len(verts) / 3
                if total_verts < 2**8:
                    face_dtype = "uint8"
                elif total_verts < 2**16:
                    face_dtype = "uint16"
                else:
                    face_dtype = "uint32"
                group.create_dataset("faces", data=faces, dtype=face_dtype, chunks=len(faces), compression="gzip", compression_opts=compression_opts)
                max_verts = max(verts)
                if (max_verts < 1):
                    dtype = "float16"
                else:
                    dtype = "float32"
                group.create_dataset("verts", data=verts, dtype=dtype, chunks=len(verts), compression="gzip", compression_opts=compression_opts)

                material_keys = []
                for material in materials:
                    alpha = 1.0
                    if material.has_transparency and material.transparency > 0:
                        alpha = 1.0 - material.transparency
                    diffuse = material.diffuse + (alpha,)
                    material_name = f"{diffuse[0]}-{diffuse[1]}-{diffuse[2]}-{diffuse[3]}"
                    try:
                        material_index = list(colours.keys()).index(material_name)
                    except:
                        material_index = len(list(colours.keys()))
                        colours[material_name] = diffuse
                    material_keys.append(material_index)
                    key = {}

                materials = np.array(material_keys, dtype='uint8') 
                group.create_dataset("materials", data=materials, dtype="uint8", compression="gzip", compression_opts=compression_opts)
                if len(material_keys) > 1:
                    group.create_dataset("material_ids", data=material_ids, dtype="uint8", chunks=len(material_ids), compression="gzip", compression_opts=compression_opts)
                # group.create_dataset("materials", data=materials, dtype="float32", compression="gzip", compression_opts=compression_opts)

            global_ids.append(uuid.UUID(ifcopenshell.guid.expand(iterator.get().guid)))
            names.append(model.by_id(iterator.get().id).Name or "")
            element_shape_ids.append(list(shape_groups.keys()).index(geom_id))
            matrices.append(m)

            if not iterator.next():
                break

    uuids_array = np.array([np.frombuffer(u.bytes, dtype=np.uint8) for u in global_ids])
    h5.create_dataset("element_global_ids", data=uuids_array, dtype=np.uint8)

    data = np.array(names, dtype=object)
    h5.create_dataset("element_names", data=data, dtype=h5py.special_dtype(vlen=str))

    total_shapes = len(list(shape_groups.keys()))
    if total_shapes < 2**8:
        dtype = "uint8"
    elif total_shapes < 2**16:
        dtype = "uint16"
    elif total_shapes < 2**32:
        dtype = "uint32"
    h5.create_dataset("element_shape_ids", data=element_shape_ids, dtype=dtype, compression="gzip", compression_opts=compression_opts)
    h5.create_dataset("element_matrices", data=matrices, dtype="float32", compression="gzip", compression_opts=compression_opts)
    h5.create_dataset("materials", data=list(colours.values()), dtype="float32", compression="gzip", compression_opts=compression_opts)

    print('#' * 100)
    print("Filesize", total_filesize)
    print('Open time', total_open_time)
    
run()
```

Moult:
OK, I think I worked out the reason there were funky results with BVH_Triangulation. When `.BVH()` is called, the LinearBuilder's `Build()` function [is called](https://github.com/Open-Cascade-SAS/OCCT/blob/20955d88da8615c54bd9f4221a3ff9fbe90fd915/src/BVH/BVH_PrimitiveSet.hxx#L100). This performs some clever radix sorting [here](https://github.com/Open-Cascade-SAS/OCCT/blob/20955d88da8615c54bd9f4221a3ff9fbe90fd915/src/BVH/BVH_LinearBuilder.hxx#L322) that [swaps the triangle elements ordering](https://github.com/Open-Cascade-SAS/OCCT/blob/20955d88da8615c54bd9f4221a3ff9fbe90fd915/src/BVH/BVH_RadixSorter.hxx#L234-L236).

However, in my code, I wasn't calling `triangulation.Elements[i]` (which are now reordered), I was calling my own `tris` and `normals` variables which weren't reordered. I did this for speed, since back when I was using BrepExtrema_TriangleSet the repeated calls to `GetVtxIndices()` was slow.

So basically the clashes were being performed on the wrong triangles and produce garbage data.

I've cleaned up the code a bit to try to match the internals of BrepExtrema_TriangleSet as much as possible as a starting point. Later on it may be tweaked and tailored to our specific requirements.

aothms:
Great work on figuring that out.

So in my attempts we also had:

1. separate product geometries into solids/shells so that we can rely on bvh-based point containment test instead of non-deterministic'ish ray hit counting

  I used topexp_explorer to first select solids then shells outside of solids, then faces outside of shells. And I augmented the template param with a size_t to distinguish between the parts in the map. What we could also do is use a multimap instead. I really think this ray hit counting is going to going to give bad results in a low, but nonetheless real number of cases. And doing twice of course only partially circumvents that.

2. think the distrinction between add/add_triangulated shouldn't be made at insertion time but rather at tree initialization.

  With a template paramater, but this is something that in the end doesn't affect end user functionality.

What are your thoughts on these items.

Moult:
Yes I don't have much to say on these: the first I don't really understand but the principle sounds good. The second is architectural so I 100% agree with anything you propose :)

I saw in your diff that you were perhaps halfway through implementing 1 and you already implemented 2. I'm happy to merge in your work on 2 but I'll need more hints before I can help implement 1 (exactly what functions can replace raycasting, and can this still be done later on when we read pure Tris from h5 without any occt shape data).

Alternatively I can continue in implementing the h5 write/load feature.

aothms:
Ok, forget about 2 for now then. I'll work this week on 1.

Moult:
The Python H5 converter script from [here](https://github.com/IfcOpenShell/IfcOpenShell/pull/4282#issuecomment-1940767853) has now been ported to C++. It's a very naive "does it compile" implementation and needs a lot of refactoring and there's at least one bug I've already noticed (I seem to have duplicate GlobalIds).

Next steps are to:

- [ ] Experiment with how to store the BVH tree data in H5
- [ ] Experiment with multithreading (the two reasons I'd like this to be in C++ is the multithreading and avoiding the need to ship with h5py)
- [ ] Begin refactoring so that it isn't just randomly dumped in the tree file.
- [ ] Clean up GlobalIds (I noticed there is an expand function in C++ somewhere I can perhaps reuse)

Isn't the existing H5 serialiser non-multithreaded? Would it be beneficial to instead put this in the geom iterator implementation instead to get multithreading for free?

aothms:
> Experiment with multithreading 

Multi-threading on IO bound tasks like this likely has very little effect. It may even make things worse because of all the locking that has to be put in place.

Moult:
Hmm, I'm looking to crush the 140 seconds of save/load time [mentioned here](https://github.com/IfcOpenShell/IfcOpenShell/issues/4279#issuecomment-1943207062). I did a couple of measurements:

108 seconds to convert to H5 via Python (89 seconds for geom iteration + ~19 seconds for H5 processing).

29 seconds to load a chunked H5 into Blender. (26 seconds of H5 loading / processing, and ~3 seconds of creating Blender objects)

I was hoping that merely using C++ / multithreading would be enough to crush either the 19 or 26 seconds.

Moult:
The good news is that with my attempt at porting the loading code to C++, the 29 seconds it took to load a H5 into Blender has now dropped to 6.6 seconds. Woohoo! (half the time loading the H5, and the other half creating Blender objects). No multithreading was used.

I found a crazy behaviour in H5 where getting a subgroup name was very, very slow. Maybe that explains why the Python code `for shape_id, shape in model["shapes"].items()` is so slow.

I also found that the casting from SWIG wrapped vector<foo> to Python list was very slow too. I got around this by implementing `numpy.i`.

The bad news is that I've gone past the point of knowing what I'm doing and I have absolutely no idea how this now compiles (I manually copied over numpy.i and the numpy include .h directory, and manually included something in CMakeLists that obviously only works on my machine. So there's a huge amount of cleanup to do ... but hey I'm still really excited about the numbers! :)

Moult:
Given there's a week left before the release, here's the coordination usecase wishlist I'd ideally like covered by then.

 1. I want to just open a model to inspect it visually as fast as possible
     - chunked loading (direct from geom iterator)
     - save sqlite
     - query sqlite
     - save to blend for future loading (no h5 necessary)
 2. I want to open many large models to inspect it visually with basic properties as fast as possible
     - Blender ui / operators to trigger preprocessing steps
     - save h5
     - save sqlite
     - load h5
     - query sqlite
 3. I want to run clash detection on a model I'm authoring
     - updated ifcclash
     - updated blender clash ui
     - build tree
     - run n:n clash functions
 4. I want to run clash detection on many large models
     - updated ifcclash
     - updated blender clash ui
     - save h5
     - load h5
     - (ideally not build) load tree
     - run n:n clash functions

Moult:
I tested with saving to H5 via C++ and it seems as though the 108 seconds have also dropped down to 89 seconds for saving a H5. I guess all the overhead in the past was in passing big lists to Python and handling those lists there.

I did a few measurements:

 - Given `%template(FloatVector) vector<float>;`, a `h5_shape.verts` is returned as a `<ifcopenshell.ifcopenshell_wrapper.FloatVector; proxy of <Swig Object of type 'std::vector< float > *' at 0x7f0f871a8ff0> >`. If I use this directly in Blender code (e.g. to create meshes with) for a random file it takes 14.7 seconds. Using the SWIG objects directly is slow it seems.
 - If I cast to list `list(e.verts)` and then create meshes using the Python list for all subsequent code it drops to 8.9 seconds.
 - If I use `numpy.i` and provide a numpy.ndarray directly from C++ then it drops to 6.5 seconds.

I wonder if this means that there could be a benefit in serving TriangulationElement geometry verts/edges/faces as numpy arrays (such as for general geometry iteration that everybody uses). I didn't look in detail as to how that's managed in SWIG but `type(shape.geometry.verts)` says tuple so maybe it's a different story.

Moult:
It's now possible to chunk directly from an IFC (instead of first having to save out a H5). This means that users can press a button in Blender and immediately load and see an IFC with a fast FPS. Something that would previously take almost 5 minutes and have them browse around with 3 FPS would now take 83 seconds and at 30 FPS.

And this should work for multiple models too! (Once I build the operator for it) Users would also be able to headlessly run it in the background and save out a Blend file and auto-link that blend file to the scene (and then memory used for ifcopenshell.open() should be freed I think). So in a single session they can load in many models and federate them conveniently.

(BTW you've probably noticed the code getting worse and worse, I'd love to clean it up but I need your guidance, and there's a lot of magic around SWIG which escapes me. I've definitely gone overscope in this and starting working on #4279 which is related but perhaps should be in a separate PR)

Moult:
OK, now when loading in a chunked model, you can activate a tool where you can "click" on objects. It'll "highlight" the object you clicked on (it's not a true editable object after all) and it'll query the SQLite db for object properties (see the right hand side of the screenshot).

2024-02-23-230139_1920x1080_scrot

Moult:
Next steps:

 1. Clash code to be reviewed by @aothms to be ready to merge. @Moult to rewrite ifcclash python frontend to work with it, and unbundle hppfcl deps. (excluding any H5 functionality related to IfcClash)
 2. @aothms to investigate alternative to numpy.i so iterator can return efficiently to Python. @Moult to write rewrite chunking in Python reading from this "buffer" and using numpy as efficiently as possible to do chunking and measure how this compares to the C++ process_chunk() / get_chunk(). Goal is to 1) reduce complexity of numpy returns 2) prevent extra code branching in iterator to maintain.
 3. @aothms to start separating H5 into its own distinct serialiser with resolving hacks like the 2 chunk definitions, the vector vs double
 4. Adding support for AABB/OBB in H5
 5. Read/write BVH tree tri swap indices
 6. Future: memory research on pure viewer

Moult:
Thanks to your awesome work @aothms I think we can close #4369, and tomorrow I'll do a quick commit to delete all the chunking C++ code, and do a bit of clean up to finish integrating it into Blender, and then I'd say action 2 is done!

PR #4320, opened 2024-02-14
Fix: add missing typename

Krande:
Hey @aothms, the failing daily conda builds are complaining about a missing typename in the hierarchy helper header file. This fixes it.


aothms:
Thanks @Krande !

Issue #4283, opened 2024-02-01
svg serializing misses a representation that consists of just a circle


Andrej730:
Noticed that svg drawing was missing a representation (Body/Plan/PLAN_VIEW) that consist of just 1 IfcCircle. In the file below it's representation `#941`. 

I've tried to narrow it down and noticed that if you add a rectangle (`IfcIndexedPolyCurve`) to the representation's items it prints both circle and rectangle.

Script and example file are below. Script is expecting a path to the `trees.ifc` at the end of the code and it prints svg generated for the representation `#941`. At the end of the code you can notice that I've added a rectangle to the representation `rep.Items = [rectangle, circle]` to make it work and commented it out, you can uncomment it for tests.

[trees.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14125404/trees.ifc.txt)

```python
import ifcopenshell
import ifcopenshell.geom
import multiprocessing
from mathutils import Vector
from collections import namedtuple
from pathlib import Path
from pprint import pprint

LineworkContexts = namedtuple(
    "LineworkContexts",
    (
        "plan_body_target plan_body_model model_body_target model_body_model plan_annotation_target "
        "plan_annotation_model model_annotation_target model_annotation_model"
    ),
)


class SvgWriter:
    def __init__(
        self,
        ifc_filepath: Path,
        ifc_file: ifcopenshell.file,
        drawing,
        drawing_document,
        drawing_elements: set[ifcopenshell.entity_instance],
    ):
        self.camera_element = drawing
        self.camera_document = drawing_document
        self.file = ifc_file
        self.drawing_elements = drawing_elements
        self.ifc_filepath = ifc_filepath

        # hardcoded
        self.scale = 0.01
        self.human_scale = "1:100"
        self.camera_width, self.camera_height = 50, 50

        self.drawing = drawing
        self.drawing_name = self.drawing.Name

        self.generate_linework()

    def get_linework_contexts(self, ifc, target_view) -> LineworkContexts:
        contexts = LineworkContexts(*[[] for _ in range(len(LineworkContexts._fields))])

        for rep_context in ifc.by_type("IfcGeometricRepresentationContext"):
            if rep_context.is_a("IfcGeometricRepresentationSubContext"):
                if rep_context.ContextType == "Plan":
                    if rep_context.ContextIdentifier in ["Body", "Facetation"]:
                        if rep_context.TargetView == target_view:
[...]

PR #4324, opened 2024-02-15
IfcGeom::util::is_manifold not to consider circles as manifold #4283

Andrej730:
Investigating I've found that SvgSerializer with prefiltering enabled (after #3359) was ignoring representations that consisted only of a circle curves since it was considered manifold by `IfcGeom::util::is_manifold(s)` in https://github.com/IfcOpenShell/IfcOpenShell/blob/0e0678b926c34f2c28e69558959d3b29e51a50ec/src/serializers/SvgSerializer.h#L433 and topology explorer `TopExp_Explorer exp(s, TopAbs_FACE)` ignored that shape as it had no faces.

IfcGeom::util::is_manifold considered a full circle manifold since it was ignoring edges with `v0.IsSame(v1)` considering them manifold but in case of a full circle, circle is a TopoDS_Edge with both vertices at the same location but it also has a circle curve and I've added check for that.

@aothms can you please take a look?

### DIFF
--- a/src/ifcgeom_schema_agnostic/base_utils.cpp
+++ b/src/ifcgeom_schema_agnostic/base_utils.cpp
@@ -11,6 +11,7 @@
 #include <gp_GTrsf2d.hxx>
 
 #include <Geom_Plane.hxx>
+#include <Geom_Circle.hxx>
 #include <Geom_OffsetSurface.hxx>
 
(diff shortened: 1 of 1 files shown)


Andrej730:
fixed with e6691989dc378956099bfe3adefeaa11874fdf88

PR #4348, opened 2024-02-21
First rough draft of alignment-based IfcFixedReferenceSweptAreaSolid

aothms:
I've implemented a mapping from IfcFixedReferenceSweptAreaSolid to `loft` (sequence of `face`) and implemented `loft` in the occt kernel. The directrix/axis is used to transform the various repetitions of face, but is not actually passed on, because the segments are interpolated linearly. The results appear reasonably correct.

I'm really quite happy with how this went. Thanks for all the great work on the C++ impl @RickBrice. I took the first the wrong column from the evaluate() matrix output at first, but after that it was really a breeze.

afbeelding

RDF.bg

afbeelding

ODA

afbeelding

afbeelding

Actually in ODA it becomes a flat surface.

### DIFF
--- a/src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp
+++ b/src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp
@@ -53,10 +53,19 @@ taxonomy::ptr mapping::map_impl(const IfcSchema::IfcFixedReferenceSweptAreaSolid
 			}
 		}
 #endif
-		auto len = end - start;
-		auto nsteps = (size_t)ceil(len);
-		for (size_t i = 0; i <= nsteps; ++i) {
-			auto m4 = pwf->evaluate(start + len / nsteps * i);
+		auto curve_length = end - start;
+		auto param_type = settings_.get<ifcopenshell::geometry::settings::PiecewiseStepType>().get();
+		auto param = settings_.get<ifcopenshell::geometry::settings::PiecewiseStepParam>().get();
+		size_t num_steps = 0;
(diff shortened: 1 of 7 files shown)


src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp
```diff
@@ -0,0 +1,90 @@
...
+			}
+		}
+		if (inst->EndParam() && inst->EndParam()->as<IfcSchema::IfcLengthMeasure>()) {
+			double e = *inst->EndParam()->as<IfcSchema::IfcLengthMeasure>();
+			if (e < end) {
+				end = e;
+			}
+		}
+#endif
+		auto len = end - start;
```
RickBrice, line 56: Consider using the Settings to control the loop. You can write something like this to programmatically control the fineness of the piecewise steps

~~~
	ifcopenshell::geometry::Settings settings;
	settings.get<ifcopenshell::geometry::settings::PiecewiseStepType>().value = ifcopenshell::geometry::settings::PiecewiseStepMethod::MAXSTEPSIZE;
	settings.get<ifcopenshell::geometry::settings::PiecewiseStepParam>().value = 1.0;
~~~

These settings are then used in the following code:
https://github.com/IfcOpenShell/IfcOpenShell/blob/a60e272eee613b68dd98581dfc8654a29199bf06/src/ifcgeom/taxonomy.cpp#L456-L478

You could do something similar in this loop. Alternatively, you could just call `evaluate()` without parameters and loop over the collection of edge objects returned

RickBrice (approved):
Had a break between meetings so I got this review done. Just the one comment.

I'm not too familiar with the OCC stuff yet but it looks like it works.

Issue #4312, opened 2024-02-10
BBim - UI Object Information panel feature request


trhyder:
The labeling within the Object Information panel's Object Metadata dropdown could be clearer and more intuitive. This can be achieved by addressing two key aspects:

1. Object Metadata Dropdown Labeling: The sub-dropdowns under Object Metadata currently start with the word "Object," which is redundant given their location within the Object Information tab and under the Object Metadata dropdown. To enhance clarity and streamline the interface, the labels should be adjusted as follows:
- "Object Property Sets" should be changed to "Property Sets"
- "Object Quantity Sets" should be changed to "Quantity Sets"
2. Inherited Psets Labeling and Icon Improvement: The labeling and icon for "Inherited Psets" could be improved for better comprehension and visual consistency. The suggested improvements are:
- Replace the up arrow icon associated with "Inherited Psets" with the linked icon. Currently, the arrow icon is misleading and directs the eye up rather than down. This would help the user understand that the property sets below are linked, and that changes to these will affect all others with the same link.
- Revise the label from "Inherited Psets" to "Type Linked:" to communicate that the property sets belong to their object's "type" and are not an inherited copy but a direct link.

<img width="343" alt="Screenshot 2024-02-10 at 12 11 59 PM" src="https://github.com/IfcOpenShell/IfcOpenShell/assets/59872808/ec313bba-3313-47fa-858c-b69c43ca727d">

theoryshaw:
Conceptually i like it. 

How about?...

"Object Property Sets" --> `Property Sets - Instance`
"Inherited Psets" --> `🔗 Property Sets - Type`
"Object Quantity Sets" --> `Quantity Sets - Instance`
"Inherited Qsets" --> `🔗 Quantity Sets - Type` 

note, the UI for "Inherited Qsets" doesn't exist currently.  Any Type Qsets, go under "Inherited Psets", which I think is not intended.

trhyder:
Nice, I like the use of "Instance" and "Type", I think it would help a lot. What about using "Instance" and "Type" as sub-headings under the Property Sets toggle?

## Property Sets
Instance:
* BBIM_Array
* EPset_Parametric
Type:
* Pset_CoveringCommon

## Quantity Sets
Instance:
Type:

trhyder:
[...]

PR #4352, opened 2024-02-22
Instance and Type sub-headings to UI

trhyder:
Changes to the Object Information tab:

1. Removed "Object" from the Object Property Sets and Quantity Sets headings
2. Added the sub-headings "Instance:" and "Type" to the Property Sets and Quantity Sets headings
3. Removed the parent icon.

image

Changes to the Geometry and Materials tab:

1. Changed sub-heading from "Inherited Material" to "Type:"
2. Removed duplication of type_material value
3. Removed parent icon

image

See https://github.com/IfcOpenShell/IfcOpenShell/issues/4312

Note: there is a bug with the Qtos appearing in the PSet "Type" list

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/spatial/workspace.py
+++ b/src/blenderbim/blenderbim/bim/module/spatial/workspace.py
@@ -100,31 +100,27 @@ class SpatialToolUI:
         row = cls.layout.row(align=True)
         row.prop(data=cls.model_props, property="rl3", text="RL")
         row = cls.layout.row(align=True)
-        row.label(text="", icon="EVENT_SHIFT")
-        row.label(text="", icon="EVENT_A")
-        row.operator("bim.generate_spaces_from_walls")
-        row = cls.layout.row(align=True)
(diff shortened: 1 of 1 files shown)


Andrej730:
Hi! Thank you for PR. 

Couple things I've noticed:
- it will still show "Instance" even if it's not a typed element or if it is a type
- it shows "Instance" even if there is no psets/qsets
- For psets/qsets it works fine without icon because "Type" / "Instance" label is outside of the `layout.box` that's containing the pset data. But for materials without icon it seems now harder to notice that material comes from type.

trhyder:
@Andrej730 thank you for the feedback. I'll remove the changes to the materials section from my pull request. It's my opinion, that the Materials section of the Geometry and Materials tab looks cluttered. My gut instance is that the Object Materials sub-section should be separated and be in its own section and moved to the Object Information tab, but I need to look more into this to make a more confident suggestion.

I'll fix the "Instance" errors you observed. Thank you

PR #4379, opened 2024-02-29
Make-bsdd-installable-and-add-tests

jgunstone:
hi there, 
I wanted to try this package out so downloaded and set up a testing environment. 
I thought that might be worthwhile to share back? 

I also added a `pyproject.toml` file, making it possible to install it as a standalone package straight from github - I added the install command and testing setup notes to the README.md. 

I setup the `pyproject.toml` using hatch, and tried to get rid of all of the generic stuff... if its not set up in the way you want it feel free to edit / ask me to edit

thanks - the package seems v useful!

### DIFF
--- a/src/bsdd/__about__.py
+++ b/src/bsdd/__about__.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: 2024-present jgunstone <j.gunstone@maxfordham.com>
+# SPDX-FileCopyrightText: 2024-present IfcOpenShell Developers
 #
-# SPDX-License-Identifier: MIT
-__version__ = "0.7.0"
+# SPDX-License-Identifier: LGPL-3.0-or-later
+__version__ = "0.1.0"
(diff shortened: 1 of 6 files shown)


src/bsdd/__about__.py
```diff
@@ -0,0 +1,4 @@
+# SPDX-FileCopyrightText: 2024-present jgunstone <j.gunstone@maxfordham.com>
+#
+# SPDX-License-Identifier: MIT
```
Moult, line 3: bSDD is under LGPL-3.0-or-later - would it be possible to change this?

Also the version is not tied to IfcOpenShell, so I guess we can start at whatever number we want (probably not 1.0.0 because it's probably not "stable" as yet, still experimental and bSDD is changing continuously, so maybe 0.1.0 or 0.0.1?)

Moult:
Great work! Only one minor comment above ;)

src/bsdd/__about__.py
jgunstone: yikes! apologies, didn't mean to change the license! will update now

jgunstone:
updated as per your comments - let me know if there is anything else

PR #4397, opened 2024-03-04
Fix proposal for # 3515 

bdamay:
I finally simply feeded operator bl_description to inform user on the way the list is ordered


bdamay:
@Moult Didn't mean to request a review - i misclicked somehow. I'm a bit confused with github PR workflow sometimes

Moult:
The failing test is our bad, I guess our testing system doesn't have support for `ifc4x3_rc2`. You can safely ignore it.

I noticed some commented out code, would you like to either uncomment it or delete it before merging?

PR #4404, opened 2024-03-07
Replace the specific "libpython3.8" debian package dependency for the…

fescale:
… generic "libpython3-dev".

When you build the debian package with a recent ubuntu version, the build uses a newer python but the package dependency is incorrectly set to libpython3.8. This dependency can't be satisfied in recent ubuntu releases, so the package can't be properly installed.

Using the generic "libpython3-dev" solves the issue and should also work with the older ubuntu releases.


aothms:
That makes a lot of sense. The .deb output from ctest/cmake is not something I personally regularly use. Thanks for looking into this.

PR #4409, opened 2024-03-09
Inline help feature

bdamay:
This PR provides a help button in Grouping and Filtering section - The operator opens webbrowser to relevant documentation section on blenderbim.org

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -1172,7 +1172,7 @@ endforeach(COMPONENT)
 
 set(CPACK_DEBIAN_PACKAGE_NAME "${PROJECT_NAME}")
 set(CPACK_DEBIAN_PACKAGE_MAINTAINER "${CPACK_PACKAGE_CONTACT}")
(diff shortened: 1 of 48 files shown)


bdamay:
@Moult I prefer you take a look before merging. Thank you for your help !

Moult:
Can you reuse the existing bim.open_uri operator instead of creating a new one?

```
row.operator("bim.open_uri", text="", icon="HELP").uri = "https://blahblah..."
```

bdamay:
Hi @Moult , thank you for your help,  i removed my extra operator, i didn't kow about generic operator for opening url, sorry. 

Doing that, i have lost the ability to set bl_description to that operator that then shows "(undocumented operator)" instead of something like "Get help on grouping and filtering" 
I tried to set thisvalue dynamically, but i actually didn't find how in blender api. Seems a can't set description attribute in operator function nor get an instance of the operator object on which i could set custom bl_description after creation. 

It might not be possible, but may be there is a trick ?

PR #4417, opened 2024-03-11
changed comparison for type relationships

### LINKED ISSUE #4402: IfcDiff on type relationships
The IfcDiff code that compares types reports changes even if their are no type changes.

This is because the comparison isn't done on the type id's but on the objects returned by get_type()

Current code:
```
            if relationship == "type":
                if ifcopenshell.util.element.get_type(old) != ifcopenshell.util.element.get_type(new):
                    self.change_register.setdefault(new.GlobalId, {}).update({"type_changed": True})
                    return True
```

Proposed code:
```
            if relationship == "type":
                if ifcopenshell.util.element.get_type(old) is not None and ifcopenshell.util.element.get_type(new) is not None:
                    if ifcopenshell.util.element.get_type(old).GlobalId != ifcopenshell.util.element.get_type(new).GlobalId:
                        self.change_register.setdefault(new.GlobalId, {}).update({"type_changed": True})
                        return True
                elif ifcopenshell.util.element.get_type(old) != ifcopenshell.util.element.get_type(new):
                    self.change_register.setdefault(new.GlobalId, {}).update({"type_changed": True})
                    return True
```

I do realise that this proposal is naive because the id of the type might not change while the type information does. Perhaps compare the string representation of the type?

JordanMcManus:
The original code compares the type objects returned by get_type(). I am not 100% sure if this is the cause, but the __eq__ method of entity_instance compares the file_pointer. Since the types come from different files they will never be equal.

This proposal compares the result of get_info() on the types instead.

fixes #4402

### DIFF
--- a/src/ifcdiff/ifcdiff.py
+++ b/src/ifcdiff/ifcdiff.py
@@ -280,7 +280,7 @@ class IfcDiff:
                 old_type = ifcopenshell.util.element.get_type(old)
                 new_type = ifcopenshell.util.element.get_type(new)
                 if old_type is not None and new_type is not None:
-                    if old_type.get_info() != new_type.get_info():
+                    if old_type.GlobalId != new_type.GlobalId:
                         self.change_register.setdefault(new.GlobalId, {}).update({"type_changed": True})
                         return True
                 elif old_type != new_type:
--- a/src/ifcdiff/ifcdiff.py
+++ b/src/ifcdiff/ifcdiff.py
@@ -277,7 +277,14 @@ class IfcDiff:
             return
         for relationship in self.relationships:
             if relationship == "type":
-                if ifcopenshell.util.element.get_type(old) != ifcopenshell.util.element.get_type(new):
+                old_type = ifcopenshell.util.element.get_type(old)
+                new_type = ifcopenshell.util.element.get_type(new)
+                if old_type is not None and new_type is not None:
+                    if old_type.get_info() != new_type.get_info():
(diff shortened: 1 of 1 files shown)


src/ifcdiff/ifcdiff.py
```diff
@@ -277,7 +277,14 @@ def diff_element_relationships(self, old, new):
             return
         for relationship in self.relationships:
             if relationship == "type":
-                if ifcopenshell.util.element.get_type(old) != ifcopenshell.util.element.get_type(new):
+                old_type = ifcopenshell.util.element.get_type(old)
+                new_type = ifcopenshell.util.element.get_type(new)
+                if old_type is not None and new_type is not None:
+                    if old_type.get_info() != new_type.get_info():
```
aothms, line 283: I don't have a good overview of the entire chain of logic here, but `get_info()` will return a dict with values, which may still be ifcopenshell.,entity_instances. So if the type has an IfcOwnerHistory or RepresentationMaps and these are from different files, than the dictionary comparison will still evaluate to false. The dict also contains the numeric instance identifier. I don't know if it's intended to include these in the comparison. Consider: `recursive=True` for the first, `include_identifier=False` for the second.

JordanMcManus:
My first instinct was to compare the GlobalId.

When I looked at the __eq__ method on entity_instance I saw part of the resolution involved get_info(), that's why I changed it. But now I see that only shovels the problem further out.

I'll change it back to comparing the GlobalId.

PR #4428, opened 2024-03-13
Add offset and limit to bsdd api queries

jgunstone:
a minor PR - adds limit and offset variables to Client function calls where applicable.
also adds IncludeReverseRelations and ReverseRelationDictionaryUris attributes to get_class call


jgunstone:
thanks @Moult

Issue #609, opened 2019-06-07
faces in .dae have inverted/missing faces after conversion from .ifc 


MassMachinery:
I was converting the model attached below with an outdated version of IfcOpenShell and experienced missing elements, probably relating to #466 so I updated to the latest version of IfcOpenShell but then I discovered that some elements seem to be inverted (see screenshots below). I added the model as .ZIP file. 

As clarification, in the older version of IfcOpenShell the inverted elements were showing correctly.
Could you please have a look at what could cause this issue? Thanks in advance!
 
Couch
![image](https://user-images.githubusercontent.com/7532951/59104137-03b35d00-8931-11e9-9cc0-94650cbbd14c.png)

Cupboards
![image](https://user-images.githubusercontent.com/7532951/59104097-e8e0e880-8930-11e9-8f98-414790ff58bd.png)

[Duplex_A_20110907_optimized.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/3265864/Duplex_A_20110907_optimized.zip)

aothms:
What version of IfcOpenShell are you using exactly? My recommendation is the head of v0.6.0 and invoke IfcConvert.exe with `--orient-shells` to fix faces incorrectly oriented in the file.

MassMachinery:
Awesome! Using  --orient-shells fixed it. This didn't seem necessary in older versions though. I was using  v0.6.0. Thanks for the speedy reply!

aothms:
> This didn't seem necessary in older versions though. I was using v0.6.0.

There was a couple of months time window in which this was indeed enabled by default. But it turned out too invasive still in case of extremely detailed meshes (see https://github.com/IfcOpenShell/IfcOpenShell/issues/578)

Issue #607, opened 2019-06-05
Boolean Operands like ST_SHAPELIST should not append all shapes


ninichki:
![image](https://user-images.githubusercontent.com/1086380/58967688-ef4a5580-87b4-11e9-99a4-a03072c0fd97.png)

ninichki:
Fixes maybe this Issue
"Timeout / performance regression on wall"

aothms:
Many thanks. If you already have it as a commit locally, isn't it easier to just send a Pull Request? These are valuable corrections and I really prefer to keep the author history correct.

ninichki:
Hi,
for me it's not allowed to create a new branch. So i can't make a pull request?

aothms:
> for me it's not allowed to create a new branch. So i can't make a pull request?

You can fork the repository first so that you have a full copy under your own account. There you can create a branch and create a pull request with the original parent repository. It's a couple clicks worth of work, but really pays off in the long run. Thanks in advance!

BIMvoice:
Superseded: the code this screenshot refers to no longer exists — `ST_SHAPELIST` has no occurrence in `src/ifcgeom/` on current `v0.8.0` after the taxonomy-based kernel rewrite. If the underlying concern (boolean operands aggregating all shapes) persists in the new kernel it would need re-diagnosis against the current code; as filed, this can be closed as superseded.

Issue #605, opened 2019-06-04
Regression: No geometry output from wall (example file)


johanrd:
Regression: Segfault on wall

1. Convert [604--wall--borked.ifc](https://github.com/IfcOpenShell/files/blob/master/604--wall--borked.ifc) with 68151b9 (v0.6.0 Feb 12 – OCCT 7.3.0p3) and experience that the wall converts successfully:
<img width="282" alt="Screenshot 2019-06-04 at 09 11 57" src="https://user-images.githubusercontent.com/4601554/58859141-6e526780-86a9-11e9-8b47-00622922a419.png">

2. Convert the same file with 1a5716287343163d4dc137dea762553f91833ac6 (v0.6.0 May 24 OCCT 7.3.0p3) and experience that there is no geomtetry output:
```Internal error, missing edge from triangulation```

johanrd:
Update: This file segfaults on OCCT 7.4.0 (with e37c882)

PR #4433, opened 2024-03-15
IfcSectionedSolidHorizontal

aothms:
I hope this is a step in the right direction.

I've tested a bit on the following file, with makes the spec example a bit more interesting with various interpolations, and it seems to match my understanding of what it should look like. @civilx64 @RickBrice hope you can have a look, curious to hear your comments and questions.

[sectioned-solid-horizontal-multiple.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14618218/sectioned-solid-horizontal-multiple.ifc.txt)

afbeelding

### DIFF
--- a/src/ifcgeom/kernels/opencascade/loft.cpp
+++ b/src/ifcgeom/kernels/opencascade/loft.cpp
@@ -24,6 +24,8 @@
 #include <BRepFill_Filling.hxx>
 #include <BRepTools_WireExplorer.hxx>
 #include <BRepBuilderAPI_MakeEdge.hxx>
+#include <BRepBuilderAPI_MakePolygon.hxx>
+#include <BRepBuilderAPI_MakeFace.hxx>
 
 using namespace ifcopenshell::geometry;
 using namespace ifcopenshell::geometry::kernels;
@@ -56,6 +58,7 @@ bool OpenCascadeKernel::convert(const taxonomy::loft::ptr loft, TopoDS_Shape& re
 			if (shps[i].ShapeType() != TopAbs_FACE) {
 				return false;
 			}
+			// @todo this is only outer wire
 			ws[i] = BRepTools::OuterWire(TopoDS::Face(shps[i]));
 		}
 		if (it == loft->children.begin()) {
@@ -69,10 +72,10 @@ bool OpenCascadeKernel::convert(const taxonomy::loft::ptr loft, TopoDS_Shape& re
 		BRepTools_WireExplorer a(ws[0]);
 		BRepTools_WireExplorer b(ws[1]);
 		for (; a.More() && b.More(); a.Next(), b.Next()) {
-			BRepFill_Filling fill;
 			auto& e1 = a.Current();
-			auto e3 = TopoDS::Edge(b.Current().Reversed());
-			
+			// auto e3 = TopoDS::Edge(b.Current().Reversed());
+			auto& e3 = b.Current();
+
 			// Documentation says unconnected edges are automatically connected, but this is not the case
 			TopoDS_Vertex e1a, e1b, e3a, e3b;
 			TopExp::Vertices(e1, e1a, e1b, true);
@@ -80,6 +83,8 @@ bool OpenCascadeKernel::convert(const taxonomy::loft::ptr loft, TopoDS_Shape& re
 			auto e2 = BRepBuilderAPI_MakeEdge(e1b, e3a).Edge();
 			auto e4 = BRepBuilderAPI_MakeEdge(e3b, e1a).Edge();
 			
+			/*
+			BRepFill_Filling fill;
 			fill.Add(e1, GeomAbs_C0);
 			fill.Add(e2, GeomAbs_C0);
 			fill.Add(e3, GeomAbs_C0);
@@ -87,6 +92,12 @@ bool OpenCascadeKernel::convert(const taxonomy::loft::ptr loft, TopoDS_Shape& re
 			fill.Build();
 			// faces.Append(fill.Face());
 			BB.Add(comp, fill.Face());
+			*/
+
+			auto f = BRepBuilderAPI_MakeFace(BRepBuilderAPI_MakePolygon(e1a, e1b, e3b, true).Wire()).Face();
+			BB.Add(comp, f);
+			auto g = BRepBuilderAPI_MakeFace(BRepBuilderAPI_MakePolygon(e3b, e3a, e1a, true).Wire()).Face();
+			BB.Add(comp, g);
 		}
 	}
 
--- a/src/ifcgeom/mapping/IfcSectionedSolidHorizontal.cpp
+++ b/src/ifcgeom/mapping/IfcSectionedSolidHorizontal.cpp
@@ -1,0 +1,197 @@
+/********************************************************************************
+ *                                                                              *
+ * This file is part of IfcOpenShell.                                           *
+ *                                                                              *
+ * IfcOpenShell is free software: you can redistribute it and/or modify         *
+ * it under the terms of the Lesser GNU General Public License as published by  *
+ * the Free Software Foundation, either version 3.0 of the License, or          *
+ * (at your option) any later version.                                          *
+ *                                                                              *
+ * IfcOpenShell is distributed in the hope that it will be useful,              *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 *
+ * Lesser GNU General Public License for more details.                          *
+ *                                                                              *
+ * You should have received a copy of the Lesser GNU General Public License     *
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.         *
+ *                                                                              *
+ ********************************************************************************/
+
+#include "mapping.h"
+#define mapping POSTFIX_SCHEMA(mapping)
+using namespace ifcopenshell::geometry;
+
+#include "../../ifcgeom/profile_helper.h"
+
+#include <boost/range/combine.hpp>
+
+#ifdef SCHEMA_HAS_IfcSectionedSolidHorizontal
+
+namespace {
+	// std::lerp when upgrading to C++ 20
+	template <typename T>
+	T lerp(const T& a, const T& b, double t) {
+		return a + t * (b - a);
+	}
+}
+
+taxonomy::ptr mapping::map_impl(const IfcSchema::IfcSectionedSolidHorizontal* inst) {
+	auto dir = map(inst->Directrix());
+	auto css = inst->CrossSections();
+	auto csps = inst->CrossSectionPositions();
+	std::vector<taxonomy::face::ptr> cross_sections;
+
+	// The PointByDistanceExpressesions are factored out into (a) a cartesian offset relative to the
+	// reference frame along a certain curve location (b) the longitude.
+
+	// The longitudes determine the range of the sweep and the offsets are interpolated in between
+	// sweep segments. 
+	std::vector<Eigen::Vector3d> profile_offsets;
+	std::vector<double> longitudes;
+
+	auto pwf = taxonomy::dcast<taxonomy::piecewise_function>(dir);
+	if (!pwf) {
+		// Only implement on alignment curves
+		return nullptr;
+	}
+
+	for (auto& cs : *css) {
+		cross_sections.push_back(std::move(taxonomy::cast<taxonomy::face>(map(cs))));
+	}
+	for (auto& csp : *csps) {
+		auto pbde = csp->Location()->as<IfcSchema::IfcPointByDistanceExpression>(true);
+		
+		longitudes.push_back(*pbde->DistanceAlong()->as<IfcSchema::IfcLengthMeasure>(true) * length_unit_);
+
+		// Corresponds to the profile X, Y directions (hopefully).
+		Eigen::Vector3d po(
+			pbde->OffsetLateral().get_value_or(0.),
+			// @todo I don't understand whether vertical is an offset relative to the tangent plane or to the global XY plane
+			pbde->OffsetVertical().get_value_or(0.),
+			0.
+		);
+
+		profile_offsets.push_back(po);
+	}
+	if (cross_sections.size() != profile_offsets.size()) {
+		Logger::Error("Expected CrossSections and CrossSectionPositions to be equal length, but got " + std::to_string(cross_sections.size()) + " and " + std::to_string(profile_offsets.size())  + " respectively", inst);
+		return nullptr;
+	}
+	if (cross_sections.size() < 2) {
+		Logger::Error("Expected at least two cross sections, but got " + std::to_string(cross_sections.size()), inst);
+		return nullptr;
+	}
+
+	auto loft = taxonomy::make<taxonomy::loft>();
+	// @todo intialize as default
+	loft->axis = nullptr;
+
+	// @todo currently only the case is handled where directrix returns a piecewise_function
+	if (pwf) {
+		double start = std::max(0., longitudes.front());
+		double end = std::min(pwf->length(), longitudes.back());
+
+		auto curve_length = end - start;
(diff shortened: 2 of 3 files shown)


aothms:
I'm going to merge this in to start a build, but comments still welcome :)

RickBrice:
Took me a little digging through the IFC to understand the image. There are 3 solids. One is along the gradient curve, one has a lateral offset of 50 at the end of the gradient curve, and one has a vertical offset at the mid-point. Yes?

I see lots of IfcLinearPlacement and IfcPointByDistanceExpression. Cool! That seems to be working.

@aothms Can you try this with an IfcOffsetCurveByDistances?

IfcSegmentedReferenceCurve isn't going to work yet. That's still a work in progress.

aothms:
> @aothms Can you try this with an IfcOffsetCurveByDistances?

afbeelding

[sectioned-solid-horizontal-ocbd.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14633755/sectioned-solid-horizontal-ocbd.ifc.txt)

Seems to work! Awesome. (None of the other viewers I tried render something for this though...)

I didn't realize though we already have logic for this interpolation. I also didn't realize the conceptual overlap between SectionedSolidHorizontal and OffsetCurveByDistances. They both are a curve with a list of linearly placed offsets. The only difference is that the solid also defines an equal-cardinality list of profiles.

The code we wrote is a bit different. You have a really interesting approach of defining the offsets as another piecewise function and returning the composition of both. I just iterate over the curve domain, lookup the span I'm currently in, and then apply the offset by lerp'ing. The consequence of that is that I will probably never hit exactly the locations at which the profiles/offsets are defined. I assumed that's not a big problem. And would also be the case if I use your approach, because ultimately the spans are hidden at the call site. It also makes sense that our approaches are different, because you return a new piecewise function and I just use the evaluation to position the profiles. But good to be aware of the conceptual overlap maybe and something to revisit/harmonize later.

> Took me a little digging through the IFC to understand the image. There are 3 solids. One is along the gradient curve, one has a lateral offset of 50 at the end of the gradient curve, and one has a vertical offset at the mid-point. Yes?

Yeah sorry, I was a bit in a rush. Yes for the offsets. Two of them also have a different end profile.

RickBrice:
> I also didn't realize the conceptual overlap between SectionedSolidHorizontal and OffsetCurveByDistances

Neither did I.

> You have a really interesting approach

Interesting good or interesting bad - If you have an alternative, I'm open to making the implementation better.

How are you making the IFC files and displaying them in Blender? (sorry, still bit of novice). Being able to visualize a solid might help me understand cant and whether or not I'm doing it right.

aothms:
> Interesting good or interesting bad 

Haha I don't know. I guess yours is computationally more efficient because you calculate the slope and multiplication is cheap. I'm lerping for every interpolated point. But maybe there is some overhead in your case for returning full matrices or std::function.

> How are you making the IFC files and displaying them in Blender?

Mostly hand editing actually. If you have some files in mind that you want me to wrap a solid around I don't mind. I'm curious to have a look as well.

RickBrice:
I've got a program that exports bridge superstructure models that uses IfcSectionedSolidHorizontal. I just don't what tools (IfcConvert maybe?) and what options to use to create something that can be opened and visualized in Blender.

aothms:
IfcConvert v0.8 to convert to Collada .dae which can be imported easily in Blender.

    IfcConvert.exe sectioned-solid-horizontal.ifc -yv sectioned-solid-horizontal.dae --use-element-guids

for the curve geometry:

    IfcConvert.exe sectioned-solid-horizontal.ifc -yv sectioned-solid-horizontal-curve.dae --use-element-guids --plan

RickBrice:
Thanks - my IFC bridge models are pretty weak. I've got an IFCOS build going right now. As soon as it's done I'll generate a post a couple of models.

They are precast, prestressed concrete bridges and post-tensioned spliced precast bridges. Deck, girders, traffic barriers, and prestress reinforcement only.

RickBrice:
Guess I'm not as far along as I thought - the beams, deck, and railings are coming out with decent geometry but their placement is a mess.

This file has a bunch of errors when running through IfcConvert

> [Notice] [2024-03-18 14:08:31] exclude entities IfcOpeningElement IfcSpace
> [Error] [2024-03-18 14:08:31] No operation defined for:
> #608=IfcSectionedSolidHorizontal(#113,(#195,#236,#277,#318,#359,#400,#441,#482,#523,#564,#605),(#197,#238,#279,#320,#361,#402,#443,#484,#525,#566,#607))
> [Error] [2024-03-18 14:08:31] Failed to convert:
> #609=IfcShapeRepresentation(#119,'Body','AdvancedSweptSolid',(#608))
> [Error] [2024-03-18 14:08:31] No operation defined for:
> #609=IfcShapeRepresentation(#119,'Body','AdvancedSweptSolid',(#608))

I need to be way more efficient with the beam cross section as well.

[I405-SR167 Flyover.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14642568/I405-SR167.Flyover.ifc.txt)

aothms:
> [Error] [2024-03-18 14:08:31] No operation defined for:
> #608=IfcSectionedSolidHorizontal(#113,(#195,#236,#277,#318,#359,#400,#441,#482,#523,#564,#605),(#197,#238,#279,#320,#361,#402,#443,#484,#525,#566,#607))

Are you sure you pulled the latest changes? IfcSectionedSolidHorizontal is exactly the scope of this PR.

But yeah... we have some work todo. 

afbeelding

RickBrice:
Oh my - that's embarrassing :)

aothms:
> I've got your commits on v0.8 branch from 2/21 and 3/13-3/15

That seems ok, but then you really shouldn't get an error about No operation defined for IfcSectionedSolidHorizontal

I also tried with https://github.com/IfcOpenBot/IfcOpenShell/commit/0d3c73de822d1b11ebc31e21d3b946e2df15d30b / https://s3.amazonaws.com/ifcopenshell-builds/IfcConvert-v0.8.0-0d3c73d-win64.zip and I don't get the error on the automated build either.

RickBrice:
I'm really stuck here with bridge models using IfcSectionedSolidHorizontal. I re-build and ran the INSTALL project from visual studio. This seems to have updated IfcConvert so the previous error for `No operation defined for IfcSectionedSolidHorizontal` has gone away.

Using a fresh build of IfcConvert (also using the pre-built one) I get

> Log:
> [Notice] [2024-03-20 12:23:21] exclude entities IfcOpeningElement IfcSpace
> [Error] [2024-03-20 12:23:21] Failed to convert:
> #487=IfcShapeRepresentation(#88,'Body','AdvancedSweptSolid',(#486))
> [Error] [2024-03-20 12:23:21] No operation defined for:
> #487=IfcShapeRepresentation(#88,'Body','AdvancedSweptSolid',(#486))

That tells me I'm doing something wrong with the shape representation and representation contexts. Using your sectioned-solid-horizontal-multiple.ifc file as a guide, I updated my coding so the representation context, representation subcontext, and shape representations are the same. 

Using IfcConvert without the --plan option, there are no errors and no dae file created. With the --plan option I get the 'No operation defined for' error and no geometry.

Any ideas?

The attached is a more simple model than before - one span, 5 beams, straight, flat road.

[PGSuper1.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14671671/PGSuper1.ifc.txt)

RickBrice:
Is this an error in the IfcSectionedSolidHorizontal documentation?

> The profile X axis is the direction of RefDirection from IfcAxis2PlacementLinear

If RefDirection is omitted in IfcAxis2PlacementLinear it defaults to the tangent of the curve. This would put the section profile x-axis along the curve making the sweep nonsensical. 

Maybe it should be the profile x-axis is aligned with the placement x-axis defined by the cross product of Axis and RefDirection.

aothms:
This is strange. For me that model fails on a null ObjectPlacement in the mapping of IfcProduct. And it's not an exception that can be caught. I've fixed it in here 4b32381192fcd4d2b77831a18ab5f5081ed0e5d8

After that output looks like this on my end:

afbeelding

> Is this an error in the IfcSectionedSolidHorizontal documentation?

>> IfcAxis2PlacementLinear:
>> 
>> RefDirection: The direction used to determine the direction of the local X Axis... If RefDirection is omitted, the direction is taken from the curve tangent at Location.

>> IfcSectionedSolidHorizontal:
>>
>> The profile X axis is the direction of RefDirection from IfcAxis2PlacementLinear, and the profile Y axis is the direction of Axis.

To be honest I just randomly tried a couple of axis permutations until it looked well, but maybe I should have had a proper look.

At least linguistically it seems to make sense. X matches X. The other axis doesn't linear Z corresponds to profile Y. 

My problem is that I don't understand linear placement I guess.

- What is the value of Axis/Z when null in the model?
- I find X the least sensible axis to align to curve tangent. In most conventions either Z (e.g opengl camera) or Y corresponds to the "forward" axis. Also in the name: "RefDirection" is a supplementary direction, but isn't in a curve the tangent the primary direction.

But this is probably something to bring up with the IF again..

RickBrice:
> For me that model fails on a null ObjectPlacement in the mapping of IfcProduct

I should have run my file through the validation service. It found that issue.

The bridge model is what I expected. Simple, but a step in the right direction.

The attached files are:
1) sectioned-solid-horizontal.ifc from the Ifc documentation, except that I changed the cross section shape so it is asymmetric.
2) sectioned-solid-horizontal_With_PlanOption_Output.txt - output from running IfcConvert with the --plan option
3) sectioned-solid-horizontal_With_PlanOption.dae - the Collada file - when imported in Blender there is no visible geometry
4) sectioned-solid-horizontal_Without_PlanOption_Output.txt - output from running IfcConvert without the --plan option
5) sectioned-solid-horizontal_Without_PlanOption.dae - the Collada file - when imported in Blender there is visible geometry

With the --plan option there are 'Failed to convert' with 'No operation defined for' `IfcShapeRepresentation(#17,'Body','AdvancedSweptSolid',(#116))`.

There is also a bunch of warnings about `No material and surface styles for' the IfcCompositeCurve, IfcGradientCurve, and IfcCurveSegment but I think this is expected.

Without the --plan option there is a bunch of errors for `Failed to convert` with `No operation defined for` IfcShapeRepresentation for 'Axis' and 'FootPrint' curves. I think this is probably expected too, but not sure.

Are there bugs here?

[sectioned-solid-horizontal.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14697209/sectioned-solid-horizontal.ifc.txt)
[sectioned-solid-horizontal_With_PlanOption.dae.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14697210/sectioned-solid-horizontal_With_PlanOption.dae.txt)
[sectioned-solid-horizontal_With_PlanOption_Output.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14697211/sectioned-solid-horizontal_With_PlanOption_Output.txt)
[sectioned-solid-horizontal_Without_PlanOption.dae.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14697212/sectioned-solid-horizontal_Without_PlanOption.dae.txt)
[sectioned-solid-horizontal_Without_PlanOption_Output.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14697213/sectioned-solid-horizontal_Without_PlanOption_Output.txt)

RickBrice:
> My problem is that I don't understand linear placement I guess.

My problem is that I don't understand placement in general I guess.

For sectioned-solid-horizontal.ifc file in the previous comment, I changed the shape of the cross section so the left side is really wide compared to the right side, when looking in the X-Y plane the section is defined in. I changed (-5,-1) to (-50,-1).

I expected the wide side of the cross section to be on the left in the IfcSectionedSolidHorizontal but it was on the right. After reviewing the documentation, this is expected.

> The profile normal is derived from the associated [IfcAxis2PlacementLinear](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/lexical/IfcAxis2PlacementLinear.htm), not necessarily the tangent of the Directrix

It seems that the profile X-axis maps to the linear placement Y-axis and the profile Y-axis maps to the linear placement Z-axis. Doing the right-hand rule cross product the profile normal is the profile Z-axis and is pointing out of the screen. For this axis to align with the linear placement X-axis the mirror image of the profile shape is projected along the curve.

This brings me to the conclusion that the documentation in IfcSectionedSolidHorizontal is not correct.

> The profile X axis is the direction of RefDirection from [IfcAxis2PlacementLinear](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/lexical/IfcAxis2PlacementLinear.htm)

This should be the profile normal is the direction of RefDirection from IfcAxis2PlacementLinear.

There is still an open question in the IF about the default Axis and RefDirection https://github.com/buildingSMART/IFC4.x-IF/issues/125#issue-1995149407. I'll ping the IF to see if there are any resolutions.

RickBrice:
Here is that first bridge model I sent you. Can you see how it converts and renders in Blender?

I don't seem to get any geometry in Blender, but do in RDF and ODA.

[I405-SR167 Flyover.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14705221/I405-SR167.Flyover.ifc.txt)

aothms:
> This should be the profile normal is the direction of RefDirection from IfcAxis2PlacementLinear.

I think that's a good way to put it, but strictly speaking the profile is 2d, so even I perfectly understand what you mean, maybe it needs to be augmented a bit to indicate that this normal is obtained when embedding the 2d profile in 3d space. I would also add another clarification indicating that RefDirection is the curve tangent.

---

I must say, I'm not getting more optimistic about the state of 4.3 support in software. But yeah, good news is compared to the rest, we, with 0 buget, are doing ok :)

BimVision. A bit wobly

afbeelding

ODA. Weird plates.

afbeelding

rdf.bg. Only the beams.

afbeelding

---

IfcOpenShell. Discontinuities in gradient curve.

afbeelding

For us the problem appears to be in the gradient curve evaluation. In order for that to show up I had to uncomment the footprint. For some reason it takes priority for visualization. Multiple representation and the selection of curve/surface types is still not ideal in v0.8 as you also observed in the logs.

If I comment out the footprint:

```
#147=IFCPRODUCTDEFINITIONSHAPE('Alignment Product Definition Shape',$,(/*#122,*/#123));
```

And then run

```
IfcConvert I405-SR167.Flyover.ifc -yvv I405-SR167.Flyover.dae --use-element-guids --plan --include attribute GlobalId 3rzI5aF7nFWABTp2u297Ph
```

I get the discontinuous gradient curve to display in Blender.

RickBrice:
Thanks for creating those images for me. ODA version 24 did a better job than version 25 on the same file.

I don't get the same result for Blender. I just get an empty canvas with the default cube.
The steps I'm using are:
1) Remove FootPrint representation
2) Fire up Blender 3.6
3) File > Import > Collada
4) Select the attached dae file
5) Look around the scene... nothing but the default cube

It appears to be loading the alignment
image

[I405-SR167.Flyover.dae.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/14724282/I405-SR167.Flyover.dae.txt)

I'll take a look at the gradient issues. It was correct before, but with all the guessing on implementation I probably broke something.

aothms:
When you select it from the "outliner" (your image), on the 3d view menu select Frame selected. Make sure that you have reasonable clipping planes in the view menu as well. Use the little arrow to get to that menu if you don't see it.

afbeelding

afbeelding

afbeelding

> Edit - the gradient curve issue is bad IFC coding, not the IFCOS implementation.

But how come than the other viewers don't have the weird jumps?

RickBrice:
Thanks - I thought I tried Frame Selected, but obviously not. I don't have any real experience with Blender. Obviously, there is still more work to do. Being able to visualize the geometry will help tremendously.

image


> But how come than the other viewers don't have the weird jumps?

Maybe I'm not doing the IFC coding and the IFCOS implement correctly in general. There could be offsetting errors that end up looking ok when I dump the piecewise function points and plot them in Excel.

PR #4436, opened 2024-03-16
when `bim.select_aggregate` is run, it deselects the initial object

theoryshaw:
ping @brunoperdigao 

After `bim.select_aggregate`, i often reassign the aggregate to another parent aggregate. 

The PR just streamlines things, as I don't have to deselect the initial object, before assigning the aggregate.


brunoperdigao:
I think this is addressed in #4437

PR #4437, opened 2024-03-17
shift + bim.select_aggregate = Selects all related aggregates from a …

theoryshaw:
ping @brunoperdigao 

- shift + bim.select_aggregate = Selects all related aggregates from a selection set, and parts within.
- shift + bim.select_linked_aggregates = Selects all related aggregates, and parts within.

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/9357b5be-facd-4314-9db9-15c64e35daa4

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/aggregate/operator.py
+++ b/src/blenderbim/blenderbim/bim/module/aggregate/operator.py
@@ -201,17 +201,87 @@ class BIM_OT_select_aggregate(bpy.types.Operator):
     bl_label = "Select Aggregate"
     bl_options = {"REGISTER", "UNDO"}
     obj: bpy.props.StringProperty()
+    bl_description = (
+        "Selects the part's aggregate empties.\n\n"
+        + "SHIFT+CLICK - selects all the aggregate empties and all the parts within"
+    )
+
+    select_all: bpy.props.BoolProperty(name="Select All Parts", default=False, options={"SKIP_SAVE"})
+
+
+    def invoke(self, context, event):
+        if event.type == "LEFTMOUSE" and event.shift:
+            self.select_all = True
+        return self.execute(context)    
 
     def execute(self, context):
         self.file = IfcStore.get_file()
-        obj = bpy.data.objects.get(self.obj) or context.active_object
-        aggregate = ifcopenshell.util.element.get_aggregate(tool.Ifc.get_entity(obj))
-        aggregate_obj = tool.Ifc.get_object(aggregate)
-        if aggregate_obj in context.selectable_objects:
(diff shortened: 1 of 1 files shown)


src/blenderbim/blenderbim/bim/module/aggregate/operator.py
```diff
@@ -201,18 +201,88 @@ class BIM_OT_select_aggregate(bpy.types.Operator):
...
+            for part in parts:
+                if part.IsDecomposedBy:
+                    for subpart in part.IsDecomposedBy[0].RelatedObjects:
+                        parts.append(subpart)
+                parts_objs.append(part)
+
+            for element in parts_objs:
+                obj = tool.Ifc.get_object(element)
+                if obj:
+                    obj.select_set(True)
```
brunoperdigao, line 246: There is no need to create a list, for the parts, you can select them in the end of loop. See the last comment example.

src/blenderbim/blenderbim/bim/module/aggregate/operator.py
```diff
@@ -290,9 +372,26 @@ def _execute(self, context):
...
+                    parts_objs.append(part)
+
+                for element in parts_objs:
+                    obj = tool.Ifc.get_object(element)
+                    if obj:
+                        obj.select_set(True)
+            else:
+                for element in parts:
+                    obj = tool.Ifc.get_object(element)
+                    obj.select_set(True)
```
brunoperdigao, line 394: - I suggest to use underscore, because it's a Python convention. Use `group_link`
- I would change some other variable names. What you called `parts` is actually a list of aggregates that belong to the same group, right? So it should be called "linked_aggregates". And what you called `subpart` is actually a `part`.
- Also, there is no need to create a list, for the parts, you can select them in the end of loop.

So it would become something like this:
```python
        for group_link in group_rel:
            linked_aggregates = list(group_link.RelatedObjects)
            if self.select_all:
                for aggregate in linked_aggregates:
                    if aggregate.IsDecomposedBy:
                        for part in aggregate.IsDecomposedBy[0].RelatedObjects:
                            print(part)
                            obj = tool.Ifc.get_object(part)
                            if obj:
                                obj.select_set(True)
            else:
                for aggregate in linked_aggregates:
                    obj = tool.Ifc.get_object(aggregate)
                    obj.select_set(True)

```

brunoperdigao (commented):
@theoryshaw The PR is working fine and it's a great addition. My comments are just small things to improve in the code.

theoryshaw:
@brunoperdigao very cool... Thanks for the comments!  I'll go through and make adjustments.

PR #4456, opened 2024-03-21
Unescaped ; was causing problems in Inkscape

FabioPiccinini:
An unescaped `;` was causing problems when opening svgs that included this style with Inkscape


aothms:
This is in a comment, maybe instead reply `;` with `-` or `,` as it leaves the textual intent more in tact?

I think most people here would use inkscape as well though.

FabioPiccinini:
I agree, the solution I proposed makes the comment more difficult to read.
If it were possible to replace the ; with - or , it would be more readable.

aothms:
Thanks!

FabioPiccinini:
Glad to help! I am happy to have made my first contribution, even if it's only a comma 😄

PR #4507, opened 2024-04-03
Add CCI Classification schema

ppaawweeuu:
I'd like to add CCI classification to BBIM. It is very similar to CCS, developed by CCI-C and licensed under CC BY 4.0 More details here: https://cci-collaboration.org/the-standard/

This is my first commit, I used CCS(EN).ifc file as reference. Any feedback welcomed.


Moult:
Awesome thank you! If you could also share the script you used to generate the classification that would be awesome! That way it can be regenerated in the future when the classification is updated.

PR #4516, opened 2024-04-05
BBIM tree names, make it possible to adjust layout

berndhahnebach:
https://community.osarch.org/discussion/comment/20038/#Comment_20038

PR #4517, opened 2024-04-05
BBIM tree names, make it possible to adjust layout

berndhahnebach:
https://community.osarch.org/discussion/comment/20038/#Comment_20038

duplicate of #4516 but I can not reopen it for some reason ...

### DIFF
--- a/src/blenderbim/blenderbim/bim/import_ifc.py
+++ b/src/blenderbim/blenderbim/bim/import_ifc.py
@@ -800,7 +800,8 @@ class IfcImporter:
         for axis in axes:
             shape = self.create_generic_shape(axis.AxisCurve)
             mesh = self.create_mesh(axis, shape)
-            obj = bpy.data.objects.new(f"IfcGridAxis/{axis.AxisTag}", mesh)
+            obj_name = tool.Loader.get_name_layout("IfcGridAxis", axis.AxisTag)
+            obj = bpy.data.objects.new(obj_name, mesh)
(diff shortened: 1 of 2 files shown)


Andrej730:
Renaming objects should also be in sync with this - https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/blenderbim/blenderbim/bim/handler.py/#L41-L42

berndhahnebach:
I did not rename anything. I just created some method to be able to reaname all tree obj. at once.

This was not used at the code I changed. How would this be integrated ? 

Is this needed for the PR allready ?

Andrej730:
> I did not rename anything. I just created some method to be able to adjust the layout of the tree object names at once.
> 
> This was not used at the code I changed. How would this be integrated ?
> 
> Is this needed for the PR allready ?

I think if we're moving object naming at import to a separate function in this PR then to keep it consistent we should also use the same function when user would try to rename object in outliner.

berndhahnebach:
Do you have an example how to make use of this class to check names?

berndhahnebach:
rebased on latest HEAD

How about an example to use the class for checking the names?

Andrej730:
> How about an example to use the class for checking the names?

Check the handler function above, it gets triggered every time you change object's name. If we plan to allow different approach how objects named during import then we also need to make sure this approach is maintained when users will be renaming them later on.

Moult:
We now use the loader function for names so closing.

Issue #4520, opened 2024-04-05
Error on import of ifcopenshell.util.unit 0.7.0.240405 vs Python 3.9


sebphi:
I just upgraded to `0.7.0.240405` and got the following error:

```
$ python
Python 3.9.18 (main, Aug 25 2023, 13:20:14) 
[GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ifcopenshell.util.unit
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/sebphi/.local/share/virtualenvs/phportal-oMXq_Eu-/lib/python3.9/site-packages/ifcopenshell/util/unit.py", line 671, in <module>
    content_type: ifcopenshell.ifcopenshell_wrapper.named_type | ifcopenshell.ifcopenshell_wrapper.type_declaration,
TypeError: unsupported operand type(s) for |: 'type' and 'type'
```

I would be happy if the previous version 0.7.0.231218 was available on pypi again as i have currently no need for a newer version. Cheers!

KaareH:
I'm facing the same issue:

Quote from closed duplicate issue #4522 :

> In commit 8aecee583677ab999a549f40436fd2711ab13fa1 the union operator `|` is used, which is only supported in Python > 3.9.
> 
> If one tries to `import ifcopenshell.util.unit`, `TypeError: unsupported operand type(s) for |: 'type' and 'type'` is raised.
> 
> Using `ifcopenshell==0.7.0.240405` and `Python==3.9.10`.

aothms:
Related #4447 also regarding compatibility of typing annotations. I think py39 compatibility is something we probably want to retain for the time being. Feel free to send in a pull request (`from typing import Union; Union[ifcopenshell.ifcopenshell_wrapper.named_type | ifcopenshell.ifcopenshell_wrapper.type_declaration]` probably)

sebphi:
Thanks for your swift response @aothms , here is the PR: https://github.com/IfcOpenShell/IfcOpenShell/pull/4523

Moult:
There's now a new release on PyPI. Can this be closed?

sebphi:
Yes, works fine on my end now. Cheers!

PR #4523, opened 2024-04-05
Fix import of ifcopenshell.util.unit in Python <3.10 #4520

sebphi:
This one-line fix makes ifcopenshell work for me again in Python < 3.10. I suppose it will also work for > 3.9.


aothms:
Thanks!

PR #4573, opened 2024-04-23
Fix HDF5 conditional inclusion in GeomTree

stefkeB:
IfcGeomTree uses HDF5, but this is an optional module, so we have to wrap it with #ifdef WITH_HDF5


aothms:
Thanks @stefkeB, much appreciated.

PR #4583, opened 2024-04-24
changed IfcGeom to link SCHEMA_AGNOSTIC_FILES shared

FlorianB-DE:
changed IfcGeom to link SCHEMA_AGNOSTIC_FILES shared

I had the issue that in my statically linked the linker would not link with "-Wl,--no-undefined" and therefore it was not portable.
Now it should be


aothms:
I don't think this is necessary. With `-DBUILD_SHARED_LIBS=On` when configuring CMake all libraries are set to `shared`.

FlorianB-DE:
Ima go verify. 1 Sec

FlorianB-DE:
Whoopsy. Thank you :D

PR #4590, opened 2024-04-27
Comment std::cout causing console spam

KaareH:
Prevent console spam when using `tree.add_element`.


aothms:
Thanks. Indeed quite an oversight to leave that in.

Issue #4608, opened 2024-05-04
IfcCSV - importing enumerated values doesn't work


ppaawweeuu:
When we export enumerated values from e.g. Pset_WallCommon.Status we get list of list of values. Properties separated by `;` (semicolon) and enumerated string values separated by `,` (comma). But when we import that csv file, using the same settings, enumerated values stay with no changes, error in console.

Enumerated values:
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/61344631/a0f778e3-85e1-463f-bff7-5008f9faaa95)

Export setting:
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/61344631/fcd41a27-ed6f-4758-b508-6245a59c8f69)

Csv before changes:
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/61344631/5822edec-7547-4270-9b1b-e10b171ea09d)

Csv after changes:
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/61344631/4dcfd2a3-dc8a-402f-a6bf-2872c2c2b049)

Console after importing csv:
```
Traceback (most recent call last):
  File "C:\Users\user_name\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\blenderbim\bim\prop.py", line 86, in get_attribute_enum_values
    data = json.loads(prop.enum_items)
  File "C:\Program Files\Blender Foundation\Blender 4.0\4.0\python\lib\json\__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "C:\Program Files\Blender Foundation\Blender 4.0\4.0\python\lib\json\decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "C:\Program Files\Blender Foundation\Blender 4.0\4.0\python\lib\json\decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
File "C:\Users\user_name\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\blenderbim\bim\prop.py", line 83, in get_attribute_enum_values
[...]

PR #4626, opened 2024-05-06
ifccsv - support importing enum properties #4608

Andrej730:


ppaawweeuu:
Works fine, Can we merge it, pls :)

Moult:
I just took a look at this, is there a reason this isn't implemented in ifcopenshell.util.selector.set_element_value and uses this hack instead?

Andrej730:
> I just took a look at this, is there a reason this isn't implemented in ifcopenshell.util.selector.set_element_value and uses this hack instead?

to use the same concatentation string that was used to serialize the list for enum property

Andrej730:
Superseded by c9847ca

PR #4693, opened 2024-05-19
Update installation.rst refering to System not Import-Export

eshattow:
Blender BIM addon is listed by Blender 4.1 as `System: BlenderBIM`


Andrej730:
Thank you!

PR #4743, opened 2024-05-26
fix #3802 - Copy array attributes from another array

### LINKED ISSUE #3802: Pick up the array attributes of another array
Would be nice to pick up the array attributes of another array

image

theoryshaw:
https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/7fae2a1e-476a-42ea-8db2-a137b8e288bd

### DIFF
--- a/src/blenderbim/blenderbim/bim/module/model/array.py
+++ b/src/blenderbim/blenderbim/bim/module/model/array.py
@@ -83,17 +83,32 @@ class EnableEditingArray(bpy.types.Operator, tool.Ifc.Operator):
     def _execute(self, context):
         obj = context.active_object
         element = tool.Ifc.get_entity(obj)
-        data = json.loads(ifcopenshell.util.element.get_pset(element, "BBIM_Array", "Data"))[self.item]
         props = obj.BIMArrayProperties
-        props.count = data["count"]
-        si_conversion = ifcopenshell.util.unit.calculate_unit_scale(tool.Ifc.get())
-        props.x = data["x"] * si_conversion
-        props.y = data["y"] * si_conversion
-        props.z = data["z"] * si_conversion
-        props.use_local_space = data.get("use_local_space", False)
-        props.sync_children = data.get("sync_children", False)
-        props.method = data.get("method", "OFFSET")
+        
+        relating_obj = props.relating_array_object
+        
+        if relating_obj:
+            relating_props = relating_obj.BIMArrayProperties
+            
+            props.count = relating_props.count
+            props.x = relating_props.x
+            props.y = relating_props.y
+            props.z = relating_props.z
+            props.use_local_space = relating_props.use_local_space
+            props.sync_children = relating_props.sync_children
+            props.method = relating_props.method
+        else:
+            data = json.loads(ifcopenshell.util.element.get_pset(element, "BBIM_Array", "Data"))[self.item]
+            si_conversion = ifcopenshell.util.unit.calculate_unit_scale(tool.Ifc.get())
+            props.x = data["x"] * si_conversion
+            props.y = data["y"] * si_conversion
+            props.z = data["z"] * si_conversion
+            props.use_local_space = data.get("use_local_space", False)
+            props.sync_children = data.get("sync_children", False)
+            props.method = data.get("method", "OFFSET")
+        
         props.is_editing = self.item
+        
         return {"FINISHED"}
 
 
@@ -137,6 +152,10 @@ class EditArray(bpy.types.Operator, tool.Ifc.Operator):
 
         tool.Blender.Modifier.Array.set_children_lock_state(element, self.item, True)
         tool.Blender.Modifier.Array.constrain_children_to_parent(element)
+
+        #clears the relating_array_object so it doesn't show again next time
+        props.relating_array_object = None
+        
         return {"FINISHED"}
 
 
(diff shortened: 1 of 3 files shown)


src/blenderbim/blenderbim/bim/module/model/prop.py
```diff
@@ -87,6 +87,41 @@ def update_type_page(self, context):
...
 
 
+def update_relating_array_from_object(self, context):
+    if self.relating_array_object is None:
+        return
+    element = tool.Ifc.get_entity(self.relating_array_object)
+    if not element:
+        return
+    # If a child is selected, the following assigns the parent to the relating_array_object
+    parent_globalid = element.IsDefinedBy[0].RelatingPropertyDefinition.HasProperties[0].NominalValue.wrappedValue
```
brunoperdigao, line 97:
```suggestion
    pset = ifcopenshell.util.element.get_pset(element, "BBIM_Array")
    parent_globalid = pset["Parent"]
```

src/blenderbim/blenderbim/bim/module/model/prop.py
```diff
@@ -87,6 +87,41 @@ def update_type_page(self, context):
...
+    parent_element = tool.Ifc.get().by_guid(parent_globalid)
+    parent_of_relating_array_object = tool.Ifc.get_object(parent_element)
+    #The following error handling is used to break out when an array parent is selected
+    try:
+        element.IsDefinedBy[0].RelatingPropertyDefinition.HasProperties[1].NominalValue.wrappedValue
+    except AttributeError:
```
brunoperdigao, line 103:
```suggestion
        pset['Data']
    except Exception:
```

src/blenderbim/blenderbim/bim/module/model/prop.py
```diff
@@ -87,6 +87,41 @@ def update_type_page(self, context):
...
+        pass
+    else:
+        bpy.ops.bim.enable_editing_array(item=self.is_editing)
+        return 
+
+    self.relating_array_object = parent_of_relating_array_object
+    bpy.ops.bim.enable_editing_array(item=self.is_editing)
+
+    return self.relating_array_object
```
brunoperdigao, line 112: I was getting an error that the function should return `None`. 
```suggestion
    return
```

src/blenderbim/blenderbim/bim/module/model/prop.py
```diff
@@ -87,6 +87,41 @@ def update_type_page(self, context):
...
+
+
+def is_object_array_applicable(self, obj):
+    element = tool.Ifc.get_entity(obj)
+    if not element:
+        return False
+    return element.IsDefinedBy[0].RelatingPropertyDefinition.Name == "BBIM_Array"
```
brunoperdigao, line 119:
```suggestion
    return ifcopenshell.util.element.get_pset(element, "BBIM_Array")
```

src/blenderbim/blenderbim/bim/module/model/array.py
```diff
@@ -83,17 +83,32 @@ class EnableEditingArray(bpy.types.Operator, tool.Ifc.Operator):
...
+            props.method = relating_props.method
+        else:
+            data = json.loads(ifcopenshell.util.element.get_pset(element, "BBIM_Array", "Data"))[self.item]
+            si_conversion = ifcopenshell.util.unit.calculate_unit_scale(tool.Ifc.get())
+            props.x = data["x"] * si_conversion
+            props.y = data["y"] * si_conversion
+            props.z = data["z"] * si_conversion
+            props.use_local_space = data.get("use_local_space", False)
+            props.sync_children = data.get("sync_children", False)
+            props.method = data.get("method", "OFFSET")
```
brunoperdigao, line 108: You can simplify this:

```suggestion
            element = tool.Ifc.get_entity(relating_obj)
        data = json.loads(ifcopenshell.util.element.get_pset(element, "BBIM_Array", "Data"))[self.item]
        props.count = data["count"]
        si_conversion = ifcopenshell.util.unit.calculate_unit_scale(tool.Ifc.get())
        props.x = data["x"] * si_conversion
        props.y = data["y"] * si_conversion
        props.z = data["z"] * si_conversion
        props.use_local_space = data.get("use_local_space", False)
        props.sync_children = data.get("sync_children", False)
        props.method = data.get("method", "OFFSET")
```

brunoperdigao (commented):
It wasn't working properly on my end, so I made a few suggestions that I think we'll make it work and also improve the readability. I'm still getting a warning from Blender "Failed to set value" which I don't fully understand why. However, it doesn't seem to break the functionality

theoryshaw (commented):
Thanks!  I'll try these changes.

theoryshaw:
Thanks @brunoperdigao!
The one thing that doesn't seem to work anymore, is that previously you could pick either the parent or the child, and it would work.  Now only picking the parent work.  Picking the child, does not work.

theoryshaw (commented):
Thanks!  will try.

theoryshaw:
I think we got it now. :)
much more elegant.

brunoperdigao:
@theoryshaw Nice!
This implementation has one limitation though. In cases where the relating object have more than one array, it will only copy the attributes of the first array. Not a huge problem, but it's something to keep in mind for future improvements.

PR #4745, opened 2024-05-26
Keeps selected objects selected when switching from one drawing to an…

theoryshaw:
…other


theoryshaw:
proposed: https://community.osarch.org/discussion/2179/not-unselecting-objects-when-switching-from-one-drawing-to-the-other#latest

PR #4749, opened 2024-05-27
[BBIM] Add preference option for default startup workspace, remove ambiguous wording

### LINKED ISSUE #4036: [BlenderBIM] feature request: setting option to disable BIM workspace as default
The BlenderBIM add-on introduces a new dedicated `BIM` workspace, which is activated by default. Meaning, on startup of **any** `.blend` file, the workspace `BIM` is active. This is kind of inconvinient if I want to use Blender for other stuff than BIM/IFC work.

Is there a way to add an add-on setting option to deactivate this default switching to the `BIM` tab?

dirkolbrich:
This PR adds an option to _not_ switch to the BIM workspace on startup. By default the BIM workspace is activated on Blender startup, so nothing changes for other users, but gives an option to stay on the default Blender startup workspace „Layout“.

Closes #4036

- Remove ambiguous „should“ wording from preference settings

That is the nice thing with computers, the binary 0 or 1. Either a computer does a thing or it doesn’t. Everything else is an error and can be treated as such.

There are A LOT more functions and methods which start with `should_` in the BlenderBIM source code. All of these should ( <- see what I did here? Only applicable in case of a discussion) be changed to a clearer meaning. But this is a gargantuan task for a later refactoring.

### DIFF
--- a/src/blenderbim/blenderbim/bim/handler.py
+++ b/src/blenderbim/blenderbim/bim/handler.py
@@ -301,7 +301,8 @@ def load_post(scene):
 
     if bpy.context.preferences.addons["blenderbim"].preferences.setup_workspace:
         if "BIM" in bpy.data.workspaces:
-            bpy.context.window.workspace = bpy.data.workspaces["BIM"]
+            if bpy.context.preferences.addons["blenderbim"].preferences.switch_workspace:
+                bpy.context.window.workspace = bpy.data.workspaces["BIM"]
         else:
             bpy.ops.workspace.append_activate(idname="BIM", filepath=os.path.join(cwd, "data", "workspace.blend"))
 
--- a/src/blenderbim/blenderbim/bim/handler.py
+++ b/src/blenderbim/blenderbim/bim/handler.py
@@ -296,12 +296,12 @@ def load_post(scene):
     ifcopenshell.api.owner.settings.get_application = get_application
     AuthoringData.type_thumbnails = {}
 
-    if not bpy.context.preferences.addons["blenderbim"].preferences.setup_toolbar:
+    if not bpy.context.preferences.addons["blenderbim"].preferences.should_setup_toolbar:
         tool.Blender.unregister_toolbar()
 
-    if bpy.context.preferences.addons["blenderbim"].preferences.setup_workspace:
+    if bpy.context.preferences.addons["blenderbim"].preferences.should_setup_workspace:
         if "BIM" in bpy.data.workspaces:
-            if bpy.context.preferences.addons["blenderbim"].preferences.switch_workspace:
+            if bpy.context.preferences.addons["blenderbim"].preferences.activate_workspace:
                 bpy.context.window.workspace = bpy.data.workspaces["BIM"]
         else:
             bpy.ops.workspace.append_activate(idname="BIM", filepath=os.path.join(cwd, "data", "workspace.blend"))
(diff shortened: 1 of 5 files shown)


dirkolbrich:
Would 'activate_BIM_workspace' be a better name for the variable?

I read the referenced https://www.rfc-editor.org/rfc/rfc2119 more in the form of how to use the word `SHOULD` in the context of documentation or long form text. IMHO this word is not suitable for method names as it dilutes meaning and gives functions an ambiguous state. 

As for the `should_` prefix I can remove the change from this PR and add a PR for a project wide refactor.

Moult:
Yes, please remove the should refactor from this PR and then its good to merge.

Then, let's agree on a prefix that can be consistently used, is accurate, and indicates booleaness, for a global refactor as a separate PR :)

I like "should" but happy to change. Kinda meta but it sounds like software doesn't always behave the way it should ;) I don't like "will" because it sounds like it will happen in the future, not now. I don't like "must" because it sounds as though it indicates there is no choice but for it to almost happen (i.e. not user configurable). "Can" makes no sense to me. "May" sounds extremely ambiguous, probably the reaction you has to "should".

I personally prefer "enable" instead of "activate". Enable is shorter and is more commonly used (e.g. in CLI args).

Any other words to consider?

dirkolbrich:
Thank you for the merge.

As regarding for your ides of using a prefix for variables, I think there should not be used any prefix with an unclear meaning at all.

All of the words `may`, `can`, `should` and even `must` express possibilities and uncertainty. These require decision which is a human thing, not for computers. If the name of a command is `should_set_workspace` who is it to decide if the action is to be performed? Certaintly not the computer thing. The computer needs be told with a clear command in the form of `set_workspace` what to do. 

These words are even in such a way ambiguous and uncertain, that they needed a RFC to define their meaning in a clear way -> see https://www.rfc-editor.org/rfc/rfc2119

If you want to express booleaness, which is a clear distinction between true vs. false, using any prefix with an inherently uncertain meaning like `should` is not a good idea. Because it does not indicate a clear true or false.

My view on naming is this:

Methods and functions naming should express intent, like issuing a command (to the program) in the form of „do something“. Examples are calculate_pi(), delete_data(), reorder_list(). If this command fails for whatever reason, the function must return with an error.

Variables naming should express data or state. For booleans the name must express a clear distinction of true vs. false like `setting_is_enabled`. Preference settings as boolean variables are a mix of both intent and state, so a good name would be `enable_setting` or `activate_function`.

So, for a refactor we maybe can agree on using `enable` or `activate`. `enable` if a future option to use some functionality is allowed. `activate` when some functionality will definitely be executed. But there may be a lot of other (now) ambiguous variable names, which need to be decided on case by case. Here it would be best to split the refactor in smaller commit to discuss these changes individually.

Moult:
Completely agree on methods and functions being named after intent, so "verbs" are a great way to do that.

I wasn't too stressed about the meaning of "should" personally, but I understand your point and happy to refactor if you feel strongly about it.

I think we can agree on `enable`. To me, `activate` reads more like a verb (i.e. should it hold a function?), so I'd prefer not to use it.

I'm personally OK with a bulk sed replace to replace `should_` with `enable_` and see what happens, then over time some of the more suspicious usages can be tweaked.

BTW you will also come across other boolean prefixes used here and there in the code, like `has_` or `is_`. Just a heads up.

dirkolbrich:
The prefixes `has_` and `is_` are actually perfect as they clearly indicate a definite state.

As for `enable` and `activate` I see a very fine distinction between them. `enable` allows for a future optionality to do some action, while `activate` is a finite way to start some action. But anyway, these are kind of hair-splittings in this discussion. We will see what to choose if it comes to the actual refactoring.

Well, this all proves the old programming wisdom true again - "Naming is hard."

PR #4777, opened 2024-05-31
[BBIM docs] change color of ui breadcrumb on dark background

dirkolbrich:
Make the breadcrumb text better readable on dark background.
ToDo: find a way to change the color of the SVG symbols as well.

Before:
BlenderBIM-docs_breadcrumb_before

After commit:
BlenderBIM-docs_breadcrumb_after

### DIFF
--- a/src/blenderbim/docs/_static/custom.css
+++ b/src/blenderbim/docs/_static/custom.css
@@ -63,7 +63,7 @@ section img {
     background-color: var(--color-admonition-title-background);
     border-radius: 5px;
     font-style: italic;
-    color: var(--color-content-foreground);
+    color: var(--color-admonition-text);
 }
 img.icon {
     width: auto;
--- a/src/blenderbim/docs/_static/custom.css
+++ b/src/blenderbim/docs/_static/custom.css
@@ -63,7 +63,7 @@ section img {
     background-color: var(--color-admonition-title-background);
     border-radius: 5px;
     font-style: italic;
-    color: var(--color-admonition-title);
+    color: var(--color-content-foreground);
 }
 img.icon {
     width: auto;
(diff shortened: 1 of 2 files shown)


Moult:
There is actually a dark mode vs light mode override in the sphinx conf, can you check that out? Sorry afk right now.

Moult:
Yep, check src/blenderbim/docs/conf and redefine the CSS variable for dark mode there.

dirkolbrich:
There are no overrides in the `conf.py` file for the breadcrumb style (called Admonitions in the furo theme).

The current `custom.css` uses the same color for the border, text and icons (here hardcoded into the SVG file).

https://github.com/IfcOpenShell/IfcOpenShell/blob/a5d123fe833da1e37f77c3e557811ff8bf2d946d/src/blenderbim/docs/_static/custom.css#L61-L67

There is no separate css class defined for admonitions text color and border within the Furo theme. If I  override the `color-admonition-title` class within the `conference.py` it also changes the color border of the breadcrumb. That's not what I wanted to achieve.

Moult:
Yep, I'm saying rather than change it globally, to use a CSS variable to define the separate light and dark colours. That way light can stay as the purple colour. The conf file can then define the CSS variable.

dirkolbrich:
Please reopen this PR, so Github adds the follow-up commits and makes them visible in this PR.

I think you don't understand. I just wanted to change the text color, not the border color as well. 

If the `volor-admonition-title` attribute is overridden in `conf.py`, it also changes the border color. There needs to be a distinction between text and border within `custom.css`. I added a new CSS class to `custom.css` just for the text and added color overrides to `conf.py`

Look at the other default admonitions of the Furo theme, they also use a different color class for the text and the border.

Moult:
Cheers, reopened PR. I understand what you're saying. What I'm suggesting is to define a _new_ or different CSS variable, and then define it in the conf.py for both light and dark.

dirkolbrich:
> What I'm suggesting is to define a new or different CSS variable, and then define it in the conf.py for both light and dark.

Did that with the additional commit.

Moult:
Beautiful! Thank you so much! Sorry for the hasty close!

PR #4788, opened 2024-06-01
keep font scaled correctly when changing drawing width and height

theoryshaw:
Currently when changing the width/height of the drawing camera it scales the font this way...

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/317c1a85-76b9-4e7d-89e7-2ffae747b2f8

I might be missing something, but shouldn't it scale the font this way...

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/600ec390-c524-44e1-99f2-d0edbc4f712b


Moult:
The video looks good, but the code isn't right, "get_camera_width_mm" shouldn't be changed the way it is changed, perhaps either a new function, or the way the function is called should be changed.

PR #4792, opened 2024-06-02
shift + click with bim.activate_drawing switches view without turning…

theoryshaw:
… off/off objects in scene

Proposed to quickly navigate to drawing views without turning on and off objects. 

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/a973de69-c858-42e5-b9e8-98084354d183


theoryshaw:
Proposed here, as well: https://community.osarch.org/discussion/2196/quick-way-to-switch-between-drawing-views#latest

theoryshaw:
Implemented via https://github.com/IfcOpenShell/IfcOpenShell/commit/e0f33a56ec5683a03f187a933f8f75afd6de7c37

PR #4841, opened 2024-06-11
Update README.md about extend express usage

shungeer:
1. C++ code generation by extend express.
2. python rule code generation by extend express.


Moult:
Thank you so much!

PR #4885, opened 2024-06-18
Add Web module/tool/core for Web UI

Ziad-I:
- Add a new Web module with its corresponding operators/UI/data
- Add a new Web tool/core to start/close/connect/disconnect to the Web UI


Moult:
Merged and thanks!

PR #4935, opened 2024-06-27
Update proposalcsv4d2ifc.py

DimitriosThe:
Please correct me if I am wrong:

while trying to create relationships I saw that in def create_rel_sequence (line 194) there is a parameter missing which should reflect the "rel_type", otherwise and according to "assign_sequence.py" the sequence_type remains  def assign_sequence(
    file: ifcopenshell.file,
    relating_process: ifcopenshell.entity_instance,
    related_process: ifcopenshell.entity_instance,
    sequence_type: str = "FINISH_START",


Moult:
You're absolutely right looks like it was forgotten. Thank you so much!

Issue #4351, opened 2024-02-22
Parsing IFC4.x and IFC2x3 in same executable


jonganej:
I'm trying to parse IFCs of different schemas without having to use preprocessor directives and knowing what schema each IFC uses and I haven't found out how to do that yet. Is that possible to begin with?

I stumbled across a really old issue [ #49](https://github.com/IfcOpenShell/IfcOpenShell/issues/49) and wondered if that was implemented in v0.7.0. If so, is there an example of how to use it? So far, my only understanding of how to parse IFC is from `src/examples/IfcParseExamples.cpp` and that requires knowing what schema the file in question has.

Otherwise, thanks for your effort on this extensive set of IFC tools @aothms!

aothms:
See the *Writing schema agnostic code* section at http://blog.ifcopenshell.org/2019/12/v060.html

Moult:
@jonganej would you also be interested in helping contribute to a Hello World beginners C++ tutorial in the docs?

jonganej:
> @jonganej would you also be interested in helping contribute to a Hello World beginners C++ tutorial in the docs?

Sure, I would be interested. What parts of the IFCOpenShell C++ API specifically would you see getting covered? As suggested by @aothms, I managed to use templates for schema agnostic IFC parsing and got some learnings out of it. Might be helpful for others.

Moult:
@jonganej anything is better than nothing :) As you can see here we currently have only 4 pages: https://blenderbim.org/docs-python/ifcopenshell.html

PR #4965, opened 2024-07-02
Documentation: Added getting started documentation for ifcopenshell core

jonganej:
As suggested in #4351, here is an initial version of documentation for the core module of IfcOpenShell.

Please do let me know if there are things I should add or how I could improve this.

### DIFF
--- a/src/ifcopenshell-python/docs/ifcopenshell/getting_started.rst
+++ b/src/ifcopenshell-python/docs/ifcopenshell/getting_started.rst
@@ -1,0 +1,109 @@
+Getting Started with IFC parsing
+================================
+
+The basis of all parsing and getting information from the IFC starts with obtaining a ``IfcParse::IfcFile`` object and validating that it is good for use.
+
+.. code-block:: c++
+
+    std::string input_file_path = "/path/to/my/model.ifc";
+    IfcParse::IfcFile file(input_file_path);
+    if (!file.good()) {
+      std::cerr << "Unable to parse .ifc file" << std::endl;
+      return 1;
+    }
+
+Schema-agnostic parsing of IFCs
+-------------------------------
+
+It is advisable to design your programme in a schema-agnostic fashion to be able to process all schema versions of input IFCs. 
+While different options are presented `here <http://blog.ifcopenshell.org/2019/12/v060.html>`__, the most comfortable approach for modern C++ developers
+would be with the use of templates as shown below.
+
+.. code-block:: c++
+    
+    // necessary includes
+
+    // Include all possible schema types that could be parsed
+    #include "ifcparse/Ifc2x3.h"
+    #include "ifcparse/Ifc4.h"
+    #include "ifcparse/Ifc4x3_add2.h"
+
+    template<typename Schema>
+    void parseIfc(IfcParse::IfcFile &file) {
+        const typename Schema::IfcProduct::list::ptr elements = file.instances_by_type<typename Schema::IfcProduct>();
+        for (typename Schema::IfcProduct::list::it it = elements->begin();
+            it != elements->end(); ++it) {
+            Schema::IfcProduct *ifcProduct = *it;
+            // TODO: Do something with ifcProduct
+        }
+    }
+    //
+    int main(int argc, char* argv[]) {
+        // file of IfcParse::IfcFile previously defined
+        auto schema_version = file.schema()->name();
+        schema_version = schema_version.substr(3);
+        std::transform(schema_version.begin(), schema_version.end(), schema_version.begin(), [](unsigned char c) { return std::tolower(c); });
+        if (schema_version == "4x3_rc2") {
+            parseIfc<Ifc4x3_rc2>(file);
+        }
+        else if (schema_version == "4") {
+            parseIfc<Ifc4>(file);
+        }
+        else if (schema_version == "2x3") {
+            parseIfc<Ifc2x3>(file);
+        }
+        // TODO: Handle more cases of IFC schema versions here
+    }
+
+Reading out attributes of an IfcProduct
+---------------------------------------
+
+The properties of an ``IfcProduct``, and by extension any derived class, can be read as by calling the method of the same name, e.g. ``GlobalId()``, ``Name()``. 
+Note that optional properties like name, long name, or description, among others, are wrapped with a ``boost::optional`` (`documentation <https://www.boost.org/doc/libs/1_84_0/libs/optional/doc/html/index.html>`__) container object.
+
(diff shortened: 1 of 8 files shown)


src/ifcopenshell-python/docs/ifcopenshell/getting_started.rst
```diff
@@ -0,0 +1,109 @@
...
+            // TODO: Do something with ifcProduct
+        }
+    }
+    //
+    int main(int argc, char* argv[]) {
+        // file of IfcParse::IfcFile previously defined
+        auto schema_version = file.schema()->name();
+        schema_version = schema_version.substr(3);
+        std::transform(schema_version.begin(), schema_version.end(), schema_version.begin(), [](unsigned char c) { return std::tolower(c); });
+        if (schema_version == "4x3_rc2") {
```
aothms, line 46: Do you think we should rewrite this using the Boost preprocessor mechanism (see below). Either way is a bit unfortunate, you either have a a bunch of if statements, or a incomprehensible bunch of preprocessor magic. I do think that the 2nd approach might in the end result in a better developer experience.

I think this would be the time to decide, as we're setting kind of the basis for the C++ docs here.

https://github.com/IfcOpenShell/IfcOpenShell/blob/9485190604bb3fed6f45199684609667bc2ed1df/src/examples/IfcParseExamples.cpp#L217-L227

src/ifcopenshell-python/docs/ifcopenshell/getting_started.rst
```diff
@@ -0,0 +1,109 @@
...
+            parseIfc<Ifc2x3>(file);
+        }
+        // TODO: Handle more cases of IFC schema versions here
+    }
+
+Reading out attributes of an IfcProduct
+---------------------------------------
+
+The properties of an ``IfcProduct``, and by extension any derived class, can be read as by calling the method of the same name, e.g. ``GlobalId()``, ``Name()``. 
+Note that optional properties like name, long name, or description, among others, are wrapped with a ``boost::optional`` (`documentation <https://www.boost.org/doc/libs/1_84_0/libs/optional/doc/html/index.html>`__) container object.
```
aothms, line 62: Unfortunately this is still not consistent in the C++ side.

```suggestion
Note that optional properties like name, long name, or description, among others, are wrapped with a ``boost::optional`` (`documentation <https://www.boost.org/doc/libs/1_84_0/libs/optional/doc/html/index.html>`__) container object. Except for pointer types such as IfcRoot::OwnerHistory, here the developer is still responsible for performing comparison to `nullptr` before using the value.
```

src/ifcopenshell-python/docs/ifcopenshell/getting_started.rst
```diff
@@ -0,0 +1,109 @@
...
+                     typename std::vector<typename Schema::IfcPropertySet *> &property_sets) {
+        const auto &is_defined_by = obj.IsDefinedBy();
+        if (is_defined_by == nullptr)
+            return;
+        std::vector<typename Schema::IfcRelDefines *> rels(is_defined_by->size());
+        typename std::vector<typename Schema::IfcRelDefines *>::iterator it = std::copy_if(
+            is_defined_by->begin(), is_defined_by->end(), rels.begin(),
+            [](const typename Schema::IfcRelDefines *x) {
+                if (x == nullptr) return false;
+                const typename Schema::IfcRelDefinesByProperties *defines_by_properties =
```
aothms, line 86: I would use `x->template as<typename Schema::IfcRelDefinesByProperties>()` here.

Not sure if `template as` is required or `as` is sufficient. We need to check with gcc/clang.

aothms: Same for below. Since we're setting the basis here, better do as<...> instead of dynamic_cast<...>. Functionally it's the same, but we might implement it differently later on.

src/ifcopenshell-python/docs/ifcopenshell/getting_started.rst
```diff
@@ -0,0 +1,109 @@
...
+                        return dynamic_cast< typename Schema::IfcPropertySet*>(defines_by_properties->RelatingPropertyDefinition());
+                        });
+    }
+
+Defensive programming with IfcOpenshell
+---------------------------------------
+
+The need for (down-)casting object pointers when accessing various properties in an IFC entity is evident from the previous code sample as the methods and properties usually 
+return the abstract class of the entity. It is hence important to check for ``nullptr`` when performing such casts. 
+The existence of certain fields and properties should also be checked.
```
aothms, line 109:
```suggestion
The existence of certain fields and properties should also be checked.  Also note that IfcOpenShell has as of now not been tested explicitly against malicious inputs. Schema validation (the correctness of attribute types and conformance to schema where rules) can currently only be assessed in Python (using `ifcopenshell.validate --rules`).
```

aothms (changes requested):
Super awesome. Thanks for taking the time to write this down. See my comments inline.

src/ifcopenshell-python/docs/ifcopenshell/getting_started.rst
jonganej: `template as` is necessary. `as` doesn't seem to compile

src/ifcopenshell-python/docs/ifcopenshell/getting_started.rst
jonganej: Hmm, I've never used these Boost preprocessor macros before. I see the advantage of macros, but for both cases, one still has to explicitly enumerate all the available schemas either with a `#define SCHEMA_SEQ` or a `else if `. In my production code, I chain `if`, `else if` and end with a `else` to catch the case where the input file does not match any of those schemas. Would that be possible too with the preprocessor macros?

jonganej: Ok, I'm using preprocessor macros now in an `if-else if` fashion with a final `else` to ensure that one can fail explicitly if the schema is not recognized. We still have the enumerated `#define` for the schemas, but at least it's clearer now. I tested this approach on our productive use-case and it works.

Moult:
Good to merge?

jonganej:
I would say so. I’ve done the necessary changes @aothms suggested. Was wondering if he or one of the other maintainers wanted to take a look at it.

Moult:
I'll merge it and if there are more improvements it can be fixed over time :) Thank you so much!

aothms:
Thanks again. Sorry, this got buried in my pile of holiday emails.

PR #4976, opened 2024-07-04
add choco_release git debug info

ay-ex:
adding some git debug info, in order to understand why no release is created.


ay-ex:
I think I found it: the git checkout action pulled the repo without the tags! 😑 
no wonder the script cannot find them then. 🙂

PR #4990, opened 2024-07-08
fix git clone depth

ay-ex:
@Moult : ok from the debug info it shows, I was on right right track. 
but now it needs to pull the git repo not only with the latest commit, 
but with depth=0 to also show the git commit from the day before. 🙂

### DIFF
--- a/.github/workflows/ci-blenderbim-choco.yml
+++ b/.github/workflows/ci-blenderbim-choco.yml
@@ -43,6 +43,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-tags: true
+          fetch-depth: 0
       - uses: actions/setup-python@v5 # https://github.com/actions/setup-python
         with:
           python-version: '3.12.4' # Version range or exact version of a Python version to use, using SemVer's version range syntax


Vanuan:
depth 0 will pull all the branches and history. Wouldn't last 50 commits be enough? It appears IOS gets 10-20 commits per day at max

ay-ex:
@Vanuan I completely agree that pulling a limited amount of commits would be preferrable. 
I pondered as well on that question. 🤔  however this this would create a parameter 
estimated_max_commits_per_day, that would need to be checked, maintained and possibly adjusted.
for now I would run it with depth 0, to see if choco builds can be re-enabled with that.
if that works out as expected, I plan on doin some cleanup on the code around choco releases, 
which also includes retrieving the information of the release tags via webrequest to github - 
with that we could set the depth back to the default minimum.

Vanuan:
@ay-ex Is there a master bug that describes the underlying issue you're trying to fix? I assume choco is a package manager for Windows? Is it for some automation or for end users?

ay-ex:
@Vanuan I fear I do not completely understand the question, maybe you can rephrase? what do you mean with master bug?

the problem the above pull request solved, is getting the blenderbim-nightly chocolate builds to run again: 
https://community.chocolatey.org/packages/blenderbim-nightly 
which did not build for a couple of weeks as 
* upstream blender  repo changed info on used python version
* blenderbim changed its tag version naming

as you can see from the version on the linked page, they are now up again since today. 🎉  🙂 

" I assume choco is a package manager for Windows?"
yes exactly. this allows to install newest blender, 
together with newest blenderbim (and activated) with one cli command:
`choco install blenderbim-nightly --pre`
and update them super conveniently with just `choco upgrade all`  ☺️ 

"Is it for some automation and IT operations or for end users?"
actually for both, if they like this way on installing applications and keeping the up to date conveniently.
the standard setup assumes admin privileges for choco though.

Vanuan:
Got it. I just saw the discussion on this was going for a while in the chat, but I couldn't find anything related in the issue tracker.

PR #4997, opened 2024-07-09
Select multiple files for testing and multiple test cases at once in the IFC Tester Panel

BlenderDefender:
This allows to select multiple IFC files and multiple test cases at once.
To enhance the UI and avoid confusion, the "ANIMATABLE" option has been removed from multiple test settings.


Moult:
Looks great! Is there a reason this is still marked as draft instead of ready for merge?

BTW very small nitpick, can you change the label text to be something like `IFC File(s)` (i.e. add `(s)`) to give a hint to users that they can now select multiple?

BlenderDefender:
Sure, I'll adapt that label. And yes, this PR is intentionally marked as draft, because it's lacking the actual implementation in the `bim.execute_ifc_tester` operator. The operator would still only use one file, I'll implement that next and mark the PR as ready for review as soon as that's done

BlenderDefender:
Alright, this feature should work now 👍 
I've tested this on my side, generating the reports (and opening them) work great, the feedback UI could use some more work though, as it currently just shows every item without context (which test was run on which file). For now, I'll call it a concern for another PR, as discussed previously. A tree layout might work well there, see the "mockup" below:

grafik

Moult:
Awesome! Is there a reason you've removed `bl_options = {"REGISTER", "UNDO"}` ?

BlenderDefender:
@Moult Yes, I've removed that because it's inherited from the parent class, see this line here: https://github.com/IfcOpenShell/IfcOpenShell/pull/4997/files#diff-050f51a5e11de5cb02270a2f7b47fb194c1e30262808d8980b58e47e7381fe9cR146

PR #5021, opened 2024-07-16
BBIM remove Blender manifest file

dfelinto:
BlenderBIM is not working yet for Blender 4.2. Keeping the manifest around means the add-on would try to be installed (by Blender) as an extension. Removing the add-on makes it to install correctly (as legacy add-on).

This should give the developers more time to decide how to tackle the conversion to extension.

Note: blender_version_min should have been 4.2 since extensions were not existent before that, and this range is only looked at in relation to extensions.


Andrej730:
> Note: blender_version_min should have been 4.2 since extensions were not existent before that, and this range is only looked at in relation to extensions.

Thanks, fixed!

> BlenderBIM is not working yet for Blender 4.2. Keeping the manifest around means the add-on would try to be installed (by Blender) as an extension. Removing the add-on makes it to install correctly (as legacy add-on).

At main branch BBIM currently does work as an extension but it violates lots of policies since 1) it is using a hack to keep `blenderbim` available instead of `bl_ext.user_default.blenderbim`; 2) it's messing up with `sys.path` to make it's dependencies available.

We're currently working on it on a separate branch, got sorted out all the dependencies stuff today, the main thing we're currently got stuck on is a way to make sure `blenderbim` is still available for import without `bl_ext.user_default.blenderbim` and with just  `import blenderbim`  (to avoid relative imports and make `blenderbim` generally available for user scripts). 

Is there some known good practice how to accomplish this?

What we've got so far, we came up with a sort of a stub-file that's named `blenderbim`, it's supplied as a wheel and it redirects all `blenderbim.xxx.yyy` imports to `bl_ext.user_default.blenderbim.xxx.yyy`. It does work and we don't need a hack I've mentioned above but we still get lots of policy violation warnings for the extension.


Here's an example of a simple blender extension with a similar stub file to reproduce this, if this helps: [test_addon.zip](https://github.com/user-attachments/files/16271142/test_addon.zip)

Extension is registered just fine but then we get a warning (in case of blenderbim it's a couple hundreds lines of warnings 😄 ):
image



example stub-file:
```python
import bpy
import importlib

EXTENSION = None


def find_extension():
    global EXTENSION
    if EXTENSION is not None:
        return EXTENSION
    for package_name in bpy.context.preferences.addons.keys():
        if package_name.endswith(".test_addon"):
            EXTENSION = importlib.import_module(package_name)
            return EXTENSION
    raise Exception("Extension 'test_addon' is not found.")


def __getattr__(name: str):
    ext = find_extension()
    return getattr(ext, name)
```

dfelinto:
The wheel stub is indeed the recommended way.

> example stub-file:

You can replace your for package  loop with __package__: [documentation](https://docs.blender.org/manual/en/latest/advanced/extensions/addons.html#user-preferences-and-package)

>  but we still get lots of policy violation warnings for the extension.

This is strange and could be investigated on our end. I believe the warnings are supposed to happen only when the add-on meddle with sys.path.

Could you please report this as a bug on developer.blender.org? Feel free to mention that I asked to be reported there and you can tag me and Campbell (cc @dfelinto / @ideasman42).

PR #5036, opened 2024-07-18
ci: Add black formatter to `ci.yml`

BlenderDefender:
Check, whether the files are formatted properly using the black formatter when changes are pushed or pull requests are opened.


Andrej730:
Hi! `ci.yml` is triggered only on the changes in ifcopenshell and in some related packages. I think we can create a separate workflow for this so it can cover more packages (e.g. including blenderbim).

BlenderDefender:
How should that seperate workflow be called? Something like `linter.yml`?

Andrej730:
> How should that seperate workflow be called? Something like `linter.yml`?

We can call it something like `ci-black-formatting.yml`/`ci-black-formatter.yml` so it won't get confused with pylint.

BlenderDefender:
Alright, done 👍

Andrej730:
Is there a way to set it up so it would check either only changes in the last commit or files affected by the last commit? 
We have a huge repo and this way we'll be able to properly format it in more gradual way 😁

Moult:
I evilly ran black on everything and now we pass :D

MUAHAHAHAHAHA

Andrej730:
rip git blame

PR #5043, opened 2024-07-19
New small function for calculating the length of all edges in a given geometry.

VDobranov:


src/ifcopenshell-python/ifcopenshell/util/shape.py
```diff
@@ -730,3 +730,16 @@ def get_extrusions(element: ifcopenshell.entity_instance) -> list[ifcopenshell.e
...
+    """Calculates the total length of edges in a given geometry.
+
+    :param geometry: Geometry output calculated by IfcOpenShell
+    :type geometry: geometry
+    :return: The total length of all edges in the geometry.
+    :rtype: float
+    """
+    vertices = get_vertices(geometry)
+    edges = get_edges(geometry)
+    return sum([np.linalg.norm(vertices[e[0]] - vertices[e[1]]) for e in edges])
```
aothms, line 745: You can probably resolve this all in numpy for some insignificant speedup. Untested:

```suggestion
    edges = vertices[get_edges(geometry)]
    return np.linalg.norm(vertices[:, 1] - vertices[:, 0], axis=1).sum()
```

VDobranov: Did you mean this?:
`vertices = vertices[get_edges(geometry)]`
instead of this?:
`edges = vertices[get_edges(geometry)]`

VDobranov, line 745: And it will look like this:
```suggestion
    vertices = vertices[get_edges(geometry)]
    return np.linalg.norm(vertices[:, 1] - vertices[:, 0], axis=1).sum()
```

aothms: Ah yes that's what I meant. Thanks

Issue #4721, opened 2024-05-23
IfcPatch SplitByBuildingStorey


BIMvoice:
I'm referring to our chat we had earlier. There are some issues with this patch recipe and you also mention that the code is too old, and it should be re-written.

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/139251053/db87bb69-2f79-458d-9386-fa4deb7b594a)

PR #5082, opened 2024-07-25
feat: Update the Patch Panel UI to use a dropdown for the ConvertLengthUnit, Migrate and ResetAbsoluteCoordinates

BlenderDefender:
Writing this, I stumbled upon two questions:
1. The `set_value` method in `./src/blenderbim/blenderbim/bim/prop.py`, line 294 seems to always change the type to string, for anything that's not a boolean, float or integer, so it'll never turn out to be an enum property. Is this a bug that should be fixed?
2. How should the possible units be ordered in `ConvertLengthUnit.py`? Right now, they're sorted alphabetically (except the imperial units), but it might be more sense to sort them numerically, so kilometer comes after meter, ... Also, are there any imperial units I'm missing here?


BlenderDefender:
Looking at #4721, does it make sense to write a function that transforms a variable name into a more readable name? Something along `"".replace("_", " ").capitalize()` or so?

Moult:
> The set_value method in ./src/blenderbim/blenderbim/bim/prop.py, line 294 seems to always change the type to string

I don't think this is a bug but it is a badly named function. Apart from IfcPatch, this function is only used in one other situation in `tool/pset.py` - it sets _both_ a value and autodetects an appropriate data type where there is no data type available. This function has multiple responsibilities and is misleading.

What you've done in the meantime of setting enum_value directly sounds appropriate and exactly what the rest of the codebase does, so keep your code, and when the pset module is due for an upgrade soon (to handle more property types, which would touch that part of the code) this'll be reviewed and cleaned.

> it might be more sense to sort them numerically, so kilometer comes after meter

Yes, this sounds correct. The imperial units sound enough to me.

> does it make sense to write a function that transforms a variable name into a more readable name?

Go for it :)

PR #5121, opened 2024-08-01
improve json lib discovery for cmake

autra:
# Description

While building the project, I noticed that some library discovery can be improved. Here is a fix for the json lib, basically telling cmake in which subfolder it needs to search for the hpp.

If you guys are interested, there are probably other possible fixes in the same spirit (I have one very similar for opencascade, and a wip for opencollada, which is less easy).


autra:
Which branch should I base my PR onto?

Moult:
Basing it on 0.8.0 is correct :)

autra:
Branch rebased :-)

There might be some -DJSON_INCLUDE_DIR that are now unneeded. Do you want me to remove them?

Moult:
This is something for @aothms to review, sorry, I don't know much about the C++ build.

aothms:
There's a lot of things that can be improved in the cmake script. Contributions are very welcome!

I must say I don't fully understand the distinction between:

    ${JSON_INCLUDE_DIR}/nlohmann

and

    ${JSON_INCLUDE_DIR} PATH_SUFFIXES "nlohmann"

I also don't really understand why you do

    find_path(json_header_path "json.hpp" ${JSON_INCLUDE_DIR} PATH_SUFFIXES "nlohmann")
    set(JSON_INCLUDE_DIR ${json_header_path})

instead of directly

    find_path(JSON_INCLUDE_DIR "json.hpp" ${JSON_INCLUDE_DIR} PATH_SUFFIXES "nlohmann")

Or is it not possible in cmake to have the same variable participate as input and output simultaneously? I have no idea.

The test passes so I guess it's safe to merge this in.

autra:
> 
> I must say I don't fully understand the distinction between:
> 
> ${JSON_INCLUDE_DIR}/nlohmann
> 
> and
> 
> ${JSON_INCLUDE_DIR} PATH_SUFFIXES "nlohmann"

The first version looks in ${JSON_INCLUDE_DIR}/nlohmann, then all other default lib dir (/usr/include etc...) *without* the nlohmann suffix.

The latter version also append the prefix to the default search path, which must be done since this lib is packaged this way.

> I also don't really understand why you do
> 
> find_path(json_header_path "json.hpp" ${JSON_INCLUDE_DIR} PATH_SUFFIXES "nlohmann")
> set(JSON_INCLUDE_DIR ${json_header_path})
> 
> instead of directly
> 
> find_path(JSON_INCLUDE_DIR "json.hpp" ${JSON_INCLUDE_DIR} PATH_SUFFIXES "nlohmann")

It was just to avoid overwriting what the user provided (since `JSON_INCLUDE_DIR` is a user defined variable) which can be surprising. I generally prefer treating them as if they were immutable, like function parameters for instance.

That being all said, I just read this json lib doc, and they say

> To use this library from a CMake project, you can locate it directly with find_package() and use the namespaced imported target from the generated package configuration:
> # CMakeLists.txt
> ```
> find_package(nlohmann_json 3.2.0 REQUIRED)
> ```

I'm gonna test that and maybe submit it, with other packaging improvements.

aothms:
> The latter version also append the prefix to the default search path

Thanks for clarifying :)

RickBrice:
Did this break my build?

https://github.com/IfcOpenShell/IfcOpenShell/issues/5134

Issue #5033, opened 2024-07-18
Bugs in light simulation


manuvarkey:
I was testing out the light simulation feature newly merged in BlenderBIM. Following points require correction.

1. When selecting a JSON path error is shown in linux showing both source and destination file paths used by `shutil.copy` as same. Source of the error seems to be the below line which only works on windows.
https://github.com/IfcOpenShell/IfcOpenShell/blob/9bf68db0806e451ae927ccb8a64a6c8ecbc13622/src/blenderbim/blenderbim/bim/module/light/operator.py#L156
On Linux this requires modification to `json_dest_path = os.path.join(blend_file_dir, json_file_path.split("/")[-1])` to work correctly. Probably it may be better to use a platform agnostic method to achieve the logic.

2. When path to the project directory contains white spaces, following error is reported when running 'Radiance Render'. It may be needed to escape the white spaces in the file paths before sub-process is called.

![Screenshot from 2024-07-18 11-10-35](https://github.com/user-attachments/assets/210b6bab-407a-4e81-a9b6-32ec73d532d5)

Further in the interface I find the following observations for improvement.

![Screenshot from 2024-07-18 11-14-43](https://github.com/user-attachments/assets/532c1c6c-6d7d-4e90-b3d5-8f5f721d8dfd)

1. I would question why the `IFC filename` field is required at all since the information is available for an active project. 
2. Also the `JSON file` field cannot be understood by the lay user. It should be possible to render out the project even if this field is missing by using the default parameters. Also better documentation is required in the tool tip explaining what is the JSON file.
3. I would rename the `Export to OBJ` button to `Export Geometry for Simulation` and disable the `Radiance Render` button until geometry is exported.

Platform: `Ubuntu 22.04`
Blender: `4.2`
BlenderBIM: `0.7.10.240714`

@chiragsingh1711

chiragsingh1711:
Hey Manu,
thank you so much for pointing out these corrections.
In the next push, I'll make sure all these corrections are made and will
surely improve the interface as well.
Really appreciate your input.

chiragsingh1711:
Done @manuvarkey , I have pushed the changes https://github.com/IfcOpenShell/IfcOpenShell/commit/e88abd881dd9e582845d016f87e5faec1c4c0678

chiragsingh1711:
Please review and let me know if you find more bugs. @manuvarkey

manuvarkey:
[...]

PR #5160, opened 2024-08-08
UIList + SpectralDB 

chiragsingh1711:
1. Added a Json File having SpectralDB Materials. 
2. Implemented UIList: Lists down the materials to be mapped with. 
3. By default Not mapped material is white. 
4. Added a Drop down for category and Subcategory from where the user can select the material from spectralDB 
5. Corrected hardcoded render123.hdr as highlighted by Manu in #5033


Moult:
We'll probably ship the spectraldb as a dependency and not commit directly. Can you see if you can modify the Makefile for this? There are some examples already in there for Brick and jsgantt.

PR #5164, opened 2024-08-08
Update topic.py

ejs-ejs:
setters/getters for priority, labels, modified_author, modified_date, creation_date

### DIFF
--- a/src/bcf/bcf/v2/topic.py
+++ b/src/bcf/bcf/v2/topic.py
@@ -68,6 +68,56 @@ class TopicHandler:
         """Return the comments of the topic."""
         return self.markup.comment if self.markup else []
 
+        @property
+    def priority(self) -> mdl.Topic:
+        """Return the Priority object."""
+        return self.markup.topic.priority
+    
+    @priority.setter
+    def priority(self, value: str) -> None:
+        """Set Priority object."""
+        self.markup.topic.priority = value
+
+    @property
+    def labels(self) -> mdl.Topic:
+        """Return the Labels object."""
+        return self.markup.topic.labels
+    
+    @labels.setter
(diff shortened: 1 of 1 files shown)


Andrej730:
Sorry, didn't get any chance to understand the context of the changes yesterday.

If I understand it correctly now, this PR (and #5165) would just create 2 ways to do the same thing - e.g. now if you want to access/edit topic attributes you just do

```python
topic = bcfxml.get_topic(guid).topic
print(topic.priority)
topic.priority = "High"
```

And this PR adds other shortcut that just allows accessing/editing some attributes from the handler without accessing `.topic` first. But that would make things more confusing as it will raise the questions about the difference between those two methods and we would need to define each attribute twice (in the handler and in the markup) and maintain them.

```python
topic_handler = bcfxml.get_topic(guid)
print(topic_handler.priority)
topic_handler.priority = "High"
```

ejs-ejs:
> Sorry, didn't get any chance to understand the context of the changes yesterday.
> 
> If I understand it correctly now, this PR (and #5165) would just create 2 ways to do the same thing - e.g. now if you want to access/edit topic attributes you just do
> 
> ```python
> topic = bcfxml.get_topic(guid).topic
> print(topic.priority)
> topic.priority = "High"
> ```
> 
> And this PR adds other shortcut that just allows accessing/editing some attributes from the handler without accessing `.topic` first. But that would make things more confusing as it will raise the questions about the difference between those two methods and we would need to define each attribute twice (in the handler and in the markup) and maintain them.
> 
> ```python
> topic_handler = bcfxml.get_topic(guid)
> print(topic_handler.priority)
> topic_handler.priority = "High"
> ```

Got it.

Is there some existing way to access the viewpoint and snapshot information?

PR #5165, opened 2024-08-08
Update topic.py

ejs-ejs:
Getters/setters for creation_date, modified_date, modified_author, labels, priority,


src/bcf/bcf/v3/topic.py
```diff
@@ -65,6 +65,56 @@ def comments(self) -> list[mdl.Comment]:
         """Return the comments of the topic."""
         return self.topic.comments.comment if self.topic.comments else []
 
+        @property
```
Andrej730, line 68: Indentation issue, same in #5164

Issue #4954, opened 2024-07-01
Draw wall and wall polyline tool


Moult:
This is a good issue for a new developer and can be built isolated from BBIM at first.

After 5 years we don't have a tool where you can click a point, click another, and a wall appears between those two points (or click many points to create a polyline where walls are created and auto join to neighbouring walls etc). This is embarassing :)

This is a pure UI exercise to design an interface that allows this tracking of "has clicked point", "has clicked second point" to allow users to draw a single wall or chain of walls. There should be decorators that show that you are in "select point" mode, as well as angle snapping (with custom angles), snapping to existing elements in the model, auto join, snap to Z value of active container (walls are based on 2D axes), etc.

In theory, all the components are there to do this. We have functions for wall creation, wall extension, wall from 2 point, wall joins, etc. We also have decorators which can largely be copy pasted. We don't have any snapping code.

Anybody tackling this give me a buzz before starting so we can chat :)

hydromac:
It would be good if it was a tool not only for drawing walls, but also for any type of linear object, e.g. pipelines, HVAC ducts, electrical lines, etc. I suggest that already at this stage we consider that it should be a class extending the class of a given object with the possibility of drawing it either:
1. from point to point,
2. 2. after clicking a line/polyline;
; and depending on the type of object in which it is implemented, it should have, among others, the functions mentioned, such as for walls, snapping Z (wall height) to the ceiling or roof and automatic generation of a space object (ifcSpace) when walls with floor and ceiling form a closed body, in the case of pressure pipes, drawing them along the pipe axis, in the case of gravity flow pipes, the baseline is the inner bottom of the pipe and additionally drawing with a specified slope for a given standardized pipe diameter, etc. Such modified tools would be more user-friendly when modeling BIM objects using Bonsai.

Moult:
Indeed, it will be coded in a way that is generic and reusable to be then used across a range of tools (slabs, beams, roofs, measurement tools, MEP segments, space boundary drawings, etc) - drawing walls is just the first implementation.

Moult:
Closing as this is now implemented. Other separate bugs now track individual improvements.

PR #5172, opened 2024-08-10
Feat wall polyline tool

brunoperdigao:
This PR is a draft for inserting wall with a polyline tool: https://github.com/IfcOpenShell/IfcOpenShell/issues/4954

Instructions to use (there are a lot of options, so I recommend testing those in order to understand):
1. Using only the mouse:
 - Select the wall tool a make sure you have a wall type selected.
 - Press `shift+p` to start the tool. I'll get an input panel with values from `x`, `y`, `distance` and `angle`.
 - Click a few random points.
 - Press enter and the wall should be created.
 2. Using the keyboard:
 - Select the wall tool and press `shift+p`.
 - Press "TAB" to cycle through the input options. They should be highlighted in a blue color.
 - Choose the `x` and press 0. Choose `y` and press 0. Press ENTER and it should add a point in the origin.
 - Move your mouse. Notice it will snap into a few axis. Right now they are hard coded to be every 30 degrees angle.
 - When the mouse is over the axis that you want, press 7.5. You'll notice is goes directly to the `distance` input. Press enter and it will add a point which is 7.5m away from the origin.
 - Press enter an it should add a wall.
 3. Other explanations:
 - The points are always created in the XY plane with Z=0.
 - You can move the mouse to snap in objects in the scene.
 - Backspace removes the last point.
 - You can also snap in existing polyline points.
 - If you hold `shift` while over an axis it will lock the mouse to it.
 - To use the input panel you can press "TAB" to cycle through the different values, then type the value that you want. You can also press `x`, `y`, `d` or `a` to go directly to one of those input values.
 - If you start to type a number it will go to the `distance` input directly.
 - The input will receive numbers, dot and simple arithmetic operations. Examples: `3.45`, `-3.45`, `3.45 + 2`, `3.45 - 2`, `3.45 * 2`, `3.45 / 2`.
 - The input will be evaluated if you go to another value pressing "TAB" or one of values, or if you press "RETURN".
 
 Improvements already in mind:
 - All the instructions should be on the screen when the tool is activated.
 - The walls generated should be joined.
 - The tool should add the points relative to the Z value of the default container.
 - The angles of the axis to snap should be configurable.
 - The input panel should have a background rectangle to improve visibility.
 - X and Y input should accept absolute or relative (to the last point) coordinates.
 - "C" to close the polyline.
 - Create a temporary snap system to use points as references.
 - Improve object snap options.
 - General improvements in decorator colors, text, line weight and possible new shaders.

Please feel free to test it and give feedback! 

PS - About the code, I created Snaping and Raycasting modules, but I still want to review how they are organized. Also, the input panel data is shared between wall.py and the decorator.py, but I don't know if it's better to turn them into Blender Properties. I also thought of using a data class but those are used mainly for panels and menus, not for decorators. But I guess, in this case, the input panel decorator works as a UI.


Moult:
At a brief glance this is great and should be merged - it's nice and isolated and so users can venture into it slowly as it matures.

One very minor nitpick is to rename the tools to `raycast` and `snap` (remove `ing`). As time goes on and snapping becomes more sophisticated to handle per-IFC-standard-case snapping we'll see if it needs reorganisation.

Go ahead and merge (rebase and merge) :)

Btw for "The input panel should have a background rectangle to improve visibility." in Blender 4.2 there is a new(?) shadow type which strokes the text:

```
        blf.enable(self.font_id, blf.SHADOW)
        blf.shadow(self.font_id, 6, 0, 0, 0, 1)
```

Here's what I'd say is the priority list:

 1. The walls generated should be joined.
 2. The tool should add the points relative to the Z value of the default container.
 3. Build another simple (length only, for now) measure tool (probably as a hotkey to be activated within the Explore Tool). It should just share some of the same snapping behaviour (not the Z=0 snap), as well as show additional decorators for XYZ dims.

The reason I suggest pivoting away and working on a measure tool as priority 3 is because that will be a good refactoring exercise to ensure the code is properly reusable across multiple similar features.

brunoperdigao:
@Moult Thanks for the feedback. I implemented 1 and 2, and I'll start working on the measure tool.
I've just rebased this branch. Is everything ok to merge? Just double checking :)

Moult:
Yep OK to merge (and rebase, so it stacks nicely and can read the history) :)

theoryshaw:
awesome sauce!

noticed a little glitch....

https://github.com/user-attachments/assets/474b633b-98ef-44d7-b16c-62659ce342b3

theoryshaw:
Would be awesome to be able to add lengths in feet and inches, if the ifc file is in imperial. :)

theoryshaw:
In the following video, intuitively I feel like the wall should snap vertically, as the cursor was directly below the 2nd click. 

https://github.com/user-attachments/assets/ead5bdd4-5511-47e4-95b7-c619f0fd11b5

Moult:
Maybe check whether or not the snap angle being calculated is a 2D XY angle not a 3D one.

brunoperdigao:
@theoryshaw  Thanks, I'll take a look at those issues. Imperial units will definitely be implemented.

brunoperdigao:
@theoryshaw those issues should be fixed.

Issue #5075, opened 2024-07-24
IfcConvert project appears without color, but in BlenderBIM has.


csritter:
Hi, I have a project where the colors appear correctly in BlenderBIM, but when I use IfcConvert, the project has no color (just gray).

Using IfcConvert:
![image](https://github.com/user-attachments/assets/08a6d1c3-4454-436a-928d-9a1edbf08474)

BlenderBIM:
![image](https://github.com/user-attachments/assets/07ccbd90-2432-4d52-820c-cf8cd37dc9d6)

IfcConvert version 0.8.0, commit 8b29e50f11da2730a19d21327c22a257d0434c9e
BlenderBIM version 240602

I will send an email with the project to @aothms and @Moult already to speed it up.
I hope you don't mind.

Moult:
So this is an interesting one and something most people aren't aware of. It's basically a duplicate of #2509 but long story short what IfcConvert is doing is actually correct - it's showing the rendering diffuse colour, which is defined oddly enough as a normalised ratio of 0 (e.g. always black).

So in Blender, if you switch from regular solid viewport shading to a rendered mode, you'll see that Blender also displays everything (correctly) as black! In the screenshot below you can see the black result in rendered mode, the top left shows the non-rendered pink colour, the top right shows the glTF compatible shader definition in your IFC, and the right hand side shows the original IFC definition where you'll notice that SurfaceColour is pink (slightly cut off) and DiffuseColour is an IfcNormalisedRatioMeasure of 0 (therefore 0 multiplied by anything is black).

![2024-07-25-222311_1580x1034_scrot](https://github.com/user-attachments/assets/980eca70-3825-4a4a-a7a9-7554ce955375)

csritter:
From what I understand, @aothms had the idea of ​​adding a flag to get the surface colors, but apparently, nothing was implemented.
Would you have any suggestions on how to get this using the C++ Serializer (like GlbSerializer) or Python?

Moult:
In Python it's really straightforward, prior to running [serialisation to glb](https://docs.ifcopenshell.org/ifcopenshell-python/geometry_processing.html#geometry-serialisation) do something like:

```python
for style in ifc_file.by_type("IfcSurfaceStyleRendering"):
    style.DiffuseColour = style.SurfaceColour
```

aothms:
[...]

PR #5184, opened 2024-08-15
Adding --surface-colour in conversion settings to force the use of surface color instead of diffuse color. Issue #5075

csritter:
I have now added support for using the surface color and overlaying the diffuse.

To use, just add --surface-colour to the IfcConvert command.

I thought it would be better to create 2 variables, one diffuse and one surface in the style struct because I intend to use a serializer that saves the 2 colors (if any) in a binary file. In the viewer that I will use, it will choose which color he wants to visualize. So I don't know if it's the best approach, but I've already applied this to the Serializers.


src/ifcgeom/taxonomy.h
```diff
@@ -305,24 +305,33 @@ typedef item const* ptr;
...
 
 				// @todo equality implementation based on values?
 				bool operator==(const style& other) const { return instance == other.instance; }
 
-				style() : specularity(std::numeric_limits<double>::quiet_NaN()), transparency(std::numeric_limits<double>::quiet_NaN()) {}
-				style(const std::string& name) : name(name), specularity(std::numeric_limits<double>::quiet_NaN()), transparency(std::numeric_limits<double>::quiet_NaN()) {}
+				style() : specularity(std::numeric_limits<double>::quiet_NaN()), transparency(std::numeric_limits<double>::quiet_NaN()), use_surface_color(false) {}
+				style(const std::string& name) : name(name), specularity(std::numeric_limits<double>::quiet_NaN()), transparency(std::numeric_limits<double>::quiet_NaN()), use_surface_color(false) {}
+
+				colour get_color() const {
```
aothms, line 329:
```suggestion
				const colour& get_color() const {
```

prevent a copy

src/ifcgeom/mapping/mapping.cpp
```diff
@@ -565,11 +565,13 @@ taxonomy::ptr mapping::map_impl(const IfcSchema::IfcStyledItem* inst) {
...
+	surface_style->use_surface_color = settings_.get<settings::SurfaceColour>().get();
+
+	static taxonomy::colour white = taxonomy::colour(1., 1., 1.);
+	double rgb[3];
+	if (process_colour(shading->SurfaceColour(), rgb)) {
+		surface_style->surface.components() << rgb[0], rgb[1], rgb[2];
+	}
 
     if (auto rendering_style = shading->as<IfcSchema::IfcSurfaceStyleRendering>()) {
         if (rendering_style->DiffuseColour() && process_colour(rendering_style->DiffuseColour(), rgb)) {
```
aothms, line 577: I cannot make a comment in Gh about the line below :(

`576 | const taxonomy::colour& old_diffuse = surface_style->diffuse ? surface_style->diffuse : white;`

Shouldn't that also be changed to `surface_style->surface`? To retain the old behaviour of the diffuse value?

aothms (changes requested):
Thanks for your effort on this! I made two comments. And apologies, I also added a setting. Could you resolve the conflict? Thanks again.

src/ifcgeom/mapping/mapping.cpp
csritter: You say in this case, add a check so that if there is no value in the surface, put diffuse, right?
I'll add a check with `if (!surface_style->diffuse)` or something.

csritter: But let me comment. In get_color I have checked so that when there is no surface, it gets diffuse. 
Or when there was no diffuse, use the surface (I don't know if this is a possibility)

```
const colour& get_color() const {
	if ((use_surface_color && **surface**) || !diffuse) {
		return surface;
	}
	return diffuse;
}
```

csritter:
This last commit is to resolve the conflict

csritter:
I made a small change, and I think it's best to always assign diffuse, even if it's surface, and overwrite it, if it's diffuse.
This way get_color will always return diffuse, and there will be less checking to do. 
Do you think this is valid?

aothms:
Thanks again for your work and sorry for the delay in merging. Things can always be fine-tuned later on if others have comments.

PR #5197, opened 2024-08-19
Feat measure tool

brunoperdigao:
This PR is a continuation of #5172 and implements the measure tool. It works in a similar way to the [wall polyline tool](https://community.osarch.org/discussion/2447/wall-polyline-tool). The main difference is that it works with x, y and z, and it doesn't add any object, just shows you information about coordinates, distance, angle and area. A few additional instructions:
- To start the tool you must select the Explore Tool and then press "Shift+M".
- You can press "shift+X", "shift+Y", "shift+Z" to change the snapping plane reference.
- If you have 3 or more co-planar vertices it will calculate the area a show it in green. If they are not co-planar it will show zero.
- The angle is relative to the last line in the polyline.

Because the wall polyline tool was develop taking into account only the x and y coordinates, a lot of refactor had to be done for the calculations to work also with z coordinate. So, it would be nice to keep testing the wall polyline tool as well.

Improvements already in mind:
- There are a few improvements to be made regarding variable names consistency.
- There are some precision problems with the results. For example, angles showing "179.9999" instead of "180.00".
- Support for imperial units.


theoryshaw:
Awesome!

You probably know about this already, but if you don't:
https://github.com/IfcOpenShell/IfcOpenShell/blob/2a808318bf8d3e403036f3a1160e768d3668e9fa/src/blenderbim/blenderbim/bim/module/drawing/helper.py#L123

I added a `imperial_precision` variable in the preference, that might come in hand too: https://github.com/IfcOpenShell/IfcOpenShell/commit/2a808318bf8d3e403036f3a1160e768d3668e9fa#diff-e3c2abd0076ef09787c649a0897ab0b4437d8c85dcbc6f0e8e80959955eff4b9R374

brunoperdigao:
> Awesome!
> 
> You probably know about this already, but if you don't:
> 
> https://github.com/IfcOpenShell/IfcOpenShell/blob/2a808318bf8d3e403036f3a1160e768d3668e9fa/src/blenderbim/blenderbim/bim/module/drawing/helper.py#L123
> 
> I added a `imperial_precision` variable in the preferences, that might come in handy too: [2a80831#diff-e3c2abd0076ef09787c649a0897ab0b4437d8c85dcbc6f0e8e80959955eff4b9R374](https://github.com/IfcOpenShell/IfcOpenShell/commit/2a808318bf8d3e403036f3a1160e768d3668e9fa#diff-e3c2abd0076ef09787c649a0897ab0b4437d8c85dcbc6f0e8e80959955eff4b9R374)

Related to this: https://community.osarch.org/discussion/comment/21832/#Comment_21832

Issue #4956, opened 2024-07-01
Dropdown menus for active workspace tool options


Moult:
When a tool is active (e.g the wall tool), it shows options in two locations: in the top bar as a horizontal line of options, as well as in the sidebar as a vertical list of options.

The options never fit in the top bar so the top bar is basically useless. The user has to unintuitively scroll the top bar (drag MMB) so they end up toggling the side panel. I want to take all the non-critical options and group them into menus similar to how Blender already does this in their own top bar (e.g. View / Select / Add / Object).

Anybody tackling this give me a buzz before starting so we can chat :)

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/88302/447e84e3-e12c-44a4-b5d0-a26586f7ab01)

Vanuan:
There's a third place, properties editor > tools. Though, I've just noticed it's hidden in BIM workspace. Hm, Blender icon also pushed from the subtab into dropdown. Lots of small changes.

trhyder:
@Moult I would like to have a go at this one, if no one else is working on it.

Moult:
Go for it!

trhyder:
> Anybody tackling this give me a buzz before starting so we can chat :)

@Moult do you have any additional info?

Moult:
The chat was to help point at which source code dirs should be used. Also as a discussion on what these menus should be titled. I feel as though the current grouping should be preserved, so an `Modify` menu for all the stuff like extend / butt / mitre, then an `Align` menu for alignment, `Mode` for toggling various modes, and `Aggregation` for the aggregation stuff. This'll continue to evolve over time but it's a start.

trhyder:
@Moult Once I have live dev environment setup, a few pointers to the key files would be helpful to get me started. Currently, I am waiting for an update to `dev_environment.sh`

trhyder:
Hey @Moult, I wanted to share some thoughts on this issue. What I really like about the existing setup is how, when you select an object, the available tools are right there, immediately visible. It’s a great feature that keeps things intuitive and efficient.

However, as this issue has identified, it is not quite working in its current format. The buttons are quite long, which can be a bit cumbersome. Plus, using the mouse wheel to scroll to the buttons that are off-screen isn’t very practical because it often breaks the scroll and ends up zooming in on the viewport instead.

[...]

PR #5223, opened 2024-08-25
Custom icons

trhyder:

Custom icons without text added to header
Custom icons with text and shortcuts added to sidebar

There is still a lot more I would like to do, but I believe this commit to be fully functional. Happy to receive feedback and develop further


Andrej730:
Those icons are beautiful and it unclutters the header, thank you so much!

PR #5257, opened 2024-08-29
Feature faster unit method ---> Main

raj-open:
The purpose of this pull request is to improve the method `get_property_unit` in the python api in the following ways:

## Dynamic memory allocation ##

Underlying the `get_property_unit` method was the method `get_measure_unit_type` which "opens" the underlying ifc file to obtain a dictionary (mapping type-names to unit-entities).

This is a time-expensive call to obtain a relatively small object. If one called `get_property_unit` in a loop (e.g. for all property-entities in the ifc-file), these costs accumulated significantly.

This PR firstly changes this, so that this dictionary is computed _once_ and stored within the file-class. We claim that this leads to a (time) efficiency improvement when calling the method multiple times whilst not resulting in significant memory costs.

## Heterogenous return type fixed ##

The logic of the method (`get_property_unit`) itself contained a case which returns an object whose return type (`dict<str, entity|null>`) did not match the function's type hint (`entity | null`).

We have carved this case out and placed it in a separate method called `get_property_table_unit`, just for this purpose.

We request a code review plus testing, to see if our PR is acceptable.


src/ifcopenshell-python/ifcopenshell/util/constants.py
```diff
@@ -0,0 +1,65 @@
...
+    "IfcPropertySingleValue",
+    "IfcQuantityArea",
+    "IfcQuantityCount",
+    "IfcQuantityLength",
+    "IfcQuantityTime",
+    "IfcQuantityVolume",
+    "IfcQuantityWeight",
+    "IfcSIUnit",
+    "IfcUnitAssignment",
+]
```
Moult, line 65: What are all these constants for?

raj-open: The intention there was to create a type. I should add that in my PR this is not _directly_ used.

Usage of this type would allow any developer who imports of `ifcopenshell` to for example write methods like this:
```py
from functools import reduce
from ifcopenshell.file import file as IfcFile
from ifcopenshell.util.constants import IFC_TYPES

def get_names_of_entities_by_types(
  f: IfcFile,
  types: set[IFC_TYPES],
  /,
) -> list[str]:
  """
  Obtains all entities by a given set of types
  and returns the names of these elements
  """
  # get all entities
  entities = reduce(
    lambda x, y: x + y,
    map(lambda t: f.by_type(t), types),
    []
  )
  # get names
  names = map(lambda e: getattr(e, "Name", None), entities)
  # filter out null-values
  names = filter(lambda name: isinstance(name, str), names)
  # return list
  return list(names)
```
This provides the advantage to the developer as one then has type hints, so that one nows what values can be used:
example
In fact, it would by nice at some point to implement such a `Literal`-type for methods in the repository like `(class file).by_type`, etc.

src/ifcopenshell-python/ifcopenshell/util/unit.py
```diff
@@ -445,56 +451,83 @@ def get_property_unit(
...
         return unit
-    unit_assignment = get_unit_assignment(ifc_file)
-    if not unit_assignment:
-        return
-    entity = prop.wrapped_data.declaration().as_entity()
+
+    value = None
     measure_class = None
+
+    # DEV-NOTE: Using .is_a() is wrong, as it tells us nothing about super class membership
```
raj-open, line 461: NOTE: this was more of a note to myself, as I was initially tempted to change this to
```py
match prop.is_a():
  case "IfcPhysicalSimpleQuantity":
    ...
  case "IfcPropertySingleValue":
    ...
  ...
```
(or if-elseif-...) until I realised this does not work, as `.is_a()` returns the current ifc-type, but says nothing about membership to a "super" ifc-type.

src/ifcopenshell-python/ifcopenshell/util/unit.py
```diff
@@ -445,56 +451,83 @@ def get_property_unit(
     :return: The IFC unit entity, or nothing if there is no default project
         unit defined.
     """
-    unit = getattr(prop, "Unit", None)
-    if unit:
+    unit = prop.Unit
```
raj-open, line 454: NOTE: I had considered leaving this `getattr(prop, "Unit", None)`, thinking this might be "safer". However even in the original code one often just writes `prop.UpperBoundValue` assuming the `__get_attr__` magic will just work, so for consistency I applied the same style.

If there be some particular reason to assume this is unsafe, we can definitely revert this to `getattr(prop, "Unit", None)`. Though in this case, we should probably then do the same for the other `prop. ...`-methods.

src/ifcopenshell-python/ifcopenshell/util/constants.py
Moult: Sorry for the incredibly delayed response. These names change per IFC schema version, so doesn't that limit its value as the schema version might change or even be custom? IfcOpenShell can dynamically return these values too (at runtime, so admittedly not for the type hinting that you suggest).

I see value for this if somebody wants to write some strongly typed code, but perhaps that's a separate discussion / PR.

raj-open: okay, I guess, unless the repository regularly and dynamically updates IFC models from a central source, then it is find to remove these.

Issue #5246, opened 2024-08-29
Editing project IFC header raises error for some files


manuvarkey:
Following error is raised when the IFC header editing is activated for certain files.

![image](https://github.com/user-attachments/assets/03256aac-9490-465c-9210-3d0dda407dbe)

```
Python: Traceback (most recent call last):
  File "/home/eee/.config/blender/4.2/extensions/.local/lib/python3.11/site-packages/bonsai/bim/module/project/operator.py", line 577, in execute
    author = self.file.wrapped_data.header.file_name.author
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eee/.config/blender/4.2/extensions/.local/lib/python3.11/site-packages/ifcopenshell/ifcopenshell_wrapper.py", line 6047, in author
    return _ifcopenshell_wrapper.FileName_author(self, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Type held at index 2 is 5Blank and not St6vectorINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESaIS5_EE

```

Model File:
https://github.com/user-attachments/files/16762216/Karayatta.House.zip

Andrej730:
Related to #5118

Moult:
Seems to be fixed now.

PR #5263, opened 2024-08-31
Expose header properties with invalid values #5246

Andrej730:


aothms (approved):
Cool. It would probably help if we `%ignore` the typed versions of these so that there is no C++ code generated for accessing these from the wrapper (which are overwritten now... I guess only in Python..). Ignoring these would then make the binary a bit lighter.

(We should probably review at some point the immense amount of accessors that are in there. Maybe it allows us to shave off a couple of MBs)

PR #5283, opened 2024-09-02
Update IfcGlobalId.cpp to support boost 1.86

Krande:
It looks like mt19937 is no longer a member of boost in boost 1.86. I propose we use to std::mt19937 (which is already how it is used by svgfill.cpp)

see failure log from conda forge libboost 1.86 test here: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1008654&view=logs&j=4f922444-fdfe-5dcf-b824-02f86439ef14&t=b2a8456a-fb11-5506-ca32-5ccd32538dc0&l=880

### DIFF
--- a/src/ifcparse/IfcGlobalId.cpp
+++ b/src/ifcparse/IfcGlobalId.cpp
@@ -87,8 +87,12 @@ void expand(const std::string& s, std::vector<unsigned char>& v) {
     }
 }
 
-// A random number generator for the UUID
-static boost::uuids::basic_random_generator<std::mt19937> gen;
+// Define the macro to handle different Boost versions
+#if BOOST_VERSION >= 108600
+    static boost::uuids::basic_random_generator<std::mt19937> gen;
+#else
+    static boost::uuids::basic_random_generator<boost::mt19937> gen;
+#endif
 
 IfcParse::IfcGlobalId::IfcGlobalId() {
     uuid_data_ = gen();
--- a/src/ifcparse/IfcGlobalId.cpp
+++ b/src/ifcparse/IfcGlobalId.cpp
@@ -88,7 +88,7 @@ void expand(const std::string& s, std::vector<unsigned char>& v) {
 }
 
 // A random number generator for the UUID
-static boost::uuids::basic_random_generator<boost::mt19937> gen;
+static boost::uuids::basic_random_generator<std::mt19937> gen;
 
 IfcParse::IfcGlobalId::IfcGlobalId() {
     uuid_data_ = gen();


RickBrice:
I agree with this change but it requires build scripts be updated for boost 1.86. See 0e9a8beeb85b51af9d6ae9e1436930d9c516ec2b on vs2022_boost_build_problems branch.

Krande:
Thanks for the heads up @RickBrice! 

I guess, as a starting point we could perhaps just start by adding a macro like I did in https://github.com/IfcOpenShell/IfcOpenShell/pull/5283/commits/1c19f0847125aaa092b1d5a8b83fd605b35f7590 to at least make the code compatible without having to make any larger changes to the main build setup?

I don't know enough about boost, but is there any particular differences in using `boost::random::mt19937` vs `std::mt19937`?

RickBrice:
> I don't know enough about boost, but is there any particular differences in using boost::random::mt19937 vs std::mt19937?

not sure, I just made the change to get a build to work. 

there have been issues building with visual studio- see #4744 

boost 1.86 seems to be the fix but have been unable to confirm

Krande:
I see. For what it's worth, using conda as the package manager for my IfcOpenShell dependencies, I have successfully compiled IfcOpenShell using VS2022 against both boost 1.84 and 1.86. using toolset versions 14.40 and 14.41.
If it can help in debugging the build scripts, I've made a small [recipe](https://github.com/Krande/IfcOpenShell/blob/v0.8.0-libboost-1-86-support/cmake/cmake_presets_w_conda.md) for how I used a conda [environment.yml](https://github.com/Krande/IfcOpenShell/blob/v0.8.0-libboost-1-86-support/conda/environment.yml) for my Ifcopenshell dependencies together with a [CMakePresets.json ](https://github.com/Krande/IfcOpenShell/blob/v0.8.0-libboost-1-86-support/cmake/CMakePresets.json) to quickly set up IfcOpenShell with a preset with the required dependencies and cmake variables. 

Most of the test suite passes, but I noticed that when using boost 1.86, one of the tests using IfcConvert fails because the messages in the json log file are missing newlines

```
{"level":"Notice","message":"exclude entities IfcOpeningElement IfcSpace","time":"2024-09-03 09:26:57"}{"level":"Performance","message":"file geometry conversion","time":"2024-09-03 09:26:57"}
```

instead of how it works boost<1.86

```
{"level":"Notice","message":"exclude entities IfcOpeningElement IfcSpace","time":"2024-09-03 09:26:57"}
{"level":"Performance","message":"file geometry conversion","time":"2024-09-03 09:26:57"}
```

It was easy enough to fix, just added a newline in `json_message` in the [IfcLogger.cpp](https://github.com/IfcOpenShell/IfcOpenShell/blob/f29f5d45f13eeb7105e37ed0345169a2c1046040/src/ifcparse/IfcLogger.cpp#L120-L122) 

@aothms Do you think the suggested fix in this PR is OK for now? And should I include the newline fix as well?

aothms (approved):

aothms:
@Krande I tracked down the change in boost https://github.com/boostorg/property_tree/commit/71dd6583254dc9fd6c210c2bafd0936f85c6ab6f Which indeed appears intentional. So yes, let's merge in that logger fix a well.

I think with the macro this can't result any issues also on previous boost version. I don't want to look into whether we should just always use the mersienne from `std` as opposed to boost's. Never know whether there any incompatibilities in past versions.

PR #5287, opened 2024-09-03
if boost >= 1.86 append newline to json logger

Krande:
Here is the boost 1.86 compatibility macro as mentioned in https://github.com/IfcOpenShell/IfcOpenShell/pull/5283


aothms (approved):

aothms:
Awesome thanks @Krande

PR #5289, opened 2024-09-03
fix last point coincides with first point

htlcnn:
fix last point coincides with first point @brunoperdigao


brunoperdigao:
thanks @htlcnn

PR #5293, opened 2024-09-03
Fix bug in activating drawing with exclude filter

manuvarkey:
Activating a drawing view when an element included in exclude filter is selected results in the element not being excluded.


theoryshaw:
This isn't really a bug, per say.
Selections can sometimes take time to setup, this assures things are not deselected when you navigate from one drawing to the next.

manuvarkey:
Is there a way to retain selections across drawing view changes without affecting the exclude filter ? 

Also currently activating model view when drawing view is active selects all elements. This needs to be addressed if selections are to be retained while going to drawing view and returning to model view.

manuvarkey:
What about this ? Selection is retained across drawing views.

PR #5306, opened 2024-09-04
Updated icons

trhyder:
- new master svg file for storing icons bonsai_icons.svg
- script for exporting png files and generating light mode versions
- support for light mode
#4956


duarteframos:
Not sure if you follow Blender development, or if are already aware of it, but there are potential plans for allowing SVG icons directly inside Blender in the future, no need for prior conversion to raster formats.

There is also the possibility that they may become themable at some point, dynamically taking colors from current Blender theme.
See [proof of concept](https://projects.blender.org/blender/blender/pulls/123950) and [experimental implementation](https://projects.blender.org/blender/blender/pulls/125146).

The gist of it is, if you organize your paths inside an SVG icon into groups, and name those groups after some theme color datapath like say `ThemeUserInterface.icon_scene` they will inherit that color from the theme automatically at runtime.

These are just speculative at this point and may or may not be merged in the future, but I'd say there is a good chance and the developer has a nice track record so far.
Just letting you know before investing too much time into it, or maybe you can consider taking all this into account when making new icons.

Moult:
@duarteframos thanks for the heads up!

trhyder:
@duarteframos that would be a much better way of dynamically handling different view modes. Thanks for the link

PR #5339, opened 2024-09-06
Implement sorting of spatial containers by elevation and name

manuvarkey:


manuvarkey:
@Moult as per discussion on Matrix

PR #5342, opened 2024-09-06
ifcwrap cmake - fix CMP0078 warning

Andrej730:


aothms:
Hi @Andrej730 according to https://cmake.org/cmake/help/latest/policy/CMP0078.html

> Variable SWIG_MODULE_<name>_REAL_NAME **must** be used to get real target name. This is the **default** if not specified.

Shouldn't we still keep using SWIG_MODULE_ifcopenshell_wrapper_REAL_NAME if we set CMP0078 to old?

Andrej730:
> Shouldn't we still keep using SWIG_MODULE_ifcopenshell_wrapper_REAL_NAME *if we set CMP0078 to old*?

I wanted to remove `cmake_policy(SET CMP0078 OLD)` to resolve this warning it prints:
```
CMake Deprecation Warning at CMakeLists.txt:32 (cmake_policy):
  The OLD behavior for policy CMP0078 will be removed from a future version
  of CMake.
```

But I think I've misread it thinking that `STANDARD` is the new default. Let's use the SWIG_MODULE_ifcopenshell_wrapper_REAL_NAME then as it is the current default.

aothms:
Thanks @Andrej730 that's also how I read the docs. Thanks for the quick fix.

Issue #4727, opened 2024-05-24
Decouple piecewise_function evaluation state from function definition to better support multi-threading


RickBrice:
I'm using IfcOpenShell v0.8.0, build from source, and BlenderBIM from "Live development environment". For the attached (albeit poorly constructed) model, I am expecting to see IfcAlignmentHorizontal and IfcAlignmentVertical under IfcAlignment as well as a 3D alignment curve in the graphical scene.

Any idea why they aren't there?
[Br01_UTS.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/15438180/Br01_UTS.ifc.txt)
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/37087370/6bb52a6f-1a9a-4c6c-b6d0-4a27a4a8182e)

civilx64:
To piggyback on the conversation, we would expect to see both the 2D horizontal IfcCompositeCurve in the plane of Z=0 as well as the 3D IfcGradientCurve in the graphical window. This is required to accommodate models where the limits of the vertical alignment are shorter than the extent of the horizontal - which is very common in practice.

Moult:
We load IfcElement by default. Other things are determined on a case by case basis. For example, openings are hidden, space boundaries are lazy loaded when the user requests them, etc. This means that IfcLinearElement is not included in the default project load.

I don't know enough whether we should:

 1. Load IfcLinearElement by default all the time, or...
 2. Not load it by default, but let the user load it later / toggle visibility such as by clicking a bridge segment then pressing a button / hotkey to toggle display modes to show alignments in different ways with cool visualisations (e.g. top down, or automatic elevation annotation text, etc).

My gut feel is the latter - that the user should generally see all "physical" elements loaded by default, and these "special" virtual elements can be toggled on/off later (like space boundaries, voids, structural connections...)

RickBrice:
Being a total BBim novice, @Moult is it fair to summarize your post as BBim doesn’t have support for IfcLinearElement yet (I.e., no UI toggles to display linear elements yet)?

Is there a way to show the elements via a script? What would an example look like?

You can read between the lines and assume that if I get pointed in the right direction, I would like to contribute to bridge and infrastructure capabilities of BBim.

Moult:
Yes I think currently there is no support to load it. The easiest way to load it is to modify the `process_element_filter` function probably in `import_ifc.py` to include not just IfcElement but IfcLinearElement too.

[...]

PR #5344, opened 2024-09-06
Decouples evaluation of a piecewise_function from the function itself

RickBrice:
Is this what you were thinking. 
https://github.com/IfcOpenShell/IfcOpenShell/issues/4727#issuecomment-2266276969

### DIFF
--- a/src/ifcgeom/piecewise_function_evaluator.cpp
+++ b/src/ifcgeom/piecewise_function_evaluator.cpp
@@ -1,0 +1,104 @@
+#include "piecewise_function_evaluator.h"
+#include "profile_helper.h"
+
+using namespace ifcopenshell::geometry;
+
+
+piecewise_function_evaluator::piecewise_function_evaluator(taxonomy::piecewise_function::const_ptr pwf, ifcopenshell::geometry::Settings* settings) : pwf_(pwf), settings_(settings) {
+}
+
+taxonomy::piecewise_function::const_ptr piecewise_function_evaluator::get_pwf() const {
+    return pwf_;
+}
+
+std::vector<double> piecewise_function_evaluator::evaluation_points() const {
+    if (!eval_points_.has_value()) {
+        double curve_length = pwf_->length();
+
+        auto param_type = settings_ ? settings_->get<ifcopenshell::geometry::settings::PiecewiseStepType>().get() : ifcopenshell::geometry::settings::PiecewiseStepMethod::MAXSTEPSIZE;
+        auto param = settings_ ? settings_->get<ifcopenshell::geometry::settings::PiecewiseStepParam>().get() : 0.5;
+        unsigned num_steps = 0;
+        if (param_type == ifcopenshell::geometry::settings::PiecewiseStepMethod::MAXSTEPSIZE) {
+            // parameter is max step size
+            num_steps = (unsigned)std::ceil(curve_length / param);
+        } else {
+            // parameter is minimum number of steps
+            num_steps = (unsigned)std::ceil(param);
+        }
+
+        eval_points_ = evaluation_points(pwf_->start(), pwf_->start() + curve_length, num_steps);
+    }
+    return *eval_points_;
+}
+
+std::vector<double> piecewise_function_evaluator::evaluation_points(double ustart, double uend, unsigned nsteps) const {
+    double curve_length = pwf_->length();
+    ustart = std::max(pwf_->start(), ustart);
+    uend = std::min(uend, pwf_->start() + curve_length);
+
+    nsteps = std::max(1u, nsteps); // never have fewer than 1 step
+
+    auto resolution = (uend - ustart) / nsteps;
+
+    std::vector<double> u_values;
+    u_values.reserve(nsteps);
+
+    for (unsigned i = 0; i <= nsteps; ++i) {
+        auto u = resolution * i + ustart;
+        u_values.push_back(u);
+    }
+
+    return u_values;
+}
+
+taxonomy::item::ptr piecewise_function_evaluator::evaluate() const {
+    return evaluate(evaluation_points());
+}
+
+taxonomy::item::ptr piecewise_function_evaluator::evaluate(double ustart, double uend, unsigned nsteps) const {
+    return evaluate(evaluation_points(ustart, uend, nsteps));
+}
+
+Eigen::Matrix4d piecewise_function_evaluator::evaluate(double u) const {
+    // assume monotonic evaluation and store last evaluated segment
+    if (current_span_fn_ == nullptr || (u < current_span_start_ || current_span_end_ < u)) {
+        // there isn't a current span or u is outside the range of the current span
+        // get a new "current span"
+        std::tie(current_span_start_, current_span_end_, current_span_fn_) = get_span(u);
+    }
+
+    u -= current_span_start_; // make u relative to start of span
+    return (*current_span_fn_)(u);
+}
+
+taxonomy::item::ptr piecewise_function_evaluator::evaluate(const std::vector<double>& dist) const {
+    std::vector<taxonomy::point3::ptr> polygon;
+    polygon.reserve(dist.size());
+    for (auto& u : dist) {
+        Eigen::Matrix4d m = evaluate(u);
+        polygon.push_back(taxonomy::make<taxonomy::point3>(m(0, 3), m(1, 3), m(2, 3)));
+    }
+
+    return polygon_from_points(polygon);
+}
+
+std::tuple<double, double, const std::function<Eigen::Matrix4d(double u)>*> piecewise_function_evaluator::get_span(double u) const {
+    // force u to be within bounds of the curve
+    double s = pwf_->start();
+    double e = pwf_->end();
+    u = std::max(s, u);
+    u = std::min(u, e);
+
+    double span_start = s;
+    for (auto& [length, fn] : pwf_->spans()) {
+        double span_end = span_start + length;
+        auto tolerance = settings_ ? settings_->get<ifcopenshell::geometry::settings::Precision>().get() : 0.001;
+        if (span_start <= u && u < span_end + tolerance) {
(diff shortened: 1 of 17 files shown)


aothms:
Yes 100%.

Question: can `settings_` now also be removed? https://github.com/IfcOpenShell/IfcOpenShell/blob/dfe27e557ceac8f87f6ca514a4b778b53adef1a8/src/ifcgeom/taxonomy.h#L366

Maybe later we move it out of taxonomy.h so that we have a clearer separation between the data model and the logic. And also save a bit of compilation time. Because taxonomy.h is included n_schemas x n_entities times.

aothms (approved):

RickBrice:
> Question: can settings_ now also be removed?

I will investigate, but I think yes

> Maybe later we move it out of taxonomy.h

I can do that before merging the PR. Easy enough to put the evaluator in its own cpp/h file. Any preference on namespace? Currently in `ifcopenshell::geometry::taxonomy`

RickBrice:
@aothms I made all the changes as you suggested. I've run into one problem I don't know how to solve.

Before moving `piecewise_function_evaluator` out of the `taxonomy.h/cpp` files, my python code looked like this and it worked

~~~
pwf = ifcopenshell.ifcopenshell_wrapper.map_shape(ifcopenshell.geom.settings(),inst.wrapped_data)
evaluator = ifcopenshell.ifcopenshell_wrapper.piecewise_function_evaluator(pwf)
locations = evaluator.evaluation_points()
~~~

Now, with the small change of adding settings to the evaluation, this no longer works

~~~
 pwf = ifcopenshell.ifcopenshell_wrapper.map_shape(ifcopenshell.geom.settings(),inst.wrapped_data)
 evaluator = ifcopenshell.ifcopenshell_wrapper.piecewise_function_evaluator(pwf,ifcopenshell.geom.settings())
 locations = evaluator.evaluation_points()
~~~

The error is:

~~~
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[1], line 50
     48 #print(inst)
     49 pwf = ifcopenshell.ifcopenshell_wrapper.map_shape(ifcopenshell.geom.settings(),inst.wrapped_data)
---> 50 evaluator = ifcopenshell.ifcopenshell_wrapper.piecewise_function_evaluator(pwf,ifcopenshell.geom.settings())
     51 locations = evaluator.evaluation_points()
     52 left = []

AttributeError: module 'ifcopenshell.ifcopenshell_wrapper' has no attribute 'piecewise_function_evaluator'
~~~

Clearly this is a problem with Swig and the python bindings. I have attempted the following with no solution.
1) Added `#include "../ifcgeom/piecewise_function_evaluator.h"` ever place I see `#include "../ifcgeom/taxonomy.h"`
2) Re-ran `run-cmake.bat vs2022-x64`
3) Did a complete recompile

I'm sure the problem something small but I don't know what it is. Any ideas?

aothms:
The swig C++ module comes with its own set of includes.

1. Includes that are necessary for the generated C++ code to compile: https://github.com/IfcOpenShell/IfcOpenShell/blob/e7ad71b51de955ec7d794bee3a9c1ea8d39f316f/src/ifcwrap/IfcPython.i#L86
2. Includes that SWIG actually processes to generate wrapper functions. These do not follow subsequent includes! (I think) So the new header needs to be added here: https://github.com/IfcOpenShell/IfcOpenShell/blob/e7ad71b51de955ec7d794bee3a9c1ea8d39f316f/src/ifcwrap/IfcGeomWrapper.i#L248

IfcGeomWrapper.i is not part of the solution (that's just a silly oversight, it should be) so you need to edit and then recompile IfcPython.i.

RickBrice:
@aothms thanks - that was the piece I was missing.

There is one more issue before I merge - please help me understand and provide some ideas for debugging.

When running from a jupyter notebook, I have the following code
~~~
        pwf = ifcopenshell.ifcopenshell_wrapper.map_shape(ifcopenshell.geom.settings(),inst.wrapped_data)
        evaluator = ifcopenshell.ifcopenshell_wrapper.piecewise_function_evaluator(pwf,ifcopenshell.geom.settings())
        locations = evaluator.evaluation_points()
        print(locations)
       (0.0,100.0)  <-- This is just then end points
~~~

But if I do this, the results are as expected
~~~
        pwf = ifcopenshell.ifcopenshell_wrapper.map_shape(settings,inst.wrapped_data)
        evaluator = ifcopenshell.ifcopenshell_wrapper.piecewise_function_evaluator(pwf,settings)
        locations = evaluator.evaluation_points()
        print(locations)
(0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 6.0, 6.5, 7.0, 7.5, 8.0, 8.5, 9.0, 9.5, 10.0, 10.5, 11.0, 11.5, 12.0, 12.5, 13.0, 13.5, 14.0, 14.5, 15.0, 15.5, 16.0, 16.5, 17.0, 17.5, 18.0, 18.5, 19.0, 19.5, 20.0, 20.5, 21.0, 21.5, 22.0, 22.5, 23.0, 23.5, 24.0, 24.5, 25.0, 25.5, 26.0, 26.5, 27.0, 27.5, 28.0, 28.5, 29.0, 29.5, 30.0, 30.5, 31.0, 31.5, 32.0, 32.5, 33.0, 33.5, 34.0, 34.5, 35.0, 35.5, 36.0, 36.5, 37.0, 37.5, 38.0, 38.5, 39.0, 39.5, 40.0, 40.5, 41.0, 41.5, 42.0, 42.5, 43.0, 43.5, 44.0, 44.5, 45.0, 45.5, 46.0, 46.5, 47.0, 47.5, 48.0, 48.5, 49.0, 49.5, 50.0, 50.5, 51.0, 51.5, 52.0, 52.5, 53.0, 53.5, 54.0, 54.5, 55.0, 55.5, 56.0, 56.5, 57.0, 57.5, 58.0, 58.5, 59.0, 59.5, 60.0, 60.5, 61.0, 61.5, 62.0, 62.5, 63.0, 63.5, 64.0, 64.5, 65.0, 65.5, 66.0, 66.5, 67.0, 67.5, 68.0, 68.5, 69.0, 69.5, 70.0, 70.5, 71.0, 71.5, 72.0, 72.5, 73.0, 73.5, 74.0, 74.5, 75.0, 75.5, 76.0, 76.5, 77.0, 77.5, 78.0, 78.5, 79.0, 79.5, 80.0, 80.5, 81.0, 81.5, 82.0, 82.5, 83.0, 83.5, 84.0, 84.5, 85.0, 85.5, 86.0, 86.5, 87.0, 87.5, 88.0, 88.5, 89.0, 89.5, 90.0, 90.5, 91.0, 91.5, 92.0, 92.5, 93.0, 93.5, 94.0, 94.5, 95.0, 95.5, 96.0, 96.5, 97.0, 97.5, 98.0, 98.5, 99.0, 99.5, 100.0)
~~~

Why would getting the settings object twice give different results?

From jupyter (or just a python console), how would I go about setting a breakpoint in the C++ code to debug?

Andrej730:
> From jupyter (or just a python console), how would I go about setting a breakpoint in the C++ code to debug?

@RickBrice There are couple ways to do it:
1) Start Python with Visual Studio debugger attached and add a breakpoint in IfcPythonPYTHON_wrap.cxx to some ifcopenshell wrapper's method you're about to call from Python.
2) Add a `__debugbreak()` anywhere in C++ code that you want to debug and it will open Visual Studio automatically.
3) I'm not sure if it works on other systems but I've compiled for myself [debugbreak](https://pypi.org/project/debugbreak/) package that allows you to do `debugbreak()` from python code that will open Visual Studio and then you can continue debugging.

src/ifcgeom/piecewise_function_evaluator.h
```diff
@@ -0,0 +1,58 @@
...
+    /// @return 4x4 placement matrix
+    Eigen::Matrix4d evaluate(double u) const;
+
+  private:
+    taxonomy::item::ptr evaluate(const std::vector<double>& dist) const;
+    std::tuple<double, double, const std::function<Eigen::Matrix4d(double u)>*> get_span(double u) const;
+
+    taxonomy::piecewise_function::const_ptr pwf_;
+
+    const ifcopenshell::geometry::Settings* settings_ = nullptr;
```
aothms, line 48: The issue is not 1 or 2 times. The issue is that you accept a pointer to an object managed by SWIG. So the corresponding python object is more or less immediately gc'ed because python/swig is not aware that we still have a pointer to that address.

```suggestion
    ifcopenshell::geometry::Settings settings_;
```

src/ifcgeom/piecewise_function_evaluator.cpp
```diff
@@ -0,0 +1,100 @@
+#include "piecewise_function_evaluator.h"
+#include "profile_helper.h"
+
+using namespace ifcopenshell::geometry;
+
+
+piecewise_function_evaluator::piecewise_function_evaluator(taxonomy::piecewise_function::const_ptr pwf, ifcopenshell::geometry::Settings* settings) : pwf_(pwf), settings_(settings) {
```
aothms, line 7:
```suggestion
piecewise_function_evaluator::piecewise_function_evaluator(taxonomy::piecewise_function::const_ptr pwf, ifcopenshell::geometry::Settings* settings) : pwf_(pwf), settings_(*settings) {
```

aothms (changes requested):

aothms:
What I normally do for debugging:

- In Visual Studio you can attach to running processes [ctrl] [alt] [P]. If you have multiple python proceses running they are sorted desc by lifetime so typically you need to attach to the first one. Then you can set breakpoints as you please.
- In Python I run with `python -m pdb myscript.py` and then I can set a breakpoint there as well with `b myfile.py:123, condition == True`, then run `c` to continue python execution. This helps in case you're not fast enough to attach to the running process.

Issue #1646, opened 2021-08-12
Tooltips show (undocumented operator)


onceadrog:
Mouse over on any element shows (undocumented operator), however previous versions had functioning tooltips.
e.g. 
![image](https://user-images.githubusercontent.com/6299372/129269524-1b23e36a-776c-4bec-a0ab-0d18c36a2d41.png)

Blender 2.93.2, ifcOpenShell 0.0.210812, Python 3.9.2

Moult:
No tooltips ever existed even in previous versions :) I've been holding off due to the alpha nature and the UI changing rather frequently, but if this is something you want to contribute to, please feel free, it requires relatively minimal coding knowledge - just the ability to find the code (which is pretty easy to figure out after a couple times).

Moult:
Hey @maxfb87 ! Would love contributions. Please submit a PR with your code changes so it can be merged! Pretty much any module except for model, structural, system, drawing, constraint, cost, and sequence modules would be good to go!

Moult:
Your PR has been merged. Great work! Feel free to go ahead and add tooltips everywhere you see fit!

maxfb87:
@Moult @brunopostle i'm not sure i can make it, because maybe it's too much difficult for me, but i'll try and i'll write here what i'll do

brunopostle:
@maxfb87 you don't need to make the strings translatable, this can be automated later-on. What you have done already is more useful.

maxfb87:
@brunopostle ok, so i focus on the tooltip in english at first...and after that i try to make them translatable...

maxfb87:
@brunopostle @Moult i found two tutorial about localization. I think they are very useful because they are very practice:
1 - https://b3d.interplanety.org/en/localization-of-blender-add-ons/
2 - https://b3d.interplanety.org/en/localization-of-blender-add-ons-with-api/

The first one seems much "raw" and the second seems much "integrated", as it uses the blender API.

I would go with the second one but tell me what you think.

In both cases, btw, it's necessary to define a dictionary with the translation strings: where is the best place where i can put it?

brunopostle:
@maxfb87 the second of your links indicates that you don't need to do anything special to text that doesn't have format string handling, which is basically all normal tooltips.

I would continue as you already started with your first pull request. You can ignore the dictionary that contains translation strings (since blenderbim doesn't have any translations yet).

Moult:
[...]

PR #5348, opened 2024-09-07
Update tooltips for UI elements

manuvarkey:
Related to https://github.com/IfcOpenShell/IfcOpenShell/issues/1646

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/mep.py
+++ b/src/bonsai/bonsai/bim/module/model/mep.py
@@ -648,7 +648,7 @@ class MEPAddTransition(bpy.types.Operator, tool.Ifc.Operator):
     bl_idname = "bim.mep_add_transition"
     bl_label = "Add Transition"
     bl_description = (
-        "Adds transition between two MEP elements. Elements are either provided by ID or selected in Blender"
+        "Adds transition between two selected MEP Elements"
     )
     bl_options = {"REGISTER", "UNDO"}
     start_length: bpy.props.FloatProperty(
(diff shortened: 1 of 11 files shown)


Moult:
Why are some of the labels blanked out?

manuvarkey:
> Why are some of the labels blanked out?

@Moult 

When adding text to both `bl_label` and `bl_description` for operators with `text=""` (for ex. `row.operator("bim.edit_openings", icon="CHECKMARK", text="")`), both labels are displayed as tooltips. 

Screenshots given below.

Screenshot from 2024-09-07 21-45-56
Screenshot from 2024-09-07 22-09-35
Screenshot from 2024-09-07 22-14-34

For all other operators, I have added both labels.

manuvarkey:
Adding `bl_label` alone shows `(undocumented operator)` message.

Screenshot from 2024-09-07 22-29-10

Andrej730:
`bl_label` is important as it's the operator's name appearing in the operator search and quick favorites, so it shouldn't be empty. I think there is no way in Blender to avoid "(undocumented operator)" without providing any description - so we should provide one or keep "(undocumented operator)" for now.

Moult:
Agreed, @manuvarkey do you mind ensuring that bl_label is never blank? If it duplicates the description, so be it until we find something even more descriptive. If you make a list of duplicates, I can also propose some descriptions. Happy to merge after these changes.

Moult:
A minor nitpick not to change "Opening" to "Void". It is an "Opening" element in IFC - that is the exact jargon, the "Void" is a verb, the relationship between the opening and the building element.

manuvarkey:
@Moult Made the suggested changes. I have added short identifiers for `bl_label` field in the above commits so as not to leave them blank.

Moult:
Edit openings should make all temporary openings disappear and sync changes back to the building element.

PR #5359, opened 2024-09-09
Polyline refactor

brunoperdigao:
@Moult I'm pushing this as a PR so you can take a look at how I'm organizing the code. I created a `tool/polyline.py` to accommodate a few functions and a `model/polyline.py` to create a base class that can be used by wall polyline and measure tool. Let me know if you have any suggestion.


Moult:
Totally fine by me :) Go for it.

Issue #5352, opened 2024-09-09
v0.8.0 doesn't support create_shape on IfcCircle


Moult:
```
ISO-10303-21;
HEADER;
FILE_DESCRIPTION(('ViewDefinition[DesignTransferView]'),'2;1');
FILE_NAME('test2.ifc','2024-08-18T11:03:45+02:00',(),(),'IfcOpenShell 0.7.11-alpha240815','Bonsai 0.7.11-alpha240815-4348b9b','Nobody');
FILE_SCHEMA(('IFC4'));
ENDSEC;
DATA;
#1=IfcCartesianPointList2D(((0.820209980010986,0.820209980010986),(0.820209980010986,-0.820209980010986),(-0.820209980010986,-0.820209980010986),(-0.820209980010986,0.820209980010986),(0.820209980010986,0.820209980010986)));
#2=IfcIndexedPolyCurve(#1,$,$);
#3=IfcCartesianPoint((0.0584927275776863,-0.0146500710397959));
#4=IfcAxis2Placement2D(#3,$);
#5=IfcCircle(#4,0.328083989441551);
ENDSEC;
END-ISO-10303-21;
```

```python
import multiprocessing
import ifcopenshell
import ifcopenshell.geom

import ifcopenshell.ifcopenshell_wrapper as wrapper

wrapper.turn_on_detailed_logging()

ifc_file = ifcopenshell.open('/home/dion/test.ifc')
settings = ifcopenshell.geom.settings()
shape = ifcopenshell.geom.create_shape(settings, ifc_file.by_id(5))
print(shape.verts)
```

```
Done scanning file   
[Warning] [2024-09-09 13:00:37] Not a single project or context in file
[Warning] [2024-09-09 13:00:37] Unable to detect unit information
[Error] [2024-09-09 13:00:37] Not implemented
#5=IfcCircle(#4,0.328083989441551)
Traceback (most recent call last):
  File "/home/dion/Projects/ifcopenshell/src/test.py", line 20, in <module>
    shape = ifcopenshell.geom.create_shape(settings, ifc_file.by_id(5))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dion/Projects/env/lib/python3.11/site-packages/ifcopenshell/geom/main.py", line 411, in create_shape
    ifcopenshell_wrapper.create_shape(
[...]

PR #5362, opened 2024-09-10
Fix error processing IfcCircle/IfcEllipse #5352

Andrej730:
Not sure if this is the right approach but just noticed the way IfcCircleProfileDef and IfcEllipseProfileDef work.

Btw, I was thinking maybe we can move some upgrade_xxx implementations from .h to .cpp so it will be easier to debug them (any changes to them currently recompile the entire IfcOpenShell), probably will also reduce compilation time. E.g. leave in .h. only the part that requires a template:
```cpp
				curve_to_edge_upgrade(taxonomy::ptr item) {
					if constexpr (std::is_same_v<T, edge>) {
						return curve_to_edge_upgrade_process_item(item);
					}
				}
```

and move implementation to taxonomy.cpp as something like
```cpp
boost::optional<taxonomy::edge::ptr> curve_to_edge_upgrade_process_item(taxonomy::ptr item) {
    boost::optional<taxonomy::edge::ptr> edge_;

    auto circle = taxonomy::dcast<taxonomy::circle>(item);
    auto ellipse = taxonomy::dcast<taxonomy::ellipse>(item);
    auto line = taxonomy::dcast<taxonomy::line>(item);
    auto bspline_curve = taxonomy::dcast<taxonomy::bspline_curve>(item);

    if (circle || ellipse || line || bspline_curve) {
        edge_ = taxonomy::make<taxonomy::edge>();

        if (circle) {
            (*edge_)->basis = circle;
        } else if (ellipse) {
            (*edge_)->basis = ellipse;
        } else if (line) {
            (*edge_)->basis = line;
        } else if (bspline_curve) {
            (*edge_)->basis = bspline_curve;
        }

        if (circle || ellipse) {
            // @todo
            (*edge_)->start = 0.;
            (*edge_)->end = 2 * boost::math::constants::pi<double>();
        }
    }

    return edge_;
}
```


aothms:
I think it's a good idea moving the upgrade templates to the source. If we use "explicit template instantiation" even more can be moved to the source and probably helps further reduce compiled object size. Is that something you would be willing to do as part of this PR?

That brings me to my question though. Why not rely on the upgrade mechanism for this? The reason the upgrade does not kick in is because the `dispatch_conversion<>` loop does a strict check without inheritance or upgrades.

https://github.com/IfcOpenShell/IfcOpenShell/blob/5616367a03ea397885e93523e55788da82238742/src/ifcgeom/AbstractKernel.cpp#L30

This is for performance reasons, because it's performing all these upgrades in this loop is wasteful unless it concerns those few top-level items we care about. But I think we could do two iterations. One time as is with strict enum checking. One time after that with `taxonony::dcast<>` to also factor in potential upgrades only when the strict version of the loop did not trigger any function. Thoughts?

Because the other thing is that for advanced breps and trimmed curves we need them as curves and not topological edges, so in those cases we need to "unpack" the edges than if in the mapping we already treat them as topological items. Where in fact the upgrade mechanism was intended for that.

Andrej730:
> Is that something you would be willing to do as part of this PR?

I planned to make it in a different PR but it was requiring too much recompilation so it's now part of this PR 😁


> This is for performance reasons, because it's performing all these upgrades in this loop is wasteful unless it concerns those few top-level items we care about. But I think we could do two iterations. One time as is with strict enum checking. One time after that with `taxonony::dcast<>` to also factor in potential upgrades only when the strict version of the loop did not trigger any function. Thoughts?
> 
> Because the other thing is that for advanced breps and trimmed curves we need them as curves and not topological edges, so in those cases we need to "unpack" the edges than if in the mapping we already treat them as topological items. Where in fact the upgrade mechanism was intended for that.

Makes sense, I've added another iteration after strict iterator so it will try to upgrade item to edge if nothing from strict iterator worked.

aothms:
Thanks for your continued effort on this. Did you consider the following to make it more generic instead of only for edges?

**AbstractKernel.h**

```C++
template <size_t N>
struct dispatch_with_upgrade {
	static bool dispatch_with_upgrade(ifcopenshell::geometry::kernels::AbstractKernel* kernel, ifcopenshell::geometry::taxonomy::kinds item_kind, const ifcopenshell::geometry::taxonomy::ptr item, IfcGeom::ConversionResults& results) {
		auto concrete_item = ifcopenshell::geometry::taxonomy::template dcast<ifcopenshell::geometry::taxonomy::type_by_kind::type<N>>(item);
		if (concrete_item)
			return kernel->convert_impl(concrete_item, results);
		} else {
			return dispatch_with_upgrade<N + 1>::dispatch(kernel, item_kind, item, results);
		}
	}
};

template <>
struct dispatch_with_upgrade<ifcopenshell::geometry::taxonomy::type_by_kind::max> {
	static bool dispatch(ifcopenshell::geometry::kernels::AbstractKernel*, ifcopenshell::geometry::taxonomy::kinds, const ifcopenshell::geometry::taxonomy::ptr item, IfcGeom::ConversionResults&) {
		Logger::Error("No conversion for " + std::to_string(item->kind()));
		return false;
	}
};
```

***AbstractKernel.cpp*

```
try {
	return dispatch_conversion<0>::dispatch(this, item->kind(), item, results);
} catch (std::exception& e) {
	try {
		return dispatch_with_upgrade<0>::dispatch(this, item->kind(), item, results);
	} catch (std::exception& e) {
```

aothms:
`dcast` and `cast` work the same in this regard. But when a type T is requested we first check all upgrades whether they can deliver an upgrade from the runtime type to T. If not we fall back to a normal shared_ptr cast.

https://github.com/IfcOpenShell/IfcOpenShell/blob/9be392e2f1fd9dc3694468aab3a472e054d28533/src/ifcgeom/taxonomy.h#L1374-L1407

In the preexisting `dispatch_conversion` there will never be any upgrades, because of the strict unpolymorphic type check here https://github.com/IfcOpenShell/IfcOpenShell/blob/9be392e2f1fd9dc3694468aab3a472e054d28533/src/ifcgeom/AbstractKernel.h#L88C1-L88C25 where we compare the kind (enum) to an index (where the index is the tuple index into the tuple of equivalent types.

Andrej730:
> In the preexisting `dispatch_conversion` there will never be any upgrades, because of the strict unpolymorphic type check here [`9be392e`/src/ifcgeom/AbstractKernel.h#L88C1-L88C25](https://github.com/IfcOpenShell/IfcOpenShell/blob/9be392e2f1fd9dc3694468aab3a472e054d28533/src/ifcgeom/AbstractKernel.h#L88C1-L88C25) where we compare the kind (enum) to an index (where the index is the tuple index into the tuple of equivalent types.

So, we can replace it in `dispatch_conversion` with just `std::dynamic_pointer_cast<T>` + `throw std::runtime_error("Unexpected topology")` so it won't be trying all those upgrades?

Also, just noticed `dcast` is missing `curve_to_loop_upgrade` - is it intentional?

aothms:
> So, we can replace it in dispatch_conversion with just std::dynamic_pointer_cast<T> + throw std::runtime_error("Unexpected topology") so it won't be trying all those upgrades?

No I don't think we should do that. dispatch_conversion should be as specific as possible. These enums are not in any particular order. We don't want to blindly cast every loop to a face (e.g when we have a IfcGeometricSet/taxonomy::collection of loops). Only when in a specific interpretation function we require a face, but the mapping gives us a loop we use the automatic upgrade. 

> Also, just noticed dcast is missing curve_to_loop_upgrade - is it intentional?

Probably not intentional....

Andrej730:
> No I don't think we should do that. dispatch_conversion should be as specific as possible. These enums are not in any particular order. We don't want to blindly cast every loop to a face (e.g when we have a IfcGeometricSet/taxonomy::collection of loops). Only when in a specific interpretation function we require a face, but the mapping gives us a loop we use the automatic upgrade.

I meant something like this, so it still will be strict but we'll avoid trying to run upgrades on the item as upgrades wouldn't succeed anyway.

```c++
	template <size_t N>
	struct dispatch_conversion {
		static bool dispatch(ifcopenshell::geometry::kernels::AbstractKernel* kernel, ifcopenshell::geometry::taxonomy::kinds item_kind, const ifcopenshell::geometry::taxonomy::ptr item, IfcGeom::ConversionResults& results) {
			if (N == item_kind) {
				auto concrete_item = std::dynamic_pointer_cast<ifcopenshell::geometry::taxonomy::type_by_kind::type<N>>(item);
				if (!concrete_item) {
					throw std::runtime_error("Unexpected topology");
				}
				return kernel->convert_impl(concrete_item, results);
			} else {
				return dispatch_conversion<N + 1>::dispatch(kernel, item_kind, item, results);
			}
		}
	};
```

aothms:
> I meant something like this, so it still will be strict but we'll avoid trying to run upgrades on the item as upgrades wouldn't succeed anyway due to the strictness.

I think you're right. That's probably the sensible thing to do. static_pointer_cast is sufficient once we have asserted that N == item_kind.

The reason for the current behaviour is that I had experimented with several pointer types for the taxonomy items, and taxonomy::cast<> would pick the appropriate impl. Later I think the upgrades were bolted on. But replacing this with static_pointer_cast we should probably do.

Andrej730:
Should be fine to merge?

aothms:
Oh nice! You created a tuple of all potential upgrade targets to loop over. My idea was to blindly loop over the full tuple again, just with upgrades in place. Nice!

Just one small concern for later, maybe we should not be catching arbitrary std::exception, but only a specific non_implemented_error to distinguish between other logic errors. Although even in that case it wouldn't really hurt to try again with upgrades, it just feels bit unpredictable.

PR #5406, opened 2024-09-17
WIP UI changes

trhyder:
Partial support for draw wall added
Edit Railing Path icons added
Edit Roof Path icons added
Edit Profile icons added
Other relabelling tweaks


trhyder:

Moult:
It looks like there's a conflict in the workspace.py UI. I think this is because after your previous merge, I had restructured a bunch of the code to have two separate draw functions - one for the tool header, and another for the tool sidepanel (previously, it was a single draw function with lots of if statements). It looks like this restructure wasn't captured in your latest set of changes. Is this something you can do?

trhyder:
I may have done something similar. I'll check out how you approached it.

trhyder:
@Moult I think I approached it a bit differently.

I created separate classes, one for the selected tool, see 'class AddElementUI:` at line 311, and one for the tools that can be applied to the select element, see `class ModifySelectedUI:` at line 455. From my understanding, the only differences between the Header and side panel are the row spacings and the display of shortcuts as icons.

How would you like me to proceed?

trhyder:
Actually, on reflection, I did notice some different behaviors between the tool header and the side panel. At one point I tried to put a conditional new row in the `add_layout_hotkey_operator` function but I couldn't get it to work properly. Are you ok with accepting my pr, then I'll work on separating the header from the side panel?

Moult:
> From my understanding, the only differences between the Header and side panel are the row spacings and the display of shortcuts as icons.

Icons are also grouped into align=True portions. In the future, as more features arrive, I have a gut feel that the presentation will diverge more.

trhyder:
I see. I've made the separation to the EditObjectUI. I had already structured the code to have sub-draw functions to group like tools together. Is this ok for now?

image

Moult:
Cheers, I've merged it and kept your code. The UI is going to continue changing quite a bit with the new item editing UI that's in the works, so I recommend pulling the latest changes so that in the future there are no more merge conflicts.

PR #5435, opened 2024-09-21
Cancel and Apply buttons

trhyder:
Deleted custom green apply icons
Replaced custom green apply icon with CHECKMARK
Added cancel button


trhyder:

Issue #651, opened 2019-08-01
Regression: collada serialization breaks with non-ascii filenames on osx and linux


otsoa:

BIMvoice:
This is already handled on the current v0.8 line. IfcConvert works around the serializers that cannot open unicode output paths (Collada / DAE among them) by writing to a random ASCII-only temporary file and then renaming it to the real (possibly non-ASCII) target with the unicode-aware IfcUtil::path::rename_file (see IfcConvert.cpp around the "These serializers do not support opening unicode paths" comment and the rename_file call). So a non-ASCII .dae output path no longer fails. Recommend closing as resolved.

PR #5441, opened 2024-09-22
Add AssignConstituentFractions recipe

louistrue:
### Overview

This adds a new recipe `AssignConstituentFractions` to the `ifcpatch` utility, it assigns missing fractions to material constituents in an IFC file based on their thickness.

### Details

- **Recipe File**: `src/ifcpatch/recipes/AssignConstituentFractions.py`
  - Calculates thicknesses of material constituents from associated 'Width' quantities.
  - Assigns fractions to `IfcMaterialConstituent` entities based on these thicknesses using the `Fraction` attribute.

### Motivation

This recipe enhances the `ifcpatch` utility by allowing users to accurately represent the proportional contributions of material constituents based on their physical thickness in the 'fraction' attribute if it is being exported into a 'Width' quantity (as seen with applications like Revit and Vectorworks).

It explicitly addresses the issue discussed in [Autodesk/revit-ifc Issue #651](https://github.com/Autodesk/revit-ifc/issues/651), where the `Fraction` attribute for `IfcMaterialConstituent` entities is not exported, making volume calculations and material take-offs challenging.

#### **How It Works**

The recipe operates by:
1. **Identifying Material Constituents**: It finds all `IfcMaterialConstituentSet` entities in the IFC file and collects their `IfcMaterialConstituent` entities.
2. **Associating Elements**: It identifies IFC elements (e.g., walls) that are associated with these constituent sets via `IfcRelAssociatesMaterial`.
3. **Collecting Quantities**: It gathers `IfcElementQuantity` entities associated with these elements, which contain quantities like 'Width'.
4. **Mapping by Names**: It builds a mapping between the names of the material constituents and the names of the quantities.
5. **Extracting Thicknesses**: It extracts the 'Width' values from the quantities, assuming that the 'Width' quantity represents the thickness of the constituent.
6. **Calculating Fractions**: It calculates the fractions for each constituent based on their thickness relative to the total thickness.
7. **Assigning Fractions**: It assigns the calculated fractions to the `Fraction` attribute of the `IfcMaterialConstituent` entities.

#### **Limitations**

It's important to note that this approach relies on matching names between material constituents and quantities, which introduces some limitations:

- **Name Matching Dependency**: The recipe relies on the name as the only connection between a material constituent and its thickness quantity.
- **Duplicate Names**: In cases where multiple constituents or quantities share the same name, the recipe attempts to match them based on the order of appearance. This might not always reflect the actual relationships, potentially leading to incorrect fraction assignments.

### DIFF
--- a/src/ifcpatch/ifcpatch/recipes/AssignConstituentFractions.py
+++ b/src/ifcpatch/ifcpatch/recipes/AssignConstituentFractions.py
@@ -1,0 +1,159 @@
+# IfcPatch - IFC patching utility
+# AssignConstituentFractions Recipe
+# This recipe assigns fractions to material constituents in an IFC file based on their widths.
+
+import ifcopenshell
+from ifcopenshell.util.unit import calculate_unit_scale
+from logging import Logger
+from collections import defaultdict
+from typing import List
+
+
+class Patcher:
+    """
+    Assigns fractions to material constituents in an IFC file based on their widths.
+
+    This patcher calculates the widths of material constituents based on associated quantities
+    and assigns fraction values accordingly.
+    """
+
+    def __init__(self, src: str, file: ifcopenshell.file, logger: Logger, *args):
+        self.src = src
+        self.file = file
+        self.logger = logger
+
+    def patch(self):
+        """
+        Execute the patch to assign fractions to material constituents.
+        """
+        unit_scale_to_mm = calculate_unit_scale(self.file) * 1000.0
+
+        for constituent_set in self.file.by_type('IfcMaterialConstituentSet'):
+            constituents = constituent_set.MaterialConstituents or []
+            if not constituents:
+                continue  # Skip if no constituents found
+
+            # Find elements associated with this constituent set via IfcRelAssociatesMaterial
+            associated_elements = self.get_associated_elements(constituent_set)
+            if not associated_elements:
+                continue  # Skip if no associated elements found
+
+            # Collect quantities associated with the elements
+            quantities = self.get_quantities_from_elements(associated_elements)
+
+            # Build a mapping of quantity names to quantities
+            quantity_name_map = self.build_quantity_name_map(quantities)
+
+            # Calculate widths for constituents
+            constituent_widths, total_width_mm = self.calculate_constituent_widths(
+                constituents, quantity_name_map, unit_scale_to_mm
+            )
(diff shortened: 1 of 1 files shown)


louistrue:
Sample Ifc with constituent missing fraction: https://github.com/IfcLCA/IfcLCA/blob/085aebb1259cb4d33d232ab9446aba10c3871b6d/testfiles/4_RV_Str.ifc

louistrue:
👋 just following up on this: If there’s any feedback or changes needed, I’m happy to address them. Please let me know if anything needs clarification or further work. Thanks!

Moult:
Hey @louistrue ! Thanks for the recipe. There are a few comments:

1. Can you use util.element.get_elements_by_material?
2. Can you use util.element.get_pset(qto_only=True)?
3. Why is this hardcoded to use mm?
4. Is this specific to a problem for the ReferenceView fallback for layers vs constituent sets? That should probably be communicated somehow.
5. Missing license?

louistrue:
Thanks for the review, @Moult ! Let me address each point:

1. **Using `util.element.get_elements_by_material`**:
Done! I've updated the code to use `ifcopenshell.util.element.get_elements_by_material()` for finding associated elements. 

2. **Using `util.element.get_pset(qto_only=True)`**:
While I've tried using `get_pset`, I found that for this specific case I need to work directly with `IfcPhysicalComplexQuantity` to properly handle layer information. The challenge is that I need to:
   - Find quantity sets with specific "layer" discrimination
   - Access their `HasQuantities` to find width values
   - Match these with the material constituents
The current implementation of `get_pset` doesn't give us the full structure needed to reliably match layer quantities with constituents. I'd be happy to discuss how we might enhance the utility functions to better handle complex quantities in the future.

3. **Hardcoded mm**:
You're right, this was an oversight. I've updated it to handle unit conversions properly using `ifcopenshell.util.unit.calculate_unit_scale(ifc_file)` to ensure it works with the correct units regardless of the file's unit settings.

4. **ReferenceView specificity**:
Yes, this recipe specifically addresses how Reference View MVD handles material layers through constituent sets. I've updated the docstring to better explain this context and added relevant references to the MVD specification. Hope this helps users understand when and why they might need this recipe.

5. **License**:
I'll add the standard IfcOpenShell license header to match the project's licensing.

Issue #5168, opened 2024-08-09
Container Image aecgeeks/ifcopenshell:latest seems to have broken python package and IFCConvert binary


jonganej:
I wanted to quickly try out the Python API by downloading the container image as mentioned [in the documentation](https://docs.ifcopenshell.org/ifcopenshell-python/installation.html#docker) but running that seems to give me

```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/ifcopenshell/__init__.py", line 81, in <module>
    from . import ifcopenshell_wrapper
  File "/usr/lib/python3/dist-packages/ifcopenshell/ifcopenshell_wrapper.py", line 13, in <module>
    from . import _ifcopenshell_wrapper
ImportError: libpython3.10.so.1.0: cannot open shared object file: No such file or directory
```

I then tried calling `IfcConvert` but that threw

```
IfcConvert: error while loading shared libraries: libhdf5_serial_cpp.so.103: cannot open shared object file: No such file or directory
```

Are there perhaps outdated dependencies on the docker image?

aothms:
Thanks. I think the various ubuntu image version in this workflow ran out of sync. Let's see if my commits help in that regard.

tymokvo:
Hello, also just ran into the same error. I also noticed that there only ever seems to be the `latest` tag pushed to [Docker Hub](https://hub.docker.com/r/aecgeeks/ifcopenshell/tags). Maybe publishing images tagged to match the git tags would help to have a fallback in the future?

Would be happy to contribute, if needed.

aothms:
> Would be happy to contribute, if needed.

@tymokvo that would be amazing :)

> I also noticed that there only ever seems to be the latest tag ...

We discussed some of this here: https://github.com/IfcOpenShell/IfcOpenShell/issues/3927 I guess nobody really felt the urgency to properly apply and test this on dockerhub

See here how we use https://github.com/IfcOpenShell/IfcOpenShell/blob/df494a3defe1c06ac62db2c88c385702dea2c601/.github/workflows/ci-ifcconvert.yml#L61 `steps.version.outputs.version` for inspiration.

tymokvo:
Let me know what you think of #5463 !

Issue #3927, opened 2023-10-24
Switching to semver for v0.8.0


Moult:
One of the confusing aspects of IOS is the versioning scheme. The C++ core is broadly grouped into a version number 0.X.0 which represents at least to me mostly major architectural design changes. Within the 0.X.0 version there are many -sha1 versions which represent a rolling release of features and fixes.

Then in ifcopenshell-python this is combined with yet another rolling release ... but the rolling release depends on which package you're getting. Pip is a monthly "point in time". BBIM stable is a every 2 months "point in time". I'm not entirely sure what Conda is. And there is also IfcOpenBot which matches the point in time of the last C++ -sha1 version.

I guess v0.8.0 is an opportunity to clean this up a tad. I'd like to propose to follow semver (not the major.minor.patch semver but what semver calls the "0.y.z initial development phase") - see https://semver.org/ . Here's a loose proposal that follows the rules of semver:

 - C++'s first 0.8.0 official release is called 0.8.0
 - Subsequent IfcOpenBot builds are 0.8.1, 0.8.2, 0.8.3, etc.
 - Somehow make it so that when an IfcOpenBot build is done (and succeeds), it also bumps PyPI. This makes their "point in time" equal.
 - If PyPI's workflow runs independently of IfcOpenBot, it also increment the patch number. E.g. 0.8.4, 0.8.5, 0.8.6, etc.
 - BBIM daily builds continue using its own semver e.g. 0.0.YYMMDD as it's still in alpha
 - The PyPI release date is adjusted so that every two months it also coincides with the BBIM stable release date. This way BBIM's stable point in time matches PyPI's point in time and there is a consistent simple ascending name for versions without having to lookup sha codes.

It would be nice if this could also be consolidated with docker/conda.

Moult:
> We set the version number in git, which triggers a build.

You mean you create a tag, which then triggers an IfcOpenBot / workflow run? That sounds really nice and explicit.

> I think we should move out the [BBIM daily build] tags of the repo though.

Agreed. Perhaps only tag the stable release every 2 months?

aothms:
Yes, or create a separate repo blenderbim-releases if it's convenient. That just mirrors IfcOpenShell. Or maybe the repository would be the built artefacts (don't know).

Something else is the branching. Do you think a single active branch is still ok, i.e we can't fix old features without introducing the new ones?

Moult:
Agreed, a common source of complaints has been the build noise.

[...]

PR #5463, opened 2024-09-23
Derive docker tags from git refs

tymokvo:
## Summary
- Update docker workflow to use [`github.ref_name`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) for docker tags

## Description
See discussion in 
- #5168

The change here aligns the docker tag to the git tag, which seems to be aligned with the direction from #3927. 

The only tag that is available for `aecgeeks/ifcopenshell` on the Docker registry [is `:latest`](https://hub.docker.com/r/aecgeeks/ifcopenshell/tags). Because of the `on:` value of this workflow file, this means that either
- the last pushed tag
- the last manual dispatch

of this workflow will end up as the `:latest` tag. This is _strictly_ correct (i.e. `latest` means "it happened most recently"). But, it means that breaking changes like the issues from #5168 are hard to ameliorate.

In order to keep the `latest` tag aligned with the repository, I added the condition that the `latest` tag should only be applied when the git ref is the default branch (using the [`CSV` syntax from `docker-build-push@v2`](https://github.com/docker/build-push-action/tree/releases/v2?tab=readme-ov-file#inputs)). That seems to be the case with `v0.8.0`. Let me know if this is wrong. In order for this to work, I think that future release would need to:
- push the tag
- push a branch with the same name as the tag
- update the "default branch" in the repo settings
- re-run the workflow that was started from the tag event
  - the second time the workflow runs, the second expression should evaluate to `true` and append the second tag

Note that I can't test this and it relies on stringly-typed github workflow expressions. But I'm _pretty sure_ it will work 😬 .

### DIFF
--- a/.github/workflows/ci-ifcopenshell-docker.yml
+++ b/.github/workflows/ci-ifcopenshell-docker.yml
@@ -116,6 +116,8 @@ jobs:
       with: 
         context: artifacts
         repository: aecgeeks/ifcopenshell
-        tags: aecgeeks/ifcopenshell:latest
+        # Since the dispatch is set to `tag`, `github.ref_name` should evaluate to the pushed tag
+        # On a workflow dispatch, `ref_name` will take on the value from the dispatch payload
+        tags: aecgeeks/ifcopenshell:${{ github.ref_name }}${{ github.ref_name == github.event.repository.default_branch && ',aecgeeks/ifcopenshell:latest' }}
         file: ./Dockerfile
         push: true


aothms:
> But I'm pretty sure it will work 😬 .

Thanks for all your work on this and sorry for the delay in merging. Let's find out when we tag :)

PR #5479, opened 2024-09-24
fix: compile errors qt app

jascase901:
I was able to hardcode load an ifc file, by uncommenting the hard coded parse ifcfile, but I was unable to open and load files with the qt menu.

### DIFF
--- a/src/qtviewer/IfcViewerWidget.cpp
+++ b/src/qtviewer/IfcViewerWidget.cpp
@@ -1,5 +1,4 @@
 #include "IfcViewerWidget.h"
-#include <OpenGL/OpenGL.h>
 #include <osg/ShapeDrawable>
 #include <osg/Material>
 #include <osg/MatrixTransform>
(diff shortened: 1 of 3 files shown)


dushyant-basson:
> I was able to hardcode load an ifc file, by uncommenting the hard coded parse ifcfile, but I was unable to open and load files with the qt menu.

That's strange! Load-from-menu worked on Mac and I tested it on Windows too. Which platform are you running it on?

jascase901:
Thanks for merging so quickly.

> That's strange! Load-from-menu worked on Mac and I tested it on Windows too. Which platform are you running it on?

I am using ubuntu 20.04. I also have to run with QT_QPA_PLATFORM="xcb" environment variable set, to configure qt for x11 instead of wayland. 

The following patch fixes the menu loading, but it might make the dialog look uglier. I can submit a PR if you want. But this might just be a case of my computer being a snowflake.

```diff --git a/src/qtviewer/MainWindow.cpp b/src/qtviewer/MainWindow.cpp
index fa46ed2b8..8587737a2 100644
--- a/src/qtviewer/MainWindow.cpp
+++ b/src/qtviewer/MainWindow.cpp
@@ -96,7 +96,7 @@ void MainWindow::appendToOutputText(const QString& message)
 void MainWindow::openFile()
 {
     QString filePath = QFileDialog::getOpenFileName(this, tr("Open IFC File"), 
-        ".", tr("IFC Files (*.ifc)"));
+                                                   ".", tr("IFC Files (*.ifc)"), 0, QFileDialog::DontUseNativeDialog);
 
     QFileInfo fileInfo(filePath);
     QString fileName = fileInfo.fileName();
```

PR #5481, opened 2024-09-24
Changed tools tips for commands.

Slackking1:
The current tooltip for commands, was not very logical. This is a lot more explicit. Hopefully this contribution makes sense, still getting into this amazing projekt, further more svg2pdf_command lacked "". 

Change recommended application for layouts, since this creates from what i see the fastest workflow. Eg. layouts in inkscape, and created drawings in edge.


Moult:
Thank you!

PR #5532, opened 2024-10-03
Implement rounding of elevation

manuvarkey:
Implement rounding of elevation based on precision stored in IfcGeometricRepresentationContext


src/bonsai/bonsai/core/unit.py
```diff
@@ -113,3 +114,13 @@ def edit_unit(ifc: tool.Ifc, unit_tool: tool.Unit, unit: ifcopenshell.entity_ins
         ifc.run("unit.edit_named_unit", unit=unit, attributes=attributes)
     unit_tool.import_units()
     unit_tool.clear_active_unit()
+
+
+def format_value(ifc: tool.Ifc, value: float) -> str:
```
Andrej730, line 119: We don't store this kind of methods in `core` (`core` usually doesn't contain any implementations, it's just calling methods from `tool`), can you please move it to `tool.Unit`?

manuvarkey:
@Andrej730 have moved the formatting function to tools.

Andrej730:
Thanks for the contribution.

PR #5535, opened 2024-10-04
Progress on Selected Type Preview

trhyder:


trhyder:
I moved the Add button to the bottom to allow for a top-to-bottom reading - everything above the Add button contributes values to the object that gets added

image

trhyder:
Type Manager development

image

trhyder:
image

Moult:
I'm not too keen on [this change](https://github.com/IfcOpenShell/IfcOpenShell/pull/5535#issuecomment-2393875233). I considered it but the issue is that when there are many types and it is paginated, what should happen? Should it always keep one tile free just to show that "add button"? Or force the user to switch to the last page? I'd rather keep the pagination area and the "add" separate, as is in most other UIs.

trhyder:
Ok

trhyder:
image

trhyder:
@moult I think this is ready. I have moved the add button out of the paginated list and given it a text label

PR #5573, opened 2024-10-15
Type Manager

trhyder:
Toggle enable row with evil operators


trhyder:
No it doesn't. If you have multiple types it shows you which one is active by greying out the others. No change to the functionality.

Issue #5372, opened 2024-09-12
Show IFC Logo in BB Property Types in light mode


kurb70:
If BB displays the UI in light mode, the IFC logo of the first entry “disappears”. A light gray variant (unselected) should also be shown there.

![Screenshot 2024-08-29 um 08 33 36](https://github.com/user-attachments/assets/dc8b7e6f-6ee9-43f2-bcc0-523729dc72db)
![Screenshot 2024-08-29 um 08 33 24](https://github.com/user-attachments/assets/562ff576-d0d4-42a1-b7b3-5e66449272bf)

Moult:
I noticed @trhyder had implemented light/dark modes in his UI code. It would be good to centralise the functions for light/dark detection in tool/Blender and then that could be reused to fix this.

Thoughts @trhyder ?

trhyder:
Sounds good. I can try this one

trhyder:
Hey @Moult, would you please give me a more precise location where you would like to store this light/dark detection function? I don't understand what you mean by tool/Blender

Moult:
In this file: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/bonsai/bonsai/tool/blender.py

trhyder:
@kurb70, can you share your "white mode" theme so I can test my PR works for your use-case?

Moult:
Seems to work here

PR #5579, opened 2024-10-16
Dark / Light / White mode support for icons

trhyder:
Currently the function works in Dark mode and White mode but not Light mode. This is due to Light mode having two different types of backgrounds where the icons appear - one dark and one light

https://github.com/IfcOpenShell/IfcOpenShell/issues/5372


Moult:
The detect display mode should be done in a new Data class, not run on every draw call.

trhyder:
Thank you!

PR #5598, opened 2024-10-18
Install cityjson_converter lib and exe

cjmayo:
Fixes:

```
$ IfcConvert
IfcConvert: error while loading shared libraries: libcityjson_converter.so: cannot open shared object file: No such file or directory
```

---

This is installing on Linux with shared libraries.

I don't know what `cityjson_converter_exe` is, but it was easy to add installation here! Maybe that is not appropriate.


aothms:
Thanks for this.

> I don't know what cityjson_converter_exe is

Good question, we should really work towards more of a plug-in architecture so that these kind of more academic developments can be more loosely coupled.

PR #5606, opened 2024-10-20
python: Fix ifcopenshell.guid.new()

cjmayo:
pyException:

<class 'AttributeError'>: module 'ifcopenshell' has no attribute 'guid'

---

This is having instaled IfcOpenShell using cmake.

Else you have to `import ifcopenshell.guid` and the docs don't suggest that is necessary:
https://docs.ifcopenshell.org/ifcopenshell-python/hello_world.html


aothms:
Thanks :)

PR #5619, opened 2024-10-21
Update report.html

AversioCZ:
Adding "Skipped" grey colored container for specification and requirement if no applicable element is found. Example result can be seen in attached image.
GreyForSkipped


Moult:
Thanks! This is really much more informative than the previous "green" pass colour :)

PR #5625, opened 2024-10-22
Partially fix an issue with BIMModelProperties ifc_class enum.

brunoperdigao:
The issue can be reproduced as follow:
- Select the Multi Object Tool and change the ifc class to anything other than the first option.
- Now select other tool, e.g. wall tool, and the menu is not correct.

blender_SVUDEIfaCd

This PR proposes a partial solution. It solves the issue with the menu, but it's still getting this error:
` WARN (bpy.rna): C:\Users\blender\git\blender-v420\blender.git\source\blender\python\intern\bpy_rna.cc:1366 pyrna_enum_to_py: current value '8' matches no enum in 'BIMModelProperties', '', 'ifc_class' `

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/data.py
+++ b/src/bonsai/bonsai/bim/module/model/data.py
@@ -58,6 +58,7 @@ class AuthoringData:
         cls.data["ifc_classes"] = cls.ifc_classes()
         cls.data["ifc_class_current"] = cls.ifc_class_current()
         cls.data["relating_type_id"] = cls.relating_type_id()  # only after .ifc_classes()
+        cls.data["relating_type_id_current"] = cls.relating_type_id_current()  # only after .ifc_classes()
         cls.data["relating_type_name"] = cls.relating_type_name()  # only after .relating_type_id()
         cls.data["relating_type_description"] = cls.relating_type_description()  # only after .relating_type_id()
         cls.data["predefined_type"] = cls.predefined_type()  # only after .relating_type_id()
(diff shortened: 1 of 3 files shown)


Andrej730:
Nice! I think for the completeness we can add new item to .data, something like `AuthoringData.data["ifc_class_current"]` as we're already using the snippet below 3 times (once in data.py and 2 times in this PR).
```python
        ifc_classes = cls.data["ifc_classes"]
        if not ifc_classes:
            return []
        results = []
        ifc_class = tool.Blender.get_enum_safe(cls.props, "ifc_class")
        if not ifc_class and ifc_classes:
            ifc_class = ifc_classes[0][0]
```


And I think I've found what's causing the other console errors, we can replace `.ifc_class` in them with `ifc_class_current` to fix them too.
https://github.com/IfcOpenShell/IfcOpenShell/blob/7517414552f7cbf59781109b51f335a71cb432a8/src/bonsai/bonsai/bim/module/model/data.py#L110
https://github.com/IfcOpenShell/IfcOpenShell/blob/7517414552f7cbf59781109b51f335a71cb432a8/src/bonsai/bonsai/bim/module/model/data.py#L115
https://github.com/IfcOpenShell/IfcOpenShell/blob/7517414552f7cbf59781109b51f335a71cb432a8/src/bonsai/bonsai/bim/module/model/data.py#L131

brunoperdigao:
Thanks a lot @Andrej730
I'll push another commit with your suggestion.

brunoperdigao:
Hi @Andrej730, I've pushed a commit with the fix. However, I noticed now that the problem also occurs with `relating_type_id`. To reproduce this issue:
- In the demo file, select the wall tool, and in the type manager select the type WAL300 (fourth option).
- Then change to the slab tool. It should trigger the warning, because the slab tool only have two types.

I guess I can create a similar solution, but I'm wondering if there is a general approach I'm missing.

Andrej730:
Yeah, it is the same problem. I think there is no general solution unless there is a way to change data when the tool is changed, then we'd be able to validate and reset props. Otherwise, to safely access prop from ui.py/workspace.py we need to have some data.py cache that's validating the prop and replacing it with fallback value if it's invalid.

brunoperdigao:
@Andrej730 I think it's working now. Let me know if you think I can merge this.

Andrej730:
Can't test it right now but looks good, I think we can merge it.

PR #5659, opened 2024-10-30
Update unit.py

aothms:


RickBrice (approved):
@aothms I don't fully understand the changes, but the end result appears to be correct. The IfcLengthMeasure within IfcCurveSegment are now converted correctly. Thanks.

PR #5668, opened 2024-10-31
add favicon to ifcopenshell-python and bonsaibim docs

tibicen:
Self explanatory. With collaped tabs it is easier to identify between those to documentation tabs in browsers.


Andrej730:
Thank you for the contribution!

Andrej730:
Commited as 751741455 (note that we don't use https://github.com/IfcOpenShell/IfcOpenShell/tree/master branch, we use https://github.com/IfcOpenShell/IfcOpenShell/tree/v0.8.0).

Issue #3883, opened 2023-10-13
Build error in IfcPythonPYTHON_wrap


meerfrau:
I'm building the current master with gcc 13.2.1, Python 3.10 (sic), boost 1.83.0:
```
[ 99%] Building CXX object ifcwrap/CMakeFiles/_ifcopenshell_wrapper.dir/CMakeFiles/_ifcopenshell_wrapper.dir/IfcPythonPYTHON_wrap.cxx.o
In file included from /usr/include/opencascade/../ifcgeom_schema_agnostic/../ifcgeom_schema_agnostic/../ifcgeom_schema_agnostic/IfcGeomFilter.h:34,
                 from /usr/include/opencascade/../ifcgeom_schema_agnostic/../ifcgeom_schema_agnostic/IteratorImplementation.h:25,
                 from /usr/include/opencascade/../ifcgeom_schema_agnostic/IfcGeomIterator.h:61,
                 from ./IfcOpenShell/build/ifcwrap/CMakeFiles/_ifcopenshell_wrapper.dir/IfcPythonPYTHON_wrap.cxx:3841:
/usr/include/boost/foreach.hpp: In function 'boost::foreach_detail_::auto_any<boost::iterators::reverse_iterator<T*> > boost::foreach_detail_::rbegin(auto_any_t, type2type<T*, C>*, mpl_::true_*)':
  818 |     while(0 != *p)
      | 
./IfcOpenShell/build/ifcwrap/CMakeFiles/_ifcopenshell_wrapper.dir/IfcPythonPYTHON_wrap.cxx: In function 'PyObject* _wrap_Triangulation_item_ids(PyObject*, PyObject*)':
./IfcOpenShell/build/ifcwrap/CMakeFiles/_ifcopenshell_wrapper.dir/IfcPythonPYTHON_wrap.cxx:13083: error: 'const class IfcGeom::Representation::Triangulation' has no member named 'item_ids'
13083 |       result = (std::vector< int,std::allocator< int > > *) &((IfcGeom::Representation::Triangulation const *)arg1)->item_ids();
      | 
./IfcOpenShell/build/ifcwrap/CMakeFiles/_ifcopenshell_wrapper.dir/IfcPythonPYTHON_wrap.cxx: In function 'PyObject* _wrap_new_Triangulation__SWIG_1(PyObject*, Py_ssize_t, PyObject**)':
[...]

PR #5697, opened 2024-11-05
Fix ifcwrap searching for headers in system directories

cjmayo:
This fixes building IfcOpenShell 0.7 when 0.8 is installed for me. Which looks like at least the start of #3883.


Andrej730:
Thank you!

PR #5698, opened 2024-11-05
Install headers under CMAKE_INSTALL_INCLUDEDIR/ifcopenshell

cjmayo:
IfcOpenShell is installing a number of directories and individual files into CMAKE_INSTALL_INCLUDEDIR.

set(INCLUDEDIR include) is redundant because the next command checks for relative paths.

---

Is there something I missed about why `set(INCLUDEDIR include)` is there?

This would also need applying to svgfill.

I guess it will break things for C++ users, and they would need to start using `-I/usr/include/ifcopenshell`. (or setting `-DCMAKE_INSTALL_INCLUDEDIR` when building)
A pkgconfig file could be created to help?


theoryshaw:
Hi @cjmayo,

PR #5699, opened 2024-11-05
Fix relative header include paths

cjmayo:
> For example, if /usr/include/sys/stat.h contains #include "types.h", GCC looks for types.h first in /usr/include/sys, then in its usual search path. 

https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-incform

Is there a reason why it is the way it is?


cjmayo:
I think this is maybe too much of a halfway house, leaving a lot of ".." still in place.
And that the extra directory in #5698 would provide a guard against those "escaping" the ifcopenshell directories.

PR #5740, opened 2024-11-12
Align unsupported schema message with VS

Ghesselink:
Error message aligned with 
https://github.com/buildingSMART/ifc-gherkin-rules/blob/development/features/IFC101_Only-official-ifc-versions-allowed.feature


aothms:
Thanks, but I'd prefer a strong separation between validation service and ifcopenshell

PR #5781, opened 2024-11-21
Github Action build scripts!

aothms:
Previously IfcOpenBot was a collection of physical machines and SSH/ansible scripts that ran on Amazon EC3.

I've rewritten them to run on Github Actions instead, still levering mostly the same build scripts that have been aging rather well.

You can expect a speed especially on windows because we (a) no longer need to install MSVC on an otherwise empty AMI (b) cache the compiled dependencies using git lfs in a separate repo (you can checkout yourself as well for speedier builds) and (c) because we now compile the python versions in a parallel matrix.

We only still need to figure out a place where to post notifications when builds are complete.

### DIFF
--- a/.github/workflows/build_osx.yml
+++ b/.github/workflows/build_osx.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Checkout Build Repository
         uses: actions/checkout@v3
         with:
-          repository: aothms/build-outputs
+          repository: IfcOpenShell/build-outputs
           path: ./build
           ref: ${{ matrix.os }}-${{ matrix.arch }}
           lfs: true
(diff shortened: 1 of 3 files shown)


aothms:
All,

> We only still need to figure out a place where to post notifications when builds are complete.

I changed the approach a bit, not to post messages when builds are complete, but rather list the contents of our s3 bucket and format these (should still add a cron workflow dispatch to this).

https://github.com/IfcOpenShell/build-listing/blob/87d548c6121b31276f7b353ea9b3b09d7c3bc104/.github/workflows/deploy.yml#L44-L49

Published here:

https://builds.ifcopenshell.org/

It doesn't look very cute yet and could be better integrated into our docs / download page on the website. All help welcome :)

PR #5798, opened 2024-11-25
First PR, tooltips for the drawing module added

jes-r:
Hope I did this the right way, if not please tell!

I now created a fork to my own github, but maybe it's better to create a branch?

I'm a beginner so please let me know!


Andrej730:
Thank you for the contribution.

> I now created a fork to my own github, but maybe it's better to create a branch?

Doesn't matter.

Issue #5817, opened 2024-11-27
a mode that indicated which storey was the default story. 


theoryshaw:
Would be nice to have some mode that indicated which storey was the default story. 

![image](https://github.com/user-attachments/assets/13364fa6-144f-4f50-bb77-779c031b318a)

sboddy:
Isn't the label at the top enough? I mean one _or_ the other, ok, but both? Seems to be unnecessary clutter. Speaking in a hypothetical, a project with many buildings/storeys, I think the label at the top is more usable because it is always visible. An indicator as you describe may often not be visible, so would be bad if it were the only indicator. Also, what if the building is collapsed. Now you'd have to expand everything trying to find the default container.
(Right now that top label is a little glitchy for me, and not updating when I set a new default although this _was_ working, but that is a separate issue.)

theoryshaw:
I see my confusion, it seems the field is not being updated as intended.

https://github.com/user-attachments/assets/6e151a92-f14a-4ef0-93ef-e0f2cb3c6ba0

PR #5829, opened 2024-11-30
Fix default container label #5817

sboddy:
Not sure if this is the "right way" to fix this, but it is simple and it works. There is effectively a get'er function  that works, and a stored value that never updates when you change the default container. The label was using the stored value. So perhaps the better fix is to update the stored value on change of def. cont. rather than calling the default_container() function every poll of the UI. I was struggling a bit to follow the path through the set_default_container functions. Seemed to bounce around a bit. So I picked the easy way.


Andrej730:
Indeed data.py is intended to cache the IFC data to avoid accessing IFC on every UI draw call, so cached container name should be updated during `set_default_container` to avoid this. Commited an alternative fix in 8609bd1.

Issue #3760, opened 2023-09-15
[BlenderBIM-IFC4-feature-request] Not all elements from the bSI IFC4 structural-curve-member.ifc example file are displayed in BlenderBIM.


Martin15135215:
Here I found the example:

https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/link/structural-curve-member.htm

There is this ifc file:

https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/annex/annex-e/structural-curve-member.ifc

These pictures show how the ifc file looks when opened in an ifcviewer. I also got these pictures from the bSI documentation. I myself do not know in which ifcviewer this was opened.

![](https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/figures/examples/structural_analysis_curve-1.png)

![](https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/figures/examples/structural_analysis_curve-2.png)

When I open the ifc file in BlenderBIM version https://github.com/IfcOpenShell/IfcOpenShell/releases/tag/blenderbim-230915, I only see four corner points:

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/110968398/256f3a17-fedb-47d1-83fb-7b40035cc9c0)

I briefly searched to see if there was a button so I could see the model as in the previous pictures, but didn't find it.

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/110968398/b06b348e-630b-47cd-97c1-4d081d662a08)

What I want in BlenderBIM is that all elements from the Structural analysis model are somehow displayed. For now, I am only interested in looking at the Structural analysis model. Therefore, I don't need to be able to edit it in BlenderBIM at the moment. In the future that would be cool too.

Hence my feature request to view Structural analysis model in BlenderBIM.

PR #5887, opened 2024-12-18
Decorator to show structural load representation in 3D view #3760

LM-Nascimento:
https://github.com/user-attachments/assets/5d794bea-ccaa-4545-aa5d-38e27427eb48

This can show the structural loads in 3D view, still has a few issues but it kinda works.


theoryshaw:
very cool!
would be awesome to incorporate something like this into BB, as well. 
https://community.osarch.org/discussion/2628/anastruct-2d-structural-analysis-in-python/p1

theoryshaw:
can you share your `structural-curve-member - with_point.ifc` file?  Would like to play around with it, if possible.

LM-Nascimento:
[structural-curve-member - with_point.ifc.txt](https://github.com/user-attachments/files/18189547/structural-curve-member.-.with_point.ifc.txt)
This is the file that I used. It is the example file in the IFC documentation, I added a point load and modified values for the nominal load (it is possible to edit the values using blender bim UI).

I also tested it with building 01 and building 02 from https://github.com/IfcOpenShell/analysis-models/tree/master/ifcFiles

I think we should improve structural model authoring workflow before attempting to integrate structural analysis into BB.

Moult:
Incredible! It's also an absolute behemoth of a PR so it's difficult to review, so I'm opting for the James Bond "merge first ask questions later" approach.

At a glance, I suspect there is a possibility to do some work in util.unit and use that more - right now it looks as though there is some duplication going on.

The `load_decoration_data.py` also sticks out as an anomaly - it's a new file that no other module has a similar situation. It doesn't mean it's wrong, just that it's something new :) Maybe there is also the ability have shared decoration code (if at all relevant).

Agree that structural model authoring needs to improve first before dealing with structural analysis.

LM-Nascimento:
Sorry about the size, I wasn't sure I could make it work, so I wanted to have a working prototype, that turned out to be huge. I have no background in programming, so I suppose that there is a more concise way to make this.
There is indeed duplication in the units part, the duplicate code in this PR isn't used, I just forgot to delete it.
load_decoration_data.py is responsible for getting the necessary information from the loads and members, there are many different types of loads, especially in curve members, so I decided to make a separate file due to the sheer number of lines. This decoration data  is very dependant on the custom shaders I wrote, so it would be difficult to share with other decoration code.

PR #5889, opened 2024-12-19
add new alignment function types to ifcopenshell_wrapper

civilx64:
@aothms @RickBrice this is at least a step in the right direction.  `create_shape` is not working yet but I can retrieve the new alignment function types on the python side


RickBrice:
I made the same changes (plus two or three more new function types) and committed them on my branch this morning.
What is the correct procedure here? Close the commit without merging?

civilx64:
And here I thought I was getting all fancy :grin:.  At least I found the right spot to add a few lines and got everything to re-compile - that's progress I guess!

I will close this PR.  My edits are on a branch on my fork so they won't pollute anything you are doing here in upstream.

PR #5894, opened 2024-12-20
updated restriction types

matkufe:
Could not create valid xml if one of "maxExclusive", "maxInclusive","minExclusive", "minInclusive", "totalDigits", "fractionDigits" occured in a restriction inside the ids data. Because these are floating values and can not handled, if they are of type string. They should handled the same way as "length", "minLength", "maxLength", right?


Moult:
Thanks! You're absolutely right. I think originally there was a limitation in IDS to only support a subset of restrictions.

Issue #5941, opened 2025-01-08
The creation of a parametric door, doesn't seem to work.


theoryshaw:
The creation of a parametric door, doesn't seem to work.

https://github.com/user-attachments/assets/5ec1a282-5e40-4199-9682-d0f0fd82341a

steverugi:
@theoryshaw 
odd, it works OK on my 311-0.8.1-250105 Windows

I noticed one thing though: 
When no doortype (or any other element) is available the `Add Type` button is blue, clicking on it triggers an error (obviously) 
I think it might be confusing
Also, contrary to its text, it doesn't add a type but an instance, am I correct?

![image](https://github.com/user-attachments/assets/9baee9d9-c311-4a60-a8aa-93f5d1b3c958)

thanks

Gorgious56:
I had this problem, too, reinstalling a fresh bonsai fixed it.

Gorgious56:
@steverugi I think the two bottom buttons should be changed, the leftmost one should just say "OK" and validate and close the panel, and the right most one should actually cancel any edition that was done. Or only an "OK" button to close the panel if it's not possible. Right now when you click Add Type you can't actually know which Type will be instanced so it's hit and miss, if you want to select a type it closes the panel so you can't use the blue button. The Cancel button doesn't cancel the edition so its label is misleading, it validates the edition an closes the panel.

I think clicking on a type should close the manager an go into the instanciation mode directly.

I think it's safe to do these changes, this panel is a remnant of many iterations on the type manager when there was not "SHIFT + A" shortcut or icon in the toolbar to instance an occurrence. The type manager should only allow for browsing, and editing types, not instancing them IMO. 

What are your thoughts ?

steverugi:
> @steverugi I think the two bottom buttons should be changed, the leftmost one should just say "OK" and validate and close the panel, and the right most one should actually cancel any edition that was done. Or only an "OK" button to close the panel if it's not possible. Right now when you click Add Type you can't actually know which Type will be instanced so it's hit and miss, if you want to select a type it closes the panel so you can't use the blue button. The Cancel button doesn't cancel the edition so its label is misleading, it validates the edition an closes the panel.

Yes, glad someone else noticed it :)

> 
> I think clicking on a type should close the manager an go into the instanciation mode directly.

[...]

PR #5947, opened 2025-01-09
Update Type Manager

Gorgious56:
Re #5941

Add back the type dropdown and looking glass icon in the toolbar Clicking on a type in the manager no longer closes the popup anymore. A visual indictor is displayed on the active type
Changed the parameters icon in the individual type interface

Fix bug in "Slab from Walls" Shift + A code


theoryshaw:
hello old friend. ;)

theoryshaw:
would be the _cat's pajamas_ too, if the filter list was located in the launcher too... so it could filter down the thumbnails. :)

image

Gorgious56:
Re #5941 
Here's a video illustrating the change. Any feedback before merging @Moult @theoryshaw @Andrej730 @steverugi ?

It takes a bit more space in the toolbar than before but I felt just the type manager icon wasn't visible enough.

I haven't tested the filter function on a huge database so I don't know how it fares with hundreds or thousands of types, since it triggers a bunch of redraws and recalculating of AuthoringData it might be bad. On a hundred Wall types I didn't notice any hiccup though.

Animation3

Could integrate more elaborate search terms like regex or fuzzy search (might be able to use builtin Blender search functionality) afterwards

Moult:
Things I like:

 -  It's now clear that clicking an item makes it the active type because you immediately see the "Active" text appear.
 - The new search filter.

Things I don't like:

 - There are two ways to change the active type.
 - There are two ways to add an occurrence.
 - The interface of a dropdown, followed by a search icon, followed by "fake dropdown button thing". It looks weird. I really don't have a good solution though.

Gorgious56:
Thanks for the feedback ! Another attempt.

- Went back to previous toolbar : icon + type name + dot grid icon which all launch the type manager.
- Validating the type manager popup now only closes it. Functionally is the same thing as clicking Cancel. Don't really know what to do to fix it. Does cancel make sense in the type manager ? Do we want to be able to undo what is possible to be done from within the type manager ?
- You can change the active type either with the dropdown in the popup and in the grid flow. I changed it so that changing the type in the dropdown jumps to the corresponding page. BTW first time using github copilot to figure out the maths and it worked magically :) . I'm not against removing the dropdown but I feel like it's nice to see a plain list to choose from.
- You can input a specific page number or scrub the page number field. Might be usefull if user has dozens of pages to flip through.
- Right now the search is very simple, it does not implement fuzzy search or anything fancy with *

I noticed a bug in the N panel with the icon generator which was caused by this PR.

Cheers

Animation3

Moult:
Awesome! This is the best of both worlds! A pure text dropdown, useful for those with many/complex types that use names to get around, a filter to aid it, as well as visual thumbnails all in one spot!

> Do we want to be able to undo what is possible to be done from within the type manager ?

Nah :)

The Close / Cancel is still awkward but isn't hurting anybody now. Sooner or later we'll get some inspiration of what to do with it. Feel free to merge!

PR #5957, opened 2025-01-11
Remove mathutils import from shape_builder.py

aothms:
It's imported below in the type_checking conditional, can it be removed from the top-level imports?


Andrej730:
Yeah, it was an overlook, thanks.

PR #5958, opened 2025-01-12
Icon dark/light mode function

trhyder:
New function that uses the color of text to determine if custom icons should be dark mode (dm) or light mode (mode). Applied to model and cad workspaces.


Moult:
Neat! Let me know when it's ready for review / merge.

trhyder:
@Moult The next step will be implementing this in the bonsai properties panel - but I might need some guidance. For now this works on the cad and model workspaces.

PR #5961, opened 2025-01-12
Bonsai offers to install Git and ifcmerge

brunopostle:
*****NOT READY FOR MERGING***** This has been tested on Linux, but not at all on Windows, so I have no idea if the Windows code works at all, and I don't really have a way to test it as I don't have access to a Windows development system.

How it works: if Git isn't installed on Windows, the Git panel now offers to install it from the Microsoft Store using `winget`.

Then if ifcmerge isn't installed (on either Linux or Windows), the Git panel offers to install it from copies bundled in the Bonsai ZIP package.

Note that ifcmerge.exe increases the size of the Bonsai ZIP download by about 7MB.

Note that this creates a new folder on Windows: `%USERPROFILE%\AppData\Local\Bonsai\bin` and adds it to the %PATH% (or will if it actually works). `%USERPROFILE%\AppData\Local\Bonsai` is where all the various Bonsai templates and stuff ought to be installed eventually.

### DIFF
--- a/src/bonsai/bonsai/bim/__init__.py
+++ b/src/bonsai/bonsai/bim/__init__.py
@@ -258,6 +258,9 @@ def register():
     icons = icon_preview
     bpy.app.translations.register("bonsai", translations_dict)
 
+    import bonsai.tool as tool
+    tool.Blender.ensure_bin_in_path()
+
 
 def unregister():
     global icons
--- a/src/bonsai/bonsai/bim/__init__.py
+++ b/src/bonsai/bonsai/bim/__init__.py
@@ -259,6 +259,7 @@ def register():
     bpy.app.translations.register("bonsai", translations_dict)
 
     import bonsai.tool as tool
+
     tool.Blender.ensure_bin_in_path()
 
 
--- a/src/bonsai/bonsai/tool/blender.py
+++ b/src/bonsai/bonsai/tool/blender.py
@@ -21,7 +21,6 @@ import bpy
 import bmesh
 import json
 import os
-import platform
 from ifcopenshell import entity_instance
 import ifcopenshell.api
 import ifcopenshell.util.element
@@ -437,11 +436,7 @@ class Blender(bonsai.core.tool.Blender):
         bin_dir = str(Path(__file__).parent.parent.resolve() / "libs" / "bin")
         current_path = os.environ["PATH"]
         if bin_dir not in current_path:
-            if platform.system() == "Windows":
-                path_separator = ";"
-            else:
-                path_separator = ":"
-            os.environ["PATH"] = current_path + path_separator + bin_dir
+            os.environ["PATH"] = current_path + os.pathsep + bin_dir
 
     @classmethod
     def get_default_selection_keypmap(cls) -> tuple:
--- a/src/bonsai/bonsai/tool/blender.py
+++ b/src/bonsai/bonsai/tool/blender.py
@@ -21,6 +21,7 @@ import bpy
 import bmesh
 import json
 import os
+import platform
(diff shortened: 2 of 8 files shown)


brunopostle:
I'm reconsidering this. Installing ifcmerge like this is ok on Linux where you expect `~/.local/bin` to be in your `PATH`; and it ought to be ok on Windows where `%PATH%` is a registry setting (but permissions might be tricky); but it doesn't look like there is a reliable way to get this to work on Darwin.

An alternative would be to keep ifcmerge in the Bonsai installation folder and modify the PATH to include this folder at runtime for the current session. This should work without messing around with the system.

Also the Windows command to install Git is wrong, it should be:

    winget install --id Git.Git -e --source winget

This appears to not require Administrator permissions, it would be nice to have confirmation.

brunopostle:
I think this is now ok, if there are no objections (@Moult @Andrej730) I'll merge it.

The ZIP installer now contains a `libs/bin` folder that contains `ifcmerge`. Bonsai on `register()` adds this folder to the system `PATH` for the current Blender session. Advantage of this is that we don't have to fiddle with the registry on windows, or install files to `~/.local/bin` on Linux, it _should_ work on Darwin, updates are automatic, and we have a mechanism to ship other executables if required.

(Note that ifcmerge.exe increases the size of the Windows Bonsai ZIP download by about 7MB)

If Git isn't installed on Windows, the Git panel now offers to install it from the Windows Package Manager Community Repository using `winget`.

Andrej730:
Looks good, ifcgit is very important and we need to make sure it's as easy as possible to get it for new users to take a shot on using it.

src/bonsai/bonsai/tool/blender.py
```diff
@@ -430,6 +431,18 @@ def ensure_blender_path_is_abs(cls, blender_path: Path) -> Path:
             return blender_path
         return bpy.path.abspath("//") / blender_path
 
+    @classmethod
+    def ensure_bin_in_path(cls) -> None:
+        """Check 'bin' folder is in PATH, if not add for this session"""
+        bin_dir = str(Path(__file__).parent.parent.resolve() / "libs" / "bin")
+        current_path = os.environ["PATH"]
+        if bin_dir not in current_path:
+            if platform.system() == "Windows":
```
Andrej730: Small thing - you can also get path separator with `os.pathsep`.

src/bonsai/bonsai/bim/__init__.py
```diff
@@ -258,6 +258,10 @@ def register():
     icons = icon_preview
     bpy.app.translations.register("bonsai", translations_dict)
 
+    import bonsai.tool as tool
+
+    tool.Blender.ensure_bin_in_path()
```
Andrej730, line 263: Should we also clean up `PATH` on `unregister`?

Moult:
Looks great on my end feel free to merge :)

src/bonsai/bonsai/tool/blender.py
brunopostle: Cool, now using os.pathsep

src/bonsai/bonsai/bim/__init__.py
brunopostle: I feel confident appending to the PATH, but not so much removing items. It will only be appended once even if register() unregister() is called repeatedly, and it is transient anyway - so it will be cleaned if Bonsai is uninstalled and blender restarted.

Andrej730: Oh, true, for some reason I thought it's changing PATH not just for the current session. 😅

Issue #5375, opened 2024-09-12
Bonsai - feature request - dedicated Railing & Roof tool


Plae-Blue-Dot:
The railing and roof modifiers are generally quite good and mature but they don't have default tools in the UI

I believe a roof tool icon is really important to round off the tool set
![image](https://github.com/user-attachments/assets/e8664d28-efb1-4722-a268-d3dbb1494949)
something like this? just having a dedicated tool for these means more people will use them

And a railing tool aswell which could later be part of a set of tools like the MEP tool has become.

Here is the SVG:
![roof and railing icons inkscape](https://github.com/user-attachments/assets/0bdd8210-f597-4f61-9ad6-86f6b041d4e6)

SavuGust27:
even if i don't use those toll much i agree and i would add that a dedicated icon to acces IfcFurniture would also help.

Moult:
I've just merged in @trhyder 's new roof and railing icons :)

@trhyder is it possible to also design a furniture icon? Then add a new "Furniture" icon to the workspace (as the default), and then collapse roof and railing as a submenu (similar to MEP duct/pipe/tray/cable)?

trhyder:
@Moult are you open to a different grouping of tools? Below is a suggestion with current and possible future tools 

<img width="598" alt="image" src="https://github.com/user-attachments/assets/afdf173b-b0ec-4dad-9bf4-ae961beaaf93" />

Moult:
Looks great, go for it!

trhyder:
@Moult I believe this one can be closed now

PR #5983, opened 2025-01-16
New Tools

trhyder:
- New icons: Opening, Furniture, Roof, Railing
- New groupings
- New tools: Opening tools (not working), Furniture (needs default), Roof
- New defaults: Railing, Roof (Hip Roof doesn't work)


trhyder:
#5375

Moult:
Nice! This is set to draft, let me know when ready to merge.

trhyder:
@Moult this is probably not ready for merge, but is ready for review and feedback. I tweaked some of the existing icons (I hope this doesn't offend anyone) I have tried to make some of them simpler and clearer. I believe I have achieved this, but I welcome feedback.

My open question is about tools like furniture, sanitary, light, and Appliances: What should the quick create button do? For the moment I have it just generating a square mesh.

Screenshot 2025-01-19 at 4 36 57 PM

Moult:
Subjective here, im not a fan of the changes to wall, slab, door, window, and cable. They seem too thin, considering icons can change size, and later on may have smaller icons.

I also prefer the original column because now this looks more like a downpipe.

There absolutely needs to be one for IfcMember, how about an equal angle? Commonly used for bracing.

Chair is great but would prefer if its thicker. Toilet is great, light is great. 

Id like to group stair and ramp with slab and roof to save space.

Quick create cube makes sense.

trhyder:
Thanks for the feedback.

You're right about some of the revised icons looking too thin. I'll have one more go at refining them. I want to try the window and door icons with the same 3D view as the other icons. For the column, I'd like to remove the base as, to me, it looks like a foundation element.

Great idea the IfcMember to be an equal angle - I'll add that. I'll also add an IfcPlate to the same group

I'll beef up the Chair and have another go at an electrical appliance - I had a toaster, but maybe a washing machine might be better

I'll group the stairs with the slab and roof as you request. Should there also be different tools for IfcStairFlight and IfcRampFlight?

trhyder:
I've also started to work up some additional tool icons; these can be seen in the blend file.

Is the goal to provide a tool for every ifc type?

image

Moult:
These are really quite beautiful now!

I don't think it makes sense to create an icon for every single class. There are hundreds of classes. Instead, we should target only the most common, and then generic "BIM Tool" can be used as a fallback.

The logic I'm following is that the top-level tools should have the minimum to design a small building, with a bias to the architectural discipline first, followed by structure, then followed by MEP.

This means that architecture should have, at their fingertips, walls, slabs, doors, windows, roof.

Structural should have at their fingertips, column, beam, member, footing, pile.

MEP should have duct, pipe, cable tray.

Everything else is just a bonus, for example furniture is a bonus, outlets is a bonus.

What's the difference between the lightbulb and the ceiling lamp? What's the thing to the right of the washing machine?

Stairs are an interesting one. Right now, we only do stair flight and ramp flight, we don't do stairs and ramps, so maybe put the full stair / ramp icons on hold until we build a feature for that.

I don't think fasteners deserve to be a top level icon although it looks very, very cool. Same with alarms and solar panels. There seems to be an electrical bias here, perhaps because it's more domestic. In larger jobs, mech would perhaps priority and major equipment would be DBs, SBs, dampers, fans, pumps, etc. All in all, we can leave it all to the BIM tool because we'd never cover everything.

trhyder:
I am glad you like them and thanks for the explanation on the direction.

I think the latest now only includes only the important icons - let me know if I should add any of the others that are in the blend file

The light bulb was for IfcLamp as opposed to IfcLightFitting (not in use)

Next to the washing machine  is a projector for IfcAudioVisualAppliance (just a wip piece in the blend file only)

Screen Recording 2025-01-26 at 7 10 31 PM

Moult:
This looks awesome! Let me know when it's ready to merge.

The IfcLamp is probably not required - it's quite a specialist item and actually represents a light source (not the actual light fixture) typically used for lighting simulation, not in traditional drafting.

trhyder:
@Moult, I am just testing the quick create defaults. I can't get Curtain Wall to work - should I remove this icon?

trhyder:
I removed the curtain wall icon and believe it's ready to merge

Issue #5991, opened 2025-01-17
Base64 is non-standard and should be put in line with the standard schema


raj-open:
IfcOpenShell uses a non-standard order of alphanumerical characters for their Base64<->Hex encoding of UUIDs/GUIDs.

- The Base64 standard is
    ```txt
    ABCD...Z abcd...z 01...9 +/
    ```
    and right-padding with `=`-symbols (so that the length is 0 mod 4).

    See

    - <https://base64.guru/learn/base64-characters>;
    - p. 5 in <https://www.rfc-editor.org/rfc/rfc4648.txt>.

- IfcOpenShell by contrast uses
    ```txt
    01...9 ABCD...Z abcd...z _$
    ```
    
    and no padding.

    See

    - <https://technical.buildingsmart.org/resources/ifcimplementationguidance/ifc-guid>.

Now, the final characters can be arbitrarily set and the use of padding is maybe neither here nor there (though imo one should use it). **However, what is in my opinion a serious problem** is this deviant _order of the alphanumerical symbols_. This makes base64 GUIDs from IfcOpenShell incompatible and difficult to work with in conjunction with other systems. What do I mean by this?

## Example ##

Suppose in other applications (possibly not written in either python or C) one has to convert uuids between hex and base64 and some of these come from ifc files and IfcOpenShell.

Then instead of using standard methods there, one has to reverse-engineer IfcOpenShell's peculiar convention and write a special encoder/decoder there. This is not good practice.

## Proposed Solution ##

**Pull Request** #5990 proposes a change to align IfcOpenShell with other systems. See in particular the file [guid.py](https://github.com/raj-open/IfcOpenShell/blob/guid-standards/src/ifcopenshell-python/ifcopenshell/guid.py) and the methods `compress` and `expand`.

PR #5995, opened 2025-01-18
Guid-StdLibs ---> v0.8.0

raj-open:
This PR does not change the functionality of the methods `compress` and `expand`, but improves the underlying computations.

The Base64 <-> Hex conversions here rely on standard python libraries/methods (`base64.b64encode` and `base64.b64decode` as well as `bytes.fromhex`) instead of manual computations.  This makes the algorithms more transparent compared to the previous state. This should also entail a performance boost, though that has yet to be tested.

In addition we have provided docstrings and added references to Ifc documentation.


Moult:
I'm not entirely sure how this change is a benefit? The previous code seems pretty clear and efficient to me. Am I missing something? Ping @aothms.

raj-open:
> I'm not entirely sure how this change is a benefit? The previous code seems pretty clear and efficient to me. Am I missing something? Ping @aothms.

Qualitatively there is a significant different between the old and new set of methods. The latter is much more transparent and robust.

Empirically, there is a **significant performance boost**. Using the `codetiming` module, I ran the old and new methods and the differences are as follows:

|  old/new  | method   |   num operations |   time (s) |   new : old |
|:---------:|:---------|-----------------:|-----------:|------------:|
|    old    | compress |               10 |   6.49e-05 |             |
|    new    | compress |               10 |   1.35e-05 |      20.86% |
|    old    | compress |              100 |   0.000498 |             |
|    new    | compress |              100 |   0.000106 |      21.38% |
|    old    | compress |            1e+03 |    0.00506 |             |
|    new    | compress |            1e+03 |    0.00102 |      20.14% |
|    old    | compress |            1e+04 |     0.0504 |             |
|    new    | compress |            1e+04 |     0.0104 |      20.70% |
|    old    | compress |            1e+05 |      0.502 |             |
|    new    | compress |            1e+05 |      0.104 |      20.73% |
|    old    | compress |            1e+06 |       5.03 |             |
|    new    | compress |            1e+06 |       1.04 |      20.73% |
|    old    | expand   |               10 |   7.72e-05 |             |
|    new    | expand   |               10 |   1.21e-05 |      15.66% |
|    old    | expand   |              100 |    0.00068 |             |
|    new    | expand   |              100 |    5.6e-05 |       8.24% |
|    old    | expand   |            1e+03 |    0.00686 |             |
|    new    | expand   |            1e+03 |   0.000534 |       7.79% |
|    old    | expand   |            1e+04 |     0.0695 |             |
|    new    | expand   |            1e+04 |    0.00562 |       8.08% |
|    old    | expand   |            1e+05 |      0.689 |             |
|    new    | expand   |            1e+05 |     0.0558 |       8.09% |
|    old    | expand   |            1e+06 |       6.94 |             |
|    new    | expand   |            1e+06 |      0.574 |       8.28% |

Summary:

- the new `compress` takes roughly 20–22% of the time of the old `compress`;
- the new `expansion` takes roughly 8–8.5% of the time of the old `expand`.

(I would encourage you to test this for yourselves.)

(Pseudo)code used for comparison:
```py
from uuid import uuid4
from codetiming import Timer
# import original compress as "old_compress"
# import original expand as "old_expand"
# import new compress as "new_compress"
# import new expand as "new_expand"

@Timer()
def test_old_compress(uuids: list[str], /):
    for uuid in uuids:
        old_compress(uuid)

@Timer()
def test_old_expand(guids: list[str], /):
    for guid in guids:
        old_expand(guid)

@Timer()
def test_new_compress(uuids: list[str], /):
    for uuid in uuids:
        new_compress(uuid)

@Timer()
def test_new_expand(guids: list[str], /):
    for guid in guids:
        new_expand(guid)

# prepare inputs separately
n = int(1e6)
uuids = [str(uuid4().hex) for _ in range(n)]
guids = list(map(compress, uuids))

# run and compare methods
test_old_compress(uuids)
test_new_compress(uuids)
test_old_expand(guids)
test_new_expand(guids)
```

(EDIT: added more entries to table, corrected pseudocode, added summary.)

aothms:
Couple requests:

- `b64encode` / decode accept an `altchars` kwarg, no reason to use `maketrans`

https://docs.python.org/3/library/base64.html#base64.b64encode

- We use pytest generally instead of `from unittest import TestCase`. You can just remove all the TestCase boilerplate and create functions that start with `test_` for pytest to pick those up.

- Remove unrelated changes from the PR

- The tests are nice, but it would be good to have some formal assurance that the approach for padding is compliant with the spec.

raj-open:
Hi @aothms! Thank you for looking at the request!

> Couple requests:
> 
> * `b64encode` / decode accept an `altchars` kwarg, no reason to use `maketrans`

I am aware of this. However note that this "altchars" argument is _only_ for the final 2 characters in the base64-alphabet:

> Optional altchars must be a [bytes-like object](https://docs.python.org/3/glossary.html#term-bytes-like-object) **of length 2 which specifies an alternative alphabet for the + and / characters**. This allows an application to e.g. generate URL or filesystem safe Base64 strings. The default is None, for which the standard Base64 alphabet is used.

In the case of IFC's convention, it's not merely these two characters, but the entire alphabet (cf. #5991), which deviates from the global standard of Base 64 encoding. This is the reason why something like `maketrans` is necessary.

> * We use pytest generally instead of `from unittest import TestCase`. You can just remove all the TestCase boilerplate and create functions that start with `test_` for pytest to pick those up.

Oh, my apologies. I will update this. I take it then, that I can completely do away with classes and simple write functions?

> * Remove unrelated changes from the PR

Will do! (I ideally didn't want to add this commit at all : )

> * The tests are nice, but it would be good to have some formal assurance that the approach for padding is compliant with the spec.

Okay, I will think of something. If you have any particular cases/qualities you would like me to check, I can add these.

Briefly: What I originally attempted was to properly pad (on the right with `=` as per standard). However, this led to differing result. I then reflected on IFC's algorithm and realised that it chops off the first 2 characters (in base 64) and converts/pads these. The padding that occurs in `expand` corresponds to this peculiarity. The padding in `compress` simply undoes this effect, so is not special in itself.

> [!NOTE]
> I need some time and will get on to this later this evening (UTC+1).

aothms:
> In the case of IFC's convention, it's not merely these two characters, but the entire alphabet 

Thanks. My bad.

> Okay, I will think of something. If you have any particular cases/qualities you would like me to check, I can add these.

I think I'm ok now:

- 6 bytes at a time results in groups of 4 6-byte chars that align with 3 8-byte chars as 4 * 6 == 3 * 8
- encoded this means 2 | 4 | 4 | 4 | 4 | 4 == 22
- decoded this means 1 | 3 | 3 | 3 | 3 | 3 == 16 (or double in hex)
- so that's why you leftpad with 2 bytes '0000' in hex
- and leftpad with 'AA' in base64 to get to the right number of elements per group so the left to right the marks correspond
- after leftpadding the corresponding number of chars need to be sliced off the front

I like how you came up with a general solution to something specific 👍

raj-open:
> * 6 bytes at a time results in groups of 4 6-byte chars that align with 3 8-byte chars as 4 * 6 == 3 * 8
> * encoded this means 2 | 4 | 4 | 4 | 4 | 4 == 22
> * decoded this means 1 | 3 | 3 | 3 | 3 | 3 == 16 (or double in hex)
> * so that's why you leftpad with 2 bytes '0000' in hex
> * and leftpad with 'AA' in base64 to get to the right number of elements per group so the left to right the marks correspond
> * after leftpadding the corresponding number of chars need to be sliced off the front

Yep, that covers the numerical side of things. The fiddly bit was realising that IFC pads left compared to the usual right-padding under the Base64 convention.

Please let me know if the tests are okay (I ran them separately and they were green; but I could not run them within the IfcOpenShell repository as I cannot compile ifc_wrapper, so I was unable to check if everything is compatible import-wise, etc.).

> I like how you came up with a general solution to something specific 👍

Thanks! It's _almost_ general. The 2 `AA` values / 4 `0000` values thing is a bit of magic (for hex-codes of length `n ≠ 32` this relation can slightly differ), but nonetheless the structure 

```mermaid
flowchart LR
    A["UUID.hex"]
    -->|"left-pad _m_ 0's"| B["36 chars"]
    -->|"encode"| C["UUID.64 (24 chars)"]
    -->|"left-depad _n_ chars"| D["UUID.64 (22 chars)"]
    -->|"translate"| E["UUID.64 ifc"]
```

```mermaid
flowchart LR
    A["UUID.64 ifc"]
    -->|"translate"| B["UUID.64"]
    -->|"left-pad _n_ As"| C["UUID.64 (24 chars)"]
    -->|"decode"| D["UUID.hex (36 chars)"]
    -->|"left-depad _m_ chars"| E["UUID.hex (32 chars)"]
```

appears to capture the essence of IFC's algorithms.

raj-open:
@aothms it appears that there are issues with the underlying C++ code, causing the build step in the ci pipeline to fail. I do not think it has anything to do with the two python scripts in this PR. Is it possible that something has been merged to v0.8.0 in a broken state?

aothms:
Yes, you're right. Easy to track down with the runtime of our CI action, which suddenly dropped from 20m to 2m to fail.

afbeelding

    git log d1ad082~..f0e1be765d9b2c80e39020b21503b37f7ee2064e --name-only

Doesn't show anything related to swig though so the error is a bit unexplained

    swig4.0: ../../Source/CParse/templ.c:582: Swig_cparse_template_expand: Assertion `!SwigType_isvariadic(s)' failed.

But we can merge this in and resolve the build as part of the other issue you created.

Thanks again!

PR #6006, opened 2025-01-21
adds calculations for projected_length of gradient curve segments

civilx64:
@RickBrice everything appears to check out as expected on the python end.  Let me know if anything looks suspicious.  This does not yet address IfcClothoid parent curves but I expect that IfcLine, IfcPolynomialCurve, and IfcCircle will cover the vast majority of models "in the wild".


RickBrice (approved):
Nicely done Scott - I nothing stands out as suspicious to me. 

For IfcClothoid, something similar to IfcPolynomialCurve is needed. I haven't found vertical clothoids in the wild, but I understand they are used in Japanese high speed rail.

PR #6022, opened 2025-01-22
Minor-Fixes ---> v0.8.0

raj-open:
It appears that files have been pushed to the branch without being linted causing the workflow `ci-black-formatting / lint-formatting` to consistently fail.

This PR attempts to fix this.

> [!NOTE]
> What would be even better would be the introduction of githooks that forcibly run the linter before pushing to remote.


aothms:
Thanks

PR #6024, opened 2025-01-22
Fix issue with eliminate_touching_operands

Kipre:
Hello
Yesterday I started using clipping on some of my walls and the rendering of the boolean substraction would not work for some of them in Bonsai. Since it worked fine in Solibri and Open IFC Viewer I thought it is probably a bug here. 

After a bit of investigation I think a found the bug. Looks like the coplanarity test was not restrictive enough.

{80956786-4D7A-4FAE-B229-90421D6149DE}
{465C0008-5CD2-499C-9054-8448C243D991}
[clipping_issue.ifc.txt](https://github.com/user-attachments/files/18511513/clipping_issue.ifc.txt)

I didn't add a test case for this fix but I will be happy to add one if someone points me to where it should be placed.


aothms:
Wonderful, many thanks for this. I always find it admiring when people can dive into a huge code base with fresh eyes and solve an issue with chirurgical precision.

Kipre:
Thanks, very glad the fix was relevant

PR #6041, opened 2025-01-26
Linting-Workflow ---> v0.8.0 | Make wf only fail on syntax errors instead of unformatted code

raj-open:
Since linting is not enfroced, this PR is proposes a change to a workflow to ensure failure only upon syntax errors instead of unprettified code.

## Background ##

The linter [Black](https://pypi.org/project/black) was introduced to the IfcOpenShell repository recently. Unfortunately there is no reinforcement, i.e. there are

- no **CONTRIBUTING.md** containg guidelines requiring contributors to ensure that they run the linter before pushing (or latest: merging a PR).

- no [githooks](https://pre-commit.com/#new-hooks) which force linting to be run upon pushing.

Furthermore, subsequent PRs have been merged despite containing unlinted code. This results in a systematic problem of _future_ PRs containing checks that fail not due to their own deficits, but rather because unlinted changes had been previously merged. So unless remedied this will lead to cascading errors and false negatives.

In light of this, it is suggested here that the maintainers of IfcOpenShell rethink the setup of the workflow in **ci-black-formatting.yaml**. Before stating what these changes are, we note some facts.

## Notes on outcomes of Linting ##

There are 3 outcomes of (black's) linting:

1. The code has _already been prettified_ -> nothing changes and the linter succeeds.

2. The developer did not prettify their code -> two subcases:

    1. **unprettified code with no syntax errors** -> linting will bring about changes and throws a non-fatal error, stating that some files changed.

    2. **code with syntax error/s** -> linting throws fatal error, stating that one or more files could not be prettified due to syntax errors.

## PR changes ##

This PR changes the the workflow in **ci-black-formatting.yaml** as follows. Previously we had (corresponding to the above cases)

1. already linted code -> workflow succeeds

2. unlinted code:

    1. **unprettified code with no syntax errors -> workflow fails**

    2. code with syntax error/s -> workflow fails

Since there is no linting policy enforced, our change turns case **2i)** into a mere warning instead of a failure.

## How to test ##

1. Create a new sub-branch from the present one (`linting-workflow`)

2. Make changes and push to verify the desired the outcome of the workflow:

    - **Case 1:** run the black linter locally, i.e. run
        ```bash
        python3 -m black .
        ```
        (or run `python ...` if using windows)

        Push the changes and wait for triggered workflow to finish.

        **Expected outcome:** Workflow succeeds with no warnings or errors.

    - **Case 2i:** Now add some bad formatting to some random **.py** file in src
        (but no syntax errors), e.g. add some extra unnecessary spaces around an `=`
        or a really long line.

        Push the changes and wait for triggered workflow to finish.

        **Expected outcome:** Workflow succeeds with warnings about unprettified code.

    - **Case 2ii:** Now introduce some syntax errors to a random **.py** file in src,
        e.g. add the line
        ```py
        if x = 0
            pass
        ```
        Push the changes and wait for triggered workflow to finish.

        **Expected outcome:** Workflow fails with notes about syntax errors.

3. Delete the temporary branch.


Moult:
Thank you! What do you think about enforcing / automating Black formatting?

raj-open:
Hi, @Moult,

> What do you think about enforcing / automating Black formatting?

I guess there are several ways you may consider doing this:

1. Automation via GitHooks

    - use e.g. the [pre-commit](https://pre-commit.com/#new-hooks) python package to install githooks.
    - this involves telling all your contributers to run a setup script to install this
    - => unless you're a small team, this could be tricky, but it's not impossible.

2. Automation via Workflows

    - set up github workflows that run the linter, then create and push a commit to the current branch
    - this can get messy, esp. for users who don't run git pull or perform reverse merges and it may end up irritating people : )
    - => I personally don't recommend this!

3. Manually: Write a CONTRIBUTING.md file and present there some guidelines about responsibilities of a contributor before creating a PR.
   In the PR check list you can add a checkpoint "was linting performed?"

4. Semi-Manually: A mix of 2+3, but instead of a workflow that automatically runs,
    you make it manually triggerable.
    If during a PR-review somebody has (despite the instructions in CONTRIBUTING.md)
    failed to lint,
    then instead of bugging them, you can just trigger the workflow manually
    (or they can be automatically alerted to do this).

4 might be nice, but I guess it's complicated. I would personally suggest 3 for now.

## Some other things to consider ##

### Choice of linter ###

Black is "easy to configure". But that's about it for the advantages.
I used to use it for years, but have now switched to [ruff](https://docs.astral.sh/ruff),
which is incredibly fast and robust.
Once you set it up, you can do things a lot more cleanly, e.g. dry-runs, auto-fixing to PEP-standards,
auto-grouping of imports (std libs, 3rd party, 1st party).

I also recommend using [uv](https://docs.astral.sh/uv) from the same developers
as a package manager (uses pyproject.toml still, but builds the dependencies a lot faster and cleaner).

### A tasks script ###

I notice you guys have a Makefile in some submodules, which is great.
Why not have a taskscripts file in the root of the repository,
which includes e.g. the following tasks:
```bash
setup # general setup after cloning <- e.g. you can add GitHook installation here : )
build # creates a venv, builds the python dependencies, builds the cpp code
clean # wipes all locally created artefacts

tests # a short for the following subtasks
tests-unit # run python (and cpp?) unit tests
tests-integration # run integration tests or whatever other tests you have
```
If you did this, you could enforce the use of GitHooks,
by informing contributors to run the `setup` (and `build`) tasks
whenever the set about to work on a new branch or something.

For this tasks script in the root repository,
I would also recommend switching from GNU Make
(which has a bunch of weird idiosyncracies like needing to use `.PHONY`, handling of sessions, etc.)
to [justfile](https://github.com/casey/just?tab=readme-ov-file#installation).
This is well establish, similar tool to Make with cleaner implementation.

## The GitHooks themselves ##

If you go for this variant, I would suggest very simple GitHooks at the start, e.g.

- commit message formatting (if you plan to enforce a particular way to write commit messages)
- linting maybe only upon push
- run some but not all tests (e.g. just unit) and just upon push

PR #6056, opened 2025-01-28
Fix transform matrix computation in Axis2Placement3D

Kipre:
Hello,
I believe there was a mistake in the computation of the transform matrix to resolve the placements of the entities. I used [this](https://github.com/g-truc/glm/blob/69b130c162e6266e07392741bd04feacc55dcda2/glm/ext/matrix_transform.inl#L178) implementation as a reference.

I stumbled upon the issue when trying to place the roof in my model (blender + bonsai on windows):
{1A4F45D9-E8D6-4BCF-8C7C-B906289921B0}
{14B7942D-954A-43A1-A9BD-35DACDF39CCE}

When transforming the ifc to obj in linux through python I had this output:
{89962E85-0733-40C0-8DEF-7752D80F9B5C}
I have no explanation as to why it does not orient at all.

After the fix, the ifc to obj starts working fine on linux:
{727344B4-A4F8-4BCA-B417-ACC5244175B0}

I have no way of testing this in blender + bonsai on windows because I don't want to install MSVC but I assume it also fixes my original issue.
I also didn't run any tests so I don't know if the change has side effects that result in the need for some adaptations.

### DIFF
--- a/src/ifcgeom/taxonomy.h
+++ b/src/ifcgeom/taxonomy.h
@@ -267,9 +267,9 @@ typedef item const* ptr;
 			struct matrix4 : public item, public eigen_base<Eigen::Matrix4d> {
 			private:
 				void init(const Eigen::Vector3d& o, const Eigen::Vector3d& z, const Eigen::Vector3d& x) {
-					auto X = x.normalized();
-					auto Y = z.cross(x).normalized();
 					auto Z = z.normalized();
+					auto Y = Z.cross(x).normalized();
+					auto X = Y.cross(Z);
 					components_ = new Eigen::Matrix4d;
 					(*components_) <<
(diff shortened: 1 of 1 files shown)


aothms:
I think you're right, I also quickly checked https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcBuildAxes.htm and https://github.com/Open-Cascade-SAS/OCCT/blob/bef42e89ef9fac7920af6ce40b74a1ff39352a9b/src/gp/gp_Ax3.hxx#L88

The current implementation is definitely wrong, because it allows for X to be non perpendicular to Z, not sure how that happened.

Thanks!

PR #6076, opened 2025-01-31
Rename dm_IFC.png to dm_ifc.png

trhyder:
renaming to lowercase to match lm_ifc presidents.


trhyder:
related to work on #5372

PR #6078, opened 2025-01-31
Unify icon modes to work with default Blender Light theme

trhyder:
Force the color of the icons in the Properties-Scene tab to match Regular Text
Use the icon display mode function to detect where to use dm_ifc.png or lm_ifc.png


trhyder:
Screenshot 2025-01-31 at 11 08 40 AM

#5372

PR #6089, opened 2025-02-02
Tree icon for new Geographic Element Tool

trhyder:
@moul, it's here if you want it. The icon was copied from the landscape library that ships with Bonsai


trhyder:
image

Moult:
Yes, vegetation needs more prominence in our built environment. This is the least we can do.

src/bonsai/bonsai/bim/module/model/product.py
```diff
@@ -114,7 +114,7 @@ def _execute(self, context):
             props.ifc_predefined_type = "FOOTING_BEAM"
             props.representation_template = "PROFILESET"
         elif self.ifc_element_type == "IfcPileType":
-            props.ifc_predefined_type = "BORED"
```
Moult, line 117: Isn't BORED a good default?

trhyder: @Moult, BORED isn't in IFC2X3, so I picked COHESION because it's in IFC2X3, IFC4, and IFC4X3

PR #6154, opened 2025-02-13
parsed mvd description in ifcopenshell.file

Ghesselink:
Implementation of grammar in ifcopenshell.file as described in
 `https://standards.buildingsmart.org/documents/Implementation/ImplementationGuide_IFCHeaderData_Version_1.0.2.pdf`
 
 New properties; 
 - `f.mvd `- The View Definition value (e.g. '`CoordinationView`')
 - `f.mvd_options` - All the values after the keyword 'Option'
 - `f.mvd_comments` - Similar, for 'Comment'
 -  `f.mvd_exchange_requirements` `-> Similar, but for 'ExchangeRequirement'
 - `f.mvd_keywords` -> List of all the keywords available in the description (e.g. ViewDefinition, Option & Comment)
 -  `f.mvd_optional_keyword_fields` -> fields from keywords other than 'ViewDefinition', 'Option', 'Comment' and 'ExchangeRequiremens'
 

----

 - Some examples; 
 
 _(f = ifcopenshell.open(file), header=f.wrapped_data.header.file_description.description_)
 

 Only View Description

```
header
'ViewDefinition [Alignment-basedReferenceView]'
f.mvd
'Alignment-basedReferenceView'
```

```
header
'ViewDefinition [CoordinationView]'
f.mvd
'CoordinationView'
```

List of keywords contains only 'MVD', other options (mvd_keywords, mvd_comments) are still empty
```
header
'ViewDefinition [CoordinationView, QuantityTakeOffAddOnView]'
f.mvd
'CoordinationView. QuantityTakeOffAddOnView'
f.mvd_options
f.mvd_keywords
"{'mvd'}"
f.mvd_comments
""
```

View definitions spread over two keywords
```
header
'ViewDefinition [CoordinationView] ViewDefinition[QuantityTakeOffAddOnView]'
f.mvd
'CoordinationView.QuantityTakeOffAddOnView'
```
Including Options
```
header
'ViewDefinition [CoordinationView,SpaceBoundary1stLevelAddOnView, QuantityTakeOffAddOnView ] Option [ExcludedObjects: Stair, Ramp, Space; SplitLevel: On]'
f.mvd
'CoordinationView.SpaceBoundary1stLevelAddOnView. QuantityTakeOffAddOnView '
f.mvd_options
"{'ExcludedObjects': [' Stair', ' Ramp', ' Space'], 'SplitLevel': ' On'}"
```

Including a Comment
```
header
'ViewDefinition [CoordinationView, SpaceBoundary1stLevelAddOnView, QuantityTakeOffAddOnView ] Option [ExcludedObjects: Stair, Ramp, Space; SplitLevel: On] Comment [This export contains Boolean Operation geometry that may not be fully supported by other application]'
f.mvd
'CoordinationView. SpaceBoundary1stLevelAddOnView. QuantityTakeOffAddOnView '
f.mvd_options
"{'ExcludedObjects': [' Stair', ' Ramp', ' Space'], 'SplitLevel': ' On'}"
f.mvd_comments
'This export contains Boolean Operation geometry that may not be fully supported by other application'
f.mvd_keywords
"{'comment', 'option', 'mvd'}"
```

Including an optional keyword field (remark)
```
header
'ViewDefinition [CoordinationView] ViewDefinition [QuantityTakeOffAddOnView] Option [ExcludedObjects: Stair, Ramp, Space; SplitLevel: On] ExchangeRequirement [CustomRequirement: Value1, Value2] REMARK [SomeKey: SomeValue; AnotherKey: AnotherValue] Comment [This is a free text comment, or a comma-separated list of items]'
f.mvd
'CoordinationView.QuantityTakeOffAddOnView'
f.mvd_options
"{'ExcludedObjects': [' Stair', ' Ramp', ' Space'], 'SplitLevel': ' On'}"
f.mvd_comments
```


src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -304,6 +407,7 @@ def __init__(
         import weakref
 
         file_dict[self.file_pointer()] = weakref.ref(self)
+        self.parsed_description = parse_mvd(self.wrapped_data.header.file_description.description)
```
aothms: We should not cache/store this here because the user can overwrite the description field. Just do the parsing as part of the mvd property getter.

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -464,6 +568,68 @@ def schema_version(self) -> tuple[int, int, int, int]:
             number = re.search(prefix + r"(\d)", schema)
             version.append(int(number.group(1)) if number else 0)
         return tuple(version)
+    
+    @property 
+    def mvd(self) -> str:
+        """
+        View Definition supported by the exporting application that is reflected in the IFC file
+        For example: 
+        “CoordinationView”
```
aothms: You have some really fancy quotes, better remove them or repalce them with the plain old ascii ones

```
>>> hex(ord('”'))
'0x201d'
```

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -205,6 +208,106 @@ def commit(self) -> None:
...
+        """
+        key = str(args[0]).lower()
+        attr_name = f"{key}"
+        if attr_name not in self.keywords:
+            setattr(self, attr_name, {})
+            self.keywords.add(attr_name)
+        dynamic_dict = getattr(self, attr_name)
+        for value_set in args[1]:
+            set_name, *values = value_set
+            dynamic_dict[set_name] = values if len(values) > 1 else values[0]
```
aothms: I guess I would just always leave it as a list

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -31,10 +31,13 @@
 from typing import Union
 from typing import overload
 from typing import Literal
+from typing import Any
```
aothms, line 35: Could you move this all to a new module. Define this as a class called `mvd_info` or sth. And make that accessible from `ifcopenshell.file.mvd. ...`. That way it's a bit grouped. Bonus points for making this degrade nicely when lark is not installed.

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -464,6 +568,68 @@ def schema_version(self) -> tuple[int, int, int, int]:
...
+    def mvd_options(self) -> str:
+        """
+        Option being used by the exporting application for e.g. quality control and debugging
+        a. this is an optional field providing informal information, the values for the Option keyword
+        reflect the export settings of the exporting software application, those settings may be
+        specific to the exporting software
+        For example: 
+        > "{'ExcludedObjects': [' Stair', ' Ramp', ' Space'], 'SplitLevel': ' On'}"
+        """
+        return str(self.parsed_description.option)
```
aothms: Please don't return these (and the ones below) as strings.

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -464,6 +568,68 @@ def schema_version(self) -> tuple[int, int, int, int]:
...
+    
+    @property
+    def mvd_optional_keyword_fields(self) -> str:
+        """"
+        Returns fields from keywords other than 'ViewDefinition', 'Option', 'Comment' and 'ExchangeRequirement'.
+        For example, in case of 
+        'REMARK [SomeKey: SomeValue; AnotherKey: AnotherValue]'
+        file.mvd_optional_keywords_fields == "{'remark': {'SomeKey': ' SomeValue', 'AnotherKey': ' AnotherValue'}}"
+        """
+        return str({kw: getattr(self.parsed_description, kw) for kw in self.parsed_description.other_keywords})
```
aothms: I think I wouldn't make the distinction between `other_keywords` and the 4 ones you mentioned. Just list them all under the dictionary. Saves some logic to remove like keeping track of the set of keywords. Maybe build the dictionary at parse time and not here in the property getter

aothms (commented):
Thanks! Overall I think this is really quite good. Most of my doubts are more from unfamiliarity with this or unclarity in the spec. @Moult or @Andrej730 anything from you guys?

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -464,6 +466,16 @@ def schema_version(self) -> tuple[int, int, int, int]:
             number = re.search(prefix + r"(\d)", schema)
             version.append(int(number.group(1)) if number else 0)
         return tuple(version)
+        
+    @property 
+    def mvd(self):
+        if not LARK_AVAILABLE:
+            return None
+        file_description = self.wrapped_data.header.file_description
```
aothms:
```suggestion
        file_description = self.header.file_description
```

src/ifcopenshell-python/ifcopenshell/file.py
```diff
@@ -464,6 +466,16 @@ def schema_version(self) -> tuple[int, int, int, int]:
...
+        return MvdInfo(
+            get_description=lambda: file_description.description, 
+            set_description=lambda d: setattr(file_description, "description", tuple(d))
```
aothms: I guess the more conventional way would be to simply pass

```suggestion
            self.header
```

And in your tests (if needed) you create a mock datastructure that mimics the layout of the file.header object

Because you don't really need a more general interface like this, get_description/set_description are encapsulated in self.header

src/ifcopenshell-python/ifcopenshell/util/mvd_info.py
```diff
@@ -0,0 +1,310 @@
...
+            value = AutoCommitList(value, self._commit)
+        super().__setitem__(key, value)
+        self._commit()
+
+    def __delitem__(self, key):
+        super().__delitem__(key)
+        self._commit()
+
+
+class AutoCommitList(list):
```
aothms, line 278: ok, this is maybe a bit more involved than I had in mind. 

A simple `to_string()` for example on MvdInfo would have been ok, so that you would have to manually call that and assign back to the header if you're mutating your MvdInfo. But I don't see downsides to your approach, so let's roll with this and see how it goes :)

aothms (commented):
Looking good, let me know what you think about the suggestion re set/get_description

src/ifcopenshell-python/ifcopenshell/util/mvd_info.py
Ghesselink: I thought about something like that, but also considered that it would be more intuitive if we could directly modify it through this mvd parser. While wondering whether it would be possible I started implementing it ..  
Maybe it is more error-prone and/or not as intuitive as expected. We can then always reverse it. Another option would be to have both methods available. 

I've also made sure we can add view_definitions this way, e.g.
`f.mvd.view_definitions.append('AlignmentBased-View')`

src/ifcopenshell-python/ifcopenshell/file.py
Ghesselink: Added 
https://github.com/IfcOpenShell/IfcOpenShell/pull/6154/commits/8983b0571b0b189d0ff1122b291f81bd25d4ebc6

src/ifcopenshell-python/ifcopenshell/file.py
Ghesselink: That makes sense, I didn't think about offloading this to the MVDInfo class. Just needed to do a couple of tweaks in the mvd parser and modifying some tests.
e.g. 
```
    @property
    def description(self) -> list[str]:
        return self._get_description()
```
must be changed to 
```
    @property
    def description(self) -> list[str]:
        return self._header.file_description.description
```
https://github.com/IfcOpenShell/IfcOpenShell/pull/6154/commits/8983b0571b0b189d0ff1122b291f81bd25d4ebc6

PR #6198, opened 2025-02-20
fix: bug in initialization of gltf serialiser fixed (documentation)

HelloJowet:
The `GltfSerializer` requires 3 parameters, but the example script in the documentation just uses two. This results into an error.

```py
class GltfSerializer(WriteOnlyGeometrySerializer):
    r"""Proxy of C++ GltfSerializer class."""

    thisown = property(lambda x: x.this.own(), lambda x, v: x.this.own(v), doc="The membership flag")
    __repr__ = _swig_repr

    def __init__(self, filename, geometry_settings, settings):
        r"""__init__(GltfSerializer self, std::string const & filename, Settings geometry_settings, SerializerSettings settings) -> GltfSerializer"""
        _ifcopenshell_wrapper.GltfSerializer_swiginit(self, _ifcopenshell_wrapper.new_GltfSerializer(filename, geometry_settings, settings))
    __swig_destroy__ = _ifcopenshell_wrapper.delete_GltfSerializer
```

I've fixed it.

### DIFF
--- a/src/ifcopenshell-python/docs/ifcopenshell-python/geometry_processing.rst
+++ b/src/ifcopenshell-python/docs/ifcopenshell-python/geometry_processing.rst
@@ -306,39 +306,43 @@ In addition to geometry settings, serialisation has its own set of
 
 .. code-block:: python
 
-    import ifcopenshell
-    import ifcopenshell.geom
-    import multiprocessing
-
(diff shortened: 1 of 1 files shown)


aothms:
Many thanks :)

PR #6234, opened 2025-02-25
Import alignment from csv

### LINKED ISSUE #5638: Alignment authoring API
Here is a very rough first cut at an alignment authoring api. I obviously need to learn a lot about python so forgive the non-standard function signatures.

I think the goal is: Keep it simple - develop an api for horizontal and vertical roadway/highway alignments. Add rail (i.e. cant) later knowing that doing so may require breaking changes to the api.

Source folder (namespace):
IfcOpenShell\src\ifcopenshell-python\ifcopenshell\api\alignment

Creation Functions
==================
`def add_horizontal_alignment(file,optional:name,optional:description,array[x,y] points, array[double] radii, optional: bool include_geometry=true)`
Adds a horizontal alignment to the file
points = array of PB, PIs, EP (begin point, intersection points, end point)
radii = array of curve radii (must be >=0, 0 means angle point), radii.size() = points.size() - 2)
include_geometry = if true, creates geometric representation as IfcComposuteCurve, otherwise only business logic is created

`def add_alignment(file,optional:name,optional:description,array[x,y] points, array[double] radii, array[d,z] vpoints, array[double] vclength, optional bool include_geometry=true)`
Adds a horizontal and vertical aligment ot the file
points = array of PB, PIs, EP
radii = array of curve radii (must be >=0, 0 means angle point), radii.size() = points.size()-2)
vpoints = array of VPB, VPIs, and VEP (vertical begin point, vertical PIs, vertical end point)
vclength = array of vertical curve lengths (must be >= 0, 0 means angle point, vclength.size() = vpoints.size()-2)
include_geometry = if true, creates geometric representation as IfcComposuteCurve + IfcGradientCurve, otherwise only business logic is created

`def add_profile(file,optional:name,optional:description,alignment,array[d,z] vpoints, array[double] vclength)`
Adds a profile to a previously defined alignment. Automatically creats IfcGradientCurve if alignment as an associated IfcCompositeCurve
vpoints = array of VPB, VPIs, and VEP
vclength = array of vertical curve lengths (must be >= 0, 0 means angle point, vclength.size() = vpoints.size()-2)

`def create_alignment_representation(file,alignment)`
Creates the geometric representation for horizontal and, if present, vertical alignment. Creates IfcCompositeCurve and 0 or more IfcGradientCurve. Does not create if they already exist.

Removal Functions
===================
`def remove_alignment(file,alignment)`
Removes alignment from the file. Removes all nested alignment segments unless the segment is nested into other alignments. Removes IfcCompositeCurve and IfcCurveSegment if associated with the alignment (only removes IfcCurveSegment if not used by other IfcCompositeCurve). Removes all nested IfcGradientCurve, if not nested into other IfcGradientCurve.

`def remove_profile(file,alignment,profile)`
Removes a profile from an alignment. Removes geometric representation if applicable. Removes nested segments and IfcCurveSegment if not used by others.

Edit Functions
===================
We should probably have some edit functions, but I think this will be a little tricky with the geometry caching implemented in IFCOS. For now, I think editing should be done by remove and create. Maybe we need some `get` functions that return the point and radii arrays for a previously created alignment.

Other Functions
==================
`def add_stationing(file,alignment,start_station)`
Creates an IfcReferent .STATION. with Pset_Stationing and positions it at the start of the alignment

`def add_referents(file,alignment)`
Defines IfcReferent .REFERENCEMARKER. for key points along the horizontal alignment such as TS, SC, PC, CS, PT, and ST
This isn't well thought out yet. Maybe do the same for vertical profile in the same or in a different function
The remove functions defined above would remove IfcReferent entities if the parent alignment/profile are removed.

RickBrice:
This PR is an attempt to provide a quick and dirty mechanism to define alignment geometry in Bonsai. The idea is to define an alignment in a CSV file and import it. This is a temporary feature and will probably go away someday once I get a better handle on how to develop for Bonsai. 

There is a new Alignment (.csv) command on the Import menu.

The CSV file format is:
X1,Y1,R1,X2,Y2,R2 .... Xn-1,Yn-1,Rn-1,Xn,Yn
D1,Z1,L1,D2,Z2,L2 .... Dn-1,Zn-1,Ln-1,Dn,Zn
D1,Z1,L1,D2,Z2,L2 .... Dn-1,Zn-1,Ln-1,Dn,Zn

The first row defines the horizontal alignment by the PI method. The circular curve radii R1 and Rn-1 are "placeholders" and are not used. A value of 0.0 is recommended.

The second and subsequent rows define vertical alignments by the PI method. D is distance along, Z is elevation, and L is the horizontal length of a parabolic vertical curve. Similar to the horizontal circular curves, L1 and Ln-1 are "placeholders" and are not used. A value of 0.0 is recommended.

Example:
Create a new project using the New Project Wizard. Use the IFC4X3 Schema
Select File > Import > Alignment (.csv)

image

Imported an alignment with two vertical profiles

image


src/ifcopenshell-python/ifcopenshell/api/alignment/segments.py
```diff
@@ -0,0 +1,268 @@
...
+# along with IfcOpenShell.  If not, see <http://www.gnu.org/licenses/>.
+
+import ifcopenshell
+from ifcopenshell import entity_instance
+from ifcopenshell import ifcopenshell_wrapper
+import math
+from typing import Sequence
+
+
+def get_axis_subcontext(file:ifcopenshell.file):
```
civilx64:
```suggestion
def get_axis_subcontext(file:ifcopenshell.file) -> entity_instance:
```

src/ifcopenshell-python/ifcopenshell/api/alignment/segments.py
```diff
@@ -0,0 +1,268 @@
...
+    axis_geom_subcontext = ifcopenshell.util.representation.get_context(file,"Model","Axis","MODEL_VIEW")
+    if(axis_geom_subcontext == None):
+        geometric_representation_context = ifcopenshell.api.context.add_context(file,context_type="Model")
+        axis_geom_subcontext = ifcopenshell.api.context.add_context(file,context_type="Model",context_identifier="Axis",target_view="MODEL_VIEW",parent=geometric_representation_context)
+   
+    return axis_geom_subcontext
+
+
+def map_alignment_vertical_segment(file: ifcopenshell.file, segment: entity_instance) -> Sequence[entity_instance]:
+    segment_type = segment.is_a().upper()
```
civilx64:
```suggestion
    """
    Creates entities for the representation of the supplied business logic entity instance.
    
    A pair of entities is returned because a single business logic segment of type HELMERTCURVE maps to two representation entities.
    """

    segment_type = segment.is_a().upper()
```

src/ifcopenshell-python/ifcopenshell/api/alignment/segments.py
```diff
@@ -0,0 +1,268 @@
...
+    start_height = segment.StartHeight
+    start_gradient = segment.StartGradient
+    end_gradient = segment.EndGradient
+    radius_of_curvature = segment.RadiusOfCurvature
+    
+    if math.isclose(horizontal_length, 0):
+        # set transition value based on whether this is the final zero-length segment
+        transition = "DISCONTINUOUS"
+    else:
+        transition = "CONTSAMEGRADIENTSAMECURVATURE"
```
civilx64: Ok as-is for now, but if modeling something like a ditch profile that has PVI stations and elevations with no curve run, we would need to set transition = 'CONTINUOUS'.

src/ifcopenshell-python/ifcopenshell/api/alignment/__init__.py
```diff
@@ -0,0 +1,24 @@
+# IfcOpenShell - IFC toolkit and geometry engine
+# Copyright (C) 2022 Dion Moult <dion@thinkmoult.com>
```
civilx64, line 2:
```suggestion
# Copyright (C) 2025 Dion Moult <dion@thinkmoult.com>
```

src/ifcopenshell-python/ifcopenshell/api/alignment/segments.py
```diff
@@ -0,0 +1,268 @@
+# IfcOpenShell - IFC toolkit and geometry engine
+# Copyright (C) 2021 Thomas Krijnen <thomas@aecgeeks.com>
```
civilx64:
```suggestion
# Copyright (C) 2025 Thomas Krijnen <thomas@aecgeeks.com>
```

src/ifcopenshell-python/ifcopenshell/api/alignment/alignment.py
```diff
@@ -0,0 +1,816 @@
+# IfcOpenShell - IFC toolkit and geometry engine
+# Copyright (C) 2021 Thomas Krijnen <thomas@aecgeeks.com>
```
civilx64:
```suggestion
# Copyright (C) 2025 Thomas Krijnen <thomas@aecgeeks.com>
```

civilx64 (commented):
Nice - thanks!!! The humble alignment.py is ready to spread its wings so to speak and become part of the api.

I made suggestions on current year in the copyright headers but not certain if those actually should be updated.

General suggestions for follow-up:
- lint with [black](https://black.readthedocs.io/en/stable/index.html)
- docstrings for everything in the api (including examples)
- unit tests for the alignment api using the FHWA data as example

Some of these may be good candidates for tasks to put on the infra authoring project backlog.

Moult:
Regarding copyright changes I think sooner or later a script that parses git logs per file and updates the copyright dates is probably something we should do project-wide.

src/bonsai/bonsai/bim/module/sequence/__init__.py
```diff
@@ -81,6 +81,7 @@
     operator.GenerateGanttChart,
     operator.GuessDateRange,
     operator.GoToTask,
+    operator.ImportAlignmentCSV,
```
Moult: This should be in module/geometry perhaps, or create your own new module like module/alignment (probably better long term as it grows in functionality).

src/bonsai/bonsai/bim/module/sequence/operator.py
```diff
@@ -654,6 +654,75 @@ def execute(self, context):
...
+
+        # IFC 4.1.5.1 alignments cannot be contained in spatial structures, but can be referenced into them
+        sites = self.file.by_type("IfcSite")
+        for site in sites:
+            ifcopenshell.api.spatial.reference_structure(self.file,products=[alignment],relating_structure=site)
+
+        # process the generated IfcReferent for the alignment
+        for rel in alignment.IsNestedBy:
+            for referent in rel.RelatedObjects:
+                if referent.is_a().upper() == "IFCREFERENT":
```
Moult: is_a("IfcReference")

Also recommend using util.element instead of manually going through rels.

src/bonsai/bonsai/bim/module/sequence/operator.py
```diff
@@ -654,6 +654,75 @@ def execute(self, context):
...
+                    referent_obj = bpy.data.objects.new(tool.Loader.get_name(referent),None)
+                    tool.Geometry.link(referent,referent_obj)
+                    tool.Collector.assign(referent_obj,should_clean_users_collection=False)
+
+        # an alignment can be an aggregation of multiple child alignments (ie. multiple verticals for a single horizontal)
+        curves = []
+        for rel in alignment.IsDecomposedBy:
+            for agg in rel.RelatedObjects:
+                if agg.is_a().upper() == "IFCALIGNMENT":
```
Moult: Same comment as above

src/bonsai/bonsai/bim/module/sequence/operator.py
```diff
@@ -654,6 +654,75 @@ def execute(self, context):
...
+                    curves.append(agg.Representation.Representations[0].Items[0]) # 3D curve
```
Moult: I find this risky, it assumes 1) there is a rep 2) the 0th rep is the appropriate context and 3) the 0th item is a 3d curve. Better to use `util.representation.get_representation`. Don't know enough whether or not the curve assumption is safe.

src/bonsai/bonsai/bim/module/sequence/operator.py
```diff
@@ -654,6 +654,75 @@ def execute(self, context):
...
+
+        # if there aren't any curves from aggregation, then there is only a single vertical or no vertical
+        if len(curves) == 0:
+            if len(alignment.Representation.Representations) == 1:
+                curves.append(alignment.Representation.Representations[0].Items[0]) # footprint
+            else:
+                curves.append(alignment.Representation.Representations[1].Items[0]) # 3D curve
```
Moult: Is footprint and 3D curve distinguished using geometric subcontext? If so get_representation utility can be used.

src/ifcopenshell-python/ifcopenshell/api/alignment/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with IfcOpenShell.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Manages alignment and alignment geometry
+"""
+
+from .alignment import *
+from .segments import *
```
Moult: The rest of the API typically has functions within each module. It may be worthwhile breaking up alignment.py and segments.py into (verb) functions, one per file like everything else. Each file would then have its corresponding test cases mirrored in the test folder.

RickBrice:
Thanks for the feedback. Looks like I’ve got some work to do before this is ready for prime time.

src/bonsai/bonsai/bim/module/sequence/__init__.py
RickBrice: @Moult I removed all the Bonsai changes I made under module/sequence and created module/alignment.
In module/alignment I have __init__.py and operator.py because these were the only files I changed in module/sequence.

The menu command for Alignment no longer shows up in the Import menu.

What am I missing?

The files are here:
https://github.com/IfcOpenShell/IfcOpenShell/tree/RAB_alignment_api/src/bonsai/bonsai/bim/module/alignment

RickBrice:
@civilx64 I've been doing a lot of refactoring of the API - it is way more granular now.
How do I test the documentation? I've been adding docstrings for all the functions.
You can see what I've been up to in https://github.com/IfcOpenShell/IfcOpenShell/tree/RAB_alignment_api/

src/ifcopenshell-python/ifcopenshell/api/alignment/__init__.py
RickBrice: Good advice. I've been doing just that. Here is the WIP https://github.com/IfcOpenShell/IfcOpenShell/tree/RAB_alignment_api/

src/bonsai/bonsai/bim/module/sequence/operator.py
RickBrice: This has been re-worked so the details are hidden behind a function and it makes use of the api.

src/bonsai/bonsai/bim/module/sequence/operator.py
RickBrice: Refactored so its not so risky.

Moult:
You may need to add your module to the list of modules in
`bim/__init__.py`

Moult:
See https://docs.ifcopenshell.org/introduction/how_to_contribute.html#writing-documentation

civilx64:
@RickBrice apologies; it looks like my docstring exmples are using a different (possibly obsolete) syntax.  The previous `ifcopenshell.alignment` docstrings get rendered rather messily:

image

when compared to other parts of the docs

image

See this example for [correct syntax in the docstring](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/ifcopenshell-python/ifcopenshell/api/root/copy_class.py#L56) and the corresponding [reference documentation](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html).

If the parameters and return types have python type hints - like this `copy_class` example - then you _should_ be able to omit the `:type:` and `:rtype:` info in the docstring.

Lastly you should be able to use the [doctest](https://realpython.com/python-doctest/) module to test any example code in docstrings.  Full disclosure - I have never utilized this myself.

LMK if you need some help with the busy work I created of reformatting `@param:` to `:param:`.  Maybe it will be a quick find/replace?

src/bonsai/bonsai/bim/module/sequence/__init__.py
RickBrice: This is now working from its own module

RickBrice:
I'm done a huge amount of refactoring and response to your comments.
I still need to clean up the documentation and implement additional unit tests.
Your review is requested.

RickBrice:
@Moult I have updated the basic documentation for the Import > Alignment comment. I have two questions for you regarding documentation

1. Where would be a good place to explain the expected values in the alignment CSV file? I couldn't find anything similar for the Cost or Work Schedule imports to use as a guide.
2. I want to update the [image that shows the import functions](https://docs.bonsaibim.org/reference/topbar.html#import). I found the style information document (line color, point size, etc), but I am uncertain as to how to create the circle, line, and rectangle with the color gradient. Is there some step-by-step instructions on how to do this with something like Gimp? I'm not at all versed with Gimp so after an hour I had to give up.

aothms:
Looks good to me. Really amazing to see the first contours of infra authoring.

Did you have a look at the failing tests:

afbeelding

I'm a fan of pytest.approx:

    assert mapped_segments[0].SegmentStart.wrappedValue == pytest.approx(1991.601501867533)

afbeelding

The `[0]` index is maybe a bit unclear that this is the 0th attribute of the direction instance. Maybe just use `DirectionRatios` but I'm guily of this myself frequently as well when I'm not patient enough to look up an attribute name. Anyway, just compare to a 2d vector instead.

aothms (approved):

RickBrice:
@aothms I updated the unit tests to use pytest.approx() as suggested. I am having difficulty reconciling the test errors. I don’t think I caused any of the mathutil errors. For geometry settings, I added a new setting, compute-curvature, but don’t know how to update the test for the new setting. There is another setting error related to cgal, and don’t think it is related to my changes. Please advise

Moult:
> Where would be a good place to explain the expected values in the alignment CSV file?

Anywhere in here is fine: https://github.com/IfcOpenShell/IfcOpenShell/tree/v0.8.0/src/bonsai/docs/guides - we can always reorganise it later but content is king :)

> I want to update the [image that shows the import functions](https://docs.bonsaibim.org/reference/topbar.html#import).

I don't have a good answer for this, because image editing takes time and a pixel-perfect attitude to give off a good impression for official docs, but it's also something that we don't have much time for. Perhaps gradiented boxes is too hard. Maybe just file a bug and let someone else help?

RickBrice:
@aothms i fixed all the unit tests related to my changes. The remaining failures seem to be related to mathutils. Is there something more I need to to do?

@civilx64 any additional comments? If not, I think this PR is ready to merge. 

We will have a simplistic method of defining an alignment and a starter API.

src/ifcgeom/function_item_evaluator.cpp
```diff
@@ -129,6 +153,15 @@ struct cant_fn_evaluator : public fn_evaluator {
         auto g = gradient_evaluator_.evaluate(u + start_);
         auto c = cant_evaluator_.evaluate(u);
 
+        
+        // curvature is stored in row 3 - capture it and remove it from the xy and uz matrices
+        // so the matrix operations (ie multiplication) works correct.y
```
civilx64:
```suggestion
        // so the matrix operations (ie multiplication) works correctly
```

src/ifcopenshell-python/ifcopenshell/api/alignment/add_stationing_to_alignment.py
```diff
@@ -0,0 +1,79 @@
...
+    .. code:: python
+
+        alignment = model.by_type("IfcAlignment")[0]
+        ifcopenshell.api.alignment.add_stationing_to_alignment(model,alignment=alignment,start_station=100.0)
+    """
+    # this commented out code is what you would do to add a geometric representation of the referent
+    # the example is a circle. a better way would be to pass a representation into the function
+    object_placement = None
+    representation = None
+    # basis_curve = ifcopenshell.api.alignment.get_basis_curve(alignment)
```
civilx64, line 46: Just fyi, for a block comment like this you can use triple quotes above and below similar to a docstring. 

Example:

```python
"""
Large
block
of
commented - out
code
goes
in
here
"""
```

civilx64 (commented):
I took a final look over everything - very impressive! Just one typo and an FYI.

src/ifcopenshell-python/ifcopenshell/api/alignment/add_stationing_to_alignment.py
RickBrice: > very impressive!

Thanks. The API is still very lacking, but it gives people the ability to define an alignment and get it into Bonsai.

RickBrice:
@Moult I think I messed up with this PR. I merged this branch into the v0.8.0 branch because I thought everything was good to go.

I have my Bonsai setup so it uses the latest build from the raw.githubusercontent.com repository. When I updated and tried to use the new alignment from CSV feature it crashes (traceback is below). The alignment feature depends on changes made in IfcOpenShell. Specifically, there is a new geometry setting to compute curvature on an alignment curve.

Sorry I messed this up. How can the problem be resolved?

~~~
  File "C:\Users\rickb\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\bonsai\tool\ifc.py", line 316, in execute
  File "C:\Users\rickb\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\bonsai\bim\ifc.py", line 458, in execute_ifc_operator
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\rickb\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\alignment\operator.py", line 67, in _execute
    ifcopenshell.api.alignment.create_geometric_representation(self.file, alignment)
  File "C:\Users\rickb\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\ifcopenshell\api\alignment\create_geometric_representation.py", line 129, in create_geometric_representation
    ifcopenshell.api.alignment.map_alignment_segments(file, layouts[0], composite_curve)
  File "C:\Users\rickb\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\ifcopenshell\api\alignment\map_alignment_segments.py", line 62, in map_alignment_segments
    ifcopenshell.api.alignment.add_segment_to_curve(file, mapped_segment, composite_curve)
  File "C:\Users\rickb\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\ifcopenshell\api\alignment\add_segment_to_curve.py", line 73, in add_segment_to_curve
    ifcopenshell.api.alignment.update_curve_segment_transition_code(prev_segment, segment)
  File "C:\Users\rickb\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\ifcopenshell\api\alignment\update_curve_segment_transition_code.py", line 51, in update_curve_segment_transition_code
    settings.set("COMPUTE_CURVATURE", True)
  File "C:\Users\rickb\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\ifcopenshell\geom\main.py", line 185, in set
    self.set_(self.name(k), v)
  File "C:\Users\rickb\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\ifcopenshell\ifcopenshell_wrapper.py", line 486, in set_
    return _ifcopenshell_wrapper.Settings_set_(self, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Setting not available
~~~

civilx64:
@RickBrice Did you recompile the C++ extension wrapper as well? If so is it linked to the blender python environment?

My hunch is that you didn't actually break anything. I don't think the daily releases include new compilations of the C++ wrapper.

RickBrice:
I can compile the C++ and copy the wrapper pyd and py files into the bonsai site-package\ifcopenshell folder and everything works.

My concern is that if someone updates bonsai with the latest build and tries to use the alignment import feature, it won't work.

Some how ifcos needs to be recompiled and packaged with bonsai.

civilx64:
> Some how ifcos needs to be recompiled and packaged with bonsai.

It will be, just takes a few days to a week for the wrapper to catch up.  CI/CD for the C++ gets kicked off manually, I believe.

@aothms  do I have that right?

aothms:
Yeah occasionally we have this as part of our imperfect distribution process. You can code more defensively with try-catch for this transition window, but that's also ugly later on. Most people in the community are used to this... I've started a new build.

RickBrice:
Thanks. I was afraid I made a major blunder on my first step into the Bonsai side of things.

PR #6292, opened 2025-03-08
Fix #5809: Add safety checks for write permissions before writing

### LINKED ISSUE #5809: ifcopenshell.file.write is unsafe if user don't have writing permission in the directory
Not sure about Unix but on Windows the snippet below will work without errors - user don't have writing permission for 'shared' folder and 'project.ifc' won't be saved silently.

```python
import ifcopenshell
from pathlib import Path

path = Path(r"\\shared-PC\shared") / "project.ifc"
ifc_file = ifcopenshell.file()
ifc_file.write(path)
```

As a partial and temporary solution I'll add a check whether file exists.

theseyan:
I've added checks for write permissions (`W_OK`) that will safely throw in case of missing permissions, before the filesystem is touched, instead of [silently failing](https://github.com/IfcOpenShell/IfcOpenShell/issues/5809).

I haven't tested this on a network share yet, as I need some pointers on running the tests with my local changes.

We can now remove the temporary workaround that checks if the file is present after the write.
Any review/feedback is very much appreciated!

Fixes #5809.


Andrej730:
I tried to test various cases of restricted access on Windows and Python is very inconsistent at detecting them, see below for the details. In theory we can just `Path.touch` file to make sure it's accessible enough for us but it's probably best just to leave it to our c++ part.
E.g. replace `write()` below
https://github.com/IfcOpenShell/IfcOpenShell/blob/069b630b160c71758e97ec8298dd2e342ccdc868/src/ifcwrap/IfcParseWrapper.i#L143-L146

with something like this as @aothms mentioned originally:
```python
	void write(const std::string& fn) {
		std::ofstream f(IfcUtil::path::from_utf8(fn).c_str());
		if (!f.good()) {
			throw std::runtime_error("Failed to write to '" + fn + "', check folder and file permissions.");
		}
		f << (*$self);
	}
```

and this should cover all restricted access cases.

-------

Testing Python access checks on Windows:

Was trying to test `if not path.exists()` block on Windows and I've created a folder with restricted access and what I'm getting weird results. For restricted folder `os.access(path.parent, os.W_OK)` still returns `True`, `path.exists()` fails for existing files and returns `False` for non-existing files.

```python
import os
from pathlib import Path

restricted = Path(r"L:\pomoika\restricted")

print("restricted folder")
print("exists", restricted.exists())  # True
# Returns True though access is restricted in the system.
print("access", os.access(restricted, os.W_OK))  # True


print("existing filepath")
# .exists() for existing filepath in restricted folder -> PermissionError
existing_filepath = restricted / "test.txt"
print("access", os.access(existing_filepath, os.W_OK))  # False
# PermissionError: [WinError 5] Access is denied: 'L:\\pomoika\\restricted\\test.txt'
print("exists", existing_filepath.exists())  # False


print("new filepath")
new_filepath = restricted / "newfile.txt"
# .exist() for non-existing filepath in restricted folder -> False
print("exists", new_filepath.exists())  # False
print("access", os.access(new_filepath, os.W_OK))  # False
# PermissionError: [Errno 13] Permission denied: 'L:\\pomoika\\restricted\\newfile.txt'
open(new_filepath, "w")
# PermissionError: [Errno 13] Permission denied: 'L:\\pomoika\\restricted\\newfile.txt'
new_filepath.touch()
```

Then I created `test.txt`, restricted access to it and got similar weird resuts - `os.access` shows `True` but editing file fails.
```python
import os
from pathlib import Path

restricted = Path(r"L:\pomoika\restricted")

print("restricted folder")
print("exists", restricted.exists()) # True
print("access", os.access(restricted, os.W_OK)) # True

print("existing restricted filepath")
existing_filepath = restricted / "test.txt"
print("exists", existing_filepath.exists()) # True
print("access", os.access(existing_filepath, os.W_OK)) # True
# PermissionError: [Errno 13] Permission denied: 'L:\\pomoika\\restricted\\test.txt'
open(existing_filepath, "w")
# PermissionError: [Errno 13] Permission denied: 'L:\\pomoika\\restricted\\test.txt'
existing_filepath.touch()
```

theseyan:
@Andrej730 Interesting find! I can reproduce the same odd behaviour on my machine.

Indeed, it's probably best to not have Python handle this at all; I've moved the access detection logic to C++ as advised.

Andrej730:
Thanks, merged.

PR #6319, opened 2025-03-11
Update profile.py to add MEPGenerator().setup_ports when the profile has only one segment (no DumbProfileJoiner called)

falken10vdl:
When adding a single duct in MEP no ports where attached. 
The function create_profiles_from_polyline does not call DumbProfileJoiner (which eventually calls MEPGenerator().setup_ports). I have added a check so if there is only one segment, the MEPGenerator().setup_ports is called explicitly.


Moult:
You've probably realised this is one of the messier areas of the codebase with lots of experimentation, incomplete features, and rewrites happening :)

falken10vdl:
Yes. I am following with detail the discussion in the forum: https://community.osarch.org/discussion/2764/i-need-a-ifcdistributionsystem-and-circuit-doctor-anyone to try to understand the topology and how SINK/SOURCE ports work so I can help if possible.
Thanks!

PR #6320, opened 2025-03-11
Update system.py so two ortogonal arrows (in x and y local) are used for MEP systems decorations

falken10vdl:
Currently the flow direction decorations run in the local Y. This means that if one is creating ducts/pipes in the xy plane, the decorations are drawn in planes perpendicular to that one and than means that the arrows are not visible from a top/bottom view. By adding another set of arrows perpendicular the arrows are always visible no matter what projection is taken.
image

### DIFF
--- a/src/bonsai/bonsai/tool/system.py
+++ b/src/bonsai/bonsai/tool/system.py
@@ -340,32 +340,37 @@ class System(bonsai.core.tool.System):
 
                 # for now it's hardcoded to local Y axis to avoid using viewport data
                 # for performance reasons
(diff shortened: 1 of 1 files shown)


Moult:
I'd like to instead investigate using local screen coordinates for this rather than this solution. But let's keep this PR open in case that isn't an option.

falken10vdl:
> I'd like to instead investigate using local screen coordinates for this rather than this solution. But let's keep this PR open in case that isn't an option.

Perfect. Thanks!

Moult:
Merging because perfect is the enemy of the good

PR #6324, opened 2025-03-11
Update add_prop_template.py so the properties are sorted

falken10vdl:
This allows to keep the pset properties sorted for easy access

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/pset_template/add_prop_template.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/pset_template/add_prop_template.py
@@ -110,5 +110,6 @@ def add_prop_template(
     )
     has_property_templates = list(pset_template.HasPropertyTemplates or [])
     has_property_templates.append(prop_template)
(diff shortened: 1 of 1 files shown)


src/ifcopenshell-python/ifcopenshell/api/pset_template/add_prop_template.py
```diff
@@ -110,5 +110,6 @@ def add_prop_template(
     )
     has_property_templates = list(pset_template.HasPropertyTemplates or [])
     has_property_templates.append(prop_template)
+    has_property_templates.sort(key=lambda pt: pt.Name)
```
Moult, line 113: or "Unnamed"?

falken10vdl: NewProperty is a nice choice. It could even be something like "000_NewProperty" if one wants to keep it almost sure at the top to easily spot the new addition to edit its name

Issue #6257, opened 2025-03-02
If local view is toggled, that it turns off grid decorations.


theoryshaw:
would be nice, if local view is toggled, that it turns off grid decorations. 

https://github.com/user-attachments/assets/3888b7c3-5a13-41f1-911f-9256799d40c1

Moult:
Already implemented.

PR #6335, opened 2025-03-13
If local view is toggled, that it turns off grid decorations. #6257

falken10vdl:
This PR addresses the issue  #6257.
Please see the discussion in https://community.osarch.org/discussion/2815/can-a-blender-addon-intercept-a-blender-function


falken10vdl:
Maybe this PR is not really needed and a hotkey to toggle grid is just a simpler already available solution.
As a beginner, it is a good exercise to understand better how operators work.

Issue #3129, opened 2023-05-09
Floor height precision loss


Gorgious56:
I assume it has to do with float precision inside of Blender, but it looks silly and doesn't export well in other softwares.
What can we do about the precision issue regarding floor height ? The docs seem to assume that only 2 digits after the point mark are sufficient, couldn't we truncate the values on validation ? I think the problem is the same for other attributes that store float values.

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/25156105/7e5103be-0ce8-4eeb-882d-da55936d7f78)
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/25156105/ba822181-c530-43b9-a9b3-751884665ff5)
`#98=IFCBUILDINGSTOREY('33m2dtCMnAuQQYgAG0o27J',$,'4 - Combles',$,$,#108,$,$,$,67.7399978637695);`
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/25156105/0bb76110-9198-4f2b-b3f1-d2368efafb5d)

Moult:
I think there's a few bits here, I think we should no longer generalize these as floats, and instead target lengths if its a length. Then we should round to the precision of the IFC.

Moult:
Right now we hardcode precision during project creation which is also wrong, it should be unit dependent.

BIMvoice:
This is float32 storage in Blender. Bonsai keeps attribute values in a Blender `FloatProperty`, which is single precision, so a clean `67.74` becomes `67.7399978637695` the moment it's loaded, and that's what gets written on export.

Following your first point, I've rounded length-measure attributes to the model's precision when writing them (derived from `IfcGeometricRepresentationContext.Precision`, in project units). Storey `Elevation` now exports as `67.74`, and genuinely precise values are kept as-is. Fixed in #8337.

Your second point still stands: `add_context` hardcodes `Precision` to `1e-5` regardless of unit. I've left that as a separate follow-up since it's a core-API change, but the rounding already reads the real precision when it's set.

PR #6349, opened 2025-03-14
[Bonsai] Add proper rounding of float attributes by model precision

theseyan:
In the `Attribute` class, `get_value` method was not properly handling `float` value attributes and simply passing them as-is, which can trigger some surprising float representation issues in Python (since it tries to represent each `float` in the highest precision).

The correct behaviour should be to round off float attributes up to the specified model precision ([documentation](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcGeometricRepresentationContext.htm)), or default to `1e-6` precision.


theseyan:
Possibly related issue: #3129

PR #6368, opened 2025-03-16
Update style.py with StyleType for "Internal"

falken10vdl:
When  clicking BIMSolarProperties.shadow_mode if RENDERING or SHADING is selected. The update sets StyleType to "Internal" which is an enum not defined.


Andrej730:
Sorry, there is no "Internal", it was just a typo :(

Issue #4771, opened 2024-05-30
expand/contract all the material categories at one time. 


theoryshaw:
Would be nice to have a way to expand/contract all the material categories at one time. 

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/2968e8bb-7a18-4590-9397-3e1075a73baf)

manuvarkey:
Works just fine. Just one minor detail. The description is shown repeated.
![WhatsApp Image 2024-06-01 at 11 09 43 AM](https://github.com/IfcOpenShell/IfcOpenShell/assets/4415312/47772716-5c4b-4287-abd8-cefe34133b16)
![WhatsApp Image 2024-06-01 at 11 09 41 AM](https://github.com/IfcOpenShell/IfcOpenShell/assets/4415312/59a8455f-3062-40de-970a-494c284fa33b)

theoryshaw:
This one doesn't seem to work for me anymore.

PR #6379, opened 2025-03-17
Issue #5130 fix + sheet all expand and contract, based on #4771

### LINKED ISSUE #5130: expand/collapse all drawing contexts at one time.
Similar to #4771

Would be nice to expand/collapse all drawing contexts at one time.

2024-08-02_10-28-14_blender_(Unsaved)_-_Blender_4 1blender

jes-r:


jes-r:
Dont know how to fix the mathutils error as i havent used it?

Andrej730:
> Dont know how to fix the mathutils error as i havent used it?

You mean the one from github workflows? Never mind it if it's unrelated to the PR.

Andrej730:
Merged, thank you.

jes-r:
> You mean the one from github workflows? Never mind it if it's unrelated to the PR.

yes that one, thanks

src/bonsai/bonsai/bim/module/drawing/operator.py
```diff
@@ -3291,14 +3291,22 @@ def _execute(self, context):
 class ExpandTargetView(bpy.types.Operator):
     bl_idname = "bim.expand_target_view"
     bl_label = "Expand Target View"
-    bl_description = "Show views in this category"
+    bl_description = "\nSHIFT+CLICK to expand all view categories "
```
sboddy, line 3294: I don't think you meant to delete the original description entirely here.

jes-r: Well it was intentional though not optimal as I see it now:

image
This is how it is now with the merged version. 

I figured the old description was double and thus unneeded info, so removed it.

Looking at a earlier relatable case i think bl_label should be "Show views in this category" or "Expand View Category" instead of Target **View** as you are not expanding the view but the category.

But maybe the label needs to stay the same for other reasons @Andrej730?


TLDR: Compact / only essential info is better I think and less confusing, but maybe the label should be changed, just like material: 
image

Issue #6385, opened 2025-03-17
(Memory) Release/delete `IfcGeom::Element` earlier


sebjf:
I have been examining memory usage of the `IfcGeom::Iterator`, as we have been noticing high consumption with some larger IFCs.

I was wondering if the maintainers would allow IfcGeom::Elements to be deleted earlier in order to reduce peak memory consumption?

**Description**

My understanding is that in [`create_element_`](https://github.com/IfcOpenShell/IfcOpenShell/blob/09a43f13395ff23b237b0429a21e28fa6b190fda/src/ifcgeom/Iterator.h#L521) an `IfcGeom::Representation::BRep` is created for the representation of a `geometry_conversion_result`, and a shared pointer to this stored for each _Product_ that references that representation. The pointer is stored in both an `IfcGeom::BRepElement` and an `IfcGeom::Element`, which are stored in these vectors, respectively:

https://github.com/IfcOpenShell/IfcOpenShell/blob/09a43f13395ff23b237b0429a21e28fa6b190fda/src/ifcgeom/Iterator.h#L118-L119

The geometry iterator's `get()` method then returns a pointer to the `IfcGeom::Element` (or `IfcGeom::BRepElement` if using `get_native()`) from these.

When the Iterator [destructor](https://github.com/IfcOpenShell/IfcOpenShell/blob/09a43f13395ff23b237b0429a21e28fa6b190fda/src/ifcgeom/Iterator.h#L884) runs, each Element or BRepElement is deleted.

The implication is that the Representation::BRep (and in fact the entire Element and BRepElements) will be kept around even after a serializer has handled them and moved onto the next one.

This means for larger IFCs the Iterator can allocate quite a bit of memory that it doesn't need to.

**Validation**

As a quick test, I removed the delete calls from the destructor and instead deleted the Element and BRepElement inside the iterator loop in IfcConvert.cpp. To remove the overhead of actual serialisation I also commented out the `write()` call.

One one test IFC, the memory was reduced from 12.6 Gb to 9.3 Gb when serialising with 16 threads, and from 12.6 Gb to 5.0 Gb when serialising with 1 thread, just from moving the delete calls.

**Feature Request**

I can understand why the iterator works like this: if the API relied on the user deleting the Element, they won't! Shared pointers are also non-zero cost.

However is there another acceptable approach?

[...]

PR #6411, opened 2025-03-21
Issue 6385 - Delete IfcGeom::Elements as soon as possible

sebjf:
This PR updates the geometry iterator to delete IfcGeom::Elements once the user has passed them, thereby reducing peak memory usage.

Issue: https://github.com/IfcOpenShell/IfcOpenShell/issues/6385


aothms:
Thanks!

PR #6417, opened 2025-03-23
Resolve memory leaks when using `create_shape()` in ifcopenshell-python

shmoodyyy:
When calling the `create_shape(...)` function from the python wrapper, the kernel instance that lives in the templated helper function as a stack allocation seemed to have 100+ new leaking allocations per call according to heap profiling.

From my super basic testing script (see below) it looks like that fixed all of the leaking memory allocations, with any growth of total memory allocated stemming from just the logging buffer growing in size.

This is a very straight-forward and naive fix for the problem, as the Converter instance and the kernels that are created for **each** call of creating a shape reads to me like it should be something that's handled within the library as a singleton or static of some type, although I am very unfamiliar with the codebase as a whole so I've decided to keep it to a simple delete where it felt it makes most sense solution.

```python
import sys
import ifcopenshell
from ifcopenshell import geom
import psutil
import os
import gc

settings = geom.settings()
model = ifcopenshell.open("IfcOpenHouse.ifc")
entity = model.by_type("IfcWall")[0]

process = psutil.Process(os.getpid())
base_mem = process.memory_info().rss / 1024 / 1024
print(f"Baseline: {base_mem:.2f} MB")

for i in range(1000):
    shape = geom.create_shape(settings, entity)
    shape = None
    gc.collect() 
    mem = process.memory_info().rss / 1024 / 1024
    if i % 100 == 0:
        print(f"Iteration {i}: {mem:.2f} MB")

final_mem = process.memory_info().rss / 1024 / 1024
print(f"Final: {final_mem:.2f} MB")

input()
```

|before|after|
|-|-|
|image|image|

### DIFF
--- a/src/ifcgeom/AbstractKernel.cpp
+++ b/src/ifcgeom/AbstractKernel.cpp
@@ -206,6 +206,8 @@ ifcopenshell::geometry::kernels::AbstractKernel* ifcopenshell::geometry::kernels
 			}
 #endif
 			if (kernels.size() != n + 1) {
+                for (auto k : kernels)
+                    delete k;
 				throw IfcParse::IfcException("Invalid hybrid kernel " + geometry_library);
 			}
 		}
--- a/src/ifcgeom/AbstractKernel.cpp
+++ b/src/ifcgeom/AbstractKernel.cpp
@@ -206,8 +206,8 @@ ifcopenshell::geometry::kernels::AbstractKernel* ifcopenshell::geometry::kernels
 			}
 #endif
 			if (kernels.size() != n + 1) {
-                for (auto k : kernels)
-                    delete k;
+				for (auto k : kernels)
+					delete k;
 				throw IfcParse::IfcException("Invalid hybrid kernel " + geometry_library);
 			}
 		}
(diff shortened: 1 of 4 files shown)


aothms:
Many thanks, I'll have an actual look tomorrow, just:

> as the Converter instance and the kernels that are created for each call of creating a shape reads to me like it should be something that's handled within the library as a singleton or static of some type

Typically, users instantiate ifcopenshell.geom.iterator to iterate over a (subset of) the geometrical elements in the IFC model. As you said this would only instantiate the converter/kernel once, but also able to exploit shared geometrical definitions between elements for efficiency.

shmoodyyy:
> Many thanks, I'll have an actual look tomorrow, just:
> 
> 
> 
> > as the Converter instance and the kernels that are created for each call of creating a shape reads to me like it should be something that's handled within the library as a singleton or static of some type
> 
> 
> 
> Typically, users instantiate ifcopenshell.geom.iterator to iterate over a (subset of) the geometrical elements in the IFC model. As you said this would only instantiate the converter/kernel once, but also able to exploit shared geometrical definitions between elements for efficiency.

Sounds like a cleaner interface for dealing with a large amount of shapes as we do, which could have resolved the issue of us having to resort to killing and relaunching our ifc pods to clean up the used memory.

Although the leaks happen within the converter/kernel themselves, which in this case I feel like the deletes I've added in should cover the lifetimes of the allocations, haven't had the time to get the full tests running for this sunday afternoon.

aothms:
Thanks again. Looks good. I'll probably change some things to be unique_ptr, later on.

PR #6441, opened 2025-03-25
Fix: Add virtual destructor to AbstractKernel base class

ArjoNagelhout:
This fixes a `SIGTRAP` when destructing an `IfcGeom::Iterator`, as that calls `delete` in `ifcopenshell::geometry::Converter::~Converter()` on the field `ifcopenshell::geometry::kernels::AbstractKernel* kernel_;`. This results in UB in C++. 

Adding a virtual destructor to the base class `AbstractKernel` fixes this issue.

### DIFF
--- a/src/ifcgeom/AbstractKernel.h
+++ b/src/ifcgeom/AbstractKernel.h
@@ -44,6 +44,8 @@ namespace ifcopenshell {
 			: geometry_library_(geometry_library)
 			, settings_(settings) {}
 
+		virtual ~AbstractKernel() = default;
+
 		virtual bool convert(const taxonomy::ptr, IfcGeom::ConversionResults&);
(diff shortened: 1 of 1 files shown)


aothms:
Awesome thanks.

So actually this did get recently introduced; apparently these things were never freed before 🙈  https://github.com/IfcOpenShell/IfcOpenShell/commit/eec8c88e6f8909081705609d38d52210715410cc https://github.com/IfcOpenShell/IfcOpenShell/commit/c6e3ac4155dac0b1aa4d6d30c17d634d0b8c362a

ArjoNagelhout:
I see! Good that they're now getting freed :)

Using `std::unique_ptr` in more places might make it harder to accidentally forget calling `delete`, but I see this is already in progress.

PR #6448, opened 2025-03-26
Added c++ Hello World example

Adarshgupta1127:
Added a basic "Hello World" example in C++ to the documentation. 
This provides a starting point for new C++ users of IfcOpenShell.


aothms (changes requested):
Hi, this example does not actually seem to use IfcOpenShell. Have a look at some of the examples we currently have and generalize that into something generic. https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/examples/IfcParseExamples.cpp

PR #6450, opened 2025-03-27
[WebUI] Fix jQuery CDN links

theseyan:
The current `webui` fails to load in the browser because it imports jQuery via the jQuery CDN (https://code.jquery.com/jquery-3.6.0.min.js) which is often down or unresponsive.

This PR simply replaces all jQuery CDN links with the equivalent jsDelivr CDN links (https://cdn.jsdelivr.net/npm/jquery@3.6.0/dist/jquery.min.js) which is way faster and more reliable.
Tested to fix the loading issue.


Moult:
I didn't know that jQuery's officially recommended CDN was not reliable. I think a better solution here would be to bundle the JS offline as part of our build step in the Makefile. We shouldn't require online access to use the web UI. Similar to how we grab the jsgantt lib. Would you be interested in implementing this?

theseyan:
@Moult Even better! I've made the changes, please have a look if `/static/js` is the correct folder to put the jQuery script in.

theseyan:
Bonsai's live development doc scripts have been updated as well.

PR #6494, opened 2025-04-02
Added Quickstart guides for Linux and Windows IDEs using VSCode

falken10vdl:
These are documentation guides to setup a development environement for Linux and Windows using VSCode. Also some guidance on how to use Inkscape to generate documentation as per Bonsai doc style.
Thanks!


theoryshaw:
@Andrej730 can you pull, if this looks somewhat good to you?

theoryshaw:
Better just to get this out in the wild.:)

Thanks @falken10vdl!

falken10vdl:
@theoryshaw Thank you! Let's see peoples's feedback to improve it.

theoryshaw:
Is basically just easier to review this way. 

Of the cuff, one thing that seems in need of reconciling is the difference in .bat and .sh scripts.  There's some redundancy there....

- https://github.com/IfcOpenShell/IfcOpenShell/tree/v0.8.0/src/bonsai/scripts/installation
and
- https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/bonsai/docs/quickstart/ide/linux/dev_environment.sh and https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/bonsai/docs/quickstart/ide/windows/dev_environment.bat

I think we need some eyes on these from the core developers

falken10vdl:
I agree. I used some scripts that were available from previous docs but it would be best to get a proper view on that. I have also found that gitignore seems to not behave the same in Linux and Windows.
Thanks!

PR #6503, opened 2025-04-04
Update ui.py So clicking the gear also updates the thumbnai

falken10vdl:
In some circunstances the thumbnail of namely doors/windows are not updated when we update teh parametric information of the type. This provides a way to get the thumbnail update triggered by the user.


falken10vdl:
As received feedback from https://community.osarch.org/discussion/2858/door-type-manager-previews#latest it does not look a good option to do this.
Will look for something better :)

PR #6517, opened 2025-04-07
Added entry in type menu to call for a thumbnail refresh

falken10vdl:
Sometimes the thumbnails are not refreshed (for example when changing the geometry in the door/window parametric too). This adds an entry to call to the refresh_thumbnail function.

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/ui.py
+++ b/src/bonsai/bonsai/bim/module/model/ui.py
@@ -70,7 +70,6 @@ class BIM_MT_type_menu(bpy.types.Menu):
         op.element = props.menu_relating_type_id
 
 
-
 class LaunchTypeMenu(bpy.types.Operator):
     bl_idname = "bim.launch_type_menu"
     bl_label = "Launch Type Menu"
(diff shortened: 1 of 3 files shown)


falken10vdl:
See rationale in https://community.osarch.org/discussion/comment/24688#Comment_24688

https://github.com/user-attachments/assets/9afdf76f-6a1f-4f99-acb8-83066efb7cf4

Thanks!

Andrej730:
I'm not sure if this is the right approach, it might create an expectation for user that it's okay for thumbnails to go out of date and they cannot trust them and should alwars refresh them manually just in case. 
I think we should address the root cause and mark somehow previews as outdated when they go out of date, so they can be refreshed in time automatically.

falken10vdl:
That of course would be best. In my experience I have seen this happening in Linux mainly when using the parametric window/tool. Samppa initiated the thread https://community.osarch.org/discussion/comment/24688#Comment_24688 He might have other cases in which that is happening.

Something like this?
https://github.com/IfcOpenShell/IfcOpenShell/pull/6656

Thanks!

Andrej730:
Closing, as more proper solution is added in #6656.

Issue #6340, opened 2025-03-13
HDF5 caching throws exception when default materials are applied


sebjf:
**Error Description and Steps to Reproduce**

When an Ifc without proper material assignment is serialised with HDF5 based caching enabled, `HdfSerializer` will throw an access violation exception because the default material does not have an instance.

**Steps to Reproduce**

1. Convert this file from Ifc to Obj with IfcConver, with the `--cache` flag set:

[Column_4x3_noMaterials.ifc.txt](https://github.com/user-attachments/files/19232842/Column_4x3_noMaterials.ifc.txt)

2. Observe an unhandled exception here: 

https://github.com/IfcOpenShell/IfcOpenShell/blob/4daf4693f392732084dfd812d329722b194edf52/src/serializers/HdfSerializer.cpp#L547

**Debug Information**

This is caused because the default material applied is not an actual entity, so `s.instance` is null.

When the style object is [created](https://github.com/IfcOpenShell/IfcOpenShell/blob/4daf4693f392732084dfd812d329722b194edf52/src/ifcgeom/SurfaceStyle.cpp#L16) the superclass `item` constructor is not called/called with its default arguments, which sets instance to nullptr.

**Workaround**

Setting ApplyDefaultMaterials to false will avoid the error, if the host application can handle the case where an `IfcGeom::Representation::Triangulation` has no materials.

**Fix**

The `item` superclass has an `_identity` member, which is independent of the `IfcBaseInterface` instance.

Would it be acceptable to use this instead for the id in `write_style`? (I can create a PR if so.)

aothms:
Thanks for the detailed report. Much appreciated. Sorry for the delay in replying. I'd rather set it to 0 when item->instance == nullptr, because the item identity doesn't have much significance outside of the current invocation. It's probably deterministic when ran from IfcConvert, but e.g from python using an arbitrary order of files being processed, the id would not identify much.

A PR would surely be appreciated!

sebjf:
@aothms, PR that sets id to 0 instead is [here](https://github.com/IfcOpenShell/IfcOpenShell/pull/6534)!

PR #6534, opened 2025-04-11
ISSUE #6340 check if material is null before trying to get id

sebjf:
This PR checks if the conversion to IfcUtil::IfcBaseClass is successful before trying to assign the id when serialising a style to HDF5.

The style instance may be null if the material is a default material.

Discussion and test case are in the issue: https://github.com/IfcOpenShell/IfcOpenShell/issues/6340


aothms:
Many thanks!

PR #6539, opened 2025-04-12
Add issue templates that allocate labels automatically for easier Triage

falken10vdl:
Here are issue templates for Bugs and Feature requests for Bonsai, IfcOpenShell-Core and IfcOpenShell-Python with severities: Critical, Major and Minor. The labels are assigned based on the selected template.

image

Thanks!


Moult:
I've updated the templates to auto fill the Type attribute, but I can't find a way to request labels specifically, and not a fan of creating a ton of templates which might be intimidating..

falken10vdl:
Sure. It was just a possibility but yes, this might not be well received.
Thanks!

Issue #6337, opened 2025-03-13
Error with relative path and linked projects when printing drawing


Breogan10:
I'm exploring the Bonsai's ability to work with linked projects, and creating drawings.

I've found that generating a drawing, which includes a linked project, works fine if linked project's PATH IS ABSOLUTE. When the linked project's path is set to be relative (as the drawing itself, for instance), NOT CHANGING ANYTHING ELSE like file path, name, etc, then the error arises as follows:

```
Python: Traceback (most recent call last):
  File "C:\Users\glopez\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\operator.py", line 298, in execute
    linework_svg = self.generate_linework(context)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\glopez\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\operator.py", line 719, in generate_linework
    IfcStore.session_files[link.name] = ifcopenshell.open(link.name)
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\glopez\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\ifcopenshell\__init__.py", line 162, in open
    raise FileNotFoundError(f"File does not exist: '{path}'.")
FileNotFoundError: File does not exist: 'untitled_.ifc'.
```

Test to reproduce this error can be done with two blank projects, one linked to the other.

```
os: Windows
os_version: 10.0.19041
python_version: 3.11.9
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: Intel64 Family 6 Model 142 Stepping 10, GenuineIntel
blender_version: 4.3.2
bonsai_version: 0.8.1
bonsai_commit_hash: None
last_actions: 
_
last_error: None
ifc: untitled.ifc
schema: IFC4X3
preprocessor_version: IfcOpenShell 0.8.1
originating_system: Bonsai 0.8.1
```

falken10vdl:
I have seen that if the line below is changed so that an absolute path is calculated form the relative one, the error is solved.

[...]

PR #6540, opened 2025-04-13
Update operator.py so relative link to ifc file works

falken10vdl:
This is a hack for 
https://github.com/IfcOpenShell/IfcOpenShell/issues/6337 If relative path is used the operator fails. By claculating the correspodning absolute path the issue is solved.

### DIFF
--- a/src/bonsai/bonsai/bim/module/project/operator.py
+++ b/src/bonsai/bonsai/bim/module/project/operator.py
@@ -1303,7 +1303,7 @@ class LoadLink(bpy.types.Operator):
     filepath_: Path
 
     def execute(self, context):
-        filepath = Path(tool.Ifc.resolve_uri(self.filepath))
+        filepath = Path(tool.Ifc.resolve_uri(tool.Ifc.get_uri(self.filepath, use_relative_path=False)))
         if not filepath.exists():
             self.report({"ERROR"}, f"File does not exist: '{filepath}'")
             return {"CANCELLED"}


Andrej730:
@falken10vdl Hi! Sorry, I don't think I understand this change and how it fixes the issue.

```python
filepath = Path(tool.Ifc.resolve_uri(tool.Ifc.get_uri(self.filepath, use_relative_path=False)))
```

1. `self.filepath = 'linked.ifc'`
2. `get_uri(linked.ifc, False)` resolves it to absolute path relative to cwd. So e.g. it's now C:\test.ifc
3. `resolve_uri(C:\test.ifc)` - the path is already absolute, so it just stays C:\test.ifc

So when I load main.ifc and try to load the link I get `File does not exist: 'L:\Blender\linked.ifc'`.

[linked.ifc.txt](https://github.com/user-attachments/files/19834276/linked.ifc.txt)
[main.ifc.txt](https://github.com/user-attachments/files/19834277/main.ifc.txt)

Shouldn't we resolve paths in drawing operator that triggers the issue instead?

falken10vdl:
Hi @Andrej730 Sorry I cannot reproduce now the issue that was solved with the simple hack. 
Testing now I see is that the implementation of relative path for linking files is not working properly. I think that is independent of the drawing operator part.
If from the relative path one recreates the absolute before calling a number of functions the issue is solved. But that forces to changes in operator, prop and uy in module/project which propably is not what it is aimed to be done.
For me, I do not understand:
class Link(PropertyGroup):
    name: StringProperty(
        name="Name",
        description="Filepath to linked .ifc file, stored in posix format (could be relative to .blend file, not to .ifc)",

what .blend file are we talking about? if we do not have cache and do not store any blend file?

@maxfb87 Can you help on this? Were you doing this part of code linking files? I think relative paths are not working regardless of drawing docs.

falken10vdl:
Ok. I found that os.getcwd() (called at some point) is returning different values if I launch blender form VSCode (my case) or if I launch it standalone.
I have now configured properly VSCode https://github.com/JacquesLucke/blender_vscode/issues/230 and will take a deeper look.

maxfb87:
@falken10vdl maybe it's helpful to mention that i actually have found some strange behavior with "use_relative_path" enables when i was looking at reloading linked file: in linux no error and on windows it triggers an error.
I wasn't able to test completely because i haven't always the windows machine but i will try.

`what .blend file are we talking about? if we do not have cache and do not store any blend file?`
It is referring to .blend file used to link a file before "link ifc" was implemented.
Not an expert on this but it is a blender feature.

Andrej730:
> description="Filepath to linked .ifc file, stored in posix format (could be relative to .blend file, not to .ifc)",

This description is no longer valid, fixed it in https://github.com/IfcOpenShell/IfcOpenShell/commit/305439d27e6b20d993fff0356ee6e7e4b60e64c7
In good old days we were supporting linking relative to the current .blend file, but not anymore, because now we store linked paths in .ifc itself, so now if they're relative they're relative to .ifc, not to .blend.

> Testing now I see is that the implementation of relative path for linking files is not working properly. I think that is independent of the drawing operator part.

Can you show an example?

---

I think the original problem lies in the drawing operator, see the traceback.

```python
  File "C:\Users\glopez\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\operator.py", line 298, in execute
    linework_svg = self.generate_linework(context)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\glopez\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\operator.py", line 719, in generate_linework
    IfcStore.session_files[link.name] = ifcopenshell.open(link.name)
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\glopez\AppData\Roaming\Blender Foundation\Blender\4.3\extensions\.local\lib\python3.11\site-packages\ifcopenshell\__init__.py", line 162, in open
    raise FileNotFoundError(f"File does not exist: '{path}'.")
FileNotFoundError: File does not exist: 'untitled_.ifc'.
```

falken10vdl:
Thanks @maxfb87 and @Andrej730 . I see there is a problem wiht linked files with relative paths both in windows and linux if the path is "moving upwards". Maybe there is still another issue with the drawing operator as well.
image

@Andrej730 This can be reproduced with the files you have attached. The only thing is that you have to move the linked.ifc file upwards relative to the main.ifc:
image

Thanks!

falken10vdl:
@Andrej730 @maxfb87 I think there is some issue on how resolve_uri and get_uri work.

- get_uri() for the absolute path is using get_cwd() data and that is not always pointing to the same directory as the ifc file directory (incidentally that is how I found that vscode was setting that to the directory root while windows was setting to the blender executable directory). That means that it will generate wrong paths for the absolute case. In order to fix it it should be something like this I believe.
image

-resolve_uri() is joining directory path with relative to get the absolute. Although I think that is valid syntax (ex: /home/falken10vdl/Documents/sampleIFC/../../Downloads/3walls.ifc) probably is just cleaner to generate a consolidated one (ex: /home/falken10vdl/Donwloads/3walls.ifc). **That solves the error in my testing**. That can be achived with something like this.
image

Let me know if you agree and I can send a PR for those changes 
Thanks!

Andrej730:
> Thanks maxfb87 and Andrej730 . I see there is a problem wiht linked files with relative paths both in windows and linux if the path is "moving upwards". Maybe there is still another issue with the drawing operator as well.

Great example, thanks. The issue here is that we wasn't resolving input filepath when trying to find linked file's scene.
So basically we were relying on something like below, that should be fixed after the commit above.
```python
Path(r"C:\Users\Andrej\..") == Path(r"C:\Users") # False
```

---

I'm not sure I want to touch get_uri / resolve_uri unless there is a specific case that's affected by their current behaviour. I think the original error #6337 is all about `bim.create_drawing` behaviour and unrelated to get_uri / resolve_uri.

falken10vdl:
@Andrej730 I agree. I saw that error happening for example here:
image

Yes maybe not touch them :) For resolve_uri the syntax looks always good although not compact but for the case of get_uri I still do not understand what is it supposed to return when you ask for a absolute path and provide a relative one with ../../ To me it will return something not right...

Currently it is behaving like this:
```
>>> import bonsai.tool as Tool
>>> filepath="../../patata.ifc"
>>> import os
>>> os.getcwd()
'/home/falken10vdl'

>>> bpy.context.scene.BIMProperties['ifc_file']
'/home/falken10vdl/Documents/sampleIFC/Testing.ifc'

>>> Tool.Ifc.get_uri(filepath,True)
'../../patata.ifc'

>>> Tool.Ifc.get_uri(filepath,False)
'/patata.ifc'

>>> Tool.Ifc.resolve_uri(filepath)
'/home/falken10vdl/Documents/sampleIFC/../../patata.ifc'


```
And this will be behaviour with the updated code:


```
>>> import bonsai.tool as Tool
>>> filepath="../../patata.ifc"
>>> import os
>>> os.getcwd()
'/home/falken10vdl'

>>> bpy.context.scene.BIMProperties['ifc_file']
'/home/falken10vdl/Documents/sampleIFC/Testing.ifc'

>>> Tool.Ifc.get_uri(filepath,True)
'../../patata.ifc'

>>> Tool.Ifc.get_uri(filepath,False)
'/home/falken10vdl/patata.ifc'

>>> Tool.Ifc.resolve_uri(filepath)
'/home/falken10vdl/patata.ifc'

```
To me Tool.Ifc.get_uri(filepath,False) does not look right but I do not know if that is expected behavior. Tool.Ifc.resolve_uri(filepath) is fine, just not compact so that was leading to the inequality error.
Thanks!

falken10vdl:
Issue solved with commit given by @Andrej730 .
Thanks!

PR #6563, opened 2025-04-16
Qty export multiple files

falken10vdl:
This is a PR following conversation https://community.osarch.org/discussion/2864/csv-export-from-multiple-models-future-patch#latest
It allows to export data from multiple files
The relevant documentation is also added.
Thanks!


Moult:
Wow this is a big PR. A huge thanks! I haven't had a chance to look in detail, but it looks as though you've created a new way to add linked models. Why don't you just reuse the existing Links panel / feature? Or have I misunderstood?

falken10vdl:
Yes, @steverugi suggested to put maybe a tick to take care of the linked files. I just felt that maybe one wanted to see in the same place all the files that one wants to use for. Qty to spreadsheet export (and the ability to select maybe some but not all). I was also thinking that if you have different LoD you might want to add files which are not linked so a UIList open to whatever one wants. Both for linked and not linked files.
Thanks!

Moult:
I'd prefer to keep it part of the "Links" system, but perhaps allow for "dormant" (? is there a better word) links. Or maybe two types of links, geometric links and data links?

A link (either geometric or data) can then be selectively enabled or disabled for IfcCSV, but there is a master list of links. This way it can also be reused for things like IfcFM, IfcClash, search filters across models, drawing visibility, IfcTester, etc, with reusable UI.

falken10vdl:
Ok. I am not sure I am following you :). The "Add linked files" is not doing anything except copying the filepaths from the links section for convenience. The files in the UIList are then treated the same way. They are processed very much like single files one after another.

falken10vdl:
After reading a little bit more maybe I understand better. You want to have the links section maybe adding enable / disable for csv export (or other actions as you say). An then keep the non liked files in the import/export panel adding a tick to take in/or not account linked files? I can update the PR like that if you guys think it is better. Let me know. Thanks!

steverugi:
@falken10vdl 
do you think it's possible to have a mockup of what the final panel is going to look like? a snapshot would do I guess
Sorry I lost track of the changes we want to implement
thanks

falken10vdl:
@steverugi Currently the PR has the panel as given in the documentation (see attached pdf)
[Spreadsheet Export - Bonsai 0.8.2 documentation.pdf](https://github.com/user-attachments/files/19835102/Spreadsheet.Export.-.Bonsai.0.8.2.documentation.pdf)

So the linked files part is as follows:
image
I simply copy the pathfiles from one panel to the other for convenience. Nothing else.
Anyhow if that is creating confusion we can adapt it as you think is easiest to understand.
Please note that my mother tongue is Spanish so maybe my words are not the best to be clearly understood :)

steverugi:
Hola @falken10vdl :)

OK so in practice the already linked models are brought in the panel where we can select files to be used in the new Spreadsheet I/O, allowing the input from multiple models instead of one as it is now. 
I don't personally see any confusion there.
It's perfect, looking forward to using it.

Thanks a lot for your help

falken10vdl:
Correcto :) Yes, it is just for convenience. Nothing is touched from the logic/handling of linked models. It is just that in spreadsheet I/O one can select whatever files one wants with the file explorer to use for exporting. One could select really files not related to the current ifc file and I was just thinking that since we already have the paths for the linked models, with the push of a button we can populate the list.
Thanks!

steverugi:
Hi @falken10vdl 
please any udpate on your PR? it's something I'd really love to see implemented
thanks for your contribution, much appreciated

falken10vdl:
Hi @steverugi . I guess the core developers are taking a look to it. I will be glad to update whatever they think is needed.
In the meantime if you want you can use it by installing the dev environment and pulling it. Find attached the documentation to get that installed in you windows pc (step 15 is exactly what you are after). Maybe you can try it and provide feedback if something is not working for you ( https://github.com/IfcOpenShell/IfcOpenShell/pull/6494  ). 
Cheers!
[Windows - Bonsai 0.8.2 IDE QS.zip](https://github.com/user-attachments/files/20263540/Windows.-.Bonsai.0.8.2.IDE.QS.zip)
[windows_files.zip](https://github.com/user-attachments/files/20263551/windows_files.zip)

falken10vdl:
Sorry. I just tried to rebase with vscode to the current base branch and it looks it has pushed everything in between instead of applying the changes to the fresh base. Clean in https://github.com/IfcOpenShell/IfcOpenShell/pull/6717

PR #6617, opened 2025-04-24
Use tool.Blender.get_view3d_area() to allow the use of separate windows

FabioPiccinini:
I replaced
next(area for area in bpy.context.screen.areas if area.type == "VIEW_3D")
with
tool.Blender.get_view3d_area()

Since the function had already been created and allows the use of separate windows.


Andrej730:
Thanks for the contribution.

PR #6638, opened 2025-04-29
User UI customization

falken10vdl:
This is a simple customization of the UI that allows to selectively hide or see TABS/Panels and also allow to move panels to a "BOOKMARK" TAB that behaves as a talisman place were one can craft a personalized TAB with the desired panels.
Seen the discussion in https://community.osarch.org/discussion/2900/simple-customization-of-panel-layout-to-clear-arrange-space-for-different-workflows

https://github.com/user-attachments/assets/17fc6de4-d34a-44f6-a8eb-91e5787bf0b4

Thanks!


falken10vdl:
@brunopostle @sboddy @theoryshaw @Andrej730 @Moult 

What do you think about this PR?

Initially it started as just UI customization https://community.osarch.org/discussion/2900/simple-customization-of-panel-layout-to-clear-arrange-space-for-different-workflows but then when I was not able to save the session information about the panel order (without too much pain) I looked to the option to just save the non bonsai info in a blend file as was discussed in https://community.osarch.org/discussion/3235/ifc-blend-sync-customer-user-ui-revisited

Since this is affecting the behavior of saving project I have put everything as an extra option for the user to decide:
image

Cheers!

PR #6650, opened 2025-05-02
validate ifcopenshell.simple_spf.file

Ghesselink:
In the validation service, we're currently using` ifcopenshell.validate` to [pre-validate IFC headers](https://github.com/buildingSMART/validate/blob/9bed38ff2a82b4ff9d070c8170ae8492200975b9/backend/apps/ifc_validation/checks/header_policy/validate_header.py#L31). However, since we're transitioning away from wrapping the entire IFC file for this task and instead adopting the lightweight `simple_spf `parser, the validate_ifc_header function should also support this new file type.

For reference, see the parse-only API introduced in [this commit](https://github.com/IfcOpenShell/step-file-parser/pull/13/commits/211a3055f437206ddcbcdaf401658b37bd170d71).

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/validate.py
+++ b/src/ifcopenshell-python/ifcopenshell/validate.py
@@ -570,7 +570,7 @@ def to_string_header_entity(header_entity):
     # Prefer native .toString() if available (native IfcOpenShell wrapper)
     if isinstance(header_entity, W.HeaderEntity):
         return header_entity.toString()
-    elif isinstance(header_entity, tuple):
+    elif hasattr(header_entity, '_fields'):
         values = [repr(getattr(header_entity, f)) for f in header_entity._fields]
         return f"{type(header_entity).__name__.upper()}({','.join(values)})"
     else:
--- a/src/ifcopenshell-python/ifcopenshell/validate.py
+++ b/src/ifcopenshell-python/ifcopenshell/validate.py
@@ -568,9 +568,9 @@ def to_string_header_entity(header_entity):
     """Recreate IFC header string representation, like FILE_NAME(...)"""
     
     # Prefer native .toString() if available (native IfcOpenShell wrapper)
-    if hasattr(header_entity, 'toString'):
+    if isinstance(header_entity, W.HeaderEntity):
         return header_entity.toString()
-    elif hasattr(header_entity, '_fields'):
+    elif isinstance(header_entity, tuple):
         values = [repr(getattr(header_entity, f)) for f in header_entity._fields]
         return f"{type(header_entity).__name__.upper()}({','.join(values)})"
     else:
@@ -581,7 +581,7 @@ def validate_ifc_header(f: Union[ifcopenshell.file, ifcopenshell.simple_spf.file
     AGGREGATE_TYPE = "LIST [ 1 : ? ] OF STRING (256)"
     STRING_TYPE = "STRING (256)"
 
-    def log_error(header_entity: W.HeaderEntity, name: str, index: int, expected_type: str, provided_type: str) -> None:
+    def log_error(header_entity: Union[W.HeaderEntity, tuple], name: str, index: int, expected_type: str, provided_type: str) -> None:
         logger.error(
             (
                 "For instance:\n    %s\n    %s\n"
--- a/src/ifcopenshell-python/ifcopenshell/validate.py
+++ b/src/ifcopenshell-python/ifcopenshell/validate.py
@@ -43,6 +43,7 @@ from typing import Union, Iterator, Any, Optional
 from logging import Logger, Handler
 
 import ifcopenshell
+import ifcopenshell.simple_spf
 import ifcopenshell.ifcopenshell_wrapper
 import ifcopenshell.ifcopenshell_wrapper as W
 import ifcopenshell.express.rule_executor
@@ -564,8 +565,11 @@ def validate_guid(guid: str) -> Union[str, None]:
     return None
 
 
-def validate_ifc_header(f: ifcopenshell.file, logger: Logger) -> None:
-    header: W.IfcSpfHeader = f.wrapped_data.header
+def validate_ifc_header(f: Union[ifcopenshell.file, ifcopenshell.simple_spf.file], logger: Logger) -> None:
+    if type(f) is ifcopenshell.file:
+        header: W.IfcSpfHeader = f.wrapped_data.header
+    else:
+        header: types.SimpleNamespace = f.header
     AGGREGATE_TYPE = "LIST [ 1 : ? ] OF STRING (256)"
     STRING_TYPE = "STRING (256)"
 
--- a/src/ifcopenshell-python/ifcopenshell/validate.py
+++ b/src/ifcopenshell-python/ifcopenshell/validate.py
@@ -566,29 +566,21 @@ def validate_guid(guid: str) -> Union[str, None]:
 
 def to_string_header_entity(header_entity):
     """Recreate IFC header string representation, like FILE_NAME(...)"""
-    def format_value(val):
-        if isinstance(val, str):
-            return f"'{val}'"
-        elif isinstance(val, tuple):
-            return "(" + ",".join(format_value(v) for v in val) + ")"
-        return str(val)
-
+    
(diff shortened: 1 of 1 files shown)


src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -563,9 +564,31 @@ def validate_guid(guid: str) -> Union[str, None]:
...
     return None
 
+def to_string_header_entity(header_entity):
+    """Recreate IFC header string representation, like FILE_NAME(...)"""
+    def format_value(val):
+        if isinstance(val, str):
+            return f"'{val}'"
+        elif isinstance(val, tuple):
+            return "(" + ",".join(format_value(v) for v in val) + ")"
+        return str(val)
```
aothms: I think this is almost equivalent to calling `repr(val)`. The only difference is how this would deal with the encoding of quotes and non-ascii characters.

```
>>> print(repr(""))
''
>>> print(repr("\""))
'"'
>>> print(repr("'"))
"'"
>>> print(repr("☺"))
'☺'
>>> print(repr("\"'"))
'"\''
```

Python switches the outermost quote type when more efficient. In SPF `"` denotes a sequence of binary digits.
Otherwise escapes with `\'`. In SPF escaping of quotes should by double single quote `''`

My point is. Since (a) this is only for informational purposes (b) your format_value anyway doesn't match the spec precisely, shouldn't we just use `repr()` instead?

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -563,9 +564,31 @@ def validate_guid(guid: str) -> Union[str, None]:
...
+    # Prefer native .toString() if available (native IfcOpenShell wrapper)
+    if hasattr(header_entity, 'toString'):
+        return header_entity.toString()
+
+    if hasattr(header_entity, '_fields'):
+        values = [format_value(getattr(header_entity, f)) for f in header_entity._fields]
+    else:
+        raise TypeError(f"Cannot stringify header_entity of type {type(header_entity)}")
+
+    return f"{type(header_entity).__name__.upper()}({','.join(values)})"
```
aothms: Please change to a single if - elif - else statement, where you return from the if and elif branches so that the control flow is immediately obvious.

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -563,9 +564,31 @@ def validate_guid(guid: str) -> Union[str, None]:
...
 
-def validate_ifc_header(f: ifcopenshell.file, logger: Logger) -> None:
-    header: W.IfcSpfHeader = f.wrapped_data.header
+def validate_ifc_header(f: Union[ifcopenshell.file, ifcopenshell.simple_spf.file], logger: Logger) -> None:
+    if type(f) is ifcopenshell.file:
```
aothms: isinstance() please

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -563,9 +564,31 @@ def validate_guid(guid: str) -> Union[str, None]:
...
+        header: W.IfcSpfHeader = f.wrapped_data.header
+    else:
+        header: types.SimpleNamespace = f.header
```
aothms:
```suggestion
    header: Union[W.IfcSpfHeader, types.SimpleNamespace] = f.header
```

They are fully compatible, no need for the if statement

aothms (changes requested):

src/ifcopenshell-python/ifcopenshell/validate.py
Ghesselink: > shouldn't we just use repr() instead?

Sure. As you mention, if we come this far it means the header field is correct and we're not using the reconstructed string anywhere else than in the console output. 

[Implemented it](https://github.com/IfcOpenShell/IfcOpenShell/pull/6650/commits/cb093001529a8d4d087d696ee16f0d0bdf70a1e7)

src/ifcopenshell-python/ifcopenshell/validate.py
Ghesselink: [Done](https://github.com/IfcOpenShell/IfcOpenShell/pull/6650/commits/cb093001529a8d4d087d696ee16f0d0bdf70a1e7)

src/ifcopenshell-python/ifcopenshell/validate.py
Ghesselink: No need anymore because of comment below

src/ifcopenshell-python/ifcopenshell/validate.py
Ghesselink: [Done](https://github.com/IfcOpenShell/IfcOpenShell/pull/6650/commits/cb093001529a8d4d087d696ee16f0d0bdf70a1e7)

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -563,9 +564,23 @@ def validate_guid(guid: str) -> Union[str, None]:
         return "Couldn't decompress guid, it's not base64 encoded."
     return None
 
+def to_string_header_entity(header_entity):
+    """Recreate IFC header string representation, like FILE_NAME(...)"""
+    
+    # Prefer native .toString() if available (native IfcOpenShell wrapper)
+    if hasattr(header_entity, 'toString'):
+        return header_entity.toString()
+
```
aothms:
```suggestion
```

First branch 2 times.

Also please change to isinstance() forgot to mention that.

src/ifcopenshell-python/ifcopenshell/validate.py
```diff
@@ -568,9 +568,9 @@ def to_string_header_entity(header_entity):
     """Recreate IFC header string representation, like FILE_NAME(...)"""
     
     # Prefer native .toString() if available (native IfcOpenShell wrapper)
-    if hasattr(header_entity, 'toString'):
+    if isinstance(header_entity, W.HeaderEntity):
         return header_entity.toString()
-    elif hasattr(header_entity, '_fields'):
+    elif isinstance(header_entity, tuple):
```
aothms:
```suggestion
    elif hasattr(header_entity, '_fields'):
    # namedtuple
```

Ok,. now I remember, isinstance doesn't work well with namedtuple, because it does some sort of wonky code generation. Sorry, then let's keep that line as is, because just tuple is even more misleading.

src/ifcopenshell-python/ifcopenshell/validate.py
Ghesselink: [Done](https://github.com/IfcOpenShell/IfcOpenShell/pull/6650/commits/6d2885ef2f7a05b7b17102233f77be4ce87a2453)

simple_spf.file doesn't have a distinctive type for header entity (`tuple` comes the closest); this
```
    if isinstance(header_entity, W.HeaderEntity):
        return header_entity.toString()
    elif isinstance(header_entity, tuple):
```
should catch it. 

Also modified type hinting in the function arguments. I'm still not really convinced about doing that, as it's not enforced anywhere. I saw this is possible at runtime with adding a [typeguard.typechecked](https://typeguard.readthedocs.io/en/stable/api.html#typeguard.typechecked) decorator, but that just results in extra overhead.

src/ifcopenshell-python/ifcopenshell/validate.py
Ghesselink: Oops, I'm just reading this. See comment above 
https://github.com/IfcOpenShell/IfcOpenShell/pull/6650#discussion_r2073853817

> simple_spf.file doesn't have a distinctive type for header entity (tuple comes the closest); this

What about 
```
if isinstance(header_entity, W.HeaderEntity)
   ... 
elif hasattr(header_entity, '_fields'):
   .. 
else
 ..
```
?

aothms: I agree good compromise

Ghesselink: [Implemented](https://github.com/IfcOpenShell/IfcOpenShell/pull/6650/commits/50e348507afcbcce4895745ab987863b617f832f)

PR #6656, opened 2025-05-05
call update_thumbnail_for_element when the parametric door/window is finished for an elementType

falken10vdl:
This is a follow up of Added entry in type menu to call for a thumbnail refresh#6517 (https://github.com/IfcOpenShell/IfcOpenShell/pull/6517/).
It just call for the thumbnail update after finishing editing Door/window wiht the parametric tool.
Thanks!


Andrej730:
Looks good, thanks for the contribution.

In theory we should invalidate existing previews instead of updating them right away, so they will be updated later, when user will actually need them, but that's another subject.

PR #6657, opened 2025-05-05
Skip non MESH and CURVE object listing elements for coloring

### LINKED ISSUE #5041: Colour by property - a way to opt out the spacial elements?
Hey,

What do you guys think about having a way to sort out the spatial elements? Most of the time I am using this to check the physical elements and it's annoying to have to remember to not consider the spatial ones.

Would there be any elegant way, simple to implement that could help with this?

image

falken10vdl:
So non MESH or CURVE objects are not listed in the Colour by property.


Andrej730:
Ahh, github didn't recognized the merge. Anyway, merged as 4ce2f81.

PR #6692, opened 2025-05-11
use rattler-build to speed up daily conda builds and fix deps

Krande:
Also added missing dep: typing_extensions.

### DIFF
--- a/.github/workflows/ci-ifcopenshell-conda-daily.yml
+++ b/.github/workflows/ci-ifcopenshell-conda-daily.yml
@@ -39,7 +39,7 @@ jobs:
 
       
   test:
-    name: ${{ matrix.platform.distver }}-${{ matrix.pyver.name }}-${{ matrix.variant }}
+    name: ${{ matrix.platform.distver }}-${{ matrix.pyver.name }}
     needs: activate
     runs-on: ${{ matrix.platform.distver }}
     defaults:
@@ -48,9 +48,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        variant: [novtk, all]
         pyver: [
-#          { name: py311, distver: '3.11'},
           { name: py312, distver: '3.12'}
         ]
         platform: [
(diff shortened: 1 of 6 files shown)


zbynekwinkler:
Would it be possible to cut a new release for `conda-forge` including the fix of the missing dependency `typing_extension`? 

Related issues in freecad which uses `ifcopenshell` from `conda-forge` to build its AppImage:
- https://github.com/FreeCAD/FreeCAD/issues/18011 : this is probably the oldest one but it has many duplicates as it was not easy to pinpoint where it comes from
- https://github.com/FreeCAD/FreeCAD/issues/21422
- https://github.com/FreeCAD/FreeCAD/issues/19051
- https://github.com/FreeCAD/FreeCAD/issues/18559
- https://github.com/FreeCAD/FreeCAD/issues/18300

Krande:
@zbynekwinkler. Sure, I started a new PR in https://github.com/conda-forge/ifcopenshell-feedstock/pull/71. Let me know if there are any other issues you guys have encountered besides the missing typing_extension dependency

Krande:
@aothms Looks like our anaconda token has expired. The new daily conda builds fail with

`Error:  ('Authentication token has expired', 401)`

I don't have access to ifcopenshell repo GA secrets. Can you update the token?

aothms:
updated please have a look

zbynekwinkler:
I see the new version at https://anaconda.org/ifcopenshell/ifcopenshell however freecad pulls in https://anaconda.org/conda-forge/ifcopenshell which has not been updated. I've managed to test-add this other channel, but could not build freecad with it since freecad builds against python 3.11 but the published packages are against python 3.12 only.

Any chance to update the conda-forge channel as well?

```
Cannot solve the request because of: freecad * cannot be installed because there are no viable options:
      └─ freecad 1.1.0dev would require
         └─ ifcopenshell >=0.8.3alpha250601, which cannot be installed because there are no viable options:
            └─ ifcopenshell 0.8.3alpha250601 would require
               └─ python_abi 3.12.* *_cp312, for which no candidates were found.
```

Krande:
@zbynekwinkler. I have restarted some of the failing conda-forge jobs in https://github.com/conda-forge/ifcopenshell-feedstock/runs/43299495324 (there is a bug that causes intermittent ci job failures in conda-forge) . Restarting it a few times should force the release of all the remaining 0.8.2 packages with the typing_extension dependency. 

The ifcopenshell conda-forge package follows the release schedule of bonsai and the rest of the ifcopenshell packages. So once a new tag is created on ifcopenshell for ifcopenshell-python it will trigger a new release. I don't think releasing one of the daily builds on conda-forge is the way to go. 

We could change to a 3.11 daily conda release if that would help out. I don't think there are that many that relies on python 3.12 of the conda daily builds of ifcopenshell. If there are we can make releases for 3.11 and 3.12

Maybe @aothms or @Moult knows when there will be a v0.8.3 release?

aothms:
> Maybe @aothms or @Moult knows when there will be a v0.8.3 release?

For me nothing prevents a releases. The binary builds are almost up and running again.

PR #6698, opened 2025-05-13
Derived dimensions: Show colored dimensions

falken10vdl:
Adds a checkbox to "Derived Dimensions" subpanel that allows to toggle showing the local gizmo and a bounding box with the edges in the same color as the gizmo axis.
image

vs
image

Thanks!


Andrej730:
I don't know if this feature is generally needed, need a second opinion on this. Can you please take a look? @theoryshaw @Moult 

For me it seems showing gizmo + showing object bounds can do the same.
image

theoryshaw:
i guess i don't think it hurts. 

Could you put the measurements in the viewport? 
If so, could you use 
image
to format distances?

image

theoryshaw:
related convo: https://community.osarch.org/discussion/2913/ifc-wall-length

maxfb87:
+1 for measurement in viewport

falken10vdl:
Sure. Let me try to add them in the viewport.
Thanks!

falken10vdl:
Please take a look to this.

image


I do not know if the colors are needed/required (color blinded people would not benefit) if the dimensions are shown in the viewport. My initial idea with the colors was that since the dimensions were already in the right "Derived Coordinates" subpanel, the color could be used as a legend. Moreover, for new comers, showing the local gizmo also would help to understand that those dimensions are local and that such reference system could very well be in different orientation to the global one.

So what do you think would make most sense?
Cheers!

Andrej730:
Can you please show an example with white text color? Maybe colored dimension lines and axis name (e.g. "X") would be enough to recognize the described axis and white colored text would have a better chance to stand out in general?

And why there is a second checkmark option? Isn't local gizmo already included in the first checkmark? It's better to keep amount of available options at minimum, so it will be easier to use.

falken10vdl:
Definitely much better. I have collapsed back to a single checkbox that toggles both the colored bounding-box with the dimensions and the local gizmo.
image
Here a video showing how the dimensions try to stay closer to the center so hopefully easier to read.

https://github.com/user-attachments/assets/9390d226-420d-45d7-9b22-5cac8f800ccc


Cheers

theoryshaw:
very cool, you might be planning this anyway, but can you use the following to format the distance?
https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/bonsai/bonsai/bim/module/drawing/helper.py#L126

falken10vdl:
thanks @theoryshaw ! I didn't know about it. Much cleaner code and coherent with the rest: 

`value_str = format_distance(value, hide_units=False)
`
Also the decimal places are wisely set based on the unit

image

Thanks!

Moult:
Sorry late to the party. I think this type of bounding box measurement belongs more as part of @brunoperdigao 's measurement tools, not as a decorator. Similar to how we have a one-click area measurement, a one-click that measures bounding box seems more valuable (especially when you can select multiple objects).

falken10vdl:
Sure, I can move it wherever you guys think is best. Where is that located :) ? In the explore tool?
I have also updated as following from your comments:
- If several objects are selected then a bounding box covering all of them is generated (using global coordinates) and the gizmo is switched to Global coordinates
- If a single object is selected the gizmo is switched to Local coordinates (to be coherent with XYZ dimensions)
- I have changed the coloring/annotation. I have found that selecting the trihedron from the bounding box corner that it is lowest in the screen (2d lowest Y) is probably best for viewing. Please let me know if you think otherwise
See an example:

https://github.com/user-attachments/assets/5fe2fb0b-d75f-4178-a8ca-9724ce0cae0f

Thanks!

brunoperdigao:
Hi @falken10vdl, you can check how face area works here: https://github.com/IfcOpenShell/IfcOpenShell/blob/5c72cf35bd03efd04ed6c41ca1557e61af3d8f98/src/bonsai/bonsai/bim/module/project/operator.py#L2675

Then you can take a look here to see how to add another type of measurement: https://github.com/IfcOpenShell/IfcOpenShell/blob/5c72cf35bd03efd04ed6c41ca1557e61af3d8f98/src/bonsai/bonsai/bim/module/project/workspace.py#L105

Let me know if you need any help.

falken10vdl:
Perfect. Thanks @brunoperdigao !

falken10vdl:
I have now moved all to the Explore Tool. Please take a look.

https://github.com/user-attachments/assets/264e7631-1934-4e3d-a742-4fc8818b370d


Thanks!

falken10vdl:
I have copied the draw_text_background and the draw_batch to have a similar style to the polyline. Maybe it would be worth to have those methods static somewhere?
image

Thanks!

theoryshaw:
possible to push your commits, so we can play around with it? :)

falken10vdl:
@theoryshaw I think I do not follow you :) not very good using github... What do you mean by pushing the commits? I can see them up there:
image

Thanks!

theoryshaw:
My apologies!... i was not pulling correctly. 
carry on. :)

brunoperdigao:
@falken10vdl A few comments:
1 - Please take a look at how measure face works. First you have to select the type of measurement, then the tool only works when you click the Measure Tool button. In your code, the tool is on in the moment you select the measurement type.
2 - The measurement tools work with the PolylineOperator, so you might want to take a look at the code there too. But I think you can make it work by using a code very similar o MeasureFaceAreaTool.
3 - Then you'll see that all the code for drawing on screen are inside `model/decorator.py`. Take a look at `PolylineDecorator` and `FaceAreaDecorator` and see how they are called inside the operator by using ".install()" and ".uninstall()".

Please ping me if you have any doubts.

falken10vdl:
@brunoperdigao First of all, thanks a lot for your comments above! 
Some questions from my side (I am quite new to blender and Bonsai programming...)
1- Yes I have seen it. I did this on purpose. The Measurement tool is active within the scope of the Explore tool. I was finding useful to have the bounding box measurements even if you move to other tools (namely during addition of walls/slabs/doors/windows. 
2, 3- Yes. Let me take a look to it

I have seen that MeasureFaceAreaTool is using modal operation. As you have seen in the code I proposed I am using depsgraph_update_post. **What is your view on pro/con of using either one?**. Since I wanted to have this bounding box generated regardless of being in the Explore Tool or not is it mandatory to use depsgraph_update_post? The modal can work outside of the Explore Tool space? I typically assign a shortcut to the operator so I can switch it on/off even if I am not in the Explore Tool.
image

Thanks!

falken10vdl:
Hi @brunoperdigao  I have redone it to split into the operator and decorator parts. Please take a look and let me know what you think.
Thanks!

brunoperdigao:
@falken10vdl Thanks, I'll try to take another look by the end of the week.
In response to your previous comment, I believe it’s a matter of keeping the consistency between tools. If you want to use it beyond the explore tool, I think it could work better as Viewport Overlay->Bonsai Decorators rather than a Measure Tool type, which aligns more with what you described.

I’d love to hear what others think about this as well!

image

falken10vdl:
@brunoperdigao I think your proposal makes more sense. Thanks!

falken10vdl:
image

theoryshaw:
Nice, makes sense to me.

brunoperdigao:
@falken10vdl Hi, everything seems good. My last comment would be to put all the imports in the beginning of the file. Also, add your decorator to this part:
https://github.com/IfcOpenShell/IfcOpenShell/blob/a531bd8c1f4e24004e5592d69d5b6e8071a9cea5/src/bonsai/bonsai/bim/handler.py#L348C3-L367C52

falken10vdl:
@brunoperdigao I moved the imports to the top (much nicer and simpler, thanks!) and also added install to the handler.py as you pointed out. Thanks a lot for your help!

brunoperdigao:
@Andrej730 @Moult Everything seems good now, can we merge this?

Andrej730:
Looks good, let's merge it.

falken10vdl:
After using this decorator some days I have seen that sometimes the labels get a little bit overlapped so I have added some algorithm to space it. Now they should be clearly visible even for small dimensions.
Cheers!

brunoperdigao:
Hi @falken10vdl, could you rebase and fix the conflicts?

falken10vdl:
@brunoperdigao Doing rebase successfully is still in my list of things to learn :)
Whenever I try it either in ui or terminal in vscode I endup messing up everything and git asking me to do a massive push of commits I have not done... there must be something on how I have my local fork setup I guess...
Anyhow, in order to try not to ruin it I have created a new PR based on a refreshed fork of the upstream.
See it here: https://github.com/IfcOpenShell/IfcOpenShell/pull/6853
btw I have added an extra commit as explained in the PR.
If you agree we can close this PR and merge from https://github.com/IfcOpenShell/IfcOpenShell/pull/6853.
Cheers!

Issue #5857, opened 2024-12-09
Bonsai | UI Feature request:  promote Grouping and Filtering to main menu


steverugi:
As previously mentioned on OSArch, it would be extremely convenient to have the `Grouping and Filtering` panel accessible directly from the main menu. Something like:

![image](https://github.com/user-attachments/assets/319f9823-80b0-4ae1-9e4a-b4da5e0e8a82)

In my opinion it is an essential feature, currently available at the very bottom, requiring frequent scrolling to access.

I hope this meets the preferences of both users and developers.
Thank you in advance

sboddy:
https://community.osarch.org/discussion/2649/bonsai-ui-feature-request-promote-grouping-and-filtering-to-main-menu#latest

Just linking some forum feedback on this proposal.

Netrieb:
I would even suggest to place it in to the `Outliner area` because Filtering, Searching and the Outliner are very similar things, wouldn't it make more sense? When you do your search, you have immediate visual feedback in the Outliner, you could have the Search tab still visible etc.

![Image](https://github.com/user-attachments/assets/926df695-fa53-49f7-8436-5d575ab7fcd8)

Moult:
Although it is possible for us to inject a new button at the top of the
outliner which will trigger a popup panel or similar to show grouping
and filtering there, it's probably not a good idea because it reduces
trust in the user on "where to look for features".

In the past we had lots of features spread about everywhere in the
Blender UI and so it wasn't clear where a feature might be. I'd say a
lot of users never discovered some of the features we had hidden in the
material node editor. Since we introduced our own tabbed UI, that made
it clear that everything in these tabs is everything we have to offer.

I think it makes sense to put these at hte top of the Quality and
Coordination tab. I've assigned this task to @blenderdefender

sboddy:
Quality and Coordination? 🤔 I still consider myself a newbie, but that strikes me as an illogical place to put Grouping and Search. I thought my suggestion in the linked discussion of moving it up in the Project Overview tab _above_ the "set and forget" stuff was an OK compromise.

Moult:
I think you're right, so just below the Spatial panel, still in Project information?

steverugi:
glad to see some action about my proposal ;)
the initial idea for me was not to have a popup window but rather a dedicated panel among the existing ones

Totally agree with @sboddy , Quality and Coordination doesn't seem the _most obvious drawer where to find your socks_

[...]

PR #6738, opened 2025-05-22
Grouping and Filtering

talukderpushpak2000:
In the Scene tab, I have repositioned the grouping and filtering icons by moving them upwards. This adjustment improves the user experience by making these frequently used actions more accessible and visible, thereby streamlining the workflow when organizing or filtering.


OrionSehn:
@talukderpushpak2000 

_(you can link pull requests to issues like this):_

This is relating to issue #5857 to try and move up the panel to make it more visible

PR #6740, opened 2025-05-22
fix to default directory on dev_environment.bat 

OrionSehn:
Default directory for the bonsai path assumed you had a fixed build, and didn't use the which is contradictory.

This fixes that, and updates the docs to make it clear to run the setup .bat script when blender is closed. Because otherwise it will deny permission to some dependency files and break your bonsai installation.


Andrej730:
Thank you for the contribution.

Issue #6684, opened 2025-05-09
Type Inference Issue with ifcopenshell.file in mypy


Martin15135215:
### Bug Description

I'm encountering type inference issues when using `mypy` with `ifcopenshell.file`. Specifically, `mypy` reports errors indicating that attributes such as `file_description` and `file_name` do not exist on objects of type `ifcopenshell.file`.

**Error Details**:

```plaintext
$ poetry run mypy 
<redacted>:9: error: Argument "schema" to "file" has incompatible type "str"; expected "Literal['IFC2X3', 'IFC4', 'IFC4X3'] | None"  [arg-type]
<redacted>:10: error: Item "function" of "Any | Callable[..., entity_instance]"  has no attribute "file_description" [union-attr]
<redacted>:11: error: Item "function" of "Any | Callable[..., entity_instance]"  has no attribute "file_description" [union-attr]
<redacted>:13: error: Item "function" of "Any | Callable[..., entity_instance]"  has no attribute "file_name" [union-attr]
<redacted>:14: error: Item "function" of "Any | Callable[..., entity_instance]"  has no attribute "file_name" [union-attr]
<redacted>:16: error: Item "function" of "Any | Callable[..., entity_instance]"  has no attribute "file_name" [union-attr]
<redacted>:17: error: Item "function" of "Any | Callable[..., entity_instance]"  has no attribute "file_name" [union-attr]
<redacted>:19: error: Item "function" of "Any | Callable[..., entity_instance]"  has no attribute "file_name" [union-attr]
<redacted>:20: error: Item "function" of "Any | Callable[..., entity_instance]"  has no attribute "file_name" [union-attr]
<redacted>:21: error: Item "function" of "Any | Callable[..., entity_instance]"  has no attribute "file_name" [union-attr]
Found 10 errors in 1 file 
```

**Code Snippet**:

```python
def create_ifc_file() -> ifcopenshell.file:
    file: ifcopenshell.file = ifcopenshell.file(schema="IFC4X3_ADD2")

    file.header.file_description.description = ("",)
    file.header.file_description.implementation_level = ""

    file.header.file_name.name = ""
    file.header.file_name.time_stamp = ""
    file.header.file_name.author = ("",)
    file.header.file_name.organization = ("",)

    file.header.file_name.preprocessor_version = ""
    file.header.file_name.originating_system = ""
    file.header.file_name.authorization = ""
    
    return file
```

**Request**:

[...]

PR #6744, opened 2025-05-23
Create ifcopenshell_wrapper.pyi

Martin15135215:
minimal stub file for swig-generated ifcopenshell_wrapper python file

https://github.com/IfcOpenShell/IfcOpenShell/issues/6684#issuecomment-2904402618

Attention: When one provides a stub file (a .pyi file) for a module or a part of a module, that stub file completely replaces the type information for that module for type checking purposes. Mypy will only use the definitions in the stub file and ignore the inline type annotations (if any) in the original .py file. See https://github.com/python/mypy/issues/7225


Andrej730:
Thank you!

PR #6752, opened 2025-05-26
Fixes: Bat script failed even when you right-click on it, and run as administrator.

theoryshaw:
Fixes via AI sleuthing:
https://chatgpt.com/share/68349a2a-bec8-8013-9e18-1667d2cdda67

ping @falken10vdl, @Andrej730

### DIFF
--- a/src/bonsai/docs/quickstart/ide/windows/dev_environment.bat
+++ b/src/bonsai/docs/quickstart/ide/windows/dev_environment.bat
@@ -106,6 +106,7 @@ echo Manually downloading some third party dependencies...
 curl https://raw.githubusercontent.com/jsGanttImproved/jsgantt-improved/master/dist/jsgantt.js -o "%PACKAGE_PATH%\bonsai\bim\data\gantt\jsgantt.js"
 curl https://raw.githubusercontent.com/jsGanttImproved/jsgantt-improved/master/dist/jsgantt.css -o "%PACKAGE_PATH%\bonsai\bim\data\gantt\jsgantt.css"
 curl -L https://github.com/BrickSchema/Brick/releases/download/nightly/Brick.ttl -o "%PACKAGE_PATH%\bonsai\bim\data\brick\Brick.ttl"
+curl https://cdn.jsdelivr.net/npm/jquery@3.7.1/dist/jquery.min.js -o "%PACKAGE_PATH%\bonsai\bim\data\webui\static\js\jquery.min.js"
 
 popd
 pause
--- a/src/bonsai/docs/quickstart/ide/windows/dev_environment.bat
+++ b/src/bonsai/docs/quickstart/ide/windows/dev_environment.bat
@@ -2,13 +2,19 @@
 
 setlocal enabledelayedexpansion
 
+:: Ensure script runs from its own directory
+cd /d "%~dp0"
+:: Use pushd/popd in case we need to return later
+pushd .
 
-rem ADAPT setting REPO_PATH to use custom path for IfcOpenShell repository.
-rem Otherwise by default it is assumed script is executed from IfcOpenShell directory.
+:: ADAPT setting REPO_PATH to use custom path for IfcOpenShell repository.
+:: Otherwise by default it is assumed script is executed from IfcOpenShell directory.
 SET REPO_PATH=%HOMEDRIVE%\Users\%USERNAME%\Documents\bonsaiDevel\IfcOpenShell
 SET BLENDER_PATH=%HOMEDRIVE%\Users\%USERNAME%\AppData\Roaming\Blender Foundation\Blender\4.2
 SET PACKAGE_PATH=%BLENDER_PATH%\extensions\.local\lib\python3.11\site-packages
 SET BONSAI_PATH=%BLENDER_PATH%\extensions\raw_githubusercontent_com\bonsai
+:: Note: If you manually installed Bonsai, the path will be the following instead...
+:: SET BONSAI_PATH=%BLENDER_PATH%\extensions\user_default\bonsai
 
 echo SETUP BONSAI ADD-ON LIVE DEVELOPMENT ENVIRONMENT
 echo Update REPO_PATH, BLENDER_PATH, PACKAGE_PATH, BONSAI_PATH in the script above.
@@ -20,7 +26,7 @@ if not defined REPO_PATH (
     echo REPO_PATH is not set, assuming we're already in IfcOpenShell directory.
     set REPO_PATH=%cd%
 )
-rem Print summary of paths
+:: Print summary of paths
 echo.
 echo Please review if the following is right:
 echo PWD: %CD%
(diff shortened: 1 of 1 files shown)


theoryshaw:
@falken10vdl 
dumb question, what does the following do? 
https://github.com/IfcOpenShell/IfcOpenShell/blob/02af46c99f94df889ffa857eefefeb137b1e8568/src/bonsai/docs/quickstart/ide/windows/dev_environment.bat#L39-L57

falken10vdl:
@theoryshaw No dumb questios at all! I changed a little bit the script that was in the doc: http://docs-unstable.bonsaibim.org/guides/development/installation.html#live-development-environment  I added that part to ignore the .so files so git will not attempt to track them. As part of the script there is copy command of some libraries: copy "%PACKAGE_PATH%\ifcopenshell\*_wrapper*" "%CD%\src\ifcopenshell-python\ifcopenshell\" 
So that will update the local Github repository and that will be seen by git as a change, so when you do a commit it will appear as a file to commit. Moreover, that .so libraries are quite big, so in my case that was not allowing to do a commit as it is a big file and that would ask for git lfs.
Does it make sense? Is the script not working for you? (I am 99% of the time in linux so I might not have seen that error)
Cheers!

falken10vdl:
I have just run as administrator the attached .bat file in a fresh bonsai installation:
imagen
[dev_environment.bat.txt](https://github.com/user-attachments/files/20445556/dev_environment.bat.txt)
Once the script is executed once, susequent executions provide an error but are harmless (the libraries already copied and the links created):
imagen

What error are you getting?

falken10vdl:
I just also tried latest bonsai with bleneder 4.4.3:
imagen

I run it by opening a windows explorer and right clicking in the file and doing "run as Administrator"
Cheers!

falken10vdl:
I have taken a look to https://chatgpt.com/share/68349a2a-bec8-8013-9e18-1667d2cdda67

I see: 
REPO_PATH is not set, assuming we're already in IfcOpenShell directory.
REPO_PATH=C:\Windows\system32

This must be populated with the GITHUB path ending in \IfcOpenshell

In my case for example:
 REPO PATH (.....\IfcOpenshell): C:\Users\falke\Documents\bonsaiDevel\IfcOpenShell

I added this (....\IfcOpenshell) as an aid to give the idea of how they shoudl look like, but I am not sure if I have created more confusion than anything...

Regards,

theoryshaw:
my path is `D:\Dropbox\GitHub\IfcOpenShell` maybe that's it. 

When i right clicked on the bat file and run with admin privileges, I get the first image, after adding the following snippet, I get the 2nd image (that is, it worked)

```
:: Ensure script runs from its own directory
cd /d "%~dp0"
:: Use pushd/popd in case we need to return later
pushd .
popd
```

falken10vdl:
I see that you have not set the REPO directory to D:\Dropbox\GitHub\IfcOpenShell  in the .bat file.
This line: SET REPO_PATH=....is either commented or does not have any information. In your case should set it something like this:
SET REPO_PATH= D:\Dropbox\GitHub\IfcOpenShell

Anyhow I think your update to the script is a nice thing to have since it will cd to the directory where the bat file is, and if that one happens to be in the root of the REPO directory, it will work even if the REPO is no set.
Nice!

Andrej730:
> Bat script failed even when you right-click on it, and run as administrator

This batch script was not intended to be executed just by itself from the explorer (at least if it's not modified), it was intended to be executed from cmd from IfcOpenShell directory.
But I agree it is convenient and sounds like a good default - `dev_environment.bat` to find root repo directory and use it for the dev environment. It wouldn't even interrupt a current workflow of running batch from repo root (at least in case if you have just 1 IfcOpenShell repo).

There's also confusion between two batch scripts. In ChatGPT logs it seems you were using `src\bonsai\scripts\installation\dev_environment.bat` (it starts with `REPO_PATH=%HOMEDRIVE%\Users\%USERNAME%\Where\Your\Git\Repository\Is\Cloned\IfcOpenShell`), but PR is changing `src/bonsai/docs/quickstart/ide/windows/dev_environment.bat`...


There is also a bug in `quickstart/ide/windows/dev_environment.bat` - `REPO_PATH` seems to be always set, so `if not defined REPO_PATH` will never get triggered and comment `:: Otherwise by default it is assumed script is executed from IfcOpenShell directory.` is incorrect.
image



I think we should reconcile the scripts first to prevent further confusion.



----

Regarding the code change:

```bat
:: Ensure script runs from its own directory
cd /d "%~dp0"
:: Use pushd/popd in case we need to return later
pushd .
```

I think it's incorrect. E.g. if you create a script like so
```bat
@echo off
cd /d "%~dp0"
pushd .
echo PWD: %CD%
popd
pause
```
and run it, you'll end up in .bat directory, not in the original folder.
```
C:\Users\Andrej>L:\oblivion\test.bat
PWD: L:\test
Press any key to continue . . .
L:\test>
```

`popd` just undoes `pushd`, but doesn't undo `cd /d "%~dp0"`.

The working script would be something like:

```bat
@echo off
pushd "%~dp0"
echo PWD: %CD%
popd
pause
```

But it seems just having `setlocal` does the same, so there's no need for `pushd` and `popd` at all.
```
@echo off
setlocal
cd /d "%~dp0"
echo PWD: %CD%
pause
```

```
C:\Users\Andrej>L:\oblivion\test.bat
PWD: L:\test
Press any key to continue . . .
C:\Users\Andrej>
```


I had an alternative idea for a while now, which would allow us not to think about this kind of system language details we don't normally care about, see https://github.com/IfcOpenShell/IfcOpenShell/issues/6754

----

> I added that part to ignore the .so files so git will not attempt to track them. As part of the script there is copy command of some libraries: copy "%PACKAGE_PATH%\ifcopenshell*_wrapper*" "%CD%\src\ifcopenshell-python\ifcopenshell"
> So that will update the local Github repository and that will be seen by git as a change, so when you do a commit it will appear as a file to commit. Moreover, that .so libraries are quite big, so in my case that was not allowing to do a commit as it is a big file and that would ask for git lfs.
> Does it make sense? Is the script not working for you? (I am 99% of the time in linux so I might not have seen that error)
> Cheers!

@falken10vdl 
I think we should address the root cause in the case like this - .so files should be ignored by .gitignore, no need to patch it just for the dev environment.
And I think this patch might be outdated at this point and the issue was already resolved after - https://github.com/IfcOpenShell/IfcOpenShell/commit/614b71e0c68d226ca066d6954ef937c5ccff0907

falken10vdl:
@Andrej730 Absolutely! If this is now already solved, better to remove the gitignore part on this .bat file and also in its linux sibling dev_environment.sh
Thanks!

theoryshaw:
superseded by: https://github.com/IfcOpenShell/IfcOpenShell/blob/d49a3486b9f5d29029b5c57408b7eeb29036e389/src/bonsai/scripts/installation/dev_environment.py

PR #6763, opened 2025-05-29
 dms2dd and dd2dms functions to use microseconds and handle negative degrees

falken10vdl:
Based on https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/schema/ifcproductextension/lexical/ifcsite.htm

I think the functions need to be updated to handle microsencods instead of miliseconds for refLatitude and refLongitude

Also negative values affect the degree number only.

### DIFF
--- a/src/ifcbimtester/bimtester/features/steps/geolocation/en.py
+++ b/src/ifcbimtester/bimtester/features/steps/geolocation/en.py
@@ -234,7 +234,7 @@ def step_impl(context, guid, number):
     site = util.assert_guid(IfcStore.file, guid)
     util.assert_type(site, "IfcSite")
     ref = util.assert_attribute(site, "RefLongitude")
-    number = ifcopenshell.util.geolocation.dd2dms(number, use_ms=(len(ref) == 4))
+    number = ifcopenshell.util.geolocation.dd2dms(number, use_us=(len(ref) == 4))
     util.assert_attribute(site, "RefLongitude", number)
 
 
@@ -244,7 +244,7 @@ def step_impl(context, guid, number):
     site = util.assert_guid(IfcStore.file, guid)
     util.assert_type(site, "IfcSite")
     ref = util.assert_attribute(site, "RefLatitude")
-    number = ifcopenshell.util.geolocation.dd2dms(number, use_ms=(len(ref) == 4))
+    number = ifcopenshell.util.geolocation.dd2dms(number, use_us=(len(ref) == 4))
     util.assert_attribute(site, "RefLatitude", number)
 
 
--- a/src/ifcopenshell-python/ifcopenshell/util/geolocation.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/geolocation.py
@@ -42,6 +42,7 @@ class HelmertTransformation(NamedTuple):
 
 def dms2dd(degrees: int, minutes: int, seconds: int, us: int = 0) -> float:
     """Convert degrees, minutes, and (micro)seconds to decimal degrees
+    All components must be either positive or negative.
 
     :param degrees: The degrees component
     :param minutes: The minutes component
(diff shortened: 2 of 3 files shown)


falken10vdl:
@Andrej730 @theoryshaw Do you agree with this? To me it looks that should be fixed?
Cheers!

Andrej730:
@falken10vdl 

I think you're correct about the typo in doc-strings, but incorrect about sign calculation.

See Signedness in
https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/schema/ifcmeasureresource/lexical/ifccompoundplaneanglemeasure.htm, in IFC all 4 components will have the same sign. I guess it's worth mentioning in docs that calculations are based on IfcCompoundPlaneAngleMeasure implementation.

Though I agree `dd2dms` can be refactored a bit - `sign = -1 if dd < 0 else 1` seems a tiny bit more readable (at least to me), `dd < 0` thing seems redundant as `dd` already has absolute value already, and we also could make a use of `seconds` calculated in `use_ms` branch.
If you want, you can help improve it and add tests to concrete expected behaviour.

falken10vdl:
@Andrej730 cool. Thanks! Yes l was aware that negative value was needed to be handled because in conversation https://community.osarch.org/discussion/comment/25363#Comment_25363 the conversion was not right.
Then yes let me change according to your comments!
Quite interesting that the sign must be put in all number (which make sense) because I think that in regular lat/long the sign is only in the degrees unit (https://en.m.wikipedia.org/wiki/Decimal_degrees)

falken10vdl:
@Andrej730 I had seen that floating point numbering was producing rounding errors which I think jeopardize the accuracy if we have microseconds. I have refactored using "Decimal" arithmetic

```
=============================================================================================================== 1 passed in 0.15s ================================================================================================================
falken10vdl@monster:~/bonsaiDevel/IfcOpenShell/src/ifcopenshell-python$ pytest test/util/test_geolocation.py::TestDMS2DDandDD2DMS -v
Blender 4.4.3 (hash 802179c51ccc built 2025-04-29 15:12:13)
module changed on disk: '/home/falken10vdl/.config/blender/4.4/extensions/blender_org/sun_position/__init__.py' reloading...
============================================================================================================== test session starts ===============================================================================================================
platform linux -- Python 3.11.11, pytest-8.4.1, pluggy-1.6.0 -- /home/falken10vdl/.local/share/applications/blender-4.4.3-linux-x64/4.4/python/bin/python3.11
cachedir: .pytest_cache
rootdir: /home/falken10vdl/bonsaiDevel/IfcOpenShell/src/ifcopenshell-python
configfile: pyproject.toml
plugins: bdd-8.1.0
collected 1 item                                                                                                                                                                                                                                 

test/util/test_geolocation.py::TestDMS2DDandDD2DMS::test_dms2dd_and_dd2dms PASSED                                                                                                                                                          [100%]

=============================================================================================================== 1 passed in 0.01s ================================================================================================================
falken10vdl@monster:~/bonsaiDevel/IfcOpenShell/src/ifcopenshell-python$ 

Cheers!

```

Andrej730:
@falken10vdl 

1. black format is missing

2. `sign = -1 if degrees < 0 or minutes < 0 or seconds < 0 or us < 0 else 1` - why checking all of them if the sign must always match by dms format definition? And what's the point in general of detaching sign from calculations in both `dms2dd` and `dd2dms`?

3. `Namely all of its are either positive or negative` - "all of its components"?

4. `dms2dd` - why converting components to decimal? They will be converted automatically on any operation with `Decimal("xxx")`. Also using `Decimal(xxx)` instead of `Decimal("xxx")` would be better, since it would avoid string parsing.

5. `dd2dms` - no need for `str` before passing value to `Decimal` if value is `int`.

falken10vdl:
@Andrej730 Thanks for the comments! Sorry I was last week in summer vacation and had no access to my development computer.
I have updated the implementations. For dms2dd I have removed decimal as the return value in the end is going to be float. It would make sense if the return value was decimal itself. I have added an assert to verify that he components all have the same sign
For dd2dms I removed str when they are int as you have rightly pointed out. I have left just the first one for dd_decimal = Decimal(str(abs(dd))) since I understand that the dd initial value will come from some string (from ifc or user input), so we better recover that string and convert it to decimal rather that using hte float which would have binary bits form the float representation.
I have kept the sign in dd2dsm as i think the logic is easier to understand but jut my opinion :) 
Please let me know your feedback.
Thanks!

Andrej730:
@falken10vdl 

> I have added an assert to verify that he components all have the same sign

Let's make it more explicit then? E.g. `(x1 >= 0 and x2>= 0 ... ) or (x1 < 0 and x2 < 0 ...)`.

> For dd2dms I removed str when they are int as you have rightly pointed out. I have left just the first one for dd_decimal = Decimal(str(abs(dd))) since I understand that the dd initial value will come from some string (from ifc or user input), so we better recover that string and convert it to decimal rather that using hte float which would have binary bits form the float representation.

Makes sense, but constant expressions like `Decimal("60")` should also be written as `Decimal(60)` to avoid string parsing.

> I have kept the sign in dd2dsm as i think the logic is easier to understand but jut my opinion :)

I think detaching the sign adds confusion - it implies that without detaching the sign calculations will work somehow differently.

falken10vdl:
@Andrej730 

> Let's make it more explicit then? E.g. (x1 >= 0 and x2>= 0 ... ) or (x1 < 0 and x2 < 0 ...).

done

> Makes sense, but constant expressions like Decimal("60") should also be written as Decimal(60) to avoid string parsing.

yes, my bad. Fixed.

> I think detaching the sign adds confusion - it implies that without detaching the sign calculations will work somehow differently.

ok. Removed the sign. it is true that the operations are the same ;) I thought I had to change the logic but it works for both cases so yes, much better to avoid confusion

Cheers

Andrej730:
Looks good, thank you.

falken10vdl:
Thanks to you!

PR #6766, opened 2025-05-30
Update code_examples.rst

piotr-smolira:
Added example about iterating entire model.

### DIFF
--- a/src/ifcopenshell-python/docs/ifcopenshell-python/code_examples.rst
+++ b/src/ifcopenshell-python/docs/ifcopenshell-python/code_examples.rst
@@ -17,21 +17,22 @@ functions, check out the API Reference.
 Get all entites (STEP ID items)
 -------------------------------
 
-IFC file opened through IfcOpenShell is iterable. This means you can get all entities (all STEP ID items) that occurs in the file with a simple loop.
+
+IFC file opened through IfcOpenShell is iterable. This means you can get all instance (all STEP ID items) that occurs in the file with a simple loop.
 
 .. code-block:: python
 
-    for entity in model:
+    for inst in model:
         # It prints out entire entity as it appears in IFC file, e.g
-        print("The STEP entity is: ", entity)
+        print("The STEP entity is: ", inst)
         # You can get info about attributes of an instance as you normally would, e.g. name, type, get_info()
-        print(entity.get_info())
-        # Get entity type
-        print(entity.is_a())
-        # Get entity name. IF statement is necessary to ensure that the entity is actual instance such as walls, properties, tasks etc.
+        print(inst.get_info())
+        # Get instance type
+        print(inst.is_a())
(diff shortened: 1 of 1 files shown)


src/ifcopenshell-python/docs/ifcopenshell-python/code_examples.rst
```diff
@@ -14,6 +14,25 @@ that you have a IFC model loaded into model variable like so:
...
+    for entity in model:
+        # It prints out entire entity as it appears in IFC file, e.g
+        print("The STEP entity is: ", entity)
+        # You can get info about attributes of an instance as you normally would, e.g. name, type, get_info()
+        print(entity.get_info())
+        # Get entity type
+        print(entity.is_a())
+        # Get entity name. IF statement is necessary to ensure that the entity is actual instance such as walls, properties, tasks etc.
+        # For more information check submodule: ifcopenshell.entity_instance
+        if isinstance(entity, ifcopenshell.ifcopenshell_wrapper.entity_instance):
```
aothms: This is always the case. IfcOpenShell is not very consistent here though:

There are entity instances and type declarations in the schema. When you iterate over the file object you only get the first category.

`model.by_type('IfcPropertySingleValue')[0].NominalValue` <- would be an example of a type declaration instance.

Type declaration instances are normally not exposed as entity_instance objects, only when they appear as a typed attribute value in the serialization which is when the schema attribute is a SELECT type (such as property NominalValue).

To distinguish:

schema_name, entity_name = inst.is_a(True).split('.')
ifcopenshell.ifcopenshell_wrapper.schema_by_name(schema_name).declaration_by_name(entity_name).as_entity() is not None # True if entity

The other (non-select) type declarations are simply mapped to python types directly. E.g `assert type(model.by_type('IfcWall')[0].GlobalId) == str`

Whether the entity has a Name attribute depends on the schema. For example `if inst.is_a('IfcRoot'): assert hasattr(inst, 'Name')` But also several other entities have a Name attribute.

I prefer to use the variable name `inst` for an arbitrary entity instance, because strictly speaking in Express terminology the entity is the class and the entity instance is the individual.

aothms (changes requested):

src/ifcopenshell-python/docs/ifcopenshell-python/code_examples.rst
piotr-smolira: Thanks Thomas for the great insight. I think it more clear now for me.
The if statement fixed issue for IfcLocalPlacement has no attribute 'Name', but at the same time I can't say if this omitted some instances that have attribute 'Name'. So being for the safe, I would change this to: ```
if hasattr(entity, "Name"):
        print(entity.Name)
```
Also the naming convention makes a lot of sense with `inst` being for an arbitrary instance.

Issue #5155, opened 2024-08-07
adding Product as resource to a Cost Schedule | wrong quantity


steverugi:
I create a 'Doors' Resource product to identify doors in my cost schedule

![image](https://github.com/user-attachments/assets/35359240-aea5-404c-8fa5-e6dfa51c6a47)

when assigned the resource to the cost item it counts as 1 before entering any quantity or assigning it from selected elements in the model

![image](https://github.com/user-attachments/assets/7516a50e-cec7-4d4e-9f8d-fd3e59000680)

when I do it (add elements from the model) the total in the BoQ is one too many

![image](https://github.com/user-attachments/assets/18489e44-b561-4c80-a2b5-af078f535900)

the same happens if I go through a work schedule task by the way, thanks

falken10vdl:
You feel this is a bug? What would be your expectation?
Thanks!

steverugi:
@falken10vdl 
I think it should work the same way as other resources

if I have 
resource blockwork  $25/m2 and assign it to a cost item with quantity= 10m2 of wall the cost item reports $25/m2x10m2 = $250

but if a have
resource door $100/U and assign it to a cost item with quantity= 5 doors I'd expect the cost item to report $100/U x 5U = $500

at the moment it reports the number of doors + 1

please correct me if wrong, thanks

falken10vdl:
Ok. Let me take a look to see if I spot the issue in the code.
Thanks!

falken10vdl:
@steverugi is it possible to share an ifc file with this issue? That way we can work on it to try to fix it. Thanks!

steverugi:
@falken10vdl 
thanks in advance for your kind assistance

![Image](https://github.com/user-attachments/assets/9323dff3-45b9-46bf-8ff6-fed28d46d398)

[doorcount.ifc.txt](https://github.com/user-attachments/files/20480717/doorcount.ifc.txt)

falken10vdl:
@steverugi Forgive me because costing and scheduling are in my "to-do" list to understand better the logic. I have however played in the UI and seen the following:

https://github.com/user-attachments/assets/8e21a3df-e416-4a80-987b-afea9f24fb86

As you can see the quantity is following the amount of  PRODUCTS + amount of RESOURCE, that is why it equates to 4 in your case.
I see that the operator called is the same bim.assign_cost_item_quantity. 
[...]

PR #6772, opened 2025-06-02
Resource should not increment quantity in cost schedule

falken10vdl:
PR as suggested in https://github.com/IfcOpenShell/IfcOpenShell/issues/5155
Regards,


maxfb87:
@falken10vdl good. What about also update the docstring accordingly?

falken10vdl:
@maxfb87 updated: 
    If cost item has just 1 quantity and it's IfcQuantityCount, API will
    assume that quantity is used for counting controlled objects
    and it will recalculate the quantity value at the end of the API call,
    _as long as the relaedObjects are not IfcConstructionResource which do not
    count towards the cost item (they only provide value)_.

Issue #6803, opened 2025-06-12
Conan center contribution


EstebanDugueperoux2:
### Feature Description

Hello IfcOpenShell community,

I open this issue to talk about a [conan recipe contribution](https://github.com/conan-io/conan-center-index/pull/27623) to [conan center](https://github.com/conan-io/conan-center-index/).
The goal is to ease IfcOpenShell C++ libraries consumption.

But to be able to do that, we have to patch the CMakeLists.txt:
- Have options moved after project() to be changed;
- Add " /permissive- /EHsc" CMAKE_CXX_FLAGS to be able to compile on Windows with MSVC for C++20 because IfcOffsetCurveByDistance.cpp use [alternative operator](https://en.cppreference.com/w/cpp/language/operator_alternative.html), no more supported with C++20 standard;
- TO_BE_COMPLETED.

Are you opened to PR to improve conan integration?

Regards.

aothms:
> Are you opened to PR to improve conan integration?

Definitely. I don't think there are heavy conan users among the core devs here, so maybe you're a bit on your own, but it's definitely a nice addition and would also improve overall code quality (like the alternative operator, probably just a typo).

EstebanDugueperoux2:
Hi @aothms some tests (TestIds and TestRemoveRepresentation) failed randomly even on v0.8.0 main branch.
 Do you have an idea of what happens?

aothms:
> some tests (TestIds and TestRemoveRepresentation) failed randomly

Do you have an idea @Andrej730 ? Seems to be rather specific failures.

Andrej730:
1) TestRemoveRepresentation - found an overlook in my recent code, I'll submit a fix.
2) TestIds - apparently this error has been there for awhile, but it was shadowed before adf8a1f. Looks like a change in `xmlschema` dependency - it's an easy fix, I'll submit it too.

EstebanDugueperoux2:
Hi @Andrej730 now it is toy.test_run which is failing all of times on main branch (v0.8.0) and my PR.
I don't know this .pyi file, it seems used only during development and not at runtime.
Do you know what option generate this .pyi file?

Andrej730:
@EstebanDugueperoux2 
it's a stub file (https://mypy.readthedocs.io/en/stable/stubs.html#stub-files), it's needed to provide typing information for the .py wrapper. 
It's not automatically generated, it's just validated by `validate_stub.py`.
You can just ignore this test for now, still need to refine it.

PR #6804, opened 2025-06-13
#6803: Update IfcOffsetCurveByDistance to be compitable on Windows

EstebanDugueperoux2:
Update IfcOffsetCurveByDistance to be compitable on Windows with MSVC and C++20

In addition add option to be able to build IfcOpenShell against other C++ standard than C++17


aothms:
Thanks, usage of `or` here is really unidiomatic probably very much unintentional. Maybe for the c++ version it would still be nice to use 17 as a backup when unset?

EstebanDugueperoux2:
Waouh thanks @aothms, it was a fast review :).
I'll retablish set(CMAKE_CXX_STANDARD 17) only if not specified from command line in my next PR.

PR #6805, opened 2025-06-14
6803: Ease conan integration

EstebanDugueperoux2:
Move option "project()" cmake instruction before main options to have them changeable from a toolchain file as conan one.


cmake/CMakeLists.txt
```diff
@@ -374,133 +386,133 @@ if(BUILD_IFCGEOM)
 
     # Open CASCADE
     if(WITH_OPENCASCADE)
-    if("${OCC_INCLUDE_DIR}" STREQUAL "")
```
JackBoosY: Since OCCT now support to export their cmake config files, we should use `find_package` instead of the manually find functions.
See https://github.com/Open-Cascade-SAS/OCCT/blob/22d437b771eb322dcceec3ad0efec6876721b8a9/CMakeLists.txt#L1357

EstebanDugueperoux2: I agree as commented in https://github.com/IfcOpenShell/IfcOpenShell/discussions/6851#discussioncomment-13598957 but on [ubuntu](ubuntu:22.04) no OpenCASCADEConfig.cmake is available while with ubuntu:24.04 it is available.
And IfcOpenShell CI installs its prerequisites using ubuntu repositories, while `nix/build-all.py` build opencascade directly from source then in this case find_package() could be used.
If we want to use conan for IfcOpenShell build, we could use find_package() also.

theoryshaw:
@EstebanDugueperoux2 

PR #6816, opened 2025-06-18
TargetScale support Subcontext

falken10vdl:
Added the field for TargetScale in Geometric RepresentationContext
image

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/context/data.py
+++ b/src/bonsai/bonsai/bim/module/context/data.py
@@ -54,18 +54,17 @@ class ContextData:
         for subcontext in context.HasSubContexts:
             target_scale_denominator = None
             if hasattr(subcontext, "TargetScale") and subcontext.TargetScale is not None:
-                try:
-                    scale_value = float(subcontext.TargetScale)
-                    if scale_value > 0:
-                        target_scale_denominator = 1.0 / scale_value
-                except (ValueError, TypeError, ZeroDivisionError):
-                    pass
-                    
-            results.append({
-                "id": subcontext.id(),
-                "context_type": subcontext.ContextType or "",
-                "context_identifier": subcontext.ContextIdentifier or "",
-                "target_view": subcontext.TargetView or "",
-                "target_scale_denominator": target_scale_denominator,
-            })
+                scale_value = float(subcontext.TargetScale)
+                if scale_value > 0:
+                    target_scale_denominator = 1.0 / scale_value
+
+            results.append(
+                {
+                    "id": subcontext.id(),
+                    "context_type": subcontext.ContextType or "",
+                    "context_identifier": subcontext.ContextIdentifier or "",
+                    "target_view": subcontext.TargetView or "",
+                    "target_scale_denominator": target_scale_denominator,
+                }
+            )
         return results
--- a/src/bonsai/bonsai/bim/module/context/data.py
+++ b/src/bonsai/bonsai/bim/module/context/data.py
@@ -49,15 +49,23 @@ class ContextData:
         return results
 
     @classmethod
-    def get_subcontexts(cls, context: ifcopenshell.entity_instance) -> list[dict[str, Any]]:
+    def get_subcontexts(cls, context) -> list[dict[str, Any]]:
         results = []
         for subcontext in context.HasSubContexts:
-            results.append(
-                {
-                    "id": subcontext.id(),
-                    "context_type": subcontext.ContextType,
-                    "context_identifier": subcontext.ContextIdentifier,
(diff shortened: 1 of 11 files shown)


Andrej730:
@falken10vdl 

From what I've seen TargetScale is used very rarily - so the most of the times users will just see `N/A` and will never use scale input when adding a new context, so it may be not worth it to introduce an unnecessary noise to the UI. 

Maybe we can add some icon before pencil button to indicate that this subcontext is special and has TargetScale set?

Regarding the TargetScale editing UI - I haven't tested it myself yet, but the idea seems reasonable - to show a denominator instead of the more obscure scale value. Can you please update tests in `tool/test_context.py` accordingly?

Regarding `context.add_context`'s new argument - please add it to the function's doc-string.

theoryshaw:
I could see how using `TargetScale` could open some interesting possibilities.

falken10vdl:
@Andrej730 and @theoryshaw what do you think about a display like this?
image
it takes maximum 6 characters. If no scale is available then it is empty.
image

This would be the docstring for the parameter:
image

tool/test_context.py.... I am not familiar with the testing environment yet...  :)   I still have to spend some time learning how it works... Anyhow I have seen that targetScale is already in one test. I have added another:

test_importing_another_subcontext_with_different_scale:
image

test_exporting_another_subcontext_with_different_scale:
image


Let me know what you think
Cheers!

Andrej730:
@falken10vdl 

> what do you think about a display like this?

Looks okay, so there won't be unnecessary "N/A" all the time. Though I still don't see the need for setting TargetScale when adding a context - it can always be set later if needed.

> tool/test_context.py.... I am not familiar with the testing environment yet... :) I still have to spend some time learning how it works... Anyhow I have seen that targetScale is already in one test. I have added another:

I believe there is no need for new tests - older import/export tests should do, they will start failing after this PR and they need to be adjusted.

falken10vdl:
@Andrej730 Ok. I remove the added tests since there are already two with TargetSacle being checked.

>  Though I still don't see the need for setting TargetScale when adding a context - it can always be set later if needed.

Do you mean removing this, right? and just having that in the edit menu.
image

Thanks!

Andrej730:
> Do you mean removing this, right? and just having that in the edit menu.

Yeah

falken10vdl:
Updated. Removed redundant tests and TargetScale prop in Geometric Representation Contexts panel. TargetScale is added/changed in editing mode.
Cheers!

Andrej730:
@falken10vdl thanks.

1. Regarding tests - still need to adjust existing tests for a new behaviour, I believe import and export attributes tests will fail, as they're expecting TargetScale attribute.

https://github.com/IfcOpenShell/IfcOpenShell/blob/9b58d4aa479bbde19fba7b99c7afed08286d26d9/src/bonsai/test/tool/test_context.py#L68

2. Property that was used for adding TargetScale for a new context is still present in the PR - see `bim.add_context`, `context/prop.py`, `context/ui.py`.

3. In ui.py `if "target_scale_denominator" in subcontext` seems unnecessary as all dictionaries will have this key either way.

4. In data.py `hasattr(subcontext, "TargetScale")` seems also unnecessary, see IfcGeometricRepresentationContext documentation (https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcGeometricRepresentationContext.htm) - HasSubContexts will always be a tuple of IfcGeometricRepresentationSubContext and they will always have TargetScale.

5. Please run black formatter on this PR.

falken10vdl:
@Andrej730 Thanks for the details. Please take a look to the update.
Thanks!

Andrej730:
@falken10vdl 

See:

> 2\. Property that was used for adding TargetScale for a new context is still present in the PR - see `bim.add_context`, `context/prop.py`, `context/ui.py`.

falken10vdl:
@Andrej730 

> 2. Property that was used for adding TargetScale for a new context is still present in the PR - see bim.add_context, context/prop.py, context/ui.py.

Sorry I do not follow you :) Can you explain it?
Thanks!

Andrej730:
@falken10vdl 

> Sorry I do not follow you :) Can you explain it?

Sure, `representation_target_scale_denominator` property that's not used anymore after https://github.com/IfcOpenShell/IfcOpenShell/pull/6816/commits/bb9cc11c82b5c305e6436a337d0accf1a23043f1 is still referenced by 3 files and those references can be removed to make it clean.

image
image
image

falken10vdl:
@Andrej730 Thanks for the explanation! I had forgotten we had removed the property from the UI.
Thanks!

Issue #6254, opened 2025-03-01
infra authoring api design


civilx64:
### Working outline for infrastructure authoring api design

This is an initial, very rough outline for planning purposes.

Consider this a Request For Comments (RFC) moreso than a set-in-stone to-do list

#### ifcopenshell.api.alignment

Well underway - see #6234 

#### ifcopenshell.api.cogo

Coordinate geometry - primarily survey points and control monuments for layout, parcels, etc.

#### ifcopenshell.api.geotech

IfcBorehole, IfcGeotechnicalStratum, etc.

#### ifcopenshell.api.infra_common

Common functionality - primarily sweeping profiles along linear elements (alignments) to generate solids and surfaces

IfcCourse is a good example of what would live here, e.g. `add_course`, `edit_course`, etc.

#### ifcopenshell.api.bridge

Bridge-specific considerations like IfcBridgePart

#### ifcopenshell.api.rail

Rail-specific considerations go here.  Volunteers with rail domain expertise welcome to lead development of this namespace!

#### ifcopenshell.api.road

Road-specific considerations like superelevation would live here.  
Rail superelevation (cant) is handled in the `alignment` api with IfcAlignmentCant

RickBrice:
Would COGO include calculations like "locate point by distance and bearing/azimuth/delta", intersect lines,circles,arcs, tangents to circles, stuff like that?

civilx64:
I typically only think of importing survey points and potentially setting key points for alignments, but we could consider all of the things you listed as in scope for this namespace.

RickBrice:
ifcopenshell.api.linear_ref - converting IfcAxis2PlacementLinear to a 4x4 matrix probably belongs in ifcopenshell.util.placement.

civilx64:
Agreed. I was thinking more in terms of having IfcReferent in this namespace. If we take the approach that linear referencing requires an alignment, then IfcReferent could stay in the `alignment` namespace and we could do away with `linear_ref`.

RickBrice:
@civilx64 I've got a project, and the current alignment API isn't getting the job done. Nothing like real work to drive requirements.

Also pinging @aothms just because.

Here is a stab and an updated API. Feedback please.

~~~
def create_alignment(
    file: ifcopenshell.file, name: str,
    include_vertical : bool = False,
    include_cant : bool = False
) -> entity_instance:
    """
[...]

PR #6817, opened 2025-06-18
Alignment api update

RickBrice:
Updates the alignment API to support layout by segments.
Geometry is now handled automatically. The API works with the business logic and the geometry curves are created and maintained behind the scenes.

Stationing referent implementation is limited. One referent is nested to the Alignment to define the stationing. One referent each is nested to curve segments to inform the meaning of the start point.


RickBrice:
@civilx64 @aothms 
An example script putting the updated API into action. Builds the alignment from the FHWA Bridge Geometry Manual
[IfcAlignmentAPI.py.txt](https://github.com/user-attachments/files/20819350/IfcAlignmentAPI.py.txt)

civilx64:
I am getting an error with the stationing - perhaps there are uncommitted changes in the `util` namespace?

```python
  File "/redacted/alignment_api_example.py", line 23, in <module>
    alignment = ifcopenshell.api.alignment.create(file, "E-Line", include_vertical=True, start_station=10000.)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/redacted/.venv/lib/python3.12/site-packages/ifcopenshell/api/alignment/create_alignment.py", line 72, in create
    __add_zero_length_segment(file, layout)
  File "/redacted/.venv/lib/python3.12/site-packages/ifcopenshell/api/alignment/__add_zero_length_segment.py", line 177, in __add_zero_length_segment
    name = f"{__get_segment_start_point_label(segment,None)} {ifcopenshell.util.stationing.station_as_string(file,0.0)}"
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/redacted/.venv/lib/python3.12/site-packages/ifcopenshell/util/stationing.py", line 29, in station_as_string
    value = math.fabs(station)
            ^^^^^^^^^^^^^^^^^^
TypeError: must be real number, not file

```

RickBrice:
Weird.... it looks like its committed

https://github.com/IfcOpenShell/IfcOpenShell/blob/e97f0cf6afdfbe97150c171b256d3eb6d2874561/src/ifcopenshell-python/ifcopenshell/util/stationing.py#L23

civilx64:
You are correct - user error on my part as I am sym-linking across a few projects and didn't grab everything I should have.

I ran the FHWA example successfully.  I had been testing with the WIP commits and had noticed a bug with .Transition = DISCONTINUOUS for the last two segments but that appears to have been fixed - thanks!

I am going to test with a few other alignments but will likely be a few days before I can do a full code review.

src/ifcopenshell-python/ifcopenshell/api/alignment/__add_segment_to_curve.py
civilx64, line 1: The typical naming convention for all of these "private" modules and functions would be to start with a single underscore, not a double underscore.  Double underscore, or "dunder", has special meaning in python since it is reserved for builtins and such.

src/ifcopenshell-python/ifcopenshell/api/alignment/__add_segment_to_curve.py
```diff
@@ -0,0 +1,134 @@
...
+        # assume IfcAxis2Placement2D
+        zero_length_segment.Placement.Location.Coordinates = (x, y)
+        zero_length_segment.Placement.RefDirection.DirectionRatios = (dx, dy)
+
+        __update_curve_segment_transition_code(curve_segment, zero_length_segment)
+
+
+def __add_segment_to_curve(file: ifcopenshell.file, segment: entity_instance, composite_curve: entity_instance) -> None:
+    """
+    Creates an IfcCurveSegment from the IfcAlignmentSegment and adds it to the composite curve. The IfcCurveSegment is added
```
civilx64: only to a composite curve or to the 'representation' curve, e.g. Composite, Gradient, or SegmentedReference curves.  (Appears to be the latter)

src/ifcopenshell-python/ifcopenshell/api/alignment/__add_segment_to_curve.py
```diff
@@ -0,0 +1,134 @@
...
+    if not composite_curve.is_a(expected_type):
+        raise TypeError(f"Expected to see '{expected_type}', instead received '{composite_curve.is_a()}'.")
+
+    # map the IfcAlignmentSegment to an IfcCurveSegment (or two in the case of helmert curves)
+    if segment.DesignParameters.is_a("IfcAlignmentHorizontalSegment"):
+        mapped_segments = __map_alignment_horizontal_segment(file, segment)
+    elif segment.DesignParameters.is_a("IfcAlignmentVerticalSegment"):
+        mapped_segments = __map_alignment_vertical_segment(file, segment)
+    elif segment.DesignParameters.is_a("IfcAlignmentCantSegment"):
+        alignment = segment.Nests[0].RelatingObject
```
civilx64: suggest 'cant_layout' variable name here instead of 'alignment'

src/ifcopenshell-python/ifcopenshell/api/alignment/__create_geometric_representation.py
```diff
@@ -25,25 +25,25 @@
...
 
 
-def create_geometric_representation(file: ifcopenshell.file, alignment: entity_instance) -> None:
+def __create_geometric_representation(file: ifcopenshell.file, alignment: entity_instance) -> None:
     """
-    Create geometric representation for the alignment.
+    Create geometric representation for the alignment and its nested layouts.
 
-    There are 5 different cases:
+    There are 5 different cases (the IfcCurve create is indicated):
```
civilx64:
```suggestion
    There are 5 different cases (the IfcCurve created is indicated):
```

src/ifcopenshell-python/ifcopenshell/api/alignment/__get_mapped_segments.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    if segment.DesignParameters.is_a("IfcAlignmentHorizontalSegment"):
+        return 2 if segment.DesignParameters.PredefinedType == "HELMERTCURVE" else 1
+    elif segment.DesignParameters.is_a("IfcAlignmentVerticalSegment"):
+        return 1
+    elif segment.DesignParameters.is_a("IfcAlignmentCantSegment"):
+        return 2 if segment.DesignParameters.PredefinedType == "HELMERTCURVE" else 1
+
+
+def __get_mapped_segments(file: ifcopenshell.file, segment: entity_instance) -> Sequence[entity_instance]:
```
civilx64:
```suggestion
def __get_mapped_segments(file: ifcopenshell.file, layout_segment: entity_instance) -> Sequence[entity_instance]:
```

Unfortunately we don't have any sort of entity typing on the python side so it is helpful to be precise with variable names.

src/ifcopenshell-python/ifcopenshell/api/alignment/__get_segment_start_point_label.py
```diff
@@ -0,0 +1,251 @@
...
+            label = lookup_table[prev_segment.DesignParameters.PredefinedType][segment.DesignParameters.PredefinedType]
+        elif segment.DesignParameters.is_a("IfcAlignmentVerticalSegment"):
+            lookup_table = {
+                "CIRCULARARC": {"CIRCULARARC": "xx", "CLOTHOID": "xx", "CONSTANTGRADIENT": "xx", "PARABOLICARC": "xx"},
+                "CLOTHOID": {"CIRCULARARC": "xx", "CLOTHOID": "xx", "CONSTANTGRADIENT": "xx", "PARABOLICARC": "xx"},
+                "CONSTANTGRADIENT": {
+                    "CIRCULARARC": "xx",
+                    "CLOTHOID": "xx",
+                    "CONSTANTGRADIENT": "P.V.I",
+                    "PARABOLICARC": "B.V.C.",
```
civilx64:
```suggestion
                    "PARABOLICARC": "P.V.C.",
```
point of vertical curvature

src/ifcopenshell-python/ifcopenshell/api/alignment/__get_segment_start_point_label.py
```diff
@@ -0,0 +1,251 @@
...
+                },
+                "PARABOLICARC": {
+                    "CIRCULARARC": "xx",
+                    "CLOTHOID": "xx",
+                    "CONSTANTGRADIENT": "E.V.C.",
```
civilx64:
```suggestion
                    "CONSTANTGRADIENT": "P.V.T.",
```
point of vertical tangency

src/ifcopenshell-python/ifcopenshell/api/alignment/__get_segment_start_point_label.py
```diff
@@ -0,0 +1,251 @@
...
+        raise TypeError(
+            f"Expected prev_segment.DesignParameters type to be one of {[_ for _ in expected_types]}, instead received {prev_segment.DesignParameters.is_a()}"
+        )
+    if segment != None and not segment.DesignParameters.is_a() in expected_types:
+        raise TypeError(
+            f"Expected segment.DesignParameters type to be one of {[_ for _ in expected_types]}, instead received {segment.DesignParameters.is_a()}"
+        )
+
+    label = "Unknown"
+    if prev_segment == None and segment != None:
```
civilx64: These abbreviations will vary between regions and agencies, so it would be nice to give the user some control over this.  The only prior art I'm aware of in bonsai is the concept of libraries.  I guess you could do that here also...?  Maybe a PropertySet of preferred abbreviations for each of these key locations along the alignment.

I added it to the Infra Authoring project as #6829.

src/ifcopenshell-python/ifcopenshell/api/alignment/get_alignment_station.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+
+    if not alignment.is_a("IfcAlignment"):
+        raise TypeError(f"Expected entity type to be IfcAlignment, instead received {alignment.is_a()}")
+
+    start_station = 0.0
+    components = ifcopenshell.util.element.get_components(alignment)
+    for c in components:
+        if c.is_a("IfcReferent") and ifcopenshell.util.element.get_predefined_type(c) == "STATION":
+            start_station = ifcopenshell.util.element.get_pset(c, name="Pset_Stationing", prop="Station")
```
civilx64, line 37: This works for alignments authored within bonsai / IFCOS but may not hold true for alignments coming from other systems.  Therefore I think the more robust solution would be to use the referent with the lowest value of IfcLinearPlacement.DistanceAlong.

src/ifcopenshell-python/ifcopenshell/api/cogo/angle_from_direction.py
```diff
@@ -0,0 +1,98 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with IfcOpenShell.  If not, see <http://www.gnu.org/licenses/>.
+
+import ifcopenshell.api.cogo.angle_from_dms
+import math
+
+def angle_from_direction(direction: str)->float:
```
civilx64:
```suggestion
def angle_from_bearing(bearing: str)->float:
```

src/ifcopenshell-python/ifcopenshell/api/cogo/angle_from_direction.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    Compute an angle, in radian, from a direction string.
```
civilx64:
```suggestion
    Compute an angle, in radian, from a quadrant bearing string.
```

civilx64 (commented):
Very nice !!!

I was struggling earlier with a scenario and I found a way past it, but it will fall down when we get to building the UI.  When you create an alignment, it first by necessity needs to be horizontal only.  

So the first step would be to call `create_alignment` and pass False for include_vertical and include_cant.  Then later you need to add the vertical layout, but the api doesn't appear to automatically add the IfcGradientCurve.  Is that by design?

src/ifcopenshell-python/ifcopenshell/api/alignment/get_alignment_station.py
RickBrice: Referents are ordered. The referent with the lowest value is the first referent. Unless get_components provides the referents out of order, this should work.

src/ifcopenshell-python/ifcopenshell/api/alignment/__get_segment_start_point_label.py
RickBrice: I struggled with this. Is there a way for a user to override the lookup tables?
Perhaps there could be a callback function (or 3) that returns the notation.

src/ifcopenshell-python/ifcopenshell/api/alignment/__get_segment_start_point_label.py
RickBrice: Changed as noted - I was using the notation I'm familiar with, and that underscores your point that we are all going to do this differently.

src/ifcopenshell-python/ifcopenshell/api/alignment/__get_segment_start_point_label.py
RickBrice: Revised as noted

src/ifcopenshell-python/ifcopenshell/api/alignment/__get_mapped_segments.py
RickBrice: good call - revised as noted

src/ifcopenshell-python/ifcopenshell/api/alignment/__create_geometric_representation.py
RickBrice: revised as noted

src/ifcopenshell-python/ifcopenshell/api/alignment/__add_segment_to_curve.py
RickBrice: Good catch - I knew there was one more to change but I couldn't find it.
Revised as noted

src/ifcopenshell-python/ifcopenshell/api/alignment/__add_segment_to_curve.py
RickBrice: Updated to say representation curve.

src/ifcopenshell-python/ifcopenshell/api/alignment/__add_segment_to_curve.py
RickBrice: You learn something new every day. File and function names updated to single underscore

RickBrice:
> I was struggling earlier with a scenario and I found a way past it, but it will fall down when we get to building the UI. When you create an alignment, it first by necessity needs to be horizontal only.

> So the first step would be to call create_alignment and pass False for include_vertical and include_cant. Then later you need to add the vertical layout, but the api doesn't appear to automatically add the IfcGradientCurve. Is that by design?

This scenario should be adequately handled. See [test_add_vertical_alignment.py](https://github.com/IfcOpenShell/IfcOpenShell/blob/rab_alignment_api_update/src/ifcopenshell-python/test/api/alignment/test_add_vertical_alignment.py)

When the alignment is created with only horizontal, the representation curve is IfcCompositeCurve

https://github.com/IfcOpenShell/IfcOpenShell/blob/b9818773b1b2431f1dfee81e00736495c6438e4a/src/ifcopenshell-python/test/api/alignment/test_add_vertical_alignment.py#L38-L46

Then a vertical layout is added and the representation curve changes to and IfcGradientCurve (whose basis curve is the IfcCompositeCurve)

https://github.com/IfcOpenShell/IfcOpenShell/blob/b9818773b1b2431f1dfee81e00736495c6438e4a/src/ifcopenshell-python/test/api/alignment/test_add_vertical_alignment.py#L48-L57

What struggle did you have? How did you resolve it?

civilx64:
> What struggle did you have? How did you resolve it?

Sorry, I did not provide a useful bug report.  Here's what I thought I had done:

1. api.alignment.create(..., include_vertical=False, include_cant=False)
2. api.alignment.add_vertical_layout()
3. api.alignment.layout_vertical_alignment_by_pi_method()
4. NoneType error in api.alignment.get_curve() because the IfcGradientCurve could not be found

My fix was to change 1. to include_vertical=True

But, now when I re-run my driver script with this same sequence, I don't get an error and everything appears to process fine.  In a earlier version I was using file.create_entity(type="IfcAlignmentVertical") for the vertical layout, without creating the corresponding IfcGradientCurve.  So quite possibly user error on my part *again*.

src/ifcopenshell-python/ifcopenshell/api/alignment/__get_segment_start_point_label.py
civilx64: You would set up separate libraries (IFC files) containing the appropriate values per region / country / agency, etc.  Or, since it's a Pset, the user can directly override the values after they have been imported into the project.

RickBrice: The abbreviations are just text used in creating the IfcReferent.Name - Not sure how a library would support that. I don't see Pset being used here except to store the station value. Pset_Stationing doesn't have any kind of name or type field.

Maybe my approach off base. For the AASHTO IDM, we need to provide key alignment points (PC, PT, etc). I've got the API creating the alignment segments and nesting an IfcReferent with the segment where the key point occurs.

Is there a different approach we should be considering?

civilx64: Let me phrase it another way:

In some BIM authoring tools, the library of standard steel shapes exists as a text-based configuration file.  In bonsai, that "configuration" file is actually an IFC model with the steel profiles defined as a library.  See [IFC4 US Steel.ifc](https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/bonsai/bonsai/bim/data/libraries/IFC4%20US%20Steel.ifc).  So that's where I was initially headed - configuration as IFC for consistency.

The custom Pset containing the abbreviations for alignment key points would be purely for the api to use when creating and editing alignments.  More specifically for the text values in IfcReferent.Name as you had noted.  If we went this route, those settings would "travel along" with models authored in bonsai, and we wouldn't need any additional UI development for loading and editing - they would just show up as any other property set attached to IfcProject.  You could maybe even have a listener attached to the UI that would update all of the referent names upon a change to the prefixes as stored in that property set.

The API is in great shape - this is just exploring alternate paths to migrate away from the values currently hardcoded in the API logic.  And from my perspective the logical alternate to the IFC library - custom Pset approach would be something like `alignment_settings.json` in the root of the api folder that the alignment api retrieves the prefixes from.

src/ifcopenshell-python/ifcopenshell/util/stationing.py
```diff
@@ -17,15 +17,29 @@
...
-    @param accuracy: number of digits following the decimal place
+    Returns a stringized version of a station. Example 100.0 is 1+00.00 as a stationing string.
+    If the project units are SI-based, the string is in the format xxx+yyy.zzz
+    If the project units are Emperial-based, the string is in the format xx+yy.zz
+    :param station: the station to be stringized
+    :return: stringized station
     """
+    
+    unit_type = ifcopenshell.util.unit.get_project_unit(file,"LENGTHUNIT")
+    if unit_type.is_a("IfcConversionBasedUnit"):
```
civilx64, line 33: Why the unit conversions in this function?  Is the station assumed to be in SI units?  It should not be.  Unless I'm missing something obvious, this function should be purely about string formatting.  If unit conversion is required (e.g. a station has been calculated based on Distance Along in m) then it should happen prior to calling this function.

Observed behavior behind this comment:

I have been testing the api with a number of different alignments from projects past and present.  I have two ifc "templates" that I have been using - one for international feet and one for US survey feet.  If I use the former, the station values end up as expected.  If I use the US survey foot template, the stations are assumed to be meters and are multiplied by 3.2808.

RickBrice: An IFC file can have any length unit (centimeter, millimeter, foot, inch, etc.) but IfcOpenShell works in meter.

The unit conversion is to convert from IfcOpenShell's units to "foot" for Imperial units and "meter" for SI units.

To determine if stations should be in foot or meter, I look at the project unit type. If IfcConversionBasedUnit I assume foot, otherwise meter.

Can you share your example with US survey foot length? This function needs a bit more work.

src/ifcopenshell-python/ifcopenshell/api/alignment/__get_segment_start_point_label.py
RickBrice: I think callbacks are the cleanest and most flexible way to handle this.

See [test_referent_names.py](https://github.com/IfcOpenShell/IfcOpenShell/blob/rab_alignment_api_update/src/ifcopenshell-python/test/api/alignment/test_referent_names.py)

Do nothing and get the hard-coded solution.

Create callback functions and use any naming prefix you want. The callback can have hard coded values, read from a JSON file, or anything else someone wants to implement.

src/ifcopenshell-python/ifcopenshell/util/stationing.py
aothms: There's a couple of things here, conversion-based not necessarily means imperial, non-conversion based also not necessarily means meter (could also have some prefix) and if the stations are length measures then the project unit factor applies incl. prefixes.

I would say that the number formatting and precision could indeed rather be stored in a custom pset, like that idea a lot.

RickBrice: I misrepresented the intent in my last post - let me restate why there is a unit conversion.

The various IfcAlignmentParameterSegment sub-types have parameters for length (curve length or horizontal length). These parameters are in project units. 

The function in question aims to create a text string that represents a station. For Imperial units, the string format is xx+yy.zz in feet. For SI units, the string format is xxx+yyy.zzz in meter. Example for a station at 100ft is 1+00.00.

When the project units are inch, the station at the start of a segment is the start station value plus the sum of all the preceding segment lengths - the resulting value in inches. To create the correct string version of the station, the length in inch needs to be converted to a length in foot (e.g. 1200.0 in = 100.0 ft -> 1+00.00). Similarly, for a project length unit of millimeter or centimeters, the station value needs to be converted to meter to create the correct string.

Two things are needed in the function. (1) Are project units Imperial or SI? (2) Need to convert the station value from project units to either foot or meter for the string.

What is a more fool-proof way to determine if project units are Imperial or SI?

The resulting string is assigned to IfcReferent.Name along with a prefix such as P.C. for Point of Curvature for horizontal.

How would a custom pset work? What would it be associated with? How does a pset help develop the referent's name?

civilx64: Sorry, this comment should have been on L34, not L33.  I understand why unit conversions are needed to arrive at the correct value of the stationing based on project units and calculating distance along the segments.  My comment was more about the contract between this function and the caller.

Looking at the function signature, one would assume that whatever is passed as `station` will be 'stringized'.  If I pass `10000.` and have a model in imperial units, I expect to get back `10+00.00` (depending on precision, of course).  If I do the same in a model with SI units, I expect to get back `1+000.00`.

Said another way, shouldn't the scope of this function purely be related to string processing?  Any unit conversions required should happen prior to calling this function, IMO.  

As for the custom Pset - I can put together an example, but the callback approach will more than suffice for the time being.

RickBrice: > Said another way, shouldn't the scope of this function purely be related to string processing? Any unit conversions required should happen prior to calling this function, IMO.

I was thinking that the unit conversion happens inside the function so the caller only needs to be concerned with passing in the station value in project units. Otherwise, it is the responsibility of the caller to do the unit conversion from every call site.

crystalinv085:
> @civilx64 @aothms An example script putting the updated API into action. Builds the alignment from the FHWA Bridge Geometry Manual [IfcAlignmentAPI.py.txt](https://github.com/user-attachments/files/20819350/IfcAlignmentAPI.py.txt)

Hello This is Jonathan here from Stefano Verugi's office. I want to run this script to test out the updated alignment api but I'm unable to do so. Since this updated api is not part of the official ifcopenshell that comes with Bonsai, How can configure my blender to point to this repository. Kindly assist me as this is my first time of doing this. I want to see how best I can contribute to the alignment api development.

aothms:
@crystalinv085 have a look at https://docs.bonsaibim.org/guides/development/installation.html#live-development-environment let us know if you run into issues

RickBrice:
I renamed the `ifcopenshell.api.cogo.angle_from_bearing` function to `ifcopenshell.api.cogo.bearing2dd` so the naming is consistent with `ifcopenshell.util.geolocation.dms2dd` and similarly named functions.

My question is, should `bearing2dd` be in the cogo API or it be moved to the geolocation API?

Background: See [https://github.com/IfcOpenShell/IfcOpenShell/issues/6254](https://github.com/IfcOpenShell/IfcOpenShell/issues/6254) for discussion of cogo API

civilx64:
dms2dd and the inverse make sense in geolocation for dealing with lat/long coordinates. dms and dd with quadrant indicators is more specific to coordinate geometry for metes and bounds surveying - including alignment layout. I say it's ok to leave as-is. Perhaps a short cross-reference in the docs for each would be useful also.

RickBrice:
That kind of sounds like "do nothing" - My favorite option

crystalinv085:
> @crystalinv085 have a look at https://docs.bonsaibim.org/guides/development/installation.html#live-development-environment let us know if you run into issues

Thanks Thomas for the link. I have gone through the section of setting up the live development environment for bonsai. I have installed the unstable version and I'm now able to get the updates. I cloned the _**rab_alignment_api_update**_ repository and copied the script _**dev_environment.bat**_ and run it as administrator. The script run without error but my Bonsai is not pointing towards the cloned repository. Any pointer on what i may be doing wrong or how to set it up?

RickBrice:
@crystalinv085 What I do is build IfcOpenShell C++ from source and then run the install script which copies all the python code from F:\IfcOpenShell\src\ifcopenshell-python\ifcopenshell to F:\IfcOpenShell\_deps-vs2022-x64-installed\Python311\Lib\site-packages\ifcopenshell

This is probably overkill for what you want to do.

Did you update the REPO_PATH variable in the dev_environment.bat script to the location where you have IFCOS cloned?

The example script does not need to run from python within Bonsai/Blender. You can run the script from Python IDLE or from the command line

~~~
python IfcAlignmentAPI.py
~~~

crystalinv085:
> @crystalinv085 What I do is build IfcOpenShell C++ from source and then run the install script which copies all the python code from F:\IfcOpenShell\src\ifcopenshell-python\ifcopenshell to F:\IfcOpenShell_deps-vs2022-x64-installed\Python311\Lib\site-packages\ifcopenshell
> 
> This is probably overkill for what you want to do.
> 
> Did you update the REPO_PATH variable in the dev_environment.bat script to the location where you have IFCOS cloned?
> 
> The example script does not need to run from python within Bonsai/Blender. You can run the script from Python IDLE or from the command line
> 
> ```
> python IfcAlignmentAPI.py
> ```

@RickBrice I uncommented and set the REPO_PATH variable to the location of the cloned repository.

RickBrice:
On my system I have 'C:\Users\redacted\AppData\Roaming\Blender Foundation\Blender\4.4\extensions\.local\lib\python3.11\site-packages\ifcopenshell' sym-linking to 'F:\IfcOpenShell\src\ifcopenshell-python'

With that, I can open the IfcAlignmentAPI.py example in the Blender Scripting window and run it.

What is your symlink for IfcOpenShell?

What errors are you getting?

crystalinv085:
> On my system I have 'C:\Users\redacted\AppData\Roaming\Blender Foundation\Blender\4.4\extensions.local\lib\python3.11\site-packages\ifcopenshell' sym-linking to 'F:\IfcOpenShell\src\ifcopenshell-python'
> 
> With that, I can open the IfcAlignmentAPI.py example in the Blender Scripting window and run it.
> 
> What is your symlink for IfcOpenShell?
> 
> What errors are you getting?

I cloned the repository to C:\Users\jaf08\Documents\GitHub\IfcOpenShell_TestBranch. and set the REPO_PATH variable to C:\Users\jaf08\Documents\GitHub\IfcOpenShell_TestBranch\IfcOpenShell. So I expect the link to point to this path after running the _dev_environment.bat_. 

This is the error i get after running IfcAlignmentAPI.py

  File "\IfcAlignmentAPI.py", line 23, in <module>
AttributeError: module 'ifcopenshell.api.alignment' has no attribute 'create'

RickBrice:
Go into your `C:\Users\jaf08\AppData\Roaming\Blender Foundation\Blender\4.4\extensions\.local\lib\python3.11\site-packages` folder, find the `ifcopenshell` folder. Look at the folder properties. It should be a link to `C:\Users\jaf08\Documents\GitHub\IfcOpenShell_TestBranch\IfcOpenShell\src\ifcopenshell-python`

Maybe your REPO_PATH in dev_environment.bat should be `C:\Users\jaf08\Documents\GitHub\IfcOpenShell_TestBranch`

crystalinv085:
> Go into your `C:\Users\jaf08\AppData\Roaming\Blender Foundation\Blender\4.4\extensions\.local\lib\python3.11\site-packages` folder, find the `ifcopenshell` folder. Look at the folder properties. It should be a link to `C:\Users\jaf08\Documents\GitHub\IfcOpenShell_TestBranch\IfcOpenShell\src\ifcopenshell-python`
> 
> Maybe your REPO_PATH in dev_environment.bat should be `C:\Users\jaf08\Documents\GitHub\IfcOpenShell_TestBranch`

Thanks Rick. I  am able to connect it now. The issue was with the REPO_PATH variable. It should have been C:\Users\jaf08\Documents\GitHub\IfcOpenShell_TestBranch` as you suggested.

RickBrice:
@crystalinv085 Awesome. I'd appreciate any feedback and ideas you have for the API.

Here is the source document for the example alignment. https://www.[fhwa.dot.gov/bridge/pubs/hif22034.pdf](https://www.fhwa.dot.gov/bridge/pubs/hif22034.pdf)

If you have real examples of railway alignments, with cant, I'd love to get them.

crystalinv085:
> @crystalinv085 Awesome. I'd appreciate any feedback and ideas you have for the API.
> 
> Here is the source document for the example alignment. [https://www.[fhwa.dot.gov/bridge/pubs/hif22034.pdf](https://www.fhwa.dot.gov/bridge/pubs/hif22034.pdf)](https://www.%5Bfhwa.dot.gov/bridge/pubs/hif22034.pdf%5D(https://www.fhwa.dot.gov/bridge/pubs/hif22034.pdf))
> 
> If you have real examples of railway alignments, with cant, I'd love to get them.

You have done a good job with the API. I have a created a simple alignment (Horizontal Alignment) in Autodesk Civil 3D. I'm going to use the api to replicate the same alignment in Bonsai so we can compare. I will give you feedback on it. 

I have railway alignments in Civil 3D. But in what form do you want the railway alignment? PDF? I can export the geometric data into excel if you prefer. Kindly let me know. I have attached a screenshot of one of the plan and profile sheets for your perusal.
Railway alignment and profile

RickBrice:
PDF's would be fine. Thanks

RickBrice:
Any more feedback on this PR? I'd like to get this wrapped up over the next couple of weeks.

The outstanding questions seem to be:
1) Should `ifcopenshell.util.stationing.station_to_string` stringize the station value provided (which makes the caller perform any necessary unit conversion), or should the provided station value be in project units and the function performs the unit conversion to foot or meter units as appropriate (this option is implemented)?
2) If the second choice of 1), what is a fool-proof method to determine if project units are Imperial or SI?
3) Are we satisfied with how key alignment point IfcReferent are nested to individual alignment segments? Current implementation is to nest the referent with the segment where it occurs. Other options include nesting with the layouts in the same nest as the segments (alternating IfcReferent and IfcAlignmentSegment), having two nests (one for referents and one for segments), and nesting referents, in one or two nests, with the IfcAlignment rather than the layouts.

crystalinv085:
> PDF's would be fine. Thanks
Rick I have attached a sample railway alignment in pdf as requested. Kindly check if it's ok for your use.
[SAMPLE RAILWAY ALIGNMENT SETTING OUT DATA.pdf](https://github.com/user-attachments/files/21106231/SAMPLE.RAILWAY.ALIGNMENT.SETTING.OUT.DATA.pdf)

RickBrice:
@crystalinv085 Thank you - that is informative and educational for me. I will try to build this layout using the alignment api.

Two questions:
Are the transitions clothoid curves?

What is the meaning of the Shift, S? (Sorry, I'm not a railway guy). Looking at the Curve Table Explanatory Scheme, my guess would be S is the distance between the circular curve and the curve that would be formed if the initial and final transition curves extended.

aothms:
For me everything good to go, we can always fix things forward as more opinions are voiced. One of the  tests seems to be failing though

afbeelding

aothms (approved):

crystalinv085:
> Any more feedback on this PR? I'd like to get this wrapped up over the next couple of weeks.
> 
> The outstanding questions seem to be:
> 
> 1. Should `ifcopenshell.util.stationing.station_to_string` stringize the station value provided (which makes the caller perform any necessary unit conversion), or should the provided station value be in project units and the function performs the unit conversion to foot or meter units as appropriate (this option is implemented)?
> 2. If the second choice of 1), what is a fool-proof method to determine if project units are Imperial or SI?
> 3. Are we satisfied with how key alignment point IfcReferent are nested to individual alignment segments? Current implementation is to nest the referent with the segment where it occurs. Other options include nesting with the layouts in the same nest as the segments (alternating IfcReferent and IfcAlignmentSegment), having two nests (one for referents and one for segments), and nesting referents, in one or two nests, with the IfcAlignment rather than the layouts.

1. I think the ifcopenshell.util.stationing.station_to_string function primarily should focus on formatting the provided station value without any unit conversion. In Civil 3D, once you set the project unit to metric, you work in meters. when you set it to imperial you work in feet. With this, all the function will do will be to format it as km+mm (00+000) or ft+in (00+00) depending on the project unit.
2. the length unit should determine whether the project is in metric or imperial. I wish we can restrict the unit for infrastructure projects to meters or feet. that will make it easy to determine.
3. The geometric points (IfcReferent) in my opinion should be nested to the horizontal alignment (horizontal layout) and not the individual segments.

crystalinv085:
> @crystalinv085 Thank you - that is informative and educational for me. I will try to build this layout using the alignment api.
> 
> Two questions: Are the transitions clothoid curves?
> 
> What is the meaning of the Shift, S? (Sorry, I'm not a railway guy). Looking at the Curve Table Explanatory Scheme, my guess would be S is the distance between the circular curve and the curve that would be formed if the initial and final transition curves extended.

The transition curves are clothoid spirals. That's what we use in railway and road alignment.
Yes, the s is the shift. When you add transitions to the in and out of the simple circular curve, it improves (flattens) the curve by shifting in inwards. the distance between the simple circular curve without the transition and the one with transition is the shift (s)

RickBrice:
@aothms does pytest run tests in parallel?

It looks like `_test_with_callbacks` is getting called before `_test_with_default_names` finishes.

https://github.com/IfcOpenShell/IfcOpenShell/blob/43110f6a70403d4924d20daab1c115c0ddae8389/src/ifcopenshell-python/test/api/alignment/test_referent_names.py#L113-L115

Is there a way to call the test functions in series? Or is it better to just put everything into a single function?

RickBrice:
@crystalinv085 Thanks for the example. I was able to build a script to create the alignment with the cant. 
[Exit_Turnout_Huni_Valley_Station.py.txt](https://github.com/user-attachments/files/21111431/Exit_Turnout_Huni_Valley_Station.py.txt)

You'll probably need to pull the latest code since I uncovered (and fixed) many problems with cant/SegRefCurves.

I can see now why you would want a CSV file, PI Method for importing alignments with transition curves. That should be easy enough to do.

civilx64:
> Is there a way to call the test functions in series? Or is it better to just put everything into a single function?

I'd recommend refactoring to a test class with methods for the individual tests.  I'll put together a quick edit and push on your branch.
Edit: actually don't need a class, just bundling the setup into fixtures helps things run as desired. See 27d3a3e.

civilx64 (approved):
Agree with Thomas, this can be merged at any time.

Issue #4067, opened 2023-12-01
Rounded objects look choppy in the svg


theoryshaw:
Rounded objects look choppy in the svg

file: 

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/d966f2aa-8400-4d69-8306-71d5e5afabf3)

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/716e59df-e861-43dc-9d0c-e25b7520ef3f)

aothms:
So two things to play with:

https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/draw.py#L121 and more importantly `setUseHlrPoly(False)`.

By default we use Polygonal HLR because anyway both svgfill nor shapely can deal with curved SVG elements (I guess...).

The other thing to try is:

```
s = ifcopenshell.geom.settings()
s.set_deflection_tolerance(0.0001)
```

theoryshaw:
Thanks @aothms 

Reporting back. 

When using `setUseHlrPoly(False)` the following is svg is output.  Neither 'Calculate Shapely Surfaces' or 'Calculate SVGFills Surfaces' is enabled. 

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/6d5ed160-4145-48b7-ab2d-648bbc1f0ede)

...

Dumb question, where do I add the following?

```
s = ifcopenshell.geom.settings()
s.set_deflection_tolerance(0.0001)
```

theoryshaw:
this ifc currently opens with a lot of stuff missing. 

![Image](https://github.com/user-attachments/assets/a23fd17f-1e91-4b35-9740-9b688fe5d0fe)

theoryshaw:
will test.

theoryshaw:
This file crashes blender: https://github.com/IfcOpenShell/IfcOpenShell/issues/6589

sboddy:
@theoryshaw Looks like Andrej fixed 6589, and my test showed your file at least loads now. It just looks messed up, with lots of parts missing like you said. I did encounter some unrelated errors trying to enable the status filters. I added some conditional checks which fixed them, but I have my suspicions that your file has issues, maybe corruption.

I had to check data existence/type to make part of the status filter enablement function on your data:

![Image](https://github.com/user-attachments/assets/92bc2b3c-1cbc-4743-8184-293afb2cabd0)

[...]

PR #6826, opened 2025-06-20
Fix #5563 inconsistent behaviour of Status filter

### LINKED ISSUE #5563: Some inconsistent UI behaviour with the Status filter under the Costing and Scheduling tab.
Referencing the discussion [here](https://community.osarch.org/discussion/2562/is-it-ok-to-ab-use-sub-storeys-to-show-hide-elements/p1?new=1).

[status_filter.ifc.zip](https://github.com/user-attachments/files/17348072/status_filter.ifc.zip)

Three "unexpected interactions" as I would call them, where the interface goes left, when you expect it to go right.

1. Using the Status filter hide each status in turn. When you click to hide the last status the eye icon will close, but the item remains visible in the 3D viewport. It is always the last visible one that won't hide. Either hide the item, or don't allow the last status to be hidden.

2.
     a) Set one status to shown and three to hidden. Now disable the status filter with the  (X) button. Should the hidden statuses be made visible?
     b) Re-enable the status filters. It now shows all statuses as visible, even though they are hidden in the 3D viewport. On it's own not too bad, but combined with 2c, it is jarring.
     c) Click to hide one of the status, and all the other hidden statuses are made visible.

~~3. In other parts of the UI I can select a hidden object. If I use the select button on a hidden status it does not select the hidden items. UI should be consistent.~~ I could have sworn I'd used select button to select hidden items, but I think I was wrong.

sboddy:
Fixes for inconsistency as detailed in #5563
- Hiding last status hides last items in viewport,
- Disabling the status filter makes all status' visible again.

### DIFF
--- a/src/bonsai/bonsai/bim/module/sequence/operator.py
+++ b/src/bonsai/bonsai/bim/module/sequence/operator.py
@@ -72,6 +72,28 @@ class DisableStatusFilters(bpy.types.Operator):
 
     def execute(self, context):
         props = context.scene.BIMStatusProperties
+
+        query = []
+        visible_statuses = {s.name for s in props.statuses}
+        for name in visible_statuses:
+            if name == "No Status":
+                q = f"IfcProduct, /Pset_.*Common/.Status=NULL, EPset_Status.Status=NULL"
+            else:
+                q = f"IfcProduct, /Pset_.*Common/.Status={name} + IfcProduct, EPset_Status.Status={name}"
+            query.append(q)
+        query = " + ".join(query)
+
+        if not query:
+            self.report({"INFO"}, "No statuses selected.")
+
+        visible_elements = ifcopenshell.util.selector.filter_elements(tool.Ifc.get(), query)
+
+        for obj in bpy.context.view_layer.objects:
+            element = tool.Ifc.get_entity(obj)
+            if not element or not element.is_a("IfcProduct"):
+                continue
+            obj.hide_set(element not in visible_elements)
+
         props.is_enabled = False
         return {"FINISHED"}
 
@@ -96,7 +118,6 @@ class ActivateStatusFilters(bpy.types.Operator):
 
         if not query:
             self.report({"INFO"}, "No statuses selected.")
-            return {"FINISHED"}
 
         visible_elements = ifcopenshell.util.selector.filter_elements(tool.Ifc.get(), query)
 
--- a/src/bonsai/bonsai/bim/module/sequence/operator.py
+++ b/src/bonsai/bonsai/bim/module/sequence/operator.py
@@ -43,17 +43,19 @@ class EnableStatusFilters(bpy.types.Operator):
     def execute(self, context):
         props = context.scene.BIMStatusProperties
         props.is_enabled = True
+        hidden_statuses = {s.name for s in props.statuses if not s.is_visible}
 
         props.statuses.clear()
 
         statuses = set()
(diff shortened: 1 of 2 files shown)


Andrej730:
Looks great, thank you for looking into this.

A few notes:
- I think we can remove `self.report({"INFO"}, "No statuses selected.")` branch entirely as it's now insignificant
- I find the approach you chose for the disabling indeed the most intuitive one. But I think it might need a second part - `bim.enable_status_filters` to remember the last applied filters and apply them. Currently `bim.enable_status_filters` just obliterates previous status values and the user would need to reapply them manually each time.
- Now `bim.activate_status_filters` and `bim.disable_status_filters` reuse a lot of code. Would be nice to move it to a common method somewhere in `tool.Sequence`.

Unrelated to this PR, just a thing I've noticed testing this - `bim.activate_status_filters` operator is not exposed to UI and it's only used internally in a few places. 
Currently, when user have filters active and some filter goes out of date (e.g. a new element introduced that wasn't affected by filters or some older element changed status and perhaps should be hidden now), they would switch `is_visible` toggle to refresh it.
It probably would be more intuitive if we would expose `bim.activate_status_filters` as a plain refresh button - so refresh would be represented more explicitly in the UI and users would be able to add this operator to Quick Favorites to update whenever they feel like it. 

----

Anyways, I'm fine merging it the way it is now and fix those caveats later, let me know if you're interested in working on this or we should proceed with the currrent version of the PR for now.

sboddy:
1. OK, Will do.
2. I certainly had that thought in mind. I was just looking at fixing the basic inconsistencies initially. But I'll come up with something for this.
3. Yeah... I knew it was ugly dup. code. Just wasn't sure where to factor it out to, but you've pointed the way :-)

I'll pop this PR into draft and do a fix it up.

sboddy:
I'm happy with this now. Please merge.

src/bonsai/bonsai/bim/module/sequence/operator.py
```diff
@@ -43,17 +43,19 @@ class EnableStatusFilters(bpy.types.Operator):
...
 
         statuses = set()
         for element in tool.Ifc.get().by_type("IfcPropertyEnumeratedValue"):
             if element.Name == "Status":
-                pset = element.PartOfPset[0]
-                if pset.Name.startswith("Pset_") and pset.Name.endswith("Common"):
-                    statuses.update(element.EnumerationValues)
-                elif pset.Name == "EPset_Status":  # Our secret sauce
-                    statuses.update(element.EnumerationValues)
+                if element.PartOfPset and isinstance(element.EnumerationValues, tuple):
```
sboddy, line 53: I've just realized I left this in this PR by mistake. I added it looking at https://github.com/IfcOpenShell/IfcOpenShell/issues/4067#issuecomment-3002194107 due to some peculiar data causing the status filter to error. I don't know if the problem was theoryshaw's file being broken somehow. Might as well leave it in as an added resilience bonus.

Andrej730: Well, technically leaving `EnumerationValues` as `None` is a valid IFC, so we might as well protect ourselves from this case.

image

Though no idea how it can occur - in theory in Bonsai if 0 elements are selected in Status enum, then `IFCPROPERTYENUMERATEDVALUE` removed completely.

Andrej730:
Looks good! 🚢

PR #6827, opened 2025-06-20
Fix #4588 fix broken rev cloud styles in sheets.

### LINKED ISSUE #4588: Revision clouds show on drawings, but don't show on the sheet
Revision clouds show on drawings, but don't show on the sheet

ifc: 

2024-04-25_22-44-18_chrome1ST_FLOOR_RCP svg_-_Google_Chromechrome
2024-04-25_22-44-44_chromeA150_-_REFLECTED_CEILING_PLAN svg_-_Google_Chromechrome

sboddy:
Beware... I have only lightly tested this on my simple test case.

It is possible that a better fix is to style the rev cloud with the embedded stylesheet, but at the moment the svg tags forming the rev cloud seem to be directly styled with a style attribute, resulting in broken clouds in sheets.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/sheeter.py
+++ b/src/bonsai/bonsai/bim/module/drawing/sheeter.py
@@ -374,6 +374,8 @@ class SheetBuilder:
             if "filter" in attrib:
                 # example use "#fill-background" filter
                 attrib["filter"] = replace_urls(attrib["filter"])
+            if "style" in attrib:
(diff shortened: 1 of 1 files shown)


Andrej730:
I think the fix is 100% correct - when we introduced separate class namespace for each drawing we updated `id` for the revision cloud marker but haven't updated `style` that's still using the old marker's `id`.

Thank you for looking into this and providing a solution.

Issue #6823, opened 2025-06-19
Right click on an instance, and be able to instantiate that instance in the scene.


theoryshaw:
### Feature Description

Would be nice to right click on an instance, and be able to instantiate that instance in the scene.  

That is similar to what happens when you `bpy.ops.bim.hotkey(hotkey="S_A")` from the type launcher. 

![Image](https://github.com/user-attachments/assets/1f77882e-e67a-49b7-8466-8afc961f5fd6)

falken10vdl:
Is just this what you are after? (invoking bpy.ops.bim.hotkey(hotkey="S_A") form a context menu)?

https://github.com/user-attachments/assets/075da525-7531-4969-9074-3f635c0cf7c5

Maybe I have not understood you  ;)

Cheers!

theoryshaw:
Yes, but if specifically, if you select a wall instance for example, and right click to `Create Instance`, then it acts in the same way if the user selected the type of that instance in 'Type Manager' and invoked `bpy.ops.bim.hotkey(hotkey="S_A")`.  Hope that make sense.

falken10vdl:
@theoryshaw  Something like this?

https://github.com/user-attachments/assets/e853c3dc-c29b-4c6b-b8ef-bfa7fa84a63a

In the Multiobject bim tool we add the right click context menu and if one object is selected then we create an instance of the same type.

Cheers!

theoryshaw:
Yes!  Can you have it active regardless of whether the multiobject bim tool is active or not?

falken10vdl:
Well if the multiobject is not active if you are in the right tool then it will work as the hotkey S_A will behave as expected
The problem is if you try to instantiate an object but you are in the wrong tool...
Let's say you select a door and do rightclick to instantiate a new one with he same type but you are in the bim.wall_tool. The ifc_classes available to the model at that state will be missing the relating_type of the selected object.
In this case what shall we do?
We could switch to the multiobject tool where we are sure that all ifc_classes are available.
What do you think?

falken10vdl:
Please take a look to https://github.com/IfcOpenShell/IfcOpenShell/pull/6830 I have not changed the tool as it could be disturbing for the end user. I just "try" and if there is error I ask the user to switch to either multiobject or the adequate tool.
Cheers!

theoryshaw:
Pure awesome... thanks!

steverugi:
sorry, what does "S_A" hotkey mean? thanks
[...]

PR #6830, opened 2025-06-22
Create instance context menu

falken10vdl:
This is a proposal to address
https://github.com/IfcOpenShell/IfcOpenShell/issues/6823
Cheers!


Andrej730:
Is this intended change? 

image

It makes search appear now for enums that have just couple elements and it's quicker to just select them from the drop-down.

image

falken10vdl:
No. You are right. I changed it while testing. If you think there should be such threshold it must be changed back to the value you decide.
Sorry.
Cheers.

PR #6831, opened 2025-06-22
Named Linked Aggregates

theoryshaw:
Hi @brunoperdigao, was wondering if you could take a look at this and let me know if it's a good approach. 

I'm basically trying to name the linked aggregates, so it can ultimately be used in an assigned IfcAnnotation tag.

### DIFF
--- a/src/bonsai/bonsai/bim/module/geometry/operator.py
+++ b/src/bonsai/bonsai/bim/module/geometry/operator.py
@@ -1161,10 +1161,10 @@ class DuplicateMoveLinkedAggregate(bpy.types.Operator):
         return OverrideDuplicateMove.execute_duplicate_operator(self, context, linked=False)
 
     def _execute(self, context):
-        return DuplicateMoveLinkedAggregate.execute_ifc_duplicate_linked_aggregate_operator(self, context)
+        return DuplicateMoveLinkedAggregate.execute_ifc_duplicate_linked_aggregate_operator(self, context, aggregate_name)
 
     @staticmethod
-    def execute_ifc_duplicate_linked_aggregate_operator(self, context, location_from_3d_cursor=False):
+    def execute_ifc_duplicate_linked_aggregate_operator(self, context, aggregate_name, location_from_3d_cursor=False):
         ifc_file = tool.Ifc.get()
         self.new_active_obj = None
         self.group_name = "BBIM_Linked_Aggregate"
@@ -1199,7 +1199,10 @@ class DuplicateMoveLinkedAggregate(bpy.types.Operator):
                 ifcopenshell.api.pset.edit_pset(
                     ifc_file,
                     pset=pset,
-                    properties={"Index": index},
+                    properties={
+                        "Index": index,
+                        "Name": "Default_Name",
+                    },
                 )
 
                 index += 1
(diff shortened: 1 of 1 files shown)


src/bonsai/bonsai/bim/module/geometry/operator.py
```diff
@@ -1485,13 +1491,75 @@ def set_new_matrix(
...
+                if not part_pset:
+                    # Create the pset if it doesn't exist
+                    part_pset = ifcopenshell.api.pset.add_pset(
+                        tool.Ifc.get(), product=part, name="BBIM_Linked_Aggregate"
+                    )
+
+                props = ifcopenshell.util.element.get_pset(part, "BBIM_Linked_Aggregate")
+                index = props.get("Index") if props else None
+                if not index:
+                    index = 1
```
theoryshaw, line 1547: i know this is not right... still figuring. :)

brunoperdigao:
I don't know if I understood your goal correctly, but I believe it would be more reliable to create a way for the IfcAnnotation tag to capture the aggregate name based on a part object, rather than adding the name to each part. This approach would help avoid duplicating information, which can lead to inconsistencies.

However, I am not familiar with how the tags work in IfcAnnotation. I will take a closer look at it soon.

theoryshaw:
Yes, you're probably right.  This seemed wasteful. :)

But yes, basically trying to find a way to name the linked aggregate so i can tag it. 

Currently, if we're using the aggregate's Name attribute each aggregate would be named like this...

- `Roof Truss_01`
- `Roof Truss_02`
- `Roof Truss_03`

.. which is not ideal to tag.  

Would like to just tag with `Roof Truss`, if possible.  And an easy way to change the name, if required. 

Any thoughts on that direction?

brunoperdigao:
@theoryshaw  I tried a similar approach by adding a field called "Annotation_Tag" in the linked aggregate Pset to be used in annotations. Its is added just to the aggregate and not to its parts. If you change the tag it will change the name of the aggregates after refreshing, but will keep the original number. In the print below the first annotation has the value of `{{parent.Name}}` while the second has `{{parent.BBIM_Linked_Aggregate.Annontation_Tag}}`.
image


There is still another problem to solve. When we refresh the linked aggregate, it means that some objects are deleted and recreated, which breaks the link with the annotation "Product Assingment". I will work on that.

theoryshaw:
Hi @brunoperdigao 

In the IFC schema, there doesn't seem to a lot of precedent, for creating an 'annotation tag' parameter. 

Why not just do something like `{{parent.BBIM_Linked_Aggregate.Name}}` and then just tag the `Name` parameter.  In the future, you could see adding more 'stuff' that you'd want to tag as well, like `{{parent.BBIM_Linked_Aggregate.Description}}`, etc. 

Thoughts?

theoryshaw:
Closing 'this' PR, as it's too cumbersome, as you mention.

theoryshaw:
> When we refresh the linked aggregate, it means that some objects are deleted and recreated, which breaks the link with the annotation "Product Assingment". I will work on that.

yes, this would be awsome.

brunoperdigao:
> Why not just do something like {{parent.BBIM_Linked_Aggregate.Name}} and then just tag the Name parameter. In the future, you could see adding more 'stuff' that you'd want to tag as well, like {{parent.BBIM_Linked_Aggregate.Description}}, etc.

Yeah, you're right. I thought it might be confusing to have the element "name" in the attributes and another "name" in the pset. But the "name" in the pset is referring to the linked aggregates name, as a group, so they really are different things. I guess it makes sense.

Issue #6777, opened 2025-06-04
if `IfcAnnotation[PLAN_LEVEL]` was contained by a spatial container, that its elevation would be reflective of the z distance from that spatial container.


theoryshaw:
### Feature Description

Would be nice if `IfcAnnotation[PLAN_LEVEL]` was contained by a spatial container, that its elevation would be reflective of the z distance from the z coordinate of that spatial container. 

And if not contained by a spatial container the elevation is from the global origin.

falken10vdl:
Please take a look to https://github.com/IfcOpenShell/IfcOpenShell/pull/6845
Should we not signal somehow that it is a relative z value? I am no Architect so I do not know if there is some nomenclature to distinguish absolute from relative magnitudes :)
Cheers!

falken10vdl:
@theoryshaw I have seen you merged the PR. Are you hapoy to close this issue?

theoryshaw:
for sure!

PR #6845, opened 2025-06-25
Z-coordinate  in PlanLevelDecorator relative to spatial container if exists. Absolute otherwise

falken10vdl:
This addresses https://github.com/IfcOpenShell/IfcOpenShell/issues/6777
Cheers!


theoryshaw:
Awesome.. thanks.

Issue #6822, opened 2025-06-19
When searching via the Ifc class search, that is looked up PredefinedTypes as well.


theoryshaw:
### Feature Description

It would be nice, when searching via the Ifc class search, that is looked up PredefinedTypes as well. 

![Image](https://github.com/user-attachments/assets/91bb517d-0382-4687-b676-26d36f9c4ee1)

similar to how it does here...

![Image](https://github.com/user-attachments/assets/b9164879-842f-4f2c-aa3c-6474b59262d4)

falken10vdl:
Are you sure that is what you want?
In the multitool space you have BIMModelProperties

![Image](https://github.com/user-attachments/assets/1a2fa8b4-4f41-4470-8434-f7fb21086ab5)
So that will allow you to search within the classes defined in the model

In the Object Metadata search field since you can reassign IFC Classes it is defined as BIMrootProperties which provides the list of all possible classes

![Image](https://github.com/user-attachments/assets/38cd5d0c-1198-47ab-8ec8-dd4758251384)
So that will allow you to search all possible ifc classes.

Why would you like to search all possible ifc classes in the multitool if afterwards you can only use in the type manager the ones that are defined (or create a new one in the type maanger and then it will be part of the search items)?

Cheers!

theoryshaw:
Here's the scenario, i guess i'm trying to address...

- User would like to create or find a 'thing', but they're not sure what class it falls under.  For example, let's say they are trying to create/find a type for a `Furnace`. 
- Currently you can type in `Furnace` in the object metadata and it will suggest `IfcUnitaryEquipmentType`.
- It would be nice if multitool search worked in the same way.

falken10vdl:
@theoryshaw I have seen the issue is that  when the MultiObject Tool is selected, we have the Model properties ifc_class available and not the root properties ifc_class (as it is in the object metadata). A solution is to offer the root properties ifc_classes to the search and if the selected one is within the list of ifc_classes of the model then ask the user to sync the model ifc_class to the selected root ifc_class (if they are different).
something like this:

https://github.com/user-attachments/assets/e493aa81-73b0-4062-9e98-327347dc61b3

 Let me know what you think.
Cheers!

theoryshaw:
Yeah i guess that works, the UI is kinda chunky however, but open source, is open source. :)

falken10vdl:
I agree. I tried to assign the model property to the root property but that is not allowed while drawing...
[...]

PR #6846, opened 2025-06-25
Multiobject search updated

falken10vdl:
This is a proposal for https://github.com/IfcOpenShell/IfcOpenShell/issues/6822
Please note that the second commit is removing a button sync operator and modifies refresh_predefined_types.
This needs careful review by core developers to assess if that could have unwanted side effects.
Cheers!


falken10vdl:
Now it is a much simpler solution without touching root.py so it is safe.

falken10vdl:
Let me rebase this as it is a little bit old snd looks it has now conflicts to be resolved.
Cheers!

falken10vdl:
No need anymore for this one as we already have https://github.com/IfcOpenShell/IfcOpenShell/pull/6962 which covers this case and it is simpler :)

PR #6849, opened 2025-06-26
Draw attributes with search

### LINKED ISSUE #4325: Have a search function for `material.Category`, in order to look up and use existing categories in the project.
Would be nice to have a search function for `material.Category`, in order to look up and use existing categories in the project. 
image

I might be able to tackle this, if easy enough.

falken10vdl:
Generalization of issue https://github.com/IfcOpenShell/IfcOpenShell/issues/4325 to add search capabilities for the attributes drawn by bonsai.bim.helper.draw_attributes
Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/__init__.py
+++ b/src/bonsai/bonsai/bim/__init__.py
@@ -127,6 +127,7 @@ classes = [
     operator.ShowSystemInfo,
     prop.StrProperty,
     operator.BIM_OT_enum_property_search,  # /!\ Register AFTER prop.StrProperty
+    operator.BIM_OT_attribute_search_values,
     prop.ObjProperty,
     prop.MultipleFileSelect,
     prop.Attribute,
--- a/src/bonsai/bonsai/bim/helper.py
+++ b/src/bonsai/bonsai/bim/helper.py
@@ -116,6 +116,7 @@ def draw_attribute(
     if enable_search and attribute.data_type in ("string", "integer", "float"):
         op = row.operator("bim.attribute_search_values", text="", icon="VIEWZOOM")
         op.attribute_name = attribute.name
+        op.attribute_ifc_class = attribute.ifc_class
         op.data_path = attribute.path_from_id(value_name)
         op.data_type = attribute.data_type
 
--- a/src/bonsai/bonsai/bim/helper.py
+++ b/src/bonsai/bonsai/bim/helper.py
@@ -53,7 +53,6 @@ def draw_attributes(
     popup_active_attribute: Optional[bonsai.bim.prop.Attribute] = None,
     callback: Optional[Callable[[bonsai.bim.prop.Attribute, bpy.types.UILayout], None]] = None,
     enable_search: bool = False,
-    should_click_ok: bool = False,
 ) -> None:
     """Draw editable UI for prop.Attributes.
 
@@ -67,7 +66,7 @@ def draw_attributes(
         row = layout.row(align=True)
         if attribute == popup_active_attribute:
             row.activate_init = True
-        draw_attribute(attribute, row, copy_operator, enable_search=enable_search, should_click_ok=should_click_ok)
+        draw_attribute(attribute, row, copy_operator, enable_search=enable_search)
         if callback:
             callback(attribute, row)
 
@@ -77,10 +76,8 @@ def draw_attribute(
     layout: bpy.types.UILayout,
     copy_operator: Optional[str] = None,
     enable_search: bool = False,
-    should_click_ok: bool = False,
 ) -> None:
     value_name = attribute.get_value_name(display_only=True)
-
 
     if value_name == "enum_value":
         prop_with_search(layout, attribute, "enum_value", text=attribute.name)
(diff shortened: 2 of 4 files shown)


Andrej730:
I like the idea - it gives more power to the users, so they can establish their own patterns in attributes and effectively maintain them. It also can be used as a quick way to search through entities attributes.

PR review notes:
1. `bim.attribute_search_values` `invoke()`:
- why `.clear()` at the start? Isn't empty by default?
- `bim.attribute_search_values` `invoke()` - why filling `collection_values` to then clear it and fill again? It seems this can be simplified.
- `bim.attribute_search_values` `invoke()` - `natsorted` would be a good fit here.

2. `bim.attribute_search_values` `get_unique_attribute_values()` :
- is `try... except` necessary here? This would make errors go silently instead of handling them.
- checks for `ifc_file` and `ifc_class` seems unnecessary, this operator is probably won't be called if no IFC file is active and ifc_class will be always provided
- why checking `ifcopenshell.entity_instance` and `(list, tuple)` attributes if operator is intended to only support `("string", "integer", "float")`?

theoryshaw:
Works great from my side!
Thank you @Andrej730 for the review.

falken10vdl:
@Andrej730 thanks a lot for the feedback! Let me go over it. Maybe it would nake sense tp add also a filter option so the user can pass a list of the attibutes that will not get a search.
Cheers!

falken10vdl:
@Andrej730 I have implemented your recommendations. I have however kept the initialization in invoke:
    self.collection_values.clear()
    self.search_value = "" 
This is needed because if we have several drawn attributes each with its search method if we do not initialize the collection will be growing with info from other attributes mixing it.
I based the code in the existing bim.enum_property_search, but i have allowed to accept not found data so no need to loop in update_attribute_search_value either.

Thanks!

Andrej730:
@falken10vdl 

1. > This is needed because if we have several drawn attributes each with its search method if we do not initialize the collection will be growing with info from other attributes mixing it.

    Wouldn't `"SKIP_SAVE"` option cover it? https://docs.blender.org/api/current/bpy_types_enum_items/property_flag_items.html#rna-enum-property-flag-items


2. `string_values = [str(value) for value in unique_values if value is not None]` - `str(value)` and `is None` checks seem unnecessary since `get_unique_attribute_values` returns `list[str]` already.

3. `should_click_ok` prop seems unnecessary, it's just always `False`.

4. > I based the code in the existing bim.enum_property_search, but i have allowed to accept not found data so no need to loop in update_attribute_search_value either.

    Makes sense. Can you please add a comment that implementations are connected? Currently it's hard to guess what `if self.first_launch` - `context.window.screen = context.window.screen` block does - `bim.enum_property_search` doesn't have it documented either, but at least there will be some connection.

5. One thing I've found that's kind of confusing - when you open search, it resets current attribute value. I think it would be more intuitive if it would keep the value until it was changed and starting value in search popup would be copied from the current attribute value.

https://github.com/user-attachments/assets/38254b96-6a03-4385-a8da-6a328fb7c222

falken10vdl:
@Andrej730 

> 1. Wouldn't "SKIP_SAVE" option cover it? https://docs.blender.org/api/current/bpy_types_enum_items/property_flag_items.html#rna-enum-property-flag-items

Indeed! Changed to use that (I did not know there was that option... I copied from bim.enum_property_search... Thanks for tip!)

> 2. string_values = [str(value) for value in unique_values if value is not None] - str(value) and is None checks seem unnecessary since get_unique_attribute_values returns list[str] already.

Agree. Removed

>3.  should_click_ok prop seems unnecessary, it's just always False.

I copied it from bim.enum_property_search. I think it is a nice option to have but up to you. I have modified to use it as True for example in the material ui call to bonsai.bim.helper.draw_attributes.

>4.  Makes sense. Can you please add a comment that implementations are connected? Currently it's hard to guess what if self.first_launch - context.window.screen = context.window.screen block does - bim.enum_property_search doesn't have it documented either, but at least there will be some connection.

Done: `class BIM_OT_attribute_search_values(bpy.types.Operator):
    """Search for attribute values. This implementation is based on bim.enum_property_search"""`

>5.  One thing I've found that's kind of confusing - when you open search, it resets current attribute value. I think it would be more intuitive if it would keep the value until it was changed and starting value in search popup would be copied from the current attribute value.

Agree. I have changed the invoke call to copy the property value to the initial search field

Andrej730:
@falken10vdl 

1. > I copied it from bim.enum_property_search. I think it is a nice option to have but up to you. I have modified to use it as True for example in the material ui call to bonsai.bim.helper.draw_attributes.

`should_click_ok=False` is nice, it works a bit quicker, I liked it too. I meant that `should_click_ok` seems unnecessary as a prop and maybe we can keep only `should_click_ok=False` behaviour in `update_attribute_search_value`, until there will be need for `should_click_ok=True`?

2. Can you please move duplicated code below to some common method? E.g. a staticmethod on `BIM_OT_attribute_search_values`

```python
    path_parts = self.data_path.split(".")
    obj_path = ".".join(path_parts[:-1])
    attr_name = path_parts[-1]
    attribute = eval(f"bpy.context.scene.{obj_path}")
```

3. `str(value := getattr(attribute, attr_name, ""))` walrus assignment seems redundant as `value` is not used later.

4. black format is a bit off after the last change

falken10vdl:
@Andrej730 I have followed your comments above. Thanks!

Andrej730:
Looks good, thank you!

falken10vdl:
Thanks to you @Andrej730 for such valuable feedback!

PR #6860, opened 2025-06-29
Basic infrastructure for IfcCostSchedule PDF export with typst

carlopav:
Pdf export of IfcCostSchedule based on typst library. The schedule is first exported in CSV into a temporary directory with Ifc5DCsvWriter. Then the template file (typst_template_ifc_cost_schedule.typ) is copied into the same directory. Then the PDF is compiled and saved into the user defined output file.
Actually supports PRICEDBILLOFQUANTITIES and SCHEDULEOFRATES.

immagine

Here you have an example of the output:
[PRICEDBILLOFQUANTITIES.pdf](https://github.com/user-attachments/files/20967535/PRICEDBILLOFQUANTITIES.pdf)

### DIFF
--- a/src/bonsai/bonsai/bim/module/cost/__init__.py
+++ b/src/bonsai/bonsai/bim/module/cost/__init__.py
@@ -59,6 +59,7 @@ classes = (
     operator.ExpandCostItemRate,
     operator.ExpandCostItems,
     operator.ExportCostSchedules,
+    operator.ExportCostSchedulesToPDF,
     operator.HighlightProductCostItem,
     operator.ImportCostScheduleCsv,
     operator.RefreshCostScheduleCsv,
--- a/src/bonsai/bonsai/bim/module/cost/operator.py
+++ b/src/bonsai/bonsai/bim/module/cost/operator.py
@@ -837,26 +837,18 @@ class ExportCostSchedulesToPDF(bpy.types.Operator, ExportHelper):
     def invoke(self, context, event):
         global cost_schedules_items
         cost_schedules_items.clear()
-
         file = tool.Ifc.get()
         schedules = file.by_type("IfcCostSchedule")
-        counter = 0
         for schedule in schedules:
-            schedule_name = getattr(schedule, "Name", "Unnamed")
-            if schedule_name is None:
-                schedule_name = "Unnamed"
-            schedule_type = getattr(schedule, "PredefinedType", "UNTYPED")
-            if schedule_type is None:
-                schedule_type = "UNTYPED"
-            cost_schedules_items.append((str(counter), "{} ({})".format(schedule_name, schedule_type), ""))
-            counter += 1
-
+            schedule_name = getattr(schedule, "Name", None) or "Unnamed"
+            schedule_type = getattr(schedule, "PredefinedType", None) or "UNTYPED"
+            cost_schedules_items.append((str(schedule.id()), "{} ({})".format(schedule_name, schedule_type), ""))
         return ExportHelper.invoke(self, context, event)
 
     def execute(self, context):
         file = tool.Ifc.get()
         self.props = tool.Cost.get_cost_props()
-        cost_schedule = file.by_type("IfcCostSchedule")[int(self.cost_schedules_enum)]
+        cost_schedule = file.by_id(int(self.cost_schedules_enum))
         options = {"should_print_summary": self.should_print_summary}
         core.export_cost_schedules_to_pdf(
             tool.Cost, filepath=self.filepath, cost_schedule=cost_schedule, options=options
--- a/src/bonsai/bonsai/bim/module/cost/operator.py
+++ b/src/bonsai/bonsai/bim/module/cost/operator.py
@@ -791,14 +791,20 @@ class ExportCostSchedulesToPDF(bpy.types.Operator, ExportHelper):
     bl_options = {"REGISTER", "UNDO"}
     bl_description = "Print chosen cost schedule to pdf.\nMake sure typst is available,\neventually installing Typst Importer extension."
     filename_ext = ".pdf"
-    filter_glob: bpy.props.StringProperty(default="*.pdf", options={'HIDDEN'}, maxlen=255)
+    filter_glob: bpy.props.StringProperty(default="*.pdf", options={"HIDDEN"}, maxlen=255)
     chosen_schedule: bpy.props.EnumProperty(
         name="",
         description="Choose IfcCostSchedule to print",
-        items=[('0','Default','Export first schedule in the IfcProject',),],
-        default='0',
+        items=[
+            (
+                "0",
+                "Default",
+                "Export first schedule in the IfcProject",
+            ),
+        ],
+        default="0",
     )
-    '''should_print_cover: bpy.props.BoolProperty(
+    """should_print_cover: bpy.props.BoolProperty(
         name="Should print document cover",
         description="Create a cover page with project data",
         default=False,
@@ -817,7 +823,7 @@ class ExportCostSchedulesToPDF(bpy.types.Operator, ExportHelper):
         name="Should print rates and totals",
         description="Print rates and totals for each voice",
         default=True,
-    )'''
+    )"""
     should_print_summary: bpy.props.BoolProperty(
         name="Should print summary",
         description="Print summary at the end of the document",
@@ -837,31 +843,29 @@ class ExportCostSchedulesToPDF(bpy.types.Operator, ExportHelper):
         # box.prop(self, "should_print_description")
         # box.prop(self, "should_print_each_quantity")
         box.prop(self, "should_print_summary")
-   
+
     @classmethod
     def poll(cls, context):
         try:
             import typst
+
             return True
         except:
             return False
-        
+
     def execute(self, context):
         file = tool.Ifc.get()
         self.props = tool.Cost.get_cost_props()
         cost_schedule = file.by_type("IfcCostSchedule")[int(self.props.cost_schedules_enum)]
         options = {
-            #"should_print_cover" : self.should_print_cover,
-            #"should_print_description" : self.should_print_description,
-            #"should_print_each_quantity" : self.should_print_each_quantity,
-            #"should_print_rates" : self.should_print_rates,
-            "should_print_summary" : self.should_print_summary
+            # "should_print_cover" : self.should_print_cover,
+            # "should_print_description" : self.should_print_description,
+            # "should_print_each_quantity" : self.should_print_each_quantity,
+            # "should_print_rates" : self.should_print_rates,
+            "should_print_summary": self.should_print_summary
         }
         r = core.export_cost_schedules_to_pdf(
-            tool.Cost, 
-            filepath=self.filepath, 
-            cost_schedule=cost_schedule,
-            options=options
+            tool.Cost, filepath=self.filepath, cost_schedule=cost_schedule, options=options
         )
         return {"FINISHED"}
 
--- a/src/bonsai/bonsai/bim/module/cost/operator.py
+++ b/src/bonsai/bonsai/bim/module/cost/operator.py
@@ -789,41 +789,10 @@ class ExportCostSchedulesToPDF(bpy.types.Operator, ExportHelper):
     bl_idname = "bim.export_cost_schedules_to_pdf"
     bl_label = "Export Cost Schedule to PDF"
     bl_options = {"REGISTER", "UNDO"}
-    bl_description = "Print chosen cost schedule to pdf.\nMake sure typst is available,\neventually installing Typst Importer extension."
+    bl_description = "Print chosen cost schedule to pdf."
     filename_ext = ".pdf"
     filter_glob: bpy.props.StringProperty(default="*.pdf", options={"HIDDEN"}, maxlen=255)
-    chosen_schedule: bpy.props.EnumProperty(
-        name="",
-        description="Choose IfcCostSchedule to print",
-        items=[
-            (
-                "0",
-                "Default",
-                "Export first schedule in the IfcProject",
-            ),
-        ],
-        default="0",
-    )
-    """should_print_cover: bpy.props.BoolProperty(
-        name="Should print document cover",
-        description="Create a cover page with project data",
-        default=False,
-    )
-    should_print_description: bpy.props.BoolProperty(
-        name="Should print description",
-        description="Export the full description if present",
-        default=True,
-    )
-    should_print_each_quantity: bpy.props.BoolProperty(
-        name="Should print each quantity",
-        description="Export the full list of quantities",
-        default=False,
-    )   
-    should_print_rates: bpy.props.BoolProperty(
-        name="Should print rates and totals",
-        description="Print rates and totals for each voice",
-        default=True,
-    )"""
+
     should_print_summary: bpy.props.BoolProperty(
         name="Should print summary",
         description="Print summary at the end of the document",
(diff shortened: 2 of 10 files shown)


maxfb87:
@carlopav in my opinion, very good! I didn't check the typist part because i don't know so much about it, but the else seems ok to me.
I leave the formal part (for example black formatting, typing, etc..) to more experience developers.
For me, we can merge it so everybody can test it and give feedback

Andrej730:
> For me, we can merge it so everybody can test it and give feedback

Couldn't we test it here to catch possible errors before merging it? 
PR can be applied with 
`curl -L https://github.com/IfcOpenShell/IfcOpenShell/pull/6860.patch | git am -3`
or 
`git pull https://github.com/carlopav/IfcOpenShell.git IfcCostScheduleToPDF` 
for tests.


@carlopav

Thanks for submitting the PR, pdf export sounds like a good quick way to export cost schedule to share it.

Code review notes:

1. `# box.prop(self, "should_print_cover")` and props in triple-quotes in `ExportCostSchedulesToPDF` - unused code?
2. `prop.py` - `get_cost_schedules_enum_items()`
    the problem with this method that it's going to access IFC very rapidly, on every UI draw call, which we usually want to avoid. E.g. it's going to run `file.by_type("IfcCostSchedule")` and access attributes of all schedules it was able to find on every draw call. 
    For this case we typically use `data.py` module - e.g. see similar enum items method `get_schedule_of_rates` in the same `prop.py` module. It's accessing data that's already cached in `CostItemRatesData`, instead of retrieving it from IFC each time.
    But another thing that those enum items are needed just for `ExportCostSchedulesToPDF` operator. Maybe we can fill those enum items once during this operator's `invoke`? Instead of creating a special property on the scene for this.
3. `chosen_schedule` - prop is unused
4. `ExportCostSchedulesToPDF.poll` - please add `poll_message_set`, so users will be able to figure why this feature is not available. Or alternatively, if `typst` is not found maybe it can still work and show a popup, describing the issue and providing `bim.pip_install` operator to install `typsts`. Otherwise most users will be lost seeing they need some kind of "typst extension".
5. `ExportCostSchedulesToPDF.execute` - saving results to `r` in `r = core.export_cost_schedules_to_pdf` seems unnecessary and kind of implies that there could be different results.
6. It seems black formatter didn't worked, there are still some issues - https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/15984077269/job/45084797105. You can run `black .` in the repository root folder and that should fix it.

    Ifc5DPdfWriter:
7.  `self.colours = self.default_colors.copy()` - `self.colours` added but never used
8. `cost_schedule: Optional[ifcopenshell.entity_instance] = None,` - perhaps type should not include `None`, since it's not supported (`self.cost_schedule.Name` in `write` is requiring cost schedule to be existing entity).
9. `check_options` - is there a need to check specific keys? Shouldn't it just replace all existing values with `"true" if value else "false"`? 
10. Can you please add `typst` as optional dependency to https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/ifc5d/pyproject.toml to keep it documented? See example https://github.com/IfcOpenShell/IfcOpenShell/blob/6c4e4e5f910a7959175f329cc159eb65f7b0f16a/src/ifcopenshell-python/pyproject.toml#L27-L30

    Tests:
11. Code assumes all cost schedules will be named and will have a predefinedtype, otherwise leads to errors (similar assumtption in `Ifc5DPdfWriter` code).
```python
nsai\bim\module\cost\prop.py", line 62, in get_cost_schedules_enum_items
    schedule.Name + " (" + schedule.PredefinedType + ")",
    ~~~~~~~~~~~~~~^~~~~~
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

12. Code is requiring a specific font and breaks if user doesn't have it.

```python
    pdf_bytes = typst.compile(typst_main_path)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: warning: unknown font family: liberation sans
    ┌─ \\?\C:\Users\Andrej\AppData\Local\Temp\tmp53emw0kq\typst_template_ifc_cost_schedule.typ:321:22
    │
321 │       #set text(font: "Liberation Sans", size: 9pt, lang: "en");
    │                       ^^^^^^^^^^^^^^^^^

error: dictionary does not contain key "BUDGET" and no default value was specified
    ┌─ \\?\C:\Users\Andrej\AppData\Local\Temp\tmp53emw0kq\typst_template_ifc_cost_schedule.typ:341:6
    │
341 │       format_table.at(schedule_type)
    │       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

13. Tried to replace `"Liberation Sans"` with `"Times New Roman"`, but noticed that code doesn't support some predefined types, though you can select them in operator's menu.

```python
    pdf_bytes = typst.compile(typst_main_path)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: error: dictionary does not contain key "BUDGET" and no default value was specified
    ┌─ \\?\C:\Users\Andrej\AppData\Local\Temp\tmphyg29e64\typst_template_ifc_cost_schedule.typ:341:6
    │
341 │       format_table.at(schedule_type)
    │       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Using `PRICEDBILLOFQUANTITIES` helped:

image

maxfb87:
@Andrej730 yeah, i wanted to merge just because more people could test it but you are right, it is better to find error before merging...so, i will try to test the PR thanks :-)

carlopav:
Hi @Andrej730 , thanks for the detailed review and sorry for the quantity of problems :)

> Thanks for submitting the PR, pdf export sounds like a good quick way to export cost schedule to share it.

Good! The aim is to provide a "professional output" for cost schedules in the common format that is used to deliver them (at least here in Italy).

>     1. `# box.prop(self, "should_print_cover")` and props in triple-quotes in `ExportCostSchedulesToPDF` - unused code?

Yes, I left it because i'll use that in the near future. Deleted.

>     2. `prop.py` - `get_cost_schedules_enum_items()`

**This is the only problem that I really don't know how to approach. I tried to setup the property in the invoke method, but I had no success... Any hint?**

>     3. `chosen_schedule` - prop is unused

Yes, sorry. Deleted.

>     4. `ExportCostSchedulesToPDF.poll` - please add `poll_message_set`, so users will be able to figure why this feature is not available. Or alternatively, if `typst` is not found maybe it can still work and show a popup, describing the issue and providing `bim.pip_install` operator to install `typsts`. Otherwise most users will be lost seeing they need some kind of "typst extension".

Done!

>     5. `ExportCostSchedulesToPDF.execute` - saving results to `r` in `r = core.export_cost_schedules_to_pdf` seems unnecessary and kind of implies that there could be different results.

Removed!

>     6. It seems black formatter didn't worked, there are still some issues - https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/15984077269/job/45084797105. You can run `black .` in the repository root folder and that should fix it.

I'm doing that now.

>     7. `self.colours = self.default_colors.copy()` - `self.colours` added but never used

Removed.

>     8. `cost_schedule: Optional[ifcopenshell.entity_instance] = None,` - perhaps type should not include `None`, since it's not supported (`self.cost_schedule.Name` in `write` is requiring cost schedule to be existing entity).

Updated "write" to handle the case cost_schedule.Name is None.

>     9. `check_options` - is there a need to check specific keys? Shouldn't it just replace all existing values with `"true" if value else "false"`?

Yes, you are right. Refactored and deleted check_options.

>     10. Can you please add `typst` as optional dependency to https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/ifc5d/pyproject.toml to keep it documented? 

Done!

>     11. Code assumes all cost schedules will be named and will have a predefinedtype, otherwise leads to errors (similar assumtption in `Ifc5DPdfWriter` code).

Updated. I do not think it's elegant, but it works now.

>     12. Code is requiring a specific font and breaks if user doesn't have it.

Provided a list of fallback fonts: ("Liberation Sans", "Roboto", "Arial", "Calibri"). Should I add more? Including a font would be the safer option, but I do not think we want to, is it?

>     13. Tried to replace `"Liberation Sans"` with `"Times New Roman"`, but noticed that code doesn't support some predefined types, though you can select them in operator's menu.
> Using `PRICEDBILLOFQUANTITIES` helped:

Oh, yes, the code "Actually supports PRICEDBILLOFQUANTITIES and SCHEDULEOFRATES."
I modified so every other PredefinedType is printed as a PRICEDBILLOFQUANTITIES, so the code does not return an error. I'll upgrade it soon to include other Types.

Andrej730:
> Yes, I left it because i'll use that in the near future. Deleted.

There's a bit unused code left - `# "should_print_cover" : self.should_print_cover,`, `# box.prop(self, "should_print_cover")`


>>     4. `ExportCostSchedulesToPDF.poll` - please add `poll_message_set`, so users will be able to figure why this feature is not available. Or alternatively, if `typst` is not found maybe it can still work and show a popup, describing the issue and providing `bim.pip_install` operator to install `typsts`. Otherwise most users will be lost seeing they need some kind of "typst extension".

> Done!

Can you please add something like this in poll message - "it can be installed from Quality and Control -> Debug and using 'typst' with Pip Install"? So non-advanced users would have at least some clue on how they can try the feature.

>>     8. `cost_schedule: Optional[ifcopenshell.entity_instance] = None,` - perhaps type should not include `None`, since it's not supported (`self.cost_schedule.Name` in `write` is requiring cost schedule to be existing entity).

> Updated "write" to handle the case cost_schedule.Name is None.

It still won't work if `cost_schedule` is `None` - `getattr(self.cost_schedule, "Name")` will fail with something like `None doesn't have 'Name' attribute` since it does exactly the same as `self.cost_schedule.Name`.

> >     11. Code assumes all cost schedules will be named and will have a predefinedtype, otherwise leads to errors (similar assumtption in `Ifc5DPdfWriter` code).
> 
> Updated. I do not think it's elegant, but it works now.

I think there's a confusion about `getattr(xxx, yyy)` - why not just use `schedule.Name`?

> Provided a list of fallback fonts: ("Liberation Sans", "Roboto", "Arial", "Calibri"). Should I add more? Including a font would be the safer option, but I do not think we want to, is it?

I think fallback fonts will do.

> >     2. `prop.py` - `get_cost_schedules_enum_items()`
> **This is the only problem that I really don't know how to approach. I tried to setup the property in the invoke method, but I had no success... Any hint?**

I think you've started implementing it originally by adding `chosen_schedule` on operator and this is how it indeed could be implemented:
- some enum property on `ExportCostSchedulesToPDF` with dynamic `items` callback
- dynamic items callback is just returning some list of enum items added to `ExportCostSchedulesToPDF`
- when user is using the operator, on `invoke` we clear the list and fill it using the code that's currently in `get_cost_schedules_enum_items`

carlopav:
> There's a bit unused code left - `# "should_print_cover" : self.should_print_cover,`, `# box.prop(self, "should_print_cover")`

Oh, sorry you are right. Deleted.

> 4. Can you please add something like this in poll message - "it can be installed from Quality and Control -> Debug and using 'typst' with Pip Install"? So non-advanced users would have at least some clue on how they can try the feature.

Nice, i didn't know that existed... very useful indeed!

> > > 8. `cost_schedule: Optional[ifcopenshell.entity_instance] = None,` - perhaps type should not include `None`, since it's not supported (`self.cost_schedule.Name` in `write` is requiring cost schedule to be existing entity).
> It still won't work if `cost_schedule` is `None` - `getattr(self.cost_schedule, "Name")` will fail with something like `None doesn't have 'Name' attribute` since it does exactly the same as `self.cost_schedule.Name`.
> I think there's a confusion about `getattr(xxx, yyy)` - why not just use `schedule.Name`?

Tried to provide a default for getattr(), perhaps there is better way?

> > > 2. `prop.py` - `get_cost_schedules_enum_items()`

I followed your advice and with a bit of AI i came up with a working solution. Please check if it could be good for you :)
Heads up:` cost_schedules_items = []` and `def get_cost_schedules_enum_items()` are in `operator.py`, is that the right place? Should we comment them somehow?

Thanks a lot @Andrej730 for your help!

PS Maybe when you think it's ready to merge we can squash the commits to one.

Andrej730:
> Tried to provide a default for getattr(), perhaps there is better way?

I think there is still confusion about what `geattr` does.

For example `getattr(self.file.by_type("IfcProject")[0], "Name", "Unnamed Project")` will never return `"Unnamed Project"` - this value will be only returned if `self.file.by_type("IfcProject")[0]` had `"Name"` attribute. 

And since IfcProject always has this attribute (https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcProject.htm), then default value will never be used.

Also `getattr` has no check for whether `self.file.by_type("IfcProject")[0]` is `None` - https://docs.python.org/3/library/functions.html#getattr

Regarding IfcProject there is no need to check whether it has a name - `Name` is required for IfcProject to be valid https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcProject.htm#5.1.3.11.4-Formal-propositions

Regarding cost schedules you can simplify it with `xxx = getattr(self.cost_schedule, "Name", None)` and `if xxx is None: xxx = "default value"`. Or with something like `xxx = getattr(self.cost_schedule, "Name", None) or "default value"`. Just to keep default value listed just once.

> I followed your advice and with a bit of AI i came up with a working solution. Please check if it could be good for you :)
> Heads up:` cost_schedules_items = []` and `def get_cost_schedules_enum_items()` are in `operator.py`, is that the right place? Should we comment them somehow?

I think we almost there. A few things I've noticed:
- we typically use `schedule.id()` instead of `counter` for enum value (first element of enum items tuple). That allows later to just do `ifc_file.by_id(int(self.enum))` to get the schedule to process
- can you please move `cost_schedules_items` and `get_cost_schedules_enum_items` to `ExportCostSchedulesToPDF` to encapsulate them on this class?

> PS Maybe when you think it's ready to merge we can squash the commits to one.

Sure, we definitely will to keep it clean 👍

carlopav:
> can you please move `cost_schedules_items` and `get_cost_schedules_enum_items` to `ExportCostSchedulesToPDF` to encapsulate them on this class?

Hello @Andrej730 I left them outside because when I incorporate them I cannot make it work... Perhaps i'm missing something about instancing and how the functions are called... I struggled again now, but  I'm definitely not able to solve this... sorry...

Andrej730:
> I left them outside because when I incorporate them I cannot make it work... Perhaps i'm missing something about instancing and how the functions are called... I struggled again now, but I'm definitely not able to solve this... sorry...

Any specific errors? 
Probably the issue is that `cost_schedules_items` is won't be available as `self.cost_schedules_items` from `get_cost_schedules_enum_items`, but you can access it from the class as `ExportCostSchedulesToPDF.cost_schedules_items`. Similarly, you'll need to use `ExportCostSchedulesToPDF.cost_schedules_items` in `invoke()`.


---

A thing about this 9c113b2989f38cfbaee36c347ac368473f654434 commit - `gettattr` is not needed in `invoke()` as IfcCostSchedule will always have `Name` and `PredefinedType` attributes

carlopav:
@Andrej730 that was it! Thanks, I think i had tried everything but this solution :)

Andrej730:
@carlopav 
```python
(str(schedule.id()), "{} ({})".format(schedule.Name, schedule.PredefinedType), "")
```

There still should be some kind handling for `None` values here as IfcCostSchedule.Name and PredefinedType can be Nones and otherwise it will appear as "None (None)`.

carlopav:
@Andrej730 Sure! Done :)

Andrej730:
Merged, thank you.

carlopav:
Thanks @Andrej730 for the invaluable support :)

Issue #6570, opened 2025-04-18
Improving ifc5d


maxfb87:
### Feature Description

I collected some features that can improve ifc5d (thanks to @steverugi) 

- [x] link an external .csv with the imported IfcCostSchedule and store the filepath in order to be able to refresh it "on the fly" (create a new refresh button) (testing) (store the .csv file as a document) (testing)
- [x] add the possibility to assign a SoR cost item also to another SoR cost item (change column Rate in Rate ID) (testing)
- [x] add the possibility to export as a pdf as it is displayed in the panel (look at this thread https://community.osarch.org/discussion/2963/small-script-to-print-pricedbillofquantities-to-pdf) (testing) done thanks to @carlopav in this commit https://github.com/IfcOpenShell/IfcOpenShell/commit/5f5c0b41d516dc503cd5bb71b4df1b3354620bc5 and this commit https://github.com/IfcOpenShell/IfcOpenShell/commit/dcf8329b04087f4a94b38495810bfaeba68a71ca
- [x] import from a csv a BoQ and a Sor connected together, maybe with 2 new columns called "RateSchedule" and "RateID"
- [ ] allow quantities also for SoR (in order to allow the possibility to create a cost rate analysis)
- [ ] when import a BoQ from the project library panel, import also the assigned cost item and have the possibility to import also only a cost item
- [ ] possibility to copy a cost item from two different cost schedules
- [ ] possibility to specify a formula in the manual quantification tab (for example 2xNetSideArea for wall painting) (related to #5846 #2397) and maybe save the formula in the formula field of an quantity (look for example here https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/schema/ifcquantityresource/lexical/ifcquantityarea.htm)
- [ ] improve the "export spreadsheet" in Bonsai (related to #6251 #5736 ). Also, unit column is not exported
- [ ] change columns name from "ID", "Name", "Quantity", "Value", Total Cost", "Rate" in
"ID",  "RateID", "Description", "Quantity", "Rate", "Amount" on order to be more adherent with practical usage
- [ ] add the possibility to divide the cost schedule in different ways, maybe using the cost item classification
- [ ] add a column in the .csv import dedicated for the classification of the cost item
- [x] export to csv also the single quantities that create the final quantity, also with a relation to the element that has generated it. Maybe with a serialized json list.
- [x] add the column Description to the exported csv
[...]

PR #6870, opened 2025-07-03
Csv cost remove file link

falken10vdl:
See https://github.com/IfcOpenShell/IfcOpenShell/issues/6570
This is a followup to that issue. In this particular case a remove csv link button is added next to the refresh csv link button.
Cheers!


falken10vdl:
What is our next step in https://github.com/IfcOpenShell/IfcOpenShell/issues/6570 :) ?

PR #6884, opened 2025-07-06
Pdf ifc cost schedule export improvements

carlopav:
Here are a bunch of improvements to export IfcCostSchedules to PDF.
I took the chanche to improve the CSV export, including a colum to identify if the cost item is a Sum and the full list of quantities included in the cost.

### DIFF
--- a/src/ifc5d/ifc5d/ifc5Dspreadsheet.py
+++ b/src/ifc5d/ifc5d/ifc5Dspreadsheet.py
@@ -159,6 +159,7 @@ class IfcDataGetter:
             "Name": cost_item.Name,
             "Description": cost_item.Description,
             "Unit": unit,
+            "Quantities": IfcDataGetter.serialise_cost_quantities(file, cost_item),
             "Quantity": quantity_data["quantity"],
             "RateSubtotal": rate_subtotal,
             "TotalPrice": total_price,
@@ -250,6 +251,49 @@ class IfcDataGetter:
             "quantity": total_cost_quantity,
             "unit": unit,
         }
+    
+    @staticmethod
+    def get_quantity_value(quantity: ifcopenshell.entity_instance) -> float:
+        if hasattr(quantity, 'LengthValue'):
+            return quantity.LengthValue
+        elif hasattr(quantity, 'AreaValue'):
+            return quantity.AreaValue
+        elif hasattr(quantity, 'VolumeValue'):
+            return quantity.VolumeValue
+        elif hasattr(quantity, 'CountValue'):
+            return quantity.CountValue
+        elif hasattr(quantity, 'WeightValue'):
+            return quantity.WeightValue
+        elif hasattr(quantity, 'TimeValue'):
+            return quantity.TimeValue
+        elif hasattr(quantity, 'NominalValue'):
+            return quantity.NominalValue
+        else:
+            return 0.0
+        
+    @staticmethod
+    def serialise_cost_quantities(file: ifcopenshell.file, cost_item: ifcopenshell.entity_instance) -> str:
+        if not cost_item.is_a("IfcCostItem"):
+            return ""
+        if cost_item.CostQuantities is None:
+            return ""
+        string = "["
+        for quantity in cost_item.CostQuantities:
+            string += '["'
+            for rel in file.get_inverse(quantity):
+                    if rel.is_a('IfcPropertySet') or rel.is_a('IfcElementQuantity'):
+                        prop_set = rel
+                        # Find elements that have this property set
+                        for prop_rel in file.get_inverse(prop_set):
+                            if prop_rel.is_a('IfcRelDefinesByProperties'):
+                                for obj in prop_rel.RelatedObjects:
+                                    if obj.is_a('IfcElement'):
+                                        string += obj.Name + " - "
+            string += quantity.Name + '", '
+            string += str(IfcDataGetter.get_quantity_value(quantity)) + "],"
+        string = string.removesuffix(",")
+        string += "]"
+        return string
 
 
 class SheetData(TypedDict):
@@ -325,6 +369,7 @@ class Ifc5Dwriter:
                 "Unit",
             ]
             if cost_schedule.PredefinedType != "SCHEDULEOFRATES":
+                headers.insert(-1, "Quantities")
                 headers.insert(-1, "Quantity")
             headers.extend(["RateSubtotal", "TotalPrice"])
 
--- a/src/ifc5d/ifc5d/ifc5Dspreadsheet.py
+++ b/src/ifc5d/ifc5d/ifc5Dspreadsheet.py
@@ -102,6 +102,20 @@ class IfcDataGetter:
         return values
 
     @staticmethod
+    def cost_item_is_a_sum(
+        cost_item: ifcopenshell.entity_instance
+    ) -> bool:
+        cost_values = []
+        if cost_item.is_a("IfcCostItem"):
+            cost_values = cost_item.CostValues
+        elif cost_item.is_a("IfcCostValue"):
+            cost_values = cost_item.Components
+        for cost_value in cost_values or []:
+            if cost_value.Category == "*":
+                return True
+        return False
+
+    @staticmethod
     def get_cost_items_data(
         file: ifcopenshell.file,
         cost_item: ifcopenshell.entity_instance,
@@ -138,6 +152,7 @@ class IfcDataGetter:
 
         data: CostItem = {
             "Index": index,
+            "Sum": int(IfcDataGetter.cost_item_is_a_sum(cost_item)),
             "Hierarchy": hierarchy,
             "Id": cost_item.id(),
             "Identification": cost_item.Identification,
@@ -301,6 +316,7 @@ class Ifc5Dwriter:
             cost_items = IfcDataGetter.get_schedule_cost_items_data(self.file, cost_schedule)
             headers: list[str] = [
                 "Id",
+                "Sum",
                 "Hierarchy",
                 "Index",
                 "Identification",
--- a/src/ifc5d/ifc5d/ifc5Dspreadsheet.py
+++ b/src/ifc5d/ifc5d/ifc5Dspreadsheet.py
@@ -36,10 +36,13 @@ class CostItem(TypedDict):
     # Exported columns.
     Index: int
     Hierarchy: str
+    ItemIsASum: int
     Id: int
     Identification: Union[str, None]
     Name: Union[str, None]
+    Description: Union[str, None]
     Unit: str
+    Quantities: str
     Quantity: Union[float, None]
     RateSubtotal: float
     TotalPrice: float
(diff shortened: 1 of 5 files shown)


Andrej730:
Hi! Looks nice, it's getting more sophisticated.

Notes reading the code:

1. Missed reading the previous PR but `CostItem` typed dictionary definition is now out of sync with how it's used. `Description` and now also `Quantities` and `Sum` attributes should be documented there.

2. Could this be simplified to some dictionary of all available options and thier default values?

```python
            typst_main_content += "  parent_to_new_page_up_to_depth: {},".format(
                str(self.options.get("parent_to_new_page_up_to_depth", 0))
            )
            typst_main_content += "  show_only_parents: {},".format(
                str(self.options.get("show_only_parents", False)).lower()
            )
            typst_main_content += "  should_print_cover: {},".format(
                str(self.options.get("should_print_cover", False)).lower()
            )
            typst_main_content += "  should_print_cost_ids: {},".format(
                str(self.options.get("should_print_cost_ids", False)).lower()
            )
            typst_main_content += "  should_print_description: {},".format(
                str(self.options.get("should_print_description", False)).lower()
            )
            typst_main_content += "  should_print_each_quantity: {},".format(
                str(self.options.get("should_print_each_quantity", False)).lower()
            )
            typst_main_content += "  should_print_each_cost_value: {},".format(
                str(self.options.get("should_print_each_cost_value", False)).lower()
            )
            typst_main_content += "  should_print_rates: {},".format(
                str(self.options.get("should_print_rates", False)).lower()
            )
            typst_main_content += "  should_print_summary: {},".format(
                str(self.options.get("should_print_summary", False)).lower()
```

3. Not removed print statement: `print(int(self.options.get("force_schedule_type")))`

4. Can you please rewrite this without `try-except` and check if monetary unit is present in the project, to make it more explicit.
```
try:
    project_currency = self.file.by_type("IfcMonetaryUnit")[0].Currency
except:
    project_currency = ""
```

5. `get_quantity_value` - we usually just do `[3]`, see below. And not sure what `NominalValue` does here, since `CostQuantities` are always `IfcPhysicalQuantity`.
 
https://github.com/IfcOpenShell/IfcOpenShell/blob/525ef2b77b58405fda1ad15d5df09f08cf490cee/src/ifcopenshell-python/ifcopenshell/util/element.py#L299-L301

carlopav:
Thanks @Andrej730! Very valuable feedback.
Don't know about point 5, it was created by Ai and I forgot to check what was it. I changed as you suggested... Perhaps the code could fail if the `quantity.is_a("IfcPhysicalComplexQuantity") `? 
Perhaps we should wait a feedback also from @maxfb87

Andrej730:
@carlopav 

1.

> 1. Missed reading the previous PR but `CostItem` typed dictionary definition is now out of sync with how it's used. `Description` and now also `Quantities` and `Sum` attributes should be documented there.

`Description` - currently defined as `str`, but it's actually optional and can be `None`. See https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcCostItem.htm

`Sum` - I'm probably missing something, but why `Sum` needs to be converted to `int` from `bool`?

---

2. 
```python
        if self.force_schedule_type == "0":
            forced_schedule_type = ""
        if self.force_schedule_type == "1":
            forced_schedule_type = "PRICEDBILLOFQUANTITIES"
        if self.force_schedule_type == "2":
            forced_schedule_type = "SCHEDULEOFRATES"
```
Why not store `PRICEDBILLOFQUANTITIES`, `SCHEDULEOFRATES` values as enum keys instead of remapping them later?


3.
Error in if-statement - `"SCHEDULEOFRATES"` will never be checked. This check basically acts like `if xxx == yyy or True`
```python
elif self.force_schedule_type == "PRICEDBILLOFQUANTITIES" or "SCHEDULEOFRATES"
```

4.

```python
project_currency = self.file.by_type("IfcMonetaryUnit")[0].Currency
```

Please reuse `project_monetary_unit` value.

5.

> Don't know about point 5, it was created by Ai and I forgot to check what was it. I changed as you suggested... Perhaps the code could fail if the `quantity.is_a("IfcPhysicalComplexQuantity") `?

I think we can skip complex quantities entirely for now, if we don't have a concrete examples and the way we want to implement them, and just print a warning in console, so it won't pass silently. Or add them with some 0 value and print a warning, so they will be present at least in some way.

6.
```python
            if quantity.is_a("IfcPhysicalSimpleQuantity"):
                string += str(quantity[3]) + "],"
```

No idea what the final serializion should look like for .typ, but wouldn't this break if it meets a complex quantity? Since then it would open a square bracket but would never close it.

carlopav:
> `Description` - currently defined as `str`, but it's actually optional and can be `None`. See https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcCostItem.htm

Ok! I'll use `Union[str, None]` as for the other optional keys.

> `Sum` - I'm probably missing something, but why `Sum` needs to be converted to `int` from `bool`?

I was thinking to shrink the text file ad "0"/"1" was shorter than "True"/"False" :) But I could have made the wrong assumption :)

> Why not store `PRICEDBILLOFQUANTITIES`, `SCHEDULEOFRATES` values as enum keys instead of remapping them later?

Yes, sure :) Done!

> Error in if-statement - `"SCHEDULEOFRATES"` will never be checked. This check basically acts like `if xxx == yyy or True`

Oh, sure :) Done!

> Please reuse `project_monetary_unit` value.

Sorry, sure! Done!

> No idea what the final serializion should look like for .typ, but wouldn't this break if it meets a complex quantity? Since then it would open a square bracket but would never close it.

Yes, you are right, I added an `else `statement so the user knows that only `IfcPhysicalSimpleQuantity` is supported.

Andrej730:
@carlopav 

> I was thinking to shrink the text file ad "0"/"1" was shorter than "True"/"False" :) But I could have made the wrong assumption :)

Okay, then it probably needs a better name - currently because it's a number and the name is "Sum", it looks like it's a sum of some kind values, not a boolean flag.


> > Why not store `PRICEDBILLOFQUANTITIES`, `SCHEDULEOFRATES` values as enum keys instead of remapping them later?
> 
> Yes, sure :) Done!

`OFF` value is currently never handled and in case of `OFF` `Ifc5DPdfWriter.write` will always fallback to `PRICEDBILLOFQUANTITIES` instead.


> Ok! I'll use `Union[str, None]` as for the other optional keys.

But why `Quantities` are now `None` also?

carlopav:
@Andrej730 
> Okay, then it probably needs a better name - currently because it's a number and the name is "Sum", it looks like it's a sum of some kind values, not a boolean flag.

Would "ItemIsASum" work?

> `OFF` value is currently never handled and in case of `OFF` `Ifc5DPdfWriter.write` will always fallback to `PRICEDBILLOFQUANTITIES` instead.

Code refactored!

> But why `Quantities` are now `None` also?

Right, Quantities are always a string. I'll change to `str`.

Andrej730:
@carlopav 

> Would "ItemIsASum" work?

Looks fine.



> > `OFF` value is currently never handled and in case of `OFF` `Ifc5DPdfWriter.write` will always fallback to `PRICEDBILLOFQUANTITIES` instead.
> 
> Code refactored!


Kind of odd that `force_schedule_type` is optional in general, but if user would leave it unfilled they'll get `ValueError`. 
Also, there's no need to check it for being `None` (`elif self.force_schedule_type is None`) if it can be only `str` - `force_schedule_type: str = "",`

carlopav:
@Andrej730 I think you are right...
Perhaps `cost_schedule` should not be optional...

Andrej730:
@carlopav 

> Perhaps `cost_schedule` should not be optional...

Then there's no need for `getattr` in `write()`, since cost schedule will always have Name and Predefined Type.

carlopav:
Good! Thanks @Andrej730 for the detailed comments, i'm learning a lot!

src/ifc5d/ifc5d/ifc5Dspreadsheet.py
```diff
@@ -101,6 +104,18 @@ def get_cost_item_values(cost_item: Union[ifcopenshell.entity_instance, None]) -
             )
         return values
 
+    @staticmethod
+    def cost_item_is_a_sum(cost_item: ifcopenshell.entity_instance) -> bool:
+        cost_values = []
+        if cost_item.is_a("IfcCostItem"):
+            cost_values = cost_item.CostValues
+        elif cost_item.is_a("IfcCostValue"):
```
maxfb87: @carlopav maybe i'm missing something, but i don't understand this: can this function be used also with non cost item as input (in this case, cost value)? If i understand it correctly, this function checks if a cost item is a sum or not, based on the category condition. Is it possible that this function is used with cost values as input?

carlopav: well... i copied it from ifc2json.py, row 83. I thought it was perhaps something useful to keep both possibilities so I didn't get rid of it.

maxfb87: My opinion is to remove that if it is not necessary, but you can just leave as it is if you want :-)

Andrej730: Good catch, agree. it can give a false impression that code that's using it can somehow provide non-ifccostitems as inputs.

carlopav: Done :)

Andrej730: But then there's also no need for `cost_item.is_a("IfcCostItem")` check.

carlopav: Sure :) My bad!

src/ifc5d/ifc5d/ifc5Dspreadsheet.py
```diff
@@ -138,12 +148,14 @@ def get_cost_items_data(
 
         data: CostItem = {
             "Index": index,
+            "ItemIsASum": int(IfcDataGetter.cost_item_is_a_sum(cost_item)),
```
maxfb87: @carlopav why it is necessary to transform the variable as a int? Couldn't it be just a boolean (True or False)?

carlopav: @maxfb87 see previous comments with @Andrej730. Was to keep It short. No problem if you both prefer to change It. Let me know.

maxfb87: I think it would be more correct

carlopav: Ok, "True" as in python or "true" as in typst? Any preference?

maxfb87: True as in Python

maxfb87:
Nice, for me it can be merged :-)

carlopav:
Updated commits to clean the revision history

carlopav:
Thanks @Andrej730 and @maxfb87 for the reviews and precious help! :)

Issue #6919, opened 2025-07-18
keep custom properties when converting blender object to ifc object


epeter-:
### Feature Description

converting existing blender objects (meshes) with custom properties to i.e. IfcBuildingElementProxies drops all existing custom properties when saving as ifc file.
keeping those properties would reduce the effort to create pset/property from existing data

sboddy:
Just a quick link to the discussion where you asked about this, as it gives a bit more context: https://github.com/IfcOpenShell/IfcOpenShell/discussions/6918

sboddy:
So there's an optional, tested version in the pull request. It works for simple properties (int, float, bool, string). Array, datablock, and python custom properties are ignored.

sboddy:
Hmmm. I never thought about it, but just tested this, and it works.
You can create a Pset template named BBIM_ImportedBlenderProps and match the names to your custom properties, and this allows you to set the Ifc data type for each property, i.e. length is an IfcLengthMeasure.

epeter-:
@sboddy thank you!

epeter-:
@sboddy  still struggeling!
i wanted to do a basic python solution and then enhance it with ui. for the moment i wanted to take all custom properties and asume they will be string.
i started from there: 
[](https://wiki.osarch.org/index.php?title=BlenderBIM_Add-on/BlenderBIM_Add-on_code_examples)

my code looks like that:
```
import bpy
import bonsai.tool as tool

for obj in bpy.context.selected_objects:  # some ifcbuildingelementproxies selected
    ifc_element = tool.Ifc.get_entity(obj)
    pset = ifc_element.BIMObjectProperties.psets.add() # also tried obj.BIMObjectProperties.psets.add()
    pset.name="MyPset"
    prop = pset.properties.add()
    prop.name ="MyProperty"
    prop.string_value = "MyValue"

    print(ifc_element)
```

error: 
AttributeError: entity instance of type 'IFC4.IfcBuildingElementProxy' has no attribute 'BIMObjectProperties'

what am i missing. 
when saving as ifc file, the blender custom properties are gone. so i have no access to an ifc file

sboddy:
I don't know what you are missing. I'm not sure why you didn't just crib from the highlighted commands in the merge request. This would give you:
``` python
import bpy
import ifcopenshell
import bonsai.tool as tool


ifc_file = tool.Ifc.get()

for obj in bpy.context.selected_objects:  # some ifcbuildingelementproxies selected
[...]

PR #6921, opened 2025-07-18
When assigning an IFC class, optionally convert custom properties to …

sboddy:
…a custom pset with SHIFT+CLICK.

Enhances as per #6918 / #6919.

image

### DIFF
--- a/src/bonsai/bonsai/bim/module/root/operator.py
+++ b/src/bonsai/bonsai/bim/module/root/operator.py
@@ -18,6 +18,7 @@
 
 import bpy
 import bmesh
+import idprop
 import ifcopenshell
 import ifcopenshell.api
 import ifcopenshell.api.geometry
@@ -370,11 +371,15 @@ class AssignClass(bpy.types.Operator, tool.Ifc.Operator):
 
             # Accomodate existing importers to Blender from other formats that set custom props
             if self.props_to_pset:
-                custom_props = {
+                custom_props = {k: v for (k, v) in obj.items() if type(v) in [bool, int, float, str]}
+                custom_array_props = {
                     k: v
                     for (k, v) in obj.items()
-                    if type(v) in [bool, int, float, str]
+                    if type(v) is idprop.types.IDPropertyArray and v.typecode in ["d", "i", "b"]
                 }
+                for k in custom_array_props.keys():
+                    for idx in range(len(custom_array_props[k])):
+                        custom_props["{}.{}".format(k, idx + 1)] = custom_array_props[k][idx]
                 pset = ifcopenshell.api.pset.add_pset(ifc_file, product=element, name="BBIM_ImportedBlenderProps")
                 ifcopenshell.api.pset.edit_pset(ifc_file, pset=pset, properties=custom_props)
 
--- a/src/bonsai/bonsai/bim/module/root/operator.py
+++ b/src/bonsai/bonsai/bim/module/root/operator.py
@@ -371,15 +371,14 @@ class AssignClass(bpy.types.Operator, tool.Ifc.Operator):
 
             # Accomodate existing importers to Blender from other formats that set custom props
             if self.props_to_pset:
-                custom_props = {k: v for (k, v) in obj.items() if type(v) in [bool, int, float, str]}
-                custom_array_props = {
-                    k: v
-                    for (k, v) in obj.items()
-                    if type(v) is idprop.types.IDPropertyArray and v.typecode in ["d", "i", "b"]
-                }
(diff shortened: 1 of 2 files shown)


Andrej730:
@sboddy 

1. Let's use alt+click? We usually use alt to add some additional behaviour and shift is for some kind of batch mode.

2. Very important to keep using `_execute`, otherwise undo will break - https://docs.bonsaibim.org/guides/development/undo_system.html
image

3. No need for `default=False`, it is `False` by default.

4. Is there need to skip `BIMObjectProperties`? It doesn't have a type `(bool, int, float, str)` either way.  Also, what is `_RNA_UI`?

    Also, depending on user addons there could be lots of garbage internal properties from other addons that makes sense to skip.

    For example:

```python
bpy.types.Object.yyy = bpy.props.StringProperty()
bpy.context.object.yyy = '42'
print(list(bpy.context.object.items()))
# [('BIMObjectProperties', <bpy id prop: owner="OBCube", name="BIMObjectProperties", address=0x0000016AE2BA99E8>), ('yyy', '42')]
```


5. Can you please add a comment with a small note why this feature needed? (e.g. to accomodate existing importers to Blender from other formats)

6. Can you please add a basic test for this feature to `test/tool/test_root.py`?

sboddy:
1. Yeah... my X11 system eats Alt-Click somehow, and I never figured out how to fix. I have to Ctrl-Alt-Click... unless that is actually assigned, in which case I can't Alt-Click things. I really need to update and switch to Wayland. **Done**.
2. I never knew the execute vs _execute thing. I just thought it related to `return {FINISHED}`. **Done**
3. **Done**
4. That expression just grew as I was poking the code. _RNA_UI (blender internals - [shrug]) was from an example. That and BIMObjectProperties were my initial attempts. Then I added the type testing, superceeding them, but I was lazy and didn't delete them. **Done**
5. **Done**
6. I've never run or written these tests. This'll take me a bit more than a couple of quick edits. Will switch the PR back to "Ready for review" once done.

Ugh! Now black is complaining about the dict comprehension in 4 now it's shorter. I'll fix that with the test commit.

sboddy:
@Andrej730 OK, so 6 is done. I improved it a little bit with arrays. I can't think of a way to sensibly handle the data_block type in psets. The Python block type can hold a dictionary or list of "stuff". I suppose this could use repr() to store a string in psets if someone wants to add it.

Also there's an option to handle the float and float_array subtype, either converting them to the right Ifc data type (i.e. IfcLengthMeasure), or in arrays naming the components appropriately (i.e. _R, _G, _B, _A, or _X, _Y, _Z, _W etc.) instead of using the position.

Andrej730:
@sboddy 


1. 
> `v.typecode in ["d", "i", "b"]`

Seems unnecessary, it's basically all possible types in `typecode` - https://docs.blender.org/api/current/idprop.types.html#idprop.types.idprop.types.IDPropertyArray.typecode


2. 
```python
                custom_props = {k: v for (k, v) in obj.items() if type(v) in [bool, int, float, str]}
                custom_array_props = {
                    k: v
                    for (k, v) in obj.items()
                    if type(v) is idprop.types.IDPropertyArray and v.typecode in ["d", "i", "b"]
                }
                for k in custom_array_props.keys():
                    for idx in range(len(custom_array_props[k])):
                        custom_props["{}.{}".format(k, idx + 1)] = custom_array_props[k][idx]
```

This can be simplified to just one for loop around `obj.items()` to keep it simple.

3.
> `bpy.ops.mesh.primitive_cylinder_add(vertices=10, location=(0, 4, 0))`

Why cylinder is added twice? Looks like duplicated code.

sboddy:
1. Didn't know that. Removed.
2. Done.
3. Was compacting repeated setup code into function, and missed those two lines. Removed.

Andrej730:
Are you sure we need custom pset name for this? It seems no one ever will use it, since the main goal was just to adapt the importers that create custom properties automatically, not create a new workflow when users would start creating custom propeties instead of psets.

sboddy:
I was just thinking of the case where someone wants to use a template with this to set the Ifc data types. If you had objects with different sets of custom properties, you would have one big template polluting each other with unused properties. I have a tendency to think of the worst possible use case. :-) You can just exclude that if it bothers you.

Andrej730:
I think it's better to keep it simple, unless there's a need for it, otherwise it will be a dead feature no one will find and use :(

sboddy:
Reverted it. Is it preferred to merge/squash PR's with a noisy log?

Andrej730:
Thank you.
It's okay, I can squash the PR during the merge.

Issue #72, opened 2016-05-19
cmake failure on windows build script


berndhahnebach:
Since the layer slicing stuff is in master I gave the windows build script a try after a few weeks of quietness. I got an error on cmake. Does someone know what's going wrong here.

Is it the missing boost one. What do I need to look for, a dll ?!

cheers bernd

~~~
C:\Users\bhb\Desktop\ifcos\ifcosdev\win>run-cmake.bat "Visual Studio 12 2013 Win64"


Script configuration:
  Generator = "Visual Studio 12 2013 Win64"
  Arguments =

Dependency Environment Variables for IfcOpenShell:
   BOOST_ROOT              = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps\boost
   BOOST_LIBRARYDIR        = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps\boost\stage\vs2013-x64\lib
   ICU_INCLUDE_DIR         = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps-vs2013-x64-installed\icu\include
   ICU_LIBRARY_DIR         = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps-vs2013-x64-installed\icu\lib
   OCC_INCLUDE_DIR         = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps-vs2013-x64-installed\oce\include\oce
   OCC_LIBRARY_DIR         = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps-vs2013-x64-installed\oce\Win64\lib
   OPENCOLLADA_INCLUDE_DIR = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps-vs2013-x64-installed\OpenCOLLADA\include\opencolla
da
   OPENCOLLADA_LIBRARY_DIR = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps-vs2013-x64-installed\OpenCOLLADA\lib\opencollada
   PYTHONHOME              = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps-vs2013-x64-installed\Python27
   PYTHON_INCLUDE_DIR      = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps-vs2013-x64-installed\Python27\include
   PYTHON_LIBRARY          = C:\Users\bhb\Desktop\ifcos\ifcosdev\deps-vs2013-x64-installed\Python27\libs\python27.lib
[...]

PR #6928, opened 2025-07-21
IfcCostSchedule PDF Export: Add Analysis of Rates

carlopav:
Add support for Analysis of Rates.
Analysis of rates can be created when printing a Schedule of Rates and is displayed at the end of the document.
Each CostItem with more than one CostValue is analyzed.

immagine

### DIFF
--- a/src/bonsai/bonsai/bim/module/cost/operator.py
+++ b/src/bonsai/bonsai/bim/module/cost/operator.py
@@ -936,7 +936,7 @@ class ExportCostSchedulesToPDF(bpy.types.Operator, ExportHelper):
     )
     should_print_analysis_of_rates: bpy.props.BoolProperty(
         name="Analysis of rates",
-        description="Print Analysis of Rates at the\n end of the document",
+        description="If printing a Schedule of Rates, display\nthe Analysis of Rates for each\nitem that has more than one Cost Value.",
         default=True,
     )
     force_schedule_type: bpy.props.EnumProperty(
--- a/src/bonsai/bonsai/bim/module/cost/operator.py
+++ b/src/bonsai/bonsai/bim/module/cost/operator.py
@@ -940,12 +940,12 @@ class ExportCostSchedulesToPDF(bpy.types.Operator, ExportHelper):
         default=True,
     )
     force_schedule_type: bpy.props.EnumProperty(
-        name="Force output type",
-        description="Force the output to this type\nalso if it is not coincident with the cost schedule Predefined Type",
+        name="Output type",
+        description='Force the output to this type.\n"Auto" defaults to selected cost schedule Predefined Type',
         items=[
             (
-                "OFF",
-                "Off",
+                "AUTO",
+                "By PredefinedType",
                 "Uses Cost Schedule Predefined Type",
             ),
             (
@@ -959,7 +959,7 @@ class ExportCostSchedulesToPDF(bpy.types.Operator, ExportHelper):
                 "Forces the output as a schedule of rates",
             ),
         ],
-        default="OFF",
+        default="AUTO",
     )
 
     def draw(self, context):
--- a/src/bonsai/bonsai/bim/module/cost/operator.py
+++ b/src/bonsai/bonsai/bim/module/cost/operator.py
@@ -924,11 +924,6 @@ class ExportCostSchedulesToPDF(bpy.types.Operator, ExportHelper):
         description="Export the full list of quantities",
         default=False,
     )
-    should_print_each_cost_value: bpy.props.BoolProperty(
-        name="Show each cost value",
-        description="Export the full list of cost values\nassociated with each cost item\nin the schedule of rates",
-        default=False,
-    )
     should_print_rates: bpy.props.BoolProperty(
         name="Rates and totals",
         description="Print rates and totals for each voice",
@@ -937,6 +932,11 @@ class ExportCostSchedulesToPDF(bpy.types.Operator, ExportHelper):
     should_print_summary: bpy.props.BoolProperty(
         name="Final Summary",
         description="Print summary at the end of the document",
+        default=True,
+    )
+    should_print_analysis_of_rates: bpy.props.BoolProperty(
+        name="Analysis of rates",
+        description="Print Analysis of Rates at the\n end of the document",
         default=True,
     )
     force_schedule_type: bpy.props.EnumProperty(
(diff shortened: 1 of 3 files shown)


carlopav:
@maxfb87 Ciao! Here another improvement. What do you think? Perhaps we can serialize more attributes of the CostValue, so we can display units and so... Also do not know how to deal with different operators (SUM, DIFFERENCE, etc...)

maxfb87:
@carlopav at the moment, what you have reported is not a "analysis of rate" because, in order to be that, every cost value should be multiplied also with quantities and it is not possible afaik...
This is the reason because, at the moment, the "analysis of rate" is done creating a new summary cost item with nested cost items...so if you want to multiply every nested cost item by quantities you can do that.
Anyway, the UI looks very good (as always :-) ) and what you have done could be useful for reporting the cost values related to a cost item...but i think it is not an analysis of rate...
This is still an unexplored territory so any idea is welcome

carlopav:
Thx @maxfb87  for commenting! 
That was a precise aim of the PR: to raise the problem and gather ideas on the topic.

Well, i disagree on the point that this Is not an analysis of rate... Its one way to do It, and IFC allows that.

I can support also rate analysis with nested cost items, but i need a flag to understand when a sumitem Is in the reality a compound cost...

A flag would be Ideal in both cases, because relying on cost items with more than one cost value could be tricky as many times cost items should carry the labor cost inside, for separated labor cost analysis...

maxfb87:
Restarting from the beginning, a rate analysis is a sum of elements where every element has price x quantities, maybe with categories.
It is necessary that the final price can be assigned as a rate for another cost item.
In order to do that, there are two solutions:
- nesting cost item (atm this way is supported with Bonsai see issue #6570)
- using components in the cost value. In this case, i don't understand if a component can be a quantity that it is not representing a monetary value. I read the schema many times, but i still struggle.
So, in the meantime, a checkbox could be a good solution i guess.

carlopav:
@maxfb87 Yes, I agree.
I'd also like to understand what should be the Ifc most suitable way to perform a rate analysis... That is to say what is usually represented by that picture:
maxresdefault
pehaps @aothms or @myoualid or @Moult could provide some insight on that :)

I also still do not know if this PR is a correct implementation of option 2 (using components in the cost value): I added a column to the csv output with a serialization of the cost values, but later I realized that the csv already export a colum for each cost value category? So I could also use that, but that colum lacks all the other attributes of the cost value (name, description, unit, etc..)... so, i'm still a bit confused :) that's because the PR is still a draft.

maxfb87:
@carlopav i tried to reproduce the example in a .ifc file and the result is the attached file.
Basically there are 3 schedules: one for elementary costs without quantities, one that combines the elementary costs with quantities, also with formulas, and the final BoQ that applies the rate with the quantities derived by the model.
With today commit, it is possible to store also the categories, in order to perform better rate analysis.
It can be improved and surely there are parts with errors, but the structure is there and this a way to perform rate analysis (already available)
[prova_rate_cost_item.txt](https://github.com/user-attachments/files/21606382/prova_rate_cost_item.txt)
Happy to receive feedback btw :-)

carlopav:
Hello @maxfb87, that's a nice hacky way to approach the topic of rate analysis :) 
I like how it fits with current possibilities (even cleaner with just one SoR that contains both elements and finished costs), but think this approach falls short when you intend to approach the `taxes/contractor overhead` part of analysis... How is the user intended to input those values? Hope not with formula on each elementary cost...
Anyway to me it's quite easy to support that with PDF export if there is a difference in data structure between a group of rates and a rate analysis. If not I can just change the SoR to look more like a BoQ (also if this is not how RateAnalysis is usually delivered).
Btw, is it possibile now to have quantities in SoR? I missed the commit :)

I still think Ifc has the data structure to deal with rate analysis in a better way using cost values, but for sure having something already working has it's advantages.

maxfb87:
@carlopav at the moment, it's not possible to use quantities in a schedule predefined type SCHEDULEOFRATES in bonsai (in fact i used predefined type USERDEFINED instead) but in the future it could be supported in Bonsai.
In the ifc schema i didn't find that a SOR can't have quantities...but it would be a big change in Bonsai so i would do it only if this method is ok also for other users...

Regarding taxes/contractor overhead, do you think it is a problem defining that number in the formula?
At the moment, there is no dedicated UI but it could be a value that the user insert one time and it could be applied to all specified cost values (for example all summary cost item in a particular schedule).

carlopav:
> in fact i used predefined type USERDEFINED instead

Oh, you fooled me :) i didn't realize that :) I do not have a strong opinion on quantities in SoR, but as you say, it's a change that Is worth discussing.

About taxes/contractor i think It would be a pity to not have It explicit and kinda automatic, It looks like that it would be kinda hidden using formulas, but perhaps i misunderstand... can you update your example also with that?

carlopav:
About data structure, I was able to create that in python:
```

#69=IFCCOSTVALUE('Total',$,IFCMONETARYMEASURE(1000.),$,$,$,$,$,.MULTIPLY.,(#70,#71));
#70=IFCMEASUREWITHUNIT(IFCLENGTHMEASURE(5.),#2);
#71=IFCCOSTVALUE('Skilled worker rate',$,IFCMONETARYMEASURE(500.),$,$,$,'Labor',$,$,$);
```
And I could go on nesting costs to reproduce all the tree (but i'm lazy).
What do you think @maxfb87 ?

maxfb87:
@carlopav yeah i understand the purpose. This is i think a more right way to do that

#72=IFCCOSTVALUE('Total',$,$,$,$,$,$,$,.MULTIPLY.,(#73,#74));
#73=IFCCOSTVALUE($,$,IFCLENGTHMEASURE(5.),$,$,$,$,$,$,$); 
#74=IFCCOSTVALUE('Skilled worker rate',$,IFCMONETARYMEASURE(500.),$,$,$,'Labor',$,$,$);

and this is what it shows in Bonsai

image

Apart from what Bonsai is showing (i think this workflow is not yet implemented, but it could i guess), my main concern is if the schema allows to store as AppliedValue something that is not representing a currency... if so, this could be a way

steverugi:
can't this be the way? (it uses resources and categories to calculate %)

image

then you add it to your schedule of rates

but as soon as doable I would remove the restriction on SCHEDULEOFRATES to not have quantities, it's not IFC standard either


image

steverugi:
sorry, I forgot to attach the example file with the rate built-up

[pricelist.ifc.txt](https://github.com/user-attachments/files/21683511/pricelist.ifc.txt)

carlopav:
I think it's better to close the pr as there is no consensus on how to deal with the topic :) and the proposed changes does not satisfy any of the possible approaches...

maxfb87:
I think the feature is useful, so let's see the best way to create a schedule of rates!

Issue #6931, opened 2025-07-22
`bim.unassign_representation_item_style()` doesn't seem to work in bulk on certain objects.


theoryshaw:
### Bug Description

`bim.unassign_representation_item_style()` seems to work in bulk when I select those (3) IfcFootings, but doesn't when i select those (5) IfcMembers. 

[stripped down - Copy.ifc.txt](https://github.com/user-attachments/files/21357971/stripped.down.-.Copy.ifc.txt)

### Attachments

https://github.com/user-attachments/assets/d80bd248-1ff6-4d41-bcf0-5ebc56d9787d

### Debug and Error Output

```true

```

theoryshaw:
I tried to debug this, but ran into [this](https://github.com/IfcOpenShell/IfcOpenShell/issues/6628) issue, unfortunately.

theoryshaw:
I 'think' i got it...

@Andrej730 can you please verify?
https://github.com/IfcOpenShell/IfcOpenShell/pull/6933

Issue #27, opened 2016-02-02
windows buildscript


berndhahnebach:
Just tried to build on windows 7 by provided buildscript. I had done this before sucessfully a few times (last one on 8th of January). I get an error on building depencieces. I do not know if it is the right place to report the problem?

Attached a file containing the last 9999 lines of compiling.  Is oce the culprit? 

bernd
[log.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/113552/log.txt)

Stinkfist0:
For starters I would probably try removing the deps\oce\build-vs20XX-xYY folder and try again. If that doesn't help, maybe try removing deps\oce altogether and rerun the script.

berndhahnebach:
Thanks for your help. Tested both, same problem.

aothms:
@Stinkfist0 I don't see any recent commits in OCE (not sure if recent changes broke the oce build), but maybe it's better to let the build script checkout a tag to be sure it's stable?

Stinkfist0:
@aothms Yeah, it would be nice to fix OpenCOLLADA and Open Cascade to known good revisions.

berndhahnebach:
+1

Stinkfist0:
I will be testing with VS 2013 x64 build on Windows 10 and will report back.

Stinkfist0:
@berndhahnebach seems to build OK. Maybe try building the INSTALL project of the oce.sln manually from the VS IDE?

berndhahnebach:
I'm not a windows guy in the regard of compiling. How would I do this?

Stinkfist0:
@berndhahnebach A-haa, I see `C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\bin\x86_amd64\` in your output - are you running the VS2013 x64 Cross Tools Command Prompt? I haven't tried that and have used VS2013 x64 Native Tools Command Prompt always.

berndhahnebach:
The window title of my comand prompt says 
VS2013 x64 Cross Tools Command Prompt

Thats the one I have been successfully build IfcOpenShell with your Build script quite a few times since August 2015

Stinkfist0:
OK, there might be something wonky in OCE regarding that. I'd recommend using Developer Command Prompt for VS2013 if doing only 32-bit builds or VS2013 x64 Native Tools Command Prompt if doing 64-bit and 32-bit builds.

berndhahnebach:
[...]

PR #6933, opened 2025-07-22
proposed fix for #6931

theoryshaw:


Andrej730:
I think it will work, but the problem was not that it wasn't handling IfcMappedItems - the problem was it was using unresolved representation to get representation items (e.g. `#245401=IfcShapeRepresentation(#27,'Body','MappedRepresentation',(#245400))` isntead of `#8165=IfcShapeRepresentation(#27,'Body','SweptSolid',(#8163))`). Representation can be resolved using `ifcopenshell.util.representation.resolve_representation`.

Haven't seen `bim.unassign_representation_item_style` implementation before, but it feels odd - it does one thing for the active object (unassigns style for active representation item specifically) and something very different for all other selected - unassigns style from all their representation items. 

It feels like batching option should be available on shift+click or something, that would also allow unassigning current style from all representation items of the active object too, to make operator's effect consistent.

Another thing noticed - it's also kind of odd that it currently unassigns the style from all available representations of the element (it's iterating over all `element.Representation.Representations`), not from just the active one.

theoryshaw:
Thanks @Andrej730  
These comments make sense to me. 
I'll see if i can dig into it more. :)

BIMvoice:
@theoryshaw @Andrej730 the bug still reproduces on current `v0.8.0`: with the sample from #6931 (5 IfcMembers), `bim.unassign_representation_item_style()` leaves 4 of the 5 members styled.

Andrej's diagnosis is the right one. The failing path took `element.Representation.Representations[0].Items[0]`, which for these Revit-family members is the unresolved `MappedRepresentation` whose `Items[0]` is an `IfcMappedItem` (its `StyledByItem` is empty). The style actually lives on the `IfcExtrudedAreaSolid` inside `MappingSource.MappedRepresentation`, so nothing matched and nothing was removed.

Implementing your "resolve the representation" suggestion at the item level with the existing helper (`ifcopenshell.util.representation.resolve_base_items`), scoped to the object's active representation (which also addresses the "iterates over all Representations" oddity you flagged):

```python
for obj in context.selected_objects:
    if obj == active_obj:
        continue

    representation = tool.Geometry.get_active_representation(obj)
    if not representation:
        continue

    # Resolve mapped items so the items actually carrying the styles are reached (see #6931)
    changed = False
    for item in ifcopenshell.util.representation.resolve_base_items(representation):
        styles = set()
        if hasattr(item, "StyledByItem"):
            for styled_by_item in item.StyledByItem:
                if hasattr(styled_by_item, "Styles"):
                    styles.update(styled_by_item.Styles)
        if styles & active_styles:
            tool.Style.assign_style_to_representation_item(item, None)
            changed = True

    if changed:
        tool.Geometry.reload_representation(obj)
```

Verified live in headless Blender 5.1.2, driven end to end through the real `bpy.ops.bim.unassign_representation_item_style()` on the actual `stripped down - Copy.ifc.txt`:

- Before: `4/5 members STILL have a style (6048)`
- After: `0/5 members still have a style`

This deliberately leaves the broader UX point (batching on shift+click, making the active-object behaviour consistent) as a separate decision for you both, since that is a design call rather than part of the reported bug.

Happy to push this onto your branch here @theoryshaw, or open a fresh PR that credits your draft, whichever you prefer.

theoryshaw:
Thanks @BIMvoice , i tweaked the PR a little.

PR #6935, opened 2025-07-22
S_Z for wall tool to change_extrusion_depth

falken10vdl:
This is a proposal to allow to extend walls in the z direction similar to how it is done with S_E


theoryshaw:
This is awesome. 
Would it be possible for this to work on all LAYER2 objects for all classes?
For example, i could see using this tool for IfcCoverings as well, among other classes.

falken10vdl:
@theoryshaw I have removed the ifcWall restriction and do bim.change_extrusion_depth for all LAYER2 objects.
Sorry I do not know exactly how these LAYER2 obejcts work :) I need to go over your tutorials to have a clearer picture.
By looking to the code I see the ChangeExtrusionDepth seems to be able to be applied to LAYER2.
I have added an ifccovering with IfcMaterialLayerUsage so it has an AXIS2 (and therefore tool.Model.get_usage_type(element) returns "LAYER2") and it is properly handled. Is this what you are looking for?
image
and after C_E...
image

Cheers!

theoryshaw:
Works great! thank you.

brunoperdigao:
@falken10vdl Nice addition! I guess is just missing that you run black formatter and we're good to go.

falken10vdl:
Cool! thanks @brunoperdigao!  I just run black formatting.
Cheers!

Issue #5696, opened 2024-11-05
Improved speed of `bim.activate_model` command


theoryshaw:
`bim.activate_model` takes a long time to run, would be nice if this could be streamlined somehow. 

I usually resort the following instead, since it's faster. 

![2024-11-05_10-06-48_blender_Lake_Kegonsa_Renovation_ DDropboxGitea_ODLake_Kegblender](https://github.com/user-attachments/assets/9801a4c1-a869-450d-96a0-40e7743ca717)

Andrej730:
@theoryshaw can you please share some model that's good for benchmarking?

theoryshaw:
NP...
- If you want a monster:) https://hub.openingdesign.com/OpeningDesign/Highland_Haven/src/commit/a4079de8fdc2f10b2d95bbb61542fe560a83b143/Open/Models_and_CAD/BlenderBIM/Highland_Haven.ifc
- If you want something more reasonable: https://hub.openingdesign.com/OpeningDesign/Verona_3_Season_Porch/src/commit/53e2affc7d78ea9f403278f77744b64096304b72/Open/Models/BlenderBIM/Verona_3_Season_Porch.ifc

sboddy:
So I just tested Highland Haven with my ~~fix~~hack:
Just to activate "FLOOR PLAN - 1ST" takes ~20 secs on my poor old tortured laptop.
Without the skip_list hack, bim.activate_model takes ~351 secs (nearly 6 minutes)
With the skip_list hack, bim.activate_model takes ~36 secs, so ~1/10th the time. Still kinda slow, but much improved.

sboddy:
Strange, this makes little to no difference to my own models bim.activate_model time.
I notice two possibilities:
1. I use a lot of include/exclude filters on my drawings. I don't think this is it, because even drawings that include everything return fast.
2. ~~Where I have lots of instances, they are things like beams, or other extruded poly's. I don't have a lot instances of tessellated meshes (IfcPolygonFaceSet). I noticed the ROSTER window in blenderarc's file was tessellated. Looking through, I see a lot of tessellated types with lots of instances in HighlandHaven,  esp. doors and windows. I suspect the IfcPolygonFaceSet is _slooooow,_ and made worse hitting every instance multiple times.~~ Scratch that. I tried converting the ROSTER window to an extrusion using the representation tools and it made little-to-no difference.

[...]

Issue #6923, opened 2025-07-19
Viewport Display


blenderarc:
### Bug Description

I entered Active Drawing mode, then went back to Active Model. What happened was that several elements changed to 'bound'. To restore them, I had to click on each one individually to make them solid again

<img width="1920" height="1020" alt="Image" src="https://github.com/user-attachments/assets/46d41907-79da-44e0-b1b8-111ba169709c" />

[AHSAN - Copy.ifc.txt](https://github.com/user-attachments/files/21329923/AHSAN.-.Copy.ifc.txt)

### Debug and Error Output

```true

```

sboddy:
I just tried to reproduce, and could not. I'm on a slightly newer 0.8.4-alpha250716 nightly. All those ROSTER blocks remain solid tessellations. Was it a specific drawing that led to this, or any of them caused this for you. Did you try exiting Blender, reloading, and retry? Still happening?

theoryshaw:
I can't replicate either. 

Have you guys noticed, however, that `bpy.ops.bim.activate_model()` <img width="22" height="25" alt="Image" src="https://github.com/user-attachments/assets/fbe87d46-50b9-413c-b19f-9960f604a388" />takes a long time?  It takes about 40 seconds on my end, for this model.


```
os: Windows
os_version: 10.0.19045
python_version: 3.11.7
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: Intel64 Family 6 Model 158 Stepping 10, GenuineIntel
blender_version: 4.2.9 LTS
bonsai_version: 0.8.4-alpha250718
bonsai_commit_hash: d49a348
bonsai_commit_date: 2025-07-18T14:47:14-07:00
```

sboddy:
Oh yeah. It was sluggish to switch back. ~33 secs to go from "ATAS PLAN" back to model mode.

sboddy:
Unrelated to the main issue _maybe,_ or maybe blenderarc has a system that the slowness causes issues with?

I did a few old-school timestamping statements.

The problem is in the bit where the function loops over all the visible object and switches representation.

Seems most of the ~180 IfcWindow objects are taking around a third of a second _each._
These are the curious ones that are becoming the bounds outline for blenderarc, which look more of a louvered building block than an actual window. I took a look and the geometry looked a bit funky. I tried fixing that, but it didn't make much (if any) difference. Windows were still taking around a quarter of a second.

[...]

PR #6940, opened 2025-07-23
Improve bim.activate_model speed ~10x when exiting a drawing #5696

sboddy:
Putting this forward as much to spur discussion, rather than in an attempt to get it merged.

Heavy files that use lots of instances of types can take a long time to exit a drawing with the bim.activate_model operator. I tracked it down to iterating over every instance, and while processing *that* instance, processing every sibling. i.e. 10 instances would result in 100 change_data calls. This dumb hack appears to work with no negative side effects in ~1/10 th the time.

A test of theoryshaw's Highland Haven torture test in #5696 on my laptop went from ~351 secs to ~36 secs.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -2204,6 +2204,7 @@ class ActivateModel(bpy.types.Operator):
                         is_global=True,
                         should_sync_changes_first=True,
                     )
+        tool.Geometry.clear_skip_list()
 
         # restore visibility after hide_view_clear()
         for obj, hide_status in visibility_status.items():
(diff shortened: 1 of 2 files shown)


sboddy:
With my new found profiling powers I re-ran the test on the associated issue #6923 OP's test file. This one drops from 30+ secs (first call graph) to <2 secs (second call graph). However, I noticed something interesting.
call_graph_without_hack
call_graph_with_hack
In _both_ cases more than 50% of the clock time is spent initializing the iterator [here](https://github.com/IfcOpenShell/IfcOpenShell/blob/5bd9a58936cd303812234f19f34deeff3839a588/src/bonsai/bonsai/tool/geometry.py#L879)
I have no idea if this can be sped up, but after applying my hacky solution to the seemingly unnecessary calls to change_data, the initialize calls are taking nearly two thirds of the remaining time.

Andrej730:
Fixed in c2730a81d

PR #6948, opened 2025-07-25
Don't mark function that throws exception noexcept

Osyotr:
Fixes annoying warning in MSVC builds:

```
src\ifcparse\variantarray.h(198): warning C4297: 'VariantArray<Blank,Derived,int,bool,boost::logic::tribool,double,std::string,boost::dynamic_bitset<unsigned long,std::allocator<Block>>,EnumerationReference,IfcUtil::IfcBaseClass *,empty_aggregate_t,std::vector<int,std::allocator<int>>,std::vector<double,std::allocator<double>>,std::vector<std::string,std::allocator<std::string>>,std::vector<boost::dynamic_bitset<Block,std::allocator<Block>>,std::allocator<boost::dynamic_bitset<Block,std::allocator<Block>>>>,aggregate_of_instance::ptr,empty_aggregate_of_aggregate_t,std::vector<std::vector<int,std::allocator<int>>,std::allocator<std::vector<int,std::allocator<int>>>>,std::vector<std::vector<double,std::allocator<double>>,std::allocator<std::vector<double,std::allocator<double>>>>,aggregate_of_aggregate_of_instance::ptr>::index': function assumed not to throw an exception but does
        with
        [
            Block=unsigned long
        ]
src\ifcparse\variantarray.h(198): note: __declspec(nothrow), throw(), noexcept(true), or noexcept was specified on the function
```

### DIFF
--- a/src/ifcparse/variantarray.h
+++ b/src/ifcparse/variantarray.h
@@ -193,7 +193,7 @@ public:
         free_();
     }
 
-    std::size_t index(std::size_t index) const noexcept {
+    std::size_t index(std::size_t index) const {
         if (index >= size()) {
             throw IfcParse::IfcException(
                 "Index " + std::to_string(index) + " is out of range for variant of size " + std::to_string(size())


Osyotr:
Note that there's another place in this file that emits warnings:
https://github.com/IfcOpenShell/IfcOpenShell/blob/1ac5a8d7855dec86a0b9d3eef60c2394c988db3f/src/ifcparse/variantarray.h#L314-L320
```
src\ifcparse\variantarray.h(318) : warning C4702: unreachable code
```

Though I'm not sure what fix would be appropriate.

aothms:
Ah thanks, that's indeed sloppy. One of the very few noexcept qualifiers in the codebase is even wrong.

> Note that there's another place in this file that emits warnings:

I remember that, that fixed compilation, because with only the exception only it wouldn't know the return type, or it wouldn't be covariant, don't remember the full details. Feel free to experiment with changing auto to void for the return type for example if the warning bothers you :)

PR #6957, opened 2025-07-28
Adjust measurement tools information so labels do not overlap

falken10vdl:
In some scenario the labels created by the measurement tool can overlap each other so it is difficult to read.
@brunoperdigao  Here is a proposal to implement a function adjust_overlapping_labels inspired in the one used for the BoundingBox decorator in order to improve readability.
Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/decorator.py
+++ b/src/bonsai/bonsai/bim/module/model/decorator.py
@@ -521,6 +521,10 @@ class PolylineDecorator:
         color = self.addon_prefs.decorations_colour
 
         blf.color(self.font_id, *color)
+
+        screen_coords = {}
+        label_index = 0
+
         for i in range(len(self.polyline_points)):
             if i < 1 and self.measure_type == "POLY_AREA":
                 continue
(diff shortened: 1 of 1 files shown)


brunoperdigao:
@falken10vdl Thanks, that's very useful. I'll take a look as soon as possible.

brunoperdigao:
@falken10vdl I tested your PR and I spotted two issues (see image below).
- The text is missing when you create just a single line with the polyline measure.
- The polyline measure shoudn't display the last angle since it's not a closed polyline.

image

brunoperdigao:
I also propose to reduce the min spacing to a low level like 2. Actually, I don't think it should be hardcoded, maybe it should based on the text size, which may vary according two blender settings.

falken10vdl:
@brunoperdigao Thanks for the feedback! Let me take a look to the issues and change the min spacing based on the text size as you say.
Thanks!

falken10vdl:
@brunoperdigao Please take a look to the updated code. I think it is addressing the problems it had before.
Thanks!

Issue #5166, opened 2024-08-08
when searching through a type, that it shortlists the type's via values in `Description` and `PredefinedType`, or `ElementType`, in addition to `Type.Name`. 


theoryshaw:
Would be nice, when searching through a type, that it shortlists the type's via values in `Description` and `PredefinedType`, or `ElementType`, in addition to `Type.Name`. 

https://github.com/user-attachments/assets/e820ecc3-dc94-44be-91fc-303eabae580f

Similar to how when searching through IfcClasses, the search results return the PredefinedType as well.

https://github.com/user-attachments/assets/e0f23b17-aa50-4ea3-963e-e5232c6cc88a

theoryshaw:
:)

theoryshaw:
@falken10vdl Was something like this implemented elsewhere?  I remember there was some similar work in this area?

falken10vdl:
Yes. I think that in a way it would be possible similar to https://github.com/IfcOpenShell/IfcOpenShell/issues/6822
Let me take a look and come back to you.
Cheers!

theoryshaw:
That would be awesome! Since the 'Name' attribute is [usually short](https://community.osarch.org/discussion/1627/what-pset-or-attribute-do-you-think-would-be-the-equivalent-of-revits-type-comments) (a number or code), it's hard sometimes finding what you're looking for.

falken10vdl:
@theoryshaw Please take a look to https://github.com/IfcOpenShell/IfcOpenShell/pull/6962
Is this what you are looking for? I had not a very comprehensive ifc file with a good bunch of examples to test. :)
I also have joined the suggestions in single line Name > Description > PredefinedType > ElementType > ObjectType instead of having many lines like Name > Description, Name > PredefinedType, etc.

theoryshaw:
Oh man, this works awesome! Thank you.

...

Would suggest having the same functionality here as well.

<img width="1588" height="665" alt="Image" src="https://github.com/user-attachments/assets/04ceb067-a1c3-4756-bd9b-9115a3b145e6" />

falken10vdl:
@theoryshaw updated the PR to include relating_type_id which is the one used in that property.
Cheers!

falken10vdl:
@theoryshaw Are we happy to close this one?
Cheers!

theoryshaw:
Yes, thanks so much!

PR #6962, opened 2025-07-30
Add suggestions for relating types in enum property search

falken10vdl:
Here is a proposal for https://github.com/IfcOpenShell/IfcOpenShell/issues/5166
Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/helper.py
+++ b/src/bonsai/bonsai/bim/helper.py
@@ -361,7 +361,7 @@ def prop_with_search(
     should_click_ok: bool = False,
     original_operator_path: Optional[str] = None,
     enable_relating_type_suggestions: bool = True,
-    search_threshold: int = 0,
+    search_threshold: int = 10,
     *,
     button_kwargs: Union[dict[str, Any], None] = None,
     **kwargs: Any,
--- a/src/bonsai/bonsai/bim/helper.py
+++ b/src/bonsai/bonsai/bim/helper.py
@@ -361,6 +361,7 @@ def prop_with_search(
     should_click_ok: bool = False,
     original_operator_path: Optional[str] = None,
     enable_relating_type_suggestions: bool = True,
+    search_threshold: int = 0,
     *,
     button_kwargs: Union[dict[str, Any], None] = None,
     **kwargs: Any,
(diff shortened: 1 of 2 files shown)


Andrej730:
@falken10vdl 

1. `if element and element.is_a().endswith("Type"):` - what this for?

2. `getattr(element, "Name", "") or "Unnamed"` - are there cases when Name is missing?

3. `value != "NOTDEFINED"` - why `NOTDEFINED` is special?

4. `next((getattr(element, "PredefinedType", None) for _ in [1]), "NOTDEFINED")` - what's this for?

5. Can you please add it as an option to `prop_with_search`, rather than hardcoding it to `("relating_type", "relating_type_id")`?

falken10vdl:
@Andrej730 My comments thanks!

>     if element and element.is_a().endswith("Type"): - what this for?
To make sure that the element are "types" and not instances. Foir example IfcWallType but not IfcWall The request was to add suggestions for ifcElmentTypes


>     getattr(element, "Name", "") or "Unnamed" - are there cases when Name is missing?
It is just defensive coding but I think there could be cases where the ifc schema has Name as optional or simply the ifc file is not well formated. This is just to ensure that the UI gets a readable friendly suggestion but I agree in most of the cases is not needed. Please advice if it is better to remove


>     value != "NOTDEFINED" - why NOTDEFINED is special?
Well because adding "NOTDEFINED" to the suggestions seems not very informative. Example:
PredefinedType = "NOTDEFINED"
Description = "Exterior Wall"
Only "Exterior Wall" would be included in the suggestion
This can be removed and the "NOTDEFINED" will appear as a suggestion if you think is more coherent



>     next((getattr(element, "PredefinedType", None) for _ in [1]), "NOTDEFINED") - what's this for?
It is the same as getattr(element, "PredefinedType", None) or "NOTDEFINED" just a way to handle the case where PredefinedType might be None and provide "NOTDEFINED" as a fallback
Changed.


>     Can you please add it as an option to prop_with_search, rather than hardcoding it to ("relating_type", "relating_type_id")?
In a smiliar way to how should_click_ok is done? Please take a look


Why only for >10?
`        if len(get_enum_items(data, prop_name, original_operator_path=original_operator_path)) > 10:`

With suggestions I believe it is usefull to get the search with suggestions even for fewer number of elements since the name could be cryptic

Cheers!

Andrej730:
@falken10vdl 

> To make sure that the element are "types" and not instances. Foir example IfcWallType but not IfcWall The request was to add suggestions for ifcElmentTypes

But how a non-type can end up in a relating type property?

> It is just defensive coding but I think there could be cases where the ifc schema has Name as optional or simply the ifc file is not well formated. This is just to ensure that the UI gets a readable friendly suggestion but I agree in most of the cases is not needed. Please advice if it is better to remove

`or "Unnamed"` makes sense, but `getattr` is misleading - it suggests that there are cases when .Name attribute will be missing completely.

> Well because adding "NOTDEFINED" to the suggestions seems not very informative. Example:

I see, basically it's the same as `None` then. Okay, let's skip it.

> It is the same as getattr(element, "PredefinedType", None) or "NOTDEFINED" just a way to handle the case where PredefinedType might be None and provide "NOTDEFINED" as a fallback

Makes sense, it's just that `next` and `for _ in [1]` construction was kind of odd.

> In a smiliar way to how should_click_ok is done? Please take a look

Yes, the same way, argument + prop.

> Why only for >10?

I agree it can be useful to have less than 10 in this case, as suggestions have additional information besides just name. Let's expose it as `prop_with_search` argument, so it would show up if there's more than 0 items.

falken10vdl:
> But how a non-type can end up in a relating type property

I guess it could be the case for a bad formed ifc file. Removed to make it simpler

> or "Unnamed" makes sense, but getattr is misleading - it suggests that there are cases when .Name attribute will be missing completely.

ok. Changed,

> I see, basically it's the same as None then. Okay, let's skip it.

Removed all the NOTDEFINED handling. It will appear the predifiendtype NOTDEFINED in the suggestions as well.

> I agree it can be useful to have less than 10 in this case, as suggestions have additional information besides just name. Let's expose it as prop_with_search argument, so it would show up if there's more than 0 items.

added search_threshold argument (default to 0)

Andrej730:
> added search_threshold argument (default to 0)

It should default to the previous value, `10`, to prevent altering existing UI.

falken10vdl:
> It should default to the previous value, 10, to prevent altering existing UI.

changed to 10

Cheers!

PR #6970, opened 2025-08-01
Add Description-ContextMenu Entry to bSDD Properties

c-mellueh:
I like the "IFC Description" Context Menu. So i added the same functionality for bSDD Properties.
image

image

### DIFF
--- a/src/bonsai/bonsai/bim/module/bsdd/__init__.py
+++ b/src/bonsai/bonsai/bim/module/bsdd/__init__.py
@@ -25,6 +25,7 @@ classes = (
     operator.LoadBSDDDictionaries,
     operator.SearchBSDDClassifications,
     operator.SearchBSDDProperties,
+    operator.BIM_OT_show_bsdd_description,
     prop.BSDDDictionary,
     prop.BSDDClassification,
     prop.BSDDProperty,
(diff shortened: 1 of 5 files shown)


Andrej730:
@c-mellueh 

Thanks for the PR, looks good in general.

A few things I've noticed:

1. Can you please submit bsdd tests fixes in a separate PR as those seem unrelated?
2. `op_description: BIM_OT_show_bsdd_description` - this is incorrect, it's a wrong type, `layout.operator` returns `OperatorProperties`. I think it would be better to annotate `bsdd_property` as `BSDDProperty` instead.
3. `# pset_name = result.get("propertySet") or "undefined"` - left out commented code.
4. `self.bl_label = result.get("name") or "undefined"` - dead code.
5. `total_results = 0` - dead code.
6. I think there should be some separation in case if there are both description and definition present:

image

c-mellueh:
@Andrej730  thank you for your feedback. 
I implemented your requests.
 Check out [5fa5e18](https://github.com/IfcOpenShell/IfcOpenShell/pull/6970/commits/5fa5e189cac007cc03f1d73932d6bcc21f9e49f7) if you're happy with my solution of separating description and definition.
I added a dashed line:
Screenshot_20250804_130552

Andrej730:
Looks good, thank you!

Issue #3254, opened 2023-06-04
Flipping a door a little glitchy


theoryshaw:
Flipping a door a little glitchy....

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/c890bafb-fbf4-4303-acff-42c1f3e5c78f

Gorgious56:
I also would like flipping a door to position the origin along the opposite side of the wall

sboddy:
Still an issue as of `bonsai_version: 0.8.3-alpha250619`

In addition, it's worth pointing out that the glitchy-ness happens when a drawing is active, and the representation is schematic with the door swing. If the flip is done in model mode with a 3D tessellated door, it flips OK. If in model mode the door is switched to the schematic representation and then flipped, the glitchy-ness is back.

Rob2309:
> I also would like flipping a door to position the origin along the opposite side of the wall

I just submitted a PR that should do exactly this #6986.
It should also fix this entire issue, as far as I can tell.

brunoperdigao:
@theoryshaw Is this issue solved?

theoryshaw:
Hi. I believe it is.
Thanks @Rob2309!

Issue #6392, opened 2025-03-18
When i duplicate a window type with a 'special' void, it resets the void to an extrusion on the new type.


theoryshaw:
When i duplicate a window type with a 'special' void, it resets the void to an extrusion on the new type. 

[House - Copy.ifc.txt](https://github.com/user-attachments/files/19328226/House.-.Copy.ifc.txt)

https://github.com/user-attachments/assets/f867f579-cd41-42b3-a280-2a84a5d52d0f

theoryshaw:
Probably somewhat related: if I import a type with a special tessellation void, it resets back to an extrusion based void in the current file. 

https://hub.openingdesign.com/OpeningDesign/OD_Project_Template/src/commit/9676d0023b2b307ae4db2d9668f250dd2f760683/Open/Models/Bonsai/OD_Bonsai_Template.ifc

In library file...
![Image](https://github.com/user-attachments/assets/c60573cc-9159-4645-bccf-93a245a208e3)

In receiving file...
![Image](https://github.com/user-attachments/assets/92e4ed13-ae85-43e0-a8df-3be71bdfead8)

theoryshaw:
It seems even with types that have extrusion-based voids, that the voids get reset incorrectly upon import to another file. 

In library file...
![Image](https://github.com/user-attachments/assets/b370ac6d-804e-46c1-a264-f92cb381fc04)


In receiving file...
![Image](https://github.com/user-attachments/assets/22c4a3a1-e06f-43f8-9c2a-fa09c45d516a)

theoryshaw:
@Moult 
You probably reverted this commit for a good reason, but i added the following PR, in the meantime, so it's available in BonsaiPR build--as not having openings duplicate correctly, causes a lot of pain in my workflow. 
https://github.com/IfcOpenShell/IfcOpenShell/pull/7802

...

Did you want me to readdress this proposed PR with changes, or does it need wholesale reevaluation?

theoryshaw:
I do think...
`marker = desc.split("||BonsaiOpeningTemplate:")[-1]`

...was a chunky approach.

theoryshaw:
I closed PR #7802 and am proposing the following PR instead: https://github.com/IfcOpenShell/IfcOpenShell/pull/8200

Issue #4432, opened 2024-03-15
If a non-material based type was mirrored, that BB would automatically create a new mirrored type. 


theoryshaw:
As a way to manage mirroring, I wondering if a system or an approach could be used where if a non-material based type was mirrored, that BB would automatically create a new mirrored type. IfcDoorType/DT01_Mirror, for example.

And it was 'hardwired' to IfcDoorType/DT01. That is, any change done to IfcDoorType/DT01 would change IfcDoorType/DT01_Mirror automatically, and vice versa.

Or, probably better than naming it will a 'Mirror' suffix, maybe the use of a `EPset_Parametric.Mirror` boolean parameter on the type.

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/dccc7a70-b50f-4036-9a76-45b0725d823d)

theoryshaw:
If you're in this head space, might want to look into this, as well...
https://github.com/IfcOpenShell/IfcOpenShell/issues/3254

Rob2309:
I just created a PR (#7003) that adds initial support for mirroring most products and product types, as well as linking mirrored types together.

PR #6986, opened 2025-08-03
Make Door / Window Tool place doors on the correct side of a wall

Rob2309:
### Changes
This PR makes the door and window tools place objects on the side of a wall that was actually clicked on.
Previously, the door / window would always be placed on the baseline of the wall, which seems unintuitive to me.
The door/window will also be oriented in a way that always makes the same side protrude out of the wall side.

The flip tool (shift + f) now also moves doors and windows to the other side of a wall when flipping them. For example, if a door protudes 1cm out of the wall, flipping it will make it protude 1cm out of the other side of the wall, instead of flipping it in place.
This could also be split out into a new tool if the behavior of the flip tool should stay unchanged.

The occurrence preview was also updated to match the actual orientation of the created occurrence (at least for every edge case I could think of). This also applies to non-door/window elements.

### Door / Window Tool

| Before | image  | image The door is placed on the wrong side of the wall. The preview does not match the final positioning. |
|--------|--------|-------|
| After | image | image The door is placed on the side of the wall that was actually clicked on. The preview matches the final positioning. |

### BuildingStorey display

This PR also fixes displaying the door/window on the default storey height, even if snapped to another storey.

| Before | image The door is shown on the default storey, which does not match its final location |
|---|---|
| After | image The door is correctly shown on the storey of wall it is snapped to |

### Flip tool

| Before | image | image The door is flipped in place. The door would now open into the wall |
|---------|------|------|
| After | image | image The door is flipped and moved to the other side of the wall. The door still opens to the outside of the wall. |

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/opening.py
+++ b/src/bonsai/bonsai/bim/module/model/opening.py
@@ -40,7 +40,7 @@ import bonsai.core.geometry
 import bonsai.bim.import_ifc as import_ifc
 from collections import defaultdict
 from math import pi, radians
-from mathutils import Vector, Matrix
+from mathutils import Vector, Matrix, Euler
 from bpy.types import Operator
 from bpy.types import SpaceView3D
 from bpy.props import FloatProperty
@@ -92,11 +92,26 @@ class FilledOpeningGenerator:
             layers = tool.Model.get_material_layer_parameters(element)
             if layers["layer_set_direction"] == "AXIS2":
                 opening_thickness_si = layers["thickness"] * 2
-                axis = tool.Model.get_wall_axis(voided_obj, layers=layers)["base"]
+                axes = tool.Model.get_wall_axis(voided_obj, layers=layers)
+                axis_base = axes["base"]
+                axis_side = axes["side"]
                 new_matrix = voided_obj.matrix_world.copy()
-                point_on_axis = tool.Cad.point_on_edge(target, axis)
-                new_matrix.translation.x = point_on_axis.x
-                new_matrix.translation.y = point_on_axis.y
+                point_on_base_axis = tool.Cad.point_on_edge(target, axis_base)
+                point_on_side_axis = tool.Cad.point_on_edge(target, axis_side)
+                if (point_on_base_axis - target).length <= (point_on_side_axis - target).length:
+                    new_matrix.translation.x = point_on_base_axis.x
+                    new_matrix.translation.y = point_on_base_axis.y
+                else:
+                    filling_min_x = min([p[0] for p in filling_obj.bound_box])
+                    filling_max_x = max([p[0] for p in filling_obj.bound_box])
+
+                    # offset that results in the same x extents when the model is rotated 180 degrees
+                    filling_opposite_x = filling_max_x + filling_min_x
+
+                    offset_dir = voided_obj.matrix_world @ Vector((filling_opposite_x, 0, 0, 0))
+                    new_matrix.translation.x = point_on_side_axis.x + offset_dir.x
+                    new_matrix.translation.y = point_on_side_axis.y + offset_dir.y
+                    new_matrix = new_matrix @ Euler((0, 0, radians(180.0))).to_matrix().to_4x4()
 
                 if should_set_z_level:
                     if filling.is_a("IfcDoor"):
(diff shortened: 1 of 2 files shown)


theoryshaw:
very cool!
...
If you're in this headspace, would suggest looking at this one too..
https://github.com/IfcOpenShell/IfcOpenShell/issues/3254

Rob2309:
> very cool! ... If you're in this headspace, would suggest looking at this one too.. #3254

Will take a look at this :)

Rob2309:
> very cool! ... If you're in this headspace, would suggest looking at this one too.. #3254

It seems like this PR also fixes #3254, at least with my quick tests.

theoryshaw:
Cool I will try too...
....
This also a relevant discussion
https://community.osarch.org/discussion/1422/how-to-mirror-an-element

Rob2309:
> Cool I will try too... .... This also a relevant discussion https://community.osarch.org/discussion/1422/how-to-mirror-an-element

I already thought about adding a separate "door mirror" operator that changes the handedness and actually mirrors the door instead of rotating it. However, this should probably be a separate PR to keep this one concise.

theoryshaw:
This might play into it as well, to have a more generic solution...
https://github.com/IfcOpenShell/IfcOpenShell/issues/4432

theoryshaw:
> It seems like this PR also fixes https://github.com/IfcOpenShell/IfcOpenShell/issues/3254, at least with my quick tests.

yes, seems to work great! thank you.

Rob2309:
I just opened another PR (#7003) that further builds on this PR and allows mirroring doors and other products.
It's a separate PR to allow merging this one without having to resolve all issues with mirroring.

theoryshaw:
I defer to @Andrej730 on the ultimate pull here.

theoryshaw:
The other aspect _(and I don't even think the original functionality accomodated)_ is when flipping a door/window/etc, it would be nice if the attached IfcOpeningElement flipped as well.  There are a lot of times in design, the IfcOpeningElement is multi-faceted, and not just a simple extrusion.

test file: 
[door flipping - Copy.ifc.txt](https://github.com/user-attachments/files/21689431/door.flipping.-.Copy.ifc.txt)

https://github.com/user-attachments/assets/6ce29474-8534-4ee4-9cd5-26e47df3a64b

Rob2309:
Did not think about this yet. I will work on that :)

theoryshaw:
Awesome. 
Also there's this that might be related: https://github.com/IfcOpenShell/IfcOpenShell/issues/6392
Sorry for piling on. :)  I just want to inform the roadmap.  As you can see the plumbing on these changes runs deep. :)

Rob2309:
> Awesome. Also there's this that might be related: #6392 Sorry for piling on. :) I just want to inform the roadmap. As you can see the plumbing on these changes runs deep. :)

Just for my understanding, how do you associate a custom opening with an object type?

Rob2309:
I just pushed a commit that copies the flipped objects matrix to the opening. This will result in it having the same position. For your test file this seems to work well. I am not quite sure if always setting the same position is correct though.

theoryshaw:
> Just for my understanding, how do you associate a custom opening with an object type?

The only way i know how, is to convert an extrusion to a tessellation...
If i understood your question, correctly. 

https://github.com/user-attachments/assets/b80043e1-c649-4599-aaeb-c0185d126869

Rob2309:
> The only way i know how, is to convert an extrusion to a tessellation...
> If i understood your question, correctly.

Ok I think I just found out how you do it. You just edit the opening and bonsai reuses this edited opening even for newly placed windows of the same type, correct?

theoryshaw:
correct.

Rob2309:
Ok I just tested a bit more, the opening positioning seems to work ok for all cases I could think of...

theoryshaw:
The only other thing i can think of is accomodating sloped walls or slabs, but what's here is awesome enough. Thanks!

Rob2309:
I would need to get familiar with sloped walls first, but I could try to add this in the future...

Rob2309:
Is there anything that needs to be done before this PR can be merged?

theoryshaw:
I think this is great, but differ to @Andrej730 for the ultimate review, and pull.

Andrej730:
@Rob2309 

Looks good to me.

A few notes:

1. Please apply black formatting - https://docs-unstable.bonsaibim.org/guides/development/code_style.html
2. ` Euler((0, 0, radians(180.0))).to_matrix().to_4x4()` - please use `Matrix.Rotation` for simplicity.

Rob2309:
Thanks for the feedback, the suggested changes should now be included.

Andrej730:
Thank you.

PR #6988, opened 2025-08-04
Fix Bonsai classification tests for newer bSDD version

c-mellueh:
separated #6970 to only the test fix


Andrej730:
@c-mellueh 

Thanks for the fixes.

Some notes:
1. `set_library_active` - can you please simplfiy this? No need to create dictionary for this - it can be just a for loop and raise `ValueError` at the end if there was no `return` from the loop.

2. In `test_add_clasification_reference_with_object_type` there's a bit of a mix-up, I believe - line `bpy.ops.bim.add_classification_reference_from_bsdd(obj="IfcSpace/Cube", obj_type="Object")` should go before `tool.Pset.get_pset_props(obj="IfcConveyorSegment/Cube", obj_type="Object").pset_name = uri` to keep it more readable.

3. Need to apply black formatter to this PR.

c-mellueh:
@Andrej730 should be fine now

PR #6991, opened 2025-08-04
fix leaking ifcgeom as_compound() calls

shmoodyyy:
Hello again, I've managed to get more of my paid-for time in the office approved for looking at the source code for our python service, some more memory leaks seemed to be popping up during long-lived service and calculations and after throwing our worker and pytest through memray the `add_element` showed up as an obvious problem (up to 2.1gB of allocated memory from `add_element` python invocations for 170mB in-memory file data)

At first I thought `IfcFile::open(...)` was leaking but turns out it was just the static schema pointers being lazy initialized, instead some places that invoked as_compound() forgot to clean up the OpenCascadeShape after copying the `TopoDS_Shape`.

I'll be looking to throw up a CICD pipeline tomorrow to package and push my fork into our private infrastructure for now and trying to monitor if any more leaking memory seems to be showing up with the little time I've got left to dedicate on this, but I don't think I'll be able to do more dedicated contributions until the next round of corpo-approvals outside of my freetime.

### DIFF
--- a/src/ifcgeom/IfcGeomRepresentation.cpp
+++ b/src/ifcgeom/IfcGeomRepresentation.cpp
@@ -222,7 +222,7 @@ IfcGeom::ConversionResultShape* IfcGeom::Representation::BRep::as_compound(bool 
 		builder.Add(compound, moved_shape);
 	}
 
-	return new ifcopenshell::geometry::OpenCascadeShape(compound);
+	return new ifcopenshell::geometry::OpenCascadeShape(std::move(compound));
 #else
         throw std::runtime_error("Not available without Open Cascade");
 #endif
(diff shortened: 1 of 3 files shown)


shmoodyyy:
addendum: I'm not sure how neccessary the move semantics for TopoDS_Shape are, I didn't look too deep into how much data each instance actually holds so I just threw it in on a gut feeling, feels right to me

aothms:
Awesome. Thanks. I don't think that the move semantics are necessary for TopoDS_Shape. It's a bit similar to a std::shared_ptr (the real data is in TShape()) but it doesn't hurt and aligns with the intent so ok to leave it in.

aothms:
There's a parenthesis too much too it seems:

```
/home/runner/work/IfcOpenShell/IfcOpenShell/src/ifcgeom/kernels/opencascade/IfcGeomTree.h:1369:122: error: expected ‘,’ or ‘;’ before ‘)’ token
 1369 |                                 TopoDS_Shape compound(std::move((ifcopenshell::geometry::OpenCascadeShape*)shp)->shape()));
```

shmoodyyy:
One opening paranthesis too little even, looks like I still haven't gotten my local build setup working properly, give me a moment.

Issue #6993, opened 2025-08-05
If you use `Bounding box Dimensions' you can't operate the move gizmo.


theoryshaw:
### Bug Description

If you use `Bounding box Dimensions' you can't operate the move gizmo. 

ping @falken10vdl 

### Attachments

https://github.com/user-attachments/assets/8c87bb46-e5f3-4574-a311-63879b119b73

### Debug and Error Output

```true

```

theoryshaw:
maybe related too, the global/local axis seems to get locked in place....

https://github.com/user-attachments/assets/ac7f487f-2e03-4e0e-aa3c-e90c00d726bd

falken10vdl:
@theoryshaw Yes I locked it on purpose so for a single object local coordinates are always displayed (and global in the case of multiple object selection) but probably not a good idea.
Let me go over it and remove this restriction so just the box is provided without touching the gizmo.
Thanks!

falken10vdl:
Here is a PR https://github.com/IfcOpenShell/IfcOpenShell/pull/6994 to address this. I just remove all locking of the gizmo. My initial idea was to force to see the local or the global gizmo so it was clear the dimensions in which direction they were written. However now that you have pointed out the awkwardness of that behaviour I think is silly to have it at all. I have removed it and now it is up tot he user to set the gizmo on whatever mode he/she likes.
Thanks!

falken10vdl:
Here a video with the gizmo working independently of the boundingbox decorator.

https://github.com/user-attachments/assets/25113e9b-5c80-4ed8-a2fc-9b3d592a6df4

Cheers!

falken10vdl:
@theoryshaw Maybe this cna no be closed?
Thanks!

theoryshaw:
Yes, thank you

PR #6994, opened 2025-08-05
Remove transform orientation settings in BoundingBoxDecorator

falken10vdl:
This is addressing https://github.com/IfcOpenShell/IfcOpenShell/issues/6993
We do not touch the gizmo at all with this PR. The colors of the bounding box should allow the user to understand if the bounding box dimensions are for the local or global coordinates.
In the case of single object selection the coordinates are local. in the case of multiple selectio the coordintes are global.
Cheers!


falken10vdl:
@theoryshaw Is this working fine? I removed all handling of gizmo. Actually like this the code is simpler/cleaner. I should have not force to mess around with the gizmo.
Thanks!

theoryshaw:
Works great!  Will keep it on, at all times now!

Issue #5776, opened 2024-11-20
Prompt for validation before deleting representation in the interface ?


Gorgious56:
While switching between representations I often miss-click and delete representations which in some cases leads to a bug. (eg deleting a representation while it is active)

![Animation3](https://github.com/user-attachments/assets/43031de9-004e-49ca-bcae-cb442070b3aa)

```
  File "C:\Users\\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\geometry\operator.py", line 366, in _execute
    core.remove_representation(
  File "C:\Users\\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\bonsai\core\geometry.py", line 180, in remove_representation
    geometry.delete_data(data)
  File "C:\Users\\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\bonsai\tool\geometry.py", line 109, in delete_data
    bpy.data.meshes.remove(data)
ReferenceError: StructRNA of type Mesh has been removed
```

So this report is twofold :

- Prevent user from deleting active representation
- Maybe add a prompt before deleting a representation ?
   Alternatively, use SHIFT + click for force deleting a representation ?

theoryshaw:
Undo, doesn't work here?

Gorgious56:
I try not using UNDO if I don't absolutely have to since it was and still is in my experience n° 1 cause for file desync. 

It seems after a bit of testing that in this case it is safe but the first point still stands IMO, it should either be forbidden to delete the active representation or be silently dealt with. Maybe it's just a UX problem and the delete button is too close to the "Make Active" button and it's a problem of buttery fingers ><

Also confirm-before-delete is the default for Blender in a lot of other instances, I think it would make sense to have it there, too.

falken10vdl:
@Gorgious56 , @theoryshaw  The code currently is robust to avoid this problem. I have however spotted that the "is_active" flag was not being updated for mapped representations. Please take a look to https://github.com/IfcOpenShell/IfcOpenShell/pull/7009
Cheers!

falken10vdl:
@Gorgious56 @theoryshaw Are you ok to clise this one? The PR (and the robust code slready in place) is already merged. This should not be a problem.
If something is removed by mistake it will not throw an error and one can UNDO safely.

[...]

PR #7009, opened 2025-08-09
set active flag for mapped representations

falken10vdl:
This was sparked by https://github.com/IfcOpenShell/IfcOpenShell/issues/5776
The code currently is much more robust than when the issue was raised and it will not break if one removed some representation. It will switch to the next one available. If there is only one representation for the object and it is removed, the object is still defined (the empty remains) without geometry.
However i have seen that for windows/doors the active representation was missing and all the representations had the "is_active" flag set to false.
Here is a proposal to update the "active_flag" in the case of Mapped Representations.

https://github.com/user-attachments/assets/2db1c93a-e117-43ce-90f3-8a916e7d71f0

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/geometry/data.py
+++ b/src/bonsai/bonsai/bim/module/geometry/data.py
@@ -118,14 +118,13 @@ class RepresentationsData:
         for representation in tool.Geometry.get_representations_iter(element):
             representation_type = representation.RepresentationType
             mapped_rep_id = None
-            
+
             if representation_type == "MappedRepresentation":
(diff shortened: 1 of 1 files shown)


Andrej730:
@falken10vdl 

Thanks, this ui bug was bugging me for awhile now, but never investigated it.

1. Black format is missing.
2. Can you please use `ifcopenshell.util.representation.resolve_representation` to simplify it?

falken10vdl:
@Andrej730 Please take a look at this updated PR to see if that is what you are asking for.
Cheers!

Andrej730:
This is it, thank you.

PR #7011, opened 2025-08-10
Fix predefined type bug

### LINKED ISSUE #7006: Incorrect PredefinedType logic
### Bug Description

In case there is Instance parameters set to:

PredefinedType = USERDEFINED
ObjectType = IfcObjectType
Image

and Type parameters set to:

PredefinedType = USERDEFINED
Element Type = Override (just as an example)
Image

It sets element type as:
IfcWall.Override - instead of IfcWall.IfcObjectType
Image

Only when value in Element Type is erased then it takes value from Object Type
Image
Image

From my perspective (if understood schema logic behind it correctly) Instance values should override type values not vice versa, right?

Uploaded ifc in .txt format to reconstuct:
[4Walls.txt](https://github.com/user-attachments/files/21681051/4Walls.txt)

### Attachments

_No response_

### Debug and Error Output

```true

```

c-mellueh:
fixes #7006

### DIFF
--- a/src/bonsai/bonsai/bim/module/root/data.py
+++ b/src/bonsai/bonsai/bim/module/root/data.py
@@ -255,6 +255,8 @@ class IfcClassData:
         if not element:
             return False
         if element_type := ifcopenshell.util.element.get_type(element):
+            if element_type == element:
+                return False
             # Allow for None due to https://github.com/buildingSMART/IFC4.3.x-development/issues/818
             return ifcopenshell.util.element.get_predefined_type(element_type) not in ("NOTDEFINED", None)
         return False
--- a/src/bonsai/bonsai/bim/module/root/ui.py
+++ b/src/bonsai/bonsai/bim/module/root/ui.py
@@ -60,15 +60,14 @@ class BIM_PT_class(Panel):
                 row.operator("bim.reassign_class", icon="CHECKMARK")
                 row.operator("bim.disable_reassign_class", icon="CANCEL", text="")
 
-                # Don't show PredefinedType  / ObjectType if Entity is Typed (See https://github.com/IfcOpenShell/IfcOpenShell/issues/7006)
                 predefined_types = root_prop.get_ifc_predefined_types(rprops, context)
-                if tool.Type.is_entity_typed(tool.Ifc.get_entity(obj)):
-                    predefined_types = []
-
+                # If Entity has inherited Predefined Type, disable PredefinedType / ObjectType dropdown to forbid double typing. See #7006)
+                enable_predef_types = not IfcClassData.data["has_inherited_predefined_type"]
                 self.draw_class_dropdowns(
                     context,
                     predefined_types,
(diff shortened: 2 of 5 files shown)


Andrej730:
@c-mellueh 

1. You can refer to issue in code as `#7006` for simplicity. Or maybe this comment can be more generic, so reader wouldn't need to investigate the issue to learn why this needed?

2. We don't access IFC directly from UI code - it would be too expensive. You need to access IFC in `root/data.py`, cache the result and access cached result from the UI. See any other example with `IfcClassData` in the ui code. 

    There's also `ifcopenshell.util.element.get_type()` for this.

3. Disabling dropdown seems to be better UX, so it will be easier for user to spot the difference. Though unfortunately, there's no poll check on properties to communicate to user why it's disabled.

4. I believe IfcWall.PredefinedType description is inaccurate (see https://github.com/buildingSMART/IFC4.3.x-development/issues/977) and it's actually referring to more general concept that does allow IfcWall.PredefinedType for typed instances when `IfcWall.PredefinedType == NOTDEFINED`. 
So this code should also follow the same logic - not just check whether entity is typed, but also check the type's predefined type.

5. `assign_type` fix makes sense to keep IFC valid, Can you please add a test for it?
But `assign_type` code should also should be adjusted for [4].

c-mellueh:
@Andrej730 thanks for the feedback! 
I implemented your requests. 
I used the allready existing `has_inherited_predefined_type` from `IfcClassData`. 
I needed to update the getter function of `has_inherited_predefined_type` a little because in the existing state all `IfcTypeObject`s return True. 
If I understand the name and intend of the function correctly than this is wrong because an `IfcWallType` doesn't inherit their predefined type, it creates it. 
If we'd keep the function as is, it won't be usable for our cause to disable the dropdown UI because the dropdown would be disabled for all `IfcTypeObject`s.

Andrej730:
@c-mellueh 

1. `if predefined_type != "NOTDEFINED":` - it also can be `None`, see https://github.com/buildingSMART/IFC4.3.x-development/issues/818 and `has_inherited_predefined_type` implementation.

2. `predefined_types = root_prop.get_ifc_predefined_types(rprops, context)` - seems unnecessary now.

3. `if hasattr(obj, "ObjectType")` - unnecessary, if it's `IfcObject`, then it's definitely has the attribute.

4. 
> I needed to update the getter function of `has_inherited_predefined_type` a little because in the existing state all `IfcTypeObject`s return True.
> If I understand the name and intend of the function correctly than this is wrong because an `IfcWallType` doesn't inherit their predefined type, it creates it.

Makes sense, I'd say it was just a bug.

c-mellueh:
@Andrej730 should work now

Andrej730:
Looks good! Can you please add a test for new behaviour of `assign_type`?

c-mellueh:
@Andrej730 done.

Andrej730:
Thank you.

Issue #6960, opened 2025-07-29
IDS ifctester Entity facet in requirements breaks on cardinality check


ay-ex:
### Bug Description

In `ifctester`, when I try to create an IDS for the following check: 
`All Walls with Attribute Name of value == "WALL 1" must be of Entity IFCWALL`
as shown below:

```python
import ifcopenshell
from ifctester import ids, reporter
from pprint import pprint


prj_ids = ids.Ids(title="class_check")

spec = ids.Specification(name="class_check_wall_1")
spec.applicability.append(ids.Attribute(name="Name", value="WALL 1"))
rule = ids.Entity(name="IFCWALL")
spec.requirements.append(rule)
prj_ids.specifications.append(spec)

ids_result = prj_ids.to_xml(r"d:\tmp\ids\class_check.ids")
ifc_model = ifcopenshell.open(r"d:\tmp\ids\sample.ifc")
prj_ids.validate(ifc_model)
results = reporter.Json(prj_ids).report()

pprint(results)
```
I get the following error:
```shell
Traceback (most recent call last):
  File "<blender_console>", line 1, in <module>
  File "C:\Users\user.name\AppData\Roaming\Blender Foundation\Blender\4.4\extensions\.local\lib\python3.11\site-packages\ifctester\reporter.py", line 256, in report
    specification_report = self.report_specification(specification)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\user.name\AppData\Roaming\Blender Foundation\Blender\4.4\extensions\.local\lib\python3.11\site-packages\ifctester\reporter.py", line 342, in report_specification
    description=requirement.to_string("requirement", specification, requirement),
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\user.name\AppData\Roaming\Blender Foundation\Blender\4.4\extensions\.local\lib\python3.11\site-packages\ifctester\facet.py", line 133, in to_string
    if requirement and requirement.cardinality == "prohibited":
                       ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Entity' object has no attribute 'cardinality'
```

The code in the offending line checks for `.cardinality`, which the `Entity` class does not have.

A proposed fix could be: Checking if the requirement is of class `Entity` and exclude it from cardinality check:

[...]

PR #7013, opened 2025-08-11
fix no cardinality check on entity requirements

ay-ex:
as proposed in #6960 
to fix the failing cardinality check in case the requirement is an Entity, 
which does not have `.cardinality`.


Moult:
The solution is correct but I'm going to commit a slightly different approach. I've just learned a nuance in IDS that requirements are totally ignored in the prohibited case that I missed in a very, very annoying `+14030 -15697` commit upstream.

ay-ex:
@Moult : thank you for the fix! 🚀

Issue #6612, opened 2025-04-23
Would be nice if links in texts were clickable.


theoryshaw:
### Feature Description

Would be nice if links in texts were clickable.

![Image](https://github.com/user-attachments/assets/074bdc8d-6285-4cf3-892f-0e63ede5edc2)

falken10vdl:
@theoryshaw Please take a look to PR https://github.com/IfcOpenShell/IfcOpenShell/pull/7044
To make things simple I have just added the property to the annottion so one URL can be stored per annotation. I hope that is not a limitation.

<img width="2773" height="1701" alt="Image" src="https://github.com/user-attachments/assets/ad4b763e-80c5-45e6-a79e-c2201a1caf85" />

Cheers!

theoryshaw:
Hi Vicente, 
IMHO, this seems a little clunky, as you might have multiple links in a text string.
...

Just thinking outloud, How about this… just use a simple regex to detect `http://` or `https://` in annotation text and wrap it in an `<a>` tag during SVG export

Or this… bring in a library like `urlextract` that specializes in finding URLs, so it handles edge cases and future-proofs link detection, though it means adding a dependency.

Or … go one step further and use a Markdown library (`markdown`, `mistune`, or `markdown-it-py`), so links can be written naturally as `[Label](https://link)` and later expanded into clickable SVG anchors — plus you gain support for richer text formatting down the road.

falken10vdl:
@theoryshaw 
Maybe the last option is the nicest? 
Is there any preference for any particular one? 
Also I think that maybe we should just do the links to target a new browser window so we remove that option and we then have no need to use EPSet and can just parse the markdown.
What do you think?
Cheers!

theoryshaw:
Using markdown would be nice!

@Andrej730 Any thoughts here?

falken10vdl:
I have refactored it using markdown-it-py. It actually has a cleaner implementation I think.

https://github.com/user-attachments/assets/591534ef-f86f-494b-b2c3-910d4cc48e13

Cheers!

theoryshaw:
@falken10vdl 
Hi, it looks like `\n` is not being taken into account. 
Not sure, if this is related to the markdown thing. 

https://github.com/user-attachments/assets/6e820148-6fe8-4317-b64c-eb4fb0a234e8

theoryshaw:
[...]

PR #7044, opened 2025-08-20
Add hyperlinks to annotations (using markdown-it-py)

falken10vdl:
This is a proposed PR to address https://github.com/IfcOpenShell/IfcOpenShell/issues/6612
The URL in the annotaion is saved in the same EPset_Annotation as Newline_At

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/data.py
+++ b/src/bonsai/bonsai/bim/module/drawing/data.py
@@ -368,10 +368,6 @@ class DecoratorData:
             }
             literals_data.append(literal_data)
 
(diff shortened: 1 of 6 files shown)


falken10vdl:
Refactor the whole PR using natural Markdown syntax. 
The new approach is much better since it has:
- Multiple links per annotation using Markdown syntax
- Natural writing: [Link Text](https://url.com)
- No UI changes needed - just write markdown in text
- Future-proof - can extend to support bold, italic, etc.

The only thing is that it adds a requirement to have markdown-it-py installed
Cheers!

falken10vdl:
combined into https://github.com/IfcOpenShell/IfcOpenShell/pull/7317
Cheers!

Issue #6157, opened 2025-02-14
IfcFixedReferenceSweptAreaSolid is not supported


Moult:
Civil pipes coming from 12D apparently use advanced swept solids using IfcFixedReferenceSweptAreaSolid

[pipe.ifc.txt](https://github.com/user-attachments/files/18793820/pipe.ifc.txt)

Krande:
Hey, I have started using IfcFixedReferenceSweptAreaSolid for weld geometries in adapy. So for the last week I have been fighting with what I believed was poorly handled transformations on my part. But reading this issue, I understand it might not be well supported in IfcOpenShell :) I tested some ifc files I created in trimble connect and to my surprise they looked correct! 

@aothms do you have any spare time to look at implementing support for it? If not I might be able to dedicate a few days of work in attempting to add support for it. 

I will nonetheless provide some sample geometries in a few hours for reference.

Best Regards
Kristoffer

aothms:
Hi @Krande since you seem to be really on top of this, I think it's most efficient if you apply the necessary changes. Honestly it shouldn't be so hard, but the complexity is in testing a sufficient amount of variations and massaging the data so that opencascade is happy, which is hard to do interactively back and forth.

There are a couple of aspects:

- FixedRefSweptArea is now only intended to be used in some "interpolated mode" as sweeps over alignment curves. Alignment curves are piecewise connected curves used in isolation to express, horizontal movement, vertical movement and angular rotation around. 
- There is already support for SurfaceCurveSweptAreaSolid (which is a more complex variant basically) by using open cascade's PipeShell which gets mapped via `taxonomy::sweep_along_curve`.


Todo:

- Implement an else block [0] to so that we handle the case where a non-alignment curve is the basis for FixedRefSweep
- Add a direction3::ptr (I think) to our taxonomy class (could also be std::variant<surface, direction> if you want to semantically that they have to be used exclusively) [1]
- Use the correct mode in [2] (see C++ header snippet in [3] - I think just pass the direction) for sweeping


[0] https://github.com/IfcOpenShell/IfcOpenShell/blob/8903d6a85c2ea045ee02c7ee3579e4909a8c54ad/src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp#L37
[...]

PR #7058, opened 2025-08-26
feat: improved support for IfcFixedReferenceSweptAreaSolid

Krande:
Hey @aothms, here's a PR based on the discussion in https://github.com/IfcOpenShell/IfcOpenShell/issues/6157. 

I made a crude hardcoded conditional simply to force the code into producing some geometry and to see if I can use it to understand how it's wired together. 

I'd appreciate any help along the way (I don't regularly write c++), so please watch out for poorly written code :) 

The pixi and cmakepresets.json stuff is simply there for me to easily create an environment with all the dependencies and start running the code as fast as possible. So that can be ignored if you don't want it.


src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp
```diff
@@ -17,106 +17,133 @@
...
+    auto ref = taxonomy::cast<taxonomy::direction3>(map(inst->FixedReference()));
+    auto profile = taxonomy::cast<taxonomy::face>(map(inst->SweptArea()));
+
+    auto loft = taxonomy::make<taxonomy::loft>();
+    // @todo intialize as default
+    loft->axis = nullptr;
+    // Check if this instance has alignment elements
+    if (has_alignment_elements(inst)) {
+        // @todo currently only the case is handled where directrix returns a piecewise_function
+        if (auto fn = taxonomy::dcast<taxonomy::function_item>(dir)) {
```
aothms: I don't think we need `has_alignment_elements()` actually. When the cast to a function_item fails we can assume it uses a regular IfcCurve as its directrix.

It's a bit tricky though as dcast also 'upgrades' (I think it might upgrade a polyline to a function item for example, so better use the more plain std::dynamic_pointer_cast from the stdlib. (Maybe that's what urged you to write your own function, not sure?)


```suggestion
        if (auto fn = std::dynamic_pointer_cast<taxonomy::function_item::ptr>(dir)) {
```

src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp
```diff
@@ -17,106 +17,133 @@
...
+                    loft->children.back()->matrix->components() = (m4b * profile->matrix->ccomponents()).eval();
+                } else {
+                    loft->children.back()->matrix->components() = m4b;
+                }
+            }
+        }
+    } else {
+        // TODO: Implement handling for non-alignment curves using sweep_along_curve
+        auto sweep = taxonomy::make<taxonomy::sweep_along_curve>(
+            nullptr, // matrix4::ptr - no transformation needed
```
aothms: There is a `Position` attribute that should probably be carried over to the sweep object, interestingly it appears to be unused here. Have a look at the handling of IfcExtrudedAreaSolid to see how it's applied.

src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp
```diff
@@ -17,106 +17,133 @@
...
+            profile, // face::ptr - the profile to sweep
+            nullptr, // item::ptr surface - not used for fixed reference sweep
```
aothms: Probably nicer to make a separate constructor for this application.

src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp
Krande: Yeah, the reason I wrote a separate function was that the conditional never moved to the else condition. I'll test your suggestion and get back to you!

Krande: using `std::dynamic_pointer_cast` seems to work! Added this in https://github.com/IfcOpenShell/IfcOpenShell/pull/7058/commits/48358297005c8955551f966bf2b632f2a07a3bd5

src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp
Krande: Okay, I think I got it working now

https://github.com/IfcOpenShell/IfcOpenShell/blob/8420b98f3fe8f9b8742982e22198fb58b2cc7b97/src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp#L118-L129

src/ifcgeom/mapping/IfcFixedReferenceSweptAreaSolid.cpp
Krande: I added a new constructor in https://github.com/IfcOpenShell/IfcOpenShell/pull/7058/commits/d01bf9690d2df49ccb6b6fb2c77586a928ce7a22#diff-dda4c9b55294de4b49f4172a6c675f50a7bfbee4d79659a3499531f2199ec048.

Krande:
Hey @aothms, 

I have updated the code based on your comments. I also added a few ifc test files using `IFCFIXEDREFERENCESWEPTAREASOLID` temporarily placed in `test/input_temp` and a few tests to assert the tesselated results. 

To be perfectly honest, I am not 100% sure about the resulting world-coord orientation of the geometries in the IFC examples I made ([simple_sweep_1.ifc](https://github.com/Krande/IfcOpenShell/blob/9dd7b488d65b67bb60ba656dd7661e17bc471542/test/input_temp/simple_sweep_1.ifc) and [simple_sweep_2.ifc](https://github.com/Krande/IfcOpenShell/blob/9dd7b488d65b67bb60ba656dd7661e17bc471542/test/input_temp/simple_sweep_2.ifc)). 

I went so much back and forth in changing the placements and orientations when trying to figure out why the tessellation of the geometries failed using IfcOpenShell that they currently are a mess :) 

Nonetheless, as you will observe, there are differences in the orientation of the tessellated geometry representations by IfcOpenShell and Trimble Connect. 

Let me know what you think and what I should prioritize next?

## simple_sweep_1



Trimble Connect
image

IfcOpenShell (https://github.com/IfcOpenShell/IfcOpenShell/pull/7058/commits/d01bf9690d2df49ccb6b6fb2c77586a928ce7a22)

image


## simple_sweep_2

In this example, it seems like Trimble Connect fails to maintain the swept profile along the sweep directrix, while IfcOpenShell makes what appears to be a successful sweep. 

Trimble Connect

image

IfcOpenShell (https://github.com/IfcOpenShell/IfcOpenShell/pull/7058/commits/d01bf9690d2df49ccb6b6fb2c77586a928ce7a22)

image


## Pipe 1

In this example, Trimble Connect seems to do fine, while IfcOpenShell creates some weird surface artifacts. Not quite sure what to make of it. I was thinking some numeric precision issue given the large x,y coordinates. But then again, they're not that large. .

Trimble Connect

image

IfcOpenShell (https://github.com/IfcOpenShell/IfcOpenShell/pull/7058/commits/d01bf9690d2df49ccb6b6fb2c77586a928ce7a22)

image

src/ifcgeom/taxonomy.h
```diff
@@ -1210,12 +1210,16 @@ typedef item const* ptr;
...
 				item::ptr curve;
+                direction3::ptr direction = nullptr;
 
 				virtual sweep_along_curve* clone_() const { return new sweep_along_curve(*this); }
 				virtual kinds kind() const { return SWEEP_ALONG_CURVE; }
 
 				sweep_along_curve(matrix4::ptr m, face::ptr basis, item::ptr surf, item::ptr crv) : sweep(m, basis), surface(surf), curve(crv) {}
 
+			    // New constructor for fixed reference swept area solid
+			    sweep_along_curve(matrix4::ptr m, face::ptr profile, item::ptr directrix) : sweep(m, profile), surface(nullptr), curve(directrix) { }
```
aothms:
```suggestion
			    sweep_along_curve(matrix4::ptr m, face::ptr profile, item::ptr directrix, direction3::ptr ref) : sweep(m, profile), surface(nullptr), curve(directrix), direction(ref) { }
```

src/ifcgeom/taxonomy.h
```diff
@@ -1210,12 +1210,16 @@ typedef item const* ptr;
 
 				item::ptr surface;
 				item::ptr curve;
+                direction3::ptr direction = nullptr;
 
 				virtual sweep_along_curve* clone_() const { return new sweep_along_curve(*this); }
 				virtual kinds kind() const { return SWEEP_ALONG_CURVE; }
 
 				sweep_along_curve(matrix4::ptr m, face::ptr basis, item::ptr surf, item::ptr crv) : sweep(m, basis), surface(surf), curve(crv) {}
```
aothms:
```suggestion
				sweep_along_curve(matrix4::ptr m, face::ptr basis, item::ptr surf, item::ptr crv) : sweep(m, basis), surface(surf), curve(crv), direction(nullptr) {}
```

aothms:
> ```
> #34=IFCCARTESIANPOINT((289081.570466963,5822857.54453541,118.886));
>                                         1234567
> ```

7 significant digits is about what you can expect for 32bit float. I don't know how you visualize, but most likely you're doing something with 32bit floats in opengl and then you can expect this. I would propose not to deal with this as part of this PR. Just change the file to have smaller polyline coordinates.

Please add the files to `src/ifcopenshell-python/test/fixtures/geom` and the test to `src/ifcopenshell-python/test/geom`, then they can stay there.

> In this example, it seems like Trimble Connect fails to maintain the swept profile along the sweep directrix, while IfcOpenShell makes what appears to be a successful sweep.

For me the transformation doesn't really help to understand, maybe remove it (should be handled correctly as you've written it). 

The thing is, it's not impossible to construct degeneracies using FixedRefSweptA. Any time the normal plane of the curve is orthogonal to the provided fixed reference attributes you'll end up with a degenerate sweep because the profile's local X should track the projection of local reference onto the local normal plane of the directrix. So if they are orthogonal you end up with a zero vector and an undefined X-axis. From reading along I couldn't really figure out if this is the point case. Maybe it wouldn't hurt to verify whether this is the case in your example. Could be Trimble is correct?

Krande:
>7 significant digits is about what you can expect for 32bit float. I don't know how you visualize, but most likely you're doing something with 32bit floats in opengl and then you can expect this. I would propose not to deal with this as part of this PR. Just change the file to have smaller polyline coordinates.

> Please add the files to src/ifcopenshell-python/test/fixtures/geom and the test to src/ifcopenshell-python/test/geom, then they can stay there.

Done in https://github.com/IfcOpenShell/IfcOpenShell/pull/7058/commits/1cd32be2fc7e2d59eb296c1fd34ae6e8ce164d7e
By simply changing the position of the pipe the geometry is fine. 

image


> The thing is, it's not impossible to construct degeneracies using FixedRefSweptA. Any time the normal plane of the curve is orthogonal to the provided fixed reference attributes you'll end up with a degenerate sweep because the profile's local X should track the projection of local reference onto the local normal plane of the directrix. So if they are orthogonal you end up with a zero vector and an undefined X-axis. From reading along I couldn't really figure out if this is the point case. Maybe it wouldn't hurt to verify whether this is the case in your example. Could be Trimble is correct?

It might be. The simple geometry cases simply use arbitrary sweep paths. So I can take a closer look and see if the path's should be slightly revised if they are poor examples to follow.

Krande:
I added a few things to please the linter and made some improvements to some of the tests using pathlib.Path and relative paths. But I think the remaining failing tests are unrelated to this PR. 

And I made a slight change to the sweep path in the `simple_sweep_2` test so that it seems like Trimble Connect no longer has an issue with the geometry and that it is identical with the geometry made by IfcOpenShell. 

@aothms, le me know if there's anything else you think is needed in this PR? 

And let me know if you prefer to keep the pixi stuff out? 

I'd recommend that we keep the pixi stuff in. First of it doesn't interfere with any other code and pixi is quite useful! If you have pixi installed and open a terminal from the Ifcopenshell root directory and run

- `pixi run test` installs the necessary dependencies, compiles and runs the test suite on the compiled code. 
- `pixi run lint` will install necessary deps and perform linting checks with ruff and black

aothms:
Nice, really nice, thanks for all the work on this.

Not sure about the pixi stuff, it looks helpful, but I'm a bit in doubt about the lock file? Do we need it to be so deterministic, are the risks to it when it gets stale (-ish..)?

Krande:
> Not sure about the pixi stuff, it looks helpful, but I'm a bit in doubt about the lock file? Do we need it to be so deterministic, are the risks to it when it gets stale (-ish..)?

Well the lock file is automatically produced similar to what nodejs does. In my mind the lock file is what makes the whole thing reproducible. The dependencies in conda-forge are immutable, so in theory it shouldn't get stale. Leaving too much wiggle room in versions of dependencies is often what breaks things. This strict pinning removes that risk. 

But if you want we can move the whole pixi stuff out and take that discussion in a different PR:)

aothms:
Ok, I believe you. Maybe in a new PR write a bit of dev docs on how people can use pixi.

PR #7068, opened 2025-08-28
Fix/issue 6099 python occ compatibility

### LINKED ISSUE #6099: PythonOCC >=7.8.0 ifcopenshell.geom.serialise doesn't work because BRepTools_ShapeSet.WriteToString signature changed
I'm trying to create `IfcSpace` from walls & slabs and have successfully created the TopoDS_Compound shape. However, while trying to serialise this shape and create a representation, I get the following error message
```
(Pdb) space_shape
<class 'TopoDS_Compound'>
(Pdb) ifcopenshell.geom.serialise(f.schema, space_shape)
*** TypeError: BRepTools_ShapeSet.WriteToString() missing 1 required positional argument: 'shape'

```
I'm currently using ifcopenshell version 0.8.0 and pythonocc version 7.8.1.1.

I'm wondering if this is due to any of the recent updates as I could do serialisation as above with ifcopenshell version 0.7.0 and pythonocc 7.7.0.

How do I serialise this shape correctly now

EDIT:
I've also tried serialising with TopoDS_Solid extracted from TopoDS_Compound, but even that has the same error output

smr02:
fixes #6099

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
+++ b/src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
@@ -18,6 +18,7 @@
 
 
 from __future__ import annotations
+import inspect
 import random
 import operator
 import warnings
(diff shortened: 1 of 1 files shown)


aothms:
Many thanks :)

Issue #7038, opened 2025-08-18
Feature Request: Enhanced Classification Display in Object Information Tab


bsmithuk:
### Feature Description

**Summary**
Add the classification system to the Object Information tab's classification display to provide better context and match functionality available in other IFC viewers like Solibri. This request complements the existing classification viewing functionality in the Project Overview tab by bringing that context to the object level.

**Current Behaviour**
In the Object Information tab, only the classification reference and name are displayed for classified objects. Users can see which classification systems exist in the IFC file via the Project Overview tab, but this information is not readily available when inspecting individual objects.

**Requested Enhancement**
Display the classification system alongside the existing classification reference and name in the Object Information tab.

Current Display Format:
Classification:
- Reference: En_20
- Name: Administrative, commercial and protective service entities

<img width="433" height="149" alt="Image" src="https://github.com/user-attachments/assets/2d6cdef9-3cdc-46e7-9e79-77fa31175393" />

Requested Display Format:
Classification:
- System: Uniclass Classification
- Reference: En_20
- Name: Administrative, commercial and protective service entities

<img width="832" height="189" alt="Image" src="https://github.com/user-attachments/assets/f74f889c-8167-4e22-b7a7-a67809064424" />

**Priority**
Low - Quality of life improvement

falken10vdl:
@bsmithuk Can you provide an ifc file with some objects that have been classified so that I could use to test this, please?
Thanks!

bsmithuk:
@falken10vdl I've submitted model example via IfcOpenShell with same ID as this issue ID.

falken10vdl:
@bsmithuk ok I do not follow you :) Do you have a link you can paste here from where I can download it?
Thanks!

bsmithuk:
Sorry @falken10vdl normally if we need to share a file that contains commercial information it can be uploaded to lead developers via a google form. @Andrej730 are you able to share this directly with @falken10vdl or I can make a bespoke example on the weekend.

falken10vdl:
Ah, now I understand. Thanks for the clarification. I think ideally an example file which is not holding any commercial information will be great as it can later be used as a reference if needed for other UI tweaks.
Thanks!

bsmithuk:
<img width="441" height="211" alt="Image" src="https://github.com/user-attachments/assets/b35ac3c1-be45-48cd-9319-bb37cd59628f" /> 
vs Solibri

[...]

PR #7072, opened 2025-09-01
Add  classification system to the Object Information, Classification References Subpanel

falken10vdl:
This is to address issue https://github.com/IfcOpenShell/IfcOpenShell/issues/7038

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/classification/data.py
+++ b/src/bonsai/bonsai/bim/module/classification/data.py
@@ -104,6 +104,17 @@ class ReferencesData:
         if element:
             for reference in ifcopenshell.util.classification.get_references(element):
                 data = reference.get_info()
+
+                if reference.ReferencedSource:
+                    if reference.ReferencedSource.is_a("IfcClassification"):
+                        data["ClassificationSystemName"] = reference.ReferencedSource.Name
+                    elif reference.ReferencedSource.is_a("IfcClassificationReference"):
+                        current = reference.ReferencedSource
+                        while current.ReferencedSource and current.ReferencedSource.is_a("IfcClassificationReference"):
+                            current = current.ReferencedSource
+                        if current.ReferencedSource and current.ReferencedSource.is_a("IfcClassification"):
+                            data["ClassificationSystemName"] = current.ReferencedSource.Name
+
                 del data["ReferencedSource"]
                 results.append(data)
         return results
--- a/src/bonsai/bonsai/bim/module/classification/data.py
+++ b/src/bonsai/bonsai/bim/module/classification/data.py
@@ -104,17 +104,7 @@ class ReferencesData:
         if element:
             for reference in ifcopenshell.util.classification.get_references(element):
                 data = reference.get_info()
-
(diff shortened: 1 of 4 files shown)


Moult:
Awesome! Can help rewrite using `util.classification.get_classifiation(reference)`?

Moult:
Why the columns and splits? Why not just another row call?

```python
row = self.layout.row(align=True)
row.label(text=reference["Classification"])
row.label(text=name, icon="ASSET_MANAGER")
row.label(text=reference["Name"] or "")
```

falken10vdl:
> Why the columns and splits? Why not just another row call?
> 
> ```python
> row = self.layout.row(align=True)
> row.label(text=reference["Classification"])
> row.label(text=name, icon="ASSET_MANAGER")
> row.label(text=reference["Name"] or "")
> ```

It was just to control the alignment. With just row calls it looks like this:
image

While if we use split/columns it looks like this:
image

So up to you. I am a telecom engineer so my aesthetics and ux design might be too table oriented :)
If you want I can update with something like:

```
    def draw_reference_ui(self, reference: dict[str, Any]) -> None:
        row = self.layout.row(align=True)


        if reference.get("ClassificationSystemName"):
            row.label(text=reference["ClassificationSystemName"], icon="OUTLINER_COLLECTION")
        else:
            row.label(text="")

        if self.file.schema == "IFC2X3":
            name = reference["ItemReference"] or "No Identification"
        else:
            name = reference["Identification"] or "No Identification"
        row.label(text=name, icon="ASSET_MANAGER")

        row.label(text=reference["Name"] or "")

        if reference["Location"]:
            row.operator("bim.open_uri", icon="URL", text="").uri = reference["Location"]
        if not self.props.active_reference_id:
            op = row.operator("bim.enable_editing_classification_reference", text="", icon="GREASEPENCIL")
            op.reference = reference["id"]
        op = row.operator("bim.remove_classification_reference", text="", icon="X")
        op.reference = reference["id"]
        op.obj = self.obj
        op.obj_type = self.data.obj_type
```

Cheers!

falken10vdl:
> Can help rewrite using `util.classification.get_classifiation(reference)`?

I am not sure I have followed you :) Do you mean to update draw_reference_ui method by using ifcopenshell.util.classification.get_classification(reference) to get the classification system name, rather than relying on the "ClassificationSystemName" key set in data extraction.
Something like this?
in data.py:
```

    @classmethod
    def references(cls) -> list[dict[str, Any]]:
        results = []
        element = cls.get_element()
        if element:
            for reference in ifcopenshell.util.classification.get_references(element):
                data = reference.get_info()
                data["ifcClassificationReference"] = reference
                del data["ReferencedSource"]
                results.append(data)
        return results
```

and then in uy.py:

```
...
        classification_entity = ifcopenshell.util.classification.get_classification(reference["ifcClassificationReference"])
        classification_name = classification_entity.Name if classification_entity else ""
...
```

Cheers!

falken10vdl:
@Moult please take a look to the updated code
Cheers!

PR #7081, opened 2025-09-02
feat: add comparison property to ifc_filter for type "query"

falken10vdl:
This is to address discussion https://community.osarch.org/discussion/3102/creating-a-query-for-search-group-qto
Currently the "query" type only allows to provide the = and the != operators within the value field.
image

This is a proposal to split and add the comparison field like for the "property" type
image

Cheers!


theoryshaw:
Thanks @brunoperdigao for the review and pull!

PR #7093, opened 2025-09-05
Organize drawings under a parent called DRAWINGS in ifc groups and ifc documents

falken10vdl:
This is a proposal to have a common parent called DRAWINGS for the ifc groups and ifc documents generated by Bonsai when dealing with drawings.
image

Cheers!

### DIFF
--- a/src/bonsai/bonsai/core/drawing.py
+++ b/src/bonsai/bonsai/core/drawing.py
@@ -302,9 +302,23 @@ def add_drawing(
         context=drawing.get_body_context(),
         ifc_representation_class=None,
     )
+    
+    drawings_parent_group = None
+    for group in ifc.get().by_type("IfcGroup"):
+        if group.Name == "DRAWINGS" and group.ObjectType == "DRAWINGS":
+            drawings_parent_group = group
+            break
+    
+    if not drawings_parent_group:
+        drawings_parent_group = ifc.run("group.add_group")
+        ifc.run("group.edit_group", group=drawings_parent_group, attributes={"Name": "DRAWINGS", "ObjectType": "DRAWINGS"})
+    
     group = ifc.run("group.add_group")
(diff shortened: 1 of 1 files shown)


falken10vdl:
@theoryshaw Do you think this makes sense? Of course it will not change the current ones but new ones would fall in the folder so removing old ones and regenerating them will bring them under the DRAWINGs group

falken10vdl:
@theoryshaw This is rather simple but old. Do you think it makes sense or should be dropped?
Cheers!

theoryshaw:
i think it's a great idea. 

for whatever reason, not all my drawings are being categorized under 'DRAWINGS' however...
https://hub.openingdesign.com/OpeningDesign/Highland_Haven/src/commit/97f158b95d543a0bd1e0ad6e897855ca86fb1618/Open/Models_and_CAD/BlenderBIM/Highland_Haven.ifc

image

falken10vdl:
The PR will not modify the parent on existing ones. Only if you create drawings once the PR is loaded they will be categorized under DRAWINGS folder. Is that what you experience?
Cheers!

theoryshaw:
I gotcha.

Does it make sense to force this on pre-PR drawings?  Not sure. 

either way, works good for me.

falken10vdl:
In my opinion, this "migration" would typically be handled if we had version control with increasing numbers. 

In the IFC file I see that there is for example: FILE_NAME('Untitled.ifc','2026-07-03T14:32:37+02:00',(''),(''),'IfcOpenShell 0.0.0','Bonsai 0.8.5-alpha251216-78cec9c','Nobody'); so we can fetch the file version. If this was incrementally handled, then we can bump up the version one and put code to do this "migration" if the file is from the older format.
This way we could keep the "migration" code for some time before simply removing that to not have code leftovers.

I do not know if there is an easy way to know if a uuid is older or newer than another so if we want this, we would need to keep this "migration" always. Is this what we want?

Cheers!

theoryshaw:
cool. :)

PR #7106, opened 2025-09-09
Operators to help add intelligent tags in Text literal annotations

falken10vdl:
This is a proposal to help generate relevant expressions for the literals in Text annotations. It provides the list of available properties or quantities if there are products assigned for the user to choose.

https://github.com/user-attachments/assets/28b85d34-ed2e-44cb-a60b-e922fc2f4886

Cheers!


theoryshaw:
Awesome!  Anyway for these helpers to be available to the tag type as well?  Most of my tags are driven by a type, and are not instances. 

...

Pulling the object's materials attributes would be awesome as well.  Perhaps for a future roadmap. :)

falken10vdl:
@theoryshaw I have added both the attributes for the assigned product and for the assigned product type. Is that what you are asking?

image

I have setup a filter (by default no filter) so that the attributes shown are meaningful. My knowledge on ifc is quite limited so I have just put a fairly simple list...

Object's materials have a lot of possibilities. What would be the most relevant info?

falken10vdl:
I am working to mimic as much as possible all options available in get elements and formatting of selector syntax so it is easy to buildup the queries

theoryshaw:
> Is that what you are asking?

I meant when you select, and are working with the 'IfcTypeProduct` itself. 

Maybe in this mode, there's an eyedropper where you can select a object to pull out the suggestions... not sure.


https://github.com/user-attachments/assets/bcdf485b-cedd-4747-8466-7b33a24c6f98

falken10vdl:
I think I understand now. Currently it is true that the helper is based on the fact that there is an object assigned from which it will pull the different parameters (psets, qsets, material, etc.) In the case there is not assigned product I agree that it makes sense to have an eyedropper to pull out suggestions...
Let me include it. Currently I have set the PR to draft as I am changing it to try to get basically as much as possible following the selector syntax:
image
Cheers!

falken10vdl:
@theoryshaw I have added the eyedropper so one can create the literals based on the attributes of other products, which is useful if one will change the assigned product, whic is the case for the IfcProductType. I am still trying to get the helper tool for all the fields of the selector syntax.

https://github.com/user-attachments/assets/19f7f3b6-08df-4e73-9eb8-759bf40b9305

Cheers!

falken10vdl:
I think I have added support for the whole getting-element-values of the selector syntax. I have also added some help information to help the user with how the syntax is used.
Cheers!

falken10vdl:
Combined into https://github.com/IfcOpenShell/IfcOpenShell/pull/7317
Cheers!

PR #7130, opened 2025-09-15
Update ifcconvert.rst with new file format support

chuongmep:
This pull request adds support for converting IFC files to DuckDB format in the documentation table, making it easier for users to see available conversion options.
https://github.com/chuongmep/ifc-2-duckdb
New conversion format added:

* Added a new row to the conversion options table in `src/ifcopenshell-python/docs/ifcconvert.rst` to indicate support for converting `.ifc` files to `.duckdb` format using the `ifc-2-duckdb` tool.


aothms:
Hi, congrats on your project

> For other formats, you may use other **IfcOpenShell utilities** as shown in the table below.
> 
> (emphases mine)

The table your adding to is for utilities in the ifcopenshell repo though. That's confusing. So I would either think about adding it as a recipe/script within the IfcOpenShell repo or advertising your project on e.g OSarch. 

Btw. it would help me understand if there are a little bit more complete query examples or benchmark comparisons to understand when duckdb would be a good fit for IFC data.

chuongmep:
Hi @aothms , 
Thank you for your reply. My work is just an experiment from a weekend hackathon, which is the main reason I don’t want to try it in the main IfcOpenShell repo. In addition, I don’t have much time to create detailed benchmark comparisons right now, but I can make sure to do that in the next stage of the repo. My goal is very clear: to enable fast data analysis with high-speed reading from DuckDB.

Moult:
This might be a good idea to put in https://docs.ifcopenshell.org/other.html - can you redo this PR for that doc page instead?

PR #7187, opened 2025-10-03
Fix warning C4702: unreachable code

Osyotr:


aothms:
Thanks, this one was high on my list as well, but never made the change dreadful of the local recompilation required..

PR #7191, opened 2025-10-03
Fix warning C4275

Osyotr:
It's not enough to disable this warning by adding /wd4257 at build time because it's triggered when the library is consumed.

Also move key functions to cpp.
https://lld.llvm.org/missingkeyfunction.html
https://releases.llvm.org/20.1.0/tools/clang/docs/DiagnosticsReference.html#wweak-vtables


aothms:
Many thanks

PR #7193, opened 2025-10-04
cmake: Simplify hdf5 linkage

EstebanDugueperoux2:


Andrej730:
I'm not sure that we should right now remove `HDF5_INCLUDE_DIR` and `HDF5_LIBRARY_DIR` completely as suggested here - we've added support for `find_package` in this case very recently and users script might still depend on previously available variables. 
We should definetely move our scripts to use packages and let some time pass to test it out and only then remove legacy variables.
Maybe we can add warnings for `HDF5_INCLUDE_DIR` and `HDF5_LIBRARY_DIR` deprecation.

EstebanDugueperoux2:
Hi @Andrej730,

Thanks for your feedback.
I understand, do you think to apply delegate pattern to cmake package, i.e. have a HDF5Config.cmake file encapsulating all HDF5 cmake stuff currently in CMakeLists.txt?

Regards.

EstebanDugueperoux2:
I have updated this PR with my proposition.
It remains the segfault to fix.
Executing `./IfcConvert -yv ../test/input/faceted_brep_csg.ifc 1032-curve.obj --validate` with gdb, I got this backtrace:

```
#0  0x00007ffff4c7b8f9 in __dynamic_cast () from /lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x000055555606a923 in Ifc2x3::IfcMeasureWithUnit::ValueComponent() const ()
#2  0x0000555555f05215 in double IfcParse::get_SI_equivalent<Ifc2x3>(Ifc2x3::IfcNamedUnit*) ()
#3  0x0000555556e1dddd in ifcopenshell::geometry::mapping_Ifc2x3::initialize_units_() ()
#4  0x0000555556e1f5c8 in boost::detail::function::function_obj_invoker2<(anonymous namespace)::factory_t_Ifc2x3, ifcopenshell::geometry::abstract_mapping*, IfcParse::IfcFile*, ifcopenshell::geometry::Settings&>::invoke(boost::detail::function::function_buffer&, IfcParse::IfcFile*, ifcopenshell::geometry::Settings&) ()
#5  0x0000555555e9a52f in ifcopenshell::geometry::impl::MappingFactoryImplementation::construct(IfcParse::IfcFile*, ifcopenshell::geometry::Settings&) ()
#6  0x0000555555e6e6be in ifcopenshell::geometry::Converter::Converter(std::unique_ptr<ifcopenshell::geometry::kernels::AbstractKernel, std::default_delete<ifcopenshell::geometry::kernels::AbstractKernel> >&&, IfcParse::IfcFile*, ifcopenshell::geometry::Settings&) ()
#7  0x0000555555e2ffc7 in IfcGeom::Iterator::Iterator(std::unique_ptr<ifcopenshell::geometry::kernels::AbstractKernel, std::default_delete<ifcopenshell::geometry::kernels::AbstractKernel> >&&, ifcopenshell::geometry::Settings const&, IfcParse::IfcFile*, std::vector<boost::function<bool (IfcUtil::IfcBaseEntity*)>, std::allocator<boost::function<bool (IfcUtil::IfcBaseEntity*)> > > const&, int) ()
#8  0x0000555555d8dd40 in main ()

```

Regards.

EstebanDugueperoux2:
@Andrej730 this PR is ready for review, I propose to divide complexity of this big CMakeLists.txt.
I start small but if this PR is accepted, I'll provide others ones in same principle.

Regards.

EstebanDugueperoux2:
@Andrej730  the CI remains KO because of ruff and TestValidateStub.

EstebanDugueperoux2:
Hi @Andrej730 @aothms I need your review to advance in the CMake files simplification to ease conan integration.

aothms:
@Andrej730 is this something you could look into, I'm really a bit out of touch with cmake

theoryshaw:
@EstebanDugueperoux2 

aothms:
I'm just going to merge this in so that we advance on this front. If people encounter issues with we'll fix them forward. Sorry for testing your patience @EstebanDugueperoux2 and thank you for your work.

EstebanDugueperoux2:
Thanks @aothms, now I try to understand the failure of TestPackageSupportedPlatforms in main branch (https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/19112208286/job/54611981992).
What is the purpose of this test?

I see it compare computed URLs with what is contained in https://builds.ifcopenshell.org/, but why?

aothms:
> I see it compare computed URLs with what is contained in https://builds.ifcopenshell.org/, but why?

The ifcopenshell python package consists of compiled C++ and python. For various reasons the coupling between the two is a bit indirect. C++ builds used to be very slow (has improved, but still 2h) and is a bit costly due to usage of LFS for caching so we don't run these too often but don't want to slow down the development on the python side. Also sometimes because the C++ code doesn't have the same kind of test coverage as we have in python, sometimes errors are apparent only on the python side. For this purpose we have a makefile that a takes a binary build of a specific commit and patches it with latest python code. This specific commit is hardcoded in the makefile. This test verifies that the binaries for this commit exist on S3. The problem why the test fails is that the referenced commit is a bit old, from before we had py3.14 builds and when we used a different pyodide stdlib, so there is a mismatch between the expected matrix and what exists on S3.

@Andrej730 I just update the makefile so that the test runs. If you disagree then feel free to rollback.

EstebanDugueperoux2:
Hi @aothms , thanks it is clearer.

I don't see why IfcOpenShell use git lfs, according to https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github, Github don't accept file bigger than 100MB but executing `find . -type f -size +100M`, I don't see big file.

Regards.

aothms:
For this step https://github.com/IfcOpenShell/IfcOpenShell/blob/593e0e28a9deb8b8a917adce8725125b2c8638a2/.github/workflows/build_rocky.yml#L36-L43 cloning this repo https://github.com/IfcOpenShell/build-outputs/tree/rockylinux8-x64/Linux/x86_64/install 

afbeelding

(this graph is cumulative)

EstebanDugueperoux2:
Ok I understand better why CI is so complex.
Would it be simpler to use find_package() cmake command to build against dependencies provided by system package manager?

aothms:
The builds we distribute sometimes have specific version requirements that haven't made it into the package manager, also many packages (the larger ones especially such as opencascade) only come with dynamic libraries in the package manager which means the resulting build would not really be portable

Andrej730:
@EstebanDugueperoux2 

Hi! Sorry for coming back to this so late, but it seems `cmake/HDF5Config.cmake` will never be used, since we're not using `CONFIG` in `find_package` and there's default `FindHDF5.cmake` bundled with cmake that will executed on the line below

https://github.com/IfcOpenShell/IfcOpenShell/blob/613015e37e4de95899aceada8fbd1172132354ca/cmake/CMakeLists.txt#L597C5-L597C49

So in theory the file should be renamed to `FindHDF5.cmake` and the command `find_package(HDF5 REQUIRED)` should be used in the main cmake.

Does this makes sense or I'm missing something? 😅

EstebanDugueperoux2:
Hi @Andrej730,

Thanks a lot for review and your commits.
Indeed building locally using nix/build-all.py, build/Linux/x86_64/install/hdf5-1.13.1/cmake/hdf5-config.cmake is used.
But perhaps it is used for CI build.
It is one thing on which I would like to works, i.e. have same build command used locally and in CI, perhaps using conan.

I continue with another PR (https://github.com/IfcOpenShell/IfcOpenShell/pull/7388).
I see that svgfill git module is used to link against ifcconvert, what is the reason to have svgfill as git module?
Is it used by another project?

@aothms on main branch I see test_serializer_settings failing because of new 'separate-z-up-node' option added through https://github.com/IfcOpenShell/IfcOpenShell/commit/9e9d66451177950c7f78f52fcaaacecb871858e4.
Should this new option be added in doc and python wrapper?

Regards.

Andrej730:
Okay, makes sense, but I think I'll rename it to `FindHDF5` to keep the legacy behaviour we had before - at least temporarily, to ensure we don't break anything during refactor.

PR #7207, opened 2025-10-06
Add unwrapping of UV for IfcReferenceImage

falken10vdl:
This is a porposal following the direction of https://community.osarch.org/discussion/3155/ in order to have the IfcReferenceImage unwrapped so it can be seen as a texture in solid mode,
Cheers!


theoryshaw:
works great!

PR #7219, opened 2025-10-08
Put Bonsai webui PID file in a writable location

brunopostle:
Currently a file called running_pid.json is written in the same folder as sioserver.py which is icky. It also doesn't work on a system where bonsai is read-only. This fix puts this file in a platform-suitable location:

  Linux: ~/.cache/bonsai/running_pid.json
  Windows: %LOCALAPPDATA%\bonsai\Cache\running_pid.json
  macOS: ~/Library/Caches/bonsai/running_pid.json

I'm adding this as a pull-request because I have other problems with the webui (Connected Blenders: 0), but I suspect this is some other issue.


theoryshaw:
Hi @brunopostle, you might have read this already...

PR #7224, opened 2025-10-10
Cmake/externalize ifcparse into its own CMakeLists.txt

EstebanDugueperoux2:
Move ifcparse cmake code to its own CMakeLists.txt

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -262,7 +262,7 @@ if(WITH_CGAL)
 	else()
 		set(CGAL_Boost_USE_STATIC_LIBS "${Boost_USE_STATIC_LIBS}")
 	endif()
-        find_package(CGAL REQUIRED)
+    find_package(CGAL REQUIRED)
         if(NOT CGAL_DIR)
             message(
                 FATAL_ERROR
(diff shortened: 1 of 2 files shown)


cmake/CGALConfig.cmake
Osyotr, line 1: This is neither a CMake find module nor a proper CMake config.
I suggest calling `find_package(CGAL REQUIRED)` directly, dropping excessive logging and checks (they are covered by the `REQUIRED` keyword).
Alternatively rename this into something like `FindLibCGAL.cmake`. Same goes for other "configs".

EstebanDugueperoux2: Hi @Osyotr thanks for the review.
I let you see with @Andrej730 who want to keep old way, based on variable, see https://github.com/IfcOpenShell/IfcOpenShell/pull/7193#issuecomment-3368280225 .

EstebanDugueperoux2: Hi @Osyotr, I have reworked this PR to move ifcparse cmake code to its own CMakeLists.txt.
I'll create another PR about  CGAL.

EstebanDugueperoux2:
@aothms @Andrej730 @theoryshaw this PR is ready for review.

aothms:
Thanks, looks good to me :)

PR #7279, opened 2025-10-22
Add functionality to deselect objects not intersected by the active camera view when annotation TEXT

falken10vdl:
This is a PR to address the discussion in https://community.osarch.org/discussion/3180/deselect-non-intersecting-objects-with-camera

https://github.com/user-attachments/assets/851c9b73-685b-4d52-b8cb-359b0febb956

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/__init__.py
+++ b/src/bonsai/bonsai/bim/module/drawing/__init__.py
@@ -89,6 +89,7 @@ classes = (
     operator.SelectAssignedProduct,
     operator.ToggleTargetView,
     operator.OpenDocumentationWebUi,
+    operator.SelectObjectsIntersectedByCamera,
     prop.Variable,
     prop.Drawing,
     prop.Document,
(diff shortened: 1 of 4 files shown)


theoryshaw:
Thanks for tackling this!

Would it be possible to have the user select their own stuff first, and the when 'Intersected by Camera' is run, it deselects those objects that don't intersect. 

...

The way it is now everything gets selected that crosses the camera.  More times than not, that's too much stuff--as often times, I'll just want to select all the walls, for example, or perhaps i'll just want to select a certain window type, etc.  Or people can use their own selection filters, etc.

falken10vdl:
Indeed! That gives the user more flexibility. Now I force a select all and then apply your deselecting code. It makes more sense what you are saying.
Just a question... what name would you give in English to this Toggle? :) 
Cheers!

falken10vdl:
I have updated the functionality to be more FilterSelectedObjectsIfIntersectedByCamera and changed it to an operator instead of a property toggle which I think is better suited for the wanted functionality

https://github.com/user-attachments/assets/4df4d27c-b7c4-4546-b1cb-ba41fc72ec47

Cheers!

falken10vdl:
black formatting has changed some portions of the code unrelated to this PR. It is just formatting.

theoryshaw:
works great!.

falken10vdl:
I have added this PR on top of 7028 since both are related to TEXT annotations and there was a conflict to merge this one on top of the previous one.
The new PR7292 combines 7028+7279.

PR #7292, opened 2025-10-26
PR7028-7279 Enhance TEXT Annotations. Added selection of similar values, bulk change to selected value and deselect objects not intersected by the active camera view

falken10vdl:
This  is a PR combining both PRs 7028 and 7279 (7279 on top of 7028 and solved the merging conflicts)
Cheers!


theoryshaw:
Nice. 

I was under the impression that the following would select all the tags that have `CurrentValue = WAL100`

https://github.com/user-attachments/assets/c4620658-a947-4258-a968-06ee9424cc3f

falken10vdl:
arghh! :) you are right! I did not test with dynamic tags and forgot to compare to the resolved values in the other tags not the literal.
I have also now looked to all the literals within the TAG not just the one with the same index. This means that it will fetch all tags where the resolved text appears in any of the literals of the tag. So for example in the case above that we have a current value WAL100, it will select TAGs that have the value WAL100 for example as a string (not only the ones for {{type.Name}} that resolve to WAL100) or  in the literal[1]... etc. 
Is this the intended behaviour?
Cheers!

theoryshaw:
That is! thank you. 

The other way, doesn't seem to work now though...


https://github.com/user-attachments/assets/d21aa653-603f-4c73-abc0-97d13cd7b2d2

theoryshaw:
Is it also possible to close out of edit mode, in bulk?  As you see here, i have to close out of edit mode, on each text instance...
 
https://github.com/user-attachments/assets/e72bb1fb-a4c0-4c11-bab3-6821b737e9b7

falken10vdl:
@theoryshaw You are right! I was confused between literal value and resolved (text) value. I have added both the literal and the currentvalue in the editing mode so it is consistent.
This way one can either search for the literal ("the formula") or the text ("the resolved value").
Please take a look:

https://github.com/user-attachments/assets/af5ac2f9-28c5-44bf-8117-a7d9b534f530

I also store the status of the tag in terms of "editing" so I restore whatever status it had the tag after the tag is processed.

Cheers!

theoryshaw:
Awesome!

One other comment.  And you can take it or leave it, but wanted to log nonetheless. 

For a consistent UI, it would seem that the following could be removed, and only have this 'select similar' happen on the 'read-only' panel.  (in an effort to not clutter the UI too much) 

Awesome PR, thank you.

falken10vdl:
@theoryshaw  Done!
See below:
image

Cheers!

falken10vdl:
@theoryshaw This is merged in the latest bonsaiPR: https://github.com/falken10vdl/bonsaiPR/releases/tag/v0.8.4-alpha251028
;)

falken10vdl:
Combined in https://github.com/IfcOpenShell/IfcOpenShell/pull/7317
Cheers!

Issue #7296, opened 2025-10-28
I get the following error when i try to `bim.perform_quantity_take_off` the following IfcSpace object.


theoryshaw:
### Bug Description

I get the following error when i try to `bim.perform_quantity_take_off` the following IfcSpace object. 

GlobalId = "3hzvE47GT7dPqZy3N4L5mn"

file: https://hub.openingdesign.com/OpeningDesign/Cool_Car_Guys/src/commit/acebe5052a4cc39109933ac8cfba9f9d018cd87a/Open/Models/Bonsai/OD_Cool_Car_Guys.ifc

### Debug and Error Output

```true
os: Windows
os_version: 10.0.19045
python_version: 3.11.11
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: Intel64 Family 6 Model 158 Stepping 10, GenuineIntel
blender_version: 4.5.2 LTS
bonsai_version: 0.0.0
bonsai_commit_hash: 5dbfe53
bonsai_commit_date: 2025-10-23T08:31:31-05:00
last_actions: 
# ifcopenshell.api: pset.edit_qto
>>> {"qto": {"cast_type": "entity_instance", "value": 4073224, "Name": "Qto_WindowBaseQuantities"}, "properties": "{'Area': 13.549737335167025, 'Height': 2.999999229832893, 'Width': 2.9999702157278954}"}
# ifcopenshell.api: pset.edit_qto
>>> {"qto": {"cast_type": "entity_instance", "value": 4073232, "Name": "Qto_WindowBaseQuantities"}, "properties": "{'Area': 13.549737335167025, 'Height': 2.999999229832893, 'Width': 2.9999702157278954}"}
# ifcopenshell.api: pset.edit_qto
>>> {"qto": {"cast_type": "entity_instance", "value": 4073240, "Name": "Qto_WindowBaseQuantities"}, "properties": "{'Area': 13.549737335167025, 'Height': 2.999999229832893, 'Width': 2.9999702157278954}"}
# ifcopenshell.api: pset.edit_qto
>>> {"qto": {"cast_type": "entity_instance", "value": 4073256, "Name": "Qto_WindowBaseQuantities"}, "properties": "{'Area': 13.549737335167025, 'Height': 2.999999229832893, 'Width': 2.9999702157278954}"}
# ifcopenshell.api: pset.edit_qto
[...]

PR #7297, opened 2025-10-28
fix for #7296: handle openings without Blender representation in get_…

theoryshaw:
…gross_top_area


theoryshaw:
@maxfb87 I vibed this fix.  Not sure it's correct.

maxfb87 (approved):

PR #7126, opened 2025-09-14
ifc5d.quantify: lookup entity names directly; index into dict rather than iterate in simple cases

aothms:
See here my call tree (from pyinstrument). A lot of time being spent in `selector.py` so I looked for some ways to reduce time spent there:

1. use by_type in simple entity type names
2. use type of elements as an index into the list (especially useful when list of elements is tiny).

```
4.186 _run_module_code  runpy.py:89
└─ 4.186 _run_code  runpy.py:63
   └─ 4.186 <module>  primitive_objects.py:1
      └─ 4.138 Element.build  primitive_objects.py:106
         └─ 4.138 Element._build  primitive_objects.py:327
            ├─ 3.805 <listcomp>  primitive_objects.py:346
            │  ├─ 2.904 Translate.build  primitive_objects.py:390
            │  │  ├─ 2.435 Boolean.build  primitive_objects.py:490
            │  │  │  ├─ 1.849 <listcomp>  primitive_objects.py:492
            │  │  │  │  ├─ 1.289 Translate.build  primitive_objects.py:390
            │  │  │  │  │  └─ 1.279 Element.build  primitive_objects.py:106
            │  │  │  │  │     └─ 1.278 Element._build  primitive_objects.py:327
            │  │  │  │  │        └─ 1.263 quantify  ifc5d\qto.py:61
            │  │  │  │  │           ├─ 1.107 filter_elements  ifcopenshell\util\selector.py:425
            │  │  │  │  │           │     [19 frames hidden]  lark
            │  │  │  │  │           └─ 0.152 IfcOpenShell.calculate  ifc5d\qto.py:259
            │  │  │  │  │              └─ 0.149 iterator.initialize  ifcopenshell\ifcopenshell_wrapper.py:1097
            │  │  │  │  │                 └─ 0.149 Iterator_initialize  <built-in>
            │  │  │  │  └─ 0.560 Element.build  primitive_objects.py:106
            │  │  │  │     └─ 0.560 Element._build  primitive_objects.py:327
            │  │  │  │        └─ 0.551 quantify  ifc5d\qto.py:61
            │  │  │  │           ├─ 0.455 filter_elements  ifcopenshell\util\selector.py:425
            │  │  │  │           │     [6 frames hidden]  lark
            │  │  │  │           └─ 0.095 IfcOpenShell.calculate  ifc5d\qto.py:259
            │  │  │  │              └─ 0.093 iterator.initialize  ifcopenshell\ifcopenshell_wrapper.py:1097
[...]

PR #7301, opened 2025-10-29
Fix QTO speed problem from https://github.com/IfcOpenShell/IfcOpenShell/pull/7126

theoryshaw:
discussed: https://matrix.to/#/!ssOmSJysZpXBupsgjO:matrix.org/$7EB0LAxm9eD920ZuafOOPA3mR5grPh0SSWsLXWI1g6w?via=matrix.org&via=tchncs.de&via=crossbach.de

### DIFF
--- a/src/ifc5d/ifc5d/qto.py
+++ b/src/ifc5d/ifc5d/qto.py
@@ -173,8 +173,7 @@ class IteratorForTypes:
     element: Union[ifcopenshell.entity_instance, None] = None
     shape: Union[ifcopenshell.geom.ShapeType, None] = None
 
-    def __init__(
-        self,
+    def __init__(self,
         ifc_file: ifcopenshell.file,
         settings: ifcopenshell.geom.settings,
         elements: Iterable[ifcopenshell.entity_instance],
--- a/src/ifc5d/ifc5d/qto.py
+++ b/src/ifc5d/ifc5d/qto.py
@@ -59,15 +59,15 @@ for name in get_args(RULE_SET):
     with open(os.path.join(cwd, name + ".json"), "r") as f:
         rules[name] = json.load(f)
 
-
 @functools.cache
 def lower_case_entity_names(schema_iden: str):
     schema = ifcopenshell.ifcopenshell_wrapper.schema_by_name(schema_iden)
-    return set(n.name().lower() for n in schema.entities())
+    return {n.name().lower() for n in schema.entities()}
 
 
 @functools.cache
 def entity_supertypes(schema_iden: str, entity_name: str):
+    """Return all supertypes (lowercased) for a given entity."""
     def visit(decl):
         yield decl.name().lower()
         if ty := decl.supertype():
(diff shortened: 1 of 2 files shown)


theoryshaw:
Defer to others.

maxfb87:
@aothms can you look at this? I don't fully understand it...

src/ifc5d/ifc5d/qto.py
```diff
@@ -78,41 +78,52 @@ def visit(decl):
...
-                        calculator.calculate(ifc_file, list(elements), qtos, results)
-        for query, qtos in queries.items():
-            if query.lower() in lower_case_entity_names(ifc_file.schema_identifier):
-                # by_type is faster than selector parsing
-                filtered_elements = ifc_file.by_type(query)
-            else:
-                filtered_elements = ifcopenshell.util.selector.filter_elements(ifc_file, query, elements)
-            if filtered_elements:
-                calculator.calculate(ifc_file, filtered_elements, qtos, results)
+            # Cache supertypes per entity to avoid recomputation
```
aothms: There is already a caching decorator on:

```
@functools.cache
def entity_supertypes(...)
```

Did you benchmark this?

src/ifc5d/ifc5d/qto.py
```diff
@@ -78,41 +78,52 @@ def visit(decl):
...
-    autodetected and loaded with the module for convenience.
+    # Precompute entity names once
+    schema_id = ifc_file.schema_identifier
+    entity_names_lower = lower_case_entity_names(schema_id)
 
-    :param rules: Set of rules from `ifc5d.qto.rules`.
-    """
-    results: ResultsDict = {}
-    elements_by_classes: defaultdict[str, set[ifcopenshell.entity_instance]] = defaultdict(set)
+    # Pre-group elements by type (avoids sorting/groupby)
```
aothms: Is that really faster? Theoretically it's n*log(n) for the sort vs just n for the defaultdict, but I don't think any of the performance here is significant wrt to the runtime of the actual quantity calculation

src/ifc5d/ifc5d/qto.py
```diff
@@ -78,41 +78,52 @@ def visit(decl):
...
+            supertype_cache = {}
+
+            for entity_type, elems in elements_by_class.items():
+                if entity_type not in supertype_cache:
+                    supertype_cache[entity_type] = entity_supertypes(schema_id, entity_type)
+
+                for sty in supertype_cache[entity_type]:
+                    if (sty in casenorm) and (qtos := queries.get(casenorm[sty])):
+                        calculator.calculate(ifc_file, elems, qtos, results)
+        else:
```
aothms: I think the important thing is the else statement here. That is really crucial otherwise we're doing it twice in the simple case.

aothms (approved):

theoryshaw:
Is the slow down, only a Windows problem?  Or are other experiencing this too? 

@aothms Sorry, Not knowledgeable enough to answer your inline questions. 
I'm happy to test any adjustments you have in mind.  

Best, Ryan

theoryshaw:
The chat i used for the last commit. 
https://chatgpt.com/share/6903e260-dcb4-8013-ab05-99df6a1e36ee

aothms:
Feel free to merge if it's better for you this way. I don't see anything prohibitve

theoryshaw:
Sounds good, will try.

PR #7308, opened 2025-10-31
Add Extras in addon preferences to conditionally add functionality. One extra added from commit from sjb007

falken10vdl:
This is in response to comment in https://community.osarch.org/discussion/2909/ifc-aggregates-vs-parent-child

> @sjb007 said:
> @falken10vdl Well this is how it used to be, and they (the main devs) intentionally moved away from this.  I didn't think it would be accepted because of this. I just ran into consistency issues with the newer method - mainly it just _doesn't_ work properly with the Status filter feature in the Scheduling tab. All I'm doing is restoring an older piece of code that worked better, at least in my usecase.

Proposal to add Extras to enable conditionally features that do not make it for main stream but that are very usefull for some users.

I have taken the example of commit by @sjb007 https://github.com/IfcOpenShell/IfcOpenShell/commit/fb77fca333a003390515bf1b220d794e5b705968.
@sjb007 I have copied your code with pride :) and add it as an Extra that can be enabled.

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/spatial/ui.py
+++ b/src/bonsai/bonsai/bim/module/spatial/ui.py
@@ -132,23 +132,16 @@ class BIM_PT_spatial_decomposition(Panel):
             op = col.operator("bim.set_default_container", icon="OUTLINER_COLLECTION", text="Set Default")
             op.container = ifc_definition_id
 
-            # Only show container visibility operators if the preference is enabled
-            from bonsai.bim.ui import BIM_ADDON_preferences
-
-            addon = BIM_ADDON_preferences.get_addon_instance()
-            if addon and getattr(addon, "preferences", None):
-                prefs = addon.preferences
-                value = getattr(prefs, "container_hide_show_isolate", None)
-                if value:
-                    op = row.operator("bim.set_container_visibility", icon="FULLSCREEN_EXIT", text="")
-                    op.mode = "ISOLATE"
-                    op.container = ifc_definition_id
-                    op = row.operator("bim.set_container_visibility", icon="HIDE_OFF", text="")
-                    op.mode = "SHOW"
-                    op.container = ifc_definition_id
-                    op = row.operator("bim.set_container_visibility", icon="HIDE_ON", text="")
-                    op.mode = "HIDE"
-                    op.container = ifc_definition_id
+            if tool.Blender.get_addon_preferences().container_hide_show_isolate:
+                op = row.operator("bim.set_container_visibility", icon="FULLSCREEN_EXIT", text="")
+                op.mode = "ISOLATE"
+                op.container = ifc_definition_id
+                op = row.operator("bim.set_container_visibility", icon="HIDE_OFF", text="")
+                op.mode = "SHOW"
+                op.container = ifc_definition_id
+                op = row.operator("bim.set_container_visibility", icon="HIDE_ON", text="")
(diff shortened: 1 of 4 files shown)


theoryshaw:
Am i doing this wrong?....


https://github.com/user-attachments/assets/e5552df8-da87-472c-93b9-d6ffbab74593

falken10vdl:
@theoryshaw I have tried in windows and it works for me.
Please note that I "improved" the automation of the bonsaiPR addon generation (doing some tests of failed PRs to verify if they are due to issue with the base or with other already loaded PRs) and by mistake was leaving the base branch instead of the weekly branch checked out before doing the addon build... now it is fixed.
Could you try with the latest addon/branch, please?

falken10vdl:
@Andrej730 and @theoryshaw Can you review this one? I think this could be a good addition to Bonsai. Allowing for further "extras" to be added that serve local/custom needs.
The rationale behind was discussed in https://community.osarch.org/discussion/3195/extra-extra
Thanks!

falken10vdl:
'+ @sboddy Are you happy with this one? I basically used your commit and wrap it under an "Extras" panel in bonsai settings.
Once in place I think other features can be turned on/off based on user needs.

Thanks!

src/bonsai/bonsai/bim/module/spatial/ui.py
```diff
@@ -132,6 +132,24 @@ def draw(self, context):
...
             op.container = ifc_definition_id
 
+            # Only show container visibility operators if the preference is enabled
+            from bonsai.bim.ui import BIM_ADDON_preferences
+
+            addon = BIM_ADDON_preferences.get_addon_instance()
+            if addon and getattr(addon, "preferences", None):
+                prefs = addon.preferences
+                value = getattr(prefs, "container_hide_show_isolate", None)
+                if value:
```
sboddy: Wouldn't it be tidier to use the idiom used elsewhere for preferences? (untested, but should work):
``` python
if tool.Blender.get_addon_preferences().container_hide_show_isolate:
```

src/bonsai/bonsai/bim/ui.py
```diff
@@ -678,6 +678,34 @@ def update_cache_dir(self, context: bpy.types.Context) -> None:
...
+            if ".bonsai" in key:
+                instance = bpy.context.preferences.addons.get(key)
+                break
+        cls.addon_instance = instance
+
+    @classmethod
+    def get_addon_instance(cls):
+        if cls.addon_instance is None:
+            cls.set_addon_instance()
+        return cls.addon_instance
```
sboddy: If you switch to use `get_addon_preferences` above, I don't think addon_instance or the get/set class methods are needed either.

sboddy (commented):
I wasn't keen on the mid-file import and wondered how preferences are accessed elsewhere. I think this method is simpler, although I didn't test it.

src/bonsai/bonsai/bim/module/spatial/ui.py
falken10vdl: Yes. Much cleaner. Thanks!

src/bonsai/bonsai/bim/ui.py
falken10vdl: yes. Removed. Thanks!

sboddy (approved):

PR #7315, opened 2025-11-03
added suggestions to search items for easy access

falken10vdl:
This is a proposal to simplify the use of search filters by adding a magnifying glass operator that lists the values available to chose or suggestions based on name/ifcclass

https://github.com/user-attachments/assets/74492d1d-1690-4ede-9608-5e2aaacfcb21

Cheers!


falken10vdl:
Added some operators for Class and GlobalId to either add/subtract/filter
Example:

https://github.com/user-attachments/assets/5dd8e35e-d022-4021-82dd-8b5a9d3c9283

Cheers!

falken10vdl:
Added a "systems" search

https://github.com/user-attachments/assets/086ee4c1-8277-468a-9f6c-b8d4e3b34460

Cheers!

theoryshaw:
works great!

falken10vdl:
@Moult @theoryshaw @Andrej730 @sboddy 
Are you guys happy with this PR? I think it could be quite useful as the search function is in quite a number of panels now that could benefit of both the suggestions and the operators for Class and GlobalId to either add/subtract/filter (that was spotted by sboddy int the selector syntax) which is a subtle behavior only seen if you read carefully the documentation. 
Thanks!

theoryshaw:
works for me!

falken10vdl:
I have converted this one to draft. I have introduced another simple way to create filter (and have suggestions as well) that leverages "extra" area in addon settings so "no change" is done to the current filter implementation.

falken10vdl:
This one is replaced by https://github.com/IfcOpenShell/IfcOpenShell/pull/7498 which includes this functionality and a new filter mode based on a chain of set operations.
Cheers!

PR #7316, opened 2025-11-04
Add attribute info for relating type

falken10vdl:
This is a proposal for https://community.osarch.org/discussion/3196/when-you-select-an-instance-that-it-shows-the-attributes-for-both-the-instance-and-the-type

https://github.com/user-attachments/assets/467a6102-f202-4ec1-8eb3-63888918da63

Cheers!


theoryshaw:
works great!

PR #7323, opened 2025-11-05
Allow_complete_aggregate_move_in_drawing_mode

falken10vdl:
Addresses https://community.osarch.org/discussion/3206/turn-on-all-objects-in-an-aggregate-if-the-drawings-camera-sees-at-least-one-aggregate-object#latest

https://github.com/user-attachments/assets/b7fd91cd-8805-413c-8c97-c2bb756e857e

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -2182,7 +2182,7 @@ class ActivateModel(bpy.types.Operator):
         start_time = time.time()
 
         dprops = tool.Drawing.get_document_props()
-        
+
         if dprops.active_drawing_id:
             active_drawing = tool.Ifc.get().by_id(dprops.active_drawing_id)
             core.sync_references(tool.Ifc, tool.Collector, tool.Drawing, drawing=active_drawing)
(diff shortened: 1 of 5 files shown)


falken10vdl:
Changed logic to only do transformation for G -> IFC Move Aggregate

https://github.com/user-attachments/assets/ccdb7278-a8af-48b5-b522-98918d60bfc3

Cheers!

theoryshaw:
This doesn't seem to be working for me...

https://github.com/user-attachments/assets/78f88902-c59a-4cfe-b421-da1fe568c55a

...
And in the following example it seems like some IfcCovering elements were deleted...

file: https://hub.openingdesign.com/OpeningDesign/Cool_Car_Guys/src/commit/da91412469158796991123b456624b4c5392b053/Open/Models/Bonsai/OD_Cool_Car_Guys.ifc

https://github.com/user-attachments/assets/6264d3b0-d58f-48df-a0bf-aadd9c515885

falken10vdl:
@theoryshaw I see an issue. I am tracking pressing the G and of course if the G is pressed but not wanted (esc) I perform wrong a translation.
I have taken a look to the initial case. In 3d view if one of the element of the assembly is hidden, if we move the rest (or the empty of the assembly) the hidden element is not moved. 

https://github.com/user-attachments/assets/a6b1e55b-586f-4f1b-a95f-d970253d152e



Is this expected behavior? I guess G should move all the elements of the assembly together regardless if they are hidden or not? And that should happen for both drawing mode and 3d view?


Thanks!

PR #7329, opened 2025-11-09
Ifc5d qto optimization

maxfb87:
following #7126 and https://github.com/IfcOpenShell/IfcOpenShell/commit/7329c485ea0e9033bf7fdcf8b0217b9ef9080c55
i tried to simplify the logic behind
Maybe i'm missing something (probably :-) ) but i think the function could do only this
ping @aothms 
I added also a profiler (timer) function (in another PR) in order to understand if the quantification time really decrease, so if anyone wants to try you are welcome (ping @theoryshaw )

### DIFF
--- a/src/ifc5d/ifc5d/qto.py
+++ b/src/ifc5d/ifc5d/qto.py
@@ -69,11 +69,6 @@ def quantify(ifc_file: ifcopenshell.file, elements: set[ifcopenshell.entity_inst
     """
 
     results: ResultsDict = {}
-    elements_by_classes: defaultdict[str, set[ifcopenshell.entity_instance]] = defaultdict(set)
-    for element in elements:
-        elements_by_classes[element.is_a()].add(element)
-
-    results: ResultsDict = {}
 #    elements_by_classes: defaultdict[str, set[ifcopenshell.entity_instance]] = defaultdict(set)
 #    for element in elements:
 #        elements_by_classes[element.is_a()].add(element)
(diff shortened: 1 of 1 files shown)


theoryshaw:

Conflict Report:

-   [**PR #7329**](https://github.com/IfcOpenShell/IfcOpenShell/pull/7329): Ifc5d qto optimization
    -   Conflicting files:  `src/ifc5d/ifc5d/qto.py`
    -   Possible breaking commits:  `72f8218fc ruff - check for more unused imports`,  `13be6ccd4 Sort imports`,  `949109648 black .`,  `e8eb5e4d0 Add continue statement to skip duplicate calculations`,  `24bee01f0 black .`

BIMvoice:
Ran this against the actual production QTO rulesets to check the dispatch change. On IFC4 it works (all 93 query keys there are plain entity names), but on IFC4X3, 100 of 112 query keys are `ifcopenshell.util.selector` expressions (e.g. `"IfcActuator + IfcActuatorType"`, `'IfcSign, PredefinedType="PICTORAL" + ...'`). The simplified `element.is_a() in queries` check drops the existing selector-expression fallback, so none of those match. In my test, 0 of 4 elements matched any query key on an IFC4X3 file, meaning `quantify()` would silently return empty results for essentially all elements on that schema.

This also reintroduces the failure mode #7301 fixed (QTOs applying to the wrong element set), since that fix is what added the selector fallback this diff removes.

Separately, `grouped_elements = defaultdict(list)` declared outside the per-calculator loop shadows the function's own `elements` parameter inside the inner loop, only latent right now because every ruleset JSON happens to have exactly one `calculators` entry.

@maxfb87, given IFC4X3 support would break, do you want to fold the selector fallback back in, or was there a different simplification angle you had in mind?

PR #7341, opened 2025-11-11
ifc5d qto add continue statement to skip duplicate calculations

kapluscheung:
ifc5d qto skip general path to avoid duplicate calculations.


aothms:
Thanks

Issue #7331, opened 2025-11-09
CostSchedules - Deriving Quantites from IfcSpace Instances / Qto_SpaceBaseQuantities  is not working


Rezan43:
[GrossFloorAreaModel_RF.txt](https://github.com/user-attachments/files/23441238/GrossFloorAreaModel_RF.txt)

### Bug Description

Hello everbody,

I got a question regarding IfcCostSchedules in relation to IfcOpenShell but also Bonsai. I got an IfcModel consisting only of IfcBuildingStoreys and IfcSpaces with a "GFA" as their PredefinedType. 

Bonsai: 
I tried following along this [video](https://www.youtube.com/watch?v=H40TJ7EH4tw) with the Intention to create one cost item whose quantity I derive from the "GrossFloorArea" Quantity from the QuantitySet "Qto_SpaceBaseQuantities"
When I clicked on "Assign Cost Item Quantity [...] from Product Property" nothing happened and when I clicked on   "Assign Cost Item Quantity [...] from Product" the Cost items Quantity value changes to "0.00". 

Is there a fix to this?

IfcOpenShell:

Here I am basically trying to do the same thing but its also not working. I've noticed that the quantity gets shown as 3 and not zero when when removing the "prop_name" Argument from the "assign_cost_item_quantity" function, 
I also tried using other Ifc-Files and Schemes but it didnt change anything. 

This is the software I am using:

- Bonsai: 0.8.3- post1
- Blender: 4.2.13 LTS
- IfcOpenShell: 0.8.3.post2

I'd appreciate any type of Input. Thanks a lot in advance. I converted the ifc file to txt file so one could download it and try for themselves. 

### Attachments

Code Excerpt from my python script 

```
#CREATED COST SCHEDULE
print("Script Started")
print("IFC Schema:\t", model.schema,end="\n\n")

#Set currency
eur = ifcopenshell.api.unit.add_monetary_unit(model, currency="EUR")
proj = model.by_type("IfcProject")[0]
existing = list(proj.UnitsInContext.Units) if proj.UnitsInContext else []
existing = [u for u in existing if not u.is_a("IfcMonetaryUnit")]  # drop old currencies
ifcopenshell.api.unit.assign_unit(model, units=existing + [eur])

schedule = ifcopenshell.api.cost.add_cost_schedule(model)
[...]

PR #7342, opened 2025-11-11
#7331 related: Modifications to include IfcSpace entities in Cost Schedules, as elements and quantities + Project Library extended support for costing and scheduling

DimitriosThe:
This is a simplified approach.

In preliminary cost evaluations in most cases the model is quite abstract to derive detailed costs, how ever by leveraging the use of spaces and working with indices cost /m2 cost/m3 and quantities from IfcSpace, one could easily generate rough budgets, aligning with conceptualization/ preliminary phase scope.

I am not aware/ do not know if there are specific restrictions for which these entities should/ must not be included in a Cost Schedule.


BIMvoice:
@DimitriosThe you're right, and thank you for flagging it, apologies for stepping on your work.

You were ~8 months first: your #7342 changed `assign_cost_item_quantity.py` the same way back on 2025-11-11, and we duplicated it in #8410 without spotting your open PR. That one's on us.

We reviewed #7342 properly, and your framing is correct, #8410 alone does not make IfcSpace quantities usable end-to-end. Your other parts are what complete it:
- `util/cost.py` (`IfcElement` -> `IfcProduct` in the PRODUCT filter): without this an IfcSpace quantity never appears in the Elements list, since IfcSpace is an `IfcSpatialElement`, not `IfcElement`. This is the piece that makes it actually show up.
- `tool/cost.py` (`IfcTypeProduct` -> `IfcProduct`): fixes a real `assert False` crash in `load_cost_item_types` for Schedule-of-Rates schedules.
- `append_asset.py`: extends Project Library appending to WorkSchedule/WorkCalendar/ConstructionResource and widens the CostSchedule whitelist, a sound addition.

Good news on landing it: despite the "conflicting" flag the real conflict is tiny, only `assign_cost_item_quantity.py`, one hunk, resolved by dropping your hunk since v0.8.0 already has that exact line now (via #8410). The other three files auto-merge cleanly.

A few things we noticed that would otherwise block CI, so you have them:
1. `append_work_calendar()` never calls/returns `self.add_element(...)` like its siblings, so appending a calendar currently no-ops, looks like a missing return.
2. `append_asset.py` isn't black-formatted to the repo's line-length 120, which fails ci-lint and inflates the diff.
3. The new appendable types + widened whitelist aren't covered by `test_append_asset.py` yet.

Happy to help however suits you: I can prepare the rebase + those fixes as a patch (or a PR into your branch) with you as author, or just leave the specifics here for you to apply, your call, it stays your work and your credit. @Moult @maxfb87 could we prioritise #7342 to finish the IfcSpace-in-cost-schedule feature? It's an easy rebase from here.

Thanks again for building the complete solution, and sorry for the duplication.

PR #7362, opened 2025-11-14
Fix #7348 ci test normalize path

### LINKED ISSUE #7348: CI test_normalize_path failure
### Bug Description

I was idly looking at what tests in CI were failing, and came across this oddity.

https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/19269554517/job/55094356251#step:5:1281

In the CI action (running on Linux) it is trying to pass a windows style path `test\test.txt` to tools/ifc.py:normalize_path class method:
``` python
    @classmethod
    def normalize_path(cls, path: Union[Path, str]) -> str:
        # Do not use `Path.resolve` as it will resolve symlinks too.
        return Path(os.path.normpath(path)).as_posix()
```
On Linux the `os.path.normpath` call will not change the string, because it will not handle the `\` as a separator. Interestingly the `/` is converted by the Windows normpath. The `pathlib.Path` also does not automatically handle a string containing a Windows path. So, at least on Linux, the `assert` statement fails.

``` python
        path = r"test\test.txt"
        assert subject.normalize_path(path) == "test/test.txt"
```

Is it valid to pass a Windows path on Linux? If not, that assert should be removed.

If normalize_path _is_ expected to convert Windows `\` into Posix `/` then the classmethod needs changing. The most reliable method I could come up with is for normalize_path to treat every path as a PureWindowsPath, convert to posix, _then_ normalise. i.e.

``` python
        return os.path.normpath(PureWindowsPath(path).as_posix())
```

The PureWindowsPath _does_ handle Posix `/` as well so it _should_ be safe to process any path, and the Windows version of os.path.normpath will convert `/` to `\` so that _should_ be safe too.

@Andrej730 Subbing you in, as you refactored all this here: https://github.com/IfcOpenShell/IfcOpenShell/commit/e792cf8fa0ed7ea7912e0f2d7865a8025ccdb678 so you're the expert 😉

Here's the dross I was using to test all the permutations I could think of, in case someone wants to run it on Windows. Unchanged, each path should "TEST" true:
``` python
for x,y in [[r"test\test.txt",    r"test/test.txt"],
            [r"test/test.txt",    r"test/test.txt"],
            [r"test\..\test.txt", r"test.txt"],
            [r"test/../test.txt", r"test.txt"],
            [r"\test\test.txt",   r"/test/test.txt"],
            [r"/test/test.txt",   r"/test/test.txt"],
            [r"D:\test/test.txt", r"D:/test/test.txt"],
            ]:
    print()
    print ("STR:", x, " ->  EXP:", y)
    for v in [
                #Path(x),
                #PurePath(x),
                #PurePosixPath(x),
                PureWindowsPath(x),
                #PurePosixPath(PureWindowsPath(x)),
                #PureWindowsPath(PurePosixPath(x)),
                #PurePath(PureWindowsPath(x)),
                ]:
        print("  NRMS:", os.path.normpath(x))
        print("  TYPE:", type(v))
        print("    PRTS:", v.parts)
        #print("    NORM:", os.path.normpath(v))
        #print("    TEST:", os.path.normpath(v) == y)
        #print("    POSX:", v.as_posix())
        print("    NRMP:", os.path.normpath(v.as_posix()))
        print("    TEST:", os.path.normpath(v.as_posix()) == y)
        
    print()
```

### Attachments

_No response_

### Debug and Error Output

```true

```

sboddy:
Fixes minor CI test failure #7348 for Linux.

### DIFF
--- a/src/bonsai/test/tool/test_ifc.py
+++ b/src/bonsai/test/tool/test_ifc.py
@@ -23,6 +23,7 @@ import bonsai.core.tool
 import bonsai.tool as tool
 import pytest
 import tempfile
+import os
 from pathlib import Path
 from bonsai.tool.ifc import Ifc as subject
 
@@ -318,8 +319,10 @@ class TestUri(test.bim.bootstrap.NewFile):
 
     def test_normalize_path(self):
         # Posix format.
-        path = r"test\test.txt"
-        assert subject.normalize_path(path) == "test/test.txt"
+        if os.name == 'nt':
+            # Limit to windows (#7348)
+            path = r"test\test.txt"
+            assert subject.normalize_path(path) == "test/test.txt"
 
         with tempfile.TemporaryDirectory() as tmp_dir:
             base_path = Path(tmp_dir)
--- a/src/bonsai/test/tool/test_ifc.py
+++ b/src/bonsai/test/tool/test_ifc.py
@@ -319,7 +319,7 @@ class TestUri(test.bim.bootstrap.NewFile):
 
     def test_normalize_path(self):
         # Posix format.
-        if os.name == 'nt':
+        if os.name == "nt":
             # Limit to windows (#7348)
             path = r"test\test.txt"
             assert subject.normalize_path(path) == "test/test.txt"


Andrej730:
Thank you.

PR #7388, opened 2025-11-19
cmake: Externalize geometry_serializer cmake config files to their ow…

EstebanDugueperoux2:
…n directory

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -205,6 +205,9 @@ if(WITH_CGAL)
 endif()
 
 if(WITH_OPENCASCADE)
+    find_package(OpenCASCADE REQUIRED)
+    # Map OpenCASCADE_LIBRARIES variable from OpenCASCADEConfig.cmake to OPENCASCADE_LIBRARIES used by kernel generic cmake file
+    set(OPENCASCADE_LIBRARIES ${OpenCASCADE_LIBRARIES})
     add_definitions(-DIFOPSH_WITH_OPENCASCADE)
     set(SWIG_DEFINES ${SWIG_DEFINES} -DIFOPSH_WITH_OPENCASCADE)
     list(APPEND GEOMETRY_KERNELS opencascade)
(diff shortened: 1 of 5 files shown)


Andrej730:
Noticed two things:

1. If we're replacing `IFC_GEOMSERIALIZATION_EXPORTS` with `IFC_GEOM_EXPORTS`, then we also need to change it here:

https://github.com/IfcOpenShell/IfcOpenShell/blob/1d01d1f18454a236a2864d70d98c3c3ad738bad6/src/ifcgeom/Serialization/ifc_geomserialization_api.h#L25


2. It seems `if((BUILD_CONVERT OR BUILD_IFCPYTHON) AND WITH_OPENCASCADE)` check should be moved `src/ifcgeom/CMakeLists.txt`, since the entire submodule depends on it and then better to add it earlier.

Besides that, all looks good.

EstebanDugueperoux2:
@Andrej730 fixed, it was a mistake during a git rebase.

cmake/CMakeLists.txt
```diff
@@ -226,6 +226,9 @@ if(WITH_CGAL)
 endif()
 
 if(WITH_OPENCASCADE)
+    find_package(OpenCASCADE REQUIRED)
+    # Map OpenCASCADE_LIBRARIES variable from OpenCASCADEConfig.cmake to OPENCASCADE_LIBRARIES used by kernel generic cmake file
+    set(OPENCASCADE_LIBRARIES ${OpenCASCADE_LIBRARIES})
```
Andrej730: It seems we don't need `OPENCASCADE_LIBRARIES` anymore, so no need to maintain it and we can just replace it with `OpenCASCADE_LIBRARIES` everywhere.

aothms:
Please check if merge is resolved correctly

EstebanDugueperoux2:
Hi @aothms,

Thanks for the reviews, I have rebased my branch, it is ready for merge now.

Regards.

aothms:
Thanks for all the recent PRs @EstebanDugueperoux2

EstebanDugueperoux2:
Thanks to you, and it is not finished :)

The next one is https://github.com/IfcOpenShell/IfcOpenShell/pull/7389

PR #7389, opened 2025-11-19
Cmake/externalize geometry mapping target cmake config

EstebanDugueperoux2:


EstebanDugueperoux2:
Thans @aothms 

Now I have:

- https://github.com/IfcOpenShell/IfcOpenShell/pull/7434
- https://github.com/IfcOpenShell/IfcOpenShell/pull/7435

Regards.

PR #7391, opened 2025-11-20
Remove spurious cd

aothms:
At this point the dir should not exist yet, right? We `pushd` into it later.


Andrej730:
Yes!! It's one of those annoying `The system cannot find the path specified` I've seen but was only able to pin down b130925c5.

Really dislike `cmd` for this as there's no straightforward way to just stop on errors without doing `IF NOT %ERRORLEVEL%==0 GOTO :Error` every line to easier narrow down issues like this.

PR #7396, opened 2025-11-21
cmake: Use same PRIVATE scope for Serializers target

EstebanDugueperoux2:
cmake: Use same PRIVATE scope for Serializers target


EstebanDugueperoux2:
Hi @Andrej730,

This is a little fix about cmake scope to fix.

Regards.

Andrej730:
Thanks.

PR #7405, opened 2025-11-23
cmake: Use OBJECT target instead of STATIC one to avoid mix static an…

EstebanDugueperoux2:
[…d shared targets in a same build](cmake: Use OBJECT target instead of STATIC one to avoid mix static and shared targets in a same build)


EstebanDugueperoux2:
Hi @aothms @Andrej730,

This is another PR to replace STATIC targets by OBJECT ones to avoid mix static and shared libraries in a same build.
Especially for conan build which doesn't accepts in conan-center-index this mix.

Regards.

Andrej730:
Looks good, more idiomatic and less libraries installed. Thank you.

PR #7411, opened 2025-11-24
feat(selector): add variable substitution and arithmetic to format()

theoryshaw:
- Add {{variable}} syntax for element property references
- Add arithmetic operators: +, -, *, / with proper precedence
- Extend format() with optional element parameter
- Variables work with all format functions (imperial_length, round, etc.)
- Maintains full backward compatibility

Example: format("imperial_length({{z}} / 2, 4)", element)


theoryshaw:
I think this PR conflicts with
https://github.com/IfcOpenShell/IfcOpenShell/commit/dffa3515c0a87d717b66ee0588d3087206de52bf

closing, unless @Moult  thinks there's anything here that could be used.

PR #7339, opened 2025-11-10
cmake: Have a dedicated CMakeLists.txt for IfcConvert & co

EstebanDugueperoux2:

PR #7414, opened 2025-11-25
Fix/include svg fill directly without using git module

EstebanDugueperoux2:


Andrej730:
Related comment - https://github.com/IfcOpenShell/IfcOpenShell/pull/7339#issuecomment-3556864468

Is this really needed for conan? I haven't yet fully wrapped my head around how conan recipe system works, but it seems it's just parsing conandata.yaml next to it. Couldn't we just parse svgfill archive source archive from another url too and extract it?

https://github.com/conan-io/conan-center-index/blob/3990da60b0b68b4b0ff49975a2a1d1caf8c9db1c/recipes/opencascade/all/conandata.yml#L1-L4

https://github.com/conan-io/conan-center-index/blob/3990da60b0b68b4b0ff49975a2a1d1caf8c9db1c/recipes/opencascade/all/conanfile.py#L132-L133

EstebanDugueperoux2:
I have never see a such hack in conandata.yml for recipe in conan-center-index.
Since svgfill is not used outside IfcOpenShell, it would simplify a lot the job to have it directly in IfcOpenShell, not only for conan integration but also for future contributor.
Typically me as a newbie, I have not the way to do a recurisve git clone with git module in my daily work, I'm not a fan of git module to say all.
In addition for conan and conandata.yml, as it use hash to check that source hasn't changed, I would have to reference svgfill for a specific commit.
While it is better to reference project by tag/release in conan-center-index.

Regards.

Andrej730:
I don't know if it's a hack - seems legal, since it's possible that source for a package is distributed between multiple repositories. We can create tags on `svgfill` during main ifcopenshell release, so tag will be the same for both repos.

Submodules are convenient in general, though may take a bit to get used to them. But I agree that currently on `svgfill` there's not much going on and we're probably bumping the submodule all the time after making commit there. I just think that publishing ifcopenshell as a package anywhere shouldn't prevent us from using submodules on the repo in general.

@aothms do you think there's still benefit from keeping `svgfill` in a separate repo?

Either way, we can decide later, when intergating ifcopenshell to conan, to see if it's actually will be a problem and may be there are some workarounds.

EstebanDugueperoux2:
@Andrej730 waiting Thomas answer, I propose to build svgfill targets only if BUILD_IFCPYTHON option is ON, to not include it by default in the conan recipe, espcially since svgfill seems only used through IfcOpenShell python wrapper.

https://github.com/IfcOpenShell/IfcOpenShell/pull/7430

aothms:
> @aothms do you think there's still benefit from keeping svgfill in a separate repo?

No, I think it's fine to integrate it. This was in a time where CGAL was not a dependency yet (I think). And maybe we thought there would be more general use out of this utility but it never happened. I think it's fine to integrate.

PR #7421, opened 2025-11-27
fix some target folders to respect parent cmake

dimateos:
Hello!

I was compiling convert and serializers and they were ignoring my bindir / libdir.
This seems to be the fix, it is actually done for other targets but not in all .cmakes

Feel free to edit/close or whatever :)

Best regards!


aothms:
Thanks. I think this makes sense, but I haven't touched the cmake files for some time. CC @Andrej730 @EstebanDugueperoux2

EstebanDugueperoux2:
Hi, 

For what I understand of https://cmake.org/cmake/help/latest/guide/tutorial/Installation%20Commands%20and%20Concepts.html, a simple call to install(TARGETS myTarget) is sufficient with default value for binary (bin/), libs (lib/) and headers files (include/).
@dimateos in which case you have to fix CMake files with this PR?

Regards.

dimateos:
Hello!

That its precisely my case, I need to use a custom lib/bin/include folder for a specific project structure, something like:
`LIB_DIRS="${INSTALL_FOLDER}/lib/${OS_NAME}/x86_64/${local_BUILD_TYPE_FOLDER}/ifcOpenShell"`
`-DINCLUDEDIR=${INSTALL_FOLDER}/external/ifcOpenShell -DLIBDIR=${LIB_DIRS} -DBINDIR=${LIB_DIRS}`

So without this PR some libs were landing in the default lib/ bin/ under `${INSTALL_FOLDER}`

Btw my lib/bin/include should be _relative_ (not prefixed by `${INSTALL_FOLDER}`) but some of the logic here was replacing my defines with defaults:

https://github.com/dimateos/Open-IfcOpenShell/blob/8757e36e2799583a54d0891e2b9f7e01073dc1ec/cmake/CMakeLists.txt#L166166-L185

I can try to find a suitable patch, but first was making this simpler PR.

Regards

EstebanDugueperoux2:
Ok I understand.
My personal goal is to simplify IfcOpenShell cmake files.
Using cmake variables like CMAKE_INSTALL_LIBDIR and CMAKE_INSTALL_INCLUDEDIR as documented in https://cmake.org/cmake/help/latest/guide/tutorial/Installation%20Commands%20and%20Concepts.html, this will do the job in a standard way.

Regards.

Andrej730:
I agree with Esteban - it would be better to if we would stick with simple `install(TARGET LIB)` installation methods and would use default cmake variables like `CMAKE_INSTALL_LIBDIR`, so it would be more standardized.

I believe when we removed `DESTINATION` args it only revealed the problems we had in our custom `LIBDIR` `INCLUDEDIR` `BINDIR` vars handling and we should address the issue there.

If possible, it would be better switch from `LIBDIR` to `CMAKE_INSTALL_LIBDIR` in the cmake scripts - `LIBDIR` and other vars is more like a legacy code.

EstebanDugueperoux2:
I prepare a PR to be reviewed by you 3.

EstebanDugueperoux2:
This is the PR https://github.com/IfcOpenShell/IfcOpenShell/pull/7429
I have leave INCLUDEDIR variable to work on it later.
@dimateos could you tell me if this is ok for you?

dimateos:
Sounds good to me. I just switched off work's PC so I will check it on Monday, I hope you dont mind.

dimateos:
Hello and sorry for the delay, 

These changes along with others in the current develop branch (v0.8.0) work nicely with my build script.
I defined CMAKE_INSTALL_XXXX and all binaries etc fall in the expected folders.

Best regards!

aothms:
Thanks for reporting back @dimateos

PR #7422, opened 2025-11-27
Remove object from non-IFC collections after IFC class assignment.

theoryshaw:
When an object is assigned to an IFC class, it gets added to the appropriate
        IFC container collection. This method removes it from any non-IFC collections
        it was previously in, ensuring the object only appears in its proper IFC
        spatial container.


Gorgious56:
Hello. Wasn't this the previously standard behaviour that was changes when collections were changed to accomodate for non-hardcoded collections to work out of the box ?

theoryshaw:
can you rephrase, didn't quite catch that?

theoryshaw:
was a little discussion here:  https://community.osarch.org/discussion/3242/applying-ifc-class-to-blender-geometry

PR #7434, opened 2025-11-30
cmake: Use correct proj target

EstebanDugueperoux2:
See-Also: https://proj.org/en/stable/development/cmake.html

### DIFF
--- a/cmake/FindCGAL.cmake
+++ b/cmake/FindCGAL.cmake
@@ -47,7 +47,7 @@ else()
         set(CGAL_Boost_USE_STATIC_LIBS "${Boost_USE_STATIC_LIBS}")
     endif()
     find_package(CGAL CONFIG)
-    if(NOT CGAL::CGAL)
+    if(NOT CGAL_FOUND)
         message(
             FATAL_ERROR
             "CGAL_SUPPORT enabled, but CGAL_INCLUDE_DIR wasn't provided and CGAL package couldn't be found."
@@ -56,5 +56,5 @@ else()
     message(STATUS "CGAL: found config at '${CGAL_DIR}'.")
 endif()
 
-target_compile_definitions(CGAL::CGAL INTERFACE IFOPSH_WITH_CGAL)
+target_compile_definitions(CGAL INTERFACE IFOPSH_WITH_CGAL)
 set(SWIG_DEFINES ${SWIG_DEFINES} -DIFOPSH_WITH_CGAL)
--- a/cmake/FindPROJ.cmake
+++ b/cmake/FindPROJ.cmake
@@ -1,0 +1,63 @@
+#
+# Input variables:
+# - `PROJ_INCLUDE_DIR`
+# - `PROJ_LIBRARIES`
+# If input variables are not specified, try to find PROJ config.
+# Input variables could also be provided as environment variables.
+#
+# Output variables:
+# - `PROJ_INCLUDE_DIR`
+# - `PROJ_LIBRARIES`
+#
+
+# To avoid cyclic calls to this file
+list(REMOVE_ITEM CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR})
+
+UNIFY_ENVVARS_AND_CACHE(PROJ_INCLUDE_DIR)
+UNIFY_ENVVARS_AND_CACHE(PROJ_LIBRARIES)
+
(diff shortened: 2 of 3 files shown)


Andrej730:
@EstebanDugueperoux2 

Makes sense, but we also have to search for `PROJ` in `find_package`, not `proj` - otherwise it errors with: `The following configuration files were considered but not accepted: L:/Software/usr/proj-9.2.1/lib/cmake/proj/proj-config.cmake, version: 9.2.1 (package = PROJ, NOT proj)`, because of the check below:

https://github.com/OSGeo/PROJ/blob/d12a8107268cc010a04b2a80e7081c3c9b3251c1/cmake/project-config-version.cmake.in#L18-L24

And let's also add `CONFIG` here? Since there's no `FindPROJ.cmake` available and we're targetting only configs here anyways.

EstebanDugueperoux2:
Hi @Andrej730 I have updated this PR.
I have also tried a build with option enabled but failed because libproj-dev (https://packages.ubuntu.com/jammy/amd64/libproj-dev/filelist) on ubuntu doesn't provide cmake file, on pkgconfig one.

Andrej730:
@EstebanDugueperoux2 
Ah, I see, we missed include/libraries variables for `proj` in 1876a788a9dbb199c9f03ae742ea40650373d7ff. 
Let's add `FindPROJ` and construct a simple `PROJ::proj` target if user provides `PROJ_INCLUDE_DIR` / `PROJ_LIBRARIES` and if they're not provided just search for config?

EstebanDugueperoux2:
Hi @Andrej730 I have added a FindPROJ.cmake and enabled proj option by default, what do you think?
@aothms I see that since https://github.com/IfcOpenShell/IfcOpenShell/commit/4fc2f623cff39dcf47bf05d7e4aa41e8c23d457f the build fails on:

`
IfcOpenShell/src/ifcgeom/mapping/IfcOpenCrossProfileDef.cpp:82:58:   required from here
/usr/include/c++/11/ext/new_allocator.h:162:11: error: no matching function for call to ‘ifcopenshell::geometry::taxonomy::point3::point3(double&, double&, double&, boost::optional<std::__cxx11::basic_string<char> >&)’
`

Regards.

Andrej730:
@EstebanDugueperoux2 

> Hi @Andrej730 I have added a FindPROJ.cmake and enabled proj option by default, what do you think?

I think it's better to keep it `OFF` the way it is now, as it's likely that most users don't need it.

`FindPROJ` code makes sense, but now it seems we're never searching for config files (`proj-config.cmake`) - I think it's the first thing we should try when neither `PROJ_INCLUDE_DIR` and `PROJ_LIBRARY` are provided, to delegate possible future library structure changes to proj's own config file.

And one other thing I think we can do - if variables are not specified and script has failed to locate config, then after we find includes and libraries with `find_path` and `find_library` we can construct a new target with `add_library(PROJ::proj UNKNOWN IMPORTED)` and use it in the main cmake script. So this target would combine both libraries and includes when we link against it.

Andrej730:
> I see that since [4fc2f62](https://github.com/IfcOpenShell/IfcOpenShell/commit/4fc2f623cff39dcf47bf05d7e4aa41e8c23d457f) the build fails on:
> 
> `IfcOpenShell/src/ifcgeom/mapping/IfcOpenCrossProfileDef.cpp:82:58: required from here /usr/include/c++/11/ext/new_allocator.h:162:11: error: no matching function for call to ‘ifcopenshell::geometry::taxonomy::point3::point3(double&, double&, double&, boost::optional<std::__cxx11::basic_string<char> >&)’`

I've submitted a temporary fix in db0974442 to keep build working

EstebanDugueperoux2:
Hi @Andrej730 

Finally ready for review, I have also updated FindCGAL.cmake to fix CI.

Regards.

cmake/FindCGAL.cmake
```diff
@@ -56,5 +56,5 @@ else()
     message(STATUS "CGAL: found config at '${CGAL_DIR}'.")
 endif()
 
-target_compile_definitions(CGAL::CGAL INTERFACE IFOPSH_WITH_CGAL)
+target_compile_definitions(CGAL INTERFACE IFOPSH_WITH_CGAL)
```
Andrej730: Let's keep it `CGAL::CGAL`? Since it is the target name defined above and also the target name exported by `CGAL` package and also exported by conan.

PR #7435, opened 2025-11-30
cmake: Link Eigen3 only against targets needing it

EstebanDugueperoux2:
Use target_link_libraries instead of link_libraries, to avoid linking Eigen3 against all targets.


EstebanDugueperoux2:
Thanks @Andrej730 another PR https://github.com/IfcOpenShell/IfcOpenShell/pull/7434
Before I create new ones.

PR #7442, opened 2025-12-02
Fix large BCF file creation by allowing Zip64

aryobenholzner:
When saving BCF files, currently, `allowZip64` is explicitly set to False. With larger BCFs this leads to the creation crashing with the error `Files count would require ZIP64 extensions` and leaving an incomplete BCF.

Allowing zip64 fixes this.


grofoli:
Same issue here. @Moult @Andrej730 @aothms Is it possible to include this adjustment in the next release?

Big thanks for the work you are doing in ifcopenshell in general!!

aothms:
Thanks :)

Issue #7338, opened 2025-11-10
Add support for IfcOpenCrossProfileDef


RickBrice:
### Feature Description

@aothms 
I'm trying to use an IfcOpenCrossProfileDef to model the surface of a roadway with superelevation transition. When I run my model through IfcConvert, I get an error "Unexpected topology". Looking at the C++ code in geometry_mapping_ifc4x3_add2 there is not a file for IfcOpenCrossProfileDef. So, I'm pretty sure IfcOpenCrossProfileDef is not supported, but I wanted to check before taking a deep dive into implementation.

If implementation is needed, I'm looking for some implementation pointers.

The IfcOpenCrossProfileDef is essentially a profile consisting of points connected by lines. In the context of IfcSectionedSurface, the profile maps to a taxonony::geom_item::ptr as a "face".

aothms:
I think it's equivalent to IfcArbitraryOpenProfileDef in that sense, which we just map to Loop basically.

And make_loft() which is used by sectioned surface can handle both loops and faces https://github.com/IfcOpenShell/IfcOpenShell/blob/080222ffe6ea3e08c1f0a4fc45589cd9a085b1ce/src/ifcgeom/infra_sweep_helper.cpp#L102

So I think the implementation consists of two parts:
- Interpretation of the slope + width into explicit polyline points. 
- Handling of the tag values
  - When mapping
    - IfcOpenCrossProfileDef
    - but also IfcCartesianPointList2D/3D actually has a TagList attribute that could affect the lofting https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcCartesianPointList2D.htm ?
  - When lofting

Are you interested in the correspondences between these tags or should we just start with converting the slope+width pairs into a simple polyline.

For the tags I guess we can add an attribute to loop in the taxonomy

RickBrice:
I will start with interpretation of the slope + width into explicit polyline points. This will address my immediate use case.

I will address tag values as a "phase 2" effort. I think IfcSectionedSolidHorizontal has the same Tag values in the mapping. That implementation should provide some guidance.

RickBrice:
@aothms I think there may be a problem with the implementation of IfcSectionedSurface but I don't know enough about it to be sure.

I have two models that should be equivalent. The surface represents the superelevation bank along a highway curve. The cross sections represent AA, CC, and DD in the figure below. Vertical dimensions are exaggerated.

[...]

Issue #7251, opened 2025-10-15
geom.create_shape() throws error for IfcSectionedSolidHorizontal with IfcArbitraryProfileDefWithVoids as CrossSection


thlemm:
### Bug Description

I recently ran into a slightly mysterious issue while trying to extract shapes from IfcSectionedSolidHorizontal with the geom.create_shape()- function.

The overall goal is to convert IFC-files into GLTF for further use. Generally, the extraction of the shapes works great. But I found an issue with skewed beams defined with IfcSectionedSolidHorizontal. 

The skew at the end of the beams only works either in IFC and GLTF when the Axis attribute in IfcAxis2PlacementLinear is set in the CrossSectionPositions. If Axis is omitted the skew disappears in the GLTF. I added snippets from the IFC-file as example below.

Setting the Axis attribute interestingly results in an unexpected error within geom.create_shape() when the profile of the CrossSection has voids. After quite some testing I couldn’t make both work at the same time: a skewed geometry and a geometry with voids. The error output from the scenario with set Axis attribute and voids (scenario D) is added below. I also added a scenario matrix to further illustrate my problem and some test files either IFC and GLTF. I am currently using 

Can you think of anything that causes this behavior? I’d really appreciate any ideas.

IfcOpenShell 0.8.3.post1
Python 3.12.4

**Example IFC-snippets CrossSectionPositions:**
Example of CrossSectionPositions with set Axis attribute in IfcAxis2PlacementLinear
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/7901638b-639c-4ec9-87db-46958e14a820" />

Example of CrossSectionPositions with no value for Axis attribute in IfcAxis2PlacementLinear
<img width="600"  alt="Image" src="https://github.com/user-attachments/assets/cb69ef2b-82ed-4020-9c11-94b0dca29d69" />

**Scenrario matrix:**
<img width="2104" height="2917" alt="Image" src="https://github.com/user-attachments/assets/ccb13679-7ba4-464c-ab54-ef4412926fab" />

**Example files for all scenarios:**
[A_without_void_without_axis_definition.gltf.txt](https://github.com/user-attachments/files/22927499/A_without_void_without_axis_definition.gltf.txt)
[A_without_void_without_axis_definition.ifc.txt](https://github.com/user-attachments/files/22927497/A_without_void_without_axis_definition.ifc.txt)
[...]

Issue #6386, opened 2025-03-17
IfcSectionedSolidHorizontal doesn't account for orientation of section


RickBrice:
IfcSectionedSolidHorizontal isn't accounting for the orientation of the cross section. Here is an example bridge with girder ends that are skewed at 45 degrees.

[PGSuper1_with_Skew.ifc.txt](https://github.com/user-attachments/files/19293455/PGSuper1_with_Skew.ifc.txt)

From [IfcSectionedSolidHorizontal](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcSectionedSolidHorizontal.htm),  The profile X axis is the direction of RefDirection from [IfcAxis2PlacementLinear](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcAxis2PlacementLinear.htm), and the profile Y axis is the direction of Axis.

The implementation in IfcSectionSolidHorizontal.cpp just gets the location from the axis2linearplacement but doesn't consider the axis or ref direction.

Any ideas how I could fix this?

From usBIM
![Image](https://github.com/user-attachments/assets/463cb4a1-199d-47e5-955b-8eb05b259e67)

![Image](https://github.com/user-attachments/assets/40b5bd17-60a4-4e08-8f9c-a1d464a5d070)

From Bonsai

![Image](https://github.com/user-attachments/assets/7b753bb1-33f1-4b3a-b9dc-3e774a35fa34)

RickBrice:
@aothms can you point me in the right direction?

In the mapping function for IfcSectionedSolidHorizontal, the cross section is mapped into a face and face->matrix is empty for all faces.

https://github.com/IfcOpenShell/IfcOpenShell/blob/3c777a0b85176f3975e35874e506559a702b0a44/src/ifcgeom/mapping/IfcSectionedSolidHorizontal.cpp#L54-L56

The cross section locations are evaluated to determine (1) the longitude (which is "distance along"), and (2) offsets defined in the IfcPointByDistanceExpression.

https://github.com/IfcOpenShell/IfcOpenShell/blob/3c777a0b85176f3975e35874e506559a702b0a44/src/ifcgeom/mapping/IfcSectionedSolidHorizontal.cpp#L58-L72

[...]

PR #7443, opened 2025-12-02
Initial investigation into lofting open profile with tags

aothms:
As discussed in #7338 here is my attempt, very rough, but the idea is simple.

width=0 segments are fused into single points with `set<string>` tags

the idea is that new tags can be introduced as long as they are connected with width=0 tags to other vertices, so that there are no gaps

tags are assumed to be unique within a profile

profile to profile connections are made when `set<string>` Tags between vertices intersect, in a lot of cases, these sets have only one member, so intersection in that case essentially means equality of the singular tags

there is a lot of code, but it comes from a large part due to duplication between the mapping layer and loft.cpp. **I think this needs to be removed so that actually make_loft() already returns a triangulated surface, so that there is no specific handling required anymore using open cascade**.

afbeelding

### DIFF
--- a/src/ifcgeom/infra_sweep_helper.cpp
+++ b/src/ifcgeom/infra_sweep_helper.cpp
@@ -14,6 +14,27 @@ namespace {
 	}
 }
 
+namespace {
+template <typename T, typename Cmp = std::less<T>>
+bool has_intersection(const std::set<T, Cmp>& A,
+                      const std::set<T, Cmp>& B) {
+    auto itA = A.begin();
+    auto itB = B.begin();
+
+    while (itA != A.end() && itB != B.end()) {
+        if (Cmp()(*itA, *itB)) {
+            ++itA;
+        } else if (Cmp()(*itB, *itA)) {
+            ++itB;
+        } else {
+            return true;
+        }
+    }
+    return false;
+}
+
+}
+
 taxonomy::loft::ptr ifcopenshell::geometry::make_loft(const Settings& settings_, const IfcUtil::IfcBaseClass* inst, const taxonomy::function_item::ptr& fn, std::vector<cross_section>& cross_sections)
 {
 	std::sort(cross_sections.begin(), cross_sections.end());
@@ -25,7 +46,7 @@ taxonomy::loft::ptr ifcopenshell::geometry::make_loft(const Settings& settings_,
 	// @todo currently only the case is handled where directrix returns a function_item
 	// @todo this "if" statement is not really required because the function returns at the start if the Directrix is not a function_item function
 	if (fn) {
-		function_item_evaluator evaluator(settings_,fn);
+		function_item_evaluator evaluator(settings_, fn);
 		double start = std::max(0., cross_sections.front().dist_along);
 		double end = std::min(fn->length(), cross_sections.back().dist_along);
 
@@ -45,6 +66,7 @@ taxonomy::loft::ptr ifcopenshell::geometry::make_loft(const Settings& settings_,
 			// parameter is minimum number of steps
 			num_steps = (size_t)std::ceil(param);
 		}
+        auto delta_step = curve_length / num_steps;
 		std::vector<double> longitudes;
 		for (auto& x : cross_sections) {
 			longitudes.push_back(x.dist_along);
@@ -52,13 +74,15 @@ taxonomy::loft::ptr ifcopenshell::geometry::make_loft(const Settings& settings_,
 		longitudes.push_back(std::numeric_limits<double>::infinity());
 		auto profile_index = longitudes.begin();
 		for (size_t i = 0; i <= num_steps; ++i) {
-			auto dist_along = start + curve_length / num_steps * i;
+            auto dist_along = start + delta_step * i;
 			while (dist_along > *(profile_index + 1)) {
 				profile_index++;
 				if (profile_index == longitudes.end()) {
 					// @todo handle this? 
 				}
 			}
+
+			const bool is_last_placement_of_this_profile = profile_index + 1 >= longitudes.end() ? false : ((start + delta_step * (i+1)) > *(profile_index + 1));
 
 			auto relative_dist_along = (dist_along - *profile_index) / (*(profile_index + 1) - *profile_index);
 			const auto& profile_a = cross_sections[std::distance(longitudes.begin(), profile_index)].section_geometry;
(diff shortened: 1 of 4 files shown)


src/ifcgeom/infra_sweep_helper.cpp
```diff
@@ -14,6 +14,27 @@ namespace {
 	}
 }
 
+namespace {
+template <typename T, typename Cmp = std::less<T>>
```
RickBrice: Does Cmp = std::less<T> assume tags have increasing value start to end along the profile? Examples have been A,B,C  and A,B,D,C. What if C,B,A and  C,D,B,A? or something more random Z, A, Q and Z, A, W, Q?

src/ifcgeom/infra_sweep_helper.cpp
```diff
@@ -143,28 +167,203 @@ taxonomy::loft::ptr ifcopenshell::geometry::make_loft(const Settings& settings_,
...
+							Logger::Warning("Mismatching availability tags on loops", inst);
 							return nullptr;
+                        }
+
+						if (w1->tags) {
+							// check uniqueness
+                            std::set<std::string> tags_seen;
+                            for (const auto& t : *w1->tags) {
+								if (tags_seen.find(t) != tags_seen.end()) {
+									Logger::Warning("Duplicate tag '" + t + "' on loft profile", inst);
```
RickBrice, line 193: * The example in https://github.com/bSI-InfraRoom/IFC-infra-unit-test/pull/106 suggests that there can be duplicate Tag values. This may work for branching, but it doesn't for string lines. Unless specifically defined somewhere, Tag values should be unique.
* Doesn't `std::set<std::string>` eliminate the possibility of duplicate tags in the set? Should this be a `std::multiset<std::string>`?

src/ifcgeom/infra_sweep_helper.cpp
```diff
@@ -14,6 +14,27 @@ namespace {
 	}
 }
 
+namespace {
+template <typename T, typename Cmp = std::less<T>>
+bool has_intersection(const std::set<T, Cmp>& A,
```
RickBrice, line 19: would `std::set_intersection` accomplish the same thing?

src/ifcgeom/kernels/opencascade/loft.cpp
RickBrice: No comment - I'm not familiar with the details of open cascade

RickBrice, line 1: No comment - I'm not familiar with the details of open cascade

RickBrice (commented):
This solution seems to be working well - better than what I came up with.

src/ifcgeom/kernels/opencascade/loft.cpp
civilx64: I'm not able to provide much help.  I recognize a few types from OCC but not experienced enough with c++ just yet to provide feedback.  I do have a functioning dev environment that can compile the python bindings and I'm happy to support with data for unit tests.

Generally speaking if there is a way to make this logic generic enough to not need this OCC-specific implementation, that feels like a win.

src/ifcgeom/infra_sweep_helper.cpp
```diff
@@ -45,21 +66,24 @@ taxonomy::loft::ptr ifcopenshell::geometry::make_loft(const Settings& settings_,
 			// parameter is minimum number of steps
 			num_steps = (size_t)std::ceil(param);
 		}
+        auto delta_step = curve_length / num_steps;
 		std::vector<double> longitudes;
```
civilx64, line 70: `longitudes` as a variable name suggests geodetic geometry with spheroids.  Maybe it's just me but seems it would be better to use different terminology.

Assuming that longitudes are mostly parallel to the alignment and connect points between profiles that have the same tage, it might be better to call them `stringlines` or `featurelines`.

src/ifcgeom/taxonomy.h
```diff
@@ -929,6 +929,7 @@ typedef item const* ptr;
 
 				boost::optional<bool> external, closed;
 				boost::optional<taxonomy::function_item::ptr> fi;
```
civilx64, line 931: not a very useful variable name.  Consider refactoring to sth like `function_item`.

src/ifcgeom/infra_sweep_helper.cpp
aothms: I had multiset initially. So I'm open to exploring that further. But I'm unsure if it changes much given how currently things are handled because the current code only cares for set intersections which would be the same for set vs multiset.

src/ifcgeom/infra_sweep_helper.cpp
aothms: I just wanted something that returns bool

src/ifcgeom/infra_sweep_helper.cpp
aothms: They are essentially the OffsetLongitudonal from successive https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcPointByDistanceExpression.htm

civilx64: Ok - makes sense and I am ok to leave as-is.

civilx64:
General comment - how much of the code in this PR overlaps with the implementation for IfcSectionedSolidHorizontal?  I ask because I believe there is a bug related to imperial unit conversions when sweeping profiles to generate sectioned solids.  I did a quick review of `infra_sweep_helper.cpp` and units appeared to be handled correctly, so I'm not sure if the bug report should be part of this PR or handled separately.

I can provide more details in this thread if that makes sense, or create a new issue if you'd prefer to keep this discussion scoped to sectioned surface only (not solids).

aothms:
If you ask me, IfcSectionSolidHorizontal and IfcSectionedSurface are basically the same thing, just at a different place in the ifc class hierarchy because of single inheritance (solid vs surface) and that they use closed vs open profile defs. They both call the function make_loft() in ifcopenshell and are sweeping a face (outer closed wire, zero or more inner wires) vs a single open wire.

RickBrice:
> IfcSectionSolidHorizontal and IfcSectionedSurface are basically the same thing

I agree. The documentation between the two should be much more similar than it is.

From IfcSectionedSurface 
> The profile normal is derived from the associated [IfcAxis2PlacementLinear](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcAxis2PlacementLinear.htm), not necessarily the tangent of the Directrix. The profile Y axis is oriented according to Axis attribute of [IfcAxis2PlacementLinear](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcAxis2PlacementLinear.htm) and the X is derived from the cross product of Directrix and the Axis attribute of [IfcAxis2PlacementLinear](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcAxis2PlacementLinear.htm).

From IfcSectionedSolidHorizontal 
> The profile normal is derived from the associated [IfcAxis2PlacementLinear](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcAxis2PlacementLinear.htm), not necessarily the tangent of the Directrix. The profile X axis is the direction of RefDirection from [IfcAxis2PlacementLinear](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcAxis2PlacementLinear.htm), and the profile Y axis is the direction of Axis."

I think the documentation for IfcSectionedSolidHorizontal is incorrect. https://github.com/buildingSMART/IFC4.x-IF/issues/147

RickBrice:
I think the code @aothms developed works good enough for now. I haven't done that much testing, but for every case I've looked at the treatment of tags for branching works. I'd be ok with merging it in and moving forward to the next topic of string lines.

src/ifcgeom/infra_sweep_helper.cpp
```diff
@@ -227,6 +431,19 @@ taxonomy::loft::ptr ifcopenshell::geometry::make_loft(const Settings& settings_,
 			auto m = (m4b * loft->children.back()->matrix->ccomponents()).eval();
 			loft->children.back()->matrix->components() = m;
 		}
+
```
RickBrice, line 434: The find_closest function isn't used

RickBrice (commented):
When I run the attached file through this simple function, it crashes in loft->print because axis is a nullptr.
[IfcSectionedSurface.ifc.txt](https://github.com/user-attachments/files/24242466/IfcSectionedSurface.ifc.txt)

Not sure what the problem is. When I open the attached file in Blender, the surface does not display. 


~~~
int main(int argc, char** argv)
{
	if (argc != 2) {
		std::cout << "usage: IfcAlignmentDriver <filename.ifc>" << std::endl;
		return 1;
	}

	// Redirect the output (both progress and log) to stdout
	Logger::SetOutput(&std::cout, &std::cout);

	// Parse the IFC file provided in argv[1]
	IfcParse::IfcFile file(argv[1]);
	if (!file.good()) {
		std::cout << "Unable to parse .ifc file" << std::endl;
		return 1;
	}

	ifcopenshell::geometry::Settings settings;
	auto mapping = ifcopenshell::geometry::impl::mapping_implementations().construct(&file, settings);

	//auto profiles = file.instances_by_type<Schema::IfcOpenCrossProfileDef>();
	//for (auto& profile : *profiles)
	//{
	//	profile->toString(std::cout); std::cout << std::endl;
	//	auto mapped_item = mapping->map(profile);
	//	auto loop = ifcopenshell::geometry::taxonomy::dcast<ifcopenshell::geometry::taxonomy::loop>(mapped_item);
	//	loop->print(std::cout, 0);
	//}

	auto surfaces = file.instances_by_type<Schema::IfcSectionedSurface>();
	for (auto& surface : *surfaces)
	{
		auto mapped_item = mapping->map(surface);
		auto loft = ifcopenshell::geometry::taxonomy::dcast<ifcopenshell::geometry::taxonomy::loft>(mapped_item);
		loft->print(std::cout, 0);
	}

   return 1;
}
~~~

theoryshaw:

Conflict Report:

-   [**PR #7443**](https://github.com/IfcOpenShell/IfcOpenShell/pull/7443): Initial investigation into lofting open profile with tags
    -   Conflicting files:  `src/ifcgeom/infra_sweep_helper.cpp`,  `src/ifcgeom/kernels/opencascade/loft.cpp`,  `src/ifcgeom/mapping/IfcOpenCrossProfileDef.cpp`,  `src/ifcgeom/taxonomy.h`
    -   Possible breaking commits:  `610d639c2 Fixes rotation lerp test in make_loft`,  `df7318973 Adds support for IfcOpenCrossProfileDef and branching with IfcSectionedSurface`,  `a34e42a51 Handle infra sweeps of open loops #7338`,  `28f5d6e91 Copy internal status on interpolated loops #7251`,  `4c2844292 Apply skew in infra lofts #6386`

PR #7449, opened 2025-12-05
cmake: Support shared build for rocksdb

EstebanDugueperoux2:


aothms:
I think this is safe since both the windows and linux build script used for the statically linked releases define `ROCKSDB_BUILD_SHARED=Off`

PR #7452, opened 2025-12-06
csv2ifc. Now it is possible to assign also a property from a pset, no…

maxfb87:
…t only from a quantity set.

@steverugi i tried what you have proposed but i'm not sure if it works correctly, but if you want to try you are welcome! :-)

ATM, DO NOT MERGE!!!


steverugi:
@maxfb87 works like a charm! (only the property name to be used, not its Pset)
thanks a lot

maxfb87:
I think this PR can be closed because there is a better implementation in another PR:
https://github.com/IfcOpenShell/IfcOpenShell/pull/7551

@steverugi what do you think?

steverugi:
> I think this PR can be closed because there is a better implementation in another PR: #7551
> 
> @steverugi what do you think?

yes! please go ahead

PR #7459, opened 2025-12-07
cmake: Add install command for qtViewer

EstebanDugueperoux2:
PR to fix qtViewer build.

Hi @aothms @Andrej730 I have this PR among others needing review.
This PR depends of https://github.com/IfcOpenShell/IfcOpenShell/pull/7434 to have CI pass on ifcopenshell (C++) build part.

Regards.

### DIFF
--- a/src/qtviewer/CMakeLists.txt
+++ b/src/qtviewer/CMakeLists.txt
@@ -17,8 +17,6 @@
 #                                                                              #
 ################################################################################
 
-cmake_minimum_required(VERSION 3.1.3)
-
 message("Running CMakeLists.txt in /src/qtviewer")
 
 # Specify the Qt version and components to use
@@ -100,3 +98,5 @@ target_include_directories(${targetName} PUBLIC
 
 get_target_property(targetIncludeDirs ${targetName} INCLUDE_DIRECTORIES)
 message(STATUS "target_include_directories: ${targetIncludeDirs}")
+
+install(TARGETS ${targetName})
--- a/src/qtviewer/ParseIfcFile.cpp
+++ b/src/qtviewer/ParseIfcFile.cpp
@@ -1,4 +1,5 @@
 #include "ParseIfcFile.h"
+#include "../ifcgeom/hybrid_kernel.h"
 
 #include "MessageLogger.h"
 #include "osg/Array"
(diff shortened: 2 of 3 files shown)


src/qtviewer/ParseIfcFile.cpp
DrInfiniteExplorer, line 1: This file and the corresponding header doesn't have any obvious connection to what the PR seem to be about?

src/qtviewer/ParseIfcFile.h
```diff
@@ -2,6 +2,7 @@
 #define PARSEIFCFILE_H
 
 #include "../ifcgeom/Iterator.h"
+#include "../ifcgeom/hybrid_kernel.h"
```
DrInfiniteExplorer: `ParseIfcFile.h` includes `hybrid_kernel.h`, but it does not use anything from that header. This unnecessarily expands the header’s transitive dependency footprint and slows compilation of all includers. Move the include to the .cpp file where it is actually needed. This reduces API exposure and improves build times.

src/qtviewer/CMakeLists.txt
```diff
@@ -17,8 +17,6 @@
 #                                                                              #
 ################################################################################
 
-cmake_minimum_required(VERSION 3.1.3)
```
DrInfiniteExplorer, line 20: Why is this removed? Other PRs you've requested still have cmake files for still have minimum required versions

DrInfiniteExplorer (commented):
This looks like a split-purpose PR?
Left a note about header include transitivity

src/qtviewer/CMakeLists.txt
EstebanDugueperoux2: Hi @DrInfiniteExplorer , this cmake command is already in root CMakeLists.txt (https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/cmake/CMakeLists.txt#L20) then no need to have it here.
I think it is a remainder perhaps of a git module.

src/qtviewer/ParseIfcFile.cpp
EstebanDugueperoux2: This is also to fix compilation issue.
As BUILD_QTVIEWER option is disabled by default, a refactoring has been done in Iterator without impacting this component.

src/qtviewer/ParseIfcFile.h
EstebanDugueperoux2: Remark taken into account.

aothms:
Thanks all

EstebanDugueperoux2:
Hi @aothms thanks for the merge.
In my opinion, the most urgent PR is https://github.com/IfcOpenShell/IfcOpenShell/pull/7434 to fix the CI about compilation.

PR #7498, opened 2025-12-16
New search tool interface

falken10vdl:
This is a proposal for a simplified search tool interface.

image

The idea is that a filter is a single chain of lines. The lines are processed from top to bottom and each line you can take three possible actions: ADD, SUBTRACT or FILTER (intersect).

The filter is a machine that starts with nothing and reads the lines in order to build a set of elements.
This is how it works:
- The filter set is initialized to an empty set
- The first row is read and it has implicitly an ADD action (you cannot change the action for the first row), so the elements of the first row are added to the filter set
- The next row is read and depending on the action the follwoign can happen:
  1. ADD: The elements of that row are added to the filter set
  2. SUBTRACT: The elements of that row are removed from the filter set 
  3. FILTER: The elements of that row are INTERSECTED with the ones of the filter set

The  process continues with the following rows

Cheers


falken10vdl:
This one superseeds https://github.com/IfcOpenShell/IfcOpenShell/pull/7315 

It provides:
- suggestions for the filter rows (based on availble information form the ifc file)
image

- Addon settings:

  image

  - toggle to enable a NEW filter mode as described above in this PR.
    image
 
  - toggle for the DEFAULT filter mode to add explictly ADD/SUBTRACAT/FILTER for Class and GlobalID rows (those are the only ones that apart from FILTER can also be used for adding and subtracting elements)
    image


Cheers!

PR #7517, opened 2025-12-26
Feat/use conan

EstebanDugueperoux2:


EstebanDugueperoux2:
Hi @aothms @Andrej730 and happy new year :)

I try to make swig generation works through conan by using latest swig release (4.4.0) but I have a strange error:

`error: ‘SwigValueWrapper<T>::SwigValueWrapper(const SwigValueWrapper<T>&) [with T = IfcGeom::OpaqueCoordinate<3>]’ is private within this context`

I have no much knowledge of swig and its usage in IfcOpenShell seems advanced.
Do you have an idea of the reason of this failure?

Regards.

Issue #7520, opened 2025-12-29
queries with backticks don't seem to be working


theoryshaw:
### Bug Description

Hi @falken10vdl 

After [this](https://github.com/falken10vdl/IfcOpenShell/commit/1fd8878796c80ca136001f47bf0ee242073d4b60) commit, it seems things like
```
``round({{Qto_SpaceBaseQuantities.NetFloorArea}}, 1)`` SF
```
...don't seem to be working. 

### Attachments

<img width="689" height="738" alt="Image" src="https://github.com/user-attachments/assets/b5ad27ac-20be-4ff1-96b6-55a2306e6373" />

<img width="721" height="733" alt="Image" src="https://github.com/user-attachments/assets/4377cfa8-33cb-41ae-b505-ecdeb64d0962" />

### Debug and Error Output

```true

```

falken10vdl:
You are right! While rebasing all this I left the point in non editing mode to pass the element so it formats properly and does not default to 0.

Please take a look to https://github.com/IfcOpenShell/IfcOpenShell/pull/7521

Thanks!

falken10vdl:
@theoryshaw In the latest https://github.com/falken10vdl/bonsaiPR/releases/tag/v0.8.4-alpha2512281535 you can test it. Please let me know if you are missing something.
Thanks!

PR #7521, opened 2025-12-29
Update formatting expression evaluation to include element context

falken10vdl:
This addresses https://github.com/IfcOpenShell/IfcOpenShell/issues/7520
In non editing mode the element context was not passed and the resolution for that format including elements information was defaulting to wrong value.

Cheers!


falken10vdl:
@theoryshaw  I have tested this PR with a number of different combinations of Intelligent TAGs:
image

Please let me know if you still see any issue.
Cheers!

theoryshaw:
yes, seems to work great!

PR #7523, opened 2025-12-30
Fix: hybrid kernel sigsegv fallback (0.8.0)

ZijianWang-ZW:
Problem: **OpenCASCADE may crash with SIGSEGV/SIGBUS on complex boolean/CSG inputs,** aborting the whole process.

Solution: Install temporary SIGSEGV/SIGBUS handlers around each kernel call and use sigsetjmp/siglongjmp to return control to HybridKernel, clear partial results, and try the next configured kernel (e.g. CGAL). Uses an alternate signal stack best-effort.

Note: Signal-based recovery is inherently unsafe and best-effort; it aims to avoid whole-process abort, not guarantee recovery.


src/ifcgeom/hybrid_kernel.h
```diff
@@ -97,9 +186,22 @@ namespace ifcopenshell {
...
+						if (sigsetjmp(hybrid_kernel_sig_jmp_buf, 1) == 0) {
+							try {
+								success = k->convert(item, rs);
+							} catch (...) {
+								success = false;
+							}
+						} else {
+							// We got here via siglongjmp after SIGSEGV/SIGBUS.
+							fprintf(stderr, "[HYBRID_KERNEL] Caught SIGSEGV/SIGBUS in kernel '%s', trying next kernel...\n",
+								k->geometry_library().c_str());
```
aothms:
```suggestion
							Logger::Error("[HYBRID_KERNEL] Caught SIGSEGV/SIGBUS in kernel '" + k->geometry_library() + "', trying next kernel...");
```

aothms: Please change the fprintfs to logger::error

src/ifcgeom/hybrid_kernel.h
```diff
@@ -53,6 +76,72 @@ namespace {
...
+		bool handlers_installed_ = false;
+
+		// Use an alternate signal stack to improve odds when stack is corrupted.
+		SigGuard() {
+			hybrid_kernel_sig_caught = 0;
+
+			// Install alt stack (best-effort)
+			stack_t ss{};
+			ss.ss_size = 64 * 1024;
+			ss.ss_sp = std::malloc(ss.ss_size);
```
aothms: I wonder if it's safer to allocate at program startup when we still have the guarantee of a sane heap

src/ifcgeom/hybrid_kernel.h
```diff
@@ -53,6 +76,72 @@ namespace {
...
+			if (ss.ss_sp && sigaltstack(&ss, &old_stack_) == 0) {
+				stack_installed_ = true;
+			} else {
+				if (ss.ss_sp) std::free(ss.ss_sp);
+			}
+
+			struct sigaction sa{};
+			sa.sa_handler = hybrid_kernel_sig_handler;
+			sigemptyset(&sa.sa_mask);
+			sa.sa_flags = SA_ONSTACK | SA_NODEFER;
```
aothms: Are you sure about SA_NODEFER here?

aothms (changes requested):
I don't have experience with this kind of signal handling. I'm not against merging this in, but I'm not convinced we should enable this by default. OCCT data is a complex ref counted data soup with potentially lots of shared data, so letting the process continue in such a corrupted state can have quite surprising consequences.

Because of this I would say move this into a setting in ConversionSettings like --handle-signals or --continue-after-signals / -crashes

Ultimately the more robust way of handling this is to use proper process isolation. I think with the RocksDB file back-end we have quite a reasonable way of delivering this because it prevents us from implementing the position-independent pointer logic in typically mmap based interprocess workflows. So the Iterator could instead interface with a secondary process because the cost of parsing essentially goes to zero in this scenario (no SPF anymore only a KV-store with all the indexes already populated).

Thoughts?

src/ifcgeom/hybrid_kernel.h
ZijianWang-ZW: Done. All 4 `fprintf(stderr, ...)` calls replaced with `Logger::Error(...)`. Also removed the now-unnecessary `#include <cstdio>` and `#include <cstdlib>`.

Additionally, extracted a `guarded_kernel_call()` template helper to avoid duplicating the signal handling pattern across all 4 methods.

src/ifcgeom/hybrid_kernel.h
ZijianWang-ZW: Good point. Replaced `std::malloc`/`std::free` with a pre-allocated `thread_local` static array:

```cpp
static constexpr size_t HYBRID_KERNEL_SIG_STACK_SIZE = 64 * 1024;
alignas(16) thread_local char hybrid_kernel_sig_stack[HYBRID_KERNEL_SIG_STACK_SIZE];
```

This is allocated once per thread at first access (well before any potential corruption), and never freed. The `SigGuard` constructor just points `sigaltstack` at this buffer. No more heap allocation during guard setup.

src/ifcgeom/hybrid_kernel.h
ZijianWang-ZW: You're right to question it — removed `SA_NODEFER`. 

Without `SA_NODEFER`, SIGSEGV/SIGBUS are automatically masked while the handler executes. Since we call `sigsetjmp(..., 1)` (with `savesigs=1`), the signal mask is saved at the `sigsetjmp` point and restored by `siglongjmp`. So the flow is:

1. `sigsetjmp` saves current mask (signals unmasked)
2. Signal fires → handler runs with SIGSEGV masked (safe)
3. `siglongjmp` restores saved mask → signals unmasked again
4. `SigGuard` destructor restores original handlers

With `SA_NODEFER`, if `siglongjmp` or subsequent code triggered another fault, the handler would re-enter and potentially infinite-loop. Removing it is strictly safer.

ZijianWang-ZW:
## Addressed all review comments

@aothms Thanks for the review! All 4 points have been addressed in commit d01b5aa:

| Review Comment | Change |
|---|---|
| **Make opt-in** | Added `--signal-handler-crash-recovery` setting in `ConversionSettings.h` (default: `false`). Signal handling is completely inactive unless explicitly enabled. |
| **fprintf → Logger::Error** | All 4 `fprintf(stderr, ...)` replaced with `Logger::Error(...)`. Extracted `guarded_kernel_call()` template helper to reduce duplication. |
| **Pre-allocate signal stack** | Replaced `malloc`/`free` with `thread_local` static 64KB array. Allocated once per thread at first access, no heap allocation during guard setup. |
| **SA_NODEFER** | Removed. `sigsetjmp(savesigs=1)` handles mask restoration via `siglongjmp`, making `SA_NODEFER` unnecessary and risky (infinite re-entry on nested faults). |

**Additional improvements:**
- `#ifndef _WIN32` platform guards around all POSIX signal code
- Removed `#include <cstdio>` and `#include <cstdlib>` (no longer needed)

**Tested** with `WSW-XX-TRA-S2-001-24-IFCV-Tragwerk.ifc` (`--kernel hybrid-opencascade-cgal --signal-handler-crash-recovery`):
- 1 SIGSEGV caught in OpenCASCADE → fell back to CGAL successfully
- Process completed in ~8 minutes, 125MB .obj output, no abort

Ready for re-review when you have time.

BIMvoice:
Thanks for tackling this. We build IfcOpenShell locally, so we compiled this branch and exercised it. Sharing findings in case they help.

**Compiles and runs fine.** Built from your PR head (IFC4, AppleClang/arm64), the `--signal-handler-crash-recovery` flag works and converts fixtures without issue. Worth noting the branch's merge-base is a fair way behind the current v0.8.0 tip now, so a rebase would be useful before merge.

**A concurrency concern in `SigGuard` (`hybrid_kernel.h`).** The jump buffer, stack and caught-flag are all correctly `thread_local`, so cross-thread `siglongjmp` isn't a problem. The issue is the handler installation itself: `SigGuard`'s constructor calls `sigaction(SIGSEGV/SIGBUS, ...)` and its destructor restores the previously-saved disposition, but signal dispositions are process-wide with no per-thread variant, and each thread's save/restore is unsynchronized against the others.

IfcOpenShell runs geometry conversion concurrently: `Iterator`'s `process_concurrently()` dispatches cloned kernels via `std::async` when threads > 1 (`-j`/`--threads`, which Bonsai and auto-detection routinely set above 1). Under that, one thread's `SigGuard` destructor can restore the pre-handler (default) disposition while another thread is still inside a guarded region; if that thread then faults, it now hits the OS default action and the whole process is killed, non-deterministically. We extracted the `SigGuard`/`guarded_kernel_call` code into a standalone harness to confirm: serial runs recovered 500/500 every time, but 4 threads x 500 forced faults killed the process outright in 9 of 10 runs.

A fix that keeps the approach: install the handler once (e.g. `std::call_once`) and track active guarded regions with an atomic refcount, only restoring the prior disposition when it returns to zero (or serialize install/restore with a mutex). Separately, and not blocking, `siglongjmp` out of arbitrary OCC/CGAL stack frames is inherently risky if a heap-allocator or internal lock was held at the fault point, so this is best kept opt-in as you've done.

The underlying goal is a real one we've run into ourselves (one bad element aborting a whole drawing), so it would be good to see this land once the threading case is handled.

PR #4637, opened 2024-05-08
Graceful handling of instance deletion in python

aothms:
Accessing an instance that got deleted now raises a catchable exception. Still need to let this percolate a bit in order to make this a bit more backwards compatible on the C++ side of things.

```
>>> import ifcopenshell
>>> f = ifcopenshell.file()
>>> inst = f.createIfcWall('a')
>>> inst2 = f.by_guid('a')
>>> f.remove(inst)
>>> inst2
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\tkrij\miniconda3\envs\ifopsh08\lib\site-packages\ifcopenshell\entity_instance.py", line 307, in __repr__
    return repr(self.wrapped_data)
  File "C:\Users\tkrij\miniconda3\envs\ifopsh08\lib\site-packages\ifcopenshell\ifcopenshell_wrapper.py", line 8045, in __repr__
    return _ifcopenshell_wrapper.entity_instance___repr__(self)
RuntimeError: No longer availabe
```

PR #7534, opened 2026-01-05
C++ datamodel v1.0

aothms:
For the past couple of days I have been rethinking what a sane v1.0 data model on the C++ side would look like that balances safety, readability, performance and a bit more modern language features. I still have work todo but wanted to share a bit of what's going on.

This has huge implications on client C++ code, so to feel the pain I went through most of the code base already. @RickBrice as one of the more active C++ devs lately I'm curious what your thoughts are.

Outcome:

```
>>> import ifcopenshell
>>> f = ifcopenshell.file(schema='ifc4x3_add2')
>>> f.createIfcWall()
#1=IfcWall($,$,$,$,$,$,$,$,$)
>>> inst = f[1]
>>> f.remove(f[1])
>>> inst
  File "C:\Users\tkrij\miniconda3\envs\ifopsh08\lib\site-packages\ifcopenshell\ifcopenshell_wrapper.py", line 5999, in __repr__
    return _ifcopenshell_wrapper.entity_instance___repr__(self)
RuntimeError: Trying to access deleted instance reference
```

Before:

```C++
	IfcSchema::IfcOpenShell* shell = new IfcSchema::IfcOpenShell(faces);
	IfcSchema::IfcConnectedFaceSet::list::ptr shells(new IfcSchema::IfcConnectedFaceSet::list);
	shells->push(shell);
	IfcSchema::IfcFaceBasedSurfaceModel* surface_model = new IfcSchema::IfcFaceBasedSurfaceModel(shells);

	IfcSchema::IfcRepresentation::list::ptr reps(new IfcSchema::IfcRepresentation::list);
	IfcSchema::IfcRepresentationItem::list::ptr items(new IfcSchema::IfcRepresentationItem::list);

	items->push(surface_model);

	IfcSchema::IfcShapeRepresentation* rep = new IfcSchema::IfcShapeRepresentation(
		0, std::string("Facetation"), std::string("SurfaceModel"), items);

	reps->push(rep);
	IfcSchema::IfcProductDefinitionShape* shapedef = new IfcSchema::IfcProductDefinitionShape(boost::none, boost::none, reps);

	return shapedef;
```

After:

```C++

    auto shell = f.create<IfcSchema::IfcOpenShell>();
    shell.setCfsFaces(faces);

    auto surface_model = f.create<IfcSchema::IfcFaceBasedSurfaceModel>();
    surface_model.setFbsmFaces({shell});

	auto rep = f.create<IfcSchema::IfcShapeRepresentation>();
    rep.setRepresentationIdentifier("Tessellation");
    rep.setRepresentationType("SurfaceModel");
    rep.setItems({surface_model});

	auto shapedef = f.create<IfcSchema::IfcProductDefinitionShape>();
    shapedef.setRepresentations({rep});

	return shapedef;
```

Rationale

reason for weak_ptr encapsulated in object:
- safety and matches semantics (the instance pointers always had their lifetime scoped to the file - this is now explicit and enforced)
- objects can cast, so variants (EXPRESS SELECT) can work with implicit casts instead of virtual inheritance
- elimination of virtual methods might help performance (although the declaration* is essentially the same mechanism) and reduce dependency on rtti in case that ever becomes a concern
- elimination of virtual methods unifies latebound and earlybound entity definitions
- which means it's possible now to do geometry interpretation on late bound schemas (previously not possible because of dependency on dynamic_cast / rtti)
- cleaner separation between early bound statically typed layer and attribute storage container (which was a bit muddy in earlier releases but essentially present)
- ability to redeclare the type of entity instances by assigning a new declaration* which was previously not possible as it was baked into the typeid

reason for creation as part of file
- might be more efficient than create() + add()
- necessary for rocksdb anyway, because file context is necessary for attribute storage in key-value store
- instances not stored in file leak easily and have missing behaviour regarding inverses which depend on the file maps

reasons for no constructor:
- readability because no named args
- easier to adapt to multiple schemas
- anyway immutability is not an option
- reduce binary size
- one way to do things
- don't like argument forwarding because is created as part of file now

no custom class anymore for aggregates (aggregate_of_instance)
- we can use aggregate initializers now and stl features
- but casts are a bit uglier


src/ifcparse/express.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include "InstanceData.h"
+
+const IfcParse::declaration& express::Base::declaration() const {
+    return *data()->declaration();
+}
+uint32_t express::Base::identity() const { return data()->identity(); }
+
+uint32_t express::Base::id() const { return data()->id(); }
+
+const InstanceData* express::Base::data() const {
```
RickBrice: This method, as well as the non-const version, defeats using smart pointers. Once the raw pointer is returned it is no longer managed and could be deleted by the caller. Why not

~~~
auto sp = data_.lock();
if (sp) {
   return sp;
} else {
   throw std::runtime_error("Trying to access deleted instance reference")
}
~~~

RickBrice:
Does the weak_ptr encapsulated in the express::Base retain file scope lifetime of objects? It seems like shared_ptr would be required. Though, I don't have a complete understanding.

Creating entities as part of the file without constructors has a trade-off. Instead of create() + add() a user of the SDK must do create() + multiple parameter initialization steps by calling set() methods. This isn't bad, but it has these consequences:
* Multiple set() methods must be invoked instead of a single constructor
* Constructors had all parameters needed for proper construction and initialization. Now the caller needs to know which set() methods to call for all of the required initialization parameters. Failure (or forgetting) to initialize required parameters will likely lead to unwanted behavior.
* Perhaps consider adding an initialize() method that has all the parameters of the old constructors. They the process is obj=f.create() + obj.initialize().

I very much like eliminating the custom classes for aggregates. I found them difficult to use. It was difficult to know which one of the classes where needed in different situations. Their interface wasn't the same as std collections, so looping wasn't as easy and they couldn't be used in the std algorithms.

This is going to break a lot of my code, but the fixes appear to be easy to deal with.

@aothms Let's set up a time for you to walk me through the bigger picture. I'm sure there are lots of details that I'm missing.

src/examples/IfcAlignment.cpp
```diff
@@ -17,478 +17,423 @@
...
-// https://www.fhwa.dot.gov/bridge/pubs/hif22034.pdf
-//
-// Sections and page number for this document are cited in the code comments.
-//
-// This examples differs from IfcSimplifiedAlignment because it builds the
-// alignment explicitly
-
-// Disable warnings coming from IfcOpenShell
-#pragma warning(disable : 4018 4267 4250 4984 4985)
+/********************************************************************************
```
RickBrice: It looks like a different example has been applied over the IfcAlignment.cpp example. The revised file is significantly different. Is this intentional?

src/ifcparse/express.cpp
aothms: That's true, but I have the feeling that maybe this should be marked private instead and then it doesn't really matter anymore.

src/examples/IfcAlignment.cpp
aothms: Sorry that was sloppy, it's back and updated for the code changes

aothms:
> Does the weak_ptr encapsulated in the express::Base retain file scope lifetime of objects? It seems like shared_ptr would be required. Though, I don't have a complete understanding.

Yes the file have the shared_ptr not so much for sharing ownership but just so that we have the option to have a weak_ptr derived from them. So that when you have instances pointing to a deleted file or to instances from the file that have been deleted.

> Now the caller needs to know which set() methods to call for all of the required initialization parameters.

Yes that's very valid. The discovery is indeed not really there.

I'm also fine with forwarding the arguments from file.create and reinstate the constructor. But as I said since the arguments are not named I find them a bit unclear especially with all the std::nullopts for all the optional arguments, it's hard to interpret. Boost has this named parameters idea, but I think it's a bit lengthy to setup https://www.boost.org/doc/libs/latest/libs/graph/doc/bgl_named_params.html

> Perhaps consider adding an initialize() method that has all the parameters of the old constructors. They the process is obj=f.create() + obj.initialize().

I think that's indeed also a nice compromise. Then we also have autocompletion in our IDE because I think with std::forward from file.create() I think IDE's will be pretty clueless.

Very flexible to have a call these days, let me know some times that suit you please.

RickBrice:
I emailed your personal account with some proposed times for next week

aothms:
@RickBrice The initialize() method is there. There are now three options:

```C++
        // 1.
        auto rel = file.create<IfcSchema::IfcRelVoidsElement>();
        rel.setGlobalId(guid());
        rel.setOwnerHistory(file.getSingle<IfcSchema::IfcOwnerHistory>());
        rel.setRelatingObject(roof);
        rel.setRelatedObjects({south_roof_part, north_roof_part});

        // 2.
        file.create<IfcSchema::IfcRelAggregates>(
            guid(), 
            file.getSingle<IfcSchema::IfcOwnerHistory>(), 
            std::nullopt, 
            std::nullopt, 
            roof, 
            std::vector<::Ifc2x3::IfcObjectDefinition>{south_roof_part, north_roof_part}
        );

        // 3.
        file.create<IfcSchema::IfcRelAggregates>().initialize(
            guid(),
            file.getSingle<IfcSchema::IfcOwnerHistory>(),
            std::nullopt,
            std::nullopt,
            roof,
            {south_roof_part, north_roof_part});
```

For option 2 I couldn't get the untyped initializer list to work. I'm not so familiar with them.

Maybe maybe if we have initialize() `return *this` again, it's also not necessary to do argument forwarding from create(), because then you can do `file::create<T>().initialize(args...)` and have the created and initialized instance returned in one anonymous statement.

@Andrej730 @Moult I think it's also a good time for the both of you to have a look. I'm getting to a point that the tests are almost passing. I still need to do quite a bit of cleaning up. I think quite some docstrings and annotations disappeared in the process of moving some things to the SWIG generated classes, I will try and get them back. Things feel quite a bit snappier without the `wrapped_data` indirections and the constant decorate-undecorate when setting and getting attributes. But I didn't yet find time to test this.

src/ifcparse/IfcUtil.cpp
```diff
@@ -122,13 +122,14 @@ void IfcUtil::unescape_xml(std::string& str) {
...
 }
 
-void express::Entity::populate_derived() {
-    for (auto it = declaration().as_entity()->derived().begin(); it != declaration().as_entity()->derived().end(); ++it) {
-        if (*it) {
-            set_attribute_value(
-                std::distance(declaration().as_entity()->derived().begin(), it),
-                Derived{}
-            );
+void InstanceData::populate_derived_() {
```
RickBrice: there is a lot of getting of the begin and end iterator. Probably not a big deal for any one entity, but over thousands of entities in a model, would it be more efficient to do something like this?

~~~
if (auto* ent = declaration_->as_entity()) {
   auto begin = ent->derived().begin();
   auto end = ent->derived().end();
   for ( auto it = begin; it != end; ++it) {
      if (*it) {
         set_attribute_value(
            std::distance(begin,it),Derived{}):
      }
   }
~~~

RickBrice:
@aothms 

I'm not as knowledgeable on C++ as you, but from a little googling, there does seem to be an untyped initializer list in c++. The closest thing I could find is
~~~
#include <iostream>
#include <initializer_list>
#include <any>
#include <typeinfo>

void printAnyList(std::initializer_list<std::any> values) {
    for (const auto& v : values) {
        if (v.type() == typeid(int))
            std::cout << std::any_cast<int>(v) << " ";
        else if (v.type() == typeid(double))
            std::cout << std::any_cast<double>(v) << " ";
        else if (v.type() == typeid(const char*))
            std::cout << std::any_cast<const char*>(v) << " ";
    }
    std::cout << "\n";
}

int main() {
    printAnyList({1, 2.5, "hello"}); // works with mixed types
}
~~~

This does not seem like a workable solution.

Returning `*this` from `initialize()` seems like a good idea. Then you can do something like:
~~~
        auto myrelagg = file.create<IfcSchema::IfcRelAggregates>().initialize(
            guid(),
            file.getSingle<IfcSchema::IfcOwnerHistory>(),
            std::nullopt,
            std::nullopt,
            roof,
            {south_roof_part, north_roof_part});
~~~

RickBrice:
Another thing comes to mind, but not related to initialization. Not sure how to say it generically, so I'll do it with an example. To add a Related Object to IfcRelAggregates after it is constructed you need to do the following
~~~
auto rel = file.create<IfcRelAggregates>().initialize(...);
auto related_objects = rel.RelatedObjects(); // causes a vector copy
related_objects.push_back(new_related_object);
rel.setRelatedObjects(related_objects); // causes another vector copy
~~~

Consider the following
~~~
class IfcRelAggregates : public...
{
public:
   const std::vector<::Ifc4x3_add2::IfcObjectDefinition>& RelatedObjects() const; // no copy and constness is preserved
   void addRelatedObject(const ::Ifc4x3_add2::IfcObjectDefinition& relobj); // relobj can be added into the internal storage without copying the vector of related objects twice
};
~~~

Before:
~~~
        auto myrel = file.create<IfcSchema::IfcRelAggregates>().initialize(
            guid(),
            file.getSingle<IfcSchema::IfcOwnerHistory>(),
            std::nullopt,
            std::nullopt,
            roof,
            {south_roof_part, north_roof_part});
       auto relobjs = myrel.RelatedObjects();
      relobjs.push_back(east_roof_part);
      myrel.setRelatedObjects(relobjs);
~~~

After:
~~~
        auto myrel = file.create<IfcSchema::IfcRelAggregates>().initialize(
            guid(),
            file.getSingle<IfcSchema::IfcOwnerHistory>(),
            std::nullopt,
            std::nullopt,
            roof,
            {south_roof_part, north_roof_part});
       myrel.addRelatedObject(east_roof_part); // easy to add... could also be myrel.addRelatedObjects({east_roof_part,west_roof_part});

      auto relobjs = myrel.RelatedObjects(); // no copy
     for(auto relobj : relobjs) {
        relobj.Name(); // do something interesting here
     }
~~~

The same concept could be applied to all of the different Rel types, and other things with lists like IfcSectionedSolidHorizontal.CrossSections

Moult:
I hope I understand what's happening here, to me it means:

 - No more segfault on accessing removed element to solve #4637 
 - No more wrapped_data (I assume anywhere we currently use wrapped_data in the code needs to change), in exchange for a general all round performance increase when accessing (direct? inverse?) attributes on entity instances?
 - Smaller binary due to no constructors so faster WASM loading?
 - ... anything else?

RickBrice:
@aothms what is the current state of this?
I’m developing more and more on the C++ side and it’s all going to break with this update. I would like to minimize the pain.

src/ifcparse/IfcUtil.cpp
aothms: I can't imagine this would matter with compiler optimizations, but would be good to check the disassembly.

aothms:
Regarding this comment below. Yes, this is something that has always been incredibly wasteful. Also on the python side. And quite problematic because there are a lot of 1-to-many objectified relationships where the many are sometimes really many many. Like the ContainedInStructure relationship in your typical building. Beyond building storey there is typically very little decomposition so in a typical bad case you can literally have a million nodes into a single aggregate attribute, where this copying really starts to exhibit some nasty non-linear behaviour if you're incrementally adding things to it.

Since this is also on the python-side. I'd prefer to solve this in the attribute value and not in the schema (python doesn't have acces to the schema). That means re-introducing some custom vector-alternative that has more context (needs to know entity instance and attribute index essentially) so that it can update the inverse relationships also when doing a push_back (or an erase).

Previously this was hardly possible at all, because the vectors returned from the attribute storage where constructed on the fly. Now there is at least a vector<Entity> type in our storage variant.

Maybe this also means we can stop with casting-by-copying of the vector types, because this is also really silly. We store them as untyped vector<Entity> and then create a copy of vector with element type of a specific subtype and do some static or dynamic cast on the elements. But we can also do this lazily with a custom iterator.

Becomes even more complex when we have to consider `vector<vector<>>` but that is really an exception maybe not worth to much time.

I certainly think this is something worth exploring.

```
template <typename T>
class ifcopenshell::vector {
  class custom_iterator_over_this_thing {};

  Entity owner;
  size_t attribute_index;

  custom_iterator_over_this_thing begin() const;
  custom_iterator_over_this_thing end() const;

  void push_back(const T& t) {
    // add to values
    // update inverses
  }
};
```
> Another thing comes to mind, but not related to initialization. Not sure how to say it generically, so I'll do it with an example. To add a Related Object to IfcRelAggregates after it is constructed you need to do the following
> 
> ```
> auto rel = file.create<IfcRelAggregates>().initialize(...);
> auto related_objects = rel.RelatedObjects(); // causes a vector copy
> related_objects.push_back(new_related_object);
> rel.setRelatedObjects(related_objects); // causes another vector copy
> ```
> 
> Consider the following
> 
> ```
> class IfcRelAggregates : public...
> {
> public:
>    const std::vector<::Ifc4x3_add2::IfcObjectDefinition>& RelatedObjects() const; // no copy and constness is preserved
>    void addRelatedObject(const ::Ifc4x3_add2::IfcObjectDefinition& relobj); // relobj can be added into the internal storage without copying the vector of related objects twice
> };
> ```
> 
> Before:
> 
> ```
>         auto myrel = file.create<IfcSchema::IfcRelAggregates>().initialize(
>             guid(),
>             file.getSingle<IfcSchema::IfcOwnerHistory>(),
>             std::nullopt,
>             std::nullopt,
>             roof,
>             {south_roof_part, north_roof_part});
>        auto relobjs = myrel.RelatedObjects();
>       relobjs.push_back(east_roof_part);
>       myrel.setRelatedObjects(relobjs);
> ```
> 
> After:
> 
> ```
>         auto myrel = file.create<IfcSchema::IfcRelAggregates>().initialize(
>             guid(),
>             file.getSingle<IfcSchema::IfcOwnerHistory>(),
>             std::nullopt,
>             std::nullopt,
>             roof,
>             {south_roof_part, north_roof_part});
>        myrel.addRelatedObject(east_roof_part); // easy to add... could also be myrel.addRelatedObjects({east_roof_part,west_roof_part});
> 
>       auto relobjs = myrel.RelatedObjects(); // no copy
>      for(auto relobj : relobjs) {
>         relobj.Name(); // do something interesting here
>      }
> ```
> 
> The same concept could be applied to all of the different Rel types, and other things with lists like IfcSectionedSolidHorizontal.CrossSections

aothms:
> @aothms what is the current state of this? I’m developing more and more on the C++ side and it’s all going to break with this update. I would like to minimize the pain.

You decide:

- Forget about this for now
- Prioritize and merge asap
- Let is percolate a bit, but at least make it somewhat stable so that you can work off of it

RickBrice:
> Since this is also on the python-side. I'd prefer to solve this in the attribute value and not in the schema (python doesn't have access to the schema). That means re-introducing some custom vector-alternative...

I liked the idea of using STL without something special. I didn't think about dealing with inverses. The fundamental question is then, how much of an actual problem is this? Is it worth the effort to solve it? 

> You decide:
> Forget about this for now

It seems like we both have other things going on that are higher priority. We can forget about it for now. While there will be pain on the C++ side, it doesn't appear to be unbearable.

aothms:
> I liked the idea of using STL without something special. I didn't think about dealing with inverses. The fundamental question is then, how much of an actual problem is this? Is it worth the effort to solve it?

Yes me too, but on the other hand STL is designed to work with custom containers because everything accepts iterators. I think people in a well-designed library would expect to be able to do `relationship.RelatedObjects().push_back(something_new)`. Since anyway we're giving ourselves some time I will see what the implications are.

> While there will be pain on the C++ side, it doesn't appear to be unbearable.

I think if we have `file.create<Something>().initialize()` the pain will be quite regexable.. hopefully..

theoryshaw:

Conflict Report:

-   [**PR #7534**](https://github.com/IfcOpenShell/IfcOpenShell/pull/7534): C++ datamodel v1.0
    -   Conflicting files:  `src/examples/IfcAdvancedHouse.cpp`,  `src/examples/IfcOpenHouse.cpp`,  `src/ifcgeom/infra_sweep_helper.cpp`,  `src/ifcgeom/kernels/opencascade/OpenCascadeKernel.cpp`,  `src/ifcgeom/kernels/opencascade/loft.cpp`,  `src/ifcgeom/kernels/opencascade/loop.cpp`,  `src/ifcgeom/mapping/IfcAxis2PlacementLinear.cpp`,  `src/ifcgeom/mapping/IfcObjectPlacement.cpp`,  `src/ifcgeom/mapping/IfcOpenCrossProfileDef.cpp`,  `src/ifcgeom/mapping/IfcRevolvedAreaSolid.cpp`,  `src/ifcgeom/mapping/IfcSectionedSolidHorizontal.cpp`,  `src/ifcgeom/mapping/IfcSectionedSurface.cpp`,  `src/ifcgeom/mapping/IfcSurfaceOfRevolution.cpp`,  `src/ifcgeom/mapping/IfcToroidalSurface.cpp`,  `src/ifcgeom/mapping/mapping.cpp`,  `src/ifcgeom/taxonomy.cpp`,  `src/ifcgeom/taxonomy.h`,  `src/ifcopenshell-python/ifcopenshell/__init__.py`,  `src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py`,  `src/ifcopenshell-python/ifcopenshell/express/header.py`,  `src/ifcopenshell-python/ifcopenshell/express/implementation.py`,  `src/ifcopenshell-python/ifcopenshell/express/mapping.py`,  `src/ifcopenshell-python/ifcopenshell/express/rule_compiler.py`,  `src/ifcopenshell-python/ifcopenshell/express/rule_executor.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC2X3.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X1.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X2.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X3.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X3_ADD1.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X3_ADD2.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X3_RC1.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X3_RC2.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X3_RC3.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X3_RC4.py`,  `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X3_TC1.py`,  `src/ifcopenshell-python/ifcopenshell/express/schema_class.py`,  `src/ifcopenshell-python/ifcopenshell/express/templates.py`,  `src/ifcopenshell-python/ifcopenshell/file.py`,  `src/ifcopenshell-python/ifcopenshell/geom/main.py`,  `src/ifcopenshell-python/ifcopenshell/sql.py`,  `src/ifcopenshell-python/ifcopenshell/stream.py`,  `src/ifcopenshell-python/ifcopenshell/util/element.py`,  `src/ifcopenshell-python/ifcopenshell/util/pset.py`,  `src/svgfill/src/arrange_polygons.cpp`,  `win/build-deps.cmd`
    -   Possible breaking commits:  `43533706a examples - build IfcOpenHouse/IfcAdvancedHouse using cmake package`,  `5a0671584 Update examples`,  `1e4693346 fix project name for IfcAdvancedHouse`,  `c78b2893d Proceed with merge`,  `a85571979 Merge branch 'v0.8.0' into v0.8.0-merge`

Moult:
I managed to get this to compile and import in Python locally mostly by letting AI work it out without any real understanding. Not sure if it's useful or not, but I've attached the findings.

[findings.md](https://github.com/user-attachments/files/26321941/findings.md)

Moult:
I experienced segfaults when doing ifcopenshell.open() (after all those compilation fixes in the previous comment) on the ISSUE_159_kleine_Wohnung_R22.ifc model (from the list of [profiling models](https://github.com/Moult/profiling/)). Here's an updated AI generated findings.md with an explanation of that fix, and a patch showing all the changes so far.

I'm also getting iterator initialisation errors on the advanced_model.ifc model saying "Trying to access deleted instance reference" (with detailed logging turned on). Haven't yet found a solution to that yet though.

Other than that, with those patches I can compile, and the profiling code is able to basically recreate the same numbers (apart from that initialisation error on that model) as the main branch.

[ai.patch](https://github.com/user-attachments/files/26322496/ai.patch)
[findings.md](https://github.com/user-attachments/files/26322497/findings.md)

aothms:
Hm yes apparently openai codex only bothered to verify compilation on windows. These template/typename things are notoriously different between compiler implementations.

Moult:
Woke up this morning and the "Trying to access deleted instance reference" was solved by AI with this explanation:

[findings.md](https://github.com/user-attachments/files/26326124/findings.md)
[ai.patch](https://github.com/user-attachments/files/26326125/ai.patch)

aothms:
Thanks! that fix we probably cannot use, because that means that every deleted instance equals every other deleted instance. We want this generally to bubble up as an exception (I think). But the root cause is super useful we just need to catch this a bit earlier, because apparently this is not a very common issue so I don't think it warrants a fix so high up in the abstraction level.

I have a bit of a backlog of other work because I missed a good couple of days due to back issues, but I'll get to this later this week. If you want to have it chew at some other tasks in the meantime, essentially:

- `ifcgeom/mapping/mapping.cpp`: `taxonomy::ptr mapping::map(const express::Base& inst)`
  
  Don't loop over all mappable geometry types for every converted instance, but rather "compile" based on IfcParse::declaration* into a function table in the mapping constructor so that we can dispatch to the appropriate call in sub linear time. Take care of subtypes: entity subtypes should be handled by the most closely defined handler in the inheritance tree. This for example applies to profiles where some subtypes are handled by their parent profile handler and some have a distinct handler for a specific subtype.

- `src/ifcgeom/AbstractKernel.h`: `dispatch_conversion`.

  Same idea, but maybe more complex. Here we do not have a `#include` based loop, but a recursive template-based one. I always expected the compiler to optimize this into some sort of jump table, but it doesn't seem it happens.

- Add a third "PassThrough" geometry kernel that only accepts `taxonomy::shell` with faces with at most 4 linear edges (so that triangulation is trivial) or extrusions of polygonal bases without openings. That writes the taxonomy::shell essentially directly to the `IfcGeom::Representation::Triangulation`, tie this in too hybrid kernel so that we can call something like `hybrid-passthrough-cgal-simple-opencascade`. Make the shape and structure similar to the occt and cgal kernel.

- Create a third paramatrization of the cgal geometry lib. Now we have `cgal` and `cgal-simple` which corresponds to Exact_predicates_exact_construction (necessary for Nef) Exact_predicates_inexact_construction (prevents some crashes). But actually the idea was that cgal-simple would correspond to Simple_cartesian&lt;double> which is much lighter, but less robust. If we at least have access to all three parametrizations we can better understand the failure modes the real cgal-simple and maybe start using it again.

- (low prio) for cgal non-simple implement an alternative boolean logic routine not based on Nef_3, but on CGAL/Polygon_mesh_processing/corefinement

Moult:
Typo perhaps: https://github.com/IfcOpenShell/IfcOpenShell/blob/datamodel-v1.0/src/ifcparse/hierarchy_helper.cpp#L628

Moult:
Maybe bug?

● Update(src/ifcgeom/mapping/IfcOffsetCurveByDistance.cpp)                                                             
  ⎿  Added 1 line, removed 1 line                                                                                      
      158                                                                                                              
      159     // at this point, next == end and prev == end-1
      160      #if defined SCHEMA_HAS_IfcDistanceExpression                                                            
      161 -    double last_distance = (*prev)->DistanceAlong() * length_unit_;                                         
      161 +    double last_distance = (*prev).DistanceAlong() * length_unit_;                                          
      162      #else                                                                                                   
      163      double last_distance = (double) prev->DistanceAlong().as<IfcSchema::IfcLengthMeasure>() * length_       
           unit_;                                                                                                      
      164      #endif

Moult:
I've noticed some discrepancies between v0.8.0 and this branch in outputs from the OpenCASCADE kernel (for the moment, ignoring all the other new kernels). See file:

[bug1.ifc.txt](https://github.com/user-attachments/files/26644400/bug1.ifc.txt)

v0.8.0 OCC:

image

Datamodel OCC:

image

Moult:
The passthrough kernel seems to be a bit funky, here's the results on AC20-FZK-Haus.ifc for hybrid-passthrough-manifold-cgal-simple-opencascade:

image

For example, for STEP ID 37582:

```
============================================================
  ios_hybrid (hybrid-manifold-cgal-simple-opencascade)
============================================================
  Vertices: 8
  Edges:    0
  Faces:    12
  Bbox min: [-0.500000, 9.729474, 3.200000]
  Bbox max: [12.500000, 9.809474, 3.360000]
  Dims:     [13.000000, 0.080000, 0.160000]
  Centroid: [6.000000, 9.769474, 3.280000]

  Vertices (8):
    [   0]    12.500000     9.809474     3.200000
    [   1]    12.500000     9.809474     3.360000
    [   2]    12.500000     9.729474     3.200000
    [   3]    12.500000     9.729474     3.360000
    [   4]    -0.500000     9.809474     3.200000
    [   5]    -0.500000     9.809474     3.360000
    [   6]    -0.500000     9.729474     3.200000
    [   7]    -0.500000     9.729474     3.360000

============================================================
  ios_hybrid (hybrid-passthrough-manifold-cgal-simple-opencascade)
============================================================
  Vertices: 1
  Edges:    0
  Faces:    12
  Bbox min: [0.000000, 0.000000, 0.000000]
  Bbox max: [0.000000, 0.000000, 0.000000]
  Dims:     [0.000000, 0.000000, 0.000000]
  Centroid: [0.000000, 0.000000, 0.000000]

  Vertices (1):
    [   0]     0.000000     0.000000     0.000000
```

Moult:
BTW I have rebased the work on the viewer on the datamodel-v1.0 branch: https://github.com/IfcOpenShell/IfcOpenShell/pull/7930

Moult:
Another observation: I think the normals coming from manifold aren't right. At least the viewer gives me funky colours :)

aothms:
> I've noticed some discrepancies between v0.8.0 and this branch in outputs from the OpenCASCADE kernel (for the moment, ignoring all the other new kernels). See file:
> 
> [bug1.ifc.txt](https://github.com/user-attachments/files/26644400/bug1.ifc.txt)
> 
> v0.8.0 OCC:
> image
> 
> Datamodel OCC:
> image

@Moult just the stupidest of mistakes in copy pasting, which is in a weird way also reassuring https://github.com/IfcOpenShell/IfcOpenShell/pull/7534/commits/a44f72287bc34e68c3bcb723fea36108cd040363

aothms:
> The passthrough kernel seems to be a bit funky

Something so simple I had better built myself. Now that I better see the code I think it needs to be thrown away, but at least this particular issue is gone.



> I think the normals coming from manifold aren't right.

It appears it doesn't populate anything so you're likely reading random data. Will have a look tomorrow.

Moult:
For the pointer issue, should it be `(*prev).DistanceAlong()` instead? Like a similar scenario on L102?

Note I also have these two hacks applied:

 - https://github.com/IfcOpenShell/IfcOpenShell/pull/7930/changes#diff-10080d9c67c533fd4678dd3679ecee57a18662516ff24dc42375787d30af8feb
 - https://github.com/IfcOpenShell/IfcOpenShell/pull/7930/changes#diff-ac16bfd5fbdc756f13333ace016627d0268bdfab7bc243e15d8e2d8e7208ef71

Explanation in [findings.md](https://github.com/IfcOpenShell/IfcOpenShell/pull/7930/changes#diff-5be547acff72c324b51d036785c5be0a125cfbbfc0f307798ccd3faeadda3a66)

Moult:
Some Python workarounds whilst things are moving: https://github.com/IfcOpenShell/IfcOpenShell/pull/7930/changes/eea2398e07a5a0ca8ef02ede8578848b702e5b70

Also I have confirmed for RDB there seem to be indeterministic shape counts:

```python
import sys
# sys.path.insert(0, "/home/dion/Projects/ios-dm/src/ifcopenshell-python")
# sys.path.insert(0, "/home/dion/Projects/ios-dm/build/ifcwrap")
sys.path.insert(0, "./build/Linux/x86_64/install/ifcopenshell")
import multiprocessing

import ifcopenshell
import ifcopenshell.geom

f = ifcopenshell.open("/home/dion/Projects/ios-dm/models/a.rdb")
print(f"schema: {f.schema}")
print(f"projects: {len(f.by_type('IfcProject'))}")
print(f"contexts: {len(f.by_type('IfcGeometricRepresentationContext'))}")
print(f"reps: {len(f.by_type('IfcRepresentation'))}")

it = ifcopenshell.geom.iterator(settings, f, multiprocessing.cpu_count(), geometry_library="opencascade")
ok = it.initialize()
print(f"initialize (opencascade): {ok}")
if ok:
    n = 1
    while it.next():
        n += 1
    print(f"  shapes: {n}")
```

With cpu count 1 I always get 107 shapes. With all CPUs, I get different numbers (101, 103, etc). I guess the mutex thing had some truth to it?

Moult:
Is it just me, or is HDF5 not available via Python? Perhaps something that got lost during the rearchitecture?

```
  File "/home/dion/Projects/ios-dm/convert_hdf5_tri.py", line 24, in convert_file
    serializer = ifcopenshell.geom.serializers.hdf5(str(h5_path), settings, serializer_settings)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'serializers' has no attribute 'hdf5'
```

Moult:
Not touching Manifold, but dumping here two issues I found: normals and a segfault when doing duplex.

confirming that i see no normals in the viewer for manifold, ai reckons it's only implemented on meshes, not extrusions (which come out with numprop=3, and never calls addnormal). it also reckons the normals aren't rotated by the placement matrix. i didn't touch anything and haven't code dived there :)
https://pastebin.com/pDeVjJ80 for full ai output

> ● Reproduced cleanly. The crash is manifold::Manifold's copy constructor inside ManifoldShape::moved() → result_to_manifold() during convert_openings (cutting door/window openings, which Duplex has). Copying an invalid Manifold suggests
>   dereferencing an empty part.solid optional. Let me read moved() and result_to_manifold.
>           
>   Ran 1 shell command 
> 
> ● Found it. In ManifoldShape::moved() (ManifoldConversionResult.cpp:505), the loop always does emplace_back(mesh, *solid) — dereferencing solid even when it's std::nullopt. That happens for a part with no solid (a shell that converted
>   "mesh-only" — non-manifold geometry, which Duplex has). Dereferencing the empty optional yields a garbage Manifold, and copying it in the ManifoldPart constructor is the segfault (manifold::Manifold copy ctor, frame #0).
>                                                                                                                                                                                                                                                
>   Let me read the exact lines and fix it.                                                                                                                                                                                                      
>                                                                                                                                                                                                                                                
>   Read 1 file                                                                                                                                                                                                                                  
>                                                                                                                                                                                                                                                
> ● Update(src/ifcgeom/kernels/manifold/ManifoldConversionResult.cpp)                                                                                                                                                                            
>                                                                                                                                                                                                                                                
> ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
>  Edit file                                                                                                                                                                                                                                     
>  src/ifcgeom/kernels/manifold/ManifoldConversionResult.cpp                                                                                                                                                                                     
> ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌
>  507    moved_parts.reserve(parts_.size());                                                                                                                                                                                                    
>  508    for (const auto& part : parts_) {                                                                                                                                                                                                      
>  509      auto mesh = transform_mesh(part.mesh, *place);                                                                                                                                                                                       
>  510 -    auto solid = part.solid ? make_manifold(mesh) : std::nullopt;                                                                                                                                                                        
>  511 -    if (part.solid && !solid) {                                                                                                                                                                                                          
>  510 +    if (!part.solid) {                                                                                                                                                                                                                   
>  511 +      // Mesh-only part (e.g. a shell that failed to close into a solid).                                                                                                                                                                
>  512 +      // Keep it mesh-only — dereferencing the empty solid optional here                                                                                                                                                                 
>  513 +      // is UB and copies a garbage manifold::Manifold (segfault).                                                                                                                                                                       
>  514 +      moved_parts.emplace_back(mesh);                                                                                                                                                                                                    
>  515 +      continue;                                                                                                                                                                                                                          
>  516 +    }                                                                                                                                                                                                                                    
>  517 +    auto solid = make_manifold(mesh);                                                                                                                                                                                                    
>  518 +    if (!solid) {                                                                                                                                                                                                                        
>  519        throw std::runtime_error("Failed to transform shape");                                                                                                                                                                             
>  520      }                                                                                                                                                                                                                                    
>  521      moved_parts.emplace_back(mesh, *solid);                                                                                                                                                                                              
> ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌
>  Do you want to make this edit to ManifoldConversionResult.cpp?

brunopostle:
## `get_info(recursive=True)` segfaults (stack overflow) on entities whose attribute graph contains a cycle

Found while testing the `ifcviewer-wgpu` branch (commit `d1027c587`, 2026-08-05), which has this datamodel-v1.0 rewrite merged in. Reproduces directly against `entity_instance.py`/`IfcParseWrapper.i` as they stand in this PR, no downstream code involved.

## Environment

- Branch: `ifcviewer-wgpu` @ `d1027c587` (2026-08-05), which includes this PR's rewrite of `entity_instance.py` / `get_info_cpp`
- Built via Fedora COPR git-snapshot package, version string `0.8.6`
- Python: 3.14.6
- OS: Linux x86_64 (Fedora)

## Summary

`entity_instance.get_info(recursive=True)` (and therefore `entity_instance.__eq__`/`__ne__`, see also #<ISSUE2>) recurses into every entity-typed attribute with no cycle/visited-instance tracking. IFC schemas legitimately allow attribute cycles (e.g. `IfcRelSpaceBoundary2ndLevel.CorrespondingBoundary` pointing at another boundary that points back), so calling `get_info(recursive=True)` on such an instance recurses without bound until the C stack is exhausted and the process segfaults. There is no Python-catchable exception — the interpreter crashes.

## Minimal reproduction

```python
import ifcopenshell

f = ifcopenshell.file()
a = f.create_entity("IfcRelSpaceBoundary2ndLevel", GlobalId=ifcopenshell.guid.new(), Name="A")
b = f.create_entity("IfcRelSpaceBoundary2ndLevel", GlobalId=ifcopenshell.guid.new(), Name="B")
a.CorrespondingBoundary = b
b.CorrespondingBoundary = a

a.get_info(recursive=True)  # segfaults
```

```
$ python3 repro.py
Fatal Python error: Segmentation fault
...
Segmentation fault (core dumped)
```

## Root cause

`src/ifcwrap/IfcParseWrapper.i`, `get_info_cpp` / `convert_cpp_attribute_to_python` (around lines 1159-1258):

```cpp
} else if constexpr (std::is_same_v<U, express::Base>) {
    if (recursive) {
        return get_info_cpp(v, recursive, include_identifier);   // <-- no cycle guard
    } else {
        return pythonize(v);
    }
}
```

`get_info_cpp` calls `convert_cpp_attribute_to_python` for every attribute, which calls back into `get_info_cpp` for every entity-typed attribute value, with no set of already-visited instance ids/pointers to stop when a cycle is revisited. Because this is native recursion (not Python-level), Python's `sys.setrecursionlimit`/`RecursionError` machinery never gets a chance to intervene — the process crashes with SIGSEGV once the C stack is exhausted.

The same crash is reachable indirectly through `entity_instance.__eq__`/`__ne__`, since both fall back to `self.get_info(recursive=True, ...) == other.get_info(recursive=True, ...)` for entities it can't cheaply prove equal (see the separate comment below on this PR about when that fallback triggers unexpectedly often).

## Suggested fix

Track visited instance ids (or pointers) for the current `get_info` recursive call and either stop recursing into an already-visited instance (returning e.g. just its id/type) or raise a catchable Python exception instead of crashing.

## How this was found

Found while investigating a real-world segfault in a downstream project's ([homemaker](https://github.com/brunopostle/homemaker-addon)) test suite while dogfooding this branch, which does `element_boundary != boundary` while iterating `IfcRelSpaceBoundary2ndLevel` instances, several of which end up with `CorrespondingBoundary` cycles. Reduced to the minimal repro above with no downstream code involved.

brunopostle:
## `ifcopenshell.file` has no value equality, so `entity_instance.__eq__` treats same-file instances as belonging to different files

Found while testing the `ifcviewer-wgpu` branch (commit `d1027c587`, 2026-08-05), which has this datamodel-v1.0 rewrite merged in — same underlying rewrite as the segfault reported above, this is the reason that crash is reachable from an ordinary `a != b` between two same-file entities.

## Environment

- Branch: `ifcviewer-wgpu` @ `d1027c587` (2026-08-05), which includes this PR's rewrite of `entity_instance.py` / `file.py`
- Built via Fedora COPR git-snapshot package, version string `0.8.6`
- Python: 3.14.6
- OS: Linux x86_64 (Fedora)

## Summary

`entity_instance.__eq__` (`ifcopenshell/entity_instance.py:211-226`) has this branch:

```python
if self.file != other.file or not self.is_entity():
    return self.get_info(recursive=True, include_identifier=False) == other.get_info(
        recursive=True, include_identifier=False
    )

return False
```

The intent is clearly: only fall back to a deep structural comparison when the two instances come from *different* files (or aren't entities); two distinct entities from the *same* file that aren't identical should just be unequal (`return False`).

But `self.file` returns a new Python wrapper object on every access, and `file_mixin` (`ifcopenshell/file.py:486`) defines neither `__eq__` nor `__hash__`. So `self.file != other.file` uses default object-identity comparison and is **always `True`**, even when `self` and `other` are two entities in the very same `ifcopenshell.file`:

```python
import ifcopenshell
f = ifcopenshell.file()
w1 = f.create_entity("IfcWall")
w2 = f.create_entity("IfcWall")
print(w1.file is w2.file)   # False
print(w1.file == w2.file)   # False (should be True, same underlying file)
```

Consequence: **every** `==`/`!=` comparison between two non-identical entity instances, even from the same file, takes the expensive recursive `get_info(recursive=True)` path instead of the cheap `return False`. This is at minimum a performance footgun, and it is what makes the crash in the companion comment on this PR reachable from completely ordinary code — any `a != b` between two distinct same-file entities recurses through `get_info(recursive=True)` on both sides, and if either side's attribute graph has a cycle, the process segfaults instead of returning `False`.

## Minimal reproduction

```python
import ifcopenshell

f = ifcopenshell.file()
w1 = f.create_entity("IfcWall")
w2 = f.create_entity("IfcWall")

assert w1.file == w2.file          # fails: file objects from the same ifcopenshell.file compare unequal
assert (w1 != w2) == True          # happens to give the right answer here, but via the wrong/expensive path
```

Combined with the companion comment on this PR, this turns an innocuous `a != b` between two same-file entities into a crash:

```python
import ifcopenshell

f = ifcopenshell.file()
a = f.create_entity("IfcRelSpaceBoundary2ndLevel", GlobalId=ifcopenshell.guid.new(), Name="A")
b = f.create_entity("IfcRelSpaceBoundary2ndLevel", GlobalId=ifcopenshell.guid.new(), Name="B")
a.CorrespondingBoundary = b
b.CorrespondingBoundary = a
c = f.create_entity("IfcRelSpaceBoundary2ndLevel", GlobalId=ifcopenshell.guid.new(), Name="C")

a != c  # segfaults, even though a and c are both plain instances in the same file
```

## Suggested fix

Give `file_mixin` (or whatever wraps the underlying C++ file pointer) proper `__eq__`/`__hash__` based on the underlying file identity/pointer, so `self.file != other.file` correctly returns `False` for two instances belonging to the same file. That alone would make `entity_instance.__eq__` take the fast `return False` path for the vast majority of real-world same-file comparisons, and would avoid triggering the recursive-cycle crash reported in the companion comment for this common case (though that recursion issue should still be fixed independently, since genuinely cross-file or non-entity comparisons can still hit a cycle).

## How this was found

Found while investigating a segfault in a downstream project's ([homemaker](https://github.com/brunopostle/homemaker-addon)) test suite while dogfooding this branch; see the companion comment on this PR for the crash this contributes to.

brunopostle:
@aothms sorry about the two bot-written comments above, I'm travelling and doing this from my phone.

Though these do appear to be genuine bugs in the ifcviewer-wgpu branch.

Moult:
@brunopostle maybe https://github.com/IfcOpenShell/IfcOpenShell/pull/8139 is a better place to comment, as the datamodel branch is superseded by ifcviewer-wgpu. BTW I also did come across the file equality issue in a failed test case and put a commit in (but I am still very scared about these commits)

brunopostle:
@Moult yes, sorry travelling so needed to put it somewhere, I will need to move it when I'm back

brunopostle:
@moult @aothms the two issues reported above seem to be fixed in a latest v0.9.0 snapshot build

aothms:
The 2nd item is likely fixed, the change in equality behaviour we had to correct anyway for the express rule tests to pass. But I don't recall seeing anything for the first item. Also this is the first time that I actually see the opportunity for pure-forward cycles, I always assumed that we tried to prevent these in the schema, but apparently I was wrong. Can we create a minimal pytest test case around this (just the two boundaries file.from_string it and then call get_info()). I think maybe the problem just got hidden somehow because we fused get_info and get_info_2 (iirc).

brunopostle:
> The 2nd item is likely fixed, the change in equality behaviour we had to correct anyway for the express rule tests to pass. But I don't recall seeing anything for the first item. Also this is the first time that I actually see the opportunity for pure-forward cycles, I always assumed that we tried to prevent these in the schema, but apparently I was wrong. Can we create a minimal pytest test case around this (just the two boundaries file.from_string it and then call get_info()). I think maybe the problem just got hidden somehow because we fused get_info and get_info_2 (iirc).

Yes, the second one seems to be fixed but the first one still segfaults, I've created an issue for this and transferred the bug report: #9334

Issue #7545, opened 2026-01-09
possible to get the search mechanism back, in the editing text dialog?


theoryshaw:
### Feature Description

@falken10vdl would it be possible to get the search mechanism back, in the editing text dialog?
I used that quite often to search for often used selection syntaxes. 

https://github.com/user-attachments/assets/8a3993c2-8c38-4fb2-a109-4325a59eff82

falken10vdl:
@theoryshaw  I have created PR https://github.com/IfcOpenShell/IfcOpenShell/pull/7547 to address it.
Can you please take a look?

https://github.com/user-attachments/assets/b12f034a-30e8-4754-9940-54c00616b42d

Thanks!

theoryshaw:
thank you!

PR #7547, opened 2026-01-09
Use helper function for search suggestions in literal property drawing

falken10vdl:
This is to address https://github.com/IfcOpenShell/IfcOpenShell/issues/7545

Cheers!


theoryshaw:
Awesome, thanks!

PR #7550, opened 2026-01-12
Fix aggregate duplication to include parts and preserve hierarchy

theoryshaw:
- Auto-include all parts when duplicating aggregates
- Preserve nested aggregate relationships during duplication
- Select all duplicated objects for immediate moving

### DIFF
--- a/src/bonsai/bonsai/bim/module/geometry/operator.py
+++ b/src/bonsai/bonsai/bim/module/geometry/operator.py
@@ -1205,13 +1205,84 @@ class OverrideDuplicateMove(bpy.types.Operator):
         for obj in objects_to_remove:
             tool.Blender.deselect_object(obj)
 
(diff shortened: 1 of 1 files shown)


Moult:
See 0b1c049e81423e45cf93abcc1bea7511ee4cf154 where basically duplication should be centralised into a self contained testable tool function. I came across this today when trying to reuse `tool.Geometry.duplicate_ifc_objects` and because this code is outside the tool, there will be inconsistencies between using the operator and other instances calling just the tool. As it stands there is currently already an inconsistency when the tool is called for duplicating drawing annotations, but hopefully nobody has noticed that bug yet :)

Can you please submit a bug about the original behaviour this was trying to solve and reference this PR? I can try to refactor this and hopefully add some tests.

theoryshaw:
sure, here you go: #7605

PR #7551, opened 2026-01-12
Ifc5d quantity formula

maxfb87:


maxfb87:
I think it can be merged now so feedback is welcome

maxfb87:
@Moult could you take a look at this PR? Just for understand if the structure is ok or if it can be implemented in a better way..thanks :-)

Moult:
Very cool, I only had a brief glance so far, but can't util.selector.format be reused? It already has operators and other things built into it (e.g. rounding).

maxfb87:
In order to add the possibility to format the result (for example to round the resulting value)?
Yeah, maybe it can be done, i'll take a look.

I was thinking if the api is the right place to store the formula "parser" or there is a better place to store it (something like util). Any hint on that?

Moult:
I mean that the util.selector.format() function already can handle fetching of variables (e.g. quantities) and arithmetic operators. So you can reuse that instead of recreating it with AST node stuff.

maxfb87:
Ah ok, thanks. It would be good to use that function.
Do you know if there is a sort of preprocessing for the variables?
Because i would like that the user could use this kind of string "2*Pset_ConcreteElementGeneral.ReinforcementVolumeRatio" instead of "2*{{Pset_ConcreteElementGeneral.ReinforcementVolumeRatio}}" (so without the double {{ }} )
and also i would like to use "2*Height" instead of "2*{{Qto_WallBaseQuantities.Height}}".
Any hint about that?

maxfb87:
I'm going to merge this because i don't want to get stuck too much on this

Issue #7555, opened 2026-01-13
Text too large in macOS Blender


jlongster:
### Bug Description

Blender: 4.2 and 5.0 (tested in both)
Bonsai: 0.8.4 
macOS: 15.3

(This was also reported by another user yesterday in the matrix channel)

Not sure what the right name for this text is, but when adding walls/measuring/etc this "annotation" text is way too big in Blender in macOS:

<img width="1158" height="818" alt="Image" src="https://github.com/user-attachments/assets/6060590c-1399-49e9-a16d-cf01f5b13278" />

All of the other text is fine in Blender, it's only this addon-specific text that is broken. I would guess this is a high-res display issue, and it needs to take the DPI of the screen into account (it probably should be rendering half this size)

### Debug and Error Output

```true

```

falken10vdl:
@jlongster I have create PR https://github.com/IfcOpenShell/IfcOpenShell/pull/7557 to address this issue.
Please note that I managed to get hold of an old intel based Mac and could reproduce the problem above (not visible in Linux nor Windows).
Now the size of the decorations are based on blenders widgets but still I had to have a little different factors in MacOS compared to Linux/Windows to make them appear better.

Cheers!

BIMvoice:
Status check: does this still reproduce on a current daily build? macOS UI scaling for viewport text has had several rounds of DPI fixes since January; if it persists, your display's scale factor (Retina vs external) is the datapoint that pins it.

falken10vdl:
In my opinion this is solved. 
Cheers!

BIMvoice:
@falken10vdl recommending this one for closing.

Your PR #7557 merged on 25 February and you confirmed on 6 July that you consider it solved. The issue is still open.

@jlongster if the oversized text is still there for you on a current build, say so and this stays open.

PR #7557, opened 2026-01-14
Scale-font-in-PolylineDecorator

falken10vdl:
Addresses https://github.com/IfcOpenShell/IfcOpenShell/issues/7555
Makes the text decorations for polyline and boundingbox decorations to follow blender's ui_style.widget.points and view.ui_scale.

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/decorator.py
+++ b/src/bonsai/bonsai/bim/module/model/decorator.py
@@ -29,7 +29,6 @@ import bpy
 import gpu
 import ifcopenshell
 import mathutils
-import sys
 from bpy.types import SpaceView3D
 from bpy_extras import view3d_utils
 from bpy_extras.view3d_utils import location_3d_to_region_2d
@@ -467,14 +466,9 @@ class PolylineDecorator:
 
         self.addon_prefs = tool.Blender.get_addon_preferences()
         self.font_id = 0
-        ui_style = context.preferences.ui_styles[0]
-        widget_font_points = ui_style.widget.points
-        ui_scale = context.preferences.view.ui_scale
-        platform_scale = 2 if sys.platform == "darwin" else 1
-
-        font_size = widget_font_points * ui_scale * platform_scale
-        offset = widget_font_points * ui_scale * (1.5 * platform_scale)
-        line_height = widget_font_points * ui_scale * (1.25 * platform_scale)
+        font_size = tool.Blender.scale_font_size(None)
+        offset = tool.Blender.scale_font_size(None) * 1.5
+        line_height = tool.Blender.scale_font_size(None) * 1.25
         blf.size(self.font_id, font_size)
         blf.enable(self.font_id, blf.SHADOW)
         blf.shadow(self.font_id, 6, 0, 0, 0, 1)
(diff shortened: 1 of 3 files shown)


jlongster:
Awesome! I don't have time to test a custom build, but would love to get this merged and I since you can repro I bet it fixes it :)

falken10vdl:
@brunoperdigao are you happy with this PR?
It should fix the problem seen in MacOS. It looks not many developers are using that OS?
image

Thanks!

brunoperdigao:
@falken10vdl Great work! My only suggestions is that you put all the logic inside `tool.Blender.scale_font_size()` instead of repeating the code

falken10vdl:
@brunoperdigao Please take a look to the code, following your suggestion. Thanks! much cleaner.

Below some pcitures of the decorators in different platforms

- MacOS
<p align="center">
Screenshot 2026-01-22 at 10 48 01<br>
Screenshot 2026-01-22 at 10 52 52<br>
Screenshot 2026-01-22 at 10 51 30<br>
Screenshot 2026-01-22 at 10 52 16<br>
</p>



- Linux OS:
<p align="center">
Screenshot 2026-01-22 at 10 48 01<br>
Screenshot 2026-01-22 at 10 52 52<br>
Screenshot 2026-01-22 at 10 51 30<br>
Screenshot 2026-01-22 at 10 52 16<br>
</p>



- Windows OS:
<p align="center">
Screenshot 2026-01-22 120936<br>
Screenshot 2026-01-22 121010<br>
Screenshot 2026-01-22 121037<br>
Screenshot 2026-01-22 121101<br>
</p>

Cheers!

brunoperdigao:
@falken10vdl, something that I noticed is that the current code is not only responsive to the blender resolution scale but also to the OS screen scale, so that if I change this in my OS, the polyline UI will change too. Can you try to keep this behavior?

falken10vdl:
@brunoperdigao Very good catch! In fact, I think that it is best to use your calculations since they cope for all cases. What I have done is to just add the default value in line with the size of the blender menus and also added a addon-setting for people to change the relative size since some people might want it bigger and some other smaller.
Please take a look. What do you think? Thanks a lot for your feedback.

https://github.com/user-attachments/assets/33967269-0269-4191-b9fd-0bc97446cac6

Cheers!

jlongster:
I'm not familiar with how any of this works, or even how native Blender UI handles text, but I would expect this text to act exactly like native Blender UI. So however that works, in terms of how I change the native Blender font size too (or if it respects the native font size or not) I think Bonsai should follow it. This will make it feel more native to Blender. I seems odd to have a separate setting to control the font size for _just that text_ instead of increasing the global Blender font size larger or smaller.

Just my opinion though, you all are deep in the weeds here so do whatever you feel is best, and whatever helps this merge sooner :)

falken10vdl:
@brunoperdigao are you happy with the PR?
Thanks!

brunoperdigao:
@falken10vdl Sorry it is taking too long, but I'd like to ask just one more thing. You can set the `size` default value as `None` like this `def scale_font_size(cls, size=None):`. That way you don't have to pass `None` as an argument every time you call the function. You can just call `tool.Blender.scale_font_size()`

falken10vdl:
@brunoperdigao Thanks! Much cleaner.
Just done it.

Cheers!

PR #7570, opened 2026-01-17
IFC link management with UUID support. Same file can be linked multiple times

falken10vdl:
A proposal to address https://community.osarch.org/discussion/3302/reflecting-ifcgroups-in-the-outliner-hierarchy#latest

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -49,6 +49,7 @@ import bonsai.bim.module.drawing.svgwriter as svgwriter
 import bonsai.bim.module.drawing.annotation as annotation
 import bonsai.bim.module.drawing.sheeter as sheeter
 import bonsai.bim.export_ifc
+import math
 from bpy_extras.io_utils import ImportHelper
 from bonsai.bim.module.drawing.decoration import CutDecorator
 from bonsai.bim.module.drawing.data import DecoratorData, ElementValuesData
@@ -57,7 +58,7 @@ from bonsai.bim.prop import StrProperty
 from typing import NamedTuple, Union, Optional, Literal, TYPE_CHECKING, Any, TypedDict, get_args
 from lxml import etree
 from math import radians
-from mathutils import Vector, Color, Matrix
+from mathutils import Vector, Color, Matrix, Quaternion
 from timeit import default_timer as timer
 from bonsai.bim.module.drawing.prop import RasterStyleProperty, RASTER_STYLE_PROPERTIES_EXCLUDE
 from bonsai.bim.ifc import IfcStore
@@ -255,6 +256,7 @@ class CreateDrawing(bpy.types.Operator):
 
     drawing_name: str
     is_manifold_cache: dict[str, bool]
+    link_transformations: dict[str, tuple]  # Maps link UUID to (ifc_file, link_matrix)
 
     @classmethod
     def poll(cls, context):
@@ -588,6 +590,8 @@ class CreateDrawing(bpy.types.Operator):
         context_type: Literal["body", "annotation"],
         drawing_elements: set[ifcopenshell.entity_instance],
         target_view: str,
+        link_matrix=None,
+        link_id=None,
     ) -> None:
         drawing_elements = drawing_elements.copy()
         contexts_: list[list[int]] = getattr(contexts, context_type)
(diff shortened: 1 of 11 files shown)


Moult:
This is great, some comments:

 1. Please remove `keep_same_ifc_file_link_as_independent`. It should not be necessary (see 3).
 2. Storing and syncing the position through custom IfcProject psets and link handle `matrix_world` I think is not ideal, as it introduces custom psets and encourages eyeball positioning of linked models. Although eyeball movement is cool and might be sufficient for @theoryshaw, we should first support a precise, georeferencing aware positioning (see 3).
 3. The definition of Identification is "a system interpretable position within the document". You've used a UUID for uniqueness, which is not particularly semantic nor does it represent a position within the document. When linking is done, there are options to choose false origin modes (auto, manual, disabled) and custom false origins. Can I propose instead storing those in the Identification field (e.g. `AUTO`, `DISABLED` or `MANUAL,12345.6,132.5,123.4,45.0`.) Using this as the key means users cannot add the same model in the same spot by definition, these origins are now stored in IFC (instead of just in the linked blend).

Can you please also keep the code about element filters sparately? I also haven't yet had the time to catch up with the new set chaining filter features sorry.

falken10vdl:
@Moult Thanks for the feedback!

1. Done
2. Perfect. Let me remove the Pset. The only thing is that it is not only position but "state" of the linked projects that is stored (if hidden, wireframe, enabled...)
3. Thanks for the explanation! It makes much more sense and it is much more meaningful. I have removed uuid and use "identification" based on the location and angle and not allow to position in the same place the same project twice. However I do not see where this information is stored. I see save_linked_models_to_ifc that saves in an IfcReference the filename but I do not see where the values AUTOMATIC or MANUAL,12345.6,132.5,123.4,45.0 can be stored in the IFC. Actually if I load the link with the cache disabled the position is lost so it looks that such information is not stored in the IFC. Am I right? This linked project position would make sense to be stored in "Identification" of the IfcDocumentReference?
4. I have removed the element filtering code to another branch.

Thanks!
PS: To retain the "state" of the linked projects what do you think would make sense? I guess that is a candidate to store in the blend.metadata and not in IFC?

Moult:
Good point about visual state, that is not something I can see stored in IFC. Maybe as a document pset (as opposed to a project pset).

The false origin mode and numbers are stored in Blender since it is a Blender / Bonsai specific origin and doesn't change the original georeferencing data of the linked IFC. That's why if you remove the cache, the effect of those numbers is "lost".

falken10vdl:
falken10vdl@monster:~/bonsaiDevel/IfcOpenShell/src/ifcopenshell-python$ python -m pytest test/util/test_geolocation.py 
Blender 5.0.1 (hash a3db93c5b259 built 2025-12-16 01:30:59)
================================================================================== test session starts ===================================================================================
platform linux -- Python 3.11.13, pytest-9.0.2, pluggy-1.6.0
rootdir: /home/falken10vdl/bonsaiDevel/IfcOpenShell/src/ifcopenshell-python
configfile: pyproject.toml
plugins: bdd-8.1.0
collected 28 items                                                                                                                                                                       

test/util/test_geolocation.py ............................                                                                                                                         [100%]

=================================================================================== 28 passed in 0.05s ===================================================================================
falken10vdl@monster:~/bonsaiDevel/IfcOpenShell/src/ifcopenshell-python$ cd /home/falken10vdl/bonsaiDevel/IfcOpenShell/src/bonsai
falken10vdl@monster:~/bonsaiDevel/IfcOpenShell/src/bonsai$  pytest -p no:pytest-blender test/core/test_georeference.py 
================================================================================== test session starts ===================================================================================
platform linux -- Python 3.11.13, pytest-8.4.1, pluggy-1.6.0
rootdir: /home/falken10vdl/bonsaiDevel/IfcOpenShell/src/bonsai
configfile: pytest.ini
plugins: blender-3.0.7, bdd-8.1.0
collected 14 items                                                                                                                                                                       

test/core/test_georeference.py ..............                                                                                                                                      [100%]

=================================================================================== 14 passed in 0.02s ===================================================================================
falken10vdl@monster:~/bonsaiDevel/IfcOpenShell/src/bonsai$  pytest -p no:pytest-blender test/core/test_project.py 
================================================================================== test session starts ===================================================================================
platform linux -- Python 3.11.13, pytest-8.4.1, pluggy-1.6.0
rootdir: /home/falken10vdl/bonsaiDevel/IfcOpenShell/src/bonsai
configfile: pytest.ini
plugins: blender-3.0.7, bdd-8.1.0
collected 5 items                                                                                                                                                                        

test/core/test_project.py .....                                                                                                                                                    [100%]

=================================================================================== 5 passed in 0.02s ====================================================================================
falken10vdl@monster:~/bonsaiDevel/IfcOpenShell/src/bonsai$  python -m pytest test/tool/test_georeference.py 
Blender 5.0.1 (hash a3db93c5b259 built 2025-12-16 01:30:59)
================================================================================== test session starts ===================================================================================
platform linux -- Python 3.11.13, pytest-9.0.2, pluggy-1.6.0
rootdir: /home/falken10vdl/bonsaiDevel/IfcOpenShell/src/bonsai
configfile: pytest.ini
plugins: blender-3.0.7, bdd-8.1.0
collected 25 items                                                                                                                                                                       

test/tool/test_georeference.py .........................                                                                                                                           [100%]

=================================================================================== 25 passed in 0.85s ===================================================================================
falken10vdl@monster:~/bonsaiDevel/IfcOpenShell/src/bonsai$  python -m pytest test/tool/test_project.py 
Blender 5.0.1 (hash a3db93c5b259 built 2025-12-16 01:30:59)
================================================================================== test session starts ===================================================================================
platform linux -- Python 3.11.13, pytest-9.0.2, pluggy-1.6.0
rootdir: /home/falken10vdl/bonsaiDevel/IfcOpenShell/src/bonsai
configfile: pytest.ini
plugins: blender-3.0.7, bdd-8.1.0
collected 31 items                                                                                                                                                                       

test/tool/test_project.py ...............................                                                                                                                          [100%]

=================================================================================== 31 passed in 1.77s ===================================================================================
falken10vdl@monster:~/bonsaiDevel/IfcOpenShell/src/bonsai$

Moult:
Super cool to see all tests passing!

I think the first time you link something and it creates the cache:

 1. The cache stores its own coordinate overrides (like it already does)
 2. The reference Identification also stores the coordinate overrides. These should match.

If you link / relink it again with different settings without purging the cache:

 1. The reference Identification stores the new coordinate overrides
 2. The Blender placing code sees that this doesn't match what's in the cache, and positions it accordingly, but doesn't recreate the cache

I think this way we maximise cache reuse whilst ensuring that the cache itself has an explicitly selected user precision tradeoff.

falken10vdl:
I was doing the placement if it was moved after creating the initial placement using the metadata empties stored.
You can see that the cache filename/collection name/Identification in IfcDocRef have the creation location. Then if we move them the empties will be updated and that will be saved in the metadata file. When loading the links (if metadata is enabled, the empties will be loaded) it checks if there are already empties named as the ones trying to load and if they match, then it uses the already available emtpies to position the linked project.
image

So in order to make it work regardless of using the metadata file, we should keep the actual position also in the reference Identification? Something like this?
image
Thanks!

theoryshaw:
> So in order to make it work regardless of using the metadata file, we should keep the actual position also in the reference Identification? Something like this?

Would it be better to store this in a more IFC way--via something like IfcLocalPlacement, instead?

Moult:
Is this PR superseded by #7607 ?

@theoryshaw the idea of storing it as an IfcLocalPlacement is something I considered. E.g. decoupling the linked model (a document reference) from the placement of a linked model (an annotation). The problem with this logic is that the only mathematically significant thing as "the placement" of a linked model is the placement of the linked model's IFC local origin. And this does _not_ seem very friendly given that this local origin may be super far away (e.g. if they are working in map coordinates directly).

As a result, my conclusion is that it is perhaps better to just say there is a linked model, and there is only one correct way two models may be overlaid with global coordinates, and you may store an arbitrary translation/rotation offset.

falken10vdl:
@Moult Yes it is superseded by https://github.com/IfcOpenShell/IfcOpenShell/pull/7607 I changed it to draft but will be removed once 7607 is merged. There is some logic in this one for handling the SurfaceStyles and for importing IfcGroups as child collections as well as it was initially asked by @theoryshaw 
Cheers!

PR #7576, opened 2026-01-17
Enhance filter value suggestions to support enumerated properties in search

falken10vdl:
Allows to populate the suggestions for search with ENUM values. Currently they are not parsed correctly and no suggestions are provided for them.

### DIFF
--- a/src/bonsai/bonsai/bim/module/search/operator.py
+++ b/src/bonsai/bonsai/bim/module/search/operator.py
@@ -445,20 +445,17 @@ class FilterValueSuggestions(Operator):
                             if pset.HasProperties:
                                 for prop in pset.HasProperties:
                                     if hasattr(prop, "Name") and prop.Name == property_name:
-                                        if hasattr(prop, "NominalValue") and prop.NominalValue:
-                                            try:
-                                                value = prop.NominalValue.wrappedValue
-                                                if value is not None and value != "":
-                                                    if not hasattr(value, "is_a") and not isinstance(
-                                                        value, (tuple, list)
-                                                    ):
-                                                        str_value = str(value)
-                                                        if not str_value.startswith("#") and not str_value.startswith(
-                                                            "("
-                                                        ):
-                                                            property_values.add(str_value)
-                                            except:
-                                                continue
+                                        if prop.is_a("IfcPropertyEnumeratedValue"):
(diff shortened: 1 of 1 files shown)


src/bonsai/bonsai/bim/module/search/operator.py
```diff
@@ -445,20 +445,17 @@ def get_property_values(self, ifc_file, pset_name, property_name):
...
-                                                    ):
-                                                        str_value = str(value)
-                                                        if not str_value.startswith("#") and not str_value.startswith(
-                                                            "("
-                                                        ):
-                                                            property_values.add(str_value)
-                                            except:
-                                                continue
+                                        if prop.is_a("IfcPropertyEnumeratedValue"):
+                                            if hasattr(prop, "EnumerationReference") and prop.EnumerationReference:
```
Moult, line 449: Why hasattr? Similarly `for prop in pset.HasProperties or []`. This'll help heavily shrink down the nesting.

Moult: Also why is the whole thing wrapped in a try except block? What is that meant to catch? In general it's good to avoid large try blocks especially without explicit exception classes as it can mask issues.

falken10vdl: Just defensive against errors in the file but sincerely I do not know enough about what errors ifcopenshell would pass and we need to handle. But I agree that what you are suggesting `for prop in pset.HasProperties or []` is better coding and probably runs faster 👍 
try/except to avoid showing to end user errors for a function that even if it fails it should not be worrying. But thanks, since bonsai is marked as alpha (to me it is much more stable than alpha :) ) I guess unless really justified, no try/except?
Thanks for the feedback. It helps me.

Moult:
Cheers, the `hasattrs` are not required as the attributes are guaranteed from the schema. In general I actively try not to support invalid IFC data (except during loading) because it masks other bugs (either from Bonsai itself or other vendors).

falken10vdl:
Ah perfect. Thanks!
Regarding try/except, the strategy is not to use it in principle unless it comes from very known errors? That is the except catchig evertthing must be avoided even if it is related to code not  esential or comestic, right?

Moult:
Yeah, try/except masks errors in general, so minimise the size of the try block, and be as specific as possible for the exception. Admittedly it is idiomatic Python sometimes to be flexible around this and we're all human :)

PR #7584, opened 2026-01-18
Implement default handling for the 'IsVentilated' property when null.

iagobaapellaniz:
Now the dropdown selector for "IsVentilated" shows "FALSE" since "IsVentilated" is an optional field of "IfcMaterialLayer" which defaults to "FALSE" (https://standards.buildingsmart.org/IFC/DEV/IFC4_2/FINAL/HTML/schema/ifcmaterialresource/lexical/ifcmateriallayer.htm). 

If someone wants to make it concrete it can still click the radio button to toggle "Is Null".

image

### DIFF
--- a/src/bonsai/bonsai/bim/helper.py
+++ b/src/bonsai/bonsai/bim/helper.py
@@ -240,6 +240,8 @@ def import_attribute(
         attribute_type = attribute.type_of_attribute()
         is_logical = str(attribute_type) == "<type IfcLogical: <logical>>"
         enum_value = data[new.name]
+        if new.name == "IsVentilated" and enum_value is None:
+            enum_value = False
         if is_logical:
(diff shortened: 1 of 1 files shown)


Moult:
This seems to just visually change the default selection when null to show False instead of True, but does not actually change any logic. This is a good improvement for usability since otherwise users won't know this without reading the docs. I'm not opposed to this, since the default for this property is indeed false if null. However, the code shouldn't belong in the helper, since that helper is generic and addresses every single attribute in IFC. The proper way to implement this would be by supplying a callback handler instead.

iagobaapellaniz:
Nice! Thanks for the feedback. I was thinking on the usability indeed. I can keep trying and learning. Any advise is very welcome.

PR #7598, opened 2026-01-22
fix: Have kernels targets built as library

### LINKED ISSUE #7560: Recover `geometry_kernel_opencascade` as a real library to link manually
### Bug Description

My project currently uses **ifcopenshell 0.8.3**, and I am attempting to upgrade the library.

In my existing setup, I link against `geometry_kernel_opencascade` in order to use `IfcGeom::Iterator`:

```cpp
IfcGeom::Iterator geomIter(settings, &ifcFile, filter_funcs, numThreads);
````

However, in version **0.8.5**, the API has changed and now requires the following approach:

```cpp
#include <ifcgeom/kernels/opencascade/OpenCascadeKernel.h>
...
std::unique_ptr<ifcopenshell::geometry::kernels::AbstractKernel> kernel(
    new IfcGeom::OpenCascadeKernel(settings));
IfcGeom::Iterator geomIter(std::move(kernel), settings, &ifcFile, filter_funcs, numThreads);
```

While investigating this change, I discovered that the commit
9d2862815741c775f2fe7891b81283badac83484 changed `geometry_kernel_opencascade` from a normal library into an **OBJECT library**.

As a result, I can no longer link against `geometry_kernel_opencascade`, which breaks my existing build setup. This change makes it impossible to use the kernel as a standalone linkable target.

Could you please advise on how to restore compatibility or otherwise recover the ability to link against `geometry_kernel_opencascade`?
From my perspective, it is unclear what benefit is gained by converting this target into an object library.

Thank you for your help.

**Why not use 0.8.4?**: I am using `MINIMAL_BUILD`, but I got problem with `GLTF_SUPPORT`. The fix for this problem is in 0.8.5: 7e48d712a20d94b2e97f0cbeb2bd85fb46804bb4

### Attachments

_No response_

### Debug and Error Output

```true

```

EstebanDugueperoux2:
Fix #7560


aothms:
Thanks @EstebanDugueperoux2 really much appreciated.

Issue #7586, opened 2026-01-19
bad assertion in add_mesh_representation


OrionSehn:
### Bug Description

I think this assertion is wrong, and the number of points is not the same as the number of faces in most meshes

https://github.com/IfcOpenShell/IfcOpenShell/blob/4c1b3221ce1416a23c048332bd47a706d53a58a2/src/ifcopenshell-python/ifcopenshell/api/geometry/add_mesh_representation.py#L72

I think this needs to be removed, should be a simple 1 line fix.

### Debug and Error Output

```true

```

Moult:
I think the assertion is correct. It's a little bit obscure but vertices and faces are _not_ a list of verts / faces directly, first it is a list of items, and then each item is vertices / faces. I.e. it is a nested list. See https://github.com/IfcOpenShell/IfcOpenShell/blame/4c1b3221ce1416a23c048332bd47a706d53a58a2/src/ifcopenshell-python/ifcopenshell/api/geometry/add_mesh_representation.py#L45-L46

aothms:
Yeah I also bumped into this in the past. I'm usually a fan of expressing this in the types: instead of `list[list[tuple[int]]], list[list[tuple[float]]]` better have `list[items]` where items is a dataclass of verts list[tuple[float, float, float]] + faces list[tuple[int, int, int]] with optional edges.

OrionSehn:
Thanks so much for the prompt responses. 💕

There was documentation on this in the geometry creation docs, which I missed on the first pass. (really easy to glance at two square brackets and assume there's just one)

Maybe we can make the simple example have an extra newline character to make this less missable for others at a glance? (and maybe remove one of the faces, so its clear its not the number of verts and faces which need to be the same)

Minor readability fix, will open a pr.  

```python
# These vertices and faces represent a 2m square 1m high pyramid in SI units.
# Note how they are nested lists. Each nested list represents a "mesh". There may be multiple meshes.
vertices = [
    [(0.,0.,0.), (0.,2.,0.), (2.,2.,0.), (2.,0.,0.), (1.,1.,1.)]
]
faces = [
    [(0,1,2,3), (0,4,1), (1,4,2), (2,4,3)]
]
representation = ifcopenshell.api.geometry.add_mesh_representation(model, context=body, vertices=vertices, faces=faces)
```
https://docs.ifcopenshell.org/ifcopenshell-python/geometry_creation.html#mesh-representations

PR #7600, opened 2026-01-22
Adjustment to geometry generation python example

OrionSehn:
Changed example to be simpler/easier to read at a glance. 

See #7586


OrionSehn:
oops made a bad merge in my fork made a new pr #7601

PR #7601, opened 2026-01-22
simplified python geometry creation mesh rep example

OrionSehn:
Adjustment to geometry generation python example


OrionSehn:
Changed example to be simpler/easier to read at a glance.

See https://github.com/IfcOpenShell/IfcOpenShell/issues/7586

Moult:
Why are one of the faces missing?

OrionSehn:
> Why are one of the faces missing?

Whoops, meant to delete two of the faces instead of one in the example. Good catch, thx

Moult:
I'm not entirely sure I agree with this, sorry. It's just less numbers but not much more. The reason I used this as an example was: 1) I wanted to show you could have both quads and tris and 2) I wanted to show that manifold meshes were possible and recommended (necessary for boolean cuts later on like plan views).

OrionSehn:
no problem

PR #7607, opened 2026-01-25
Linked IFC projects enhancement (multiple links to same project file)

falken10vdl:
- Implement link management system using UUIDs as identifiers to support multiple links to the same IFC file
- Add georeferencing compatibility detection and UI display (NONE, NOT_COMPATIBLE, PARTIAL_COMPATIBLE, FULL_COMPATIBLE)
- Support for duplicate link creation with Shift+D shortcut and automatic position offset
- Add false origin and project north calculation from 3D cursor for MANUAL mode
- Only store one cache per file, regardless of the amount of links
- Prevent duplicate links based on filepath and position comparison
- Improve error handling for missing files and loading failures
- Update tests


falken10vdl:
@Moult 

> > I try to avoid duplicate UI (especially for one that can be as
> sophisticated as georef - e.g. missing is WCS) as it means double the
> code, double the tests, potential user confusion. 

ok

> BTW the note about
> vertical datum isn't quite accurate as it may be included in the CRS
> definition so vertical datum is optional.

Is there then a way to know in the CRS Name field if the z coordinate is right in the file? I was tagging as PARTIAL_COMPATIBLE if there was not a match between parent and linked Vertical Datum data.
Thanks!

Moult:
> Is there then a way to know in the CRS Name field if the z coordinate is right in the file? I was tagging as PARTIAL_COMPATIBLE if there was not a match between parent and linked Vertical Datum data.

It's possible but probably not worth it as it requires parsing the WKT, and even then with custom WKTs I don't even want to know :)

If you're curious for a sample, the buildingSMART docs give https://epsg.io/9286 (see `VDATUM` in the WKT) and EPSG:9306.

Sorry for the delay, a bit of a big PR to chew through, and I am also new to the whole IfcDocumentReference approach for linking which I think was a previous PR which was merged. Still slowly going through it.

src/bonsai/bonsai/bim/module/project/__init__.py
```diff
@@ -94,13 +97,25 @@
...
+def update_link_ui_on_transform(scene):
+    props = tool.Project.get_project_props()
+    for link in props.links:
+        if link.is_loaded and link.empty_handle:
+            for window in bpy.context.window_manager.windows:
+                for area in window.screen.areas:
+                    if area.type == 'PROPERTIES':
+                        area.tag_redraw()
+            break
+
```
Moult, line 100: Please remove this handler. These handlers are expensive, and moving a link is a dangerous, explicit task which should be handled not via regular grab. It can be superseded by an explicit move feature - will describe.

src/bonsai/bonsai/bim/module/project/operator.py
```diff
@@ -1556,8 +1708,32 @@ def run():
...
+        print(f"DEBUG LoadLink.link_ifc: link.name='{self.link_obj.name}', link.position='{self.link_obj.position}'")
+        if self.link_obj.name == "#0" and not self.skip_position_calculation:
+            self.calculate_link_position()
+            print(f"DEBUG LoadLink.link_ifc: Calculated new position='{self.link_obj.position}'")
+        else:
+            print(f"DEBUG LoadLink.link_ifc: Using existing position from saved link or skipping calculation")
+
+        if tool.Ifc.get():
+            # Don't update positions during linking - only during save
+            success = tool.Project.save_linked_models_to_ifc(update_positions=False)
```
Moult: Please remove the call in export_ifc. We should update positions during an explicit move operation, since moving a link is a dangerous and explicit task. Will describe shortly.

src/bonsai/bonsai/bim/module/project/operator.py
```diff
@@ -1576,20 +1752,235 @@ def set_model_origin_from_link(self) -> None:
...
+        for context in ifc_file.by_type("IfcGeometricRepresentationContext", include_subtypes=False):
+            if hasattr(context, "HasCoordinateOperation") and context.HasCoordinateOperation:
+                projected_crs = context.HasCoordinateOperation[0].TargetCRS
+                return {
+                    "Name": getattr(projected_crs, "Name", ""),
+                    "Description": getattr(projected_crs, "Description", ""),
+                    "GeodeticDatum": getattr(projected_crs, "GeodeticDatum", ""),
+                    "VerticalDatum": getattr(projected_crs, "VerticalDatum", ""),
+                }
+        return {}
```
Moult: Please move this into util.geolocation similar to get_helmert_transformation_parameters.

You can also use projected_crs.get_info() rather than getattr.

src/bonsai/bonsai/tool/project.py
```diff
@@ -268,49 +269,108 @@ def load_linked_models_from_ifc(cls) -> None:
...
+                        link.position = new_position
+                        
+                        # Update the IfcDocumentReference.Identification
+                        reference_id = int(link.name[1:])
+                        reference = ifc_file.by_id(reference_id)
+                        reference.Identification = new_position
+
+        if "#0" in links:
+            new_link = links["#0"]
+            links_document = cls.get_linked_models_document()
```
Moult: I don't think this is correct. It seems as though there is a single IfcDocumentInformation being used as a "container", which holds one IfcDocumentReference per link. Instead, each link is a distinct document and should have its own IfcDocumentInformation and IfcDocumentReference. If you link the same document twice, then the IfcDocumentInformation should have two IfcDocumentReference (with different positional identifiers).

e.g.

- IfcDocumentInformation (Name=basename of filename by default, but user can change to whatever, Identification can be X, as we won't know the id scheme, Location=filepath)
- - IfcDocumentReference (Identification=coord stuff)
- IfcDocumentInformation of another linked model
- - IfcDocumentReference (Identification=coord stuff)
- - IfcDocumentReference (Identification=different coord stuff because this is linked twice in two spots)

So the custom "Bonsai internal doc..." can be removed :)

Moult (changes requested):

falken10vdl:
> Is there then a way to know in the CRS Name field if the z coordinate is right in the file? I was tagging as PARTIAL_COMPATIBLE if there was not a match between parent and linked Vertical Datum data.
> It's possible but probably not worth it as it requires parsing the WKT, and even then with custom WKTs I don't even want to know :) If you're curious for a sample, the buildingSMART docs give https://epsg.io/9286 (see `VDATUM` in the WKT) and EPSG:9306. 

It looks there is a "Compound" attribute COMPD_CS in the case everything is there. I do not know if this always the case, though.

> Sorry for the delay, a bit of a big PR to chew through, and I am also new to the whole IfcDocumentReference approach for linking which I think was a previous PR which was merged. Still slowly going through it.

No worries! Thanks for your input! very valuable to understand better how the code is setup in Bonsai. Thanks!

src/bonsai/bonsai/bim/module/project/__init__.py
falken10vdl: Perfect. Still do not have clear understanding on how much heavy the handlers are. Moving linked projects was somthign that the Architect community was asking for the use case of placing several buildings within a site in design phase. 
If I understand you right then, we should lock the position of the linked file at creation time and only allow to move it under a dedicated operator?

src/bonsai/bonsai/tool/project.py
falken10vdl: Indeed "here is a single IfcDocumentInformation being used as a "container", which holds one IfcDocumentReference per link". I just used the already available IfcDocumentInformation that was in the code and added an IfcDocumentReference per link (using Location for the filename and identification for the coordinates).
Much nicer what you say. One IfcDocumentInformation per filename and then 1 or more IfcDocumentReference hanging from that IfcDocumentInformation with just populating the identification for the coordinates

src/ifcopenshell-python/ifcopenshell/util/geolocation.py
```diff
@@ -267,6 +267,22 @@ def get_helmert_transformation_parameters(ifc_file: ifcopenshell.file) -> Option
...
+            crs = ifcopenshell.util.element.get_pset(project[0], "ePSet_ProjectedCRS")
+            return crs if crs else {}
+        return {}
+
+    # For IFC4+, get from IfcProjectedCRS
+    for context in ifc_file.by_type("IfcGeometricRepresentationContext", include_subtypes=False):
+        if getattr(context, "HasCoordinateOperation", None):
+            projected_crs = context.HasCoordinateOperation[0].TargetCRS
+            return projected_crs.get_info() if projected_crs else {}
+    return {}
```
Moult: Omit final return, that way we get a consistent signature with `get_helmert_transformation_parameters` i.e. Optional[dict...]

src/ifcopenshell-python/ifcopenshell/util/geolocation.py
```diff
@@ -267,6 +267,22 @@ def get_helmert_transformation_parameters(ifc_file: ifcopenshell.file) -> Option
...
+        project = ifc_file.by_type("IfcProject")
+        if project:
+            crs = ifcopenshell.util.element.get_pset(project[0], "ePSet_ProjectedCRS")
+            return crs if crs else {}
+        return {}
+
+    # For IFC4+, get from IfcProjectedCRS
+    for context in ifc_file.by_type("IfcGeometricRepresentationContext", include_subtypes=False):
+        if getattr(context, "HasCoordinateOperation", None):
+            projected_crs = context.HasCoordinateOperation[0].TargetCRS
```
Moult:
```
if operation := context.HasCoordinateOperation:
    return operation[0].TargetCRS.get_info()
```
No need to use getattr if the schema element is guaranteed to have the attribute. No need to check for TargetCRS because it's a mandatory attribute.

src/ifcopenshell-python/ifcopenshell/util/geolocation.py
```diff
@@ -267,6 +267,22 @@ def get_helmert_transformation_parameters(ifc_file: ifcopenshell.file) -> Option
...
 
+def get_projected_crs(ifc_file: ifcopenshell.file) -> dict[str, Any]:
+    """Get ProjectedCRS information from an IFC file."""
+    if ifc_file.schema == "IFC2X3":
+        # For IFC2X3, check ePSet_ProjectedCRS on IfcProject
+        project = ifc_file.by_type("IfcProject")
+        if project:
+            crs = ifcopenshell.util.element.get_pset(project[0], "ePSet_ProjectedCRS")
+            return crs if crs else {}
+        return {}
```
Moult: Project is mandatory, so no need to check. This can be simplified into one line:

```
if ifc_file.schema == "IFC2X3":
    return ifcopenshell.util.element.get_pset(ifc_file.by_type("IfcProject")[0], "ePSet_ProjectedCRS")
```

Moult (changes requested):

Moult:
Can you also please help implement an `is_locked` similar to how grids / spatials are by default locked? This way, the default state of a link is locked, and a user will explicitly unlock it to move it, and then when they lock it again, that's when the save position occurs.

src/bonsai/bonsai/bim/module/project/__init__.py
```diff
@@ -19,7 +19,6 @@
 import bpy
 
 from . import decorator, gizmo, operator, prop, ui, workspace
-from bpy.app.handlers import persistent
 import bonsai.tool as tool
```
Moult: I think this import can also go if not needed

src/bonsai/bonsai/bim/module/project/__init__.py
Moult: Exactly, moving should be explicit since it is such a dangerous operation for links. The handler as you had coded it effectively redrew and had a draw call for every single thing - this effectively invalidates our UI cache.

falken10vdl:
Many times I see that I am checking unnecessarily for elements regarding the IFC Schema. My knowledge in the schema is quite limited but the way to find out if I have to check or not would be somethig like this?

For example https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcCoordinateReferenceSystem.htm and see that HasCoordinateOperation is an attribute which is not optional, so no need to check its existence, right? 

Thanks!

src/ifcopenshell-python/ifcopenshell/util/geolocation.py
falken10vdl: Ok. Thanks!

src/ifcopenshell-python/ifcopenshell/util/geolocation.py
falken10vdl: All right. Thanks again ;)

src/ifcopenshell-python/ifcopenshell/util/geolocation.py
falken10vdl: ok.

src/bonsai/bonsai/bim/module/project/operator.py
falken10vdl: done.

src/bonsai/bonsai/bim/module/project/__init__.py
falken10vdl: yes.

Moult:
Yep exactly :) HasCoordinateOperation is an attribute, so getattr is unnecessary , but it has a cardinality of SET[0:1] so it might have no values, so falsey still needs checking.

falken10vdl:
> Can you also please help implement an `is_locked` similar to how grids / spatials are by default locked? This way, the default state of a link is locked, and a user will explicitly unlock it to move it, and then when they lock it again, that's when the save position occurs.

Added commit to have the is_locked property. Now the ifc saving happens when locking as suggested.
https://github.com/IfcOpenShell/IfcOpenShell/pull/7607/commits/3147506e3b53dabeb6bd1e3143cca6ba5d54f7d1

Cheers!

falken10vdl:
If you do not mind, once you are happy with the implementation of all the functionality I can then update the corresponding tests.
Thanks!

src/bonsai/bonsai/bim/module/project/operator.py
```diff
@@ -1360,29 +1370,103 @@ def draw(self, context):
...
+            self.report({"ERROR"}, "Cannot link IFC without a parent IFC project")
+            return {"CANCELLED"}
+        
         start = time.time()
         files = [f.name for f in self.files] if self.files else [self.filepath]
+        
+        if not files or all(not f or not f.strip() for f in files):
+            self.report({"ERROR"}, "No file selected")
+            return {"CANCELLED"}
+
```
Moult: What scenario triggers this code?

src/bonsai/bonsai/bim/module/project/operator.py
```diff
@@ -1360,29 +1370,103 @@ def draw(self, context):
...
+        pprops.false_origin = f"{false_origin_x:.3f},{false_origin_y:.3f},{false_origin_z:.3f}"
+        pprops.project_north = str(round(math.degrees(angle), 1))
+        
+        return super().invoke(context, event)
+
     def execute(self, context):
+        if not tool.Ifc.get():
+            self.report({"ERROR"}, "Cannot link IFC without a parent IFC project")
+            return {"CANCELLED"}
+
```
Moult: I think it's pretty important functionality to maintain to allow linking with no IFC project (this is primarily the workflow of all non-authoring coordination and artistic use). The behaviour in this scenario should be that the first linked file's false origin becomes the host false origin. In the case of no IFC, there is no tool.Ifc.link() so duplication and deletion are treated as regular Blender objects.

src/bonsai/bonsai/bim/module/geometry/operator.py
```diff
@@ -1174,6 +1174,13 @@ def poll(cls, context) -> bool:
     def execute(self, context):
         if not context.selected_objects:
             return {"FINISHED"}
+        
+        if context.active_object and tool.Ifc.get():
+            props = tool.Project.get_project_props()
+            for link in props.links:
+                if link.empty_handle == context.active_object:
+                    return bpy.ops.bim.duplicate_link(link=link.name)
+
```
Moult: This code is in execute(), which is an issue because bpy.ops.bim.duplicate_link may modify IFC data, but execute() does not track IFC undo/redo status, so if the user presses undo after this, unpredictable things happen.

Can you please refactor this to `execute_ifc_duplicate_operator`?

src/bonsai/bonsai/bim/module/geometry/operator.py
```diff
@@ -1174,6 +1174,13 @@ def poll(cls, context) -> bool:
     def execute(self, context):
         if not context.selected_objects:
             return {"FINISHED"}
+        
+        if context.active_object and tool.Ifc.get():
+            props = tool.Project.get_project_props()
+            for link in props.links:
+                if link.empty_handle == context.active_object:
```
Moult: This pattern is an anomaly in Bonsai. Now that links can be IFC entities, they should follow the same pattern as all other Blender-IFC integrations, i.e. tool.Ifc.link(document_reference, empty_handle), that way you can handle it consistently like any other IFC object.

src/bonsai/bonsai/bim/module/project/__init__.py
```diff
@@ -36,6 +36,7 @@
     operator.CreateProject,
     operator.DisableCulling,
     operator.DisableEditingHeader,
+    operator.DuplicateLink,
```
Moult: Why does this operator exist? It should just be a tool.Project function which is called by the geometry duplicate operator (which you already do).

src/bonsai/bonsai/bim/module/project/prop.py
```diff
@@ -221,9 +240,55 @@ class FilterCategory(PropertyGroup):
 class Link(PropertyGroup):
     name: StringProperty(
         name="Name",
+        description="STEP ID of the IfcDocumentReference (format: #123). Used as unique identifier for dictionary lookup",
```
Moult: I think once tool.Ifc.link() is used, using this as a public STEP ID can be removed.

Moult (changes requested):

src/bonsai/bonsai/bim/module/project/operator.py
falken10vdl: Added back the functionality to link without parent IFC: d1dc606. As you pointed out in that case shift-d does not create new link but just new collection

falken10vdl:
@Moult Since we are now storing a single file as cache but we are allowing to have several linked projects of the same file it makes sense to have the projects origin (the geometry handle) at the 3d origin, In the case of manual  you can see that the cache file has the same position relative to 3d as the original ifc file, so then when it is linked we move that to the saved position. That allows you see in the UI panel the coordinates where it is placed relative to the "host project". In the case of automatic I left it as it is since there is this calculations done with distance_limit and I have not gone over the details of it, but that means that the blend file cache has the the handle of the geometry at the 3d origin but the geometry is where the georef data is saying (so Northing & Easintg.. vey big numbers). Why is that done like so?

Moult:
Sorry I don't think I understand the question.

The cache file is an "implementation" detail specific to Blender, not semantically important to IFC/BIM. We always had a Blender cache file generated. The concept of "AUTOMATIC", "MANUAL", "DISABLED" is also a Bonsai-specific implementation detail, and ideally should not be mixed in with the semantics of IFC. The "Blender origin" is merely a detail we track to ensure we don't go out of Blender's capable ranges.

From the IFC perspective, when you federate multiple files together, there is only one set of correct coordinates to consider: the CRS coordinates. So all math must be done in CRS coordinates. These are typically large numbers.

Not entirely sure if that helped answer the question :)

falken10vdl:
@Moult I have recorded a video. Hopefully I can explain it better ;)

https://github.com/user-attachments/assets/432966a6-417d-4eaa-8f2d-25a181b450b3

I do not understand why when a chunk is created the origin for the geometry is kept to 0,0,0 and the geometry is placed according to the georef info. Why not keep all the blend cache files as if they were just saved with the geometry positioned like when they are loaded by bonsai standalone (so close to the 3d origin) and then during link load, use the location information to create an empty in that particular position and move the blend cache origin there.
Note that in the case of MANUAL, if we create several links using MANUAL, I think we have to do as I am doing (maintain the same geometry and move it based on the location info) or otherwise we would need to generate one cache file per manual link since those cache files have the geometry in different positions. Correct?

Thanks Moult for your help! All your comments are really making me learn much faster how is Bonsai set up. 
Sorry for my poor explanations. My mother tongue is Spanish and we tend to make too long sentences that probably confuse.. 

Cheers!

Moult:
Thanks, the video helps. The reason the behaviour is like that is because when the feature was first coded, there was _no_ "empty_handle" and no storage of coordinates or position. It was just baked into a Blend file and that was it. This comes with lots of limitations as you mention (you can't even logically link that same cache into another project, since it's baked into one "host model"'s coordinates). This is the time to fix it.

Now that we can move things around, multi-link IFCs, and store semantic IFC document references, definitely it should be improved.

I totally agree with your suggestion. The ability for AUTO/MANUAL/DISABLED should be retained (and AUTO tweaked to be identical to project open, i.e. it does not consider a "host" model) for the purposes of generating a blend with objects close to the Blender origin. Then, as you say, the position should be just stored semantically as a translation/rotation of CRS coordinates in the IfcDocumentReference, and the empty_handle should be used as a way for the user to move things around. This effectively downgrades AUTO/MANUAL/DISABLED from the primary way positions are baked into an [implementation detail](https://github.com/IfcOpenShell/IfcOpenShell/pull/7607#issuecomment-3871348624).

falken10vdl:
Thanks for the rationale behind it! Now it is understood why it was like that. 

As you can see the current PR already handles the origin in MANUAL mode by means of positioning the empty handle based in the IfcDocumentReference Location information. For AUTOMATIC I skipped it since I did not know the reason until know ;)  If it is fine I can try to also change that one so the cache blend file is created without any location offset and we handle the positioning with the IfcDocumentReference.

Cheers!

Moult:
Cheers, I still have a couple of mental todo items to address before looking at coordinates. Do you have a preference between me committing directly vs making comments and leaving the code changes to you?

falken10vdl:
I learn more from your commits :) but up to you. Let me know what you like best.
Cheers!

Moult:
This PR has been hanging around for a while, so I'll continue committing what I can. I think it's almost done now, now it's just addressing coordinate calculations, which is best done in tandem with test writing.

Note about the reversion of the cache clearing [21e2215](https://github.com/IfcOpenShell/IfcOpenShell/pull/7607/commits/21e2215552c5bad1518577534be2651abd1cd500) - it's a cool trick but in this case I'd first like to explicitly test with and without cache, and subsequent operations need the cache to work properly (e.g. to read coordinates from the metadata json).

falken10vdl:
@Moult Wow! So nice structured. So from the coordinates calculation, the process is that in the IfcDocumentReference a transformation matrix is stored that will be applied on top of the georeferencing done previously between the CRS of the host file and the CRS (if available) of the linked file, right? So that means that if that location info is the identity matrix (and there is CRS in the child) the child is correctly geolocated? And that means that just by testing those one can set placed_as_per_georef icon in the UI.

Moult:
Correct, testing for a non-identity transformation is all you need to do to determine `placed_as_per_georef`. Also this can allow us to remove all the unnecessary metadata about host model origin/north.

Moult:
OK, I think I'm done. There's always more to do and polish but I think it's mergeable. Let me know if you spot anything strange.

theoryshaw:
Dumb question, is the intention to add the following things, from #7570, at a later time? 
- handling SurfaceStyles
- importing IfcGroups as child collections
- being able to immediately move a manual, or disabled link. (mirroring?)
- duplicating links

Moult:
I couldn't find any posts or code related to SurfaceStyles, could you help explain what was that feature related to?

For groups, I believe the original intent was to allow for multiple links. I think this is now resolved because links are now native IFC document references.

Links when initially loaded are locked, but you can edit them to move them and yes, even mirror.

Links as I've just tweaked it cannot be duplicated (given the relatively rarity of this operation). If you want the same link twice, just add it twice.

Does that help?

theoryshaw:
> I couldn't find any posts or code related to SurfaceStyles, could you help explain what was that feature related to?

Brought up here: https://community.osarch.org/discussion/comment/27910/#Comment_27910

> 
> For groups, I believe the original intent was to allow for multiple links. I think this is now resolved because links are now native IFC document references.

The groups functionality, evolved from this thread: https://community.osarch.org/discussion/3302/reflecting-ifcgroups-in-the-outliner-hierarchy/p1

> 
> Links when initially loaded are locked, but you can edit them to move them and yes, even mirror.

Would vote to have `disabled` be unlocked by default. I can see locking `manual` and `auto` though.

> 
> Links as I've just tweaked it cannot be duplicated (given the relatively rarity of this operation). If you want the same link twice, just add it twice.

Duplicating an already linked link in the scene, was nice to have, especially if you want to duplicate something like an identical house around an entire subdivision.  If there's 20 houses, for example, having to go through the steps to link 20 times, would be a little cumbersome.

theoryshaw:
For what it's worth, here are some common test files, that have 'group options' and 'external styles': https://hub.openingdesign.com/OSArch/Community_Troubleshooting/src/branch/main/7570_IFC%20link%20management%20with%20UUID%20support.%20Same%20file%20can%20be%20linked%20multiple%20times

Moult:
Ah thanks, I had a closer look:


 - handling SurfaceStyles: yes can implement outside this PR, I suspect there is a cleaner way to implement it than was previously proposed
 - importing IfcGroups as child collections: probably warrants broader discussion here, also best addressed as a separate PR
 - being able to immediately move a manual, or disabled link. (mirroring?) - I think basically addressed.
 - duplicating links: can implement, but not now. I had an attempt but quickly realised that the duplicate function has some more fundamental issues which need to be addressed first.

PR #7609, opened 2026-01-27
Fix schema case-sensitivity and add IFC2X3 entity geometry support

HeyVico:
This commit addresses several issues in the IfcOpenShell parser:

1. Schema Case-Insensitivity Fix:
   - Fixed schema_definition constructor to use uppercase keys for the global schemas map, ensuring consistent case-insensitive lookup
   - Previously, the constructor stored schemas with original case while register_schema() and schema_by_name() used uppercase, causing potential inconsistencies (e.g., "ifc4x1" failing where "IFC4X1" worked)

2. Geometry Mapping for IFC2X3-Only Entities:
   - Added geometry mapping support for Ifc2DCompositeCurve (IFC2X3 only)
   - Added geometry mapping support for IfcBezierCurve and IfcRationalBezierCurve (IFC2X3 only)
   - These entities existed in the schema but lacked geometry processing

3. Cross-Platform Distribution Consistency:
   - Updated conda build.sh and build.bat to include all IFC4x3 schema variants
   - Changed from limited "2x3;4;4x1;4x3_add2" to full set "2x3;4;4x1;4x2;4x3;4x3_tc1;4x3_add1;4x3_add2"
   - This ensures conda packages have the same schema support as other builds

4. Test Cases:
   - Added comprehensive test suite for schema case-insensitivity
   - Added tests for IFC2X3-specific entities (Ifc2DCompositeCurve, IfcBezierCurve)
   - Added tests verifying these entities don't exist in IFC4+

https://claude.ai/code/session_01ApSKR6uq31dYyPgtMaJUto


aothms:
Hi, I appreciate the intention, but in this case all four fixes are not intentional. I think you'll have more success when trying to fix verified issues. The ifcopenshell code base is large and messy, but many things do have a reason.

1. These schemas are only written at initialisation time using uppercase names, there is no need for this.
2. Ifc2DCompositeCurve, IfcRationalBezierCurve are subtypes of entities already handled https://standards.buildingsmart.org/IFC/RELEASE/IFC2x3/TC1/HTML/ifcgeometryresource/lexical/ifcrationalbeziercurve.htm https://standards.buildingsmart.org/IFC/RELEASE/IFC2x3/TC1/HTML/ifcgeometryresource/lexical/ifc2dcompositecurve.htm There is no need to provide specific support for them
3. These schema discrepancies are intentional as the size of the binary greatly increases because of them
4. Thanks, but we do not need test cases for the entities outside of the schema

PR #7610, opened 2026-01-27
Claude/audit codebase e92 dm

HeyVico:


aothms:
There's some useful bits in here, like Pathlib.touch() but there's also many questionable items. Maybe if you break it up into more thematic clearly scoped fixes you'll have a higher chance of having people to review them. Code is always used in a particular context, this is not a webapp running untrusted input, so blanket fixes for things like sql injection should really be reviewed in context of the code paths and usage. A quick assertion would be fine for the security-conscious, but so elaborate code is really a distraction in the code base.

PR #7649, opened 2026-02-08
Fixes needed to build IfcOpenShell with ParaView

danlipsa:
We build a VTK IFC reader and we added it to ParaView. When using the ParaView superbuild we found and fixed a few minor issues as seen in this PR.

We found these issues when building as a shared library on Windows.

See:
https://gitlab.kitware.com/paraview/paraview-superbuild/-/merge_requests/1310


danlipsa:
I rebased on the current v0.8.0, hopefully this would fix the CI errors.

PR #7653, opened 2026-02-08
Add material associations to IFC4 Demo Library window/door types

jaliste:
Window and door types in the demo library had IfcSurfaceStyles for
appearance but no IfcMaterial associations. This caused materials
to be lost when converting to parametric elements.

Added:
- IfcMaterial entities for Frame, Glass, and Panel styles
- IfcMaterialDefinitionRepresentation linking styles to materials
- IfcMaterialConstituentSet with Framing/Glazing constituents
- IfcRelAssociatesMaterial associating constituent sets with types

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>


Moult:
G'day! Thanks for the PR, a few comments:

 1. You might want to edit `src/bonsai/scripts/generate_demo_library.py` instead, as it is a generated library
 2. There are specific rules on how to associate material constituents to styles using shape aspects, perhaps https://docs.ifcopenshell.org/autoapi/ifcopenshell/api/material/index.html#ifcopenshell.api.material.set_shape_aspect_constituents can help explain

jaliste:
Thanks, so I try now to do it in the genertae_demo_library.py and I added a map to translate from "Frame" to "Framing"... but maybe it's better to change the original names in the materials in the .blend file to match the IFC naming "Framing" "Glazing"?

Moult:
Thanks, this is better! I think you need to add the shape aspects first though: https://docs.ifcopenshell.org/autoapi/ifcopenshell/api/geometry/index.html#ifcopenshell.api.geometry.add_shape_aspect

BIMvoice:
Took a look at where this landed. Good news: your last 3 commits already do what Thomas asked for, the thread just went quiet before anyone confirmed it. Verified live in headless Blender, appending DT01/WT01 from the regenerated library and running the real conversion operators:

Windows are fully fixed. Frame and Glass map to the Framing/Glazing constituent names the parametric window builder recognizes, so both materials survive conversion.

Doors are 90% there. Frame carries through correctly, but the Panel style becomes a constituent literally named "Panel", which nothing downstream recognizes (`get_constituents_props_data` only knows lining/framing/glazing, and `add_door_representation` groups the door leaf under the same Framing shape aspect as the frame, there's no independent Panel shape aspect anywhere in the parametric door builder). So the Panel material gets silently dropped on conversion and the leaf falls back to the frame's color, which is the exact bug this PR set out to fix, still present for doors specifically.

Small, targeted fix: map `"Panel": "Framing"` in `constituent_name_map` (`generate_demo_library.py`), same idea as Glass to Glazing. This keeps Frame and Panel visually distinct in the library's own preview (each still gets its own IfcStyledItem) while giving the door type a single Framing constituent the existing conversion code can actually consume. Verified: before, a converted door kept only Lining/Framing both pointing at the Frame material with Panel orphaned; after, same clean Lining/Framing result with no leftover unused Panel entity.

A real independent Panel shape aspect for doors would need a change to `add_door_representation` itself, a separate, bigger change to the shared geometry builder, out of scope here.

Happy to open this as a one-line PR against your branch if that's easier, just say the word.

PR #7659, opened 2026-02-10
Fix window/door preview alignment regression in ProductDecorator

jaliste:
## Summary
- Fix preview position offset for windows/doors on walls not at origin
- Fix undefined 'context' variable (use bpy.context instead)
- Recalculate type object's inverted matrix each frame to handle representation switches

These are regressions introduced in e55955b72 ("Fix unusable lag due to dense meshes with product preview").

## Test plan
- [ ] Select Window Tool, hover over walls at different positions - preview should follow mouse correctly
- [ ] Test with walls at origin and walls far from origin - both should work
- [ ] Test window placement on both sides of a wall

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/decorator.py
+++ b/src/bonsai/bonsai/bim/module/model/decorator.py
@@ -1567,7 +1567,7 @@ class ProductDecorator:
                     obj_type,
                     representation,
                 )
-                context.view_layer.update()
+                bpy.context.view_layer.update()
                 break
 
         translate_mouse = Matrix.Translation(mouse_point)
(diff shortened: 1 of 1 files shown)


Moult:
The regression seems to have existed prior to the optimisation in [e55955b](https://github.com/IfcOpenShell/IfcOpenShell/commit/e55955b72a6f7d33ca7cbd7ca3deea4f474de8cd). I had noticed it at the time but didn't have the time to address. I think this is something best for @brunoperdigao to check.

PR #7668, opened 2026-02-12
buildingSMART Data Dictionary URL made interchangeable

Schi11er:
In order to be able to use other data dictionaries that offer an API based on the bSDD model, I have added a text field for other base URLs. This allows users to apply the functionality to their own data dictionary.

Tested with our own data dictionary.


Moult:
Isn't the whole point of the bSDD that there is a centralised database we can connect to? Ping @atomczak

Schi11er:
I totally agree with you, but we see two cases for other data dictionaries.

1. The governments of the countries want sovereignty over their national dictionaries. So some are developing their own.

2. When developing dictionaries, experts want to work together on the dictionary. For example, everyone should be able to comment and model different variants. Imports from different data formats are required. Here, we see bSDD more as a medium for publishing the final dictionary, but not as a development tool.

Therefore, our idea was to expand the existing technologies instead of developing a new plugin.

Moult:
OK, agreed, can you reimplement this using a property (similar to how we have properties like include_preview_dictionaries) and expose it in the addon settings UI instead? Have it default to `https://api.bsdd.buildingsmart.org/api/`. You can then remove `update_bsdd_baseurl`, remove `BIM_OT_bsdd_reset_baseurl` and load bsdd dictionaries can read from that prop directly and always override the baseurl.

Schi11er:
Thanks for your advice. I moved it to the addon settings as recommended.

atomczak:
> Isn't the whole point of the bSDD that there is a centralised database we can connect to?

That is the point, yes, to make implementation easier and interconnect the dictionaries. But if someone has another source of dictionaries, they are free to use it instead of bSDD. Nobody is forced to use it.

PR #7671, opened 2026-02-13
Update IfcParse.cpp

ssg3d:
IfcOpenshell read file, and write file without changes. This round trip introduces truncation noise. It should not hurt to increase the precision to keep this clean.


aothms:
Thanks

PR #7684, opened 2026-02-17
Fixes in  AddReferenceImage due to bugs added in 43b12d5

falken10vdl:
This addresses the weird pattern used in the commit that breaks Blender tests, because return context.window_manager.invoke_props_dialog(self) segfaults on headless test execution of Blender.

### DIFF
--- a/src/bonsai/bonsai/bim/import_ifc.py
+++ b/src/bonsai/bonsai/bim/import_ifc.py
@@ -117,7 +117,6 @@ class MaterialCreator:
         for texture in texture_style.Textures or []:
             if coords := getattr(texture, "IsMappedBy", None):
                 coords = coords[0]
-                # IfcTextureCoordinateGenerator handled in the style shader graph
                 if coords.is_a("IfcIndexedTextureMap"):
                     return coords
                 # TODO: support IfcTextureMap
@@ -135,6 +134,10 @@ class MaterialCreator:
                 if shape_has_openings and coords.is_a("IfcIndexedTextureMap"):
                     continue
                 tool.Loader.load_indexed_map(coords, self.mesh)
+            elif tool.Style.get_texture_style(material):
+                # No explicit coordinate mapping (e.g. IFC4 COORD relies on
+                # generated UVs). Bake XY→UV so Solid Texture mode works.
+                tool.Loader.load_generated_uv_map(self.mesh)
 
     def assign_material_slots_to_faces(self) -> None:
         if not self.mesh["ios_materials"]:
--- a/src/bonsai/bonsai/bim/import_ifc.py
+++ b/src/bonsai/bonsai/bim/import_ifc.py
@@ -85,7 +85,7 @@ class MaterialCreator:
         if element.is_a("IfcTypeProduct"):
             self.parse_element_type_material_styles(element)
         self.parsed_meshes.add(self.mesh.name)
-        if not self.ifc_import_settings.load_indexed_maps:
+        if self.ifc_import_settings.load_indexed_maps:
             self.load_texture_maps(shape_has_openings)
         self.assign_material_slots_to_faces()
         tool.Geometry.record_object_materials(obj)
--- a/src/bonsai/bonsai/bim/import_ifc.py
+++ b/src/bonsai/bonsai/bim/import_ifc.py
@@ -135,8 +135,8 @@ class MaterialCreator:
                     continue
                 tool.Loader.load_indexed_map(coords, self.mesh)
             elif tool.Style.get_texture_style(material):
-                # No explicit coordinate mapping (e.g. IFC4 COORD relies on
-                # generated UVs). Bake XY→UV so Solid Texture mode works.
+                # No explicit coordinate mapping (e.g. IFC2X3 has no IsMappedBy,
+                # and IFC4 COORD uses generated UVs). Bake XY→UV as fallback.
                 tool.Loader.load_generated_uv_map(self.mesh)
 
     def assign_material_slots_to_faces(self) -> None:
--- a/src/bonsai/bonsai/bim/import_ifc.py
+++ b/src/bonsai/bonsai/bim/import_ifc.py
@@ -892,48 +892,6 @@ class IfcImporter:
                 obj, tool.Loader.apply_blender_offset_to_matrix_world(obj, self.get_element_matrix(element))
             )
 
-        if element.is_a("IfcAnnotation") and getattr(element, "ObjectType", None) == "IMAGE":
-            image = None
-            if obj.data and obj.data.materials and obj.data.materials[0]:
-                material = obj.data.materials[0]
-                if material.use_nodes and material.node_tree:
-                    for node in material.node_tree.nodes:
-                        if node.type == "TEX_IMAGE" and node.image:
-                            image = node.image
-                            break
-            if image:
-                import bmesh
-
-                bm = bmesh.new()
-                bm.from_mesh(obj.data)
-                if not bm.loops.layers.uv:
-                    uv_layer = bm.loops.layers.uv.new()
-                else:
-                    uv_layer = bm.loops.layers.uv.active
-
-                if bm.verts:
-                    min_x = min(v.co.x for v in bm.verts)
-                    max_x = max(v.co.x for v in bm.verts)
-                    min_y = min(v.co.y for v in bm.verts)
-                    max_y = max(v.co.y for v in bm.verts)
-
-                    width = max_x - min_x
-                    height = max_y - min_y
-
-                    for face in bm.faces:
-                        for loop in face.loops:
-                            vert = loop.vert
-                            u = (vert.co.x - min_x) / width if width > 0 else 0.5
-                            v = (vert.co.y - min_y) / height if height > 0 else 0.5
-
-                            u = max(0.0, min(1.0, u))
-                            v = max(0.0, min(1.0, v))
-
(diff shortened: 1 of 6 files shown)


Moult:
Yep this is much better!  I've also just learned about the `check()` function :)

Can you add `unit="LENGTH"` to the x/y properties, then you can remove all the get_project_unit logic because it will already be in project length units. You can also remove the text describing it to be in length units. Because we already have a convention that every length dimension everywhere in the UI is in project length units.

I also see you've manually created the texture for IFC2X3. This is really, really good. A bit of history: in the past, we called the `add_representation` API which was a huge, tangled function which "magically" handled almost every time of representation possible. Because a long time ago, you just modeled a Blender thing and it "auto-converted" it intelligently to IFC, but now we have actual proper IFC item editing with proper IFC representation item primitives, so this means we built more granular code for editing geometry, materials, styles, ... and now is the time for textures too!

Long story short I'm really excited, can you have a shot at:

 - Manually create the texture always, not just for IFC2X3
 - Change `should_add_representation=True,` to `False`. Then see `root/operator.py`'s `bim.add_element` (this is how new elements of any kind are added) in the `representation_template == "MESH"` case to see how to add a representation manually. This would then allow you to remove the hacky post-processing `item.Coordinates.CoordList = new_coords` and `obj.data["ios_item_ids"] = [item_id] * num_faces`.

... in the long term, this should be just part of `add_element`. After all, this function shouldn't be so special, it's just a hardcoded IfcAnnotation class and a particular `representation_template`.

falken10vdl:
@Moult Last commit I added was just to make it faster for the user to either add a new Ifc Reference Image or to select from the IfcProject (also we avoid to check on the input from a free text field).
Cheers!

Moult:
BTW what's the point of the existing_object_by_name feature? Surely they can just edit the image path using all the existing ways we can manage texture styles? I reckon it can be removed and it's probably just a historic remnant.

Moult:
BTW this refactor I just did to use switch_geometry exposes another issue (which already existed prior to this PR), which I'm not sure if it's a Blender issue or not, that in solid viewport shading textures aren't shown if there aren't any UV coordinates, even if its "generated" UV coordinates.

Ping @Andrej730 ideas? To recreate (e.g. from the main branch): Add > Image > IFC Reference Image, then switch geometry (i.e. reload) the tessellation, and you'll see the viewport preview disappear. You can readd it by manually adding a new bunch of UVs in Data tab > UV Maps > Add, but I don't see why this is necessary, after all it works in Material preview and rendered mode.

falken10vdl:
> BTW what's the point of the existing_object_by_name feature? Surely they can just edit the image path using all the existing ways we can manage texture styles? I reckon it can be removed and it's probably just a historic remnant.

It was added in https://github.com/IfcOpenShell/IfcOpenShell/commit/da306c6141d9c8d8a577289f4f21fcb88553dcbe
It allows to select an already existing annotation and just change the image  with a new one (with the dimensions of the old one). To me just having the "add" feature would be enough becaue this "modify" can be accomplished as you say or just deleting the old and creating a new one (it is not tedious I think).

Cheers!

Moult:
Ah yeah this predates all of the new style management ui that we now have. Feel free to delete that branch.

falken10vdl:
> BTW this refactor I just did to use switch_geometry exposes another issue (which already existed prior to this PR), which I'm not sure if it's a Blender issue or not, that in solid viewport shading textures aren't shown if there aren't any UV coordinates, even if its "generated" UV coordinates.
> 
> Ping @Andrej730 ideas? To recreate (e.g. from the main branch): Add > Image > IFC Reference Image, then switch geometry (i.e. reload) the tessellation, and you'll see the viewport preview disappear. You can readd it by manually adding a new bunch of UVs in Data tab > UV Maps > Add, but I don't see why this is necessary, after all it works in Material preview and rendered mode.

I have seen that this: 
```
       if element.is_a("IfcAnnotation") and getattr(element, "ObjectType", None) == "IMAGE":
            image = None
            if obj.data and obj.data.materials and obj.data.materials[0]:
                material = obj.data.materials[0]
                if material.use_nodes and material.node_tree:
                    for node in material.node_tree.nodes:
                        if node.type == "TEX_IMAGE" and node.image:
                            image = node.image
                            break
            if image:
                import bmesh

                bm = bmesh.new()
                bm.from_mesh(obj.data)
                if not bm.loops.layers.uv:
                    uv_layer = bm.loops.layers.uv.new()
                else:
                    uv_layer = bm.loops.layers.uv.active

                if bm.verts:
                    min_x = min(v.co.x for v in bm.verts)
                    max_x = max(v.co.x for v in bm.verts)
                    min_y = min(v.co.y for v in bm.verts)
                    max_y = max(v.co.y for v in bm.verts)

                    width = max_x - min_x
                    height = max_y - min_y

                    for face in bm.faces:
                        for loop in face.loops:
                            vert = loop.vert
                            u = (vert.co.x - min_x) / width if width > 0 else 0.5
                            v = (vert.co.y - min_y) / height if height > 0 else 0.5

                            u = max(0.0, min(1.0, u))
                            v = max(0.0, min(1.0, v))

                            loop[uv_layer].uv = (u, v)

                bm.to_mesh(obj.data)
                bm.free()
```
 In create_product (import_ifc.py) bakes UV coordinates from the mesh geometry so the image texture displays correctly in Blender's Solid viewport "Texture" mode. 

So one possibility  (for IFC4. Have not taken a look to IFC2X3) would be to change IfcTextureCoordinateGenerator Mode="COORD" to use the "UV"
shader output instead of "Generated"  (The "Generated", correct me if I am wrong, works in Material/Rendered mode but produces no data for Solid mode).

The code above could be put in a single method (somthing like load_generated_uv_map) that could be used in create_product, apply_scaling, load_texture_maps and add_reference_image

Cheers!

Moult:
Yes I think you're absolutely right. That logic in `import_ifc` should be moved into `tool.Loader`, then reused in this operator, then you can delete the duplicate UV logic that is currently in `bim.add_reference_image`.

> change IfcTextureCoordinateGenerator Mode="COORD" to use the "UV" shader output instead of "Generated"

I think it should stay as Generated. It should work in solid mode, at least it does when I save one of those images then reopen the IFC.

image

falken10vdl:
> I think it should stay as Generated. It should work in solid mode, at least it does when I save one of those images then reopen the IFC.

you are right! I thought it was not possible... but following that pattern is actually cleaner. 
I have added three commits so it is clearer. Please take a look

falken10vdl:
@Moult  why do yo do  [aa39a3b](https://github.com/IfcOpenShell/IfcOpenShell/pull/7684/commits/aa39a3b42625fc0905db72e3cd8f6e7f12f7bb76) ?

load_indexed_maps False was running load_texture_maps since in prop.py you see that default = False:

    load_indexed_maps: BoolProperty(
        name="Load Indexed Maps",
        description="Load indexed maps (UV and color maps)",
        default=False,  # Very slow and hackishly implemented
    )

so when disabled (default), we use the IfcIndexedTextureMap UV.

Maybe the property should be named something like use_colour_maps_over_texture_maps ?

Moult:
`load_indexed_maps` True means that you should load UV and colour maps, False, means not to load them. By default, we don't load them because it's not fast to parse UV coordinates, and if you dig into the code there you'll see some of how its been implemented is very hackish. Ideally this should disappear entirely and be implemented in C++.

There is a subtlety here: IfcTextureCoordinateGenerator by very definition means that UV coordinates are procedurally generated. We shouldn't need to parse UV coordinates at all! Ideally, Blender's "Generated" Texture coordinate socket handles everything for us. We shouldn't need to run any code.

For reasons I don't quite understand, Blender's Generated texture socket works in Material and Rendered mode, but not in Solid mode. To get it to show up in Solid mode, you need explicit UV coordinates (which you can generate with our new `load_generated_uv_map` function). But it is a luxury (and a slow one) which I'm happy to leave off by default.

falken10vdl:
Ah Ok!

> Blender's Generated texture socket works in Material and Rendered mode , but not in Solid mode

I think that is by design.  In https://docs.blender.org/manual/en/5.0/editors/3dview/display/shading.html#solid it says clearly that "This mode utilizes the [Workbench Render Engine](https://docs.blender.org/manual/en/3.6/render/workbench/introduction.html) to render the 3D Viewport. It shows solid geometry but uses simplified shading and lighting without the use of shader nodes"
And the https://docs.blender.org/manual/en/latest/render/shader_nodes/input/texture_coordinate.html is documented under shader nodes
So since in solid mode doesn’t evaluate full material shaders, it would ignore Texture Coordinate outputs like Generated and the only way to have the image displayed would be with the uv mapping as you say.

Moult:
Yeah makes sense :)

falken10vdl:
> To get it to show up in Solid mode, you need explicit UV coordinates (which you can generate with our new load_generated_uv_map function). But it is a luxury (and a slow one) which I'm happy to leave off by default.

What if it is forced just for IfcAnnotations "IMAGE"? So the end user always sees these ones (which should not have complicated meshes, right?)

```
        # Always load texture maps for IfcAnnotation with ObjectType IMAGE (image annotation)
        if element.is_a("IfcAnnotation") and getattr(element, "ObjectType", None) == "IMAGE":
            self.load_texture_maps(shape_has_openings)
        elif self.ifc_import_settings.load_indexed_maps:
            self.load_texture_maps(shape_has_openings)
```

Cheers!

Moult:
Would prefer to avoid those types of exceptions as much as possible :)

PR #7660, opened 2026-02-10
IFC CLI interface

brunopostle:
I'm putting this on a secret branch in an attempt to avoid being slop-shamed, *I* would slop-shame me. All code here is generated by an LLM.

This branch adds two new command-line tools: `src/ifcquery` and `src/ifcedit`, they can be installed separately using something like `pip install -e src/ifcquery/` and `pip install -e src/ifcedit/`. More information about these tools is in their respective README files, here are summaries:

> ifcquery: A CLI tool for querying and inspecting IFC building models. All output is structured JSON (or human-readable text), making it easy to pipe into other tools or scripts.
> 
> ifcedit: A CLI wrapper that exposes all 350+ `ifcopenshell.api` modifier functions as shell commands. Functions are auto-discovered at runtime via introspection.

Actually ifcedit is really cool, and does a lot for such a small tool, but this is not the reason for creating this. What these tools provide is just enough discoverable functionality for an AI agent (I'm using Claude Code) to query and manipulate IFC files in the same way that the same agent can be used to work with source code.

To start you first need to have both tools working and in your `PATH`, which might be simple or hard depending on your system, then create a folder, add an IFC file to the folder, and start up your coding agent.

You need a prompt to get going, something like this (untested, but it will probably work):

> Hi there! you now have the ability to work with IFC files, these are BIM models that describe entire construction projects such as buildings. To do this you have been provided with two tools: ifcquery for inspecting and querying IFC files and ifcedit for manipulating them. These tools are all you need for this task, there is definitely and absolutely no need to ever open the IFC file directly or construct ifcopenshell python code on the fly to access it.
>
> Start by running the tools to discover their functionality, they both have a `--help` parameter that tells you what commands are available. You have been provided with a test IFC file, so try some searches, such as finding a storey and counting windows on that storey; then try some manipulation, add a pset parameter, check it worked, delete it, check that worked, or do some other tests to make sure you have a good outline of the system capabilities.
>
[...]

PR #7697, opened 2026-02-20
Package ifcmcp as a HTML client-side chat app

aothms:
For context see: https://github.com/IfcOpenShell/IfcOpenShell/pull/7660#issuecomment-3936863877


brunopostle:
@aothms all looks crazy to me, so I'll merge it. I don't have enough tokens to check the existing mcp interface right now, but any problems will be fixable. Shouldn't it be called ifcchat?

Issue #7690, opened 2026-02-19
Changing the 'custom offset' for OffsetFromReference sometimes changes other instances as well.


theoryshaw:
### Bug Description

Changing the 'custom offset' for OffsetFromReference sometimes changes other instances as well.  Seems inconsistent, when it does. 

https://github.com/user-attachments/assets/8e237ba8-c92e-4353-b259-6edf50ba102d

### Debug and Error Output

```true

```

theoryshaw:
Here's another example file. 
As you can see the layers thicknesses are not correct, either. 

[Barn_2_0 - Copy.ifc.txt](https://github.com/user-attachments/files/25428721/Barn_2_0.-.Copy.ifc.txt)

https://github.com/user-attachments/assets/8229bd55-46f0-419f-bd13-a27fd7acbce8

BIMvoice:
Cross-linking with #7531 (custom offset for slabs not saving) — same OffsetFromReferenceLine write-back path; see the note there so the pair is fixed together.

Issue #7537, opened 2026-01-06
When you rotate a slab, it makes one of the material layers fatter than the designated thickness.


theoryshaw:
### Bug Description

When you rotate a slab, it makes one of the material layers fatter than the designated thickness.

[slab slope - Copy.ifc.txt](https://github.com/user-attachments/files/24460282/slab.slope.-.Copy.ifc.txt)

<img width="348" height="323" alt="Image" src="https://github.com/user-attachments/assets/bb8a54ef-60e4-4bc4-8692-5a1dc1150337" />

<img width="1920" height="1017" alt="Image" src="https://github.com/user-attachments/assets/88a2a520-432e-4321-ac34-415c8ffcb3c2" />

### Debug and Error Output

```true

```

brunopostle:
I'm seeing two bugs with slabs: 

1. All slabs created in the GUI are 500mm thick regardless of the layerset, one of the layers is increased to make up the difference.
2. After editing the profile of the slab, the slab becomes a hipped roof, but still sliced parallel to the base plane.

brunopostle:
My problem is user error, I have a Slab Type with a BBIM_Roof Pset, which means that new slabs look fine, until I try to edit them and they turn into roofs that are sliced like slabs 🙄. Deleting the Pset makes the problem go away.

While investigating, I did discover that Shift-G to regenerate a slab doesn't actually do anything to the slab representation.

theoryshaw:
Would have taken a 1000 years to pull this off without AI. :)

PR #7704, opened 2026-02-22
Fix slab layer geometry: custom offset isolation, unit scale, and lay…

theoryshaw:
…er ordering

Three related bugs fixed in the slab material layer set workflow:

1. EditAssignedMaterial (operator.py): Applying a custom offset to one slab instance incorrectly regenerated geometry for ALL slabs sharing the same IfcMaterialLayerSet. Replaced regenerate_from_layer_set (sweeps all users) with per-element regenerate_from_occurence for AXIS3 slabs and targeted recalculate_walls for AXIS2 walls. Each element's IfcMaterialLayerSetUsage attributes are now updated individually before regeneration.

2. slice_layerset_mesh (loader.py): Loader.unit_scale is a class variable only set during full file import, so it was stale (= 1) during live geometry updates on foot-based IFC files. Layer bisect planes were being computed in IFC feet while the mesh was in Blender metres, placing all cuts completely outside the mesh. Fixed by computing unit_scale fresh from the IFC file on each call via ifcopenshell.util.unit.calculate_unit_scale.

3. OffsetFromReferenceLine stale / layer order reversed (slab.py, loader.py): A guard (and custom_offset is None) in change_thickness prevented writing the correct OffsetFromReferenceLine (position.z) to the usage when a custom offset was active. This left the value at 0.0 instead of the actual slab bottom (e.g. -1.0 IFC units for a TOP-reference slab), so the bisect starting point co was at the reference plane rather than the slab bottom, reversing layer assignments or missing layers entirely. Removed the guard — safe because each element has its own IfcMaterialLayerSetUsage instance. Reverted the AXIS3 bisect normal back to (0,0,1) (upward from co at slab bottom) which is correct once OffsetFromReferenceLine is properly set.

### DIFF
--- a/src/bonsai/bonsai/bim/module/material/operator.py
+++ b/src/bonsai/bonsai/bim/module/material/operator.py
@@ -614,25 +614,29 @@ class EditAssignedMaterial(bpy.types.Operator, tool.Ifc.Operator):
                     attributes=attributes,
                 )
 
-                layer_sets_to_regenerate = set()
+                slab_planer = slab.DumbSlabPlaner()
+                wall_objs = []
 
                 for obj in objects:
                     obj_element = tool.Ifc.get_entity(obj)
                     obj_material_usage = ifcopenshell.util.element.get_material(obj_element)
-
                     if obj_material_usage and obj_material_usage.is_a("IfcMaterialLayerSetUsage"):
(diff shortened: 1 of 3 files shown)


theoryshaw:
Seems to fix #7690, but not #7537 yet.

theoryshaw:
@brunoperdigao i'm going to close this one too, and see how it goes. :)

PR #7705, opened 2026-02-22
 fix for #7537: When you rotate a slab, it makes one of the material layers fatter than the designated thickness.

theoryshaw:
Fix slab layer geometry for rotated and angled slabs
Three bugs fixed:

1. EditAssignedMaterial was sweeping all slabs that share a layer set
   when changing material properties, instead of updating only the
   selected element's assigned material.

2. slice_layerset_mesh (loader.py): layer bisect positions were scaled
   incorrectly due to a stale unit_scale and a reversed/missing
   DirectionSense guard. Now always recalculates unit_scale fresh and
   correctly applies sense_factor.

3. change_thickness (slab.py): extrusion depth was computed as
   `thickness / cos(obj.rotation_euler.x)`, which incorrectly scaled
   ObjectPlacement-rotated slabs (where the extrusion direction is
   local Z and no scaling is needed). The correct formula is
   `thickness / extrusion_vec.z`, which handles both ObjectPlacement
   rotation (extrusion_vec.z ≈ 1.0 → no scale) and ExtrudedDirection
   tilts (extrusion_vec.z < 1.0 → scale up) uniformly. The resulting
   slab was 1.414× too thick for 45°-rotated slabs, making both
   material layers appear fatter than specified.

   slice_layerset_mesh retains a depth_scale safety factor
   (extrusion_vec.z × ifc_depth / total_layer_thickness) as a
   robustness guard for IFC files from other authoring tools where
   extrusion depth may not match the sum of LayerThicknesses.

### DIFF
--- a/src/bonsai/bonsai/bim/module/material/operator.py
+++ b/src/bonsai/bonsai/bim/module/material/operator.py
@@ -614,25 +614,29 @@ class EditAssignedMaterial(bpy.types.Operator, tool.Ifc.Operator):
                     attributes=attributes,
                 )
 
-                layer_sets_to_regenerate = set()
+                slab_planer = slab.DumbSlabPlaner()
+                wall_objs = []
 
                 for obj in objects:
                     obj_element = tool.Ifc.get_entity(obj)
                     obj_material_usage = ifcopenshell.util.element.get_material(obj_element)
-
                     if obj_material_usage and obj_material_usage.is_a("IfcMaterialLayerSetUsage"):
                         obj_material_usage.OffsetFromReferenceLine = material.OffsetFromReferenceLine
                         obj_material_usage.DirectionSense = material.DirectionSense
                         obj_material_usage.ReferenceExtent = material.ReferenceExtent
 
-                        layer_sets_to_regenerate.add(obj_material_usage.ForLayerSet)
-
                         # Save custom offset to BBIM_MaterialLayer pset
                         tool.Model.save_custom_offset_to_pset(obj_element, obj)
 
-                for layer_set in layer_sets_to_regenerate:
-                    wall.DumbWallPlaner().regenerate_from_layer_set(layer_set)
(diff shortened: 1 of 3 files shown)


theoryshaw:
possible fix for #7537

brunoperdigao:
@theoryshaw Is this PR still relavant? I wasn't able to reproduce #7537

brunoperdigao:
@theoryshaw I see what is going on. See this [commit](https://github.com/IfcOpenShell/IfcOpenShell/commit/67b0ab6685f771ab9111219f0000fb3e5c5ed2a6)  for the explanation. That's why x angle is locked in Blender.
I'll have to revisit this. I'll take a look at the PR, but slabs have been really tricky. One change can affect multiple parts of the code.

theoryshaw:
Yes it's still relevant from my perspective.  If you open the test file `slab slope - Copy.ifc.txt` shared here, https://github.com/IfcOpenShell/IfcOpenShell/issues/7537#issue-3786638210 you'll see the different layer thicknesses.  I believe this PR addresses that condition.

brunoperdigao:
@theoryshaw Just a heads up that I'm going to start working on this.

theoryshaw:
Sounds good. Were you just going to review this pr, or come up with something from scratch?

brunoperdigao:
@theoryshaw A bit of both, I guess. I'm just trying to figure out the best approach since this part of the code is very prone to unintended consequences.

brunoperdigao:
@theoryshaw Could you test now after the latest commits?

There are a few things about the PR that I didn't fully comprehend. I'm not sure if it is necessary to change loader.py and the way the layer set is regenerated. Maybe I missed something. Let me know what you think.

theoryshaw:
@brunoperdigao cool, it seems to work.  I'm going to close this PR, so it doesn't build to BonsaiPR then.  Will keep testing via that build.

PR #7706, opened 2026-02-22
Self-heal inconsistent slab extrusion depth on import

theoryshaw:
Self-heal inconsistent slab extrusion depth on import

In slice_layerset_mesh (loader.py), detect when an AXIS3 slab's stored
extrusion.Depth is inconsistent with the sum of its LayerThicknesses
(i.e. depth_scale != 1.0). This was caused by old Bonsai code in
change_thickness that incorrectly scaled extrusion.Depth by
1/cos(obj.rotation_euler.x) for ObjectPlacement-rotated slabs, making
the mesh 1.414× too tall for 45°-rotated slabs.

Two-part fix applied on first import of affected files:
1. Scale mesh vertices in Z (from the mesh bottom) by 1/depth_scale so
   the local Z span equals total_perp_thickness × unit_scale, giving
   the correct physical slab geometry immediately.
2. Write the corrected extrusion.Depth = total_perp_thickness /
   extrusion_vec.z back to the IFC data so future imports load the
   correct geometry without requiring this correction.

Files saved after this fix will open correctly with depth_scale = 1.0
and no vertex adjustment needed.[Self-heal inconsistent slab extrusion depth on import](https://github.com/IfcOpenShell/IfcOpenShell/pull/7706/changes/fb5b6def61d0a798a03d7282b83663a1e3475898)
In slice_layerset_mesh (loader.py), detect when an AXIS3 slab's stored
extrusion.Depth is inconsistent with the sum of its LayerThicknesses
(i.e. depth_scale != 1.0). This was caused by old Bonsai code in
change_thickness that incorrectly scaled extrusion.Depth by
1/cos(obj.rotation_euler.x) for ObjectPlacement-rotated slabs, making
the mesh 1.414× too tall for 45°-rotated slabs.

Two-part fix applied on first import of affected files:
1. Scale mesh vertices in Z (from the mesh bottom) by 1/depth_scale so
   the local Z span equals total_perp_thickness × unit_scale, giving
   the correct physical slab geometry immediately.
2. Write the corrected extrusion.Depth = total_perp_thickness /
   extrusion_vec.z back to the IFC data so future imports load the
   correct geometry without requiring this correction.

Files saved after this fix will open correctly with depth_scale = 1.0
and no vertex adjustment needed.

### DIFF
--- a/src/bonsai/bonsai/bim/module/material/operator.py
+++ b/src/bonsai/bonsai/bim/module/material/operator.py
@@ -614,25 +614,29 @@ class EditAssignedMaterial(bpy.types.Operator, tool.Ifc.Operator):
                     attributes=attributes,
                 )
 
-                layer_sets_to_regenerate = set()
+                slab_planer = slab.DumbSlabPlaner()
+                wall_objs = []
 
                 for obj in objects:
                     obj_element = tool.Ifc.get_entity(obj)
                     obj_material_usage = ifcopenshell.util.element.get_material(obj_element)
-
                     if obj_material_usage and obj_material_usage.is_a("IfcMaterialLayerSetUsage"):
                         obj_material_usage.OffsetFromReferenceLine = material.OffsetFromReferenceLine
                         obj_material_usage.DirectionSense = material.DirectionSense
                         obj_material_usage.ReferenceExtent = material.ReferenceExtent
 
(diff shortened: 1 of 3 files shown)


theoryshaw:
I'm going to close this one too.  Ping @brunoperdigao 
Will reopen, if i run into this problem again.

PR #7713, opened 2026-02-24
Fix #7712: Fix text annotation alignment not saved on edit

### LINKED ISSUE #7712: text alignment doesn't seem to work.
### Bug Description

text alignment doesn't seem to work.

### Attachments

https://github.com/user-attachments/assets/14bfe0b7-5777-4207-9169-d9b9d58d181d

### Debug and Error Output

```true

```

theoryshaw:
edit_text was missing a call to edit_text_alignment, so align_vertical/align_horizontal prop changes were never exported to IFC.

Additionally, edit_text_alignment was called before edit_text_literals, which deletes and recreates the IFC literal entities — discarding the alignment change immediately. Moved the call to after edit_text_literals.

EditText operator was also being replayed by Blender's undo/register stack with stale prop values, overwriting the correct alignment. Guarded _execute with is_editing to make replayed calls no-ops.

Also fixed a ValueError in is_camera_moved where
np_frombuffer_legacy returns a flat (9,) array but the rotation dot product requires (3,3). Fixed with reshape.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/decoration.py
+++ b/src/bonsai/bonsai/bim/module/drawing/decoration.py
@@ -1788,7 +1788,7 @@ class CutDecorator:
 
         # Handle both old float64 and new float32 checksums for version compatibility
         rot_checksum_bytes: bytes = eval(DecoratorData.camera_rotation_checksum)
-        rot_check = tool.Blender.np_frombuffer_legacy(rot_checksum_bytes, 9)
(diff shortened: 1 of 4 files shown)


theoryshaw:
Maybe fixed via. https://github.com/IfcOpenShell/IfcOpenShell/commit/273ecfe8e4ee3701c54c5be248f37508cbd89830 instead.

PR #7719, opened 2026-02-24
Fix #7718: Fix FallDecorator label calculation for all slope annotation types

### LINKED ISSUE #7718: The 'FALL' tag shows 'NO DATA'.
### Bug Description

The 'FALL' tag shows 'NO DATA'. 

### Attachments

_No response_

### Debug and Error Output

```true

```

theoryshaw:
Fix #7718: Fix FallDecorator label calculation for all slope annotation types

- Fix wrong dict key type in decoration.py: DecoratorData.data["fall"] is keyed by obj.name (str) but was looked up with obj (Object), causing object_type to always be None
- Apply obj.matrix_world transform to spline points before computing rise/run in both decoration.py and svgwriter.py; local coordinates have Z=0 for flat annotations, world coordinates correctly reflect elevation change
- Use hypotenuse (segment_length) instead of run as the denominator for SLOPE_FRACTION label display

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/decoration.py
+++ b/src/bonsai/bonsai/bim/module/drawing/decoration.py
@@ -997,7 +997,7 @@ class FallDecorator(BaseDecorator):
         # generate label text
         # same function as in svgwriter.py
         def get_label_text():
-            B, A = [v.co.xyz for v in spline_points[:2]]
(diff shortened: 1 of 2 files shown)


theoryshaw:
when tabbing in/out, however, for SLOPE_FRACTION and SLOPE_PERCENT and SLOPE_ANLE the text decorations for those elements don't turn on after exiting edit mode, but for 'FALL', it works fine.

PR #7720, opened 2026-02-24
Fixes example in documentation

RickBrice:
The small example for ifcopenshell.api.spatial.reference_structure doesn't work. I've fixed up the example.

I am wondering if the example is the problem or the function itself.

It seems like it would be convenient to reference multiple objects in multiple spatial structures at the same time.


aothms:
I think your fix matches the type hint and also the types of the relationships in ifc https://github.com/IfcOpenShell/IfcOpenShell/blob/2d7a556dd5c24f1ba2d7ab0beb58078ebfbf5669/src/ifcopenshell-python/ifcopenshell/api/spatial/reference_structure.py#L30

PR #7735, opened 2026-03-01
ifcquery plot subcommand

aothms:
Probably concatenating the images is not the right way, probably drawing selection needs to be more interactive, but I think a 2d view could be nice to have as a command as well.

AI generated, poorly tested. Essentially just a wrapper around ifcopenshell.draw but with the (untested) ability also for highlighting.

    $ python -m pdb -m ifcquery ..\..\files\nibs\duplex_a.ifc plot --view auto

d


brunopostle:
@aothms oops, didn't see this pull request, will look

PR #7740, opened 2026-03-03
Add Support for IfcOpenShell on Windows ARM64

tsomanna:
PR Description:

- Adds full Windows ARM64 build support for IfcOpenShell, including native compilation on the ARM64 toolchain.

- Applies ARM64‑specific patches `(win/arm64-patches)` for MPIR `(mpir-arm64-changes)` and MPFR `(mpfr-arm64-changes)` through an updated `build-deps.cmd` that conditionally applies these fixes for Windows on ARM builds.

- Introduces a new CI workflow `(build_win_arm64)` to help automate the build Windows ARM64 build process with build-win-arm64.py helper file


aothms:
@Andrej730 was this intentional?

tsomanna:
Hi @aothms,
I believe the previous PR was closed because the target branch `(windows-test)` was deleted. I initially attempted to apply my changes directly to the `v0.8.0` branch, but I ran into an issue where the generated **IfcGeom.lib file exceeded 4 GB,** which caused the Windows runners to fail since this exceeds the platform’s library size limits.

I also noticed that the same issue occurred in one of @Andrej730’s runs
(https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/21701943474/job/62584150452).
This was fixed on the` windows-test `branch but not on `v0.8.0`, which is why I switched to using windows-test for my builds.

I am currently preparing a new PR targeting the v0.8.0 branch.
Please let me know if this is the correct approach or if I should target a different branch.

aothms:
> generated IfcGeom.lib file exceeded 4 GB, 

Yes we should move to more like a plug-in architecture the sooner the better, as we're running into more and more variations of this problem. Unfortunately I don't have much time the coming months.

>  Please let me know if this is the correct approach or if I should target a different branch.

Sounds great. Thanks for your continued effort on this. Much appreciated.

tsomanna:
Hi @aothms @Andrej730, I've opened a new PR targeting the v0.8.0 branch: [PR 7761](https://github.com/IfcOpenShell/IfcOpenShell/pull/7761)
Please take a look when you have a moment. Let me know if this aligns with the expected workflow.
Thank you!

PR #7744, opened 2026-03-04
Fix #7681: Fix isolate_objects ignoring hide_select/hide_viewport (#7710)

### LINKED ISSUE #7681: When 'disable selection' is on, and you activate a drawing, it turns off the object.
### Bug Description

When 'disable selection' is on, and you activate a drawing, it turns off the object.  I would assume it should have no effect on the blender only object. 

https://github.com/user-attachments/assets/75a44671-cf08-48c0-8009-ec9589190964

### Attachments

_No response_

### Debug and Error Output

```true

```

theoryshaw:
Objects with hide_select=True could not be selected during isolation, causing hide_view_set to incorrectly hide them. Objects with hide_viewport=True had their H-key hide state modified as a side effect of hide_view_clear/hide_view_set. Both are now left unaffected by bim.activate_drawing.

### DIFF
--- a/src/bonsai/bonsai/tool/blender.py
+++ b/src/bonsai/bonsai/tool/blender.py
@@ -1962,14 +1962,32 @@ class Blender(bonsai.core.tool.Blender):
         previously_active = bpy.context.view_layer.objects.active
 
         override = cls.get_viewport_context()
+
+        # Save H-key hide state for globally-restricted objects so we don't change it.
+        # hide_viewport=True means the object is globally hidden via the outliner restriction;
+        # hide_view_clear/hide_view_set should not add or remove an additional H-key hide on them.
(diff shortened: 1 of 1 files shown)


theoryshaw:
@Moult you probably reverted the following for a good reason, but I pushed this PR in the meantime to 'fix' this bug in the BonsaiPR builds. 

https://github.com/IfcOpenShell/IfcOpenShell/commit/41afaaec0d49aa06a840ee34e14ac59c57bb61e1

theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/627ae39868d9ce4c72939eb6869eb5f499bb95f6/logs/summaries/2026-05-25-PR-7744.md

PR #7764, opened 2026-03-09
buildingSMART Data Dictionary module: use pSets from different data dictionary sources

Schi11er:
I made the identifier_url for querying pSets dynamic so that not only classification but also attribution can be done from other data dictionary sources.

In addition to merged pull request #7668


Schi11er (commented):
removed checking errors

Issue #6420, opened 2025-03-24
Blender crash using file browser in Blender 4.4.0


Andrej730:
> This is crash from last night, but I don't remember what caused it:
> 
> ```
> # Blender 4.4.0, Commit date: 2025-03-17 17:00, Hash 05377985c527
> bpy.ops.bim.override_move_macro(BIM_OT_override_move={}, TRANSFORM_OT_translate={"value":(-0.086393, 0.0696743, 0.191748), "orient_type":'GLOBAL', "orient_matrix":((1, 0, 0), (0, 1, 0), (0, 0, 1)), "orient_matrix_type":'GLOBAL', "constraint_axis":(False, False, False), "mirror":False, "use_proportional_edit":False, "proportional_edit_falloff":'SMOOTH', "proportional_size":1, "use_proportional_connected":False, "use_proportional_projected":False, "snap":True, "snap_elements":{'VERTEX', 'EDGE', 'FACE', 'EDGE_MIDPOINT', 'EDGE_PERPENDICULAR'}, "use_snap_project":False, "snap_target":'CLOSEST', "use_snap_self":True, "use_snap_edit":True, "use_snap_nonedit":True, "use_snap_selectable":False, "snap_point":(0, 0, 0), "snap_align":False, "snap_normal":(0, 0, 0), "gpencil_strokes":False, "cursor_transform":False, "texture_space":False, "remove_on_cancel":False, "use_duplicated_keyframes":False, "view2d_edge_pan":False, "release_confirm":False, "use_accurate":False, "use_automerge_and_split":False})  # Operator
> bpy.context.space_data.system_bookmarks_active = 1  # Property
> bpy.context.space_data.params.filename = "untitled.ifc"  # Property
> 
> # backtrace
> Exception Record:
> 
> ExceptionCode         : EXCEPTION_ACCESS_VIOLATION
> Exception Address     : 0x00007FF72DAB7453
> Exception Module      : blender.exe
> Exception Flags       : 0x00000000
> Exception Parameters  : 0x2
> 	Parameters[0] : 0x0000000000000000
> 	Parameters[1] : 0x0000000000000000
> 
> 
> Stack trace:
> blender.exe         :0x00007FF72DAB7400  file_execute_get_description
> blender.exe         :0x00007FF72CA96A30  WM_operatortype_description
> blender.exe         :0x00007FF72D8BCB10  UI_but_string_get_rna_tooltip
[...]

PR #7785, opened 2026-03-12
fix: Correct vertical/cant segment handling in alignment API

DesertSpringsCivil:
## Summary

Two bugs in the alignment API prevented `IfcAlignmentVertical` (and cant) segments from being added when a geometric representation exists.

### Bug 1 — `_add_segment_to_curve.py`

Removed an unconditional `if not curve.is_a("IfcCompositeCurve")` guard that was left over from when the function only supported horizontal segments. The preceding `if/elif/elif` chain already validates the correct curve type for each segment type; the redundant check always raised `TypeError` for vertical (`IfcGradientCurve`) and cant (`IfcSegmentedReferenceCurve`) segments.

### Bug 2 — `add_zero_length_segment.py`

Added a `None` guard before the recursive `add_zero_length_segment(file, layout.BaseCurve)` call for `IfcGradientCurve`. When an `IfcGradientCurve` is created before a horizontal representation exists, `BaseCurve` is `None` and the recursive call crashed with `AttributeError`.

## Test plan

- [ ] Verify `layout_vertical_alignment_by_pi_method` no longer raises `TypeError: Expected to see IfcCompositeCurve` when adding vertical segments to an alignment with a representation
- [ ] Verify `add_vertical_layout` no longer crashes with `AttributeError` on `BaseCurve` when called on an alignment without a horizontal representation

### DIFF
--- a/src/bonsai/bonsai/tool/alignment.py
+++ b/src/bonsai/bonsai/tool/alignment.py
@@ -388,6 +388,68 @@ class Alignment:
         ifc_file = tool.Ifc.get()
         align_api.layout_horizontal_alignment_by_pi_method(ifc_file, layout, hpoints, radii)
 
+    @classmethod
+    def create_representation_structure(cls, alignment: "ifcopenshell.entity_instance", template: str) -> None:
+        """Create semantic layout containers and geometric representation for an IfcAlignment.
+
+        Called from the AddElement operator when the user selects a representation type
+        for a newly-created IfcAlignment.  For layout-based templates (HORIZONTAL, GRADIENT,
+        CANT), the appropriate IfcAlignmentHorizontal / Vertical / Cant entities are nested to
+        the alignment and ifcopenshell.api.alignment.create_representation is invoked to build
+        the matching geometric curves.  For polyline templates, an IfcPolyLine with two
+        placeholder origin points is created directly.
+
+        Layout containers are IFC entities only — no Blender objects are created for them.
+
+        Args:
+            alignment: The IfcAlignment entity that was just created by AddElement.
+            template:  One of ALIGNMENT_HORIZONTAL, ALIGNMENT_GRADIENT, ALIGNMENT_CANT,
+                       ALIGNMENT_POLYLINE_3D, or ALIGNMENT_POLYLINE_2D.
+        """
+        import ifcopenshell.api.alignment as align_api
+        import ifcopenshell.api.nest
+        import ifcopenshell.guid
+        from ifcopenshell.api.alignment._create_polyline_representation import (
+            _create_polyline_representation,
+        )
+
+        ifc_file = tool.Ifc.get()
+
+        if template == "ALIGNMENT_HORIZONTAL":
+            h_layout = ifc_file.createIfcAlignmentHorizontal(GlobalId=ifcopenshell.guid.new())
+            ifcopenshell.api.nest.assign_object(ifc_file, related_objects=[h_layout], relating_object=alignment)
+            align_api.create_representation(ifc_file, alignment)
+
+        elif template == "ALIGNMENT_GRADIENT":
+            h_layout = ifc_file.createIfcAlignmentHorizontal(GlobalId=ifcopenshell.guid.new())
+            ifcopenshell.api.nest.assign_object(ifc_file, related_objects=[h_layout], relating_object=alignment)
+            v_layout = ifc_file.createIfcAlignmentVertical(GlobalId=ifcopenshell.guid.new())
+            ifcopenshell.api.nest.assign_object(ifc_file, related_objects=[v_layout], relating_object=alignment)
+            align_api.create_representation(ifc_file, alignment)
+
+        elif template == "ALIGNMENT_CANT":
+            h_layout = ifc_file.createIfcAlignmentHorizontal(GlobalId=ifcopenshell.guid.new())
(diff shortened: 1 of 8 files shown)


src/ifcopenshell-python/ifcopenshell/api/alignment/_add_segment_to_curve.py
RickBrice, line 1: I don't agree with this change. curve must be an IfcCompositeCurve or one of its derived types.
At the end of this function, _add_curve_segment_to_composite_curve is called. curve must be a composite curve.

src/ifcopenshell-python/ifcopenshell/api/alignment/add_zero_length_segment.py
RickBrice, line 1: While technically ok, I don't think this is the right solution. The IfcGradientCurve is dependent on its BaseCurve. The root problem is that the gradient curve isn't properly created. A gradient curve can't survive on its own and be useful.

src/ifcopenshell-python/ifcopenshell/api/alignment/create_representation.py
RickBrice, line 1: When an alignment is created with ifcopenshell.api.alignment.create(), the nests for the layouts are created. If the IfcAlignment is created another way, it would be better to also create the IfcRelNest that its layouts are going to need. Skipping the call to _add_segment_to_curve if the layout nest does't exist avoids a crash, but masks a more serious problem. create_representation wont properly create the geometry representation of the alignment. In fact, if there isn't a layout from which to generate a geometric representation, then there is no need to call this method. The alignment is probably just a polyline.

src/bonsai/bonsai/tool/alignment.py
```diff
@@ -388,6 +388,68 @@ def layout_by_pi_method(cls, layout: "ifcopenshell.entity_instance", hpoints: li
...
+        import ifcopenshell.api.alignment as align_api
+        import ifcopenshell.api.nest
+        import ifcopenshell.guid
+        from ifcopenshell.api.alignment._create_polyline_representation import (
+            _create_polyline_representation,
+        )
+
+        ifc_file = tool.Ifc.get()
+
+        if template == "ALIGNMENT_HORIZONTAL":
```
RickBrice: It would be better to use ifcopenshell.api.alignment.create with the appropriate value for include_vertical, include_cant, and include_geometry. The api sets up everything you need to at semantic segments to the horizontal, vertical, and cant layouts with the geometric representations being automatically created if geometry is included.

RickBrice (changes requested):

src/bonsai/bonsai/tool/alignment.py
```diff
@@ -388,6 +388,68 @@ def layout_by_pi_method(cls, layout: "ifcopenshell.entity_instance", hpoints: li
...
+            h_layout = ifc_file.createIfcAlignmentHorizontal(GlobalId=ifcopenshell.guid.new())
+            ifcopenshell.api.nest.assign_object(ifc_file, related_objects=[h_layout], relating_object=alignment)
+            align_api.create_representation(ifc_file, alignment)
```
RickBrice: https://github.com/IfcOpenShell/IfcOpenShell/blob/cf42f986c41c6b9ee61d27b8f1800b5ea37cf732/src/ifcopenshell-python/ifcopenshell/api/alignment/create_representation.py#L33-L35

Perhaps the function is poorly named, but it is intended to generate the geometry representation from a fully defined alignment consisting of only the semantic definition.

Use align_api.create(file,"Name",include_geometry=True) to create an alignment with a horizontal layout and its geometric representation. Similar for all the others below.

DesertSpringsCivil:
Thanks for the comments Rick! Sorry, Claude got away from me on this one. I'm building tests for the alignment tools and this was AI's attempt to "fix" a couple tests that weren't working. Let me know if this "fix of the fix" resolves things for you.

RickBrice:
No worries. To me, alignment seem a bit picky and need to be carefully constructed. Some of the geometric entities, like IfcGradientCurve, don't really stand on their own as independent objects. I think most people also think of alignments in terms of the semantic definition since it is so closely related to the geometric representation. This seems to be an oddity in IFC. With that in mind, I tried to make the api handle all the messy details of alignment, layouts, nests and representations behind the scenes.

The api is certainly open to improvements, and I know I'm not the best at choosing function names and writing the documentation. My lack of skill in this area is probably not helpful for AI.

I'm looking forward to your next installment. Hopefully some of what is described in our OSArch discussion (https://community.osarch.org/discussion/comment/28182#Comment_28182) is making it into the UI.

theoryshaw:

Conflict Report:

-   [**PR #7785**](https://github.com/IfcOpenShell/IfcOpenShell/pull/7785): fix: Correct vertical/cant segment handling in alignment API
    -   First detected failing: 2026-03-13
    -   Base commit at first detection:  `cf42f986c41c6b9ee61d27b8f1800b5ea37cf732`
    -   Conflicting files:  `src/bonsai/bonsai/bim/module/alignment/operator.py`
    -   Possible breaking commits:  `ff3933a11 Remove some unused imports`,  `953c5fddc bonsai - sort imports`,  `40f92d15c Alignment api update (#6817)`,  `02d359d0e Fix Ruff UP035 (also deprecated annotation symbols)`,  `ecee3781e Fix more missing Blender descriptions leading to crashes #6420`

PR #7805, opened 2026-03-16
Fix recursion cost item bug

maxfb87:
There is a subtle bug that affect the cost panel.
When you disable a cost item editing, it appears in the console but it doesn't stops blender (so the user doesn't see anything).
It is caused by the update function defined for property change_cost_item_parent because blender activate that also when that variable has the same value of the assignment (for example it is already false and i assign false).
So in order to prevent that, i fixed with an if statement and also added a test for it.


maxfb87:
I think I'm going to merge this because i don't want to get stuck too much on this

PR #7820, opened 2026-03-19
Fix #7275: Make remove_drawing_from_sheet undoable

### LINKED ISSUE #7275: If you `bim.remove_drawing_from_sheet` and then undo, the drawing is still removed from the sheet, when printing the svg.
### Bug Description

If you `bim.remove_drawing_from_sheet` and then undo, the drawing is still removed from the sheet, when printing the svg. 

### Attachments

_No response_

### Debug and Error Output

```true

```

theoryshaw:
## Problem

`bim.remove_drawing_from_sheet` writes the drawing's `<g>` element out of the sheet's layout SVG immediately on execution. Blender's undo mechanism restores IFC data (the `IfcDocumentReference`) but cannot restore files written to disk. This means pressing Ctrl+Z after a remove correctly restores the IFC reference, but the layout SVG on disk is already missing the drawing group. Any subsequent `bim.create_sheets` call reads the stale layout SVG and produces output without the drawing, even though the IFC says it should be there.

The problem is compounded when `bim.create_sheets` is called between the remove and the undo: each call reads the layout SVG from disk, so the missing group is never recovered.

## Solution

Stop modifying the layout SVG in `remove_drawing_from_sheet`. The layout SVG is now only written by additive operations (`add_drawing`, `add_document`, `change_titleblock`, `update_sheet_drawing_sizes`). Stale groups (those whose `IfcDocumentReference` no longer exists in the IFC) are handled lazily at build time:

-   **`build_drawings()`**  — when  `ifc.by_id(data-id)`  or  `ifc.by_guid(data-drawing)`  raises  `RuntimeError`, the group is removed from the in-memory tree so it does not appear in the output sheet SVG. The layout SVG file on disk is left untouched, preserving the group for if/when undo restores the IFC reference.
-   **`next_drawing_location()`**  — filters stale groups out before computing where to place a newly added drawing, so the removed drawing's slot does not affect auto-placement of subsequent drawings.
-   **`_ifc_entity_exists()`**  — small static helper used by  `next_drawing_location()`  to safely probe the IFC without raising.

## Why not clean up the layout SVG during  `build()`?

An earlier iteration of this fix performed a pre-build pass that removed stale groups from the layout SVG and wrote it back to disk. This caused a second regression: if the user called `bim.create_sheets` _between_ the remove and the undo, the pre-build cleanup would permanently delete the `<g>` from the layout file before the undo could restore the IFC reference. After the undo the layout SVG was empty and the drawing was still missing. The current approach avoids all writes to the layout SVG during `build()`, so the group survives any number of `create_sheets` calls and is available whenever undo is triggered.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/sheeter.py
+++ b/src/bonsai/bonsai/bim/module/drawing/sheeter.py
@@ -132,9 +132,22 @@ class SheetBuilder:
         self.add_view_title(x, view_height + y + VIEW_TITLE_OFFSET_Y, view, layout_dir)
         layout_tree.write(layout_path)
 
+    @staticmethod
+    def _ifc_entity_exists(ifc: ifcopenshell.file, entity_id: int) -> bool:
+        try:
+            ifc.by_id(entity_id)
+            return True
+        except RuntimeError:
+            return False
+
     def next_drawing_location(self, layout_root: ET.Element, next_width: float) -> list:
         titleblocks = layout_root.findall(f'{SVG}g[@data-type="titleblock"]')
-        drawings = layout_root.findall(f'{SVG}g[@data-type="drawing"]')
+        ifc = tool.Ifc.get()
+        drawings = [
+            g
+            for g in layout_root.findall(f'{SVG}g[@data-type="drawing"]')
+            if self._ifc_entity_exists(ifc, int(g.attrib.get("data-id", "0")))
+        ]
 
         # how wide is the title block frame
         try:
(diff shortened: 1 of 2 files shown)


Moult:
I don't think this is the right approach. Can you try again telling the AI to instead learn from our undo redo documentation, and perhaps look at how undo and redo is implemented in project operators or the brickschema module for inspiration? Then we can properly edit the <g> element and also make undo redo work because the undo / redo is stored in the operator undo / redo function.

theoryshaw:
Will do.

PR #7826, opened 2026-03-21
Suppress reportRedeclaration in Pyright config

falken10vdl:
The `if TYPE_CHECKING:` redeclaration pattern used in Bonsai `prop.py` files triggers Pyright/Pylance's `reportRedeclaration` diagnostic. Suppressing it globally in `pyproject.toml` avoids scattering `# type: ignore` comments throughout the code.


Andrej730:
Thanks, I'll get rid of those pyright ignores, they indeed are getting too noisy 😬

falken10vdl:
Thank you!

Issue #7800, opened 2026-03-16
Gridline Location | not editable in 4x3


steverugi:
### Bug Description

Adding Grids to an IFC 4x3 and later change (G+x ot y) their location is not possible even if I unlock them from relevant panel.
But it is possible in IFC4.

Workaround: select them, go to Transform and unlock the X,Y Location padlocks (holding Alt to operate on all)

Thanks

Windows 10, Bonsai v0.8.5 alpha260315

<img width="1057" height="563" alt="Image" src="https://github.com/user-attachments/assets/85a85729-4949-416f-8751-d1fcbafe2acb" />

### Debug and Error Output

```true

```

BIMvoice:
Mechanism pointer, also for #6808 (same underlying difference): moving a grid axis funnels through `sync_grid_axis_object_placement` (`src/bonsai/bonsai/bim/export_ifc.py:125`), which ends in `bpy.ops.bim.update_representation` on the axis object. The IFC4-vs-4X3 divergence is therefore in how `update_representation` regenerates the axis `AxisCurve` — IFC4 grid axes round-trip as simple `IfcPolyline`s, while 4X3 grids created by Bonsai carry different curve geometry that the regeneration path does not rewrite, so the move silently doesn't stick (matching your G+X/Y symptom and the padlock having no effect). A fix needs someone with a GUI to trace `update_representation` on a 4X3 axis, but that is the place to look.

Issue #6808, opened 2025-06-16
Gridline editing - different behavior between IFC4 and 4x3


steverugi:
### Bug Description

when I work with an IFC4 model I can move gridline without problem, provided padlock is 'open'
if I use IFC4x3 gridlines, once saved and reloaded, they can only be G+X or Y by entering in Edit Mode (Tab)

thanks

Bonsai v0.8.3 - 250606

### Debug and Error Output

```true

```

franseoane:
Bonsai v0.8.4 - 251107
Error persists for IFC 4x3.

Option to "Edit mode" to move/copy the grids: 
Select an axis + side panel "N" / Element: Transformation / Position (unlock position manually X/YZ)

BIMvoice:
Cross-linking: the same IFC4-vs-4X3 axis-curve regeneration difference is traced in #7800 (see the mechanism pointer there — the move path ends in `update_representation`, which does not rewrite the 4X3 axis curve). These two issues share one root cause and could be merged.

Issue #6231, opened 2025-02-25
Bonsai - Only the location of last duplicated grid gets saved


mdjska:
When duplicating IfcGridAxis with shift-D, and repeating the duplication a couple of times in between saving the IFC file, only the last grid axis that was duplicated before saving gets saved with its own location. The others just reference the IfcPolyline and IfcCartesianPoints of the parent grid axis, the are duplicates of. This happens, specifically when creating 'duplicates of duplicates'.

Reproduced on Windows w. Blender  4.2.3, 4.3.2 and on Macos w. Blender 4.3.2.

1. To reproduce this, create a new project and "Add Grids"
2. Duplicate an IfcGridAxis.
3. Duplicate the new, duplicated IfcGridAxis.
4. Save IFC model (see result below)
5. When loading the IFC project again, the grid axis will of course have moved back to the location that was saved in the model, giving the appearance of moving randomly.

```
#69=IFCGRIDAXIS('A',#72,.T.); -> original grid axis
#70=IFCCARTESIANPOINT((-2000.,0.));
#71=IFCCARTESIANPOINT((22000.,0.));
#72=IFCPOLYLINE((#70,#71));
[...]
#98=IFCGRIDAXIS('A',#72,.T.); -> first duplicate, referencing its parents location
#99=IFCGRIDAXIS('A',#102,.T.); -> second duplicate
#100=IFCCARTESIANPOINT((-2000.,10000.));
#101=IFCCARTESIANPOINT((22000.,10000.));
```

theoryshaw:
i can replicate

mdjska:
This issue still persists in the IFC4x3 version along with that the grids cannot be moved with G shortcuts. The only fix is to save in between every move.

BIMvoice:
@mdjska this looks fixed: theoryshaw's PR #7833 addresses this exact duplicate-of-duplicate bug, which we independently verified (reproduced before/after on your sample file, confirmed the fix is genuine). It also includes a fix for IfcGridAxis unlock not working in IFC4X3, which matches your March follow-up about it persisting there plus the G-shortcut issue.

The PR is still open and awaiting merge. If you get a chance to test it against your own workflow, that would help move it along.

PR #7833, opened 2026-03-23
Grid issues

### LINKED ISSUE #7639: IfcAnnotatoin/GRID doesn't seem to show up in the following file.
### Bug Description

IfcAnnotatoin/GRID doesn't seem to show up in the following file. 

[Clover_Lane_Residence - Copy.ifc.txt](https://github.com/user-attachments/files/25086024/Clover_Lane_Residence.-.Copy.ifc.txt)

Image

<br>
To test, it shows up in V0.8.1, however...
Image

### Attachments

_No response_

### Debug and Error Output

```true

```

theoryshaw:
fixes: #7639, #7800, #6231, #6808,

### DIFF
--- a/src/ifcopenshell-python/test/api/grid/test_create_axis_curve.py
+++ b/src/ifcopenshell-python/test/api/grid/test_create_axis_curve.py
@@ -1,0 +1,88 @@
+# IfcOpenShell - IFC toolkit and geometry engine
+# Copyright (C) 2021 Dion Moult <dion@thinkmoult.com>
+#
+# This file is part of IfcOpenShell.
+#
+# IfcOpenShell is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# IfcOpenShell is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with IfcOpenShell.  If not, see <http://www.gnu.org/licenses/>.
+
+import numpy as np
+
+import ifcopenshell.api.grid
+import test.bootstrap
+
+
+class TestCreateAxisCurve(test.bootstrap.IFC4):
+    def make_grid_with_axis(self, axis_tag="A"):
+        grid = self.file.createIfcGrid()
+        grid.ObjectPlacement = self.file.createIfcLocalPlacement(
+            RelativePlacement=self.file.createIfcAxis2Placement3D(
+                Location=self.file.createIfcCartesianPoint([0.0, 0.0, 0.0])
+            )
+        )
+        axis = ifcopenshell.api.grid.create_grid_axis(
+            self.file, axis_tag=axis_tag, same_sense=True, uvw_axes="UAxes", grid=grid
+        )
+        return grid, axis
+
+    def test_creates_a_polyline_axis_curve(self):
+        _, axis = self.make_grid_with_axis()
+        ifcopenshell.api.grid.create_axis_curve(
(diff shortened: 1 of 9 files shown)


BIMvoice:
Verified this independently: the fix genuinely resolves the duplicate-of-duplicate bug, confirmed live before/after on the sample file from the issue.

One gap worth flagging though: the new Gherkin test scenario added in this PR doesn't actually exercise the fix. It goes through a duplicate path that never creates a new IFC entity, so it would pass even without the fix applied. Might be worth adding a scenario that duplicates a grid that's already a duplicate, to actually cover the regression.

BIMvoice:
Correction to my earlier comment: I was wrong about the new scenario not exercising the fix. I had been reading `src/bonsai/test/bim/bootstrap.py`, which is dead code that nothing imports; the live harness is `test_feature.py`, where "I duplicate the selected objects" correctly calls `bim.override_object_duplicate_move()`. The scenario is sound.

Re-verified on current v0.8.0: the bug still reproduces (a chain of 4 duplicates leaves all 4 axes sharing one IfcPolyline, only the last keeps its location), and this branch fixes it (each duplicate gets its own curve, all 4 positions survive save and reload). It also still merges cleanly onto v0.8.0 with no conflicts.

Sorry for the noise from that first review. #8809 removes the dead Gherkin runner that misled me, since 38 of its step definitions share names with live steps while behaving differently.

BIMvoice:
@theoryshaw we built and verified this PR as offered. Summary in case it helps it land:

**#7639 (imperial grid annotations) is genuinely fixed** — confirmed live on the reporter's own imperial file: all 9 plan-view drawings went from 0 grid annotations to 3, correctly labeled `01`/`02`/`A`. Preferring the already-unit-scaled Blender representation is the right fix. Metric projects are a clean no-op (verified 6/6 unchanged).

**#7800 / #6808** (4X3 grid axis lock/unlock): the `prop.py` fix is correct. `IfcGridAxis` isn't an `IfcPositioningElement` in IFC4X3, so the toggle was skipping axes entirely.

**#6231** (duplicate-of-duplicate grid): the `geometry.py` + `create_axis_curve.py` pair correctly gives each duplicated axis its own curve.

Two small things before merge: (1) it's ~162 commits behind v0.8.0 now, so a rebase would be worth it (no conflicts apparent in the touched files); (2) the PR's own new test `test_replaces_existing_curve_when_unshared` asserts `by_id(old_id) is None`, but `by_id` raises `RuntimeError` for a missing entity, so that assertion should be `pytest.raises(RuntimeError)`. The fix logic itself is sound; that's just the test.

Happy to see this merged once rebased.

src/ifcopenshell-python/test/api/grid/test_create_axis_curve.py
```diff
@@ -0,0 +1,88 @@
...
+        _, axis = self.make_grid_with_axis()
+        ifcopenshell.api.grid.create_axis_curve(
+            self.file, p1=np.array([0.0, 0.0, 0.0]), p2=np.array([10.0, 0.0, 0.0]), grid_axis=axis
+        )
+        old_curve_id = axis.AxisCurve.id()
+        ifcopenshell.api.grid.create_axis_curve(
+            self.file, p1=np.array([1.0, 0.0, 0.0]), p2=np.array([11.0, 0.0, 0.0]), grid_axis=axis
+        )
+        assert axis.AxisCurve.id() != old_curve_id
+        assert self.file.by_id(old_curve_id) is None
```
BIMvoice, line 59: @theoryshaw following up on the earlier note with a ready-to-apply fix. `file.by_id` raises `RuntimeError` for a missing id (it never returns `None`), so this assertion is what's producing the two `RuntimeError: Instance #8 not found` failures in this PR's compile-and-test CI. The error actually proves the removal worked, only the assertion style is off. With this change the file passes 6/6 locally. The remaining CI failures (test_element, test_schema, rocksdb streaming, wall_opening, validate_stub) are all pre-existing v0.8.0 fixes your branch predates, so a plain rebase clears them, and the branch still merges cleanly.

```suggestion
        with pytest.raises(RuntimeError):
            self.file.by_id(old_curve_id)
```
(plus `import pytest` at the top)

We also re-verified today on latest v0.8.0: in IFC4X3 the panel unlock still leaves axes locked before this PR and unlocks them after, and a moved axis survives save and reload. This comment was written with AI assistance.

PR #7851, opened 2026-03-24
Fixed missing definition resulting in USD support not being built

rares-radu:
This commit fixes the issue where enabling `USD_SUPPORT` does not add the `WITH_USD` definition, resulting in USD functionality not being built.


cmake/CMakeLists.txt
```diff
@@ -234,6 +234,7 @@ set(USD_LIBRARIES "")
 if(USD_SUPPORT)
     find_package(USD REQUIRED)
     set(USD_LIBRARIES pxr::USD)
+    add_definitions(-DWITH_USD)
```
aothms, line 237:
```suggestion
    add_definitions(-DWITH_USD)
    set(SWIG_DEFINES ${SWIG_DEFINES} -DWITH_USD)
```

I think it makes more sense to have it here. Is there a reason why your change is scattered over two files? May thanks!

rares-radu: I saw that that's how it's done for OpenCOLLADA, so I kept the style, but now that I took a closer look, the others add the definitions as you've mentioned.

PR #7858, opened 2026-03-27
Fix #6602: Print IfcSpaces in projection.

### LINKED ISSUE #6602: View IfcSpaces in projection.
### Feature Description

Would be nice to view IfcSpaces in projection. 
Would allow for these types of drawings: https://community.osarch.org/discussion/comment/24594/#Comment_24594

theoryshaw:
Fix #6602: Print IfcSpaces in projection.

### Summary

Previously, `IfcSpace` elements were excluded from the standard drawing element pipeline and rendered through a separate ad-hoc loop in `generate_linework()`. This loop only handled spaces that intersected the cut plane (cut view), leaving spaces entirely invisible in projection — a significant limitation for any drawing where a space sits outside the cut range.

This PR removes that special case entirely, integrating `IfcSpace` into the same pipeline as all other elements.

### Changes

-   **`tool/drawing.py`  —  `get_drawing_elements()`**: Removed the explicit  `IfcSpace`  exclusion filter. Also fixed the initial element set (used for linked file processing) to include  `IfcSpatialElement`  alongside  `IfcElement`, so spaces are not inadvertently dropped before the intersection step.
-   **`tool/drawing.py`  —  `activate_drawing()`**: Removed the separate  `get_drawing_spaces()`  call; spaces now come through  `get_drawing_elements()`  like everything else.
-   **`operator.py`  —  `generate_linework()`**: Added a guard to create an empty  `<g>`  element when the OpenCASCADE serializer produces no group (which occurs in BISECT cut mode when the serializer output contains only  `<defs>`). Without this, the function returned early before any space geometry could be added.
-   **`operator.py`  —  `generate_bisect_linework()`**: Extended the visible-object loop to handle  `IfcSpatialElement`  types that do not intersect the cut plane. These are rendered by projecting mesh edges onto the camera plane. Back-face culling (via face normal dot product against the camera direction) is applied so only visible faces contribute edges. A coplanarity threshold (`dot > 0.99`) suppresses triangulation edges within flat faces, producing clean output that matches the HLR edge style of regular projected elements.
-   **`default.css`  /  `sample.css`**: Added  `.IfcSpace.projection`  two-class selector, which has higher CSS specificity than the existing  `.IfcSpace { fill: none; stroke: none; }`  rule and makes projected spaces visible with a standard stroke.

### Why this matters

The old approach treated `IfcSpace` as a special drawing concern, hard-coding the assumption that spaces only matter at the cut plane. In practice, users compose drawings at many scales and section heights — a reflected ceiling plan, a high-level section, or an elevation may need to show spaces that are entirely above or below the cut. Under the old code, those spaces were simply absent from the SVG output with no way for the user to include them.

By routing spaces through the standard element pipeline, users gain the same compositional freedom they have with any other element: spaces respond to `EPset_Drawing` include/exclude filters, appear correctly in all target views, and can be styled through CSS like any other projected element. This aligns `IfcSpace` behaviour with the broader drawing system rather than maintaining a separate code path that had to be manually kept in sync.

### DIFF
--- a/src/bonsai/bonsai/bim/data/assets/default.css
+++ b/src/bonsai/bonsai/bim/data/assets/default.css
@@ -78,6 +78,7 @@ text.GRID, tspan.GRID { /* 5mm */ font-size: 8.25px; }
 .material-sand { fill: url(#sand); }
 .material-concrete { fill: url(#concrete); stroke-width: 0.5; }
 .IfcSpace { fill: none; stroke: none; }
+.IfcSpace.projection { fill: white; stroke: black; stroke-linecap: round; stroke-width: 0.25; }
 .PredefinedType-STUD { stroke: black; stroke-width: 1; }
 .PredefinedType-WOOD { fill: url(#wood); stroke: black; stroke-width: 0.5; }
 .PredefinedType-STEEL { fill: url(#steel); stroke: black; stroke-width: 0.5; }
--- a/src/bonsai/bonsai/bim/data/assets/sample.css
+++ b/src/bonsai/bonsai/bim/data/assets/sample.css
@@ -28,3 +28,4 @@
 .break { fill: white; }
 .breakline { fill: none; stroke: black; stroke-linecap: 'round'; stroke-width: 0.25; marker-mid: url(#breakline-marker); }
 .IfcSpace { fill: none; stroke: none; }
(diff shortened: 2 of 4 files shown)


theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/501aeca300d88564931c7757a7b8b240531093b6/logs/summaries/2026-07-10-rebase-print_ifcspaces_in_projection.md

PR #7860, opened 2026-03-27
ifcquery, ifcedit, ifcmcp: add Makefiles and PyPI publish workflows

brunopostle:
[@aothms @Moult This would add new packages to the [IfcOpenShell](https://pypi.org/org/IfcOpenShell/) project on pypi, so I won't do this without your approval]

These three packages were added to src/ but lacked the Makefile needed by common.mk to build distribution wheels, and the GitHub Actions workflow to publish them to PyPI.

Adds make dist / make test / make qa targets and ci-*-pypi.yaml workflows matching the pattern used by ifcpatch, ifcclash, etc.


Moult:
Looks good to me. Happy to merge if @aothms is OK with it too. Will probably also need adding to docs and readme.

brunopostle:
OK, have updated the docs and added .rst files for the three tools

aothms:
The `ifcmcp` workflow failed with a 400.

```
Uploading ifcmcp-0.8.5-py3-none-any.whl
INFO     Response from https://upload.pypi.org/legacy/:                         
         400 Bad Request                                                        
INFO     <html>                                                                 
          <head>                                                                
           <title>400 The name 'ifcmcp' is too similar to an existing project.  
         See https://pypi.org/help/#project-name for more information.</title>  
          </head>                                                               
          <body>                                                                
           <h1>400 The name 'ifcmcp' is too similar to an existing project. See 
         https://pypi.org/help/#project-name for more information.</h1>         
           The server could not comply with the request since it is either      
         malformed or otherwise incorrect.<br/><br/>                            
         The name &#x27;ifcmcp&#x27; is too similar to an existing project. See 
         https://pypi.org/help/#project-name for more information.              
                                                                                
                                                                                
          </body>                                                               
         </html>                                                                
```

https://pypi.org/search/?q=ifcmcp

https://pypi.org/search/?q=ifc-mcp

What do we do? Reach out to pypa saying we are OG? Rename to ifcopenshell-mcp or ifc-model-context-protocol?

brunopostle:
I did a search for the exact name, but there is now an ifcMCP: https://pypi.org/project/iflow-mcp_smartaec-ifcmcp/

aothms:
Their naming rules are not really documented, but I think the more likely collision is https://pypi.org/project/ifc-mcp/

aothms:
Went with https://pypi.org/project/ifcopenshell-mcp/ in case of better ideas do let me know :)

PR #7882, opened 2026-04-01
Fix IfcSurfaceStyleRendering colour reset on save

falken10vdl:
## Problem

When an `IfcSurfaceStyleRendering` had `DiffuseColour`, `SpecularColour`, or `SpecularHighlight` set to `null` in the IFC file, loading the style into the Bonsai panel set the corresponding `is_*_null` flag to `True`. Editing the colour picker did not clear the flag. On save, `get_shading_style_data_from_props` read the flag and wrote `None` back, discarding the user's change and resetting the material to its default colour.

1. **`tool/style.py` — `get_shading_style_data_from_props`**: before reading a colour prop, check the corresponding null flag. If it is set, write `None` and skip. This makes live-preview and save agree on null vs. non-null.

2. **`bim/module/style/prop.py`**: a `_make_clear_null_updater` factory produces named callbacks that clear the null flag via dict access (`self[is_..._null] = False`) before delegating to `update_shader_graph`. The three aliases (`update_diffuse_colour`, `update_specular_colour`, `update_specular_highlight_value`) are wired to all colour/ratio/class pickers. The null `BoolProperty` toggles are also wired to `update_shader_graph` so toggling them triggers an immediate preview rebuild.

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/module/style/prop.py
+++ b/src/bonsai/bonsai/bim/module/style/prop.py
@@ -118,6 +118,19 @@ def update_shader_graph(self: Union["Texture", "BIMStylesProperties"], context: 
     tool.Loader.create_surface_style_with_textures(material, shading_data, textures_data)
 
 
+def _make_clear_null_updater(null_prop: str):
+    def _update(self: "BIMStylesProperties", context: bpy.types.Context) -> None:
+        self[null_prop] = False
+        update_shader_graph(self, context)
+
+    return _update
+
+
+update_diffuse_colour = _make_clear_null_updater("is_diffuse_colour_null")
+update_specular_colour = _make_clear_null_updater("is_specular_colour_null")
+update_specular_highlight_value = _make_clear_null_updater("is_specular_highlight_null")
+
+
 UV_MODES = [
     ("UV", "UV", _("Actual UV data presented on the geometry")),
     ("Generated", "Generated", _("Automatically-generated UV from the vertex positions of the mesh")),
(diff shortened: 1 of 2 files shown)


theoryshaw:
Nice!  I ran into this quite a bit.  It's funny how some bugs you just live with, without reporting. :)

falken10vdl:
@Andrej730 Are you happy with this fix?
Moreover I am planning to propose some other enhancements to styles based on extracting elements of https://github.com/IfcOpenShell/IfcOpenShell/pull/7938 and also adding some other as discussed in https://community.osarch.org/discussion/3409/surface-styles-revisited
Namely:


* [Fix IfcSurfaceStyleRendering colour reset on save#7882](https://github.com/IfcOpenShell/IfcOpenShell/pull/7882 "https://github.com/IfcOpenShell/IfcOpenShell/pull/7882"). This is this PR that was driving me crazy when looking to styles. It is a fix.
* New PR to establish a precedence of Blender representation based on what styles are available (this can be extracted from  https://github.com/IfcOpenShell/IfcOpenShell/pull/7938 ). Currently if you have a "Shade" style but no "Render" style, Blender will default to "white" if your Viewport Shading is set to Material_Preview or Rendered. The idea is that in this case it uses principled BSDF from "Shade" so in the end we have as proposed above but using the already used terminology in Bonsai: 
    * "Shading (maybe this one could be renamed Shade/Render or Flat)" (Precedence chain: Render → Shade)
        Solid: Flat colour (color_type=MATERIAL)
        Material Preview/Rendered: IFC node graph (Render, or Shade colour)
    * "External (maybe this one could be renamed Texture/External or Pretty)" (Precedence chain: External → Render+Texture → Render → Shade):
        Solid: UV texture (color_type=TEXTURE, UVs on demand)
        Material Preview/Rendered: External .blend → IFC node graph fallback

  The bpy.ops.bim.toggle_prefer_ifc_shading operator would be removed and we would use the existing bpy.ops.bim.update_current_style operator to switch between the Shade/Render (i.e. Flat) and the Texture/External (i.e. the Pretty)
* New PR To generate Shade base color from external (this can be extracted from  https://github.com/IfcOpenShell/IfcOpenShell/pull/7938 )
* New PR to generate Render+Texture from external (this one not yet implemented)


Cheers!

falken10vdl:
@Moult I have looked to "blame" in the code around the fix I am proposing and found you. Are you happy with the proposed fix? 
Thanks!

Moult:
Looks great! I'm personally not a fan of these wrapper functions because they move the codebase towards magic but here it's pretty docile :)

falken10vdl:
Yes I am also not so much of a fan of it but it looks more pythonic I guess :)
I will issue some more incremental PRs as discussed in https://community.osarch.org/discussion/3409/surface-styles-revisited with each of the addressed topics
Thanks!

PR #7893, opened 2026-04-02
Fix TypeError in ray_cast_by_proximity_2d degenerate edge

theoryshaw:
A degenerate edge (zero-length segment) caused an early `return` of a tuple instead of continuing the loop, resulting in a TypeError when snap.py iterated the result and tried to assign `point["group"]` on a float.

### DIFF
--- a/src/bonsai/bonsai/tool/raycast.py
+++ b/src/bonsai/bonsai/tool/raycast.py
@@ -434,11 +434,8 @@ class Raycast(bonsai.core.tool.Raycast):
             seg_len_sq = sx * sx + sy * sy
 
             if seg_len_sq == 0.0:
-                # degenerate segment: return distance to p0
-                dx = px - p0x
-                dy = py - p0y
-                dist = math.hypot(dx, dy)
-                return dist, (p0x, p0y), 0.0
+                # degenerate segment: skip it
+                continue
 
             # project (p - p0) onto seg: t = dot(p-p0, seg) / |seg|^2
             apx = px - p0x


theoryshaw:
## Fix  `TypeError`  crash in snapping when mesh contains degenerate edges

### Summary

Fixes a crash that occurs when using polyline drawing tools (e.g. in plan view) over mesh objects that contain degenerate edges — edges where both vertices share the same position, giving a zero-length segment.

### Root Cause

In `tool/raycast.py`, `ray_cast_by_proximity_2d` iterates over candidate edges to find snap points. For degenerate edges (`seg_len_sq == 0.0`), the code had an early exit that returned a raw tuple `(dist, (p0x, p0y), 0.0)` from inside the `for e in edges:` loop — short-circuiting the rest of the function and returning a tuple instead of the expected `list[dict]`.

Back in `tool/snap.py`, `detect_snapping_points` receives the result and iterates with:

```python
for point in snap_points:
    point["group"] = "Object"

```

When iterating over `(dist, (p0x, p0y), 0.0)`, the first `point` is the bare `float` value `dist`. Attempting `point["group"] = "Object"` on a float raises:

```
TypeError: 'float' object does not support item assignment

```

This crash propagates up through `handle_mouse_move` and kills the modal operator, breaking the polyline tool entirely for any scene containing degenerate geometry.

### Fix

Replace the erroneous `return` with `continue`, skipping the zero-length edge and allowing the loop to process remaining edges normally. The function then returns its accumulated `points` list as expected.

### Testing

Reproduce by entering a polyline drawing tool and hovering over a mesh object with a degenerate edge (two vertices at identical coordinates). Before the fix the tool crashes immediately; after the fix snapping works correctly and the degenerate edge is silently skipped.

----------

_This PR was generated with the assistance of an AI coding tool._

```
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\project\operator.py", line 3012, in modal
    self.handle_mouse_move(context, event)
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\model\polyline.py", line 411, in handle_mouse_move
    detected_snaps = tool.Snap.detect_snapping_points(context, event, self.objs_2d_bbox, self.tool_state)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\tool\snap.py", line 405, in detect_snapping_points
    point["group"] = "Object"
    ~~~~~^^^^^^^^^
TypeError: 'float' object does not support item assignment
```

theoryshaw:
Hi, @brunoperdigao Can you take a look? :)

PR #7894, opened 2026-04-02
Feat: Join coplanar SVG projection cells using normal vector evaluation

### LINKED ISSUE #3742: Join projection planes that are coplanar.

image

dipayansardar73-decode:
Resolves #3742

This PR addresses the issue where `IfcOpenShell` SVG outputs failed to join cells forming coplanar faces if they were angled relative to the projection.

The previous logic inside `draw.py` relied on identical `ray_distance` (depth). This fails on any surface that is non-orthogonal to the view frustum.

**Implementation Details:**
1. Updated `format(x)` to extract `x.position` and `x.normal` instead of ray distance from the BRep queries.
2. Replaced the scalar depth check with point-to-plane vector operations:
   - Verify parallel alignment using the dot product of both normals: `numpy.dot(n1, n2)`
   - Verify geometric coplanarity by taking the vector between both trace points against the normal: `numpy.dot((p1 - p2), n1)`

This ensures cells properly merge regardless of arbitrary projection angles.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/draw.py
+++ b/src/ifcopenshell-python/ifcopenshell/draw.py
@@ -445,8 +445,8 @@ def main(
                             # Check if styles match, normals are parallel and points are coplanar
                             # Issue #3742: Replaced simplistic distance check with vector mathematics and material check
                             if s1 == s2:
-                                if abs(1.0 - numpy.dot(n1, n2)) < 1.0e-5:
-                                    if abs(numpy.dot(p1 - p2, n1)) < 1.0e-5:
+                                if abs(1.0 - abs(numpy.dot(n1, n2))) < 1.0e-4:
+                                    if abs(numpy.dot(p1 - p2, n1)) < 1.0e-4:
                                         
                                         def get_cached_material(inst):
                                             id_ = inst.id()
--- a/src/ifcopenshell-python/ifcopenshell/draw.py
+++ b/src/ifcopenshell-python/ifcopenshell/draw.py
@@ -426,14 +426,26 @@ def main(
                             # found to be inside element using tree.select() no face or style info
                             return x
                         else:
-                            return (x.instance.is_a(), x.ray_distance, tuple(x.position))
+                            return (x.instance.is_a(), tuple(x.position), tuple(x.normal))
 
                     pp = pairs[he_idx : he_idx + 2]
                     if pp == (-1, -1):
                         continue
                     data = list(map(format, map(semantics.__getitem__, pp)))
-                    if None not in data and data[0][0] == data[1][0] and abs(data[0][1] - data[1][1]) < 1.0e-5:
-                        to_remove.append(he_idx // 2)
+                    if None not in data and data[0][0] == data[1][0]:
+                        if len(data[0]) == 2 and len(data[1]) == 2:
+                            # Both from tree.select() -> (instance, -1)
+                            to_remove.append(he_idx // 2)
+                        elif len(data[0]) == 3 and len(data[1]) == 3:
+                            # Both from tree.select_ray() -> (type, position, normal)
+                            p1, n1 = numpy.array(data[0][1]), numpy.array(data[0][2])
+                            p2, n2 = numpy.array(data[1][1]), numpy.array(data[1][2])
+                            
+                            # Check if normals are parallel and points are coplanar
+                            # Issue #3742: Replaced simplistic distance check with vector mathematics
+                            if abs(1.0 - numpy.dot(n1, n2)) < 1.0e-5:
+                                if abs(numpy.dot(p1 - p2, n1)) < 1.0e-5:
+                                    to_remove.append(he_idx // 2)
                         # Print edge index and semantic data
                         # print(he_idx // 2, *data)
 
--- a/src/ifcopenshell-python/ifcopenshell/draw.py
+++ b/src/ifcopenshell-python/ifcopenshell/draw.py
@@ -414,6 +414,7 @@ def main(
             if iteration != num_passes:
                 to_remove = []
 
+                material_cache = {}
                 for he_idx in range(0, len(pairs), 2):
                     # @todo instead of ray_distance, better do (x.point - y.point).dot(x.normal)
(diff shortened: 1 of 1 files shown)


theoryshaw:
@dipayansardar73-decode This is awesome, thank you. 
Would it be possible to only join these coplanar surfaces if they share the same material?  
If not, they should be separated.

dipayansardar73-decode:
> @dipayansardar73-decode This is awesome, thank you. Would it be possible to only join these coplanar surfaces if they share the same material? If not, they should be separated.

@theoryshaw Yes, absolutely! That's a great point. I noticed the original logic only checked the instance class types but skipped over the materials entirely. 
Since the underlying BRep query already exposes the `style_index` for the intersecting faces, it's very straightforward to add. I'll update the mathematical check to ensure that faces only merge if both their normals are perfectly coplanar and they share the exact same `style_index`.
I'll push the updated commit to this PR right away!

dipayansardar73-decode:
@theoryshaw Just pushed the update! It now checks `style_index` before merging. Does the logic in that commit look good to you, or is there another material attribute you'd prefer I check instead?

theoryshaw:
I just tested and i couldn't get this to work. 

I created some test files we can commonly use to test: 
https://hub.openingdesign.com/OSArch/Community_Troubleshooting/src/branch/main/7894_Join%20coplanar%20SVG%20projection%20cells%20using%20normal%20vector%20evaluation


I think both materials and styles should play into when edges merge. 

I'm thinking the following logic would be a good approach: 
See red 'X' for conditions they should join. 
If there's (2) different materials with the same style, i don't think they should join. 
see the drawing called `NORTH SECTION` in the `coplanar join.ifc` file.
image

image

dipayansardar73-decode:
> I just tested and i couldn't get this to work.
> 
> I created some test files we can commonly use to test: https://hub.openingdesign.com/OSArch/Community_Troubleshooting/src/branch/main/7894_Join%20coplanar%20SVG%20projection%20cells%20using%20normal%20vector%20evaluation
 

@theoryshaw  Ah, that makes perfect sense. Thank you for the detailed diagram and the test file. You are absolutely right—relying on the visual style_index alone is a drafting hazard. If two adjacent walls are made of Concrete and Plaster respectively, their physical material boundary line must remain visible on the drawing, even if an architect accidentally painted them both with the same grey visual style. 

I have just pushed an extensive update to this PR to implement exactly the logic you asked for.

dipayansardar73-decode:
@theoryshaw 

**Implementation Details**: I upgraded the pipeline inside draw.py to directly fetch and enforce real-world physical material boundaries during the SVG path merge.

1. Modified the format(x) BRep payload to push the actual x.instance (the live IFC object) down into the vector looping engine.
2. Implemented ifcopenshell.util.element.get_material(inst) to cross-reference the exact physical IfcMaterial driving both intersecting polygons.
3. Added a highly optimized material_cache dictionary locally inside the sweep loop so we aren't repeatedly taxing the IFC graph by fetching materials for thousands of connected edges.
Now, the engine strictly obeys your matrix rule. If the faces are perfectly coplanar and share the exact same style, it performs a final check on mat1.id() == mat2.id(). If they represent different structural materials, it aggressively preserves the separation line.

Could you re-run your coplanar join.ifc test file against this latest commit? It should map to your Red "X" Diagram completely now.

theoryshaw:
Hm, i couldn't get it to work. 

image

image

dipayansardar73-decode:
> Hm, i couldn't get it to work.

@theoryshaw 
Thanks for taking a look at this. I ’ve spent a solid few hours today digging into why this isn't working as expected in the viewport, and I think I’ve found the root of the problem.

The issue isn't in the coplanar joining logic itself, but in how the drawing engine handles the material-to-style mapping. Currently, different materials (like your Concrete and Brick examples) are colliding and falling back to a single generic style because the CSS class names aren't being generated uniquely or accurately for each material.

dipayansardar73-decode:
@theoryshaw 
I have successfully patched this locally with the following changes:

1. Updated ifcopenshell.draw: To extract and sanitize material names into unique .mat-[MaterialName] and .material-[MaterialName] classes (preserving CamelCase for accuracy).
2. Updated operator.py: To ensure these classes are actually injected into the viewport linework groups (generate_material_layers).
3. Fallback Protection: Added a .no-material class to separate elements with no assignments from those that should have unique styles.

dipayansardar73-decode:
@theoryshaw 
The reason it still looks grey in your screenshot is that the default default.css assets are lowercase (e.g., .material-concrete) while the engine is now producing CamelCase (e.g., .mat-Concrete). Since CSS is case-sensitive, they don't match.

I’ve been working on this for hours across both the Python engine and the Blender addon layers to get the mapping perfect. I'd love to hear your thoughts on standardizing these material classes so i can finally fix this 'Same Style' collision for everyone.

theoryshaw:
Hi @dipayansardar73-decode 
I started vibe coding to help me understand your approach, as well as try to answer your questions... but I ended up coming up with another approach: https://github.com/IfcOpenShell/IfcOpenShell/pull/7908

Feel free to comment and review. 

I would defer to an core maintainer on the ultimate pull.

dipayansardar73-decode:
@theoryshaw 
I have been testing your PR locally and it's a massive improvement. Moving the merging to the SVG layer finally lets us join non-wall elements like actuators, which were previously skipped by the hardcoded checks. Your 3D adjacency check handles the coplanar cases perfectly. 
I noticed the CI is failing on linting, but the code itself is rock solid. Thanks for the effort. Great Work !

theoryshaw:
Thank Claude. :)
Thank you for reviewing!

dipayansardar73-decode:
> Thank Claude. :) Thank you for reviewing!

Haha! Claude is blushing 😂
Welcome :)

theoryshaw:
@dipayansardar73-decode okay to close this PR?  If you have some other intent, please open again.

dipayansardar73-decode:
> @dipayansardar73-decode okay to close this PR? If you have some other intent, please open again.

@theoryshaw  Yes yes, that's perfectly fine 👍

PR #7898, opened 2026-04-03
Fix IfcOpeningElement using wrong geometry context

theoryshaw:
When creating an IfcOpeningElement via 
image
, the context was read from props.contexts which could accidentally or intentionally through previous operation be set to IfcGeometricRepresentationContext ("Model") rather than the Body subcontext ("Model/Body/MODEL_VIEW"). The boolean engine only searches subcontexts for opening geometry, so the void was silently never applied.

Fix: always resolve Model/Body/MODEL_VIEW directly from the IFC file when the class is IfcOpeningElement, bypassing the context dropdown entirely.

Also add warnings in AddOpening when the opening has no Body representation, and when switch_representation produces no IfcBooleanResult after the opening is applied.

### DIFF
--- a/src/bonsai/bonsai/bim/module/root/operator.py
+++ b/src/bonsai/bonsai/bim/module/root/operator.py
@@ -27,6 +27,7 @@ import ifcopenshell.api.material
 import ifcopenshell.api.pset
 import ifcopenshell.api.root
 import ifcopenshell.util.element
+import ifcopenshell.util.representation
 import ifcopenshell.util.schema
 import ifcopenshell.util.shape_builder
 import ifcopenshell.util.type
@@ -530,7 +531,13 @@ class AddElement(bpy.types.Operator, tool.Ifc.Operator):
             return self.report({"WARNING"}, "A featured element must be nominated.")
 
         ifc_context = None
-        if get_enum_items(props, "contexts", context):
+        if props.ifc_class == "IfcOpeningElement":
+            ifc_context = ifcopenshell.util.representation.get_context(tool.Ifc.get(), "Model", "Body", "MODEL_VIEW")
+            if ifc_context is None:
+                ifc_context = ifcopenshell.util.representation.get_context(tool.Ifc.get(), "Model", "Body")
+            if ifc_context is None:
+                self.report({"WARNING"}, "No Model/Body context found. Opening representation may be on the wrong context.")
+        elif get_enum_items(props, "contexts", context):
             ifc_context = int(props.contexts or "0") or None
             if ifc_context:
                 ifc_context = tool.Ifc.get().by_id(ifc_context)
--- a/src/bonsai/bonsai/bim/module/void/operator.py
+++ b/src/bonsai/bonsai/bim/module/void/operator.py
@@ -156,6 +156,14 @@ class AddOpening(bpy.types.Operator, tool.Ifc.Operator):
                     should_add_representation=True,
                     context=body_context,
                 )
+            if element2:
+                opening_body_rep = ifcopenshell.util.representation.get_representation(element2, "Model", "Body")
+                if opening_body_rep is None:
+                    self.report(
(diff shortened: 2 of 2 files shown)


theoryshaw:
## Problem

When an `IfcOpeningElement` is created via **Add Element** 
(`bpy.ops.bim.add_element(ifc_product="IfcFeatureElement", 
ifc_class="IfcOpeningElement", ...)`), its geometry representation 
can end up assigned to the root `IfcGeometricRepresentationContext` 
(e.g. `ContextType='Model'`, no `ContextIdentifier`) instead of the 
correct `IfcGeometricRepresentationSubContext` 
(`ContextIdentifier='Body'`).

This happens because `AddElement._execute` reads `ifc_context` from 
`props.contexts` — a dropdown that lists root contexts **before** 
subcontexts. When `skip_dialog=True` is used (bypassing the UI), 
whatever context was last selected (or the dropdown default) is used 
without validation. A user creating an opening interactively could 
also accidentally pick `"Model"` from the top of the list instead of 
`"Model/Body/MODEL_VIEW"`.

The result is an `IfcOpeningElement` that looks correct in Blender 
(its mesh is visible) and has a valid `IfcRelVoidsElement` 
relationship to the target element, but **never cuts a void**. No 
error is raised. The boolean subtraction silently fails because 
`ifcopenshell.util.representation.get_representation(opening, 
"Model", "Body")` returns `None` — it only searches 
`IfcGeometricRepresentationSubContext` instances, so the 
representation on the root context is invisible to it.

This bug persists across sessions: the wrong context is baked into 
the saved IFC file. On reload, Bonsai faithfully imports the opening 
with its incorrect context, and subsequent calls to 
`bpy.ops.bim.add_opening` continue to silently fail.

## Root Cause

`AddElement._execute` ([root/operator.py](src/bonsai/bonsai/bim/module/root/operator.py)):

```python
ifc_context = None
if get_enum_items(props, "contexts", context):
    ifc_context = int(props.contexts or "0") or None
    if ifc_context:
        ifc_context = tool.Ifc.get().by_id(ifc_context)

```

`contexts()` in `root/data.py` populates the dropdown with root contexts first, subcontexts second. For `IfcOpeningElement`, there is only one correct choice — `Model/Body/MODEL_VIEW` — but nothing enforced this, and the wrong context produced no error.

## Fix

### `root/operator.py`  — Force Body subcontext for IfcOpeningElement

When `ifc_class == "IfcOpeningElement"`, bypass `props.contexts` entirely and resolve the context directly from the IFC file using `get_context`. A two-level fallback handles files that define the Body subcontext without a `TargetView`:

1.  Try  `Model / Body / MODEL_VIEW`  (most specific)
2.  Fall back to  `Model / Body`  (no target view)
3.  Warn if neither exists

### `void/operator.py`  — Warn when opening has no Body representation

`AddOpening._execute` now validates the opening's representation before calling `feature.add_feature`. If no Body representation is found, a `WARNING` is reported in the Blender header bar explaining that the void will not be cut and what to check in the IFC file.

A second warning fires after `switch_representation` if the voided element's updated representation contains no `IfcBooleanResult` — catching cases where the relationship was created correctly but the geometry engine still did not apply the cut (e.g. non-intersecting geometry).

## Impact

-   Existing IFC files with openings already on the wrong context are unaffected by this fix (the saved data is not modified), but users will now receive a clear warning when they attempt to apply them, rather than a silent no-op.
-   All other element types continue to use the context dropdown as before.
-   `IfcOpeningElement`  creation via both the UI dialog and  `skip_dialog=True`  now always produces a correctly-contexted representation.

🤖 Generated with the assistance of an AI coding tool.

theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/25d34fc7ae945b6214054223dd57fda65004454f/logs/summaries/2026-07-10-rebase-fix-opening-element-body-context.md

PR #7899, opened 2026-04-03
Center IfcOpeningElement bbox on 3D cursor

theoryshaw:
When placing an IfcOpeningElement via add_element, the extruded void is now centered on the 3D cursor rather than starting at it, so the cursor sits at the depth midpoint of the opening.

### DIFF
--- a/src/bonsai/bonsai/bim/module/root/operator.py
+++ b/src/bonsai/bonsai/bim/module/root/operator.py
@@ -622,14 +622,16 @@ class AddElement(bpy.types.Operator, tool.Ifc.Operator):
                     elif usage.LayerSetDirection == "AXIS3":
                         z_axis = tuple(local_z) if direction_sense == "POSITIVE" else tuple(-local_z)
 
+                    half_mag = 0.25 / unit_scale
                     item = builder.extrude(
                         profile,
                         magnitude=0.5 / unit_scale,
+                        position=tuple(-Vector(z_axis) * half_mag),
                         position_x_axis=tuple(local_x),
                         position_z_axis=z_axis,
                     )
                 else:
-                    item = builder.extrude(curve, magnitude=0.5 / unit_scale)
+                    item = builder.extrude(curve, magnitude=0.5 / unit_scale, position=(0.0, 0.0, -0.25 / unit_scale))
 
             representation = builder.get_representation(ifc_context, [item])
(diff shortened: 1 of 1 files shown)


theoryshaw:
## Center IfcOpeningElement on 3D cursor at creation

### Problem

When adding an `IfcOpeningElement` via `bpy.ops.bim.add_element(ifc_product="IfcFeatureElement", ifc_class="IfcOpeningElement", skip_dialog=True)`, the newly created void's geometry is built as an extruded solid starting at the local origin `(0, 0, 0)` and extending outward by 0.5 m. Because the Blender object's origin is placed at the 3D cursor, this means one face of the void sits flush with the cursor — the entire depth of the opening extends away from it in one direction.

image


In practice this forces a manual move after every placement: the user positions the cursor at the face of a wall or slab, creates the opening, then has to slide it back so it overlaps sufficiently on both sides to produce a clean boolean cut.

### Root Cause

In `AddElement._execute` (operator.py), the `EXTRUSION` path for `IfcOpeningElement` calls `builder.extrude(curve, magnitude=0.5 / unit_scale)` with no position offset. The default `position=(0, 0, 0)` means the solid occupies `[0, magnitude]` along the extrusion axis in the object's local space. Since the IFC object placement is set to the 3D cursor location, the solid starts at the cursor rather than being centered on it.

Two code paths are affected:

-   **Non-layered elements**  (generic featured element, no  `IfcMaterialLayerSetUsage`): simple Z-axis extrusion.
-   **Layered elements**  (walls with  `IfcMaterialLayerSetUsage`): extrusion along a custom axis derived from the wall's local Y or Z, with explicit  `position_x_axis`  /  `position_z_axis`  overrides.

### Fix

For both paths, the extrusion start position is offset by half the magnitude in the negative extrusion direction, so the solid spans `[-0.25 m, +0.25 m]` around the local origin — and therefore around the 3D cursor.

**Non-layered case:**

```python
item = builder.extrude(curve, magnitude=0.5 / unit_scale, position=(0.0, 0.0, -0.25 / unit_scale))

```

**Layered (wall) case:**

```python
half_mag = 0.25 / unit_scale
item = builder.extrude(
    profile,
    magnitude=0.5 / unit_scale,
    position=tuple(-Vector(z_axis) * half_mag),
    position_x_axis=tuple(local_x),
    position_z_axis=z_axis,
)

```

The non-opening `EXTRUSION` path (walls, slabs, etc.) is unchanged.

PR #7897, opened 2026-04-03
feat: Add Anthropic Claude API support to ifcchat

DesertSpringsCivil:
## Summary
- Adds a **Provider** dropdown (OpenAI / Anthropic) to the ifcchat sidebar
- When Anthropic is selected, the UI updates the API key label, placeholder, and available models (claude-sonnet-4-6, claude-opus-4-6, claude-haiku-4-5)
- Implements `anthropicMessagesCreate()` calling Claude's Messages API (`/v1/messages`) with proper `tool_use` / `tool_result` format
- Converts existing OpenAI tool schemas to Anthropic format (`input_schema` instead of `parameters`)
- Existing OpenAI functionality is completely unchanged

## Context
Per discussion at https://community.osarch.org/discussion/3396/ — @theoryshaw requested Claude API key support. Tested locally against the Pyodide worker: loaded an IFC model, ran tool calls through Claude, received structured responses correctly.

## Test plan
- [ ] Select "Anthropic (Claude)" provider — verify label changes to "Anthropic API key" with `sk-ant-...` placeholder
- [ ] Verify model selector switches to Claude models (sonnet-4-6, opus-4-6, haiku-4-5)
- [ ] Enter Anthropic API key, load an IFC file, ask "Summarize the model"
- [ ] Verify tool-calling loop works (Claude calls ifc_summary, result fed back, response displayed)
- [ ] Switch back to OpenAI — verify original behavior unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7904, opened 2026-04-03
feat: Reduce ifcchat token usage and default to IFC4X3

DesertSpringsCivil:
## Summary
- **Anthropic prompt caching**: Mark system prompt and tool definitions with `cache_control` so they're served from cache on subsequent turns (~90% cost reduction on static tokens)
- **Tool result truncation**: Cap tool results at 2000 chars in conversation history — the full result is still shown in the UI, but only the truncated version is sent back to the LLM on subsequent turns
- **Conversation history sliding window**: Keep last 40 messages, trim at user message boundaries to avoid breaking mid-tool-call sequences
- **Default to IFC4X3**: "New IFC" button now creates IFC 4.3 models; `ifc_new` tool schema uses `enum` to constrain valid schema values and prevent errors like `"IFC4X3ADD2"`

## Context
Follow-up to #7897. When using Claude via Anthropic API, the rate limit (30k tokens/min on lower tiers) was being hit within a single prompt due to unbounded conversation history and large tool results being resent every turn. The prompt caching change is Anthropic-specific; all other changes benefit every provider.

## Test plan
- [ ] Anthropic: verify multi-turn conversation lasts longer before rate-limiting
- [ ] All providers: verify large tool results (e.g. `ifc_tree` on a big model) show full output in UI but don't bloat subsequent API calls
- [ ] Click "New IFC" — verify model is created with IFC4X3 schema
- [ ] Ask the LLM to create a new IFC model — verify it uses valid schema values (IFC4, IFC2X3, IFC4X3)
- [ ] OpenAI/Gemini/OpenRouter: verify no regressions

### DIFF
--- a/src/ifcchat/api_anthropic.js
+++ b/src/ifcchat/api_anthropic.js
@@ -146,15 +146,22 @@ function toChatCompletionResponse(response) {
 
 export async function chat({ apiKey, model, messages, tools }) {
     const request = splitSystemAndMessages(messages);
+    const anthropicTools = toAnthropicTools(tools);
+
+    // Mark the last tool with cache_control so the entire tool list is cached
+    if (anthropicTools.length > 0) {
+        anthropicTools[anthropicTools.length - 1].cache_control = { type: "ephemeral" };
+    }
+
     const body = {
         model,
         max_tokens: 4096,
         messages: request.messages,
-        tools: toAnthropicTools(tools),
+        tools: anthropicTools,
     };
 
     if (request.system) {
-        body.system = request.system;
+        body.system = [{ type: "text", text: request.system, cache_control: { type: "ephemeral" } }];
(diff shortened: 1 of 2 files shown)


aothms:
> Tool result truncation: Cap tool results at 2000 chars in conversation history — the full result is still shown in the UI, but only the truncated version is sent back to the LLM on subsequent turns

I'm not so sure on this one. This could really provide misleading data back on things like counts. We'd rather look for other ways to improve the terseness of the mcp outputs. I'll revert that one if you don't mind.

aothms:
- Maybe we can make these end-user settings so that it's a bit configurable and people can report what works and doesn't work
- Maybe there can be a message compaction step instead of just truncation to keep the last 40 messages alive.

PR #7908, opened 2026-04-04
* Fix #3742: Remove coplanar boundary lines between adjacent same-mat…

### LINKED ISSUE #3742: Join projection planes that are coplanar.

theoryshaw:
…erial elements in Bonsai SVG drawings

Adds `remove_coplanar_boundary_lines()` to operator.py (Bonsai uses this path, not draw.py's main()). After `merge_linework_and_add_metadata()` assigns material CSS classes, this post-processes the SVG to delete projection line segments that appear in two or more adjacent, coplanar elements with the same material and presentation style.

Key design decisions:
- Material identity: compared via sorted IFC material ID tuples from `get_materials()`, not CSS class names — avoids false matches between unrelated `material-null` elements.
- Presentation style identity: compared via IFC IfcPresentationStyle IDs from `StyledByItem` on geometry representation items — handles elements with no material but distinct visual styles.
- Physical adjacency: confirmed by a 3D shared-vertex test (tol=0.01 m) after a quick AABB guard, rejecting elements whose 2D projections overlap but sit at different depths.
- Coplanarity: determined by the dominant (largest-area) face normal of each Blender mesh object — area-weighted averages are unreliable for slabs whose equal top/bottom faces cancel out. Folded walls sharing an edge but meeting at an angle are correctly rejected (normal dot ≪ 1.0).

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -1668,7 +1668,7 @@ class CreateDrawing(bpy.types.Operator):
             # Quick AABB guard
             corners_a = [obj_a.matrix_world @ Vector(c) for c in obj_a.bound_box]
             corners_b = [obj_b.matrix_world @ Vector(c) for c in obj_b.bound_box]
-            _verbose = {"1eW6OACmXD1hWJLDeB6erW", "2x0$XNlFH6FxlvS7eYJttx"} == {guid_a, guid_b}
+            _verbose = {"3S7QeSPVn3EB3T9TkDUy1e", "0YyL6RtCL8axn6AG9ItFhs"} == {guid_a, guid_b}
             for axis in range(3):
                 min_a = min(c[axis] for c in corners_a)
                 max_a = max(c[axis] for c in corners_a)
@@ -1789,6 +1789,39 @@ class CreateDrawing(bpy.types.Operator):
                     pass
             return None
 
+        def segs_bbox(segs):
+            """Compute AABB (min_x, max_x, min_y, max_y) from a parsed segment list."""
+            if not segs:
+                return None
+            xs = [c for _, (p0, p1) in segs for c in (p0[0], p1[0])]
+            ys = [c for _, (p0, p1) in segs for c in (p0[1], p1[1])]
+            return (min(xs), max(xs), min(ys), max(ys))
+
+        def seg_on_boundary_of(seg, bbox):
+            """True if seg is axis-aligned, lies on a bbox face, and meaningfully overlaps it.
+
+            Used for unilateral matching: when one element's shared edge is not
+            explicitly drawn (it's implicit or on the interior), the other element's
+            segments that lie on that missing edge can still be detected and removed.
+            """
+            (x0, y0), (x1, y1) = seg
+            min_x, max_x, min_y, max_y = bbox
+            is_v = abs(x0 - x1) < TOL
+            is_h = abs(y0 - y1) < TOL
+            if is_v:
+                x_mid = (x0 + x1) / 2
+                if abs(x_mid - min_x) < TOL or abs(x_mid - max_x) < TOL:
+                    y_lo, y_hi = min(y0, y1), max(y0, y1)
+                    overlap = min(y_hi, max_y) - max(y_lo, min_y)
+                    return overlap > TOL
+            if is_h:
+                y_mid = (y0 + y1) / 2
+                if abs(y_mid - min_y) < TOL or abs(y_mid - max_y) < TOL:
+                    x_lo, x_hi = min(x0, x1), max(x0, x1)
+                    overlap = min(x_hi, max_x) - max(x_lo, min_x)
+                    return overlap > TOL
+            return False
+
         def lines_match(a, b):
             """Return True if segments a and b represent the same physical edge.
 
@@ -1883,7 +1916,7 @@ class CreateDrawing(bpy.types.Operator):
                 for j, (grp_j, mat_j, style_j, segs_j, guid_j) in enumerate(group_data):
                     if j <= i:
                         continue
-                    _target = {"1eW6OACmXD1hWJLDeB6erW", "2x0$XNlFH6FxlvS7eYJttx"} == {guid_i, guid_j}
+                    _target = {"0QpDRRuif0R80d4CYPEQ$L", "0YyL6RtCL8axn6AG9ItFhs"} == {guid_i, guid_j}
                     if mat_j != mat_i:
                         if _target:
                             print(f"[TARGET PAIR] SKIPPED: different mat {mat_i!r} vs {mat_j!r}")
@@ -1904,6 +1937,26 @@ class CreateDrawing(bpy.types.Operator):
                                 to_remove.add(id(path_i))
                                 to_remove.add(id(path_j))
                                 matched += 1
+                    # Unilateral fallback: when one element's shared edge is implicit
+                    # (not explicitly drawn as a path segment), segments from the other
+                    # element that lie on the boundary of that element's SVG bbox are
+                    # still on the shared face and should be removed.
+                    if matched == 0 and segs_i and segs_j:
+                        bbox_i = segs_bbox(segs_i)
+                        bbox_j = segs_bbox(segs_j)
+                        if bbox_i and bbox_j:
+                            for path_j, line_j in segs_j:
+                                if seg_on_boundary_of(line_j, bbox_i):
+                                    if _target:
+                                        print(f"[TARGET PAIR] UNILATERAL j-on-i: {line_j}")
+                                    to_remove.add(id(path_j))
+                                    matched += 1
+                            for path_i, line_i in segs_i:
+                                if seg_on_boundary_of(line_i, bbox_j):
+                                    if _target:
+                                        print(f"[TARGET PAIR] UNILATERAL i-on-j: {line_i}")
+                                    to_remove.add(id(path_i))
+                                    matched += 1
                     if _target:
                         print(f"[TARGET PAIR] adj=True, matched={matched} segment pair(s)")
                         if matched == 0:
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -1747,7 +1747,7 @@ class CreateDrawing(bpy.types.Operator):
                 adjacency_cache[key] = True
                 return True
             dot = abs(n_a.dot(n_b))
-            if dot <= 1.0 - 1e-3:
+            if dot <= 1.0 - 3.8e-5:
                 adjacency_cache[key] = False
                 return False
             # Normals are parallel — also verify the elements share a face plane.
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -1655,8 +1655,6 @@ class CreateDrawing(bpy.types.Operator):
 
         adjacency_cache = {}
 
-        _watch_adj = frozenset({"39Iwvbn41B7OZ72qQGJq6v", "08yv0FO$j2AwVYFWEkXEuh"})
-
         def are_coplanar_and_adjacent(guid_a, guid_b, tol=0.01):
             """True if the two meshes share a vertex AND have parallel face normals.
 
@@ -1665,14 +1663,11 @@ class CreateDrawing(bpy.types.Operator):
             shared face is coplanar — elements meeting at a fold angle are rejected.
             """
             key = (min(guid_a, guid_b), max(guid_a, guid_b))
-            _dbg = frozenset({guid_a, guid_b}) == _watch_adj
             if key in adjacency_cache:
                 return adjacency_cache[key]
             obj_a = get_obj(guid_a)
             obj_b = get_obj(guid_b)
             if obj_a is None or obj_b is None or obj_a.type != "MESH" or obj_b.type != "MESH":
-                if _dbg:
-                    print(f"[WATCH] {guid_a} vs {guid_b}: no mesh obj -> True")
                 adjacency_cache[key] = True
                 return True
             # Quick AABB guard
@@ -1684,13 +1679,9 @@ class CreateDrawing(bpy.types.Operator):
                 min_b = min(c[axis] for c in corners_b)
                 max_b = max(c[axis] for c in corners_b)
                 if min_a > max_b + tol:
-                    if _dbg:
-                        print(f"[WATCH] {guid_a} vs {guid_b}: AABB separated on axis {axis} -> False")
                     adjacency_cache[key] = False
                     return False
                 if min_b > max_a + tol:
-                    if _dbg:
-                        print(f"[WATCH] {guid_a} vs {guid_b}: AABB separated on axis {axis} -> False")
                     adjacency_cache[key] = False
                     return False
             # Proximity check: vertex-to-vertex OR vertex-to-edge.
@@ -1724,8 +1715,6 @@ class CreateDrawing(bpy.types.Operator):
             if not has_shared:
                 # Slower vertex-on-edge check for containment cases
                 has_shared = vertex_near_edges(verts_b, obj_a, tol_sq) or vertex_near_edges(verts_a, obj_b, tol_sq)
-            if _dbg:
-                print(f"[WATCH] {guid_a} vs {guid_b}: has_shared={has_shared}")
             if not has_shared:
                 adjacency_cache[key] = False
                 return False
@@ -1740,10 +1729,7 @@ class CreateDrawing(bpy.types.Operator):
                         return False
                 return True
 
-            contained = aabb_contains(corners_a, corners_b) or aabb_contains(corners_b, corners_a)
-            if _dbg:
-                print(f"[WATCH] {guid_a} vs {guid_b}: aabb_contained={contained}")
-            if contained:
+            if aabb_contains(corners_a, corners_b) or aabb_contains(corners_b, corners_a):
                 adjacency_cache[key] = True
                 return True
             # Coplanarity check: use the largest-face normal for each object.
@@ -1758,17 +1744,10 @@ class CreateDrawing(bpy.types.Operator):
             n_a = dominant_world_normal(obj_a)
             n_b = dominant_world_normal(obj_b)
             if n_a is None or n_b is None:
-                if _dbg:
-                    print(f"[WATCH] {guid_a} vs {guid_b}: no dominant normal -> True")
                 adjacency_cache[key] = True
                 return True
             dot = abs(n_a.dot(n_b))
-            if _dbg:
-                import math
-                print(f"[WATCH] {guid_a} vs {guid_b}: n_a={tuple(round(x,4) for x in n_a)} n_b={tuple(round(x,4) for x in n_b)} dot={dot:.6f} angle={math.degrees(math.acos(min(dot,1.0))):.3f}°")
-            if dot <= 1.0 - 3.8e-5:
-                if _dbg:
-                    print(f"[WATCH] {guid_a} vs {guid_b}: not coplanar (dot too low) -> False")
+            if dot <= 1.0 - 3.8e-5:  # ~0.5° tolerance
                 adjacency_cache[key] = False
                 return False
             # Normals are parallel — also verify the elements share a face plane.
@@ -1782,8 +1761,6 @@ class CreateDrawing(bpy.types.Operator):
             range_b = (min(projs_b), max(projs_b))
             overlap = min(range_a[1], range_b[1]) - max(range_a[0], range_b[0])
             same_plane = overlap > tol
-            if _dbg:
-                print(f"[WATCH] {guid_a} vs {guid_b}: range_a={range_a} range_b={range_b} overlap={overlap:.5f} same_plane={same_plane}")
             adjacency_cache[key] = same_plane
             return same_plane
 
@@ -2003,6 +1980,7 @@ class CreateDrawing(bpy.types.Operator):
 
                 Compatible means the visible cross-section at the shared face is
                 the same material.  Checks in order:
+
                   - Exact match (identical layer sequences).
                   - Reversed match (same assembly in opposite orientation).
                   - Suffix match: one sequence ends with all layers of the other.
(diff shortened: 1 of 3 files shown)


theoryshaw:
## Summary

Fixes #3742. When a plan or RCP drawing is generated, adjacent elements that are **coplanar and share the same material** produce a visible line between them in the SVG output. This line exists because each IFC element is projected independently into its own `<g class="projection">` group — the shared edge appears in both groups and is rendered twice. This PR removes those duplicate segments after linework generation so the surfaces read as a single continuous plane.

## Background

Bonsai's drawing pipeline does not use `draw.py`'s `main()` function — it has its own equivalent path in `CreateDrawing.execute()` inside `operator.py`. The `draw.py` implementation in the referenced PR (#7894) therefore has no effect on Bonsai drawings. This fix is applied at the correct location in `operator.py`.

## What was added

A new method `remove_coplanar_boundary_lines(root)` is called on the SVG root immediately after `merge_linework_and_add_metadata()` (which is where IFC GUIDs and material CSS classes are stamped onto the `<g>` elements). It runs for both `HALFEDGE` and `OPENCASCADE` cut modes.

The method works in three stages:

**1. Group collection**
All `<g class="... projection ...">` elements are grouped by their immediate SVG parent, so only elements that belong to the same drawing view are ever compared against each other.

**2. Visual identity check (material + style)**
Two elements are candidates for merging only if they carry exactly the same visual appearance:
- **Material key**: a sorted tuple of IFC material IDs from `ifcopenshell.util.element.get_materials()`. Using IFC IDs (not CSS class names) avoids false matches between unrelated `material-null` elements that happen to share the same CSS class but are stylistically distinct.
- **Style key**: a sorted tuple of `IfcPresentationStyle` IDs taken directly from `StyledByItem` on the element's geometry representation items. This handles elements that have no material assignment but do have distinct IFC presentation styles (e.g. different line weights or colours via direct style association).

**3. Coplanarity and adjacency check (`are_coplanar_and_adjacent`)**
Even with matching material and style, a boundary line should only be removed if the two elements are genuinely coplanar and touching in 3D space. Three sub-checks are performed in order:

- **AABB guard**: bounding boxes must overlap in all three world axes (cheap early-out).
- **Shared vertex**: at least one pair of vertices from the two Blender mesh objects must be within 0.01 m of each other in world space. This rejects elements whose 2D projections overlap because they sit at different depths — a common case with stacked floor slabs.
- **Dominant face normal**: the normal of each object's **largest polygon** (by area) must be nearly parallel (`|n_a · n_b| > 0.999`). Area-weighted average normals were considered but are unreliable for slabs: the top and bottom faces have equal area and opposite normals, so they cancel and the average drifts toward the side faces. Using the single largest face gives the true orientation. This check also correctly rejects folded walls that share an edge but meet at an angle.

Results are cached per GUID pair to avoid redundant 3D queries.

**4. Segment matching**
Surviving candidate pairs have their `<path d="Mx0,y0 Lx1,y1">` segments compared endpoint-to-endpoint (in both directions) with a 0.01 SVG-unit tolerance. Matching paths are removed from both groups.

## Cases correctly handled

| Scenario | Result |
|---|---|
| Two coplanar slabs, same material, touching edge | ✅ Boundary removed |
| Two walls in line, same material, touching edge | ✅ Boundary removed |
| Two elements, different materials | ✅ Boundary kept |
| Two elements, no material but different IFC styles | ✅ Boundary kept |
| Two elements, same material, overlapping in 2D but different depths | ✅ Boundary kept (no shared vertex) |
| Two walls sharing an edge but meeting at a fold angle | ✅ Boundary kept (normals not parallel) |

theoryshaw:
test files: https://hub.openingdesign.com/OSArch/Community_Troubleshooting/src/commit/3e9111482f429b2877a9f310d89a215ef1e774d0/7894_Join%20coplanar%20SVG%20projection%20cells%20using%20normal%20vector%20evaluation

image


image

theoryshaw:
## Background

Bonsai generates SVG drawings by projecting 3D IFC elements into 2D. When two
adjacent elements share the same material/style and lie on the same face plane,
the boundary line between them appears in **both** SVG projection groups as a
duplicate — creating a visible internal line that should not exist. The goal of
this fix is to detect and remove those duplicate lines automatically.

The core function is `remove_coplanar_boundary_lines` inside the
`CreateDrawing` operator in `operator.py`. It iterates over all pairs of
projection `<g>` elements, checks whether the two elements are adjacent and
coplanar, and removes matching SVG `<path>` segments from both groups.

---

## Changes Made This Session

### 1. Vertex-to-Edge Proximity Check (Containment Case)

**Problem:** The original adjacency check used a vertex-to-vertex shared-point
test: if no vertex of element A was within tolerance of any vertex of B, the
pair was rejected as not adjacent. This failed for **containment** — where a
smaller element sits physically inside a larger one (e.g., a pilaster embedded
in a wall). In that scenario, the inner element's corners lie on the *edges* of
the outer element, never at its corner vertices.

**Fix:** After the vertex-to-vertex check fails, fall back to a
**vertex-to-edge** check: for every vertex of B, compute the perpendicular
distance to every edge of A (and vice versa). If any vertex lies within
tolerance of any edge, the elements are considered adjacent.

```
point_to_seg_dist_sq(p, a, b):
    projects p onto segment a–b, clamps to [0,1], returns squared distance
```

This correctly handles T-junctions, embedded elements, and other cases where
shared corners don't coincide with outer-element vertices.

---

### 2. AABB Containment Bypass for the Normal Check

**Problem:** After proximity passes, the code checks coplanarity using the
**dominant face normal** (the normal of the largest polygon in each mesh). For
typical walls this is the large side face. But for a short, flat inner element
(e.g., a thin wall stub), the largest polygon is the *top face*, giving a
normal pointing in Z (up). The adjacent outer wall's normal points in Y (wall
face direction). The dot product is 0 → coplanarity check fails → boundary not
removed.

**Fix:** Before the normal check, detect whether one element's AABB is **fully
contained** within the other's on all three axes (within tolerance). If so,
skip the normal check entirely — the proximity check already guarantees they are
physically touching, and the dominant-normal heuristic is unreliable for
geometrically compact inner elements.

```python
def aabb_contains(outer, inner):
    for axis in range(3):
        if min(inner[axis]) < min(outer[axis]) - tol: return False
        if max(inner[axis]) > max(outer[axis]) + tol: return False
    return True

if aabb_contains(corners_a, corners_b) or aabb_contains(corners_b, corners_a):
    return True  # skip normal check
```

This is safe because two elements at different depth layers (which the normal
check was protecting against) can never have one AABB fully nested inside the
other.

---

### 3. Face-Plane Position Check (Parallel Offset Walls)

**Problem:** Two walls with identical orientation (same face normal direction)
but at *different positions* along that normal (e.g., two offset walls meeting
at a corner in an L-junction) were incorrectly accepted as coplanar. The
dot-product normal check only verifies that normals are parallel — it says
nothing about whether the elements are on the **same plane**.

**Fix:** After confirming the normals are parallel (dot > 0.999), project all
vertices of both elements onto the shared normal direction and compare the
resulting scalar positions. At least one face-plane position of A must be within
tolerance of one face-plane position of B for the pair to be accepted.

```python
plane_pos_a = {round(v.dot(n_a), 5) for v in verts_a}
plane_pos_b = {round(v.dot(n_a), 5) for v in verts_b}  # both use n_a
same_plane = any(abs(pa - pb) < tol for pa in plane_pos_a for pb in plane_pos_b)
```

Note that **both** sets are projected onto `n_a` (not each onto its own
normal). This is required for correctness when the two normals are
anti-parallel (+Y vs −Y): projecting onto each object's own normal produces
mirrored values (opposite signs) that would never match, even though the
elements are genuinely on the same plane.

**Example rejected:** Wall A at Y ∈ [−0.25, +0.05], Wall B at Y ∈ [−0.41,
−0.10]. Both face Y, but no Y position from A matches any from B within 0.01 m
→ correctly rejected.

**Example accepted:** Wall A and Wall B both at Y ∈ [−0.41, −0.10], normals
+Y and −Y respectively. Projecting both onto +Y gives identical sets of
positions → correctly accepted.

---

### 4. Collinear-Overlap Segment Matching (L-Shaped Walls)

**Problem:** The segment matching function `lines_match` required exact
endpoint correspondence (within tolerance, in either direction). This failed for
**L-shaped or notched walls** whose boundary at a shared interface is split into
multiple sub-segments in the SVG, while the adjacent simple rectangular wall
has a single full-length segment covering the same edge.

**Example:**
- Simple wall `2x0$XNl` has one segment at x ≈ 41.91, spanning y = 68.48 → 87.53
- L-shaped wall `1eW6OAC` has *two* segments at x ≈ 41.91:
  - y = 87.53 → 85.57 (upper portion)
  - y = 76.82 → 68.48 (lower portion)
  - (the gap y = 85.57 → 76.82 is the interior notch, handled by horizontal segments)

No exact match is possible, but all three segments represent the same physical
interface and should be removed.

**Fix:** Added a **collinear-overlap** check as a second case in `lines_match`.
For two axis-aligned segments (both horizontal or both vertical):

1. They must lie on the same line (same perpendicular coordinate within TOL).
2. Their extents along that line must overlap by more than TOL (not just touch
   at a single point).

```python
# Vertical case
if abs((x0a + x1a)/2 - (x0b + x1b)/2) < TOL:
    return max(min_ya, min_yb) < min(max_ya, max_yb) - TOL
```

The `- TOL` threshold prevents false matches between segments that merely share
an endpoint.

---

## Summary of Guards in `are_coplanar_and_adjacent`

The full decision pipeline is now:

```
1. Look up Blender objects by GUID (cache miss → query IFC + tool.Ifc)
2. If either object is missing or non-mesh → assume adjacent (True)
3. AABB guard: if bounding boxes are separated on any world axis → False
4. Proximity: vertex-to-vertex within tol?
      No → vertex-to-edge within tol?
          No → False
5. AABB containment: if one bbox fully contains the other → True (skip 6–7)
6. Dominant-normal check: largest-polygon normal for each object
      dot product < 0.999 → False (elements meet at an angle)
7. Face-plane check: project all vertices onto n_a
      no position from A within tol of any position from B → False
8. → True (adjacent and coplanar)
```

And `lines_match` now handles both:
- **Exact match**: same endpoints within TOL (either direction)
- **Collinear overlap**: both axis-aligned, on the same line, overlapping range > TOL

---

## Known Remaining Debug Output

The code still contains `[TARGET PAIR]`, `[MERGED]`, and `[COPLANAR DEBUG]`
print statements left in for ongoing troubleshooting. These should be removed
before merging to `v0.8.0`.

theoryshaw:
updated test files: https://hub.openingdesign.com/OSArch/Community_Troubleshooting/commit/30ced2a2bcb6b0dc104b162d7519385c7ff10406

image

image

theoryshaw:
## What the problem was

When two adjacent walls/slabs share a boundary, Bonsai generates an SVG where both elements have a `<path>` drawn along that shared edge — one from each element. The algorithm removes these duplicate lines by finding segments that appear in **both** elements' path lists and deleting them.

The problem: sometimes only **one** element actually draws the shared boundary as an explicit path segment. The other element's path simply doesn't include it — the shared edge is implicit in the polygon closure or owned entirely by the neighboring element. So the search for a matching pair comes up empty, and neither segment gets removed.

In the failing case (`0QpDRRuif0R80d4CYPEQ$L` vs `0YyL6RtCL8axn6AG9ItFhs`):

-   Element  **j**  drew two explicit vertical segments along the shared edge at x≈89.73 (one from y=62→99, another from y=111→155)
-   Element  **i**  had  **no segments**  at those y-ranges along x=89.73 — just a tiny stub at the very bottom
-   The bilateral search found zero matches and stopped — those lines stayed visible in the drawing

## What the fix does

After the bilateral search finds nothing, a fallback check kicks in. It computes a simple bounding box from each element's drawn segments, then asks:

> _Does any segment from element j lie on the edge of element i's bounding box, with meaningful overlap?_

Element j's two vertical segments at x≈89.73 land exactly on element i's right bbox edge, and their y-ranges sit well within element i's y-extent — so they pass the test and get removed, even though element i had no matching segment to pair them against.

The `overlap > TOL` guard prevents false positives from segments that merely graze a bbox corner without sharing real length.

test file: 
[Highland_Haven_old.zip](https://github.com/user-attachments/files/26491797/Highland_Haven_old.zip)

theoryshaw:
Updated test files:
https://hub.openingdesign.com/OSArch/Community_Troubleshooting/src/commit/93bf4bce5fba986c126dd542d3d0921aa32e2d19/7894_Join%20coplanar%20SVG%20projection%20cells%20using%20normal%20vector%20evaluation



image

theoryshaw:
Added in segments for portions of the shared edge that are not shared....



<br>
image

theoryshaw:
updated test files: https://hub.openingdesign.com/OSArch/Community_Troubleshooting/src/commit/6b6ef543ac4ec136f812703d11ce024c933111bd/7894_Join%20coplanar%20SVG%20projection%20cells%20using%20normal%20vector%20evaluation

theoryshaw:
updated test files: https://hub.openingdesign.com/OSArch/Community_Troubleshooting/src/commit/68e8ea9dab0b4dfa4b26a8b3229e4618845ac34b/7894_Join%20coplanar%20SVG%20projection%20cells%20using%20normal%20vector%20evaluation

theoryshaw:
Updated test files: https://hub.openingdesign.com/OSArch/Community_Troubleshooting/src/commit/42fdfcf3185f0d3b1be1e57696e1527d1f0ec4fb/7894_Join%20coplanar%20SVG%20projection%20cells%20using%20normal%20vector%20evaluation

I also created an architectural document for a  _dense knowledge-transfer reference for AI agents or developers who need to continue work on the "Join Coplanar Surfaces" feature for Bonsai's SVG drawing generation._

https://hub.openingdesign.com/OSArch/Community_Troubleshooting/src/commit/42fdfcf3185f0d3b1be1e57696e1527d1f0ec4fb/7894_Join%20coplanar%20SVG%20projection%20cells%20using%20normal%20vector%20evaluation/fix-3742-architectural-decisions.md

The following highlighted tests still fail, and I probably will not address them in the near future, since a lot of AI time was burned trying to solve and it continued to run into dead ends. 

`drawing: PLAN_VIEW`

`drawing: NORTH SECTION`

BIMvoice:
@theoryshaw @Moult @sboddy I ran a verification pass on this PR against current v0.8.0 so it does not stay stuck. Summary of what I found.

**It merges cleanly** against v0.8.0 at 55a2430d71, no conflicts. The `lint-formatting` failure is 31 cosmetic lines that one `black .` fixes. The `compile-and-test` failure is pre-existing on main (every ci.yml run on v0.8.0 in April 2026 was red) and unrelated, since this PR touches no C++.

**It composes with sboddy's merged #8608.** They sit at different layers: #8608 classifies edges within a single element in the C++ serializer, this PR removes duplicate segments between elements in Python after merge. #8608 keeps the `d="Mx,y Lx,y"` format and the per-product group structure that this PR's parser relies on, so nothing here is superseded. Both default off, so the combination is untested.

**It genuinely works, on the cases it reaches.** Using your own Highland_Haven_old.ifc in headless Blender 5.1.2, with the toggle on it removed real seams on 6 of 6 drawings I tried, including the exact `0QpDRRuif0R80d4CYPEQ$L` / `0YyL6RtCL8axn6AG9ItFhs` pair you called out. With the toggle off the output is byte-identical to baseline across all 76 projection groups and 3674 paths, so the opt-in gating is clean.

**Three things I think need addressing before merge:**

1. Elements with no material match every other element. `mat_keys_match((), anything)` returns True because `long_[:0] == ()`. In this model `IfcDoor` and `IfcSanitaryTerminal` have empty material keys, and on 1ST FLOOR RCP 31 of 61 touched segments were door outlines being dissolved into walls. This defeats the material-identity goal described in the PR body. Looks like a one-line guard.

2. Two walls with different exterior siding (`SIDING - BATTEN_2A` vs `SIDING - BATTEN_1`) had their shared boundary removed in ELEVATION - WEST. The exact/reversed/prefix/suffix checks correctly reject them, then it falls through to `return face_a == face_b`. The docstring says that camera-face comparison is for AXIS3 slabs, but the code does not restrict it to AXIS3, so walls reach it too.

3. The toggle can silently reset itself. `join_coplanar_surfaces` is not persisted to EPset_Drawing and has no update callback, so when `CreateDrawing` calls `update_representation` the camera data is recreated and the flag goes back to False. On two of the six drawings I tested the pass was skipped entirely despite the setting being on, with no feedback to the user. Forcing the default to True made both run. This may explain part of what you experienced as "still fails".

One more thing worth a maintainer's eye: the `fill_mode == "SVGFILL"` changes are not behind the toggle, and `num_passes` going from 0 to 1 activates a code path that could never execute before. That is the draw.py-style approach aothms suggested in #3742 and may be the more promising direction, but it currently ships ungated. I could not A/B it because SVGFILL already crashes on baseline on this model (`A linearring requires at least 4 coordinates`), which looks like a separate pre-existing bug.

On cost, the new pass added 7.5s to a 76-element plan and 5.5s to NORTH SECTION. Fixing item 1 should bring that down, since empty material keys currently force the expensive geometry test on every unmaterialised element.

Happy to help with any of the above if useful. Test artifacts and the exact before/after segment lists are reproducible from your Highland_Haven_old.ifc.

sboddy:
So I had a crack at this. On testcase output I've gotten to...

Selection_428

1. Looks like a regression.
2. Looks like a fail. Removed one correct line, and one incorrect line.
3. Regression...  I think... maybe. Are we supposed to merge styles now too?

Selection_429

1. Looks like a success.
2. Regression... I think... maybe. Are we supposed to merge styles now too?
3. An improvement, but no gold star. There are still bits that should not be there.

I think my Claude also picked up some of the stuff BIMvoice outlines:

- Empty-material wildcard in mat_keys_match (elements with no material no longer match everything)
- Camera-face fallback restricted to AXIS3 slabs only (was wrongly reachable by walls)
- join_coplanar_surfaces/generate_material_layers toggles now persist across camera representation recreation (confirmed via a live round-trip test)

sboddy:
https://github.com/sboddy/IfcOpenShell/tree/fix-3742-coplanar-bonsai-projection_2_sjb

If you want to take a look. I'll see if I (or more accurately, Claude) can't fix the outstanding issues.

theoryshaw:
awesome, will take a look, this weekend.

sboddy:
Wow! This thing is a **severe** PITA. Multiple rounds of trying to fix this thing. I got it to fix some of the above problems, but damn that wall is evil... _evil_ I tell you!

sboddy:
The more I've fought with this the more I think (1) is the correct interpretation for both PLAN and NORTH SECTION.

PLAN:
Selection_430

(2) Is now correct. If you look real close the top lines look to have different thicknesses. However if you zoom right in on the actual model you will in fact see a difference between the two sides. So I'm calling this as right.

NORTH SECTION:
Selection_431

(2) This is the best I've got so far. These lines are really tricky little buggers!

I don't want to "steal your thunder" by posting my own PR, but there's quite a lot of useful background in my drafted PR description that might be worth preserving, and feeding to the next rounds of trying to beat this into submission. I can either create a draft PR, or post the draft .md directly here as a _very_ long comment. Your choice.

BIMvoice:
@sboddy this is genuinely appreciated, thank you for grinding through this and for folding in the three things we flagged (the empty material wildcard, the AXIS3 restriction on the camera-face fallback, and the toggle persistence). Please go ahead and open the draft PR, not just a comment. Reasons: it turns your write up into a referenceable, versioned artifact instead of one very long comment buried in an already long thread, CI can actually run against your branch so we get a real signal instead of screenshots, and it gives theoryshaw a clean diff to review against his original when he gets to it this weekend, rather than a wall of text to parse. We would rather build on a draft PR than a comment, and we will actively use it, it slots straight into the linework work we're doing on #8633.

On your test rig calls: tracing your circled cases against the labeled grid, (1) being correct in both PLAN and NORTH SECTION lines up with how the code is actually meant to work. The style key check (`IfcPresentationStyle` IDs from `StyledByItem`) was written into the original PR specifically to give material less elements an identity to match on, so two undecorated elements sharing the same style merging is intended behavior, not scope creep. So no, you're not accidentally merging styles, that's the design working as intended for the no material case.

(2) in PLAN, keeping the line where there is a real but subtle difference, is the right default for a drawing tool. Erring toward a spurious line beats erring toward erasing a real boundary, so that call looks right to us too.

(2) in NORTH SECTION, the complicated wall with voids and infills, matches what theoryshaw already flagged back in April as one of the two tests he could not crack. That is the genuine frontier here, not a regression you introduced, and it is fine if that one stays open for another round.

One thing worth knowing before you push: we checked and your branch and #8633 do not touch any of the same code paths (#8633 only changes the grouping criteria in `merge_linework_and_add_metadata`, your work is entirely in `remove_coplanar_boundary_lines`/`mat_keys_match`), so there is no conflict either direction and they can land in any order. Once the draft PR is up we are happy to run the same before and after regression pass on your branch that we ran on theoryshaw's, using Highland_Haven_old.ifc plus your synthetic test rig, so there is a reproducible verification alongside the screenshots.

BIMvoice:
@sboddy I independently re-verified your branch (`fix-3742-coplanar-bonsai-projection_2_sjb`, currently at `4bc3e808d9`) against Highland_Haven_old.ifc in headless Blender.

All three things we flagged are fixed exactly as described. `mat_keys_match((), x)` now returns False (operator.py line 2130), the camera-face fallback now returns None for anything but AXIS3 (operator.py lines 2084 to 2085), and both toggles are now wired to EPset_Drawing the same way as has_linework/linework_mode (prop.py lines 500 to 511, tool/drawing.py lines 1012 to 1013 and 1049 to 1052).

Live, on the real model: with the toggle off, output is path-count identical to v0.8.0 baseline on every drawing I tried, so the opt-in gating stays clean. With it on, the door and sanitary terminal outlines that were wrongly dissolved on theoryshaw's original branch are now correctly kept (confirmed the exact pair theoryshaw called out in April too, `0QpDRRuif0R80d4CYPEQ$L` and `0YyL6RtCL8axn6AG9ItFhs`, now resolves). On ELEVATION - WEST, the different-siding wall pair that was wrongly merged before is now correctly kept apart. I did not find any case where a boundary that should stay visible got newly removed.

I was not able to fetch your synthetic test rig, hub.openingdesign.com blocks both git clone and raw file access for me, so I could not independently replay the numbered grid cases from your screenshots. From the screenshots themselves, the COMPLICATED WALL WITH VOIDS AND INFILLS case in NORTH SECTION does look like the same genuinely hard frontier theoryshaw flagged in April, not a regression, so it seems reasonable for that to stay open for another round.

This is solid work, thank you for grinding through it. Given theoryshaw's own suggestion, this seems ready for a draft PR so CI can run and the diff is easy to review against the original.

sboddy:
I'm making progress, but it is a torturous journey filled with lurking traps. I'm pushing fixes to my branch, and currently having yet another go at removing those lingering lines. I've not (so far) addressed the no-material, same style item. I'm still not sure if the intent is to merge styles or not. I've been ignoring that for now in favour of solving the real knotty problem wall.

When I reach the point that I feel I can't prompt Claude to further resolve it (getting real close now) I'll create the draft PR with all the gory details.

theoryshaw:
Closing mine, in favor of @sboddy's solution! https://github.com/IfcOpenShell/IfcOpenShell/pull/9330

PR #7940, opened 2026-04-15
Concatenate selections

theoryshaw:
### Overview

This branch standardizes how Bonsai's "select similar" operators handle multiple selected objects. Previously, most operators either used only the active/first object, ignored duplicates, or built clipboard queries inside loops (producing overwrites or single-item output). All affected operators now collect their relevant entities from the full selection, deduplicate by entity id, and write a single clipboard query joining all values with `+` — consistent with IFC selector query syntax.

----------

### `bim.select_similar`  —  search/operator.py

Previously only `reference_values[0]` was passed to `_generate_clipboard_query`, so only the first selected object's value appeared in the clipboard.

Now all reference values are passed and joined with `+`.

**Example:** Select two walls and click "Select Similar" on `GlobalId`:

```
# Before
GlobalId = "1BDpUFH1P7$8huSfztD7Kp"

# After
GlobalId = "1BDpUFH1P7$8huSfztD7Kp" + GlobalId = "2yrtTmJoD5gvf_wIjwsNxp"

```

----------

### `bim.select_ifc_class`  —  search/operator.py

The clipboard was set and reported inside the per-class loop, overwriting on each iteration. The separator was also `,` rather than `+`.

Now the clipboard is written once after the loop with `+` as the separator.

**Example:** Select an `IfcWall` and an `IfcSlab`, then click "Select IFC Class":

```
# Before (last iteration wins, reported twice)
IfcWall, IfcSlab

# After (written once)
IfcWall + IfcSlab

```

----------

### `bim.select_similar_container`  —  spatial/operator.py

Only `context.active_object`'s container was used. No clipboard output existed at all.

Now all selected objects are iterated, containers are deduplicated by entity id, elements from all containers are selected, and a clipboard query is generated.

**Example:** Select objects from Level 1 and Level 2, then click "Select Similar Container":

```
# Before
(no clipboard output, only active object's container used)

# After
location = "Level 1" + location = "Level 2"

```

----------

### `bim.select_aggregate`  —  aggregate/operator.py

When multiple parts of the same aggregate were selected, the aggregate was appended to a plain list multiple times, producing duplicate entries in the clipboard query.

Aggregates are now collected into a dict keyed by entity id before selection and query generation.

**Example:** Select three parts all belonging to the same `IfcElementAssembly "Truss-A"`:

```
# Before
parent = "Truss-A" + parent = "Truss-A" + parent = "Truss-A"

# After
parent = "Truss-A"

```

----------

### `bim.select_by_material`  —  material/operator.py

The most involved change. The operator only used the explicit `material` prop (a single material ID set by the UI button), with no awareness of other selected objects. Layer set usages were not resolved to specific layers.

Now the operator derives materials from all selected objects. When the clicked material is an `IfcMaterial` within a layer set, its layer index is detected and the corresponding layer material is pulled from each other selected object's layer set. Falls back to the explicit prop if selection yields nothing.

**Example 1 — simple materials:** Select a concrete wall and a steel column, click "Select By Material" on either:

```
# Before
material = "Concrete"

# After
material = "Concrete" + material = "Steel"

```

**Example 2 — layer sets:** Select two walls, click "Select By Material" on Layer 1 (the exterior CMU layer). Wall A has `[CMU, Air, GWB]`, Wall B has `[Brick, Air, GWB]`:

```
# Before
material = "IfcMaterialLayerSet"

# After
material = "CMU" + material = "Brick"

```

The layer index from the clicked material is used to look up the matching layer on every other selected object's layer set.

### DIFF
--- a/src/bonsai/bonsai/bim/module/aggregate/operator.py
+++ b/src/bonsai/bonsai/bim/module/aggregate/operator.py
@@ -407,13 +407,18 @@ class BIM_OT_select_linked_aggregates(bpy.types.Operator):
     bl_label = "Select linked aggregates"
     bl_options = {"REGISTER", "UNDO"}
     select_parts: bpy.props.BoolProperty(default=False)
+    should_unhide: bpy.props.BoolProperty(default=False, options={"SKIP_SAVE"})
 
     @classmethod
     def description(cls, context, properties):
         if properties.select_parts:
-            return "Select all aggregates, subaggregates and all their parts"
+            return "Select all aggregates, subaggregates and all their parts\n\nALT+Click to also unhide hidden objects (viewport and local hide)"
         else:
-            return "Select all aggregates"
+            return "Select all aggregates\n\nALT+Click to also unhide hidden objects (viewport and local hide)"
+
+    def invoke(self, context, event):
+        self.should_unhide = event.alt
+        return self.execute(context)
 
     def execute(self, context):
         for obj in context.selected_objects:
@@ -446,11 +451,18 @@ class BIM_OT_select_linked_aggregates(bpy.types.Operator):
                     for element in parts_objs:
                         obj = tool.Ifc.get_object(element)
                         if obj:
+                            if self.should_unhide:
+                                obj.hide_viewport = False
+                                obj.hide_set(False)
                             obj.select_set(True)
                 else:
                     for element in parts:
                         obj = tool.Ifc.get_object(element)
-                        obj.select_set(True)
+                        if obj:
+                            if self.should_unhide:
+                                obj.hide_viewport = False
+                                obj.hide_set(False)
+                            obj.select_set(True)
 
         return {"FINISHED"}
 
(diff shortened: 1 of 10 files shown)


theoryshaw:
**Summary of rebase**

-   `Concatenate_selections`  is rebased onto  [Unhide_with_alt_click](https://github.com/IfcOpenShell/IfcOpenShell/pull/7781)  — same 5 commits, replayed cleanly on top
-   `SelectSimilarContainer.execute()`: supports both multi-container iteration (from Concatenate) and  `should_unhide`  (from Unhide), with the  `self.container`  panel shortcut preserved
-   `SelectByMaterial.execute()`: keeps Concatenate's full multi-object material loop, but now passes  `should_unhide=self.should_unhide`  to each  `core.select_by_material()`  call
-   `Unhide_with_alt_click`  is untouched

The next build should merge both PRs in order — Unhide first (it's older, lower PR number), then Concatenate on top — with no conflicts.

PR #7941, opened 2026-04-17
Fix #7942: Fix extend_walls_to_underside ridge artifact

### LINKED ISSUE #7942: `bim.extend_walls_to_underside` seems to glitch out
`bim.extend_walls_to_underside` seems to glitch out on the following file. 

[extend to roof - Copy.ifc.txt](https://github.com/user-attachments/files/26807578/extend.to.roof.-.Copy.ifc.txt)

https://github.com/user-attachments/assets/2d21a5a2-0b8a-4b4e-bcab-5a51761b15ec

theoryshaw:
## Fix #7942 :  `ExtendWallsToUnderside`  produces wrong geometry on second call

**Closes:** _(link issue here)_

### Summary

`bpy.ops.bim.extend_walls_to_underside` glitched when applied twice to the same wall — once per slope of a ridge roof. The first call produced correct geometry, but the second call introduced spurious extra vertices (e.g. 14 instead of 10), resulting in a malformed gable profile.

### Root Cause

`clip_wall_to_slab` builds an `IfcPolygonalFaceSet` clip solid from the downward-facing slope face by extruding its exact vertex footprint upward. When the operator is applied twice for a symmetric ridge roof, the two clip solids share an **exact ridge edge** — their boundaries are coincident in 3D.

OCCT's boolean engine treats this as a **kissing-solid** condition: two volumes that touch at a shared face or edge rather than overlapping volumetrically. This is numerically degenerate and causes OCCT to introduce extra vertices at the coincident boundary, resulting in a corrupted gable profile.

### Fix

Instead of extruding the exact face footprint, `clip_wall_to_slab` now builds the clip solid's bottom face as a **rectangle on the slope plane**, sized to the tight bounding box of the original face vertices in slope-plane coordinates, **extended by a small margin** (1 project unit) in each tangent direction.

This margin pushes each clip solid slightly past the ridge into the adjacent slope's territory, turning the shared ridge edge into a **volumetric overlap** between the two solids. OCCT handles overlapping `DIFFERENCE` operands correctly — material already removed by the first boolean simply remains absent.

The construction uses an orthonormal basis spanning the slope plane (`tangent1`, `tangent2` derived from the face normal), projects the face vertices onto those axes to find the tight extents, then adds the margin before building the clip bmesh. The extrusion height (`max_z - min_z`) and all downstream logic (`add_boolean`, `mark_manual_booleans`) are unchanged.

### Changes

File

Change

`tool/model.py`  —  `clip_wall_to_slab`

Build per-face clip solid from a margined slope-plane rectangle instead of exact face footprint; one operand per face instead of one combined mesh

`tool/model.py`  —  `get_slab_clipping_bmesh`

Add  `faces.ensure_lookup_table()`  after  `recalc_face_normals`

`core/model.py`  —  `extend_wall_to_slab`

Minor cleanup

`bim/module/model/wall.py`  —  `ExtendWallsToUnderside`

Minor cleanup

### Testing

Tested against a ridge roof with two slopes applied sequentially to both a wall (`IfcWall`) and a covering (`IfcCovering`). Both now produce correct gable profiles after two operator calls.

----------

_Generated with the assistance of an AI coding tool._


theoryshaw:
conflict write-up: https://github.com/falken10vdl/bonsaiPR/blob/main/logs/summaries/2026-05-21-PR-7941.md

theoryshaw:
closing, this pr was part of https://github.com/IfcOpenShell/IfcOpenShell/pull/7944

PR #6853, opened 2025-06-26
Derived dimensions2

falken10vdl:
PR that replaces https://github.com/IfcOpenShell/IfcOpenShell/pull/6698 so it is cleaner.
There is also the addition of [Commit 3f0101e](https://github.com/IfcOpenShell/IfcOpenShell/commit/3f0101e701b48339797458b0bf417b558c70afde) to handle the gizmo.
Cheers!

Issue #6835, opened 2025-06-23
Bonsai | IFCx3BaseQuantities-Blender showing wrong calculation for IfcDoor


Arunarchitect:
### Bug Description

When I use IFCx3BaseQuantities-Ifcopenshell, its shown as 1of them were not quantified,

![Image](https://github.com/user-attachments/assets/970b743c-55d4-4334-b60c-037bae445b91)
when I use IFCx3BaseQuantities-Blender, i get both width and height as equal values(in reality its not)

![Image](https://github.com/user-attachments/assets/8eed3bca-c2cb-4343-8824-47801b4fb6c2)

[test.ifc.txt](https://github.com/user-attachments/files/20865232/test.ifc.txt)

### Debug and Error Output

os: Windows                                                                                                             os_version: 10.0.19044                                                                                                  python_version: 3.11.7                                                                                                  architecture: ('64bit', 'WindowsPE')                                                                                    machine: AMD64                                                                                                          processor: Intel64 Family 6 Model 165 Stepping 3, GenuineIntel                                                          blender_version: 4.2.5 LTS                                                                                              bonsai_version: 0.8.3-alpha250613                                                                                       bonsai_commit_hash: 9b8357c                                                                                             bonsai_commit_date: 2025-06-13T19:33:30+05:00                                                                           last_actions:                                                                                                           last_error: None                                                                                                        ifc: Sreekesh.ifc                                                                                                       schema: IFC4X3                                                                                                          preprocessor_version: IfcOpenShell 0.8.3-alpha250613                                                                    originating_system: Bonsai 0.8.3-alpha250613-9b8357c    

```true

```

franseoane:
Also, multiple errors in the calculations of quantities for windows, with IFC4x3 Base Quantities – IfcOpenShell  and  Blender
Example: 
Wall 5x3x0.2 m
Parametric Window: 1x1 m

Qto_WindowBaseQuantities: 
	Area=1.19 (must be 1.00 m2)
Qto_WallBaseQuantities: Ok
Qto_OpeningElement:  
	Area=1.2 (must be 1.00 m2)
	Height/Altura=1.2 (must be 1.00 m)

Blender 4.5.4 LTS
Bonsai v0.8.4 - alpha251108

![Image](https://github.com/user-attachments/assets/4da23e8f-3c66-4696-b8d7-9825d7de8ce7)

[...]

PR #7974, opened 2026-04-20
See #6853. Minor fix for IfcDoor with IFC4x3 quantity calculation wit…

maxfb87:
…h blender engine


maxfb87:
I don't know why but the link points to a wrong issue...
This is the right one
https://github.com/IfcOpenShell/IfcOpenShell/issues/6835

PR #7986, opened 2026-04-21
Use get class properties instead of get class

### LINKED ISSUE #7985: bSDD ratelimit in Bonsai
### Bug Description

When trying to add Properties by using the bSDD Classification. It says for every Classification "No bSDD Props Found". After some digging i found out that `tool.Bsdd.bsdd_classes` returns a class with a Rate limit warning.
 
```python
>>> bonsai.tool.Bsdd.bsdd_classes 
{'https://identifier.buildingsmart.org/uri/hw/som/0.2.2/class/Pumpe': {'statusCode': 429, 'message': 'Rate limit is exceeded. Try again in 3 seconds.'}}
```
This happens in `import_bsdd_classes` of bonsai.core.bsdd when `bsdd.import_class_properties()` uses
```python
if not (bsdd_class := cls.get_bsdd_class(active_class.uri)):
```
That loads the whole class and leads somehow to a ratelimit. (If i run it in VSCode Debug mode this wont happen but for me it happens every time if I run it normally)

The function uses `/api/Class/v1`
In this case it would make more sense to use `/api/Class/Properties/v1` because we just request the Class Properties not the whole class. In my tests this also avoids the rate limiting somehow.

If you want i create a PR that adds an `get_bsdd_class_properties() `function to the Bsdd tool and implement it in `import_class_properties()`

### Attachments

_No response_

### Debug and Error Output

```true

```

c-mellueh:
fixes #7985

### DIFF
--- a/src/bonsai/bonsai/tool/bsdd.py
+++ b/src/bonsai/bonsai/tool/bsdd.py
@@ -150,7 +150,6 @@ class Bsdd(bonsai.core.tool.Bsdd):
             statuses.append("Inactive")
         return list(filter(lambda d: d["status"] in statuses, dicts))
 
-
     @classmethod
     def get_related_ifc_entities(cls) -> list[str]:
         active_object = bpy.context.active_object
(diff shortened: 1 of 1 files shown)


c-mellueh:
@aothms is there a problem with the PR?
The bug #7985 destroys the whole bSDD workflow because i cant give bsdd psets to elements anymore. I can't tell my coworkers to go work on bonsaiPR till this is fixed  :( 
I dont want to be annoying if the PR needs to pass some further inspections please lmk :)

PR #8012, opened 2026-04-26
Closes #7691: Enable bim.measure_tool in edit mode

### LINKED ISSUE #7691: Have the 'Measure Tool' work in edit mode, as well.
### Feature Description

I would be nice if the 'Measure Tool', within the Explore Tool, worked in edit mode, as well.

theoryshaw:
Register Shift+M in the Mesh keymap so the measure tool can be invoked while in EDIT_MESH mode without requiring the Explore workspace tool. Append measure controls (type selector + clear button) to VIEW3D_HT_tool_header so they are visible in edit mode.

Guard the object deselect in hotkey_S_M to OBJECT mode only, since deselecting objects has no meaning in edit mode.

Fix snapping: obj.ray_cast() has no valid BVH tree for the active edit-mode mesh, so vertex/edge snapping was silently failing. Replace with a BMesh proximity pass via bmesh.from_edit_mesh() when context.mode == "EDIT_MESH".

### DIFF
--- a/src/bonsai/bonsai/bim/module/project/__init__.py
+++ b/src/bonsai/bonsai/bim/module/project/__init__.py
@@ -124,8 +124,16 @@ def register():
         kmi.properties.should_save_as = False
         addon_keymaps.append((km, kmi))
 
+        km = wm.keyconfigs.addon.keymaps.new(name="Mesh", space_type="VIEW_3D")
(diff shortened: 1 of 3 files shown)


c-alpha:
Any progress on this PR?

I am sorely missing this feature for positioning objects at a specified distance from one another.

Will the measurements remain attached to the objects, and update their values while moving the objects?

PR #8014, opened 2026-04-27
Surface styles fixes and message warnings

falken10vdl:
Here are a  number of commits to fix some minor issues in the handlign of Surface Styles. Also some watning messages are added to help the user understand what combinations of modes are supported.

Cheers!


falken10vdl:
Details in https://community.osarch.org/discussion/3409/surface-styles-revisited/p2

PR #8030, opened 2026-04-30
Test modal pr

brunoperdigao:
I've drafted no headless tests for modal operators. However, I wasn't able to integrate with the current pytest setup, because I think pytest-blender doesn't recognize the flag `--enable-event-simulate`.
The current setup uses `MODULE` to run test groups. There are only two options: `make test-modal MODULE=snap` and `make test-modal MODULE=wall`.
This works for isolated testing during development/refactoring of modal operators and related tools like the polyline tool and snap.
Should we proceed with this limited integration, or is there a way to get pytest-blender to recognize the event simulation flag? 
Any thoughts on this @Andrej730 @Moult


brunoperdigao:
I'm going to merge since it is an isolated implementation. I'll keep improving this for the next days.

Issue #4014, opened 2023-11-14
Issues with annotation having multiple products


theoryshaw:
Changing the 'product assignment' is not updated. 

Had to delete the  'product assignment' first, for it to then take on a new product.  

file: https://hub.openingdesign.com/OpeningDesign/Monica_Residence/src/commit/bf25b828213ff5896ce6793aeeec56385d5609d9/Models/BlenderBIM/Monica_Residence.ifc

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/f447e9a5-8c89-4169-9284-9f01504d4b2f

theoryshaw:
maybe there's multiple product assignments per tag?

Andrej730:
Makes sense - issue caused by IfcAnnotation having multiple assigned products. In IFC there is no limitation to how many products are assigned but in IfcAnnotation we kind of assume that it's going to be just 1.
 
The question is how that happened, any ideas how to assign multiple products from UI that will lead to that issue?

theoryshaw:
no clue, any ideas @chopinregis?

chopinregis:
> The question is how that happened, any ideas how to assign multiple products from UI that will lead to that issue?

Before @theoryshaw showed me that i could cancel out the objects it had stored, which i didn't know previously, everytime that I would use the picker i'll select and ojbect and validate, if it doesn't show me what i wanted, I'll activate the picker again until i manage to get something. So this seemed to be storing materials continuously.

Andrej730:
@chopinregis on what project I can reproduce it consistently? 
Tried on the one Ryan attached and it seems to work fine - if there are no multiple products assigned before then it unassigns and assigns 1 product maximum.

theoryshaw:
I've been lazy, and haven't reported this bug in awhile, although it happens a lot.  I'm usually able to fix it, if i repeat assigning the product multiple times. 

...

If you go to `RCP - FRAMING - ROOF` and try to assign the annotation (GlobalId = "3SRzCJePPEKQfMyu13_WEu") to the header object right next to it, you see that it doesn't assign it to that header, but another one. 

file: https://hub.openingdesign.com/OpeningDesign/Cool_Car_Guys/src/commit/6160e32f1912670e91c4748f82fde1cefac43846/Open/Models/Bonsai/OD_Cool_Car_Guys.ifc

https://github.com/user-attachments/assets/92e519bb-822c-4dfa-9884-c059ae8643e4

[...]

Issue #1427, opened 2021-04-10
Proper way to handle exception of cpp/c++ in python script.


OkanPinar:
Hello everyone.

I am trying to parse ifc model on [pythonocc examples](https://github.com/tpaviot/pythonocc-demos/blob/master/assets/ifc_models/IFC%20Schependomlaan.ifc) with  ifcopenshell in python and i created upcoming  script which is creating geometries in three different way. First two method is not using pythonocc and they failed entity #46535, third method is failing after entity #275367. Somehow using pythonocc is passing entity #46535 without any problem . TopoDs compound object of entity #46535 have a child but display of it is empty. It is obvious that there is a problem in that entity but there is no python exception at first and second method. Is there any proper way to handle this kind of exceptions?

Platform details:
Env: Windows-Anaconda.
Python: 3.8
OCCT: 7.4.1
IfcOpenshell: v0.6.0

Thanks in advance.

```
import ifcopenshell
import ifcopenshell.geom
import os

ifc_filename = "./IFC_Schependomlaan.ifc"
assert os.path.isfile(ifc_filename)
ifc_file = ifcopenshell.open(ifc_filename)

settings = ifcopenshell.geom.settings()


def method_1():
    current_shape_index = 0
    iterator = ifcopenshell.geom.iterator(settings, ifc_file)
    valid_file = iterator.initialize()
    if valid_file:
        while True:
            shape = iterator.get()
            current_shape_index += 1
            if current_shape_index > 228:
                print("index: %s" % current_shape_index, shape.product)
            else:
                print("index: %s"% current_shape_index)
            if not iterator.next():
                break


def method_2():
    current_shape_index = 0
    products = ifc_file.by_type("IfcProduct")
    for product in products:
        if hasattr(product, 'Representation') and product.Representation:
            shape = ifcopenshell.geom.create_shape(settings, inst=product)
            current_shape_index += 1
            if current_shape_index > 226:
                print("index: %s" % current_shape_index, product)
            else:
                print("index: %s" % current_shape_index)


def method_3():
    current_shape_index = 0
    products = ifc_file.by_type("IfcProduct")
    settings.set(settings.USE_PYTHON_OPENCASCADE, True)
    for product in products:
        if hasattr(product, 'Representation') and product.Representation:
            shape = ifcopenshell.geom.create_shape(settings, inst=product)
            current_shape_index += 1
            if current_shape_index > 957:
                print("index: %s" % current_shape_index, product)
            else:
[...]

Issue #1421, opened 2021-04-07
Importing Tessellation meshes is slow and generates additional geometry


brunopostle:
Attached are two versions of the same IFC mesh, this mesh has:

- 1948 unique vertices
- 2251 faces, consisting of:
  - 1977 quad faces
  - 274 triangles

A simple conversion to triangles would return 1977*2 + 274 = 4228 triangles

[_test_tessellation.ifc.gz](https://github.com/IfcOpenShell/IfcOpenShell/files/6269744/_test_tessellation.ifc.gz) is a Tessellation constructed using IfcCartesionPointList3D, IfcIndexedPolygonalFace and IfcPolygonalFaceSet. IfcConvert takes 6.5 seconds to process this into a DAE file with 4240 triangles, there are more triangles than expected and additionally the resulting mesh is non-manifold in blender.

[_test_brep.ifc.gz](https://github.com/IfcOpenShell/IfcOpenShell/files/6269743/_test_brep.ifc.gz) is a Brep constructed using IfcCartesianPoint, IfcFace and IfcClosedShell. It is a much larger file because this format is very inefficient. IfcConvert takes 3 seconds to process this into a DAE file with 4228 triangles, this is the expected result.

In principle, the Tessellation should be simpler to process since there is no need to identify and merge duplicate points, but the result is slower and less accurate.

Moult:
See #1391 and #841

Amoki:
Can https://github.com/IfcOpenShell/IfcOpenShell/issues/1411 be related to it too?
The example model uses Tessellation too.

brunopostle:
Tessellation handling seems ok here using current git

Moult:
@brunopostle does "ok" refer to speed or the additional geometry issue?

brunopostle:
@Moult Speed with Tessellation seems to be almost exactly the same as IfcClosedShell Brep. The geometry issues are fixed, at least the test mesh that was becoming non-manifold before is now ok.

Moult:
Awesome! Closing issue :D

PR #8032, opened 2026-05-01
Closes #6621: Fix IfcRelAssignsToProduct on duplicate

### LINKED ISSUE #6621: Duplicated object and annotation should get unique IfcRelAssignsToProduct
### Feature Description

Would be nice if you duplicate an object and annotation that's attached via `IfcRelAssignsToProduct`, that the new object/annotation are attached via their own unique `IfcRelAssignsToProduct`.

theoryshaw:
Closes #6621: Fix IfcRelAssignsToProduct on duplicate

When duplicating a product+annotation pair linked via IfcRelAssignsToProduct, copy_class's generic else-branch appended the new annotation to every existing rel that referenced the original, leaving both old and new products sharing stale RelatedObjects.

Adds fix_assigns_to_product_after_duplication to
tool.Geometry, called at the end of duplicate_ifc_objects. It corrects the three cases: new annotation in old product's rel (remove it), stale original in new product's rel (remove it), and annotation-only duplicate (build a fresh unique rel).

### DIFF
--- a/src/bonsai/bonsai/tool/geometry.py
+++ b/src/bonsai/bonsai/tool/geometry.py
@@ -2533,6 +2533,7 @@ class Geometry(bonsai.core.tool.Geometry):
         # is complete.
         cls._recalculate_walls_with_new_connections(old_to_new)
 
+        cls.fix_assigns_to_product_after_duplication(old_to_new)
         bonsai.bim.handler.refresh_ui_data()
         tool.Root.reload_grid_decorator()
         return old_to_new, new_active_obj or active_object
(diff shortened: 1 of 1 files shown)


theoryshaw:
## Summary

When duplicating an IFC product (e.g. `IfcActuator`) that has an annotation (`IfcAnnotation`) assigned to it via `IfcRelAssignsToProduct`, the new product and new annotation were left sharing stale relationship data with the originals instead of getting a clean, independent assignment.

## Root Cause

`ifcopenshell.api.root.copy_class` handles relationship copying in `copy_indirect_attributes`. For most relationship types there is explicit handling (psets, aggregates, materials, voids, etc.), but `IfcRelAssignsToProduct` falls through to the generic `else` branch:

```python
else:
    for i, value in enumerate(inverse):
        if value == from_element:
            new_inverse = ifcopenshell.util.element.copy(self.file, inverse)
            new_inverse[i] = to_element
        elif isinstance(value, (tuple, list)) and from_element in value:
            new_value = list(value)
            new_value.append(to_element)  # ← the problem
            inverse[i] = new_value

```

When `copy_class` runs for the **product** (`IfcActuator #1368 → #1421`), `RelatingProduct` matches directly so a new rel is correctly created:

-   `rel #1427`:  `RelatingProduct=#1421, RelatedObjects=[#1400]`  ✅

When `copy_class` then runs for the **annotation** (`IfcAnnotation #1400 → #1438`), it finds two `IfcRelAssignsToProduct` inverses and hits the list-append branch for each:

-   `rel #1410`  (old product):  `RelatedObjects`  becomes  `[#1400, #1438]`  ❌
-   `rel #1427`  (new product):  `RelatedObjects`  becomes  `[#1400, #1438]`  ❌

The result is that both the old and new product point to both annotations, and both annotations consider themselves assigned to both products.

## Fix

A new classmethod `fix_assigns_to_product_after_duplication` is added to `tool.Geometry` and called at the end of `duplicate_ifc_objects`, after all `copy_class` calls have completed. It iterates over every newly created element and corrects its `IfcRelAssignsToProduct` memberships by classifying the `RelatingProduct` into one of three cases:

Case

Condition

Action

**A**

`RelatingProduct`  is an  **old**  (original) element

The new annotation was incorrectly appended to the original product's rel — remove it

**B**

`RelatingProduct`  is a  **new**  (duplicated) element

The stale original annotation was carried into the new product's rel — remove it

**C**

`RelatingProduct`  is  **unchanged**  (annotation-only duplicate)

The new annotation was appended to the shared rel — remove it and create a fresh  `IfcRelAssignsToProduct`  unique to the new annotation

After the fix, each product's rel contains only its own annotation, matching the behavior of `bim.assign_selected_as_product`.

## Test Plan

-   Duplicate a product+annotation pair linked via  `IfcRelAssignsToProduct`  (e.g. using  `bim.assign_selected_as_product`  to create the link first, then Shift+D). Verify each copy has its own independent  `IfcRelAssignsToProduct`  with the correct  `RelatingProduct`  and  `RelatedObjects`.
-   Duplicate the duplicated pair again — verify the chain stays clean across multiple generations.
-   Duplicate only the annotation (not the product) — verify the new annotation gets a fresh  `IfcRelAssignsToProduct`  pointing to the same product, and the original annotation's rel is untouched.
-   Duplicate only the product (not the annotation) — verify existing behavior is unchanged (both old and new product reference the same annotation).
-   Verify unrelated duplications (no  `IfcRelAssignsToProduct`  involved) are unaffected.

theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/bf48dff6052d0e7f21fe38b6f38abf98d7452639/logs/summaries/2026-07-10-rebase-fix-duplicate-assigns-to-product.md

theoryshaw:
Pushed df35a895, which updates case A so this PR keeps working alongside #9054.

**Why it was needed.** #9054 fixes the other half of #4014: `copy_class` was cloning a product's `IfcRelAssignsToProduct` onto the duplicate while keeping the *original's* annotations in `RelatedObjects`, so both products claimed the same annotation. This PR never covered that path — its guard `if not rel.is_a("IfcRelAssignsToProduct") or new_elem not in rel.RelatedObjects: continue` requires the new element to be a related object, and there it is the `RelatingProduct`.

But the two interact. Case A here only removed the new element from the original product's relationship, relying on `copy_class` having cloned that relationship onto the new product so case B could strip the stale original from it. Once `copy_class` stops cloning it, nothing re-links the annotation copy:

```
duplicate wall + annotation together, with #9054 applied
  after copy_class : annotation -> [wall] | annotation_copy -> [wall]
  before this commit, after case A : annotation -> [wall] | annotation_copy -> NOTHING
  after this commit                : annotation -> [wall] | annotation_copy -> [wall_copy]
```

**What changed.** Case A now follows the product: when the relating product was duplicated in the same operation, the new element is assigned to the corresponding copy, paired by index so the i-th copy of an annotation labels the i-th copy of the product. It appends to the new product's existing relationship if it has one, creates a fresh one otherwise, and skips when the element is already assigned to that product.

**Merge order does not matter.** That last guard is the reason. Verified over five scenarios against both the old and the new `copy_class`: product + annotation together, annotation alone, product alone, and the two task equivalents so the 4D Output link is not disturbed. Against the old `copy_class` every case this PR already handled still passes; the only failure is duplicating the product alone, which is exactly the bug #9054 fixes and which this PR never claimed.

Case B is still live, incidentally — #9054 keeps copying non-annotation related objects, such as the task that outputs the product, so a duplicated product + task pair still goes through it.

PR #8048, opened 2026-05-11
Fix lint failures and add missing pyparsing dependency

Ghesselink:
Trying to get the CI checks green (works on fork now) 

- pyparsing: it's needed by ifcopenshell.express at import time but wasn't declared anywhere. Added it to pyproject.toml and the CI install line so the express tests can run.
- Black / ruff: a few files weren't formatted.
- ty stub: the .pyi signature for arrange_polygons didn't match the wrapper anymore (1 arg vs. 2). 
- ty version: pinned to 0.0.34 in ci-lint.yaml. uv tool install ty was unpinned, so CI was grabbing the latest ty. See also https://github.com/IfcOpenShell/IfcOpenShell/pull/8033


aothms:
Thanks Geert!

PR #8055, opened 2026-05-13
Add export option to DebugActiveDrawing

theoryshaw:
Fix failure detection by switching from `except Exception` to checking the return value, then back to `except Exception` once it was confirmed that `bpy.ops` re-raises inside another operator. Identify the drawing via the active scene camera (consistent with CreateDrawing) to avoid the `ifc_definition_id=0` crash on non-drawing list entries.

Improve the binary search to test each candidate in isolation, enabling multiple independent failures to be found in one pass. Suppress drawing-generation log noise during tests and print a clean summary with element id, type, and name.

Add a `should_export_isolated_file` operator property that extracts the failing elements into a minimal standalone IFC file using ExtractElements, making it easy to attach a reproduction case when filing a bug report.

### DIFF
--- a/src/bonsai/bonsai/bim/module/debug/operator.py
+++ b/src/bonsai/bonsai/bim/module/debug/operator.py
@@ -16,6 +16,8 @@
 # You should have received a copy of the GNU General Public License
 # along with Bonsai.  If not, see <http://www.gnu.org/licenses/>.
 
+import contextlib
+import io
 import logging
 import os
 import platform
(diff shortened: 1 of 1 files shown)


theoryshaw:
## Summary

Overhauls `DebugActiveDrawing` (`bim.debug_active_drawing`) to reliably
find failing drawing elements and optionally export them for bug reporting.

### Bug fixes

- **Drawing lookup crash** (`Instance #0 not found`): The operator was
  reading `props.drawings[active_drawing_index].ifc_definition_id`, which
  returns `0` when the active list entry is a group/folder header rather
  than a real drawing. Switched to `tool.Blender.get_ifc_definition_id(
  context.scene.camera)`, consistent with how `CreateDrawing` identifies
  the active drawing.

- **Failure detection never triggered**: `drawing_fails_to_load` used
  `except Exception` to detect crashes. When `bpy.ops.bim.create_drawing`
  is invoked directly by the user, Blender swallows the C++ SWIG exception
  at the operator level and the `except` block never fires. When invoked
  from *within* another operator (as the binary search does), Blender
  re-raises it as a Python `RuntimeError` — so `except Exception` is
  correct in this context. The original approach was right; it was
  temporarily broken during investigation and is now restored.

- **Only one failing element reported**: The final per-element confirmation
  test called `drawing_fails_to_load(all_elements - {element})`, which asks
  "does excluding this one element fix everything?" For multiple independent
  failures this always returns `True` (still broken), giving the misleading
  impression that an element is not the cause. Changed to
  `drawing_fails_to_load({element})` — "does this element fail in
  isolation?" — so each bad element is confirmed independently and all
  failures are surfaced in one pass.

### Improvements

- **Noise suppression**: Drawing generation profile logs (`Processing body
  context`, `Generate linework`, etc.) are suppressed during binary search
  iterations using `contextlib.redirect_stdout`, keeping the console output
  focused on search progress.

- **Readable output**: Replaced raw IFC entity reprs with formatted
  `#id IfcClass 'Name'` labels. Added colored `FAIL`/`ok` per-element
  confirmation lines and a bordered summary block at the end.

- **Input guards**: Added early `{"CANCELLED"}` returns with clear error
  messages for no IFC file loaded, no active camera, and camera not linked
  to an IFC drawing element.

- **Export to IFC**: Added a `should_export_isolated_file` boolean property
  (shown via `invoke_props_dialog` before the search runs). When enabled,
  after finding failing elements the operator extracts them — along with
  their full spatial hierarchy, materials, and representations — into a
  minimal standalone `.ifc` file saved alongside the original, using
  `ifcpatch`'s `ExtractElements.Patcher`. The console output then directs
  the user to attach this file when filing a GitHub issue, lowering the
  barrier to reporting geometry bugs with a reproduction case.

## Test plan

- [ ] Open a drawing that generates successfully — operator should report
  "No errors creating drawing, nothing to investigate."
- [ ] Open a drawing with a known bad element — operator should narrow down
  to that element and print its id/class/name in the summary.
- [ ] Confirm multiple independently bad elements are all reported (not just
  the first).
- [ ] Run with "Export Failing Elements to IFC" enabled — verify a
  `_failing_elements.ifc` file is created next to the original IFC and
  contains the identified elements with valid spatial structure.
- [ ] Run with no active camera / no IFC loaded — verify graceful
  `{"CANCELLED"}` with an informative error message.

PR #8065, opened 2026-05-15
Fix #8046 - Entity xs:restriction name crashes with AttributeError on…

### LINKED ISSUE #8046: AttributeError: 'Restriction' object has no attribute 'endswith' when an element fails an entity requirement expressed as xs:restriction
### Bug Description

**Description**
When an element matches the applicability but fails an entity requirement expressed as <xs:restriction><xs:enumeration>, ifctester crashes instead of reporting a failure. The <xs:restriction> form for entity names is valid per the IDS 1.0 schema and is covered by existing test cases, but those cases only test the pass path. The fail path is untested and broken.

**Steps to reproduce**
- Use the attached IDS file against any IFC2X3 model containing at least one IfcColumn
- Run validation via ifctester

### Attachments

[ifctester-restriction-bug-demo.txt](https://github.com/user-attachments/files/27587358/ifctester-restriction-bug-demo.txt)

The IDS (Renamed extension to .txt) has a broad applicability (walls and columns) and a narrower requirement (walls only). Columns match the applicability but fail the requirement, triggering the crash.

**Two issues in facet.py:**
- Restriction.__eq__ has no explicit return True at the end of the method. When all constraints pass, it returns None implicitly. This means a comparison like inst.is_a().upper() == self.name evaluates to None (falsy) even for a matching element, causing is_pass to be falsy when it should be True.
- Because is_pass is falsy, Entity.__call__ enters the IFC2X3 type-inference block and hits line 237 (self.name.endswith("TYPE")), which raises AttributeError because Restriction has no endswith method. Line 240 (f"{self.name}TYPE") would produce an incorrect string for the same reason.

**Suggested fix:**
Add return True at the end of Restriction.__eq__ so matching constraints produce a truthy result. Additionally harden Entity.__call__ against non-string names:
# Restriction.__eq__ — add explicit return
def __eq__(self, other):
    ...
    return True  # currently missing, falls through to implicit None

# Entity.__call__ — extract name_str before the IFC2X3 block
name_str = self.name if isinstance(self.name, str) else next(
    iter(self.name.options.get("enumeration", [])), ""
)
if (
    not is_pass
    and inst.file.schema == "IFC2X3"
    and not name_str.endswith("TYPE")
    and (element_type := ifcopenshell.util.element.get_type(inst))
):
    is_pass = element_type.is_a().upper() == f"{name_str}TYPE"

### Debug and Error Output

```true
Error

```

bsmithuk:
Fix #8046   - Entity xs:restriction name crashes with AttributeError on IFC2X3 fail path 

When an IDS entity requirement uses xs:restriction (e.g. an enumeration of entity names) and an IFC2X3 model element fails that requirement, Entity.__call__ entered the IFC2X3 type-inference block and called self.name.endswith("TYPE"), which raises AttributeError because self.name is a Restriction object, not a str. Fix by guarding the IFC2X3 block with isinstance(self.name, str), consistent with the existing guard in Entity.filter(). IFC2X3 type inference is intentionally skipped for Restriction names as filter() does the same and adds regression tests covering the IFC2X3 + Restriction pass and fail paths.

### DIFF
--- a/src/ifctester/ifctester/facet.py
+++ b/src/ifctester/ifctester/facet.py
@@ -234,6 +234,7 @@ class Entity(Facet):
         if (
             not is_pass
             and inst.file.schema == "IFC2X3"
+            and isinstance(self.name, str)
             and not self.name.endswith("TYPE")
             and (element_type := ifcopenshell.util.element.get_type(inst))
         ):
--- a/src/ifctester/test/test_facet.py
+++ b/src/ifctester/test/test_facet.py
@@ -240,6 +240,23 @@ class TestEntity:
         wall3 = ifcopenshell.api.root.create_entity(ifc, ifc_class="IfcWall", predefined_type="BAZFOO")
         run("Restrictions an be specified for the predefined type 3/3", facet=facet, inst=wall3, expected=False)
 
+        restriction = Restriction(options={"enumeration": ["IFCWALL"]})
+        facet = Entity(name=restriction)
+        ifc = ifcopenshell.file(schema="IFC2X3")
+        run(
+            "Restriction entity names do not crash on IFC2X3 fail path 1/2",
+            facet=facet,
+            inst=ifc.createIfcWall(),
+            expected=True,
+        )
+        ifc = ifcopenshell.file(schema="IFC2X3")
+        run(
+            "Restriction entity names do not crash on IFC2X3 fail path 2/2",
+            facet=facet,
+            inst=ifc.createIfcColumn(),
+            expected=False,
+        )
+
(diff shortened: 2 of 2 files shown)


Moult:

Thank you for the report and the fix. Your guard is right that the crash comes from string methods on a `Restriction`, and your two test cases are the exact fail and pass paths. #9540 takes the fix one step further so a restriction can still pass through the IFC2X3 type mapping rather than skipping it, and extends `filter()` the same way. Closing this in favour of #9540, with credit for the report in #8046.

Issue #1320, opened 2021-02-13
BlenderBIM: parametric dimensions strings


theoryshaw:
Related to #1153 and I'm sure it's on the roadmap, but thought I'd start an issue for atomized discussion and tracking. 

Undoubtedly part of _**making construction drawing generation like, really, really awesome**_, would be to have parametric dimensions strings. :)

Something, perhaps, similar to the animation below.  
Perhaps a future roadmap could even include using the dimension string to enter in new lengths that move the object.

Since BlenderBIM's dimensions are part of the IFC file, it would seem to make sense to figure out a way that these parametric connections could be retained upon import as well. 

![2021-02-13_11-01-52](https://user-images.githubusercontent.com/507113/107856170-1069a300-6dec-11eb-8eff-6c8df130068a.gif)

theoryshaw:
somewhat related to #1142

qwiglydee:
### Desired usecases and workflows 

Need to be specified. 
The animation is cool, but not quite comprehensive.

### My vision of functionality

1. measurement annotations:
    - indicate some dimension of existing geometry (of one or multiple objects), which is a calculated value
    - have a lable and a geometric shape consistent with the dimension
    - should track changes of a created geometry in real time, including edit mode
    - should track changes of a generated geometry
1. parametric annotations:
   - indicate parameter of some generator (modifier/nodes/whatever)
   - changing their label value should trigger regeneration
   - the shape of annotation should be consistent with resulting geometry (just like for normal annotation)
   - might serve as handles to change parameter visually (like Blender's "control bones")
1. annotations should exist as objects in scene so that users can grab an manipulate them 
    - although it only makes sense to change the distance or orientation
    - or to move control handles
1. they should be storeable and retrievable in IFC file
 
### Technical considerations

The geometry data of objects are somewhat ephemeral - we cannot reference to vertices/edges/faces or track changes of them.
(The MeasureIt addon breaks or screws up after some geometry changes and it doesn't work with modifiers).

However, it's possible to use vertex groups or custom vertex attributes to mark some key vertices to denote they're need an annotation. They are more stable with respect to modifications.

The best way to recalculate all required geometry of annotations seems to be via depsgraph_update handler triggered every time something changes in scene. This is better than doing it on every frame (as MeasureIt does). But that might not work well for updating real mesh objects in scenes (as they currently exist in BlenderBIM).

[...]

PR #8083, opened 2026-05-19
Parametric dimensions

theoryshaw:
addresses #1320

Still some things to iron out, but putting this out there to test on BonsaiPR.

https://github.com/user-attachments/assets/f67afeee-a203-4189-ba0a-1e29b553d3f8


theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/5e54af71d9217d74e40b4dd6329d97e44a2fb076/logs/summaries/2026-06-07-rebase-parametric_dimensions.md

theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/dfba54470ee3bf0d4aaaf92cb74769e318918d31/logs/summaries/2026-06-07-PR-8083.md

theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/b3868e8f5536516b73b49cf0a43088edbae70b44/logs/summaries/2026-06-28-PR-8083.md

Plae-Blue-Dot:
@theoryshaw is this accessible on BonsaiPR? Would love to try it out

theoryshaw:
@Plae-Blue-Dot Hi, let me try to fix the current conflict with v0.8.0 so it works again with BonsaiPR.

theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/a4ee7fa6007b74c39dc37e4d5d45c8e1bbea42e3/logs/summaries/2026-08-04-rebase-parametric_dimensions.md

theoryshaw:
@Plae-Blue-Dot I think this one has it now: https://github.com/falken10vdl/bonsaiPR/releases/tag/v0.8.6-alpha2608042349

In general, i would be hesitant about using BonsaiPR for any production work, but for testing, i think it will be fine.

Plae-Blue-Dot:
> @Plae-Blue-Dot I think this one has it now: https://github.com/falken10vdl/bonsaiPR/releases/tag/v0.8.6-alpha2608042349
> 
> In general, i would be hesitant about using BonsaiPR for any production work, but for testing, i think it will be fine.

Absolutely just to test for now, I think this + copy annotations changes the game

Thanks for getting it going again Theoryshaw!
I loaded the latest BonsaiPR today and it doesn't seem to be active there, I'll have a look again on the next update

theoryshaw:
@Plae-Blue-Dot you actually have to manually download & install the build from the link above. 
Here it is again:  https://github.com/falken10vdl/bonsaiPR/releases/tag/v0.8.6-alpha2608042349

At any one time, BonsaiPR might have (3) different builds...

| Build   | Merge order                    | In a conflicting pair, the winner is… |
|---------|--------------------------------|---------------------------------------|
| `[asc]` | lowest PR number first (oldest) | the **lower**-numbered PR             |
| `[desc]`| highest PR number first (newest)| the **higher**-numbered PR            |
| `[upd]` | most-recently-updated first     | the PR touched most recently          |

When you update BonsaiPR from 'Get Extensions' tab in Blender, it pulls the absolute latest build, whether it's the [asc], [desc], or [upd] one. 

For this particular PR, you wanted to use the [upd] one, so the PR doesn't conflict with any later updated PRs. 
...

To see if a particular PR made a particular build check here: https://github.com/falken10vdl/bonsaiPR/releases

Each release then links to a report like this https://github.com/falken10vdl/bonsaiPR/blob/main/automation/reports/archive/v0.8.6-alpha2608042349.md, where you can verify which PR made it, or not. 

...

@falken10vdl I wonder if there's anything in the 'Get Extensions' UI that would allow people to toggle which build they wanted.

falken10vdl:
I doubt It is possible to parametricallly select. Letme double check but It was not possible some months ago when We implementes It. We could eventually create bonsaiPRup, bonsaiPRdesc, bonsaiPRasc as workarround?

Plae-Blue-Dot:
> ensions' UI th

Thanks Theoryshaw, I tried again and it works! 
for some reason the download number: 
image
and version number in blender:
image
are different, threw me off a bit
 
 The dimensioning is great!
 the editing without going into item mode/ edit mode is streets ahead, and way more intuitive, 
 It's way quicker and There's less chance (or zero chance rather) or corrupting the dimension 
Also the instant preview after clicking is much clearer

 I'm going to compare it to the normal editing and see the time difference,
 Would it be possible to add a plus icon or something to add a value to an existing dim?
 
 And is this planned to be included in the 0.8.6 release? I think it is a super important feature

theoryshaw:
> Would it be possible to add a plus icon or something to add a value to an existing dim?

yes i would like to add that eventually. 

>And is this planned to be included in the 0.8.6 release? I think it is a super important feature

I'm not sure what the maintainers have in mind or bandwidth to review. 

Over the next couple weeks I might look into creating a system like this https://community.osarch.org/discussion/comment/29896/#Comment_29896
Where powerusers can curate their own builds.  That way other users can use, and have a little more confidence to use them for production, and also, to compose their own, etc.

Plae-Blue-Dot:
Ah I think I see, I hadn't realised how fragmented these features are, stacked on other features not yet merged. Bonsai's architecture and the reviewing process are the bottleneck and drawing is not a priority there. While users used to how bonsai works can manage, this is too much of a hurdle for newcomers who would benefit the most.

I tried out the dimensioning a bit more and the dims can't snap to points since the parametriscism uses faces to remain perpendicular, any wall that has an angled edge can't be dimensioned at the moment:
image
I could get around it by extending the wall and breaking the join, dimensioning and then 'bake to static' which is the old/ current dimensioning method
image

So thanks theoryshaw! This is a great feature, I see no downside to updating the dimensioning to this from a user point of view. I think your approach is good regarding the PR's, I'll keep an eye out!

theoryshaw:
use tab to enter different selection modes, that is, face/layer/edge/vertex.

Plae-Blue-Dot:
> use tab to enter different selection modes, that is, face/layer/edge/vertex.

You've thought of it all! Very nice!

theoryshaw:
some added functionality: https://community.osarch.org/discussion/comment/29924/#Comment_29924

BIMvoice:
@theoryshaw a user hit a crash in a BonsaiPR build that traces back to this branch, reported in #9290. It looks like a small slip rather than anything in the design.

In `AddAnnotation._execute` (`src/bonsai/bonsai/bim/module/drawing/operator.py`) the diff changes

```python
enable_editing=True,
```

to

```python
enable_editing=object_type not in ("ELEVATION", "SECTION"),
```

In that scope the local is `props.object_type`, so the bare `object_type` raises `NameError: name 'object_type' is not defined`. It fires for every annotation object type, so creating an ordinary line fails too, not just elevations and sections. The reporter's traceback in #9290 shows it landing there.

`props.object_type` should sort it. The other `object_type` uses in the diff are inside `core.add_annotation`, where it is a real parameter, so I think this is the only occurrence of the problem.

One other small thing while you are in there: there are a few leftover debug prints in the core module, `print(f"[SECTION] core.add_annotation called: ...")` and similar, which you probably want to drop before this merges.

Issue #7408, opened 2025-11-24
Blender freezes when opening IFC file with Bonsai


jijidom:
### Description

When I open a specific IFC file in Blender using the Bonsai add-on, Blender completely freezes (UI not responding). I have waited for a long time, but it never recovers, and I have to kill the Blender process manually. No error popup is shown.

Other IFC files sometimes work, so it looks like this may be related to certain models / geometry.

### Environment

- OS: Windows
- Blender: 4.5.3 LTS
- Bonsai version: 0.8.3-post1

### IFC file details

- IFC schema: IFC4
- File size: 16KB
- Source of the model: custom generator

### Steps to reproduce

1. Start Blender with a fresh configuration (factory settings) and enable only the Bonsai add-on.
2. Go to **File → Import → IFC (Bonsai)**.
3. Select the IFC file: `[db792773-fb89-4137-91d7-58a5c4ef202f.ifc]`.
4. Click **Import**.

### Actual behavior

- Blender UI freezes and becomes unresponsive.
- I do **not** see any clear error in the Blender UI.
- I have to kill Blender from the OS task manager / `kill` command.

### Expected behavior

- The IFC should either:
  - import successfully, or
  - show an error message explaining what went wrong,
  instead of freezing Blender indefinitely.

### What I have tried

- Reset Blender to factory settings and re-enabled only Bonsai.
- Tried other IFC files: some smaller/simple models import correctly, this specific model always freezes.

Thanks in advance for your help!

### Attachments

[db792773-fb89-4137-91d7-58a5c4ef202f.txt](https://github.com/user-attachments/files/23700764/db792773-fb89-4137-91d7-58a5c4ef202f.txt)

### Debug and Error Output

```true

```

jijidom:
### Additional Information
﻿
I have tested this again with the latest released Bonsai **0.8.4**, but the issue still occurs. The Blender freeze happens in exactly the same way. There is no improvement or change in behavior compared to previous versions.
﻿
This suggests that the freeze may not be related to an older Bonsai version and might instead be caused by:
- a specific geometry or topology pattern in the IFC file,
- a performance regression in the importer,
- or a problem inside the hybrid-cgal-simple-opencascade kernel (just speculation).

aothms:
> a specific geometry or topology pattern in the IFC file

IfcConvert has the same issue, OpenCASCADE gets stuck building the sweep over polyline `#171`, I'll have a look.

aothms:
[...]

PR #8118, opened 2026-05-29
Fix sign of temporary offset restore in sweep_along_curve

tiagoazvdo:
## Summary

Fixes the sign of the temporary-offset restore in `OpenCascadeKernel::convert(sweep_along_curve)`.

The temporary-offset workaround introduced in #7408 (commit `bd57cc8735`) subtracts the directrix centroid (`mean`) from the curve points so the sweep is built near the origin (OCCT is unstable for geometry far from origin). After building, it must translate the result **back** by `+mean` to restore the original position. The restore negated the sign:

```cpp
trsf.SetTranslation(gp_Vec(-mean.x(), -mean.y(), -mean.z()));  // before
trsf.SetTranslation(gp_Vec( mean.x(),  mean.y(),  mean.z()));  // after
```

`Move(-mean)` places the swept solid at `-mean` (mirrored through the origin) instead of its true location.

## Impact

The buggy branch only runs when:
- the directrix is a pure polyline (`is_polyhedron()`), and
- its centroid is more than 100 m from the origin (`mean.norm() > 1e2`).

So models centered near the origin are unaffected, which is why this went unnoticed. But IFC files that keep **absolute site coordinates** (common in Revit/ODA exports) render affected swept solids — reinforcing bars, pipes — at a mirrored phantom location hundreds of meters away from the rest of the model.

## Testing

Reproduced with a real reinforced-concrete beam IFC (rebar as `IfcSweptDiskSolid` with composite-curve directrixes, exported from Revit via ODA, absolute site coordinates):
- **Before:** 18 of 54 bars (the straight ones — polyline directrix) placed in a phantom cluster ~300 m from the other 36.
- **After:** all 54 bars co-located correctly.

Straight bars/stirrups hit the polyline branch; bent stirrups (whose directrix contains arc segments) are not `is_polyhedron()` and were already correct — consistent with the diagnosis.

### DIFF
--- a/src/ifcgeom/kernels/opencascade/sweep_along_curve.cpp
+++ b/src/ifcgeom/kernels/opencascade/sweep_along_curve.cpp
@@ -300,7 +300,11 @@ bool OpenCascadeKernel::convert(const taxonomy::sweep_along_curve::ptr scs, Topo
 
 	if (applied_temporary_offset) {
         gp_Trsf trsf;
-        trsf.SetTranslation(gp_Vec(-mean.x(), -mean.y(), -mean.z()));
+        // Restore original position: add back the mean subtracted from the
+        // directrix points above. Previously negated, which placed the swept
+        // solid at -mean instead of its original location for geometry far
+        // from the origin.
+        trsf.SetTranslation(gp_Vec(mean.x(), mean.y(), mean.z()));
         result.Move(trsf);
     }
 


aothms:
Many thanks. This fix indeed appears correct to me. Probably the fix was only tested on a model with a singular element and never got noticed.

PR #7930, opened 2026-04-11
Ifcviewer - an ultra fast ifcopenshell viewer and app

Moult:
Edit: blockers for release

- [x] Fix mutex issue with RDB parallel https://github.com/IfcOpenShell/IfcOpenShell/pull/7534#issuecomment-4308484504
- [x] Get automatic builds working
- [x] Fix RDB read only issue causing crash
- [x] Geometry correctness benchmark matrix to make sure manifold / passthrough / etc is actually verified / correct in new datamodel branch (vs old 0.8.0 branch)

From https://community.osarch.org/discussion/3386/addressing-some-core-ifcopenshell-issues

> A desktop viewer. Right now, Blender is really not optimised for a viewer and people don't realise how fast IOS really is because Blender itself imposes a crazy overhead for loading Blender meshes. We need a way on the desktop to view 50 models for simple coordination.

So the high level proposal is to make:

 1) A big viewer tool. Basically Bonsai but not for authoring, it's for coordination. It's for viewing, BCF issue tracking, clash detection, viewing with drawings, connection to CDEs, all the stuff that we currently have to do with proprietary software (BIMVision, Revizto, ...).
 2) A framework for other people to build their own tools. So if you want to build your own kiosk app, tablet app, desktop app, phone app, ... you have a starting point. Kind of like the non-web equivalent of "npm install web-ifc-component"

My general impression is that all the strategies of making a high performance viewer is well established and heavily documented (i.e. well-trained AI). So I'm fairly confident we can quickly get to a state where all the basic tricks are implemented with vibe coding and start to approach the more cutting edge like whatever nanite is doing.

Also, I think the general approach of a Qt app with dockable windows and standard properties viewers, checkboxes, settings window, status bar, etc is also very well established and AI will do a pretty good job in a greenfield situation, so I hope to vibe away.

I hope to get to a milestone where the bare viewer is ready with a good foundation. After that, hopefully with the work done with IfcZero, datamodel refactoring, new kernels, etc especially with Python utils upstreamed to C++, I hope especially with the guidance from the Bonsai data classes and UI layer, to replicate some of the more important readonly properties. I think this somewhat low risk, so long as the AI helps with all the Qt stuff, I'll be able to look after the IFC-related domain logic.

Obviously, don't merge :) Comments very very welcome.

See README.md for latest full explanation:

https://github.com/IfcOpenShell/IfcOpenShell/tree/ifcviewer/src/ifcviewer

(Also ignore the Python stuff and build hacks that's just a mess I made in my local dev)

PR #8139, opened 2026-06-04
Ifcviewer wgpu

Moult:
Do not merge

This is a continuation of the work done in https://github.com/IfcOpenShell/IfcOpenShell/pull/7930 and supersedes https://github.com/IfcOpenShell/IfcOpenShell/pull/8057 (which was an original continuation prior to the webgpu redo).

Things that need review before merging:

 - all changes in src/serializers
 - all changes in cmake/
 - this race condition fix https://github.com/IfcOpenShell/IfcOpenShell/pull/8057/commits/19bff92a47ea2397af3c94cfb3dccbe1f75f4def
 - all the updated github workflows, including the changes to win/ and nix/ to try to get things to build including (note this decreases the Linux size down to much more reasonable amounts on par with other OSes) https://github.com/IfcOpenShell/IfcOpenShell/pull/8057/commits/de7520418b7208bba31272bccc72831ffc63e1c2 https://github.com/IfcOpenShell/IfcOpenShell/pull/8057/commits/a91b1da28b926e3b1ce697463ed9d9be5c90b8b3

This is a BIG pr, but there is reason to the bigness. It contains:

 - Length, area, volume measurement
 - Marquee selection
 - RDB conversion tools exposed in the UI
 - Model group management in the viewer
 - Removal of OpenGL viewport rebuilt with WebGPU (which is more verbose than OpenGL)
 - With WebGPU, entirely new streaming approach (no more one big VBO, it's chunked spatially)
 - MacOS support
 - Viewport overlay code to generically draw text, rects, lines, etc (used in marquees, length tool, etc)
 - A bunch more Qt UI, including quite complex ones like tables of georeferencing coordinates, settings and tabs, UI components and widgets ...
 - Getting builds to work
 - Rename to Bonsai Viewer
 - A cloud connection app in Rust (FLTK UI, REST API, RPC protocol, OAuth, ...)
 - Developer docs being written, an entire new Sphinx docs for src/bonsaiviewer/docs/


Moult:
TODO:

- [x] rebase
- [x] clean up stuff like wgpu mem probe etc
- [x] set licensing properly (GPL for bonsaiviewer, LGPL for all viewercode)
- [x] update websites to reflect new bonsaiviewer
- [ ] add coming soon bonsaiweb page
- [ ] update ifcopenshell docs to talk about new ifcviewer capabilities
- [x] implement basic attribute / property / spatial tree viewing for minimum viable product
- [x] add docs about ifcview format details
- [ ] add docs about ifcviewer performance flow chart / considerations
- [ ] build ifcview conversion into ifcconvert (Moult to have an attempt)
- [ ] python bindings to convert to ifcview (Moult to have an attempt)
- [ ] thomas to review those things at original post above
- [ ] let thomas know when safe to merge into datamodel

Once merged into data model...

- [x] confirm that ifcopenshell python works for mac (i.e. full build matrix we need for bonsai)
- [ ] update bonsai to work with datamodel branch

Aimed release end of this month - future targets include mainstreaming manifold kernel, ifccsv / ifctester / etc in bonsaiviewer, merging iosweb + ifcviewer, etc.

aothms:
I had a look at the per mesh quantization. Doesn't seem this is an issue anymore in modern wgpu pipelines. 

```
@group(1) @binding(0) var<storage, read> vertices:      array<u32>;
@group(1) @binding(1) var<storage, read> meshes:        array<MeshQuant>;
@group(1) @binding(2) var<storage, read> instances:     array<InstanceRecord>;
@group(1) @binding(3) var<storage, read> indices:       array<u32>;
@group(1) @binding(4) var<storage, read> visible_draws: array<VisibleDraw>;
@group(1) @binding(5) var<storage, read> prefix_sums:   array<u32>;
@group(1) @binding(6) var<uniform>       u_model:       PerModel;
```

(ViewportCore.cpp)

I have some reading up to do on what this means exactly, but it seems pretty optimal in terms of storage - where this is just stored and looked up once per mesh. Going the route of quantization in glTF where the vert positions are just in (0,1) and then the dequantization transform is factored into the regular object transform is also complex because we depend on this transform for the normals as well.

Things really look different from the 10y ago when I toyed with opengl

aothms:
I also had a look at the ifcview format. Not everything is immediately obvious to me.

afbeelding

I wonder if interleaved ascii JSON inside the binary blob isn't more future proof, extensible and maybe even more efficient instead of the string pool [maybe not for entity type names if they are deduplicated] (alternatively, maybe zero-terminated strings in the string pool would also be better then you can skip the length and have compatibility with c-style strings without having to make copies).

afbeelding

I think I'd find GLB's approach also a bit more flexible. So define a binary buffer and accessors over that with semantics instead of relying on predefined offsets and fields. So then you could have arbitrary lod levels and maybe even add an error metric in there over time to better decide which lod level for a chunk to pick based on camera distance (not sure how that works now).

afbeelding

And then I don't understand why there are three arrays. I understand there is objects and meshes, but which of these are InstanceCpu and SidecarChunk?

---

An orthogonal or maybe even opposite design concern could be memory mapping. Currently the code relies a lot on std::string/vector that point to heap-allocated data - these would need to become std::span (?). Also, compression is a bit of an orthogonal idea to memory mapping, not necessarily incompatible. Maybe there could be two formats .ifcview and .ifcviewc where the first is more suitable for desktop usage where you can swap in and out files without latency and the the second is more suitable for web where there is no mmapping and network latency is the deciding factor. With a clever design the data and code could be more or less identical, just a little bit more metadata on whether a section is compressed or not. 
 Maybe I'm being naive

aothms:
I added a couple of things, most notably the psets

You can see how it can be used in IfcParseExamples

```C++
element_properties props = get_psets(element);

for (auto& ps : props) {
    std::cout << ps.first << std::endl;
    std::cout << std::string(ps.first.size(), '=') << std::endl;
    size_t max_key_len = 0;
    for (auto& p : ps.second) {
        if (p.first.size() > max_key_len) {
            max_key_len = p.first.size();
        }
    }
    for (auto& p : ps.second) {
        std::cout << p.first << std::string(max_key_len - p.first.size(), ' ') << ":" << p.second << std::endl;
    }
    std::cout << std::endl;
}
```

Which prints something like this

```
Pset_WindowCommon
=================
FireRating:FireRating
IsExternal:true
Reference :M_Casement:819mm x 759mm
id        :23750
```

Codex invented a bit much typing around the return values because of the flexibility that python offers. Not really to my liking but there are no truely easy answers there.

I also had Codex generate statically typed variants to the helper functions you already generated. At the call site nothing should have changed (except that I renamed ifcutil -> helpers, both are not great, but I really don't want ifc prefixes for code anymore).

One important thing that I forgot is that the schemas became plug-ins and this makes the static dispatch more or less impossible. I'm not entirely sure how this is solved. Probably this means that for the viewer the schema plug-ins can no longer be true plug-ins (not nice for the web) unless we create interfaces around that or use the same mechanism for hierarchy_helper. Tbh my understanding has also been a bit eroded here due to AI usage. If all fails we can always revert to the latebound helper implementations you had. Or we implement an abstract schema that can be used on top of all concrete schemas and make that an interface by which the util functions be static and generic at the same time.

Moult:
@aothms @Andrej730 I've brought this branch up to speed with the changes made on v0.8.0. So I _think_ this branch is now ready to become the new v0.8.0 (or v0.9.0?).

Here's a summary because it took me a while to get my bearings on the lots of changes happening everywhere.

Firstly, I've checked using `git log origin/datamodel-v1.0 ^ifcviewer-wgpu` (i.e. show all commits in datamodel excluding all commits in ifcviewer-wgpu) that this branch contains all the changes from the datamodel branch. The command gives an empty result which is awesome - so I think the datamodel branch can be dropped.

For the record, the last commit on ifcviewer-wgpu before I began this exercise was 018695a, and the last commit merged from v0.8.0 --> ifcviewer-wgpu before I began this exercise was therefore 206cd6bbe. So what I had to merge was `git log 206cd6b..origin/v0.8.0 --oneline` (248 commits).

Unfortunately those 248 commits included things that make conflict resolution a bit painful or risky like:

 1. Updates in files that have since been renamed (e.g. IfcParse->parse.cpp)
 2. Formatting / lint / typing on things that have been totally changed
 3. Generated files which don't make sense to merge
 4. Python or C++ code that uses things that have changed (e.g. Logger->logger* and dropping use of wrapped_data in Python) which might merge, but will not compile or break on runtime
 5. Stuff like [these 7 commits](https://github.com/IfcOpenShell/IfcOpenShell/pull/8608/changes) which have chunks of C++ that need to be looked at carefully and tested (e.g. with visual SVG result) when porting over to this branch.

So rather than a merge, I decided to chronologically cherry pick starting from after `206cd6b` from v0.8.0 --> ifcviewer-wgpu. Some apply cleanly, but others I decided either to skip (e.g. just formatting or code gen - we can rerun formatters / codegen at the end, or things that were already implemented in this branch such as version bumps). And for some, I manually ported (and squashed into a single commit) and tested such as the SVG edge feature.

So the git incantation to check for any commits in v0.8.0 that aren't in ifcviewer-wgpu is:

```
$ git log origin/v0.8.0...ifcviewer-wgpu --oneline --cherry-pick --left-only --no-merges
# the triple dot ... means commits in either branch but not both
# cherry-pick means exclude identical commits that are cherry picked (since cherry picked commits have different hashes)
# left-only means commits present in origin/v0.8.0 and not in ifcviewer-wgpu
# no-merges means to exclude merge commits
```

This results in 70 commits. These commits therefore were either skipped (e.g. formatting or not applicable or already present) or ported manually (e.g. SVG edge feature), or had some conflict that needed resolving (usually very minor) during the cherry picking process.

I have compiled locally to catch any build issues. Things I haven't done:

 - Run all test suites
 - Run Github actions
 - Rerun and check all code formatters / regen
 - Investigate stuff that went missing due to file renames. For example I found that ee2b357 was merged into this branch, but because that commit edited IfcParse.cpp, the merge deleted the file and so the actual fix was lost (`git show -m 8c9c3cde2 -- src/ifcparse/IfcParse.cpp` shows the merge commit deleting IfcParse.cpp). I only stumbled across this one and it's pushing my knowledge of Git here to track it down.

Moult:
OK I've migrated all the type check fixes, rerun formatters, and fixed main OS builds so fingers crossed things are green (at least related to those things).

@aothms can you take a look at these commits, I'm not sure if they got lost in the merge:

 - f3dcf0b53
 - d2381ad6c
 - dcebf23af
 - a9d677687

aothms:
Many thanks @Moult. I personally know how painful this is. This is the approach I use myself as well. Replaying commit by commit is also the only way for me to retain some context.

- https://github.com/IfcOpenShell/IfcOpenShell/commit/f3dcf0b5395f314de54aa428308a4444b9628c58 Should be fine - now that instances are always created as part of a file - this commit was to fix some failing tests (I think the subcontexts that were not properly purged because they had more inverses registered then were actually present in the model)
- https://github.com/IfcOpenShell/IfcOpenShell/commit/d2381ad6c627f9629b4fa8e704164379b0786a48 Should be fine - seems that the logic is no longer to strip away the leading/trailing `.` `"` but just not to copy them over https://github.com/IfcOpenShell/IfcOpenShell/blob/5739a24808b3180b57e92c728312bb9db30c2ba7/src/ifcparse/parse.cpp#L349
- https://github.com/IfcOpenShell/IfcOpenShell/commit/dcebf23af85b36f22950ea1d880e0383159c5ff6 - I think this still applied, this was about initialisation order basically, that the header is constructed before file.storage is created, so if the header creates instances during construction they are based on a garbage storage. Added 69dd3a060
- https://github.com/IfcOpenShell/IfcOpenShell/commit/a9d67768754f55b8643b17aa8cf6c1def0fac390 - Missing. Added in 45eeb0b4e

---

Do you think we can ask AI to rebase/rework all the open PRs to 0.9, do a final release of 0.8 and then make the switch?

Moult:
> Do you think we can ask AI to rebase/rework all the open PRs to 0.9, do a final release of 0.8 and then make the switch?

I would prefer to switch to v0.9.0 (so long as all test cases pass, which I haven't yet run) and then rework the open PRs to v0.9.0 (the pure AI PRs can use AI to rework, but the human PRs I think we should leave to the responsibility of the submitter or the maintainer to figure out what to do).

Things left to do:

 - Make sure all tests pass, I haven't yet run them, maybe I can start tonight
 - Build WASM action is failing, maybe that's OK, maybe we just wait until Sayan's work gets merged in the next release (perhaps within a month?)
 - ci-ifcwrap-standalone is failing
 - ci is failing
 - Build Bonsai Viewer Autodesk Connector is failing
 - Finish updating the website with new content / screenshots

There's so much more left to do too but those are probably the showstoppers for a release. The rest can happen after I think.

Moult:
For the failing tests that rely on `ifcopenshell.create_entity` (which @Andrej730 also stumbled across) any ideas what to do? There are a few other failing ifcopenshell python tests like `f.create_entity('IfcDescriptiveMeasure', 'asdf') == 'asdf'` is now false, not sure the intended behaviour there.

aothms:
Here's proposal for the easy way of just a global file dict that we can create instances in. This is what I said to @Andrej730 on Matrix [added one bit]

---

The reason is twofold: (a) rocksdb, you cannot create a rocksdb instance without a file/database context - and it would be really silly to have to create everything in memory first and then add it to the file having to convert the storage back-end for everything you create (b) memory ownership: entity instances are now a weak_ptr so that deleting a file reliably cleans up memory without crashing. You cannot have a weak_ptr without a shared_ptr. `file` owns the shared_ptr. I discussed this also with @RickBrice we can create our own hybrid ptr that can be either shared or weak or use a std::variant so that instances can be self-owned until a file takes owership. But this was a lot of overhead/complexity for something that only half-worked anyway, e.g inverses also depended on a file context.

I think there are two options. Temporary file as you said - we can also make it a bit more special, have one global temp file per schema that is somehow managed automatically so that ifcopenshell.create_entity() still works. Or something pure python like {"type": "IfcCar.Pnt", "Coordinates": [1.0, 2.0, 3.0]} and make this a bit more seamless with file.add() and entity_instance.eq. It would be nice to have entity instances as something "pickleable".

---

Note that with the global file object I'm not entirely sure whether the semantics of `__eq__` and friends are still ok to have the express rules pass. Could be that this depended on the instances being ownerless and that instances from distinct files never compared equal. But I need to run, just wanted to leave this in here for discussion. 

---

Given this design it would be nice to have a ifcopenshell.file.clear() method. Or otherwise have a method to throw away the global module-level files so that we do not leak memory.

---

https://github.com/IfcOpenShell/IfcOpenShell/pull/8139/commits/6bcfb175a3bb1b6c165186690f727100165a2dd9

Moult:
Most tests are starting to turn green, the biggest bunch left that I'm aware of is `make test-bim` where there's a bunch of failures. I'll slowly work through them.

aothms:
ci is back to running the tests at least

aothms:
@Andrej730 any particular reason for https://github.com/IfcOpenShell/IfcOpenShell/pull/8139/commits/b6fb5e6cdc68925c13aa1f06ac8686ceb16bf8d7 file.header.assign(header&) is a recent addition but I think it's useful also from the python side?

aothms:
I've got the C++ documentation generation unblocked, went through a bunch of style unification things on the geom side (hope I'm not in the way), fixed a couple tests.

aothms:
Got closed automatically because I just now renamed the branch to v0.9.0

PR #8141, opened 2026-06-04
Fix CGAL 6.x build: add Point_d_4d_Less comparator for std::map

brunopostle:
CGAL 6.x deleted operator< from Point_d, so std::map<Point_d, ...> no longer compiles.

This pull request adds a custom lexicographic comparator and updates the three affected maps in snap_halfspaces and snap_halfspaces_2.

This is a claude-suggested fix

### DIFF
--- a/src/ifcgeom/kernels/cgal/nef_to_halfspace_tree.h
+++ b/src/ifcgeom/kernels/cgal/nef_to_halfspace_tree.h
@@ -116,6 +116,18 @@ template <typename Kernel>
 using plane_map = std::map<typename Kernel::Plane_3, typename Kernel::Plane_3, PlaneLess<Kernel>>;
 // using plane_map = std::unordered_map<typename Kernel::Plane_3, typename Kernel::Plane_3, PlaneHash<Kernel>>;
 
+// Lexicographic comparator for CGAL Point_d (operator< is deleted in CGAL 6.x)
+struct Point_d_4d_Less {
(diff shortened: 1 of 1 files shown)


src/ifcgeom/kernels/cgal/nef_to_halfspace_tree.h
```diff
@@ -116,6 +116,18 @@ template <typename Kernel>
 using plane_map = std::map<typename Kernel::Plane_3, typename Kernel::Plane_3, PlaneLess<Kernel>>;
 // using plane_map = std::unordered_map<typename Kernel::Plane_3, typename Kernel::Plane_3, PlaneHash<Kernel>>;
 
+// Lexicographic comparator for CGAL Point_d (operator< is deleted in CGAL 6.x)
+struct Point_d_4d_Less {
+	using Point_d = CGAL::Epick_d<CGAL::Dimension_tag<4>>::Point_d;
+	bool operator()(const Point_d& a, const Point_d& b) const {
+		for (int i = 0; i < 4; ++i) {
```
aothms: It's minor but I prefer to use

std::lexicographical_compare

on the ranges a.cartesian_begin() / end b.cartesian_begin() / end

without the hardcoded 4.

aothms (changes requested):

brunopostle:
Ok, fixed in 2857302, build is ok across three fedora versions (but all using -DCMAKE_CXX_STANDARD=20)

aothms:
Many thanks!

PR #8142, opened 2026-06-04
Add missing standard library includes for self-sufficient headers

brunopostle:
Fixes builds with newer GCC/libstdc++ that no longer provide <cstdint>, <cstring>, <cfloat>, <memory>, <algorithm> etc. transitively.

This was needed because the latest RocksDB 11.x headers require c++20

Also disambiguates visit<> calls in taxonomy.h with the full namespace and casts the character value in IfcCharacterDecoder to uint32_t to silence ambiguous overload warnings.

These are claude-suggested fixes


aothms:
Thanks!

PR #8143, opened 2026-06-04
Fix HDF5 config-mode detection to use shared library when static is absent

brunopostle:
When HDF5 is found via its CMake config file, the code previously hardcoded the hdf5_cpp-static target.

On distributions that ship only shared HDF5 (e.g. Fedora rawhide where the config file was added in a newer package), this caused a link failure. Now checks for hdf5_cpp-static, hdf5_cpp-shared, and hdf5::hdf5_cpp-shared in order, falling back to module-mode discovery.

this is a claude-suggested fix


aothms:
Thanks :) note that we want to get rid of hdf5 in v0.9 since we only use it for caching and we have this new viewer format.

PR #8144, opened 2026-06-04
Support RocksDB shared library and new unique_ptr DB::Open API

brunopostle:
Some distributions (e.g. Fedora) ship only a shared RocksDB that exports RocksDB::rocksdb-shared rather than RocksDB::rocksdb. The CMake target selection now falls back to the shared target when the static one is absent.

Newer RocksDB also changed DB::Open and DB::OpenForReadOnly to take std::unique_ptr<DB>* instead of DB**. IfcFile.cpp uses SFINAE tag dispatch to build against both old and new APIs without version detection.

These are claude-suggested fixes

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -258,10 +258,14 @@ if(WITH_ROCKSDB)
     set(ROCKSDB_LIBRARIES "IFCOPENSHELL_RocksDB")
     target_compile_definitions(IFCOPENSHELL_RocksDB INTERFACE IFOPSH_WITH_ROCKSDB)
     set(SWIG_DEFINES ${SWIG_DEFINES} -DIFOPSH_WITH_ROCKSDB)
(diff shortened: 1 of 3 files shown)


aothms:
Thanks.

Normally we just do preprocessor switches on the version. I think that's a bit simpler to digest. Seems that the unique_ptr open() calls are available since 9.11.0 which is older than the versions used in our build scripts.

> Offer new DB::Open and variants that use std::unique_ptr<DB>* output parameters and deprecate the old versions that use DB** output parameters.
> 
> https://github.com/facebook/rocksdb/releases?page=2

I also prefer to turn the compatibility layer the other way around. Retain a unique_ptr on our end and convert the naked pointer into it for versions pre 9.11. Could you instruct claude adapt these two things?

brunopostle:
Fixed in 99cfa2b, build is ok across three fedora versions (but all using -DCMAKE_CXX_STANDARD=20)

aothms:
perfect thanks

PR #8147, opened 2026-06-04
fix #8146: Remove slab connection when clip FaceSet is deleted

### LINKED ISSUE #8146: Deleting a clip FaceSet item should remove the wall-slab connection
When `bim.extend_walls_to_underside` clips a wall to a slab, it creates an `IfcPolygonalFaceSet` boolean item and an `IfcRelConnectsElements(TOP)` relationship linking the wall to the slab.

If the user enters item mode (`Import Representation Items`) and manually deletes the `IfcPolygonalFaceSet`, only the geometry item is removed. The `IfcRelConnectsElements` remains, so the next call to `bim.regenerate_wall_to_underside` finds the wall still connected to the slab and silently re-creates the deleted clip.

## Expected Behavior

Deleting a clip `IfcPolygonalFaceSet` in item mode should also remove the `IfcRelConnectsElements(TOP)` for that slab, so that `regenerate_wall_to_underside` no longer considers the wall connected to it.

## Steps to Reproduce

1.  Select a wall and a slab → run  **Extend Walls to Underside**
2.  Select the wall → run  **Import Representation Items**  to enter item mode
3.  Delete the  `IfcPolygonalFaceSet`  item
4.  Run  **Regenerate Wall to Underside**

**Result:** The deleted clip comes back.  
**Expected:** The wall is no longer clipped by that slab.

theoryshaw:
When a user deletes an IfcPolygonalFaceSet item in item mode, remove the corresponding IfcRelConnectsElements(TOP) so that regenerate_wall_to_underside no longer finds the slab connected and does not re-create the boolean.

BBIM_Boolean pset data is migrated from a flat list of boolean IDs to a dict mapping str(boolean_id) → slab GlobalId. This enables per-slab disconnect: deleting a FaceSet for slab A removes only slab A's connection, leaving slab B's boolean and connection intact. Legacy pset files (no slab guid) fall back to unmarking only the matched boolean and leaving connections untouched.

### DIFF
--- a/src/bonsai/bonsai/core/model.py
+++ b/src/bonsai/bonsai/core/model.py
@@ -183,7 +183,7 @@ def regenerate_wall_to_underside(
         for slab_obj in slab_objs:
             clip = model.get_slab_clipping_bmesh(slab_obj)
             if clip:
-                model.clip_wall_to_slab(wall, clip)
+                model.clip_wall_to_slab(wall, clip, ifc.get_entity(slab_obj))
         clipped_objs.append(obj)
     if clipped_objs:
         model.reload_body_representation(clipped_objs)
@@ -221,7 +221,7 @@ def extend_wall_to_slab(
             clip = model.get_slab_clipping_bmesh(slab_obj)
             if not clip:
                 continue
-            model.clip_wall_to_slab(wall, clip)
+            model.clip_wall_to_slab(wall, clip, ifc.get_entity(slab_obj))
             model.connect_wall_to_slab(wall, ifc.get_entity(slab_obj))
             did_clip = True
         if did_clip:
--- a/src/bonsai/bonsai/core/tool.py
+++ b/src/bonsai/bonsai/core/tool.py
@@ -673,7 +673,7 @@ class Misc:
 
 @interface
 class Model:
-    def clip_wall_to_slab(cls, element, bm): pass
+    def clip_wall_to_slab(cls, element, bm, slab=None): pass
     def connect_wall_to_slab(cls, wall, slab): pass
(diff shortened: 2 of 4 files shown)


Gorgious56:
Hello. Wondering if maybe it's an XY problem, wouldn't this be solved by an explicit "Disconnect connection" button rather than having to manually enter representation edition to do so ? I don't think it's immediately evident that removing the IfcPolygonalFaceSet item should remove the connection, for example in slanted walls removing the boolean geometry doesn't remove the connection either. Maybe user wants to retain the connection but have a specific geometric behaviour they want to manually bake in that is not the auto-generated one. 

for reference I added in last commit a "disconnect wall" button at wall connections. Maybe we need something like that ?

https://github.com/user-attachments/assets/05be60b6-7b2f-4edd-b24e-71c7648e0f57

theoryshaw:
@Gorgious56 Hi, yes that makes sense!  It probably make sense to have a 'disconnect slab' gizmo/operator as well. Since you might have 1 wall connected to 10 slabs, but you only want to remove 1 slab, not the other 9. 

Since your head's in this world right now (considering your recent commits), any interest in taking on this scope?  
I can close this PR, if so.  

Cheers.

Gorgious56:
Noted, I'll integrate the changes in next batch, should ocme by end of the week. Would suggest keeping PR open in case functionality is not up to par. Cheers

Gorgious56:
Came up with something like that. Wall regeneration is a bit buggy atm, I think we need a regeneration manager or something. Will investigate later.

https://github.com/user-attachments/assets/7010f721-9ffc-4a31-97f5-3e268fbe495a

PR #8171, opened 2026-06-13
Add per-drawing render overrides (status_render module)

theoryshaw:
Adds a new "Render Overrides" panel under the active drawing that applies render effects to IFC elements selected by a filter. Each drawing holds a list of rules; each rule has its own filter (reusing the shared Search filter system) plus exposure, gamma, and transparency.

Effects use the mechanism that can actually show them:
- Exposure/gamma: compositor nodes masked per rule via Cryptomatte, built just before a render and torn down after. Cryptomatte (not the Object Index pass) is used because EEVEE Next always renders Object Index as 0 (Blender #121690); it works in both EEVEE and Cycles.
- Transparency: a temporary Transparent-BSDF material applied to the matched objects so the geometry behind shows through. Because it is a real material it also shows live in the Rendered viewport (WYSIWYG), and is restored after the render / on toggle off / before save.

The single enable toggle drives both render and live viewport preview. Switching the active drawing (msgbus on scene.camera) re-syncs the preview and auto-enables a drawing that already has rules. Editing a rule's filter query (including clearing it) updates the preview immediately.

Rules are persisted to the IFC model at EPset_Drawing.RenderOverrides (JSON, declared in the pset template as IfcText), so they travel with the drawing: written on add/remove and on save, read back on file load and IFC import (via data.refresh / refresh_ui_data, since load_post fires before the import creates the cameras).

The override only runs where the compositor does (F12 and Default-render drawing underlays); the panel disables the toggle when the applied CurrentShadingStyle is not a Default render type.

Files:
- bim/module/status_render/: new module (operator, prop, ui, data, NOTES)
- bim/__init__.py: register the module
- bim/data/pset/EPset_Drawing.ifc: add RenderOverrides property template
- tool/search.py: status_render filter resolver, on_filter_query_edited dispatch, graceful empty-query handling
- bim/module/search/operator.py: notify on_filter_query_edited after edits

### DIFF
--- a/src/bonsai/bonsai/bim/__init__.py
+++ b/src/bonsai/bonsai/bim/__init__.py
@@ -91,6 +91,7 @@ modules = {
     "light": None,
     "alignment": None,
     "clip_box": None,
+    "status_render": None,
     # Uncomment this line to enable loading of the demo module. Happy hacking!
     # The name "demo" must correlate to a folder name in `bim/module/`.
     # "demo": None,
--- a/src/bonsai/bonsai/bim/data/pset/EPset_Drawing.ifc
+++ b/src/bonsai/bonsai/bim/data/pset/EPset_Drawing.ifc
@@ -5,7 +5,7 @@ FILE_NAME('EPset_Drawing.ifc','2020-01-01T00:00:00',$,$,'EPset_Drawing','EPset_D
 FILE_SCHEMA(('IFC4'));
 ENDSEC;
 DATA;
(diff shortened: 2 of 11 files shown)


theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/efe3d23bb6888e92b65742a71ba8bc22653091d5/logs/summaries/2026-06-17-rebase-drawing_render_overrides.md

theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/2b8cc7a6b654020621b8cd16a102c803c9755b82/logs/summaries/2026-07-05-PR-8171.md

PR #8200, opened 2026-06-24
Preserve custom opening geometry via a type-level Reference template

theoryshaw:
Custom IfcOpeningElement voids (e.g. an IfcPolygonalFaceSet / tessellation) were lost - reset to a default extrusion - on bim.duplicate_type, project append, and type switching, because the void lived only on occurrences and nothing carried it to a new type.

Anchor the shared opening body on the filling type as a 'Reference' representation map (per IfcShapeRepresentation, 'Reference' is geometry "not part of the Body representation", used for opening geometries excluded from an implicit Boolean operation). bim.duplicate_type and append copy a type's RepresentationMaps, so the template survives; generate_opening_from_filling consults it before falling back to a generated extrusion.

- map_type_representations: skip 'Reference' maps so occurrences don't receive the opening shape as their own Body geometry.
- opening.py: get_/set_type_opening_representation, promote_opening_to_type, update_type_template_from_opening; pre/post type.assign_type listeners (anchor the old type's void before a switch; regenerate to the assigned type's void afterwards, replacing the previous "preserve custom" guard).
- DuplicateType promotes the void before copy; AppendLibraryElement harvests the template cross-file from a library instance.
- Write-back on void edit, hooked at both commit paths (UpdateRepresentation and OverrideModeSetObject).
- reimport_element_representations renders the requested representation, so switching a type to its Reference row shows the void rather than the body.
- Representations panel shows RepresentationIdentifier plus column headers.

### DIFF
--- a/src/bonsai/bonsai/bim/module/geometry/__init__.py
+++ b/src/bonsai/bonsai/bim/module/geometry/__init__.py
@@ -69,6 +69,7 @@ classes = (
     operator.RemoveRepresentation,
     operator.RemoveRepresentationItem,
     operator.RemoveRepresentationItemFromShapeAspect,
+    operator.SelectByRepresentationType,
     operator.SelectConnection,
     operator.SelectRepresentationItem,
     operator.SwitchRepresentation,
--- a/src/bonsai/bonsai/bim/module/geometry/data.py
+++ b/src/bonsai/bonsai/bim/module/geometry/data.py
@@ -138,6 +138,11 @@ class RepresentationsData:
                 "ContextType": representation.ContextOfItems.ContextType or "",
                 "ContextIdentifier": "",
                 "TargetView": "",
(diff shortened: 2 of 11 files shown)


theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/8d445b6c607f0d7d4256caf0723dbcaefafa29dd/logs/summaries/2026-06-24-PR-8200.md

theoryshaw:
conflict resolution https://github.com/falken10vdl/bonsaiPR/blob/bfb7d3303d32a7553b2673efa0c32b3e01ec03de/logs/summaries/2026-07-05-rebase-opening-template-on-type.md

theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/8bee21c42224b5377f06dff2c4584b9688bd8fc7/logs/summaries/2026-07-06-PR-8200.md

PR #8206, opened 2026-06-25
Fix stale object reference crash in measure tool snapping

theoryshaw:
Drop deleted Blender objects from the snap cache before comparing names. Accessing an attribute on a removed object raised ReferenceError ("StructRNA of type Object has been removed"), crashing detect_snapping_points after an object was deleted.


theoryshaw:
I think this one can be closed as well, relateve to https://github.com/IfcOpenShell/IfcOpenShell/pull/9337

@brunoperdigao can you confirm?

Issue #8199, opened 2026-06-24
Extract elements recipe messes up georeferencing


BIMvoice:
### Bug Description

Georeferencing is removed when exctracted any elements using IfcPatch with this recipe.

The eatings and northings are converted into Blender offset.

Check the images, first one, original model, second one extracted element model.

### Attachments

<img width="1440" height="803" alt="Image" src="https://github.com/user-attachments/assets/bd902b49-99f1-459c-8164-132fc1c36041" />
<img width="1108" height="913" alt="Image" src="https://github.com/user-attachments/assets/b0d42114-205b-4652-9c77-577e62f24db0" />

### Debug and Error Output

```true

```

aothms:
Same as #8204 could you encode this in a failing test. Thanks

aothms:
Thanks @BIMvoice the test passes now - but you mentioned something else:

> The eatings and northings are converted into Blender offset.

I wonder if this was a by product of append_asset() from a library with georef to a model without - and should hence be fixed automatically - or whether this was caused by something else. Maybe double check and/or refine the tests.

BIMvoice:
Double-checked as requested. `append_asset()` is not the culprit: it has no georeferencing logic at all and copies placements verbatim, so it never converts eastings/northings into anything. The "converted into Blender offset" I reported was purely downstream of the georef loss: once `ExtractElements` dropped `IfcMapConversion`, Bonsai's import fell back to automatic false-origin guessing (`false_origin_mode == "AUTOMATIC"` → `tool.Loader.guess_false_origin`), which turned the large coordinates into a session Blender offset. With your recipe fix the map conversion survives, so that symptom is fixed automatically. No change needed in `append_asset`.

I refined the test per your suggestion in #8279: the georef regression test now also asserts the extracted element's local placement coordinates are byte-identical to the source, so a future regression that bakes map coordinates into placements would be caught. That PR also carries a one-line follow-up fix: e6dc582 iterates `IfcProject.RepresentationContexts` unconditionally, but the attribute is OPTIONAL, so projects without contexts currently crash every extraction (8 of the existing ExtractElements tests fail on `v0.8.0` right now with `TypeError: 'NoneType' object is not iterable`).

BIMvoice:
[...]

PR #8223, opened 2026-06-30
fix(ifcpatch): guard None RepresentationContexts; add placement roundtrip test (#8199)

BIMvoice:
Follow-up to #8199 and the fix in e6dc582, addressing Thomas's comment.

### 1. Bug fix — `TypeError` on projects with no contexts

`e6dc582` iterates `element.RepresentationContexts` directly. In IFC this attribute is optional and will be `None` for any `IfcProject` that has no representation contexts (e.g. the minimal project in `test_basic`). This crashes all tests that do not explicitly create a context:

```
TypeError: 'NoneType' object is not iterable
```

Fix: `for ctx in (element.RepresentationContexts or []):`

### 2. New test — element local placement is not baked with the georef offset

Answers Thomas's question:

> I wonder if this was a byproduct of `append_asset()` from a library with georef to a model without — and should hence be fixed automatically — or whether this was caused by something else.

**Answer: yes, it was automatic.** `append_asset()` does `auto_local2global(library) → auto_global2local(target)`. Before the fix, the target had no `IfcMapConversion`, so `auto_global2local` was a no-op and global coordinates landed as local coordinates (the "Blender offset"). After the fix the target has georef, so the roundtrip is correct.

`test_georeferencing_offset_not_baked_into_local_placement` verifies this: a wall at local (5, 10, 0) in a model with Eastings=100000, Northings=200000 must still be at (5, 10, 0) in the output, not (100005, 200010, 0).

All 12 tests pass, 2 correctly skipped for IFC2X3.

### DIFF
--- a/src/ifcpatch/ifcpatch/recipes/ExtractElements.py
+++ b/src/ifcpatch/ifcpatch/recipes/ExtractElements.py
@@ -110,7 +110,7 @@ class Patcher(ifcpatch.BasePatcher):
             pass
         if element.is_a("IfcProject"):
             proj = self.new.add(element)
-            for ctx in element.RepresentationContexts:
+            for ctx in (element.RepresentationContexts or []):
                 for coop in ctx.HasCoordinateOperation:
                     self.new.add(coop)
             return proj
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing as a stale duplicate: the None-`RepresentationContexts` guard and the placement roundtrip test here already landed via the merged #8279 (and #8220). #8199 is verified fixed and closed.

PR #8229, opened 2026-07-01
Make LAYER2 parametric wall tools IFC-class agnostic

theoryshaw:
The wall gizmos, joins, extend/regenerate-to-underside and fillet-corner machinery are driven entirely by placement + axis + IfcMaterialLayerSetUsage, never the entity type. The IfcWall restriction lived only in gate predicates and a wall-specific pset, so any AXIS2 LAYER2 element (e.g. vertical IfcCovering siding/cladding) was excluded despite working end-to-end.

- Drop the is_a("IfcWall") gate from is_wall and is_path_connectable_wall; gate on LAYER2 usage (plus fillet corner) instead.
- Resolve the wall/underside gizmo partner pair via the predicate, not is_a.
- Move fillet-corner state (IsFilletCorner / FilletRadius) off the wall-only BBIM_Wall pset onto the class-agnostic EPset_Parametric, via a shared get_parametric_prop helper that falls back to BBIM_Wall for existing files.

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/wall.py
+++ b/src/bonsai/bonsai/bim/module/model/wall.py
@@ -3158,7 +3158,12 @@ def _apply_fillet_corner_geometry(
 
 
 def _resolve_two_walls(context: bpy.types.Context) -> tuple[bpy.types.Object, bpy.types.Object] | None:
-    """``(active, other)`` from a 2-wall selection, both LAYER2 with straight axes."""
+    """``(active, other)`` from a 2-element selection, both LAYER2 with straight axes.
+
+    IFC-class agnostic: the fillet only needs two straight LAYER2 bodies (the
+    corner it creates inherits the active element's type), so coverings / siding
+    fillet the same way walls do. The ``has_layer2_usage`` check below is the
+    real gate, not entity type."""
     selected = list(tool.Blender.get_selected_objects())
     if len(selected) != 2:
         return None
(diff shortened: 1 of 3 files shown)


Gorgious56:
Interesting. As a pure coding perspective I would also rename functions to be coherent with their internal changes. eg regenerate_fillet_corner_wall > regenerate_fillet_entity (or something like that), is_wall > is_layer2_entity, is_path_connectable_wall > is_path_connectable_entity - otherwise we'll have a discrepancy and risk running into bugs. Cheers

I rarely ever user LAYER2's apart from walls so won't be able to crash test this. Do you mind sharing here a stripped down ifc file with a few test elements I can use ?

Not sure about this, but I'm wondering if we risk a namespace collision by naming the Pset "EPset_Parametric", maybe use "EPset_BBIM_Parametric" or something like that ? Again, I am not sure about that, do we risk getting conflicting custom psets from other vendors ?

Cheers

theoryshaw:
Cool. 

Here's a troubleshooting file...
https://hub.openingdesign.com/OSArch/Community_Troubleshooting/src/branch/main/_Agnostic_LAYER2_tools/Agnostic_LAYER2_tools.ifc

It actually has a couple glitches worth exploring at some stage...

image

Feel free to push changes, with other testing conditions--you have push privileges. 

...

> As a pure coding perspective I would also rename functions to be coherent with their internal changes. eg regenerate_fillet_corner_wall > regenerate_fillet_entity (or something like that), is_wall > is_layer2_entity, is_path_connectable_wall > is_path_connectable_entity - otherwise we'll have a discrepancy and risk running into bugs. Cheers

Agreed.  Do you want me to start tackling this?  Or did you want to stab it?

...

@Moult are you okay putting these fillet properties in EPSet_Parametric?
image

theoryshaw:
Logging this convo, here, as well: https://community.osarch.org/discussion/comment/29548/#Comment_29548

PR #8231, opened 2026-07-01
Updated documentation on Introduction to IFC 

Tyzze86:
The documentation on introduction to IFCstill have IFC4x3 as a draft. This is updated to being a published version from 2024. Updated tip on using IFC4 to have a second paragraph that recommend using IFC 4x3 for infractructure

### DIFF
--- a/src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
+++ b/src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
@@ -67,7 +67,7 @@ Begin learning IFC
 ------------------
 
 IFC has three versions published by ISO: **IFC2X3** from 2007, **IFC4** from
-2017, and **IFC4X3** in draft form. Each version improves on the previous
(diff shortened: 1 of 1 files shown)


Gorgious56:
Hello ! Thanks for the PR. If you're open to suggestions before merge:

The blank line after `elements.` has three trailing spaces that are worth trimming.

There is a contradiction between existing tip and infrastructure related tip, could be worth rewording everything, e.g.:

   > For most buildings, IFC4 is recommended. For infrastructure projects (road, railway, bridge, and other civil elements), use IFC4X3. The IFC4X3 documentation is also generally more newcomer-friendly.

Otherwise LGTM

Cheers

Tyzze86:
@Gorgious56 Thanks for the feedback. Great catching does trailing backspaces. I had missed those competely. 
Changed the tip to your suggested phrasing. I agree with that and makes the tip both shorter and more readable

PR #8242, opened 2026-07-03
Linked file features

theoryshaw:
dev notes: https://github.com/IfcOpenShell/IfcOpenShell/blob/f6cec0c9a5fb59453bf7cd6b44a2323fdaf81a8a/docs/dev-notes/linked-file-features.md

### DIFF
--- a/docs/dev-notes/linked-file-features.md
+++ b/docs/dev-notes/linked-file-features.md
@@ -1,0 +1,204 @@
+<!-- This file was generated with the assistance of an AI coding tool. -->
+
+# Linked file features — queries, styles, transforms, and multi-linking for linked IFC models
+
+> **Living dev note** for the `Linked_File_Features` branch/PR. Read before working
+> on the feature; append decisions and findings as the PR is refined. This is *not* user
+> documentation — at merge it is removed or its durable parts promoted to code comments.
+> See [README.md](README.md) for the convention (introduced on the
+> `opening-template-on-type` branch; not yet on this branch's base).
+
+## Problem
+
+Linked IFC models (`bim.link_ifc`) had several gaps that made them hard to use as a
+"reference in other trades' models" workflow:
+
+- One shared `.ifc.cache.blend` per IFC file meant the **same file could not be linked
+  twice with different selector queries** — both links showed whichever query was cached
+  first in-session, and whichever was cached last after reopening (Blender reuses one
+  library datablock per path).
+- The selector query was not durably stored anywhere in the host IFC, so save → reopen
+  lost or cross-wired the filter; a scripted `bpy.ops.bim.reload_link()` also wiped it.
+- Linked geometry got **flat diffuse-only materials** — external `.blend` styles
+  (`IfcExternallyDefinedSurfaceStyle`) and per-layer materials (layerset slicing) that
+  the normal import applies were ignored.
+- Moving a linked model required an explicit enable-edit → move → save dance on the
+  active link only, with save/cancel buttons in the panel header.
+- The Explore tool's highlight broke (GPU type errors), drew at the link's *original*
+  location when the link had been moved, and `bim.append_inspected_linked_element`
+  placed appended elements at the original location too.
+
+## Key facts established
+
+- **Cache architecture**: `LoadLink.link_ifc` generates a Python script and runs a
+  background Blender subprocess that executes `bim.load_linked_project` and saves a
+  `.ifc.cache.blend`. The host session then *links* (not appends) the `IfcProject/...`
+  collection from that blend and instances it via an empty (the link "handle").
+  Georeferencing metadata lives in a sidecar `.cache.json`; extracted properties in
+  `.cache.sqlite` (whole file, query-independent — deliberately shared across queries).
+- **Blender reuses an in-session library per path.** Loading the same blend path twice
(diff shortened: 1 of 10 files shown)


falken10vdl:
@theoryshaw some comments form my side:
- It looks that path handling is not fully consistent across the pr. Could absolute vs relative variants fail to match and end up creating duplicate linked document records?
- Is IfcDocumentReference.Description intended to be the long-term storage for selector query metadata? This may mix implementation metadata with IFC document semantics and potentially conflict with user made descriptions.
- Is md5(query)[:8] a sufficiently collision-resistant suffix for very long-lived IFC projects, or should a longer hash segment be allocated? Do not know how frequent is this in the industry...
- Would it not be better to persist link transformations when locking (DisableEditingLink) rather than on depsgraph autosave events, since those callbacks can run very frequently even if they return quickly?

Cheers!

theoryshaw:
Thanks for the review @falken10vdl!

Admittedly the following is an AI response, as this is pushing outside my understanding.  Feel free to push back if appropriate.   And also feel free to modify this PR as well.  If all this looks good, please pull, if it makes sense.  Thank you kindly for the review. 

...

**Path handling** — good catch, this was a real gap: `get_linked_models_documents()` keyed documents by the stored `Location`, so linking the same file first relative then absolute (or vice versa) created a duplicate document record. Fixed by resolving both sides to absolute paths before matching (⟨commit⟩).

**`Description` for the query** — agreed it's implementation metadata in an IFC attribute, but it follows the existing convention on these same references: `Identification` has always stored the link's 4×4 transformation, which is arguably a bigger stretch. These references are Bonsai-managed (`Scope="LINKED_MODEL"` documents created by Link IFC), so collisions with user-authored descriptions should be rare. If we want a cleaner long-term home, I'd suggest a follow-up that consolidates query + transform + options into one serialized convention rather than growing this PR.

**`md5(query)[:8]`** — 32 bits means a birthday collision needs on the order of ~65k _distinct queries against the same file_; real projects use a handful. More importantly a collision isn't silent: the cache JSON stores the full query string and `should_clear_cache()` compares it against the requested query, so a colliding cache is simply detected and rebuilt — it self-heals.

**Autosave vs save-on-lock** — save-on-lock alone loses the "what you see is what's saved" guarantee: moving a link and saving the project without locking would silently drop the move on next load. The handler is cheaper than it looks: it early-outs when there are no links, and only does real work on ticks that contain an object-transform update while a link is unlocked. It also gives undo consistency for free — undoing a move fires another depsgraph update, so the stored transform follows the undo, which a lock-time-only save wouldn't.

falken10vdl:
> Thanks for the review @falken10vdl!
> 
> Admittedly the following is an AI response, as this is pushing outside my understanding. Feel free to push back if appropriate. And also feel free to modify this PR as well. If all this looks good, please pull, if it makes sense. Thank you kindly for the review.
> 
> ...
> 
> **Path handling** — good catch, this was a real gap: `get_linked_models_documents()` keyed documents by the stored `Location`, so linking the same file first relative then absolute (or vice versa) created a duplicate document record. Fixed by resolving both sides to absolute paths before matching (⟨commit⟩).

Cool!


> **`Description` for the query** — agreed it's implementation metadata in an IFC attribute, but it follows the existing convention on these same references: `Identification` has always stored the link's 4×4 transformation, which is arguably a bigger stretch. These references are Bonsai-managed (`Scope="LINKED_MODEL"` documents created by Link IFC), so collisions with user-authored descriptions should be rare. If we want a cleaner long-term home, I'd suggest a follow-up that consolidates query + transform + options into one serialized convention rather than growing this PR.

OK


> **`md5(query)[:8]`** — 32 bits means a birthday collision needs on the order of ~65k _distinct queries against the same file_; real projects use a handful. More importantly a collision isn't silent: the cache JSON stores the full query string and `should_clear_cache()` compares it against the requested query, so a colliding cache is simply detected and rebuilt — it self-heals.

OK. Did not go to the details. Good it is complete.


> **Autosave vs save-on-lock** — save-on-lock alone loses the "what you see is what's saved" guarantee: moving a link and saving the project without locking would silently drop the move on next load. The handler is cheaper than it looks: it early-outs when there are no links, and only does real work on ticks that contain an object-transform update while a link is unlocked. It also gives undo consistency for free — undoing a move fires another depsgraph update, so the stored transform follows the undo, which a lock-time-only save wouldn't.

Another option is to drop depsgraph autosave, and then Save Project first call save_link_transformation for all loaded editing links so save semantics remain “what you see is what gets saved”, right? @Moult what do you think? I remember that when you reviewed the PR I created for linked files I think I had an initial depsgraph which was called a  lot and I remember that the lock was something you added so it was when locking that the update was taken in consideration.
Cheers!

theoryshaw:
conflict resolution: https://github.com/falken10vdl/bonsaiPR/blob/30195867a94659a869496bb71941fd2362c0f311/logs/summaries/2026-07-09-rebase-Linked_File_Features.md

theoryshaw:
@falken10vdl 

>  IfcDocumentReference.Description intended to be the long-term storage for selector query metadata? This may mix implementation metadata with IFC document semantics and potentially conflict with user made descriptions.

perhaps....
https://github.com/IfcOpenShell/IfcOpenShell/pull/8242/changes/cd5897d10d954cdedbf9842f091816f5305d1a81

PR #8243, opened 2026-07-03
Feature #5753 - Autosave for ifc files

### LINKED ISSUE #5753: Request for an ifc autosave feature
Relative new comers are told not to try and maintain both blender and ifc files. However, a couple of times now I've been caught up in the work - you know that zone where time disappears - and boom, a segfault burns two hours of work. This has happened to me a couple of times in the last few days.

It would be really nice to have an optional autosave with a timer to make a backup ifc file that can be checked for at load time. If the backup is newer than the save file it can ask if you want to load that instead.

sboddy:
Implemented as described in https://github.com/IfcOpenShell/IfcOpenShell/issues/5753, with two options:
- A nag dialog with save or cancel options.
- An autosaved file.

Settings are in preference to activate the feature (default: off), the period before prompting/saving,
and choosing between the two methods.

I have reviewed and tested the code, and I understand it, and it works and does not introduce any obvious bugs.

This commit was created using AI assistance. Cursor for the initial code, then Grok and I fixing all the errors
that Cursor made.

### DIFF
--- a/src/bonsai/bonsai/bim/handler.py
+++ b/src/bonsai/bonsai/bim/handler.py
@@ -320,9 +320,11 @@ def loadIfcStore(scene: bpy.types.Scene) -> None:
     IfcStore.purge()
     refresh_ui_data()
     if not tool.Ifc.get():
+        tool.Autosave.cancel_timer()
         return
     tool.Ifc.schema()
     IfcStore.relink_all_objects()
+    tool.Autosave.reset_timer()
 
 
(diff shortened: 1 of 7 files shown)


sboddy:
Maybe let this stew in BonsaiPR for a little bit, or until someone else has had a chance to glance over it.

I should add that I've only tested this with straight IFC files. I have _not_ tested with .blend files, or the metadata .blend files.

Issue #8100, opened 2026-05-24
Search Group True-False to 1-0 bug


ppaawweeuu:
### Bug Description

I use search group with very basic setup. I'm looking for LoadBearing walls.
True and False values I can chose from list do not work. No objects are selected.
When I manually write 1 (as True) or 0 (as False) for search value - everything works fine.
Applies to current and NEW filter mode.
BTW: I love search menu with available options :D

### Attachments

`True` finds nothing
<img width="875" height="343" alt="Image" src="https://github.com/user-attachments/assets/6aef1f57-e5d2-429b-afb2-f7fdbb190a24" />

`1` finds 3 walls
<img width="880" height="330" alt="Image" src="https://github.com/user-attachments/assets/9b8389ec-2481-4f5e-b7a2-f09e4cdf9e7b" />

### Debug and Error Output

```true
os: Windows
os_version: 10.0.26200
python_version: 3.13.9
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: Intel64 Family 6 Model 140 Stepping 1, GenuineIntel
blender_version: 5.1.2
bonsai_version: 0.8.5
bonsai_commit_hash: None
bonsai_commit_date: None
bonsai_git_branch: None
last_actions: 
last_error: None
blend_file_path: Not saved
blend_file_dirty: N/A
ifc_file_path: C:/Users/.../wall_attributes.ifc
ifc_is_dirty: False
ifc: wall_attributes.ifc
schema: IFC4
preprocessor_version: IfcOpenShell 0.8.5
originating_system: Bonsai 0.8.5
```

BIMvoice:
Root-caused. The Search UI writes the list choice as `True`/`False` into the filter query, but the `filter_elements` grammar only accepted uppercase `TRUE`/`FALSE` as boolean keywords — anything else fell through to a string comparison against a Python bool, which never matches. That is also why manually typing `1`/`0` works: numbers take the int-cast path.

This is fixed in #8248 (boolean keywords are now case-insensitive: `true`/`True`/`TRUE`), verified with exactly your scenario — `Pset_WallCommon.LoadBearing=True` matches after the fix. Once that merges, the list values work as-is.

BIMvoice:
Confirmed on `v0.8.0` and fixed in #8288.

[...]

Issue #6116, opened 2025-02-08
Bugs with ExtractElements and other things


BIMvoice:
A list of bugs we discovered in a meeting, @Moult has more details:

Bug 1: Incorrect Deletion of Opening Elements When Filtering Load-Bearing Elements

Description:

When filtering the IFC model to isolate load-bearing elements, the deletion operation also removes opening elements (e.g., windows, doors) that should not be deleted.

Steps to Reproduce:
	1.	Apply a filter to select only load-bearing elements (walls, slabs, beams, etc.).
	2.	Use the deletion function to remove non-load-bearing elements.
	3.	Observe that openings (which should remain) are also deleted.

Expected Behavior:

The filter should preserve openings, as they are linked to load-bearing elements (walls, slabs, etc.), instead of removing them.

Actual Behavior:

Openings are deleted along with non-load-bearing elements.

Potential Cause:

The selection and filtering mechanism might not be distinguishing between standalone elements and dependent elements (such as openings).

Screenshots / Video Evidence:

Recording with timestamp

Bug 2: Exported Model Shows Incorrect Geometry / Misaligned Elements

Description:

When extracting elements using the filtering function, the exported IFC model has misaligned geometry. Some elements appear shifted or incorrectly placed, even though they were properly positioned before extraction.

Steps to Reproduce:
	1.	Apply a filter to extract only load-bearing elements.
	2.	Use the “Extract Elements” function.
	3.	Open the resulting IFC model.
	4.	Observe that slabs, walls, or other elements appear in incorrect positions.

Expected Behavior:

The exported model should retain the original geometric positions of elements.

Actual Behavior:

Some elements (e.g., slabs) appear displaced after export.

Potential Cause:

The transformation matrix or georeferencing information may not be correctly preserved during the export process.

Screenshots / Video Evidence:

Recording with timestamp

Bug 3: Filter Mechanism Doesn’t Always Apply Correctly to Property Sets

Description:

When applying a filter based on property sets (e.g., Pset_Common), the filter does not always return the correct elements. The filtering mechanism seems to be case-sensitive or inconsistent in reading Boolean values (TRUE vs. true).

Steps to Reproduce:
	1.	Apply a filter using Pset_Common.LoadBearing = TRUE.
	2.	Observe that some expected load-bearing elements are not selected.
	3.	Change the filter to use true instead of TRUE and test again.
	4.	The behavior may vary inconsistently.

Expected Behavior:

[...]

PR #8248, opened 2026-07-05
Selector: accept case-insensitive TRUE/FALSE in filter queries (#6116)

BIMvoice:
Part of #6116 (Bug 3: boolean filters are case-sensitive).

The `filter_elements` grammar defined boolean keywords as uppercase only:

```
true: "TRUE"
false: "FALSE"
```

So a query like `Pset_WallCommon.LoadBearing=true` did not parse as a boolean. It fell through to a string comparison against the property's Python `bool`, which never matches, so the query silently returned nothing while `=TRUE` worked. That inconsistency is what the issue reports.

The format grammar in the same file already treats booleans case-insensitively (`"true" | "True" | "TRUE"`). This change makes the filter grammar consistent with it:

```
true: "true" | "True" | "TRUE"
false: "false" | "False" | "FALSE"
```

Extended `test_selecting_by_property` to cover all three cases for both `TRUE` and `FALSE`. Full `test_selector.py` suite passes (37 tests).

### DIFF
--- a/src/bonsai/bonsai/bim/module/search/operator.py
+++ b/src/bonsai/bonsai/bim/module/search/operator.py
@@ -128,6 +128,20 @@ class FilterValueSuggestions(Operator):
     suggestion_type: StringProperty(default="value")
 
     first_launch: BoolProperty(default=True, options={"SKIP_SAVE"})
+
+    @staticmethod
+    def value_to_query_string(value: Any) -> str:
+        """Stringify a raw IFC value for use as a filter query suggestion.
+
+        Booleans are rendered as the query language's TRUE / FALSE keywords
+        rather than Python's str(bool) capitalisation ("True" / "False"),
+        since the filter grammar only accepts the former as a literal and
+        would otherwise silently fail to match the property (#6116, #8100).
+        """
+        if isinstance(value, bool):
+            return "TRUE" if value else "FALSE"
+        return str(value)
+
     search_value: StringProperty(
         name="Search",
         description="Search for filter values",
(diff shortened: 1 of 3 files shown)


BIMvoice:
This also fixes #8100 (`LoadBearing=True` returning nothing from the search UI), which is the same boolean-filter case-sensitivity bug. I closed my other PR #8288 in favour of this one to avoid two overlapping changes.

Moult:
I would prefer to be explicit and make it exactly TRUE / FALSE. It is a query language after all. It seems to me as a bug in the Blender UI instead.

I don't understand why the format grammar is so lax?

BIMvoice:
Thanks for the pushback, that is a fair read and I agree the query language should stay strict.

I reworked the PR to match. The filter grammar is back to exactly TRUE / FALSE, nothing else, and I added assertions the other way so the strict behaviour is pinned by a test (lowercase or mixed case true/True/false/False is not a boolean literal and does not match).

You were right that the real bug is in the Search UI, not in the selector. It was in the value suggestion list: FilterValueSuggestions in bim/module/search/operator.py built its suggestions with Python's str(value), so a genuine boolean property came out as the strings "True"/"False" (Python capitalisation), not the query language's TRUE/FALSE. Picking "LoadBearing: True" from the dropdown therefore produced a query like Pset_WallCommon.LoadBearing = "True", a quoted string compared against a Python bool, which can never match.

The fix is a small type-aware helper that renders a genuine bool as TRUE/FALSE and leaves everything else as str(value). I deliberately kept it type-aware rather than normalising any string that happens to be spelled "true"/"false"/"null", so a property whose value is really the string "true" is still filtered as a string and not silently coerced into the boolean keyword. That keeps the strictness you want and avoids a new ambiguity.

Reproduced live in headless Blender with a wall carrying Pset_WallCommon.LoadBearing = True. Before: get_property_values returned {'True','False'} and the exported query matched nothing. After: get_property_values returns {'TRUE','FALSE'}, the exported query is Pset_WallCommon.LoadBearing = TRUE, and it returns exactly the load bearing wall. ifcopenshell.util.selector is untouched apart from reverting the grammar to strict.

On the format() grammar question: that lenient boolean rule comes from a different feature, the optional suppress_zero_inches flag on imperial_length() (commit 3f2cad049), where a user hand writes something like imperial_length({{...}}, 2, "'", "\"", True) inside a format template. That is user authored template text rather than a query filter, which is why the boolean there was made lenient. I do not think it needs changing, but if you would prefer it also be strict TRUE/FALSE for consistency I am happy to tighten it in a follow up.

This was prepared with the assistance of an AI tool.

Issue #4110, opened 2023-12-13
IfcDiff - Results not as expected


vulevukusej:
I made some changes to the sample file (wall coloured in yellow) below to test IfcDiff:

1. Added a dummy pset "as"
2. Changed IFCCLASS from IfcWallStandardCase to IfcController
3. Changed the building storey from "Erdgeschoss" to "Dachgeschoss"

I tested the diff both in BBIM (v230902) and in my own tool.  The results generated are identical, i show screenshots from my own tool just for clarity. 

The changes in properties seem to be correctly identified.  

If i leave the arguments empty, i get no results. 
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/83825269/d53d294a-c739-4bd8-8dbd-d347063ed96c)

But If change the changes to "Container" I get 82 changes ? 
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/83825269/02359916-54e2-4cb0-8269-ee96db287e93)

If I combine "Container" and "IfcController" I get 1 change:
![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/83825269/52364e72-c446-4e24-98d4-ae57111aae40)

[FzkHaus.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/13659695/FzkHaus.txt)
[FzkHaus_rev2.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/13659697/FzkHaus_rev2.txt)

sennhvi:
> In diff.py, diff_element_relationships(self, old, new) shows why too many changes were made.
> 
> ```
> for relationship in self.relationships:
>     if relationship == "type":
>         if ifcopenshell.util.element.get_type(old) != ifcopenshell.util.element.get_type(new):
>             self.change_register.setdefault(new.GlobalId, {}).update({"type_changed": True})
>             return True
> ```
> 
> Same IfcXXXType entity instance in 2 different ifc files refer to two objects with different identity in python , which makes != operator returns True, and self.change_register got a new member.
> 
> To compare two entities like IfcXXXType we may need to parse the entity object's property tree and compare each property , after what we can get the result.

And `GlobalId` may take an irreplaceable role to identify whether these 2 objects share same value?

vulevukusej:
Thanks @sennhvi - Are you willing to create a PR?

falken10vdl:
@vulevukusej was PR created for this?
Thanks!

[...]

PR #8249, opened 2026-07-05
IfcDiff: compare container and aggregate by GlobalId (#4110)

BIMvoice:
Part of #4110.

## Problem

The `container` and `aggregate` relationship checks compared the results of `get_container`/`get_aggregate` between the two files directly:

```python
if ifcopenshell.util.element.get_container(old) != ifcopenshell.util.element.get_container(new):
```

Those are entity instances belonging to two different `ifcopenshell.file` objects, so `!=` is always true even when the container is unchanged. Every element with a container or aggregate was reported as changed. On the model in #4110 this produced 82 false "changes" for `container`.

The `type` check in the same method already does the right thing (compares `GlobalId`). This makes `container` and `aggregate` consistent with it.

## Verify

On the #4110 sample models, `relationships=["container"]` goes from 82 changes to 1, and that 1 is a real change (an element whose container GlobalId actually differs). Reproduced before/after.

Added `test_unchanged_container_is_not_reported` and `test_changed_container_is_reported`. Full `test.py` suite passes (6 tests).

## Not included here

The same models also show that a pure class change (`IfcWallStandardCase` -> `IfcController`, same GlobalId, same attribute values) is not detected, because `diff_element` never compares `is_a()`. That is a separate concern with an output-shape question (new key vs fold into `attributes_changed`); happy to send a follow-up PR if you want it addressed.

### DIFF
--- a/src/ifcdiff/ifcdiff.py
+++ b/src/ifcdiff/ifcdiff.py
@@ -323,11 +323,25 @@ class IfcDiff:
                     self.change_register.setdefault(new.GlobalId, {}).update({"properties_changed": diff})
                     return True
             elif relationship == "container":
-                if ifcopenshell.util.element.get_container(old) != ifcopenshell.util.element.get_container(new):
+                old_container = ifcopenshell.util.element.get_container(old)
+                new_container = ifcopenshell.util.element.get_container(new)
+                if old_container is not None and new_container is not None:
+                    if old_container.GlobalId != new_container.GlobalId:
+                        self.change_register.setdefault(new.GlobalId, {}).update({"container_changed": True})
+                        return True
+                elif old_container != new_container:
+                    # one of the containers is None while the other is not None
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing this in favour of #9296, which is a strict superset.

This PR fixed the container and aggregate comparison, which compared `entity_instance` objects across two different `ifcopenshell.file` instances and so were never equal. That fix is included verbatim in #9296, on the same lines.

While working in the same function I found six further defects of the same kind, all causing a revision diff to silently under-report: class changes never detected, placement never compared, material never compared, the `id` exclusion regex never matching, attribute detail computed then discarded, `is_shallow=False` never accumulating, and geometry never queued for models without a `MODEL_VIEW` subcontext. Keeping both open would mean two PRs conflicting on the same lines for one bug, so folding is cleaner than competing.

Nothing is lost by closing this. If the larger PR is unwelcome for any reason, say so and I will reopen this one on its own.

PR #8256, opened 2026-07-05
Sequence: sub-day durations occupy their day when cascading (#8245)

### LINKED ISSUE #8245: wrong durations in schedule
### Bug Description

Hi,

I have made a construction schedule animation in Bonsai. I figured most of it out but I have an error in the duration calculations of the tasks I do not understand. Maybe a bug.
As long as i put durations of one or more days (1d), the successors starting date is fine. When I then give durations of 5 hours (5h) to some tasks, the date is not calculated anymore. I have a work calendar with worktime from 07:00-12:00 and 13:00-17:00 (9 hours/day). In my case I have three tasks 5 hours each. So the date should switch but doesn't.
Am I doing it wrong or is it a bug???

thanks.

Image

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Fixes #8245.

## Problem

With a task duration of `PT5H`, cascading produced `ScheduleFinish == ScheduleStart` — the five hours vanished — and the FINISH_START successor was scheduled from that wrong finish (same day as the predecessor's start). Reproduced on current `v0.8.0`:

```
T1: 2026-07-06T09:00 -> 2026-07-06T09:00  dur PT5H     # finish == start
T2: 2026-07-06T09:00 -> ...                            # successor starts same day
```

Two places assume day-only durations:
- `ifcopenshell.util.sequence.get_start_or_finish_date` returns `start` unchanged whenever `duration.days == 0` (intended for milestones, but `PT5H` has `days == 0, seconds == 18000`).
- `cascade_schedule`'s FINISH_START branch computes `days = 0 if predecessor_duration.days == 0 else 1`, so a sub-day predecessor was treated as a milestone.

## Fix

The engine is deliberately day-granular (times snap to 09:00/17:00), so the minimal consistent behaviour is that a sub-day duration occupies its start day:

- `get_start_or_finish_date`: only a true zero duration (no days, no seconds) is a milestone; a sub-day duration counts as one day.
- FINISH_START cascade: only true zero durations let the successor start the same day.

After the fix:

```
T1: 2026-07-06T09:00 -> 2026-07-06T17:00  dur PT5H
T2: 2026-07-07T09:00 -> 2026-07-07T17:00
```

The stored `ScheduleDuration` keeps its full precision (`PT5H`); milestone behaviour (`P0D`) is unchanged.

## Verify

Added `test_cascading_finish_to_start_for_sub_day_durations`, mirroring the existing FS test style. Red-green verified: fails on the previous code (`finish == 09:00`), passes with the fix. Full `test/api/sequence/` suite passes (67 tests).

Hour-granular scheduling (finishing at 14:00 after 5h) would be a larger redesign of the day-walking calendar logic; this PR makes sub-day durations behave consistently within the existing day-granular model.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/sequence/cascade_schedule.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/sequence/cascade_schedule.py
@@ -156,7 +156,10 @@ class Usecase:
                 finish = self.get_task_time_attribute(predecessor, "ScheduleFinish")
                 if not finish:
                     continue
-                days = 0 if predecessor_duration.days == 0 else 1
+                # A sub-day duration (e.g. PT5H) still occupies its day, so
+                # only true zero durations (milestones) start the same day.
+                is_milestone = predecessor_duration.days == 0 and not getattr(predecessor_duration, "seconds", 0)
+                days = 0 if is_milestone else 1
                 duration_type = "WORKTIME"
                 if rel.TimeLag:
                     # updated to handle IfcRatioMeasure as a TimeLag value
--- a/src/ifcopenshell-python/ifcopenshell/util/sequence.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/sequence.py
@@ -115,13 +115,18 @@ def get_start_or_finish_date(
     calendar: ifcopenshell.entity_instance,
     date_type: Literal["START", "FINISH"] = "FINISH",
 ):
-    if not duration.days:
+    seconds = int(getattr(duration, "seconds", 0))
+    if not duration.days and not seconds:
         # Typically a milestone will have zero duration, so the start == finish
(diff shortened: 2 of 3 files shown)


BIMvoice:
Closing this in favour of #8313.

@SautaRoc told us directly on #8245 that this approach was not the solution: rounding a sub-day duration up to occupy its whole day still loses the real timing, and his example (a bridge push carried out in a single night) needs actual intraday resolution.

#8313 does that. A `PT5H` task starting 09:00 now finishes at 14:00, and `PT10H30M` finishes at 19:30, with a regression test. It also makes the extra cascade_schedule change in this PR unnecessary, since a correctly computed finish already cascades to the right successor start.

#8313 remains open for review.

Issue #8187, opened 2026-06-19
Polyline insertion near heavy linked IFC models: ray_cast_by_proximity_2d crash on None verts + severe slowdown from snapping against linked geometry


valentenet:
### Bug Description

**Summary**
Inserting an IFC element via the polyline tool (Shift+A) in a scene containing a heavy linked IFC model has two distinct problems, both in the snap raycast path. Both have a confirmed fix below (tested on my project — crash gone, insertion went from ~30s to instant).

Problem 1 — Crash: TypeError: cannot unpack non-iterable NoneType

In Raycast.ray_cast_by_proximity_2d (bonsai/tool/raycast.py), edge vertices are projected to 2D via view3d_utils.location_3d_to_region_2d, which returns None for vertices behind the camera / outside the projection. The block that builds edge_verts only handles the case where exactly one vertex is None (the ^ XOR at ~line 396). When both vertices project to None — an entire edge behind the camera, common when the viewport is close to or inside dense linked geometry — it stores (None, None), and the following loop does p0x, p0y = p0 (line 424), unpacking None and raising the TypeError. The crash fires on every handle_mouse_move, making placement impossible.

Fix (guard the loop):
###################################
python        

for e in edges:
            p0, p1 = edge_verts[e]
            if p0 is None or p1 is None:
                continue
            p0x, p0y = p0
            p1x, p1y = p1
###################################
These edges aren't on screen, so no usable snap is lost. Also covers the case where intersect_edge_region_border returns None.

**Problem 2 — Severe slowdown: snap evaluates heavy linked models**

Snap.detect_snapping_points calls Raycast.filter_objects_to_raycast, which adds every bounding-box-hit object to the raycast list, including heavy linked reference models. On a federated project this makes each insertion take ~30s before the modal even responds. Excluding the linked collection from the View Layer avoids it, but also hides the model.

Suggested fix — let filter_objects_to_raycast skip linked reference objects, ideally behind a user-facing toggle:
###################################
python    

@classmethod
    def _is_linked_reference(cls, obj):
        if getattr(obj, "library", None) is not None:
            return True
        if getattr(obj, "override_library", None) is not None:
            return True
        inst = getattr(obj, "instance_collection", None)
        if inst is not None and getattr(inst, "library", None) is not None:
            return True
        return False
###################################

and, inside the loop in filter_objects_to_raycast, before processing each object:
###################################
python        

for obj, bbox_2d in objs_2d_bbox:
[...]

PR #8261, opened 2026-07-05
Bonsai: skip fully off-screen edges in 2D proximity snap (#8187)

BIMvoice:
Implements the crash fix (Problem 1) from #8187, diagnosed and field-tested by @valentenet — all credit for the analysis to them.

## Problem

In `Raycast.ray_cast_by_proximity_2d`, edge endpoints are projected with `location_3d_to_region_2d`, which returns `None` behind the camera. The builder only handles the XOR case (exactly one endpoint off-screen); when **both** endpoints are off-screen — an entire edge behind the camera, common when the viewport sits inside dense linked geometry — `(None, None)` lands in `edge_verts` and the proximity loop crashes on `p0x, p0y = p0` with `TypeError: cannot unpack non-iterable NoneType`, on every `handle_mouse_move`.

## Fix

Guard the loop and `continue` on a `None` endpoint. Such edges are not on screen, so no usable snap is lost; this also covers `intersect_edge_region_border` returning `None` in the XOR branch.

## Verify

Logic guard in Blender-bound code; the reporter confirmed exactly this change eliminates the crash on their federated project (~1.42M verts). Syntax-checked; no headless test is possible for the modal path.

## Not included

Problem 2 from the issue (snap evaluating heavy *linked* models, ~30s stalls) is a performance/design change with a user-facing toggle proposal — left for a follow-up so this crash guard can land independently.


carlopav:
With recent commits in v0.8.0 by @brunoperdigao this PR seems outdated and should probably be closed? Can you check @BIMvoice ?

theoryshaw:
> With recent commits in v0.8.0 by @brunoperdigao this PR seems outdated and should probably be closed? Can you check @BIMvoice ?

namely this one i think: https://github.com/IfcOpenShell/IfcOpenShell/pull/9337

BIMvoice:
@carlopav you are right, checked and closing.

The fix guarded against `None` endpoints inside a `selected_edges` loop in `tool/raycast.py`. That code no longer exists: `selected_edges` now appears **zero** times in `raycast.py` on `v0.8.0`, and `git merge-tree` against the current branch conflicts rather than applying.

@brunoperdigao's snap rewrite between 16 and 19 August is what removed it, in particular `9673f01c21` ("remove custom edge BVH") and `209d487f5f` ("use GPU detection for solid and wireframe objects"). The projection path this PR was defending no longer works the way it did, so the guard has nothing left to guard.

Thanks for flagging it rather than leaving it sitting in the queue. If #8187 turns out to still reproduce on the new snap code, it would need a fresh diagnosis against the GPU detection path rather than a rebase of this.

PR #8264, opened 2026-07-05
Selector: fix material != and NULL filtering (#6787)

### LINKED ISSUE #6787: Search Selector filter for material issues
### Bug Description

* when I filter elements with material it doesn't seem to work with 'not equal' (!=) operator
* if I use 'NULL' it returns all elements (useful to check if elements don't have material assigned, if working)

https://github.com/user-attachments/assets/5782bebf-27d2-4ce0-b241-d6d68ca708a8

thanks

Blender 4.4.0 Bonsai v0.8.3 250606

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Fixes #6787.

## Problem

With walls carrying CON01 / STE01 / no material, on current `v0.8.0`:

```
material=CON01   -> [w1]            OK
material!=CON01  -> [w3]            wrong: w2 (STE01) excluded
material=NULL    -> [w1, w2, w3]    wrong: returns everything
material!=NULL   -> []              wrong: returns nothing
```

Two defects in the `material` facet:

1. **Double negation**: `compare()` already applies `!` per material, and the facet then flips the aggregate again (`return result if comparison == "=" else not result`). Any element whose materials differ from the value passes the inner compare and is then flipped to excluded.
2. **NULL compared per-field**: `NULL` was compared against each material's `Name` and `Category`, so any material with no `Category` (`None is None`) matched `material=NULL` — i.e. nearly every element.

## Fix

Compare each material positively and negate the aggregate once (`!=` now means "no material's name or category matches", consistent with the equivalent fix for list values in #8254), and define `NULL` as "the element has no materials", which is what the existing test already asserted for the no-materials case.

## Verify

All four cases above now return the expected sets. Extended `test_selecting_by_material` with a second material and a material-less element; red-green verified (fails on the previous code, passes with the fix). Full `test_selector.py` passes (37 tests).

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/selector.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/selector.py
@@ -1017,15 +1017,21 @@ class FacetTransformer(lark.Transformer):
 
         def filter_function(element: ifcopenshell.entity_instance) -> bool:
             materials = ifcopenshell.util.element.get_materials(element)
-            result = False if materials else None
-            for material in materials:
-                if self.compare(material.Name, comparison, value):
-                    result = True
-                if self.compare(getattr(material, "Category", None), comparison, value):
-                    result = True
-            if result is not None:
-                return result if comparison == "=" else not result
-            return self.compare(None, comparison, value)
+            positive = comparison.lstrip("!")
+            if value is None:
+                # NULL means the element has no materials at all.
+                result = not materials
+            else:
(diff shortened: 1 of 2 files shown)


BIMvoice:
Rebased onto current v0.8.0. The earlier lint-formatting and compile-and-test failures were pre-existing base breakage from before the CI cleanup, not from this change; test_selector.py passes locally on the rebased branch (39 passed).

PR #8278, opened 2026-07-05
pset.edit_pset: keep booleans for LOGICAL measures (#7475)

### LINKED ISSUE #7475: Pset_BuildingStoreyCommon, parameter AboveGround not éditable and incorrect
### Bug Description

Hello,
When starting a new project or using an existing IFC file, the parameter AboveGround in the Pset_BuildingStoreyCommon does not propose the right values (TRUE, FALSE, and UNKNOWN).
If the IFC file is exported from Revit, it is impossible to change the value of the parameter; it is always "UNKNOWN".
Is there a way to fix this problem?

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Fixes #7475.

`cast_value_to_primary_measure_type` maps `LOGICAL` through `str`, so setting `Pset_BuildingStoreyCommon.AboveGround = True` stored the string `"True"`, which `IfcLogical` reads as **UNKNOWN** — reproduced headless; the property can never take a real value from any caller. Genuine bools now pass through unchanged; string inputs keep existing behaviour.

Regression test added; full `test_edit_pset.py` passes (29).

### DIFF
--- a/src/bonsai/bonsai/bim/helper.py
+++ b/src/bonsai/bonsai/bim/helper.py
@@ -247,12 +247,7 @@ def import_attribute(
         is_logical = str(attribute_type) == "<type IfcLogical: <logical>>"
         enum_value = data[new.name]
         if is_logical:
-            new.special_type = "LOGICAL"
-            enum_items = ("TRUE", "FALSE", "UNKNOWN")
-            new.enum_items = json.dumps(enum_items)
-            if enum_value is not None and enum_value != "UNKNOWN":
-                # IfcOpenShell returns bool if IfcLogical is True/False.
-                enum_value = "TRUE" if enum_value else "FALSE"
+            enum_value = set_logical_enum_items(new, enum_value)
         else:
             enum_items = ifcopenshell.util.attribute.get_enum_items(attribute)
             new.enum_items = json.dumps(enum_items)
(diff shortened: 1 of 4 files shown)


Moult:
This looks incomplete:

1) the test suite does not test all three logical values
2) the original bug is still unaddressed, the UI doesn't display logical options for the user to choose.

BIMvoice:
Thanks for the review, both points are fair. I have pushed a rework that addresses each.

1) Test coverage for all three LOGICAL values

test_edit_pset.py now has a dedicated test for each state of IfcLogical (TRUE, FALSE, UNKNOWN), all round tripped through edit_pset and read back from the IFC entity:

- test_editing_a_logical_property_with_true: AboveGround = True stores IfcLogical(.T.), wrappedValue is True.
- test_editing_a_logical_property_with_false: AboveGround = False stores IfcLogical(.F.), wrappedValue is False.
- test_editing_a_logical_property_with_none_is_unknown: AboveGround = None stores IfcLogical(.U.), wrappedValue == "UNKNOWN", both for a brand new property and for editing an existing one back to None.

While adding the None/UNKNOWN case I found the original fix did not actually cover it: edit_pset already treats None as "delete/blank this property" for every data type, so a LOGICAL property set to None was either removed or left with no NominalValue at all, never an explicit UNKNOWN. I extended edit_pset so that, specifically for a LOGICAL property, None means the real UNKNOWN state rather than delete/blank (documented in the docstring). I also found that passing Python None straight into the underlying IfcLogical constructor crashes the interpreter; the value the wrapper actually accepts for UNKNOWN is the string "UNKNOWN", so edit_pset now translates None to "UNKNOWN" internally before it reaches that boundary.

2) The UI not exposing the logical options

You are right that this was still broken. IfcOpenShell's own ifcopenshell.util.attribute.get_primitive_type already classifies IfcLogical as an "enum" (TRUE/FALSE/UNKNOWN), and Bonsai already has a working general pattern for exactly this (bonsai/bim/prop.py's Attribute class plus bonsai/bim/helper.py's import_attributes, used for ordinary entity attributes). The pset editing panel, however, never wired a property's LOGICAL type into that pattern, in two places in bonsai/tool/pset.py:

- get_special_type_for_prop never recognised IfcLogical, so LOGICAL properties never got tagged as special_type "LOGICAL".
- import_single_value_from_template (used for a property proposed from a pset template that has not been added yet) only had branches for string/integer/float/boolean, nothing for "enum".

Concretely, before the fix, enabling editing on a fresh AboveGround property produced: data_type "enum", special_type "" (empty), enum_items "" (empty), and get_value() returned None. In other words, the widget rendered as an enum dropdown with zero options, so there was nothing for the user to pick, exactly matching what you reported.

I added a small shared helper, bonsai.bim.helper.set_logical_enum_items, that configures an Attribute as a LOGICAL enum with items TRUE/FALSE/UNKNOWN (the same translation import_attributes already used), and wired it into both import_pset_from_existing (editing an already-set property) and import_single_value_from_template (a freshly proposed template property, which now defaults to UNKNOWN until the user picks a value). get_special_type_for_prop now also recognises IfcLogical on both the template and the existing-property paths.

After the fix, the same property shows data_type "enum", special_type "LOGICAL", enum_items ["TRUE", "FALSE", "UNKNOWN"], and the user can pick any of the three from the dropdown. I drove this end to end in headless Blender through the real bim.enable_pset_editing / bim.edit_pset operators (not just unit tests): picking TRUE, FALSE, and UNKNOWN from the property panel each round tripped correctly to IfcLogical(.T.), IfcLogical(.F.), and IfcLogical(.U.) on the actual IFC entity.

Let me know if you would rather see the third state exposed a different way (e.g. keeping a checkbox for TRUE/FALSE plus a separate "set unknown" toggle) instead of the three way enum dropdown; I went with the enum since it already matches Bonsai's existing convention for other LOGICAL attributes elsewhere in the addon.

This was prepared with the assistance of an AI tool.

Issue #6436, opened 2025-03-25
[IfcPatch]Create proper CLI entrypoint / script


sanzoghenzo:
**Feature Description**

Currently, ifcPatch CLI can only be run via `python -m ifcpatch`.
The current instructions to use an alias may work, but its doesn't play well with multiple environments, especially now that we have easy to use isolated environment managers such as pipx and uv.

Using the wonderful uv as an example, to run ifcpatch in an ephemeral environment, one needs to:

```shell
uvx --with ifcpatch python -m ifcpatch ...
```

instead of just `uvx ifcpatch ...` as one would expect.

```shell
$ uvx ifcpatch -h
The executable `ifcpatch` was not found.
warning: Package `ifcpatch` does not provide any executables.
```

this also makes it impossible to install ifcpatch with `uv tool install` which errors out with the `No executables are provided by ifcpatch` message.

Python already has all we need to create a proper cross platform executable via the [`project.scripts` table](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#creating-executable-scripts)

since this config needs a function, it cannot points to the `__main__` module directly, unless we wrap everything inside it in a `main` function 

```python
def main():
    # current contents of __main__.py

if __name__ == "__main__":
    main()
```

and then use it in the pyproject.toml

```toml
[project.scripts]
ifcpatch = "ifcpatch.__main__:main"
```

to be more explicit, the `main` function could be named `cli` and put into a `cli` module.

BIMvoice:
Implemented in #8281: `main()` wrapper plus `[project.scripts] ifcpatch = "ifcpatch.__main__:main"`, same convention as `ifcquery`/`ifcedit`/`ifcmcp`. Once merged and released, `uvx ifcpatch ...` and `pipx install ifcpatch` work as you'd expect; `python -m ifcpatch` unchanged.

sanzoghenzo:
I realized that most of the CLI utilities don't have the `project.scripts` entrypoints.
Should I create an issue (and PR when I find the time) for each of them, or a single one is enough? Can I just edit the title/message of this one?

BIMvoice:
@sanzoghenzo I measured the current state across the repo, in case it helps you scope it. Nineteen packages ship a `pyproject.toml`.

**Already have `[project.scripts]`:** `ifcedit`, `ifcmcp`, `ifcquery`, all using the `<pkg> = "<pkg>.__main__:main"` form.

[...]

PR #8281, opened 2026-07-05
ifcpatch: add console script entry point (#6436)

BIMvoice:
**Request** (#6436): `ifcpatch` can only be run as `python -m ifcpatch`, which doesn't play well with pipx/uvx (`uvx --with ifcpatch python -m ifcpatch ...` instead of `uvx ifcpatch ...`).

**Change:** wrap the module-level CLI code in `main()` (behavior unchanged) and register `ifcpatch = "ifcpatch.__main__:main"` under `[project.scripts]` — the exact convention already used by `ifcquery`, `ifcedit`, and `ifcmcp` in this repo.

**How verified:** `python -m ifcpatch` still parses args and runs recipes end-to-end (ExtractElements on a small model, output checked), and the entry-point callable was exercised the way the generated script does it (`from ifcpatch.__main__ import main; main()`) with a second recipe run.

### DIFF
--- a/src/ifcpatch/ifcpatch/__main__.py
+++ b/src/ifcpatch/ifcpatch/__main__.py
@@ -24,23 +24,28 @@ import ifcopenshell
 
 import ifcpatch
 
-parser = argparse.ArgumentParser(description="Patches IFC files to fix badly formatted data")
-parser.add_argument("-i", "--input", type=str, required=True, help="The IFC file to patch")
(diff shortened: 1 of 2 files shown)


sanzoghenzo:
It would be nice to also update the documentation at src/ifcopenshell-python/docs/ifcpatch.rst, removing the instruction for the alias and the packaging hint

Issue #6489, opened 2025-04-02
Issue filtering by pset prop value


Andrej730:
> Good evening,
> For over a week or so I have not been able to filter by the Property as below;
> 
> ![Image](https://github.com/user-attachments/assets/a4ab44a2-2ef8-4924-bc7e-8fe7619ba5fb)
> 
> I have a Pset_WallCommon with the 'IsExternal' property checked but when I try to filter as above it returns "0 results". I have been updating with the unstable releases but still not working. Thank you. 

 _Originally posted by @Owuraqu in [#5959](https://github.com/IfcOpenShell/IfcOpenShell/issues/5959#issuecomment-2738667247)_

steverugi:
Hi @Andrej730 it works when using all caps TRUE

BIMvoice:
Root cause found and fixed in pending #8248: the `filter_elements` grammar only defined the boolean keywords in uppercase (`TRUE`/`FALSE`), so `IsExternal=TRUE` parsed as a boolean but `IsExternal=true` fell through to a string comparison against a Python bool — which never matches, silently returning 0 results. That's exactly why @steverugi's all-caps workaround works. After #8248 the keywords are case-insensitive, with regression tests. Safe to close once that merges.

PR #8288, opened 2026-07-05
Selector: match boolean properties by True/False spelling (#8100)

BIMvoice:
## Problem

Filtering by a boolean property (e.g. `LoadBearing`) with the search UI or a query only worked when the value was written as the special token `TRUE`/`FALSE` or as `1`/`0`. Typing `True` or `False` matched nothing, which is what the Bonsai search dropdown offers by default. Reported in #8100.

## Cause

In `util.selector.FacetTransformer.compare`, a boolean property value is a Python `bool`, and `bool` subclasses `int`. A string query value therefore entered the `isinstance(element_value, int)` branch and ran `int("True")`, which raises and is swallowed, so the comparison fell through to no match. `int("1")` succeeds and `True == 1`, which is why only the numeric spellings worked.

## Fix

Handle `bool` before the numeric branches and coerce the common string spellings (`True`/`true`/`TRUE`/`1` and `False`/`false`/`FALSE`/`0`) to the matching boolean, then compare by equality. The special `TRUE`/`FALSE` tokens and `1`/`0` still work exactly as before.

## Verification

Red-green regression test added in `test/util/test_selector.py::test_selecting_by_boolean_property`. Before the fix, `True`/`true`/`False`/`false` return the empty set; after, all four spellings resolve to the correct element. Numeric, string, and NULL comparisons were re-checked for regressions and pass unchanged.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/selector.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/selector.py
@@ -1225,12 +1225,22 @@ class FacetTransformer(lark.Transformer):
             return any(self.compare(ev, comparison, value) for ev in element_value)
         elif isinstance(value, str):
             try:
-                if isinstance(element_value, int):
+                if isinstance(element_value, bool):
+                    # bool subclasses int, so handle it before the numeric
+                    # branches. A boolean property should match every common
+                    # spelling ("True"/"False") the same way "1"/"0" already do.
+                    if value in ("True", "true", "TRUE", "1"):
+                        value = True
+                    elif value in ("False", "false", "FALSE", "0"):
+                        value = False
+                elif isinstance(element_value, int):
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing as redundant. This overlaps with #8248, which fixes the same root cause (the `filter_elements` grammar defined the boolean keywords as uppercase-only) at the grammar layer and already covers this exact case: with #8248, `LoadBearing=True`/`=true`/`=TRUE` all parse as booleans. #8248 is the more idiomatic fix (it makes the filter grammar consistent with the format grammar in the same file) and keeps `compare()` untouched, so I'd rather land that one than carry two overlapping changes. #8100 is the same boolean-filter bug as #6489 and #6116 (Bug 3), all resolved by #8248.

Issue #8129, opened 2026-05-30
Incorrect filter behavior for multi-selected enumerated values


ppaawweeuu:
### Bug Description

Filters for multi-selected enumerated property values behave counterintuitively for operators such as 'equal to', 'not equal to', 'contains', and 'does not contain'. Logic that works correctly when comparing strings seems inappropriate when comparing enumerated types.

My IfcProject
I have only one element in the project (IfcWall) with one property set and two enumerated values selected:
Wall01: Pset_ConcreteElementGeneral.CastingMethod = **['INSITU', 'PRINTED']**

**equal to**
Fileter: `IfcWall, "Pset_ConcreteElementGeneral"."CastingMethod" = "INSITU"`
Finds Wall01 – I think the result is misleading. It should be empty because the property contains that value, but is not equal to a single value.

**not equal to**
Filter: `IfcWall, "Pset_ConcreteElementGeneral"."CastingMethod" != "INSITU"`
Finds Wall01 – I think this is OK, but the result is oddly the same for both '=' and '!='.

**contains**
Filter: `IfcWall, "Pset_ConcreteElementGeneral"."CastingMethod" *= "INSITU"`
Finds Wall01 – I think this is correct. The list of values contains INSITU.

**does not contain**
Filter: `IfcWall, "Pset_ConcreteElementGeneral"."CastingMethod" !*= "INSITU"`
Finds Wall01 – I think this is completely misleading.

-
I understand that enums—especially multi-selected ones—are an edge case, but filters support other modules across all schemas and with the implementation of IfcQuery, IfcEdit, IfcMCP, IfcChat and prompted operations, there could be many unexpected outcomes.
I used ifc4x3 file, but problem applies to all schemas and all enumerated properties in psets.

I issued similar problem #8100 with processing bool values, I think it is related, to some degree.

### Debug and Error Output

```true
os: Windows
os_version: 10.0.26200
python_version: 3.13.9
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: Intel64 Family 6 Model 140 Stepping 1, GenuineIntel
blender_version: 5.1.2
bonsai_version: 0.8.5
bonsai_commit_hash: None
bonsai_commit_date: None
bonsai_git_branch: None
last_actions: (...)
last_error: None
blend_file_path: Not saved
blend_file_dirty: N/A
ifc_file_path: C:/Users/.../wall_props_filter2.ifc
ifc_is_dirty: True
[...]

PR #8289, opened 2026-07-05
Selector: negated operators on list properties mean none-match (#8129)

BIMvoice:
## Problem

For a list-valued property such as a multi-selected enumeration (`CastingMethod = ['INSITU', 'PRINTED']`), the negated filter operators behaved counterintuitively (#8129):

- `!= "INSITU"` returned the element even though it contains `INSITU`
- `!*= "INSITU"` (does not contain) returned the element even though it does contain `INSITU`
- `=` and `!=` gave the same result

## Cause

`util.selector.FacetTransformer.compare` distributes the comparison over each list value with `any(...)`. For a negated operator that pushes the negation inside the `any`, so the element matched whenever it held a single value that differed. `['INSITU','PRINTED'] != 'INSITU'` became `any(v != 'INSITU')`, which is true because `PRINTED` differs.

## Fix

Handle a negated operator as `not any(positive match)`, i.e. "none of the values match", which is the intuitive reading of `!=` and `!*=` on a list. Positive operators (`=`, `*=`) are unchanged.

Resulting semantics for a list property:
- `=`   element has at least one value equal
- `!=`  element has no value equal
- `*=`  element has at least one value containing the substring
- `!*=` element has no value containing the substring

## Verification

Red-green regression test added in `test/util/test_selector.py::test_selecting_by_multivalue_property`. Scalar comparisons are unaffected (single values never enter the list branch).

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/selector.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/selector.py
@@ -1222,6 +1222,14 @@ class FacetTransformer(lark.Transformer):
 
     def compare(self, element_value, comparison, value) -> bool:
         if isinstance(element_value, (list, tuple)):
+            if comparison.startswith("!"):
+                # For a negated operator the intuitive meaning is "none of the
+                # values match", i.e. NOT (any value matches the positive form).
+                # Distributing the negation into `any` instead would return the
+                # list whenever it holds a single differing value, so "!=" and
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing as superseded: #8254 (the aggregate-negation fix for #8129) is now merged into v0.8.0, which covers this. This PR was a parallel approach to the same fix and now conflicts, so it is redundant. Thanks.

PR #8292, opened 2026-07-05
ifctester: entity requirement as xs:restriction no longer crashes on fail (#8046)

BIMvoice:
## Problem

When an element matches the applicability but fails an entity requirement expressed as `<xs:restriction><xs:enumeration>`, ifctester crashed instead of reporting a failure (#8046):

```
```

The `<xs:restriction>` form for entity names is valid per the IDS 1.0 schema and is covered by existing pass-path tests, but the fail path was untested and broken.

## Cause

In `Entity.__call__`, `self.name` can be a `Restriction` rather than a string. On the IFC2X3 fail path the type-inference branch called `self.name.endswith("TYPE")` and built `f"{self.name}TYPE"`, both of which assume a string. (The reporter also noted a missing `return True` in `Restriction.__eq__`; that is already present on `v0.8.0`, so only the string-method crash remained.)

## Fix

Rewrite the IFC2X3 branch so it never calls string methods on `self.name`. It derives the occurrence name from the element's type (`IfcWallType` -> `IFCWALL`) and compares it with `==`, which works for both plain strings and `Restriction`s. This keeps the existing string-name behaviour (direct match and type inference) and additionally makes restriction-based entity requirements resolve correctly.

## Verification

Red-green test added in `test_facet.py::TestEntity::test_ifc2x3_restriction_entity_requirement`, covering both the previously crashing fail path and the pass-via-type-mapping path. Verified against `v0.8.0`: before, an IFC2X3 element failing a restriction entity requirement raised `AttributeError`; after, it reports a clean fail, and string-name direct/inference and IFC4 restriction paths are unchanged.

### DIFF
--- a/src/ifctester/ifctester/facet.py
+++ b/src/ifctester/ifctester/facet.py
@@ -231,16 +231,21 @@ class Entity(Facet):
         is_pass = inst.is_a().upper() == self.name
         reason = None
 
-        if (
-            not is_pass
-            and inst.file.schema == "IFC2X3"
-            and not self.name.endswith("TYPE")
-            and (element_type := ifcopenshell.util.element.get_type(inst))
-        ):
-            is_pass = element_type.is_a().upper() == f"{self.name}TYPE"
-            reason = {"type": "NAME", "actual": element_type.is_a().upper()[:-4]}
-        elif not is_pass:
-            reason = {"type": "NAME", "actual": inst.is_a().upper()}
+        if not is_pass:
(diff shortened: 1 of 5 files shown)


Moult:

Thanks Petru. Same diagnosis and the same shape of fix, deriving the occurrence class from the type and comparing with `==`. #9540 lands that without the extra fixture files, adds the matching fallback to `filter()` so applicability and requirement agree, and handles `IfcDoorStyle`. Closing this in favour of #9540.

PR #8294, opened 2026-07-05
ifcdiff: compare container and aggregate by GlobalId, not identity (#4110)

BIMvoice:
## Problem

Running IfcDiff with the `container` relationship reported far too many changes (the reporter saw 82 spurious changes for a file where only a handful of things changed). #4110.

## Cause

`diff_element_relationships` compared the container and aggregate of the old and new element directly:

```python
if ifcopenshell.util.element.get_container(old) != ifcopenshell.util.element.get_container(new):
```

Those two instances come from two different `ifcopenshell.file` objects. `entity_instance.__eq__` treats a proper instance's identity as its step id within its file, so instances from different files never compare equal. Every element that had a container (or an aggregate) was therefore flagged as changed, even when it had not moved. The `type` branch had already been fixed to sidestep exactly this by comparing GlobalIds; `container` and `aggregate` were left behind.

## Fix

Compare the container and aggregate by GlobalId (with `None` handling for the added/removed case), mirroring the `type` branch.

## Verification

Red-green regression tests added in `test.py`:
- `test_unchanged_container_is_not_a_change` and `test_unchanged_aggregate_is_not_a_change`: a round-tripped copy with the same containment now yields no changes (before the fix both failed, reporting a spurious `*_changed`).
- `test_changed_container`: genuinely moving a wall to another storey is still reported as `container_changed`.

Full `ifcdiff` suite passes (7 tests). Confirmed the two new "unchanged" tests fail on `v0.8.0` without the fix and pass with it.

### DIFF
--- a/src/ifcdiff/ifcdiff.py
+++ b/src/ifcdiff/ifcdiff.py
@@ -323,11 +323,21 @@ class IfcDiff:
                     self.change_register.setdefault(new.GlobalId, {}).update({"properties_changed": diff})
                     return True
             elif relationship == "container":
-                if ifcopenshell.util.element.get_container(old) != ifcopenshell.util.element.get_container(new):
+                # The two containers come from different files, so their instances
+                # never compare equal. Compare by GlobalId, as the type branch does.
+                old_container = ifcopenshell.util.element.get_container(old)
+                new_container = ifcopenshell.util.element.get_container(new)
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing as a duplicate of #8249, which fixes the same cross-file GlobalId comparison with equivalent logic and is already mergeable.

Issue #4112, opened 2023-12-13
Error when classifying elements using bSDD


BIMvoice:
This is the error I get:

`Python: Traceback (most recent call last):
  File "/Users/petru/Library/Application Support/Blender/4.0/scripts/addons/blenderbim/tool/ifc.py", line 201, in execute
    IfcStore.execute_ifc_operator(self, context)
  File "/Users/petru/Library/Application Support/Blender/4.0/scripts/addons/blenderbim/bim/ifc.py", line 336, in execute_ifc_operator
    result = getattr(operator, "_execute")(context)
  File "/Users/petru/Library/Application Support/Blender/4.0/scripts/addons/blenderbim/bim/module/classification/operator.py", line 337, in _execute
    ifcopenshell.api.run("pset.edit_pset", tool.Ifc.get(), pset=pset, properties=properties)
  File "/Users/petru/Library/Application Support/Blender/4.0/scripts/addons/blenderbim/libs/site/packages/ifcopenshell/api/__init__.py", line 66, in run
    result = usecase_class(ifc_file, **settings).execute()
  File "/Users/petru/Library/Application Support/Blender/4.0/scripts/addons/blenderbim/libs/site/packages/ifcopenshell/api/pset/edit_pset.py", line 162, in execute
    new_props = self.add_new_properties()
  File "/Users/petru/Library/Application Support/Blender/4.0/scripts/addons/blenderbim/libs/site/packages/ifcopenshell/api/pset/edit_pset.py", line 346, in add_new_properties
    value = self.cast_value_to_primary_measure_type(value, primary_measure_type)
  File "/Users/petru/Library/Application Support/Blender/4.0/scripts/addons/blenderbim/libs/site/packages/ifcopenshell/api/pset/edit_pset.py", line 410, in cast_value_to_primary_measure_type
    return type_fn(value)
ValueError: could not convert string to float: ''
`

![Screenshot 2023-12-13 at 15 39 11](https://github.com/IfcOpenShell/IfcOpenShell/assets/139251053/c23429a5-77cf-42db-9829-b468aa1e5e0d)

BIMvoice:
Re-checked against the current `v0.8.0`. The original 2023 traceback came through the old `blenderbim` bSDD classification operator, which has since been rewritten, so that exact path is gone. But the underlying library behaviour still reproduces:

```python
import ifcopenshell, ifcopenshell.api.pset, ifcopenshell.api.root
[...]

PR #8296, opened 2026-07-05
pset.edit_pset: empty string for a numeric property stores no value (#4112)

BIMvoice:
## Problem

Classifying elements via bSDD crashed when a numeric property was left empty (#4112):

```
File ".../ifcopenshell/api/pset/edit_pset.py", line 410, in cast_value_to_primary_measure_type
    return type_fn(value)
ValueError: could not convert string to float: ''
```

## Cause

When a property is covered by a pset template, `get_primary_measure_type` returns the template's numeric `PrimaryMeasureType`. If the supplied value is an empty string (as bSDD produces for an unfilled numeric field), `cast_value_to_primary_measure_type` runs `float("")` and raises.

## Fix

Catch the `ValueError` at the call site and, for an empty-string value, store the property with no nominal value (`IfcPropertySingleValue` with `$`) instead of crashing. The change is surgical:
- an empty string on a **numeric or boolean** measure now yields a property with no value,
- an empty string on a **string** measure still stores the empty string,
- a genuinely invalid value (e.g. `"abc"` for a real) still raises.

## Verification

Red-green test added in `test_edit_pset.py::test_an_empty_string_for_a_numeric_property_stores_no_value` (runs for both IFC2X3 and IFC4): `ThermalTransmittance=""` no longer raises and yields `NominalValue is None`, while `Reference=""` keeps an empty `IfcIdentifier`. Confirmed the crash reproduces on `v0.8.0` before the fix and is resolved after, with the valid-value and invalid-value paths unchanged.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/pset/edit_pset.py
@@ -386,8 +386,18 @@ class Usecase:
                 if value is None:
                     nominal_value = value
                 else:
-                    value = self.cast_value_to_primary_measure_type(value, primary_measure_type)
-                    nominal_value = self.file.create_entity(primary_measure_type, value)
+                    try:
+                        cast_value = self.cast_value_to_primary_measure_type(value, primary_measure_type)
+                        nominal_value = self.file.create_entity(primary_measure_type, cast_value)
+                    except ValueError:
+                        if value == "":
+                            # An empty string is not a meaningful value for a
+                            # numeric or boolean measure (e.g. an unfilled bSDD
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing this as a duplicate of #8251, which we opened for the same bug earlier the same day and then lost track of. Apologies for the noise.

#8251 is the more complete fix. It handles the empty-numeric-value case in both `update_existing_prop_single_value` and `add_new_properties`, and factors the check into a reusable `is_empty_numeric_value` predicate. This PR only covers `add_new_properties`, so editing an already existing property with an empty string would still raise.

#8251 remains open for review.

Issue #8207, opened 2026-06-25
Measure tool crashes with ReferenceError: StructRNA of type Object has been removed


theoryshaw:
### Bug Description

Describe the bug

Activating the measure tool (S M hotkey) throws a ReferenceError and aborts when the snap cache holds a reference to a Blender object that has since been deleted. The snapping system caches SnapObj entries that each keep a direct reference to a Blender object; once that object is deleted, the next snap pass accesses snap_obj.obj.name on the freed object and crashes.

To Reproduce

Open a model and use the measure tool / polyline snapping so that objects get cached for snapping.
Delete one of the objects that was previously hovered/snapped to.
Invoke the measure tool again (S M) and move the mouse over the viewport.
Crash.
Expected behavior

Deleted objects are silently dropped from the snap cache and snapping continues without error.

Console output

File ".../bonsai/tool/raycast.py", line 973, in create_snap_obj
    if obj.name == snap_obj.obj.name:
                   ^^^^^^^^^^^^^^^^^
ReferenceError: StructRNA of type Object has been removed
(Full traceback originates from workspace.py → measure_tool → polyline.invoke → Snap.detect_snapping_points → Raycast.filter_objects_to_raycast → Raycast.create_snap_obj.)

Root cause

Raycast.create_snap_obj iterates cls.snap_objs and compares snap_obj.obj.name without validating that the cached object still exists. A deleted object's StructRNA is invalidated, so any attribute access raises ReferenceError.

### Debug and Error Output

```true
Python: Traceback (most recent call last):
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\project\workspace.py", line 109, in execute
    getattr(self, f"hotkey_{self.hotkey}")()
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\project\workspace.py", line 135, in hotkey_S_M
    bpy.ops.bim.measure_tool("INVOKE_DEFAULT", measure_type=measure_type)
  File "C:\Program Files\Blender Foundation\Blender 4.5\4.5\scripts\modules\bpy\ops.py", line 107, in __call__
    ret = _op_call(self.idname_py(), kw, C_exec, C_undo)
[...]

PR #8301, opened 2026-07-06
Bonsai: drop deleted objects from the snap cache (#8207)

BIMvoice:
## Problem

Activating the measure tool (S M) after deleting an object that was previously snapped to crashes with `ReferenceError: StructRNA of type Object has been removed` (#8207). Thanks to @theoryshaw for the precise diagnosis.

## Cause

Snapping caches `SnapObj` entries in `tool.Raycast.snap_objs`, each holding a direct reference to a Blender object. `create_snap_obj` iterates that cache and reads `snap_obj.obj.name`; once the cached object has been deleted, that access hits a freed `StructRNA` and raises `ReferenceError`, aborting the snap pass (and the same stale reference would break the raycast passes downstream).

## Fix

At the start of `create_snap_obj`, prune cache entries whose Blender object no longer exists (detected by the `ReferenceError` on attribute access). Deleted objects are silently dropped and snapping continues, which is the expected behaviour. Because `filter_objects_to_raycast` only appends live `create_snap_obj` results to the raycast list, pruning here also protects the downstream passes.

## Verification

I don't have a Blender session to run the modal tool, but I verified the pruning logic in isolation: a cache with a deleted entry (which raises `ReferenceError` on `.name`, exactly as a freed `StructRNA` does) is reduced to only the live entries, and subsequent `.obj.name` access no longer raises.

### DIFF
--- a/src/bonsai/bonsai/tool/raycast.py
+++ b/src/bonsai/bonsai/tool/raycast.py
@@ -969,6 +969,18 @@ class Raycast(bonsai.core.tool.Raycast):
     def create_snap_obj(cls, obj):
         if obj.data is None or not isinstance(obj.data, bpy.types.Mesh):
             return None
+        # Drop cache entries whose Blender object has been deleted. Accessing the
+        # freed object below (and in the raycast passes) otherwise raises
+        # ReferenceError ("StructRNA ... has been removed") and aborts snapping.
+        # See #8207.
+        valid_snap_objs = []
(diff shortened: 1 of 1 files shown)


BIMvoice:
Confirmed, closing this in favor of #8206: same fix (the valid_snap_objs prune for ReferenceError cache entries), and yours was first. Sorry for the duplicate @theoryshaw.

Issue #8227, opened 2026-07-01
Error when printing


theoryshaw:
### Bug Description

I get the following error when trying to print `SECTION - KITCHEN - NS - NORTH` in the following drawing. 

[Highland_Haven - Copy.ifc.txt](https://github.com/user-attachments/files/29531450/Highland_Haven.-.Copy.ifc.txt)

### Debug and Error Output

```true
File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\operator.py", line 385, in execute
    annotation_svg = self.generate_annotation(context)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\operator.py", line 2294, in generate_annotation
    self.svg_writer.create_blank_svg(svg_path).draw_annotations(annotations, precision, decimal_places).save()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\svgwriter.py", line 375, in draw_annotations
    self.draw_dimension_annotations(obj)
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\svgwriter.py", line 1571, in draw_dimension_annotations
    self.draw_dimension_annotation(
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\svgwriter.py", line 1617, in draw_dimension_annotation
    v0 = self.project_point_onto_camera(v0_global)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #8228, opened 2026-07-01
Debug drawing faster failure search

theoryshaw:

PR #8302, opened 2026-07-06
Bonsai: don't crash drawing generation on an unprojectable dimension (#8227)

BIMvoice:
## Problem

Printing a drawing crashed (#8227) with:

```
File ".../drawing/svgwriter.py", in project_point_onto_camera
    return self.camera.matrix_world.inverted() @ geometry.intersect_line_plane(...)
TypeError: Matrix multiplication: not supported between 'Matrix' and 'NoneType' types
```

## Cause

`mathutils.geometry.intersect_line_plane` returns `None` when the line is parallel to the plane. `project_point_onto_camera` fed that straight into `Matrix @ ...`, so a single dimension annotation whose endpoint is parallel to the camera plane took down the entire drawing generation.

## Fix

Return `None` from `project_point_onto_camera` when the intersection is `None`, and in `draw_dimension_annotation` skip that one dimension when either endpoint fails to project. The rest of the drawing still generates instead of the whole print aborting.

## Verification

I don't have a Blender session to run the full print, but I verified the guard logic in isolation: with `intersect_line_plane` returning `None`, the projection returns `None` and the dimension is skipped with no `TypeError`, and drawing continues. @theoryshaw a run on the `SECTION - KITCHEN - NS - NORTH` drawing would confirm end to end.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/svgwriter.py
+++ b/src/bonsai/bonsai/bim/module/drawing/svgwriter.py
@@ -1564,6 +1564,11 @@ class SvgWriter:
         offset = Vector([self.raw_width, self.raw_height]) / 2
         v0 = self.project_point_onto_camera(v0_global)
         v1 = self.project_point_onto_camera(v1_global)
+        if v0 is None or v1 is None:
+            # A dimension endpoint is parallel to the camera plane and cannot be
+            # projected. Skip this one annotation rather than aborting the whole
+            # drawing generation. See #8227.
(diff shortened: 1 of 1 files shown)


BIMvoice:
Closing in favour of #8228 per the discussion in #8227. Skipping the unprojectable dimension hides corrupt geometry instead of surfacing it, which is the wrong tradeoff for catching how the geometry became degenerate. Deferring to the debug-and-fail approach.

Issue #8122, opened 2026-05-29
Can't `bim.remove_drawing` a corrupt drawing


theoryshaw:
activate `Type A2` drawing. 
And then try to `bim.remove_drawing` it. 

get the following error.

[Highland_Haven - Copy.ifc.txt](https://github.com/user-attachments/files/28402374/Highland_Haven.-.Copy.ifc.txt)

```
os: Windows
os_version: 10.0.19045
python_version: 3.11.11
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: Intel64 Family 6 Model 158 Stepping 10, GenuineIntel
blender_version: 4.5.7 LTS
bonsai_version: 0.0.0
bonsai_commit_hash: 7489d08
bonsai_commit_date: 2026-05-29T09:49:27-05:00
bonsai_git_branch: BonsaiPR_base
last_actions: 
# operator: BIM_OT_remove_drawing
last_error: Traceback (most recent call last):
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\ifc.py", line 531, in execute_ifc_operator
    result = getattr(operator, "_execute")(context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\operator.py", line 3287, in _execute
    sheet_references = tool.Drawing.get_sheet_references(drawing)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\tool\drawing.py", line 3180, in get_sheet_references
    if reference.Location == drawing_reference.Location:
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'Location'

blend_file_path: D:\Dropbox\Gitea_OD\Highland_Haven\Open\Models_and_CAD\BlenderBIM\temp.blend
blend_file_dirty: True
[...]

PR #8303, opened 2026-07-06
Bonsai: allow removing a drawing with no associated document (#8122)

BIMvoice:
## Problem

`bim.remove_drawing` on a corrupt drawing crashed (#8122):

```
File ".../bonsai/tool/drawing.py", in get_sheet_references
    if reference.Location == drawing_reference.Location:
AttributeError: 'NoneType' object has no attribute 'Location'
```

## Cause

`get_sheet_references` calls `get_drawing_document(drawing)`, which returns `None` when the drawing has no `IfcRelAssociatesDocument` (i.e. a corrupt drawing). The code then dereferenced `drawing_reference.Location`, so the drawing could not be removed at all.

## Fix

Return an empty list of sheet references when `get_drawing_document` returns `None` — a drawing with no document has no sheet references — so `remove_drawing` proceeds instead of aborting.

## Verification

Confirmed that `get_drawing_document` returns `None` for a drawing with no `IfcRelAssociatesDocument`, and the guard then returns an empty list rather than dereferencing `None.Location`. @theoryshaw a run of `bim.remove_drawing` on the `Type A2` drawing in your file would confirm end to end.

### DIFF
--- a/src/bonsai/bonsai/tool/drawing.py
+++ b/src/bonsai/bonsai/tool/drawing.py
@@ -2830,6 +2830,10 @@ class Drawing(bonsai.core.tool.Drawing):
     def get_sheet_references(cls, drawing: ifcopenshell.entity_instance) -> list[ifcopenshell.entity_instance]:
         sheet_references: list[ifcopenshell.entity_instance] = []
         drawing_reference = cls.get_drawing_document(drawing)
+        if drawing_reference is None:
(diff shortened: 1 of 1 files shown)


BIMvoice:
Closing in favour of #8123, which was submitted first and is the more complete fix, it also guards the downstream document removal in `remove_drawing` and handles the no-group case, not just `get_sheet_references`. Thanks @theoryshaw.

Issue #8205, opened 2026-06-25
Drawings & BCF from Georeferenced model


Yurialo:
## Description
I have encountered a critical bug regarding camera management, the Drawings tool, and the BCF tool, which seems to be strictly related to georeferenced models (large coordinates). 

There are two main issues that occur after saving and reopening the file:

1. **Loss of Zoom and Crop settings:** 
When setting up a camera view (e.g., in a plan view or drawing layout), customizing the zoom factor and the print crop area, these settings are completely lost upon saving and reopening the file. The camera automatically resets to its default viewport values, losing the focus and framing previously established. This also affects the BCF tool camera states.

2. **Camera Coordinate Offset (Shift):** 
In models with georeferencing/large coordinates, the camera responsible for generating planimetric sections suffers a severe coordinate offset (shift) after saving and reopening the file. The camera physically moves away from its original position. This exact same shifting issue occurs when saving and restoring viewpoints via the BCF tool.

## Steps to Reproduce
1. Open a georeferenced IFC model in Bonsai.
2. Create a drawing/section or set up a camera view using the BCF tool.
3. Adjust the camera zoom, framing, and position.
4. Save the project and close the file.
5. Reopen the file -> The camera zoom/crop is reset to default, and the camera position suffers an offset/shift.

## Expected Behavior
The camera should maintain its exact position, coordinates, zoom factor, and crop settings across saves, regardless of whether the model is georeferenced or uses large coordinates.

### Debug and Error Output

```true

```

BIMvoice:
Mechanism pointers for whoever picks this up. Both symptoms fit one storage split: drawing/BCF cameras live in Blender scene space, while georeferenced models get a session-computed model offset ("Calculate model offset" step on load). Anything stored in Blender world coordinates but not re-anchored to the IFC origin will shift when the offset is recomputed on reopen — exactly your symptom 2, and the same reason BCF viewpoints (which are absolute map-space in the file) come back displaced. Symptom 1 (zoom/crop reset) is the camera props not being serialized into the IFC/drawing definition at all, so they only ever lived in the .blend session.

Fix direction: persist camera zoom/crop into the drawing's stored definition, and store/restore camera positions relative to the IFC origin (applying `auto_local2global`/`auto_global2local` on save/load) rather than raw Blender world space. Needs someone with a GUI + a georeferenced model to implement and verify.

BIMvoice:
[...]

PR #8304, opened 2026-07-06
Bonsai: apply georeferencing offset to imported drawing cameras (#8205)

BIMvoice:
## Problem

On georeferenced / large-coordinate models, drawing and BCF section cameras are shifted away from their correct position after saving and reopening, and drift further on each save/reopen cycle (#8205, symptom 2).

## Cause

Every object imported from IFC has the model/georeferencing offset applied through `tool.Loader.apply_blender_offset_to_matrix_world` (`import_ifc.py:786/835/886`, etc.). The drawing camera import does not:

```python
# import_drawing / import_temporary_drawing_camera
obj.matrix_world = cls.get_camera_shape_matrix(drawing, shape)   # raw absolute map coords
```

So on a georeferenced model the rest of the scene is shifted to Blender-local via `global2local`, while the camera keeps absolute map coordinates and lands far away. Worse, on save `edit_object_placement` re-applies `local2global` to an already-absolute matrix, so the camera drifts further every save/reopen.

## Fix

Route the camera matrix through the same offset helper every other object uses, in both `import_drawing` and `import_temporary_drawing_camera`.

## Verification

- For **non-georeferenced** models the helper returns the matrix unchanged (camera has `obj.data`, no cartesian-point offset, `has_blender_offset` false), so behaviour is identical to before — no regression.
- For **georeferenced** models it applies `global2local` exactly like the 8 existing call sites; verified in isolation that an absolute placement (e.g. `1234567, 7654321`) is brought into the scene-local frame, matching the rest of the scene.

I don't have a Blender session with a georeferenced model to run the full save/reopen round-trip, so that end-to-end confirmation is welcome. This addresses symptom 2 (camera shift); symptom 1 (ortho zoom/crop not persisted) is a separate matter — `ortho_scale` is not stored in `EPset_Drawing` the way perspective `shift_x/shift_y` are — and is left for a maintainer to decide whether manual ortho zoom should be persisted.

### DIFF
--- a/src/bonsai/bonsai/tool/drawing.py
+++ b/src/bonsai/bonsai/tool/drawing.py
@@ -1007,7 +1007,14 @@ class Drawing(bonsai.core.tool.Drawing):
         cls.import_camera_props(drawing, camera)
         tool.Ifc.link(drawing, obj)
 
-        obj.matrix_world = cls.get_camera_shape_matrix(drawing, shape)
+        # Apply the georeferencing/model offset like every other imported object
+        # (see import_ifc.py). Without this the drawing camera keeps absolute map
+        # coordinates while the rest of the scene is shifted to Blender-local, so
+        # on a georeferenced model the camera lands far away and drifts further on
+        # each save/reopen. See #8205.
+        obj.matrix_world = tool.Loader.apply_blender_offset_to_matrix_world(
+            obj, np.array(cls.get_camera_shape_matrix(drawing, shape))
+        )
 
         tool.Geometry.record_object_position(obj)
         tool.Collector.assign(obj)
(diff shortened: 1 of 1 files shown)


BIMvoice:
Verified this end to end on a georeferenced reproduction.

Setup: a model with geometry baked at map coordinates so the blender offset engages on load (`has_blender_offset` true, offset E 158123.224 / N 6406512.544). Create a plan drawing, save, reopen, activate the drawing.

- without this PR: the model loads correctly local (slab at 8.6, 9.8) but the drawing camera lands at **(158123.22, 6406512.5)**, the raw projected coordinates. This is the "camera opens in nowhere" from #8205, and because the next save re-projects the already-absolute camera, it is also the doubling that was reported.
- with this PR: the same reopen puts the camera back at **(0.0, 0.0, 1.6)**, exactly where it was created.

So routing the imported drawing camera through `apply_blender_offset_to_matrix_world` fixes the shift and breaks the doubling cycle. Companion PRs for the rest of #8205: #8328 (drawing zoom/crop reset on save) and #8329 (BCF viewpoints on georeferenced models).

Issue #8203, opened 2026-06-25
Typo in resource.assign_resource duplicate guard causes duplicate RelatedObjects


john-rogers:
### Bug Description

In `src/ifcopenshell-python/ifcopenshell/api/resource/assign_resource.py`, the duplicate-assignment guard appears to contain a typo:

```python
if assignment.is_a("IfclRelAssignsToResource") and assignment.RelatingResource == relating_resource:
    return assignment
```

I believe `"IfclRelAssignsToResource"` should be `"IfcRelAssignsToResource"`.

This affects behaviour when assigning the same related object to the same resource more than once. The function does not create a second `IfcRelAssignsToResource`, but it appends the same object twice to `RelatedObjects`.

Minimal observed result:

Number of IfcRelAssignsToResource: 1
Same relationship returned: True
Related object ids: [8, 8]
Occurrences of test proxy: 2

Expected behaviour:

Number of IfcRelAssignsToResource: 1
Same relationship returned: True
Occurrences of test proxy: 1

Suggested fix:

```diff
- if assignment.is_a("IfclRelAssignsToResource") and assignment.RelatingResource == relating_resource:
+ if assignment.is_a("IfcRelAssignsToResource") and assignment.RelatingResource == relating_resource:
```

Tested locally in Bonsai / IfcOpenShell installed under Blender 4.5:

../.config/blender/4.5/extensions/.local/lib/python3.11/site-packages/ifcopenshell/api/resource/assign_resource.py

### Debug and Error Output

```true

```

BIMvoice:
Confirmed on the current `v0.8.0` and fixed in #8252. The stray `l` in `is_a("IfclRelAssignsToResource")` meant the guard never matched, so a repeat assignment appended the object to `RelatedObjects` twice. Red-green regression test added (fails before, passes after). Thanks for the clear write-up.

BIMvoice:
Fixed in #8305 — corrected the class name to `IfcRelAssignsToResource`. Verified red-green: before, `RelatedObjects` had the object twice (`[2, 2]`); after, once (`[2]`). Thanks for the clean report.

PR #8305, opened 2026-07-06
resource.assign_resource: fix typo in duplicate guard (#8203)

BIMvoice:
## Problem

Assigning the same object to the same resource twice appends it to `RelatedObjects` again, producing duplicates like `[8, 8]` (#8203). Thanks @john-rogers for the precise report.

## Cause

The duplicate guard checks `assignment.is_a("IfclRelAssignsToResource")` — a stray `l`. No such class exists, so the guard never matched and the early return was dead code. The second call then fell through to the append branch.

## Fix

Correct the class name to `"IfcRelAssignsToResource"` so the existing relationship is found and returned.

## Verification

Red-green with a minimal repro (assign the same actor to the same crew resource twice):
- Before: `RelatedObjects` ids `[2, 2]`
- After: `RelatedObjects` ids `[2]`, single `IfcRelAssignsToResource`

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/resource/assign_resource.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/resource/assign_resource.py
@@ -81,7 +81,7 @@ def assign_resource(
     """
     if related_object.HasAssignments:
         for assignment in related_object.HasAssignments:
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing: superseded by 21ae78fbc2 (merged), the identical typo fix with an equivalent regression test.

Issue #6398, opened 2025-03-19
Bug when add a IFC object


Arturo-arch5:
Python: Traceback (most recent call last):
  File "C:\Users\Arturo\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\root\operator.py", line 631, in execute
    bpy.ops.bim.add_element("INVOKE_DEFAULT")
  File "C:\Program Files\Blender Foundation\Blender 4.2\4.2\scripts\modules\bpy\ops.py", line 107, in __call__
    ret = _op_call(self.idname_py(), kw, C_exec, C_undo)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Error: Python: Traceback (most recent call last):
  File "C:\Users\Arturo\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\bonsai\bim\helper.py", line 300, in prop_with_search
    if len(get_enum_items(data, prop_name, original_operator_path=original_operator_path)) > 10:
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Arturo\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\bonsai\bim\helper.py", line 345, in get_enum_items
    items = items(data, context or bpy.context)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Arturo\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\root\prop.py", line 98, in get_ifc_products
    return IfcClassData.data["ifc_products"]
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
KeyError: 'ifc_products'
Ubicación: C:\Users\Arturo\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\root\operator.py:376

Andrej730:
@Arturo-arch5 
Couldn't reproduce the issue.

[...]

PR #8308, opened 2026-07-06
Bonsai: don't cache a half-built IfcClassData on load failure (#6398)

BIMvoice:
## Problem

Adding an IFC object could crash (#6398) with:

```
File ".../bonsai/bim/module/root/prop.py", line 98, in get_ifc_products
    return IfcClassData.data["ifc_products"]
KeyError: 'ifc_products'
```

## Cause

`IfcClassData.load()` set `is_loaded = True` as its first line, then populated `cls.data` key by key. If any of those populate calls raised (for instance an enum callback firing before a file is fully ready), `is_loaded` was already `True` with an incomplete (often empty) `data` dict. Because `get_ifc_products` only calls `load()` when `not is_loaded`, every subsequent call skipped the reload and dereferenced the missing key, raising `KeyError`.

## Fix

Set `is_loaded = True` only after all keys are populated. A failed load then leaves `is_loaded = False`, so the next call retries `load()` instead of returning a half-built dict. This matches how other `*Data` modules order the flag.

## Verification

Reproduced the pattern in isolation: with the flag set first, a transient first-call failure leaves the class poisoned and the second call raises `KeyError 'ifc_products'`; with the flag set last, the second call retries and succeeds.

### DIFF
--- a/src/bonsai/bonsai/bim/module/root/data.py
+++ b/src/bonsai/bonsai/bim/module/root/data.py
@@ -42,25 +42,36 @@ class IfcClassData:
 
     @classmethod
     def load(cls):
-        cls.data = {}
-        cls.data["ifc_products"] = cls.ifc_products()
-        cls.data["ifc_classes"] = cls.ifc_classes()
-        cls.data["ifc_classes_suggestions"] = cls.ifc_classes_suggestions()  # Call AFTER cls.ifc_classes()
-        cls.data["representation_template"] = cls.representation_template()
-        cls.data["contexts"] = cls.contexts()
-
-        cls.data["has_entity"] = cls.has_entity()
-        cls.data["name"] = cls.name()
-        cls.data["has_inherited_predefined_type"] = cls.has_inherited_predefined_type()
-        cls.data["ifc_class"] = cls.ifc_class()
-        cls.data["ifc_predefined_types"] = cls.ifc_predefined_types()
-        cls.data["can_reassign_class"] = cls.can_reassign_class()
-        cls.data["profile"] = cls.profile()
-        # Set only after every key is populated. If a populate call above raises
-        # (e.g. before a file is fully ready), is_loaded stays False so the next
-        # call retries load() instead of returning a half-built dict, which
-        # otherwise surfaced as KeyError 'ifc_products'. See #6398.
+        # is_loaded must be set True *before* the helpers below read the ifc_product /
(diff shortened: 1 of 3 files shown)


theoryshaw:
typing in any text field lagged on every keystroke because panels redraw per keystroke, and the `ifc_product`/`ifc_class` enum callbacks re-ran the expensive `IfcClassData.load()` (a full schema traversal) each time. This happened because PR #8308 moved `is_loaded = True` to the end of `load()`, so a stale enum value that made `declaration_by_name()` raise left `is_loaded`永 `False` — meaning the cache never took and the heavy load repeated on every redraw.

the following PR on this PR addresses that issue...

https://github.com/BIMvoice/IfcOpenShell/pull/1

theoryshaw:
converting to 'draft' so it's not pushed to the BonsaiPR build, in the meantime.

BIMvoice:
Thanks @theoryshaw, that diagnosis is exactly right and your fix is the correct shape. I have folded your two commits into this PR branch (authorship preserved) and taken it back out of draft.

Root cause confirmed: deferring `is_loaded = True` to the end of `load()` meant that on a file whose `ifc_product` / `ifc_class` enum holds a stale/out-of-range stored index, `declaration_by_name()` raised before the flag was set, so `is_loaded` stayed `False`. The dynamic enum `items=` callbacks then re-ran the full `load()` on every panel redraw (the per-keystroke freeze), and the nested re-entry surfaced as `KeyError: 'ifc_products'`.

The fix restores the up-front `is_loaded = True` re-entrancy guard, keeps the #6398 half-built-cache guarantee via `try/except` (clear the flag and re-raise on a genuine populate failure), and guards the two `declaration_by_name` calls so a stale enum yields empty lists instead of raising.

Verified live in headless Blender 5.1.2, running your `test_root_data.py` scenarios against both states:

- Regression state (this PR before your fix, `is_loaded` set at the end, unguarded `declaration_by_name`): both the valid-product load and the stale-enum load raise `RuntimeError: Entity with name '' not found in schema 'IFC4'` from the unresolved enum, so `load()` never completes and `is_loaded` stays `False` (the reload-every-redraw path).
- Fixed state (your commits): both pass. A valid product resolves and populates the class list; a stale enum index (`props["ifc_product"] = 9999`) completes without raising, keeps the fully-built dict, sets `is_loaded = True`, and returns empty `ifc_classes` / `ifc_predefined_types`.

black and ruff (line-length 120) are clean on `data.py` and the new test.

Issue #7696, opened 2026-02-20
Bonsai/IfcOpenShell rewrites all float values


brunopostle:
### Bug Description

This is a change from sometime in the last month (since 2026-01-13). If I open a file and save it, all the float values are rewritten:

``` diff
diff --git a/model.ifc b/model.ifc
index fac1d2d..f013509 100644
--- a/model.ifc
+++ b/model.ifc
@@ -1,7 +1,7 @@
 ISO-10303-21;
 HEADER;
 FILE_DESCRIPTION(('ViewDefinition[DesignTransferView]'),'2;1');
-FILE_NAME('model.ifc','2026-01-17T11:28:20+00:00',(''),(''),'IfcOpenShell 0.0.0','Bonsai 0.0.0','Nobody');
+FILE_NAME('model.ifc','2026-02-20T14:42:55+00:00',(''),(''),'IfcOpenShell 0.0.0','Bonsai 0.0.0','Nobody');
 FILE_SCHEMA(('IFC4'));
 ENDSEC;
 DATA;
@@ -11,14 +11,14 @@ DATA;
 #4=IFCSIUNIT(*,.VOLUMEUNIT.,$,.CUBIC_METRE.);
 #5=IFCDIMENSIONALEXPONENTS(0,0,0,0,0,0,0);
 #6=IFCSIUNIT(*,.PLANEANGLEUNIT.,$,.RADIAN.);
-#7=IFCMEASUREWITHUNIT(IFCREAL(0.0174532925199433),#6);
+#7=IFCMEASUREWITHUNIT(IFCREAL(0.017453292519943299),#6);
 #8=IFCCONVERSIONBASEDUNIT(#5,.PLANEANGLEUNIT.,'degree',#7);
 #9=IFCUNITASSIGNMENT((#8,#3,#4,#2));
 #10=IFCCARTESIANPOINT((0.,0.,0.));
 #11=IFCDIRECTION((0.,0.,1.));
 #12=IFCDIRECTION((1.,0.,0.));
 #13=IFCAXIS2PLACEMENT3D(#10,#11,#12);
-#14=IFCGEOMETRICREPRESENTATIONCONTEXT($,'Model',3,1.E-05,#13,$);
+#14=IFCGEOMETRICREPRESENTATIONCONTEXT($,'Model',3,1.0000000000000001E-05,#13,$);
[...]

PR #8309, opened 2026-07-06
IfcParse: serialize REALs with shortest round-trip form (#7696)

BIMvoice:
## Problem

Opening a file and saving it rewrites every REAL value with extra noise digits (#7696), producing enormous diffs for anyone version-controlling IFC:

```diff
-#7=IFCMEASUREWITHUNIT(IFCREAL(0.0174532925199433),#6);
+#7=IFCMEASUREWITHUNIT(IFCREAL(0.017453292519943299),#6);
-#14=IFCGEOMETRICREPRESENTATIONCONTEXT($,'Model',3,1.E-05,#13,$);
+#14=IFCGEOMETRICREPRESENTATIONCONTEXT($,'Model',3,1.0000000000000001E-05,#13,$);
```

Confirmed still present on a current 0.8.6-alpha build.

## Cause

`format_double` (`src/ifcparse/IfcParse.cpp`) serialized doubles with `std::setprecision(std::numeric_limits<double>::max_digits10)` (17 significant digits). max_digits10 guarantees a round-trip but is not the *shortest* such representation, so clean values like `0.0174532925199433` gained trailing noise (`...99`).

## Fix

Use `std::to_chars`, which emits the shortest string that round-trips exactly (the same thing Python's `repr` does), then keep the existing mantissa/exponent post-formatting unchanged.

## Verification

Same standalone-isolation approach used for other IfcParse fixes: I compiled the exact function logic (old vs new) over a range of doubles.

| value | old (max_digits10) | new (to_chars) | round-trips |
|---|---|---|---|
| 0.0174532925199433 | 0.017453292519943299 | **0.0174532925199433** | yes |
| 1e-05 | 1.0000000000000001E-05 | **1.E-05** | yes |
| 0.1 | 0.10000000000000001 | **0.1** | yes |
| 2.2250738585072014e-308 (denormal) | same | same | yes |
| 0, 1, 123456.789, pi | correct | correct | yes |

Every tested value produces the shortest form and `strtod` parses it back to the identical double. Added `#include <charconv>`. A full build is the final integration check, but the function logic is verified.

### DIFF
--- a/src/ifcparse/IfcParse.cpp
+++ b/src/ifcparse/IfcParse.cpp
@@ -39,6 +39,7 @@
 #include <stdlib.h>
 #include <string>
 #include <iomanip>
+#include <charconv>
 
 #ifdef USE_MMAP
 #include <boost/filesystem/path.hpp>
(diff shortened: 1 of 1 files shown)


aothms:
I think this makes sense, but then please also remove the dependency on

std::ostringstream oss;
            oss.imbue(std::locale::classic());

BIMvoice:
Done. Removed the `std::ostringstream` and `imbue(std::locale::classic())` from `format_double` and build the string with plain `std::string` operations, since `std::to_chars` is already locale-independent. Verified in a standalone compile that the output is unchanged (same shortest values, all round-trip). (`format_binary` still uses an ostringstream, so the `<sstream>` include stays for that one.)

Issue #5873, opened 2024-12-13
Assigning numpy array directly to IFC attribute


Andrej730:
Noticed that it doesn't seem to be possible to assign a numpy array directly to the attribute. At first I thought it's swig failing to convert non-default (tuples/list) Python types to c++ vectors but `mathutils.Vector` seems to work too. Any ideas what's the difference?

```python
import ifcopenshell
import numpy as np
from mathutils import Vector
from array import array

ifc_file = ifcopenshell.file()
# # 1=IfcCartesianPoint((3.,4.))
print(ifc_file.create_entity("IfcCartesianPoint", Vector([3.0, 4.0])))
# # 2=IfcCartesianPoint((3.,4.))
print(ifc_file.create_entity("IfcCartesianPoint", array("f", [3.0, 4.0])))
# TypeError: attribute 'Coordinates' for entity 'IFC4.IfcCartesianPoint' is expecting value of type 'AGGREGATE OF DOUBLE', got 'ndarray'.
pt = ifc_file.create_entity("IfcCartesianPoint", np.array([3.0, 4.0]))
```

aothms:
See https://github.com/IfcOpenShell/IfcOpenShell/blob/7f1b234045380163ab8fdc50afb2d5b6d02e1a45/src/ifcwrap/utils/type_conversion.i#L32-L45

I'm actually really surprised that mathutils.Vector would work, because the check is very restrictive.

In particular not considering inheritance: `element->ob_type == type_obj`

Most likely that is culprit. Indexing into numpy arrays does not return direct Floats, but numpy's own scalar type, I guess that's where it's different from mathutils.Vector.

I also don't know why I made that so restrictive to begin with.. I think because I wanted to educate people on the difference between REAL and INTEGER in express, because SPF is strict about them. `1` is not `1.`. But that's hardly a good reason to influence the behaviour of our library...

Feel free to make this a bit more lenient.

BIMvoice:
Fixed in #8310 (merged): `entity_instance` now accepts numpy arrays when setting attributes. Closing.

BIMvoice:
Reopening, on reflection I shouldn't close someone else's report myself. The fix is merged in #8310, so this should be resolved on a recent build, but I'll leave confirming and closing it to you.

BIMvoice:
[...]

PR #8310, opened 2026-07-06
entity_instance: accept numpy arrays when setting attributes (#5873)

BIMvoice:
## Problem

Assigning a numpy array directly to an aggregate attribute failed (#5873):

```python
ifc_file.create_entity("IfcCartesianPoint", np.array([3.0, 4.0]))
# TypeError: attribute 'Coordinates' ... is expecting 'AGGREGATE OF DOUBLE', got 'ndarray'
```

A `list`, `array.array`, and `mathutils.Vector` all work; only numpy is rejected. As @aothms noted on the issue, the strict SWIG check compares exact Python types (`ob_type == type_obj`), so an `ndarray` container and numpy scalar elements (`np.float64`) are refused, and he invited making it more lenient.

## Fix

In `entity_instance.walk` (the traversal used when unwrapping values for the setter), convert any value exposing `tolist()` to native Python types before it reaches the C++ setter. numpy's `tolist()` returns real Python `float`/`int`, so both the container and its elements become acceptable. This also covers numpy scalars and `array.array`, and recurses for nested arrays.

## Verification

Red-green on current code (file-swapped into the installed package):
- `np.array([3.0, 4.0])` -> `#1=IfcCartesianPoint((3.,4.))`
- `[np.float64(3.0), np.float64(4.0)]` -> `IfcCartesianPoint((3.,4.))`
- `np.array([[0,0,0],[1,1,1]])` -> `IfcCartesianPointList3D(((0.,0.,0.),(1.,1.,1.)))` (nested)
- plain lists and string attributes (e.g. GlobalId) unchanged.

### DIFF
--- a/src/ifcwrap/utils/type_conversion.i
+++ b/src/ifcwrap/utils/type_conversion.i
@@ -34,9 +34,19 @@
 		if (PySequence_Size(aggregate) == -1) return false;
 		for(Py_ssize_t i = 0; i < PySequence_Size(aggregate); ++i) {
 			PyObject* element = PySequence_GetItem(aggregate, i);
-			// This is equivalent to the PyFloat_CheckExact macro. This means
-			// that direct instances of int, float, str, etc. need to be used.
-			bool b = element->ob_type == type_obj;
+			// Accept the exact type or, for the numeric types, a subclass such
+			// as a numpy scalar (numpy.float64 subclasses float), so that numpy
+			// arrays can be assigned. The REAL vs INTEGER distinction is kept: a
+			// float is not accepted where an int is expected and vice versa, and
+			// bool (a subclass of int) is still rejected for INTEGER. See #5873.
+			bool b;
+			if (type_obj == static_cast<void*>(&PyFloat_Type)) {
+				b = PyFloat_Check(element);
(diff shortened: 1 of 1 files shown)


aothms:
I'd rather not. Either fix it in the C++ side or leave it as is. In v0.9 we don't have the walk() call anymore, so it'd be weird to introduce this for a couple of months and then take it away.

BIMvoice:
Good call, moved it to the C++ side and dropped the `walk()` change. `check_aggregate_of_type` now accepts subclasses for the numeric types: `PyFloat_Check` for double (`numpy.float64` subclasses `float`) and `PyLong_Check` (minus `bool`) for int, keeping the REAL vs INTEGER distinction. Verified with a runtime red-green built as a shared lib and called via ctypes: the old exact check rejects `np.array([3.0, 4.0])`, the new one accepts it, plain lists still work, an int list is still rejected where a REAL is expected, and `bool` stays rejected for INTEGER. Note `numpy.float32` and numpy int scalars are not subclasses of `float`/`int` so they still won't pass without a numpy dtype dependency, but the common float64 case (which the report used) works.

Issue #8237, opened 2026-07-02
Regression incorrectly generating space boundaries for windows (`IfcRelSpaceBoundary`)


CyrilWaechter:
### Bug Description

Space boundaries for windows are now incorrectly generated:

1. Open attached IFC file

<img width="564" height="536" alt="Image" src="https://github.com/user-attachments/assets/77df67ca-5a26-43a9-8d34-6f36da919aca" />

2. In spatial tools. Generace `IfcRelSpaceBoundary2ndLevel` for `Space`
3. See that windows space boundaries are misplaced (they are supposed to be placed aligned to corresponding windows)

<img width="575" height="580" alt="Image" src="https://github.com/user-attachments/assets/fd5c2233-9725-4cd2-b6d1-701955570b5d" />

### Attachments

[RegressionInWindowsSpaceBoundariesGeneration.zip](https://github.com/user-attachments/files/29602337/RegressionInWindowsSpaceBoundariesGeneration.zip)

### Debug and Error Output

```
os: Linux
os_version: #1 SMP PREEMPT_DYNAMIC Wed, 03 Jun 2026 22:05:15 +0000
python_version: 3.14.6
architecture: ('64bit', 'ELF')
machine: x86_64
processor: 
blender_version: 5.1.2
bonsai_version: 0.0.0
bonsai_commit_hash: a65f291
bonsai_commit_date: 2026-06-30T10:23:06+02:00
bonsai_git_branch: v0.8.0
last_actions: 
# operator: BIM_OT_show_boundaries
# operator: BIM_OT_save_project
# operator: BIM_OT_add_boundary
# ifcopenshell.api: root.create_entity
>>> {"ifc_class": "IfcRelSpaceBoundary2ndLevel"}
# ifcopenshell.api: owner.create_owner_history
>>> {}
# ifcopenshell.api: root.create_entity
>>> {"ifc_class": "IfcRelSpaceBoundary2ndLevel"}
# ifcopenshell.api: owner.create_owner_history
>>> {}
# ifcopenshell.api: root.create_entity
>>> {"ifc_class": "IfcRelSpaceBoundary2ndLevel"}
# ifcopenshell.api: owner.create_owner_history
>>> {}
# ifcopenshell.api: root.create_entity
>>> {"ifc_class": "IfcRelSpaceBoundary2ndLevel"}
# ifcopenshell.api: owner.create_owner_history
>>> {}
# ifcopenshell.api: root.create_entity
[...]

PR #8311, opened 2026-07-06
Bonsai: place auto-generated opening boundaries at their real position (#8237)

BIMvoice:
## Problem

Generating 2nd-level space boundaries by selecting a space (the single-space auto mode) placed window and door boundaries in the wrong location, not aligned to their openings (#8237). @MDHering pinned it down: selecting a space **plus one element** (`create_element_boundary`) gives correct window placement, while selecting the **space alone** (`auto_generate_boundaries`) does not. Those are two separate code paths.

## Cause

In `auto_generate_boundaries`, each opening boundary is built from the opening's geometry:

```python
shape = ifcopenshell.geom.create_shape(settings, opening)
mat = Matrix(ifcopenshell.util.shape.get_shape_matrix(shape))
mat.translation = (0, 0, 0)   # <-- discards the opening's position
...
verts = ifcopenshell.util.shape.get_vertices(shape.geometry)   # LOCAL coords
...
opening_face_verts = [space_matrix_world_i @ mat @ v.co for v in opening_face.verts]
```

`get_vertices` returns the opening's **local** geometry, so `mat` (its placement, translation included) is exactly what carries it to its real position. Zeroing the translation collapsed every opening onto the origin. The wall itself is placed with its real `building_obj.matrix_world`, so openings were handled inconsistently.

## Fix

Keep the full placement matrix (drop the `mat.translation = (0, 0, 0)` line). The rotation-only use of `mat` (for the face-normal test) is unaffected.

## Verification

On the reporter's `RegressionInWindowsSpaceBoundariesGeneration.ifc`, the first opening's real placement translation is `(0.1, 1.5, 1.0)`. A vertex is placed at `(0.6, 0, 0)` under the old code (collapsed toward origin) and at `(0.7, 1.5, 1.0)` with the fix, i.e. shifted by exactly the `(0.1, 1.5, 1.0)` that was being discarded. That offset is the misplacement. A Blender run on the file is the final visual confirmation; @MDHering / @CyrilWaechter, this should align the window boundaries in the single-space mode.

### DIFF
--- a/src/bonsai/bonsai/bim/module/boundary/operator.py
+++ b/src/bonsai/bonsai/bim/module/boundary/operator.py
@@ -842,13 +842,6 @@ class AddBoundary(bpy.types.Operator, tool.Ifc.Operator):
                         # Create shape of opening as a dissolved BMesh
                         settings = ifcopenshell.geom.settings()
                         shape = ifcopenshell.geom.create_shape(settings, opening)
-                        # get_vertices returns the opening's LOCAL geometry, so mat
-                        # (its placement, including translation) is what carries the
-                        # opening to its real position below. Zeroing the translation
-                        # collapsed every opening onto the origin, which is why
-                        # auto-generated window/door boundaries were misplaced while
-                        # the single-element path (create_element_boundary) was
-                        # correct. See #8237.
                         mat = Matrix(ifcopenshell.util.shape.get_shape_matrix(shape))
                         opening_bm = bmesh.new()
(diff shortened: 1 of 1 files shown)


CyrilWaechter (commented):
Code change seems good to me. I have tested in on several test model from https://github.com/CyrilWaechter/IfcRelSpaceBoundary_TestFiles without issue. Although the long comment seems superfluous and not complying to project standards. Keeping it in the commit message is enough.

Issue #8074, opened 2026-05-18
Bonsai: Error when opening project with defult LengthUnit set to DECIMETRE


semhustej:
### Bug Description

Bonsai is not able to open file with default LengthUnit set to `DECIMETRE`.

I am able to reproduce in Bonsai 0.8.5.

The issue does not happen when testing with METRE, MILIMETRE, or CENTIMETRE units

### Attachments

**Steps to reproduce:**

- open sample file (remove the .txt extension). The file was created using Bonsai

[dm.ifc.txt](https://github.com/user-attachments/files/27952246/dm.ifc.txt)

- file does not open, error shows up

### Debug and Error Output

```true
Error message:


os: Linux
os_version: #1 SMP PREEMPT_DYNAMIC Debian 6.12.88-1 (2026-05-15)
python_version: 3.13.9
architecture: ('64bit', 'ELF')
machine: x86_64
processor: 
blender_version: 5.1.0
bonsai_version: 0.8.6-alpha260506
bonsai_commit_hash: c197a45
bonsai_commit_date: 2026-05-06T13:32:05+02:00
bonsai_git_branch: v0.8.0
last_actions: 
# ifcopenshell.api: owner.update_owner_history
>>> {"element": {"cast_type": "entity_instance", "value": 33, "Name": "My Building"}}
# ifcopenshell.api: root.create_entity
>>> {"ifc_class": "IfcBuildingStorey", "predefined_type": "None", "name": "My Storey"}
# ifcopenshell.api: owner.create_owner_history
>>> {}
# ifcopenshell.api: geometry.edit_object_placement
>>> {"product": {"cast_type": "entity_instance", "value": 39, "Name": "My Storey"}, "matrix": {"cast_type": "ndarray", "value": [[1.0, 0.0, 0.0, 0.0], [0.0, 1.0, 0.0, 0.0], [0.0, 0.0, 1.0, 0.0], [0.0, 0.0, 0.0, 1.0]]}}
# ifcopenshell.api: owner.update_owner_history
>>> {"element": {"cast_type": "entity_instance", "value": 39, "Name": "My Storey"}}
# ifcopenshell.api: aggregate.assign_object
>>> {"products": [{"cast_type": "entity_instance", "value": 27, "Name": "My Site"}], "relating_object": {"cast_type": "entity_instance", "value": 1, "Name": "My Project"}}
# ifcopenshell.api: spatial.unassign_container
>>> {"products": "[]"}
[...]

PR #8316, opened 2026-07-06
Handle unsupported length unit prefixes on project load (#8074)

BIMvoice:
## Problem

Bonsai crashes when opening any IFC whose default `LengthUnit` uses an SI prefix that Blender's `length_unit` enum does not support. Reported in #8074 for `DECIMETRE` (reproducible on the attached `dm.ifc`), but the same crash applies to `DECA`, `HECTO`, `MEGA`, `NANO`, and other valid IFC prefixes.

## Root cause

`import_ifc.py` `set_units()` builds the Blender enum value as `f"{unit.Prefix}METERS"`. For `DECI` this produces `"DECIMETERS"`, which is not a member of Blender's enum:

```
TypeError: bpy_struct: item.attr = val: enum "DECIMETERS" not found in
('ADAPTIVE', 'KILOMETERS', 'METERS', 'CENTIMETERS', 'MILLIMETERS', 'MICROMETERS')
```

The exception propagates out of `load_project_elements` and aborts the whole import, so the file never opens.

## Fix

Wrap the `length_unit` assignment in `try/except TypeError` and fall back to `METERS` for display when the prefix is not a valid Blender enum. This mirrors the existing `try/except` fallbacks for `AREAUNIT` (`SQUARE_METRE`) and `VOLUMEUNIT` (`CUBIC_METRE`) a few lines below in the same function. Only the Blender display unit falls back; the model's real unit assignment is unchanged.

## Verification

Confirmed headless in Blender 5.1.2: assigning `length_unit = "DECIMETERS"` raises the reported `TypeError`, the `METERS` fallback is accepted, and the supported prefixes (KILO/CENTI/MILLI/MICRO) still apply. Files that previously failed to open now load.

### DIFF
--- a/src/bonsai/bonsai/bim/import_ifc.py
+++ b/src/bonsai/bonsai/bim/import_ifc.py
@@ -978,10 +978,16 @@ class IfcImporter:
                 if unit.is_a("IfcSIUnit"):
                     bpy.context.scene.unit_settings.system = "METRIC"
                     if unit.Name == "METRE":
-                        if not unit.Prefix:
+                        try:
+                            if not unit.Prefix:
+                                bpy.context.scene.unit_settings.length_unit = "METERS"
+                            else:
+                                bpy.context.scene.unit_settings.length_unit = f"{unit.Prefix}METERS"
+                        except TypeError:
+                            # Blender's length_unit enum only supports a subset of SI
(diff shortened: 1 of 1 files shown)


BIMvoice:
Closing: superseded by 69a4be68e8 (merged), a functionally equivalent fix for #8074 (whitelist check with ADAPTIVE fallback at the same location).

Issue #7147, opened 2025-09-21
Error w/ manual quantification


theoryshaw:
### Bug Description

When i try and do  `bim.calculate_single_quantity` on something like the following object, I get the following error

GlobalId = "3Bwmk9gjDFHgxXMQJ71Nuk"

file: https://hub.openingdesign.com/OpeningDesign/Restaurant_Brookfield/src/commit/c9c115a985b3babcca740f46be994595108edd88/Open/Models/Bonsai/Restaurant_Brookfield.ifc

<img width="292" height="149" alt="Image" src="https://github.com/user-attachments/assets/56bc2a65-5ef8-476c-b3b3-672291041e8d" />

### Debug and Error Output

```true
Python: Traceback (most recent call last):
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\ifc.py", line 526, in execute_ifc_operator
    result = getattr(operator, "_execute")(context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\qto\operator.py", line 151, in _execute
    results = ifc5d.qto.quantify(ifc_file, elements, rules)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\ifc5d\qto.py", line 112, in quantify
    calculator.calculate(ifc_file, filtered_elements, qtos, results)
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\ifc5d\qto.py", line 355, in calculate
    value = formula_functions[formula](geometry)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\ifcopenshell\util\shape.py", line 118, in get_z
[...]

PR #8318, opened 2026-07-06
Skip quantity takeoff for elements with empty geometry (#7147)

BIMvoice:
## Problem

Quantity takeoff crashed in #7147:

```
File ".../ifcopenshell/util/shape.py", line 118, in get_z
    return (np.max(verts_flat[2::3]) - np.min(verts_flat[2::3])).item()
ValueError: zero-size array to reduction operation maximum which has no identity
```

An element can be included by the geometry iterator yet still produce empty geometry (no vertices). The bounding-box / elevation helpers in `util.shape` then call `np.max`/`np.min` on a zero-size array and raise, aborting the whole takeoff run.

## Fix (revised after review)

The first version made the `util.shape` helpers return `0.0` for empty geometry. As @aothms pointed out, that is wrong: no geometry is not the same as a bounding box or bottom elevation of `0.0`, and silently fabricating that value corrupts quantities. Reverted.

Instead this handles it at the domain layer: `ifc5d` qto skips an element's geometric quantity when the geometry has no vertices, matching the existing pattern of `continue`-ing past quantities that cannot be computed (`get_segment_length`/`get_weight` returning `None`). The `util.shape` helpers are left unchanged.

## Verification

Confirmed `geometry.verts` is the right accessor and `len(geometry.verts) == 0` cleanly flags empty geometry (0 for empty, 24 for a real wall) via the geom iterator, so the guard skips only genuinely empty elements and does not alter normal takeoff.

### DIFF
--- a/src/ifc5d/ifc5d/qto.py
+++ b/src/ifc5d/ifc5d/qto.py
@@ -366,6 +366,13 @@ class IfcOpenShell(QtoCalculator):
                                 if value is None:
                                     continue
                             else:
+                                # An element can be included by the iterator yet still
+                                # produce empty geometry (no vertices). Bounding-box and
+                                # elevation helpers can't be evaluated on it, and emitting
+                                # a fabricated value (e.g. 0.0) would silently corrupt the
+                                # quantity, so skip it instead.
+                                if len(geometry.verts) == 0:
+                                    continue
                                 value = formula_functions[formula](geometry)
                                 assert isinstance(value, (float, int))
                                 value = cls.unit_converter.convert(value, IfcOpenShell.raw_functions[formula].measure)


aothms:
I don't think this is correct. No geometry is not the same as a bottom elevation of 0.0. I would be really careful in blindly patching these things. Especially since they affect quantities.

BIMvoice:
You're completely right, thanks. I reverted the `util.shape` changes. Fabricating `0.0` for missing geometry is wrong, especially for elevations, and it would corrupt quantities silently.

I moved the fix into `ifc5d` qto instead: it now skips an element's geometric quantity when the geometry has no vertices, matching the existing pattern where `get_segment_length`/`get_weight` return `None` and the loop continues. That avoids the crash without inventing a value. The `util.shape` helpers are left unchanged. Let me know if you'd prefer it handled elsewhere.

PR #8320, opened 2026-07-06
ifcpatch(ExtractElements): defer shared relationship members to fix O(n²) extraction

BIMvoice:
## Problem

`ExtractElements` is O(n^2) and becomes unusable at scale. A reporter's extraction of 2004 walls from a 495 MB IFC2X3 model ran for over 12 hours without finishing.

## Root causes

Two, both in `project.append_asset`.

**1. Shared relationship member lists were reassigned per element.** When a relationship is shared by many extracted elements, for example one `IfcRelAssociatesMaterial` linking thousands of products, `add_inverse_element` re-scanned and re-assigned that relationship's whole member list on every single append. Reassigning an N member list N times is O(n^2), dominated by entity hashing during set de-duplication.

**2. The asset uniqueness flag was not forwarded to the nested type append.** `append_product` recursed into `append_asset` to add an element's type but dropped `assume_asset_uniqueness_by_name`, which defaults to `True`. `ExtractElements` passes `False`, so the nested call silently took the slow path. That armed `SafeRemovalContext`, whose `__exit__` full scans the shared, ever growing `reuse_identities` once per occurrence. Profiling the railing baseline: `entity_instance.__hash__` 412M calls for 109.1s, and `__exit__` accounting for **378.2s of 430.7s**.

The second cost is per distinct type, so element classes with many types suffer most: this model has 97 window types against 12 railing types.

## Fix

- An opt-in `deferred_relationship_members` accumulator. Member lists are assigned once at the end via `flush_deferred_relationship_members` rather than per element. Opt-in, so `append_asset` default behaviour is unchanged for all other callers.
- The uniqueness flag is forwarded to the nested type append.
- Deferred members that are `IfcRoot` subtypes are matched by GlobalId; members that are not, which today is only `IfcRelOverridesProperties.OverridingProperties` holding `IfcProperty`, are matched by identity. Without this they resolved to nothing and the list was written empty, violating the `[1:?]` cardinality and producing invalid IFC.

## Measured

On the reporter's 519 MB IFC2X3 model, real `Patcher.patch()`:

| element class | before | after |
|---|---|---|
| `IfcWindow` (3557) | 1339.7s | **102.9s** |
| `IfcRailing` (2583) | 430.7s | **35.4s** |

Synthetic, N walls sharing a relationship: 500 3.86s to 0.24s, 1000 15.55s to 0.66s, 2000 64.47s to 2.36s. Base quadruples per doubling; the fix does not.

Peak memory on the real model fell from 4.19 GB to 2.58 GB.

## Correctness

Verified against a `v0.8.0` baseline on the real model: identical wall GlobalId sets, identical per class entity counts, identical relationship member counts, 0 dangling references, 0 empty member lists, and `ifcopenshell.validate` error counts no worse.

An earlier revision of this branch did lose data. It dropped the property sets of 26 spatial containers, one site, one building and 24 storeys, and emitted 26 `IfcRelDefinesByProperties` with empty `RelatedObjects`. It also dropped element types from shared associations and removed `IfcDistributionPort` entities entirely. All three are fixed here and covered by tests.

The old eager path was additionally emitting duplicate geometry: every window occurrence pointed at a representation map its own type did not declare, with 1313 orphaned `IfcRepresentationMap`. Output on this model falls from 1,044,914 to 688,899 entities with no loss of information.

## Known caveat

Presentation layer assignments change. Railings strictly improve, 3 empty and 0 items becoming 0 empty and 3 items. Windows are mixed: all 10 empty assignments become non-empty, but total membership falls from 85 to 12. The structural cause is `add_inverse_element` returning early when revisiting a non-relationship inverse, which makes layer membership order dependent in both the old and new code. Fixing that properly needs a wider change than belongs in this PR.

## Tests

Cover shared relationship members surviving, spatial container property sets surviving, element types surviving shared associations, ports and populated `IfcRelNests` surviving, and non `IfcRoot` deferred members resolving.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/project/__init__.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/project/__init__.py
@@ -27,7 +27,7 @@ into your project.
 """
 
 from .. import wrap_usecases
-from .append_asset import append_asset
+from .append_asset import append_asset, flush_deferred_relationship_members
 from .assign_declaration import assign_declaration
 from .create_file import create_file
 from .unassign_declaration import unassign_declaration
@@ -36,6 +36,7 @@ wrap_usecases(__path__, __name__)
 
 __all__ = [
     "append_asset",
+    "flush_deferred_relationship_members",
     "assign_declaration",
     "create_file",
     "unassign_declaration",
--- a/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py
@@ -472,6 +472,7 @@ class Usecase:
                 library=self.settings["library"],
                 element=element_type,
                 reuse_identities=self.reuse_identities,
+                assume_asset_uniqueness_by_name=self.assume_asset_uniqueness_by_name,
             )
             ifcopenshell.api.type.assign_type(
                 self.file,
@@ -612,8 +613,14 @@ class Usecase:
                         continue
                     if skip_not_reused_entities_attr_i is not None and i == skip_not_reused_entities_attr_i:
                         identity = item.wrapped_data.identity()
-                        if (item := self.reuse_identities.get(identity)) is None:
+                        reused = self.reuse_identities.get(identity)
+                        if reused is None:
+                            # reuse_identities is only filled by file_add's per-attribute
+                            # path; the native file.add fast path records added_elements.
+                            reused = self.added_elements.get(item.id())
+                        if reused is None:
                             continue
+                        item = reused
                     else:
                         item = self.add_element(item)
                     new_attribute.append(item)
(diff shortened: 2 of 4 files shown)


BIMvoice:
Closing this as superseded by #9050, to take one PR out of the review queue rather than leave two that do the same thing.

**#9050 fully contains this PR.** Verified rather than assumed:

```
git merge-base --is-ancestor fe08cec4ee 5fe943835f   ->  true
```

This PR's head commit is a genuine ancestor of #9050's head, and #9050 is exactly this branch plus one further commit ("keep every element on its presentation layer"). Every file touched here is also touched there. So merging #9050 delivers all of this work, and merging this one afterwards would be a no-op.

Nothing is lost by closing it. If #9050 is ever rejected on the strength of that extra commit, this can be reopened as the smaller change.

For context on the work itself, the O(n^2) claim in these commits was independently measured and holds up: extracting elements that share a relationship rebuilt the whole member list once per element. Timings on walls sharing one `IfcRelAssociatesMaterial`, median of five runs each:

| Elements | Before | After | Ratio |
|---|---|---|---|
| 100 | 0.118s | 0.042s | 2.8x |
| 500 | 2.269s | 0.209s | 10.9x |
| 2000 | 32.80s | 0.922s | **35.6x** |

At 5000 the unfixed version did not finish within six minutes, against 2.2s fixed, so no figure is quoted for it rather than extrapolating one.

That evidence belongs to #9050 now.

Issue #5113, opened 2024-07-31
Error when opening file


theoryshaw:
Error when trying to open the following file:
https://hub.openingdesign.com/OpeningDesign/Rae_Living_Room_Ext/src/commit/06ae2f09ecb9d7d008cc6f7d6908e1a186405f48/Open/Models/BlenderBIM/Rae_Living_Room_Ext.ifc

```
os: Windows
os_version: 10.0.19045
python_version: 3.11.7
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: Intel64 Family 6 Model 158 Stepping 10, GenuineIntel
blender_version: 4.2.0
blenderbim_version: 0.7.11-alpha240731
blenderbim_commit_hash: 6cb063e
last_actions: 
# operator: BIM_OT_activate_external_style
# operator: BIM_OT_activate_external_style
# operator: BIM_OT_activate_external_style
# operator: BIM_OT_activate_external_style
# operator: BIM_OT_activate_external_style
# operator: BIM_OT_activate_external_style
# operator: BIM_OT_activate_external_style
# operator: BIM_OT_activate_external_style
# operator: BIM_OT_activate_external_style
# operator: BIM_OT_activate_external_style
last_error: Traceback (most recent call last):
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\blenderbim\bim\module\project\operator.py", line 745, in finish_loading_project
    bpy.ops.bim.load_project_elements()
  File "C:\Program Files\Blender Foundation\Blender 4.2\4.2\scripts\modules\bpy\ops.py", line 109, in __call__
    ret = _op_call(self.idname_py(), kw)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Error: Python: Traceback (most recent call last):
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\blenderbim\bim\module\project\operator.py", line 824, in execute
    ifc_importer.execute()
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\blenderbim\bim\import_ifc.py", line 244, in execute
    self.create_elements()
[...]

PR #8321, opened 2026-07-06
Guard get_mappeditem_transformation against a removed MappingSource (#7170)

BIMvoice:
## Problem

Opening the file in #7170 crashed during import:

```
File ".../ifcopenshell/util/placement.py", in get_mappeditem_transformation
    m4 = get_axis2placement(item.MappingSource.MappingOrigin)
AttributeError: 'NoneType' object has no attribute 'MappingOrigin'
```

## Root cause

The file contains an `IfcMappedItem` whose `MappingSource` (an `IfcRepresentationMap`) has been removed, leaving a dangling reference. `item.MappingSource` then reads as `None`, and `get_mappeditem_transformation` dereferences `.MappingOrigin` on it. This aborts the whole file load (`bonsai` `is_native` calls it during import).

## Fix

Return an identity transform when `MappingSource` (or its `MappingOrigin`) is missing. A mapped item with no source has no mapping to resolve, so identity is the correct no-op, and the malformed entity no longer takes down the entire import. The valid path is unchanged; the guard only triggers on the `None` case.

## Verification

Reproduced by creating an `IfcMappedItem`, removing its `IfcRepresentationMap`, and calling the function: raises the reported `AttributeError`. With the fix it returns identity and does not raise. A well-formed mapped item still computes its real transform.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/validate.py
+++ b/src/ifcopenshell-python/ifcopenshell/validate.py
@@ -398,6 +398,121 @@ def get_entity_attributes(schema: schema_definition, entity: str) -> tuple[entit
     return entity_attrs
 
 
+def validate_instance(
+    inst: ifcopenshell.entity_instance,
+    logger: Union[Logger, json_logger],
+    schema: Optional[schema_definition] = None,
+) -> None:
+    """Validate the attributes and inverse attributes of a single entity instance.
+
+    This runs the same per-instance checks as :func:`validate` (abstract entity,
+    unreadable attribute values, attributes derived in a subtype, non-optional
+    attributes that are `None`, attribute type validity and inverse cardinality),
+    but for one instance only. It does not run the cross-instance checks (GlobalId
+    uniqueness) or the express rules.
+
+    It is factored out so that other tooling, e.g. :func:`diagnose`, can validate
+    a specific suspect instance without iterating the whole file.
+
+    :param inst: The entity instance to validate.
+    :param logger: A standard :class:`logging.Logger` or a :class:`json_logger`.
+    :param schema: The schema definition. Derived from `inst` when not supplied.
+    """
+    if schema is None:
+        schema = ifcopenshell.ifcopenshell_wrapper.schema_by_name(inst.file.schema_identifier)
+
+    entity, attrs = get_entity_attributes(schema, inst.is_a())
+
+    if entity.is_abstract():
+        e = "Entity %s is abstract" % entity.name()
+        if isinstance(logger, json_logger):
+            logger.set_state("attribute", None)
+            logger.error(e)
+        else:
+            logger.error("For instance:\n    %s\n%s", inst, e)
+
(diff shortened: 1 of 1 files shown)


aothms:
These are not optional attributes. I don't think the code should be defensive against every kind of schema violation. Rather we should build something generic:

- Try-except at a rather high level / maybe the root exception handler
- Figure out ifcopenshell.entity_instances in the traceback local variable frames
- Validate these using ifcopenshell.validate
- Somehow provide a report to the user with an estimated confidence of one being the cause of the other

BIMvoice:
You're right, `MappingSource`/`MappingOrigin` are not optional and the code shouldn't be defensive against every schema violation. I've dropped the guard and built the generic mechanism you described instead.

Two additions in `ifcopenshell/validate.py`:

- `validate_instance(inst, logger, schema=None)`: the per-instance checks from `validate()`, factored out so a single suspect can be validated without walking the whole file. `validate()` now calls it, behaviour is unchanged and the existing `test_validate` suite passes.
- `diagnose(exception, ifc_file=None)`: walks the traceback frames, collects the `entity_instance` objects in their locals, validates each with `validate_instance`, and ranks them by likelihood. A schema-invalid entity near the raise is the strongest suspect; confidence is reported by proximity to the failing frame. `str()` of the result is a human-readable report, `.likely_cause` is the top suspect.

On the #7170 file it turns the crash into:

```
IFC processing raised AttributeError: 'NoneType' object has no attribute 'MappingOrigin'
Likely cause (high confidence, frame depth 0): #6=IfcMappedItem($,#5)
    - IfcMappedItem.MappingSource: Attribute not optional
```

I kept it as a standalone utility rather than wiring a root exception handler, since the natural place to catch is in the consumers (Bonsai file load, ifcpatch, ifcconvert) rather than inside `util`. Happy to add the high-level hook wherever you think it belongs, or to move `diagnose` if `validate` isn't the right home. Does this match what you had in mind?

BIMvoice:
Same principle as #8359, you are right. Guarding MappingSource at this one call site is treating a symptom of an invalid file, and there are many such call sites. Closing this.

I am going to prototype the generic approach you described instead: a high level handler that on a load or geometry crash walks the traceback frames, collects the ifcopenshell.entity_instance objects in the local variables, runs ifcopenshell.validate on them, and reports the most likely culprit entity with a confidence estimate. I will put a first version up against the #7170 (removed MappingSource) and #5113 (null DirectionRatios) files so there is something concrete to react to, rather than guessing at the shape you want.

BIMvoice:
First version of the generic mechanism is up as #8378. It walks the traceback, validates the entity_instance objects found in the frames, and names the likely culprit with a confidence estimate. On the #7170 shape it points straight at the IfcMappedItem with a null MappingSource. It is a standalone utility for now, with a note proposing where the high level hook should live. Would appreciate your read on the shape before I take it further.

PR #8328, opened 2026-07-07
bonsai(drawing): persist drawing camera zoom and crop on save (#8205)

BIMvoice:
Part of #8205.

## Problem
On a drawing, set a zoom factor / crop area for the print sheet, Save, then reopen: the zoom and crop reset to the scale default. Reported for georeferenced models but it is not georeferencing specific, it happens on any model.

## Root cause
The drawing camera's bounds (`ortho_scale`, raster, `clip_end`, matrix) are written to the IFC representation only inside `ActivateDrawing`, via `BIMCameraProperties.update_representation`. Saving the project while staying in a drawing never flushed them, so the drawing's `IfcBlock` kept its creation size and the camera rebuilt from it on reload.

## Fix
Flush any loaded drawing camera whose bounds changed at the start of the save operator, using the same `update_representation` check activation already relies on. No-op for cameras that have not changed and for projects without drawing cameras.

## Verification (headless)
Create a plan drawing, set `ortho_scale` to 12.5, save, reopen and activate:
- before: reopens at 50.0 (reset)
- after: reopens at 12.5 (preserved)

Existing `test/core/test_drawing.py` failures (`tool.Ifc has no attribute 'by_type'`) are pre-existing on `v0.8.0` and unrelated to this change.

### DIFF
--- a/src/bonsai/bonsai/bim/module/project/operator.py
+++ b/src/bonsai/bonsai/bim/module/project/operator.py
@@ -2018,6 +2018,12 @@ class ExportIFC(bpy.types.Operator, ExportHelper):
             if not cam_element or not cam_element.is_a("IfcAnnotation"):
                 continue
             cam_props = tool.Drawing.get_camera_props(cam_obj.data)
+            # Skip cameras without an active representation: bim.update_representation
+            # asserts on one (geometry/operator.py) and would otherwise abort the whole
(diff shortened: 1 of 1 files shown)


theoryshaw:
hi, 
a pr on your pr: https://github.com/BIMvoice/IfcOpenShell/pull/3

PR #8332, opened 2026-07-07
ifc5d: measure segment length from geometry when it is not a supported extrusion

BIMvoice:
## Problem
`IfcOpenShell.get_segment_length` (used by the `net_get_segment_length` QTO rule for `IfcFlowSegment` / `IfcCableSegment` etc.) only handles a single `IfcExtrudedAreaSolid`. When a segment is stored as tessellated / Brep geometry, which some authoring tools export (for example Allplan cable/conductor segments), it returns `None`, so **no Length quantity is produced at all** and the QTO silently yields nothing for those elements.

This also produces an inconsistency: the Blender calculator (`get_length`) returns a value for the same element while the ifcopenshell calculator returns nothing.

## Fix
When the representation is not a supported extrusion, fall back to the geometry's **principal-axis extent**, i.e. the true longest dimension of the element measured along its own direction (using the already-tessellated body geometry the iterator provides, so no re-meshing).

This correctly measures straight and diagonally-run segments, round or flat, that the axis-aligned stored quantity under-reports.

Honest limitation, noted in the code: a genuinely **curved/bent** segment stored as a Brep can still under-report, because the mesh no longer carries the original routing curve.

## Verification
On a real Allplan export of grounding conductors (8 `IfcCableSegment` Breps, stored `Length` = 12.895 m, which is only the axis-aligned extent):
- **before:** `net_get_segment_length` returns `None` for all 8 → no Length written.
- **after:** each returns its true length from geometry, e.g. 14.74 m for the element in question, matching BIMvision, the Bonsai Measure tool, and the ifclite viewer (14.74–14.77). The stored 12.9 was the authoring tool's axis-aligned value.

Extruded segments are unaffected (the fallback only runs when it is not a clean extrusion).

### DIFF
--- a/src/ifc5d/ifc5d/qto.py
+++ b/src/ifc5d/ifc5d/qto.py
@@ -35,6 +35,7 @@ import ifcopenshell.util.representation
 import ifcopenshell.util.selector
 import ifcopenshell.util.shape
 import ifcopenshell.util.unit
+import numpy as np
 
 
 class Function(NamedTuple):
@@ -76,13 +77,26 @@ def entity_supertypes(schema_iden: str, entity_name: str):
     return list(visit(decl))
 
 
-def quantify(ifc_file: ifcopenshell.file, elements: set[ifcopenshell.entity_instance], rules: dict) -> ResultsDict:
+def quantify(
+    ifc_file: ifcopenshell.file,
+    elements: set[ifcopenshell.entity_instance],
+    rules: dict,
+    *,
+    estimate_length_from_geometry: bool = False,
+) -> ResultsDict:
     """Quantify elements from a rules using preset quantification rules
 
     Rules placed as a JSON configuration file in the ``ifc5d`` folder will be
     autodetected and loaded with the module for convenience.
 
     :param rules: Set of rules from `ifc5d.qto.rules`.
+    :param estimate_length_from_geometry: Opt-in fallback. When ``True``, a flow
+        segment stored as tessellated / Brep geometry (not a supported
+        extrusion) gets its ``get_segment_length`` estimated from the mesh
+        instead of yielding no Length quantity. Off by default because the
+        estimate is not exact for bent segments or for footprints whose
+        extrusion depth is not the longest dimension, and Qto is meant to stay
+        reliable. Only turn it on when an approximate length is acceptable.
     """
     results: ResultsDict = {}
     elements_by_classes: defaultdict[str, set[ifcopenshell.entity_instance]] = defaultdict(set)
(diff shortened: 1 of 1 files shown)


aothms:
I don't think this is a safe default. PCA here is not bad per se as a fallback, but again silently adding this to Qto which is supposed to be as reliable as possible is just too risky. This doesn't only produce wrong results for bent elements, but also for elements where the depth dimension we're looking for isn't the principal (i.e longest) direction.

I would much rather: Identify the faces top and bottom faces / caps. Find the average arclength over edges that connect these faces. That gives you proper length in case of bent elements and works in case of a non-rectangular footprint extruded over a small length (because then the equal top-bottom faces can only be the actual extrusion basis).

There is basis for this kind of mesh analysis here https://github.com/IfcOpenShell/IfcOpenShell/discussions/7734#discussioncomment-15970370 It would be nice if these kind of dual graph / mesh analysis tools are implemented in C++ on top of the TriangulationElement.

But regardless of the approach this should be a flag that is not enabled by default.

BIMvoice:
Thanks, this is fair and I've reworked it around your two points.

1. It's no longer a default. The geometry estimate is now fully opt-in behind a keyword-only flag on the public entry point, `quantify(..., estimate_length_from_geometry=False)`, threaded down through `calculate` into `get_segment_length`. With the flag off (the default), `get_segment_length` gets no geometry and returns `None` exactly as before this PR, so the shipped rule sets and Qto stay as reliable as they were. The default JSON rule sets are unchanged.

2. On the method: you're right that PCA is wrong not just for bent elements but also when the extrusion depth isn't the principal direction. I've documented both cases in the code as the reason it's opt-in and approximate. I did not swap in the cap-face method here on purpose, doing it well means normal-based face clustering, a face-adjacency dual graph, cap identification, then average arclength over the connecting edges, close to the prototype in discussion #7734, and I don't want to land that half-tested when the only geometry I can verify against is straight segments. So I'd keep this PR as the small, safe, opt-in fallback and do the cap-face method as a dedicated follow-up, ideally the C++ dual-graph / mesh-analysis tooling on `TriangulationElement` you mentioned. Happy to open that follow-up.

Verified on an Allplan grounding-conductor export (8 IfcCableSegment Breps): flag off, no Length is written for any of them, same as before; flag on, each gets its length from geometry, 14.74 m for the element in question, matching BIMvision, the Bonsai Measure tool, and ifclite.

BIMvoice:
Rebased onto current v0.8.0. Conflicts were with the new opening and footing quantity functions in qto.py; resolved as a union keeping upstream's dispatch structure. ifc5d tests show no new failures and the cap face measurement verifies exactly on a synthetic segment.

PR #8356, opened 2026-07-07
Bonsai: store the drawing camera outside the Body context so viewers don't show boxes (#4800)

### LINKED ISSUE #4800: IFC file from blenderbim with drawings shows up with boxes in solibri
When an IFC file containing drawings is opened in Solibri, the views are represented as boxes. Hopefully, this bug can be swiftly fixed as I need to submit IFC files frequently.

Näyttökuva 2024-06-04 152311

BIMvoice:
Fixes #4800.

Drawing cameras (`IfcAnnotation`/`DRAWING`) had their view-volume CSG solid created in `Model/Body/MODEL_VIEW`; IFC viewers like Solibri render Body-context solids, so the drawing "views" show up as boxes over the model.

To answer @Moult's old question: no C++ hardcodes the Body context for the camera. The kernel/serializers are context-driven via settings; the Body coupling was purely Python, one writer (`core/drawing.py`) and two readers.

This moves the camera's crop-volume solid to a dedicated `Model/Clearance/MODEL_VIEW` context. `Clearance` is a non-Body identifier viewers skip, and it's one of the identifiers `add_representation` routes to `create_variable_representation`, so the camera block/pyramid is built and rebuilt by the kernel unchanged, unlike `Box`, which becomes an un-rebuildable `IfcBoundingBox` and breaks the camera round-trip. It also stays out of the drawing SVG serialization (which only iterates Body and Annotation contexts). Readers resolve via a new `get_drawing_camera_representation` that falls back to `Model/Body` so drawings authored by older Bonsai still open.

Verified headless (Blender 5.1.2): new drawings land in Clearance (not Body), reimport/rebuild works and survives save/reopen, and legacy Body cameras still import via the fallback. No C++ change needed.

Scope: existing files aren't migrated (the camera moves to Clearance only when a drawing is recreated); a migration pass could be added if wanted. Happy to swap `Clearance` for a purpose-named context if you'd prefer.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -966,6 +966,17 @@ class CreateDrawing(bpy.types.Operator):
             self.serialiser.setFile(ifc)
             drawing_elements = tool.Drawing.get_drawing_elements(self.camera_element, ifc_file=ifc)
 
+            # The camera defines the drawing's view crop, it is not a drawn
+            # element. Keep it out of the context-specific linework passes below
+            # and instead always feed it to the serialiser through the
+            # context-agnostic iterator further down, so the crop/frustum is
+            # built regardless of which representation context the camera lives
+            # in. Otherwise a camera stored outside the Body context (see #4800)
+            # would never reach the serialiser and the whole drawing would come
+            # out empty.
+            if tool.Ifc.get() == ifc:
+                drawing_elements.discard(self.camera_element)
+
             if self.cprops.fill_mode == "SHAPELY":
                 for element in drawing_elements.copy():
                     if element.is_a("IfcAnnotation"):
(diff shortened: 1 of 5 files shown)


aothms:
> To answer @Moult's old question: no C++ hardcodes the Body context for the camera. The kernel/serializers are context-driven via settings; the Body coupling was purely Python, one writer (core/drawing.py) and two readers.

No there is, implictly. The SVG serializer does expect the geometry for the camera to be iterated over so that it can construct the camera frustrum from the box:

https://github.com/IfcOpenShell/IfcOpenShell/blob/b2d58d0b812f93b64c5b56bd3fd67057138d51d0/src/serializers/SvgSerializer.cpp#L631

BIMvoice:
You're right, thank you, I'd missed that. The serialiser itself is context-agnostic (it matches the camera by GlobalId and builds the crop from its box), but Bonsai only fed the camera to it through the body-context linework pass: the camera is part of the drawing group so `get_drawing_elements` returns it, which meant the context-agnostic force-include in `generate_linework` was always skipped. Moving the camera out of Body therefore starved the serialiser and produced empty drawings, exactly the implicit dependency you pointed at.

Fixed by treating the camera as the view definition it is: it's now excluded from the context-specific passes and always fed to the serialiser via the no-context-filter iterator, so the crop is built regardless of context. No C++ change needed. Verified end-to-end through the real `create_drawing` (the C++ HLR serialiser, not just the import path): output is byte-identical for cameras still in Body, and drawings generate correctly for the camera stored in `Model/Clearance` (72 paths, 500×500mm, same as the Body baseline). Updated the PR.

PR #8359, opened 2026-07-07
util.placement: tolerate null direction ratios in get_axis2placement (#5113)

BIMvoice:
Related to #5113.

While investigating #5113 (files with null `IfcAxis2Placement3D` components) I found that on current 0.8.x the exact `IFCAXIS2PLACEMENT3D($,$,$)` from that issue already loads without error, the C++ kernel tolerates the fully-null placement now. But a closely related null-placement case still crashes in the Python util: when an `Axis` or `RefDirection` entity exists but its `DirectionRatios` attribute is null (`IFCDIRECTION()`), `get_axis2placement` throws an opaque `TypeError: unsupported operand type(s) for *: 'NoneType' and 'NoneType'`, because the guard only checked whether the direction entity was present, not whether it carried ratios, so `np.array(None)` flowed into the matrix math. Affects `IfcAxis2Placement3D`, `IfcAxis2Placement2D`, and `IfcAxis1Placement`.

This falls back to the schema default axes (Z = 0,0,1, X = 1,0,0) when the ratios are absent, matching the existing handling of a missing `Axis`/`RefDirection`. Adds regression tests for the null-placement family (`test/util/test_placement.py`, 8 pass). No fabricated transform, pure schema defaults.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/placement.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/placement.py
@@ -62,8 +62,8 @@ def get_axis2placement(placement: ifcopenshell.entity_instance) -> MatrixType:
     """
     ifc_class = placement.is_a()
     if ifc_class in ("IfcAxis2Placement3D", "IfcAxis2PlacementLinear"):
-        z = np.array(placement.Axis.DirectionRatios if placement.Axis else (0, 0, 1))
-        x = np.array(placement.RefDirection.DirectionRatios if placement.RefDirection else (1, 0, 0))
+        z = np.array((placement.Axis and placement.Axis.DirectionRatios) or (0, 0, 1))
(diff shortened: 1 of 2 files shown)


aothms:
We already discussed this in #8321 let's not be more permissive then the schema dictates.

BIMvoice:
Fair enough, agreed. DirectionRatios is a required attribute, so a per-site fallback is the wrong layer and makes the util quietly tolerate an invalid file. Closing this.

The generic direction you sketched in #8321 is the better one: catch at a high level, pull the ifcopenshell.entity_instance objects out of the traceback frames, run ifcopenshell.validate on them, and report the likely culprit with a confidence estimate. That would cover this whole family of null-attribute crashes in one place instead of guarding each call site. If there is appetite for it I am happy to take a run at a first version.

PR #8360, opened 2026-07-07
util.placement: fix double unit scaling of linear placement fallback position

BIMvoice:
`get_axis2placement()` evaluated `IfcAxis2PlacementLinear` locations (`IfcPointByDistanceExpression`) through the geometry engine with `convert-back-units` enabled. For files whose length unit is not metres, that setting double-applies the unit scale, so the returned `CartesianPosition` coordinates came out scaled by `1/unit_scale` (e.g. `3.28084` for a file in feet). Metre-unit files hid the bug because the scale is 1.

The engine now runs with default SI settings and the translation is converted back to the file's length unit explicitly via `calculate_unit_scale()`. Direction ratios are unitless and unaffected. Metre-unit files are unchanged.

Fixes the persistently-failing `test/api/alignment/test_update_fallback_position.py` (a foot-unit file produced `(12405.06, 8737.81, 0.0)` instead of `(3781.06, 2663.29, 0.0)`), which currently fails on every PR's CI. Full alignment suite green (26 passed); `test/util/test_placement.py` green. Black/Ruff/ty clean.

### DIFF
--- a/src/ifcwrap/IfcGeomWrapper.i
+++ b/src/ifcwrap/IfcGeomWrapper.i
@@ -977,12 +977,9 @@ struct ShapeRTTI : public boost::static_visitor<PyObject*>
 			if (item == nullptr) {
 				throw IfcParse::IfcException("Failed to convert placement");
 			}
-            if (st.get<ifcopenshell::geometry::settings::ConvertBackUnits>().get()) {
-                // we pass the settings to the Transformation object, but access the data just offloads to the
-                // generic cartesian_base<Matrix4> so there's no time to apply the settings to the translation part.
-                item = ifcopenshell::geometry::taxonomy::matrix4::ptr(item->clone_());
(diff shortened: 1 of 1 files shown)


aothms:
> that setting double-applies the unit scale

Then please fix it there

BIMvoice:
Reworked per your review. The double unit scaling is now fixed at the source: create_shape in IfcGeomWrapper.i no longer divides the placement translation by the length unit, so IfcGeom::Transformation applies it exactly once, and the earlier workaround in ifcopenshell.util.placement is gone. The branch is rebased on current v0.8.0, so it also carries the PlacementRelTo fallback fix. Verified on a foot unit alignment: the engine SI translation converts once to (3781.06, 2663.29), the expected fallback position, and a parent PlacementRelTo still composes on top.

BIMvoice:
Closing: superseded by 52d894298e (RickBrice, merged 2026-07-10), which removes the same double-unit-scaling block this PR targeted.

PR #8361, opened 2026-07-07
Bonsai: scope schedule/document styles on a sheet so they keep their own font (#5221)

### LINKED ISSUE #5221: When a schedule is built, it has the correct font, but when it's placed on a sheet, it takes on the default font
When a schedule is built, it has the correct font...

but when it's placed on a sheet, it takes on the default font

BIMvoice:
Fixes #5221.

Drawings on a sheet run their embedded CSS through `ensure_drawing_unique_styles`, isolating styles per reference. Schedules/references in `build_documents` were appended **unscoped**, so a schedule's global `text, tspan { font-family: ... }` rule collided with other documents on the sheet, and with two or more documents the last one in document order won for every schedule, so a schedule rendered in the wrong font (the reported symptom; theoryshaw's `!important` was forcing the rule to win a cascade it was otherwise losing).

This scopes documents like drawings (`text.dNN, tspan.dNN { ... }`), a clean specificity/scoping fix, not `!important`. `ensure_drawing_unique_styles` now also tolerates an embedded SVG with no `<style>` (plain references) instead of asserting one is present.

Verified with two schedules (distinct fonts) on one sheet: before, both rules are global and the later wins for both; after, each is scoped and keeps its own font, with no unscoped `text` rule remaining. Drawing scoping output is unchanged.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/sheeter.py
+++ b/src/bonsai/bonsai/bim/module/drawing/sheeter.py
@@ -335,49 +335,51 @@ class SheetBuilder:
         """
         prefix = f"d{drawing_id}"  # just number doesn't work
 
-        # add .prefix class to all css selectors
-        style = svg.find(f"{SVG}defs/{SVG}style")
-        assert style is not None
-        style_data = style.text
-        assert style_data is not None
-        text = ""
-        brackets_level = 0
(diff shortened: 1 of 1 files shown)


BIMvoice:

PR #8363, opened 2026-07-07
Bonsai: green the ty type-check for src/bonsai (fixes a latent missing-import bug)

BIMvoice:
Greens the `ty` type-check step for `src/bonsai` (`poe ty` -> `ty-bonsai` now prints "All checks passed!"). The 17 diagnostics fell into three buckets:

**Real fixes (latent bugs / wrong references):**
- `model/__init__.py` never imported the `decorator` submodule it uses in `unregister()`, a `NameError` waiting to happen on addon disable. Added the import.
- Added missing imports actually used at runtime: `bonsai.core.geometry` (mep.py), `bonsai.core.model` (tool/model.py), `bmesh` (gizmos.py, used in an annotation), and `ifcopenshell.api.pset` / `util.element` in three tests.
- `railing.py` referenced `"BIMRailingProperties"` where only `prop` is imported -> `"prop.BIMRailingProperties"`.

**Tooling false positives (narrow `# ty: ignore` with notes, no logic changed):**
- The `Surveyor.*` "missing-argument" and the `get_geom_generation` "empty-body" diagnostics stem from the `@interface` decorator building classmethods dynamically via `type()`, which ty can't follow.
- `gizmo_textures` is a module not present in the static tree.

Scope: this greens the bonsai half of `ty`. The `ty-ios` half (`ifcopenshell-python`, mostly the WIP `alignment` code) has its own ~17 pre-existing diagnostics, including a genuine bug (`util/alignment.py` passes an `include_referent` kwarg to a function that doesn't accept it), and `poe ruff` is independently red there. Those are left for the maintainers, as they touch unfinished code; this PR is the clean, self-contained bonsai portion.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/gizmos.py
+++ b/src/bonsai/bonsai/bim/module/drawing/gizmos.py
@@ -85,6 +85,7 @@ from enum import Enum
 from typing import Any, ClassVar, Literal, Optional, Protocol, runtime_checkable
 
 import blf
+import bmesh
 import bpy
 import gpu
 import ifcopenshell.util.element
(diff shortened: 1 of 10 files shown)


BIMvoice:
Superseded by #8364, which contains this bonsai ty cleanup plus the ios ty subtask and the ruff cleanup (greening ty and ruff together). Closing this in favour of #8364.

PR #8362, opened 2026-07-07
Format repository with current Black to green the lint-formatting CI check

BIMvoice:
Per @aothms's invite to fix the drifted formatting. The `lint-formatting` job runs Black over the whole repo with `psf/black@stable`, and the repo had drifted from the current stable Black (25.11.0), so ~27 pre-existing files were flagged "would reformat" on every PR, making the check red for everyone.

- **Ruff commit:** `ruff check --fix` (safe fixes only) removes 26 pre-existing unused imports (F401) in the alignment API and one test.
- **Black commit:** `black .` reformats the 27 drifted files (line-length 120, repo `pyproject.toml` config).

Formatting + dead-import removal only, no behavioural changes. `black --check .` and `poe ruff` now pass.

Note: the separate `ty` step (in `ci-lint.yaml`) is still red on ~17 pre-existing type diagnostics; those need human judgement and are out of scope here, left for a maintainer call.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8364, opened 2026-07-07
Green the ci-lint ty and ruff checks (src/bonsai + ifcopenshell-python)

BIMvoice:
## What this does

Greens the `ci-lint / lint-formatting` job's **`ty` type-check** and **`ruff`** steps, which are both currently red on `v0.8.0`.

A note on why the ty problem was bigger than it looked: `poe ty` runs `ty-bonsai` and then `ty-ios`, and poe aborts on the first failing subtask. So while `ty-bonsai` was failing, `ty-ios` never ran and its diagnostics were invisible. Fixing `src/bonsai` (17 diagnostics) unmasks 17 more in `src/ifcopenshell-python`. Both are addressed here.

After this PR: `poe ty` (bonsai + ios) passes, and `poe ruff` passes repo-wide. **Black is not touched here** — it is red for a separate set of ~16 files handled by #8362. Both PRs are needed for the job to go fully green.

## Changes, grouped honestly

**Safe missing imports** (code relied on transitive imports, or a name was outright undefined):
- `bonsai/bim/module/model/mep.py`: `import bonsai.core.geometry`
- `bonsai/tool/model.py`: `import bonsai.core.model`
- `bonsai/bim/module/drawing/gizmos.py`: `import bmesh` (used only in an annotation)
- three `src/bonsai` test files: `ifcopenshell.api.pset` / `ifcopenshell.util.element`
- `api/alignment/_get_segment_endpoint.py`: `import ifcopenshell.geom`
- `api/alignment/update_end_point.py`: `import ifcopenshell.api.alignment`
- `test/api/alignment/test_create_representation.py`: `ifcopenshell.api.{aggregate,context,spatial}` and `ifcopenshell.util.unit`

**Annotation fix**
- `bonsai/bim/module/model/railing.py`: `"BIMRailingProperties"` → `"prop.BIMRailingProperties"` (only `prop` is imported in that module).

**Ruff `unused-import` cleanup** (all in the WIP `api/alignment` code): removed leftover unused imports via `ruff check --fix`. No behaviour change.

**Narrow `# ty: ignore` for genuine tooling false positives** (no rule disabled globally, no runtime behaviour changed):
- `core/product.py`, `core/tool.py`: the `@interface` decorator in `core/tool.py` builds classmethods dynamically via `type()`, which ty cannot follow. It reports false `missing-argument` on `Surveyor.*` calls and a false `empty-body` on `get_geom_generation`. Suppressed with notes.
- `bonsai/bim/module/drawing/gizmos.py`: `gizmo_textures` is provided at runtime and is not in the static tree.
- `ifcopenshell/__init__.py`, `geom/main.py`: calls into the compiled `ifcopenshell_wrapper` pass an optional trailing `logger`, but the hand-maintained `ifcopenshell_wrapper.pyi` stub omits it, so ty reports `too-many-positional-arguments`. Suppressed at the call sites, matching the `# ty: ignore` pattern already present in `__init__.py`. **The real fix is to add the optional `logger` parameter to those stub signatures** — left to the maintainers since the stub is validated by `util/scripts/validate_stub.py`. Also removed one now-unused `# ty: ignore[unknown-argument]`.
- `api/cost/assign_cost_item_quantity.py`: `OPERATORS` maps binary ast ops plus the unary `ast.USub` (`operator.neg`); only binary ops reach the 2-arg call in `visit_BinOp`, so ty's union-based arity check is a false positive. Suppressed.

## Two real bugs found and fixed — please sanity-check

Both are runtime crashes on reachable paths; each fix is the single unambiguous correction, not a guess.

1. **`bonsai/bim/module/model/__init__.py`** used `decorator.MEPSystemPathDecorator.uninstall()` / `WallSystemPathDecorator.uninstall()` in `unregister()`, but never imported the `decorator` submodule → `NameError` on addon disable / Blender shutdown. Added the missing sibling import.
2. **`ifcopenshell/util/alignment.py`** called `add_zero_length_segment(patched_file, layout, include_referent=False)`, but `add_zero_length_segment(file, layout)` has no such parameter → `TypeError`, reachable through the `AddZeroLengthSegmentToAlignment` ifcpatch recipe. Removed the bogus keyword to match the correct sibling call on the next line. If `include_referent` was actually intended as an API parameter, the correct fix is yours to make in the API instead.

## Verification

- `poe ty` → passes (bonsai + ios)
- `poe ruff` → passes repo-wide
- `black` → clean on every file this PR touches (`assign_cost_item_quantity.py` was already black-dirty and is reformatted here; that one file overlaps #8362)
- `python -m compileall` → clean on all changed `ifcopenshell-python` files

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/gizmos.py
+++ b/src/bonsai/bonsai/bim/module/drawing/gizmos.py
@@ -85,6 +85,7 @@ from enum import Enum
 from typing import Any, ClassVar, Literal, Optional, Protocol, runtime_checkable
 
 import blf
+import bmesh
 import bpy
 import gpu
 import ifcopenshell.util.element
@@ -2035,7 +2036,9 @@ class TexturedQuadGizmoMixin(StaticTrisGizmoMixin):
 
     def setup(self) -> None:
         super().setup()
-        from bonsai.bim.module.drawing import gizmo_textures
+        from bonsai.bim.module.drawing import (
+            gizmo_textures,  # ty: ignore[unresolved-import]
+        )
 
         self._quad_batch = batch_for_shader(
             gizmo_textures.get_shader(),
@@ -2044,7 +2047,9 @@ class TexturedQuadGizmoMixin(StaticTrisGizmoMixin):
         )
 
     def draw(self, context: bpy.types.Context) -> None:
-        from bonsai.bim.module.drawing import gizmo_textures
+        from bonsai.bim.module.drawing import (
+            gizmo_textures,  # ty: ignore[unresolved-import]
+        )
 
         texture = gizmo_textures.get_icon_texture(self.icon_name)
         if texture is None:
--- a/src/bonsai/bonsai/bim/module/model/__init__.py
+++ b/src/bonsai/bonsai/bim/module/model/__init__.py
@@ -27,6 +27,7 @@ import bonsai.tool as tool
 from . import (
     array,
     covering,
+    decorator,
     door,
     external,
     grid,
--- a/src/bonsai/bonsai/bim/module/model/mep.py
+++ b/src/bonsai/bonsai/bim/module/model/mep.py
@@ -38,6 +38,7 @@ import numpy as np
 from ifcopenshell.util.shape_builder import ShapeBuilder
 from mathutils import Matrix, Vector
 
+import bonsai.core.geometry
 import bonsai.core.root
 import bonsai.tool as tool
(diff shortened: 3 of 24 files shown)


BIMvoice:
Superseded by https://github.com/IfcOpenShell/IfcOpenShell/pull/8367, which combines this ty/Ruff cleanup with the Black reformat so the whole lint-formatting job passes in one PR.

Issue #4783, opened 2024-06-01
[BBIM] IfcFooting | Extruded profile elements orientation


steverugi:
When creating an IfcFooting the are two predefined types that look similar (FOOTING_BEAM and STRIP_FOOTING) and work in real life horizontally, or _beam-like_.  

Instead they are created vertically, or _column-like_.

I mainly see two cases:

1. PAD_FOOTING, and PILE_CAP, a `horizontal layer` does the job, they both "behave" like a thick slab  
2. STRIP_FOOTING and FOOTING_BEAM, an `extruded profile` is used instead, since they both are, and need to be represented like, a "beam"

It may be worth noting that, depending on the cases, the QTO set needs to update accordingly, not so important but somehow noticeable, related to the above cases:
1. `Length` and `Width` represent the horizontal axis, `Height` the vertical (extruded) one, or the thickness of the element
2. `Length` needs to represent, like with beams and columns, the _extruded_ value of a chosen profile

I hope the above makes sense, thanks

dapenamr:
@steverugi , thanks for mentioning this bug in your contribution on community.osarch.org [(Post from 20.09.2024, https://community.osarch.org/discussion/2511/bonsai-ifcopenshell-v-bonsai-qto-engine-comparison/p1)].
I have tried to register in the osarch.org community to be able to contribute and ask questions about the topic you are currently working on (QTO) but I still do not have access... however, there is always a way ; )

In my case, the model was not made in Blender+Bonsai, but from another software.
I concluded that the IfcOpenShell calculator is used for these cases and when modeling directly in Blender+Bonsai the Blender calculator is used. I would like this to be confirmed by @Moult .

Taking as an example the thin slab (concrete subbase) of the attached retaining wall foundation, the manual quantities obtained individually with the IfcOpenShell calculator do not correspond to the real ones. I have tested with each LENGTH option.

I think this is due to the orientation of the objects. Maybe the IfcOpenShell calculator is taking the orientation of the “move select items” vector and not the object itself? (see picture attached).

![image](https://github.com/user-attachments/assets/b2075fbc-f301-4853-aee2-93b28e409f0f)

@Moult , I attach the IFC file for your evaluation, only for the purpose of documenting the bug and looking for solutions. Please be discreet with this model as it corresponds to an ongoing project.

[Regelblock-IFC4.zip](https://github.com/user-attachments/files/17095009/Regelblock-IFC4.zip)

Best regards and I keep in touch.

[...]

PR #8365, opened 2026-07-07
Fix IfcFooting Qto_FootingBaseQuantities axis mapping (#4783)

### LINKED ISSUE #5507: IfcOpenShell Quantity calculation issues
On creating an IfcFooting[STRIP_FOOTING] or IfcFooting[FOOTING_BEAM] Bonsai models a 'column-like' element (vertical extrusion) but it should be more  a 'beam-like' one (horizontal extrusion)

image

It can be rotated to its expected position but when quantity take-off is run using IfcOpenShell engine it produces some odd results:

image

the Volume is OK since it's the product of the dimensions on the three axis but Length and Height are wrong (I suspect because they come from the original orientation the element started with)

In normal practice Height should be 0.35, Width 0.20 (not calculated), and Length 3.00m

thanks

BIMvoice:
Fixes the QTO half of #4783 (the orientation was already correct on v0.8.0).

Qto_FootingBaseQuantities was computed wrongly for extruded-profile
footings (STRIP_FOOTING / FOOTING_BEAM), which Bonsai authors as a
profile swept along the representation-local Z axis:

- Length was net_get_max_xy, so it grabbed a profile cross-section
  dimension instead of the extruded (beam) axis.
- Height was net_get_z, so it grabbed the extruded axis length instead
  of the profile height.
- Width was null and never reported at all, even though it is a member
  of Qto_FootingBaseQuantities.

The ifc5d IfcOpenShell Qto engine evaluates geometry in
representation-local coordinates (ObjectPlacement is ignored), so for a
profile-extruded footing the local axes are: Z = extruded/beam axis,
X = profile width, Y = profile height. Map accordingly, matching the
IFC schema semantics of Qto_FootingBaseQuantities (Length measured along
the path for strip footings, Width perpendicular to the path, Height the
nominal vertical height):

- Length -> net_get_z
- Width  -> net_get_x
- Height -> net_get_y

Applied to both IFC4 and IFC4X3 IfcOpenShell rule sets.

Blender Qto engine: Width used the generic get_width (min of X/Y), which
for a beam-like footing collapses onto the height. Add get_footing_width
(predefined-type aware, mirroring get_footing_length/get_footing_height)
and wire it into the IFC4 and IFC4X3 Blender rule sets. Also align the
IFC4X3 Blender footing rules to the predefined-type-aware footing
functions so both schemas behave consistently.

Verified headless against the ifc5d IfcOpenShell engine: a 0.6 x 0.4
profile extruded 2.5 m now reports Length 2.5, Width 0.6, Height 0.4,
and ifcopenshell.validate reports no schema issues on the written
IfcElementQuantity for both IFC4 and IFC4X3.

Corrected for both the IfcOpenShell and Blender Qto engines, IFC4 and IFC4X3. Schema-adherent against Qto_FootingBaseQuantities (verified: Length 2.5 / Width 0.6 / Height 0.4, ifcopenshell.validate reports 0 issues). black + ruff clean.

Also fixes #5507, which reports the same Qto axis mapping defect for IfcFooting STRIP_FOOTING / FOOTING_BEAM (odd Length/Width/Height while Volume stays correct).

### DIFF
--- a/src/bonsai/bonsai/bim/module/qto/calculator.py
+++ b/src/bonsai/bonsai/bim/module/qto/calculator.py
@@ -104,7 +104,7 @@ def get_footing_length(o: bpy.types.Object) -> float:
         return get_length(o)
     if predefined_type == "FOOTING_BEAM" or predefined_type == "STRIP_FOOTING":
         return get_z(o)
-    elif predefined_type == "PAD_FOOTING":
+    elif predefined_type == "PAD_FOOTING" or predefined_type == "PILE_CAP":
         return max(get_x(o), get_y(o))
     else:
         return get_length(o)
@@ -197,10 +197,24 @@ def get_footing_height(o: bpy.types.Object) -> float:
         return get_height(o)
     if predefined_type == "FOOTING_BEAM" or predefined_type == "STRIP_FOOTING":
         return get_y(o)
-    elif predefined_type == "PAD_FOOTING":
+    elif predefined_type == "PAD_FOOTING" or predefined_type == "PILE_CAP":
         return get_z(o)
     else:
         return get_height(o)
+
+
+def get_footing_width(o: bpy.types.Object) -> float:
+    element = tool.Ifc.get_entity(o)
+    assert element
+    predefined_type = ifcopenshell.util.element.get_predefined_type(element)
+    if not predefined_type:
+        return get_width(o)
+    if predefined_type == "FOOTING_BEAM" or predefined_type == "STRIP_FOOTING":
+        return get_x(o)
+    elif predefined_type == "PAD_FOOTING" or predefined_type == "PILE_CAP":
+        return get_width(o)
+    else:
+        return get_width(o)
 
 
 def get_height(o: bpy.types.Object) -> float:
--- a/src/ifc5d/ifc5d/IFC4QtoBaseQuantities.json
+++ b/src/ifc5d/ifc5d/IFC4QtoBaseQuantities.json
@@ -318,12 +318,12 @@
                     "GrossSurfaceArea": null,
                     "GrossVolume": null,
                     "GrossWeight": "gross_get_weight",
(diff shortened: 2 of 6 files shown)


BIMvoice:
Force-pushed a corrected mapping. The earlier version swapped Length and Height for beam-like footings (STRIP_FOOTING / FOOTING_BEAM), caught in live testing.

The corrected fix is predefined-type-aware and now verified three ways:
- Spec: Qto_FootingBaseQuantities defines Length as "measured along the path" for strip footings and Width "perpendicular to the footing path (or longitudinal axis)", with Height the total nominal (vertical) height.
- Driven against the real Bonsai generators (DumbProfileGenerator / DumbSlabGenerator) for beam-like and pad footings, on both engines and both schemas.
- Decisive non-square orientation check: a strip footing with a 0.3 wide x 0.6 tall cross section and a 6.0 run gives Length 6.0, Width 0.3, Height 0.6, and the world bounding box confirms the 0.3 is physically horizontal (perpendicular to the run) and the 0.6 is physically vertical. So Width and Height are not swapped.

Beam-like (STRIP_FOOTING, FOOTING_BEAM): Length = net_get_z (the run), Width = net_get_x (horizontal), Height = net_get_y (vertical). Slab-like (PAD_FOOTING, PILE_CAP): Length/Width = the footprint dimensions, Height = net_get_z (thickness). Both engines and both schemas; ifcopenshell.validate clean.

maxfb87:
@BIMvoice could you resolve the conflict before proceed with merge? Thanks

BIMvoice:
@maxfb87 conflicts resolved: rebased onto current v0.8.0 including the merged opening-orientation change (#8834), footing values re-verified for all three predefined-type cases (STRIP_FOOTING, FOOTING_BEAM, PAD_FOOTING), no new test failures. Ready for merge, thanks!

PR #8368, opened 2026-07-07
cgal: deflection-based floor on conic arc tessellation so large-radius profile arcs aren't straightened (#8051)

### LINKED ISSUE #8051: Incorrect geometry with cgal-simple vs opencascade
### Bug Description

Coming out of the benchmarking tool (for now, manual bug reports whilst I get a feel for how to separate noise, duplicates, from actionable things), I see two very different results for opencascade vs cgal-simple.

Use https://cloud.thinkmoult.com/models/ifclite/dental_clinic.ifc and look at these IDs:

```
Surface area changed (top 20):
  dental_clinic.ifc  express_id=2395  ifc_type=IfcMember  baseline_surface_area=0.42241815899942176  subject_surface_area=0.41883835443507295
  dental_clinic.ifc  express_id=2396  ifc_type=IfcMember  baseline_surface_area=0.42241815899945107  subject_surface_area=0.41883835443510753
  dental_clinic.ifc  express_id=2397  ifc_type=IfcMember  baseline_surface_area=0.754631863770941  subject_surface_area=0.7502858212140717
```

Here's one example - the curved one is from OCC and the straight one is from cgal-simple:

Image

Note: this is not a regression, I can reproduce this in 0.8.5.

BIMvoice:
Fixes #8051.

The CGAL kernel disagrees with OpenCASCADE because it **under-tessellates profile arcs**, not because of any swept/directrix difference. The `IfcMember`s (2395/2396/2397) are `IfcExtrudedAreaSolid`s whose profile OuterCurve is an `IfcCompositeCurve` with `IfcTrimmedCurve` arcs on ~6.5 m-radius `IfcCircle`s spanning only ~9°.

In `src/ifcgeom/kernels/cgal/CgalKernel.cpp` `evaluate_conic`, the arc segment count is `ceil(span / 2π · circle-segments)` — a fraction of the **full** circle, and **radius-agnostic**. So a 9° arc on a 6.5 m radius gets `ceil(0.16/2π · 16) = 1` segment: a single chord (~20 mm sagitta) that goes straight. OpenCASCADE meshes by deflection, so it stays curved. This affects the **full `cgal` kernel too**, not just `cgal-simple` (both share this path). Raising `circle-segments` converges CGAL back onto the OCC areas, confirming tessellation is the sole cause.

Fix: add a deflection-based floor to the segment count in `evaluate_conic` — the max of the `circle-segments` count and the count needed to keep chord deviation within `mesher-linear-deflection` (default 1 mm). Small circles are unchanged (the circle-segments count still dominates); only large-radius curves densify, matching OCC.

**Verification (honest):** a full CGAL kernel build was not feasible in my environment (Homebrew ships CGAL 6.2; this v0.8.0 targets CGAL 5.x, and the TU also needs SWIG/schema-generated sources). So this is verified by: (1) the edited logic compiles clean under `-Wall` in an isolated harness mirroring the real `taxonomy::circle/ellipse` types + settings; (2) it yields 5 segments/arc (was 1) for these mullions and leaves a small R=0.05 circle at 16 (no regression); (3) via the `circle-segments` knob (the identical code path), 5 segments/arc gives area 0.42235 (0.017% vs OCC) and 0.75461 (0.003%). CI / a maintainer with a CGAL build does the final end-to-end verification.

Tradeoff to flag: large-radius curves now tessellate finer to honor the deflection setting, which raises vertex counts for big arcs and could affect CGAL boolean cost; a cap could be added if that's a concern (I couldn't measure it without a build).

### DIFF
--- a/src/ifcgeom/ConversionSettings.h
+++ b/src/ifcgeom/ConversionSettings.h
@@ -361,8 +361,8 @@ namespace ifcopenshell {
 
 			struct CircleSegments : public SettingBase<CircleSegments, int> {
 				static constexpr const char* const name = "circle-segments";
-				static constexpr const char* const description = "Number of segments to approximate full circles in CGAL kernel.";
-				static constexpr int defaultvalue = 16;
+				static constexpr const char* const description = "Number of segments to approximate full circles in the CGAL kernel. When 0 (the default) the segment count is derived from mesher-linear-deflection instead, so curves stay within the deflection tolerance regardless of radius.";
+				static constexpr int defaultvalue = 0;
 			};
 
 			struct CgalSmoothAngleDegrees : public SettingBase<CgalSmoothAngleDegrees, double> {
--- a/src/ifcgeom/kernels/cgal/CgalKernel.cpp
+++ b/src/ifcgeom/kernels/cgal/CgalKernel.cpp
@@ -391,6 +391,11 @@ namespace {
 		}
 	};
 
+	// Representative radius used to size the polygonal approximation of a conic.
+	// For an ellipse the larger semi-axis is the conservative choice.
+	inline double conic_radius(const taxonomy::circle::ptr& c) { return c->radius; }
+	inline double conic_radius(const taxonomy::ellipse::ptr& e) { return e->radius > e->radius2 ? e->radius : e->radius2; }
+
 	struct cgal_curve_creation_visitor {
 		Settings& settings_;
 		parameter_range param;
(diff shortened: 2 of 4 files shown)


aothms:
Please make it one or the other. When CircleSegments == 0 -> tesselate based on MesherLinearDeflection otherwise use CircleSegments. Make CircleSegments = 0 the new default and update the setting docs.

BIMvoice:
Reworked as requested. It is now one or the other, no floor on top of a fixed count.

When circle-segments is 0 the CGAL kernel sizes the conic tessellation from mesher-linear-deflection, so the chord stays within the deflection tolerance at any radius. When circle-segments is non zero it is used directly as a fixed segment count for a full circle. I made 0 the new default and updated the setting description plus the ifcconvert and geometry-settings docs.

circle-segments is only read by the CGAL kernel. OpenCascade meshes by deflection and never reads it, so the new default changes nothing on the OCC side. I verified the segment-count logic for both branches: the #8051 large radius arc now tessellates smoothly under the default and a fixed count still yields exactly that many segments. I did not run a full CGAL geometry build locally.

PR #8370, opened 2026-07-07
ifcgeom: clone mapped items before in place mutation to fix nondeterministic multithreaded geometry (#8050)

### LINKED ISSUE #8050: Nondeterministic geometry output with multithreading
### Bug Description

This came about after finding strange results from benchmarking. First, the good news:

 - 0.8.5 from PyPI vs 0.8.6-94f4dbc (2026-05-10T17:14:28.000Z) gives identical output from opencascade, cgal, and cgal-simple. Hooray! No regressions!
 - This bug seems to exist in both 0.8.5 and 0.8.6-94f4dbc, so it's not a regression
 - It is somewhat rare it seems

To reproduce https://cloud.thinkmoult.com/models/ifclite/dental_clinic.ifc and use https://github.com/IfcOpenShell/ifcbench/blob/master/scripts/test_multi.py , which will run the iterator N times and see if there any differences between them.

```
python test_multi.py --model models/dental_clinic.ifc --kernel opencascade --runs 10 --threads 12
```

I've tested with opencascade, cgal, and cgal-simple. All three reliably produce either a missing (i.e. not yielded) STEP ID, or different geometry (different n_verts). The STEP ID 2752 and 2735 seem to come up quite commonly. Changing threads to 1 causes the issue to go away, so this suggests that it's not kernel-specific, but is a general multithreading thing.

Example output:

```
...
changed_id=2752  baseline=(verts=8, tris=12, hash=2a8dd1ea4d5a445cdcd69568c65c3f24)  run=(verts=12, tris=20, hash=6fcc089a48cd1e502cf280d03994aefc)
...
missing_ids=[2735]
...
```

A neat thing with the ifcbench tool is that it has a convenient helper to run a script against a particular environment BTW, e.g.:

```
ifcbench execute-py ifcopenshell-python-313-20260510-0.8.6-94f4dbc-linux64 scripts/test_multi.py --model models/dental_clinic.ifc --kernel cgal-simple --runs 10 --threads 12
```

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Fixes #8050.

The geometry iterator produced nondeterministic output under multithreading (elements intermittently missing or with different vertex counts between runs; single threaded is fine).

## Root cause

The mapping cache returns the same shared taxonomy item on a cache hit (`mapping::map()` returns the identical `shared_ptr`), and several `map_impl` handlers mutate a `map()` result in place without cloning first. That shared item is reused across the representations a worker processes, and across workers, so the in place write corrupts it. Because task distribution to workers is scheduler dependent, the corruption varies run to run.

## Fix

Clone before mutating at every site that writes into a `map()` result, matching the idiom already used in `IfcOrientedEdge`, `IfcDerivedProfileDef` and `IfcArbitraryOpenProfileDef`:

- `IfcFace.cpp`: clone the mapped loop before `reverse()` and setting `external`.
- `IfcCompositeCurve.cpp`: clone the mapped ParentCurve before `reverse()`.
- `IfcEdge.cpp`: clone the mapped loop before `calculate_linear_edge_curves()` writes into its child edges.

A full audit of every `map()` result mutation in `src/ifcgeom/mapping` is included; the sites that only read or store the result are left unchanged. Because `clone_()` deep clones children, the mutations that reach into child edges are also contained.

This fixes the race at the mutation site and keeps the mapping cache enabled, so the earlier approach of disabling the worker cache is no longer needed.

## Verification

Correctness is argued by equivalence with the existing clone before mutate idiom: after the change no cached shared item is ever mutated, so no two readers can observe each other's writes regardless of thread scheduling. A multithreaded before and after on a compiled kernel was not re run in this iteration, so I am happy to run the dental_clinic comparison again on a build if you want the empirical confirmation as well.

### DIFF
--- a/src/ifcgeom/mapping/IfcCompositeCurve.cpp
+++ b/src/ifcgeom/mapping/IfcCompositeCurve.cpp
@@ -53,6 +53,9 @@ taxonomy::ptr mapping::map_impl(const IfcSchema::IfcCompositeCurve* inst) {
 			auto crv = map(segment->as<IfcSchema::IfcCompositeCurveSegment>()->ParentCurve());
 			if (crv) {
 				if (!segment->as<IfcSchema::IfcCompositeCurveSegment>()->SameSense()) {
+					// Clone before reversing so we do not corrupt the shared cached
+					// curve that map() may return for other segments.
+					crv.reset(crv->clone_());
 					crv->reverse();
 				}
 				if (crv->kind() == taxonomy::EDGE) {
--- a/src/ifcgeom/mapping/IfcEdge.cpp
+++ b/src/ifcgeom/mapping/IfcEdge.cpp
@@ -43,6 +43,9 @@ taxonomy::ptr mapping::map_impl(const IfcSchema::IfcEdge* inst) {
 		auto basis = map(inst->as<IfcSchema::IfcEdgeCurve>()->EdgeGeometry());
 		auto loop = taxonomy::dcast<taxonomy::loop>(basis);
 		if (loop && loop->children.size() == 1) {
+			// Clone before mutating so calculate_linear_edge_curves does not write
+			// into the child edges of the shared cached loop returned by map().
+			loop.reset(loop->clone_());
 			loop->calculate_linear_edge_curves();
 			basis = loop->children[0]->basis;
 		}
--- a/src/ifcgeom/mapping/IfcFace.cpp
+++ b/src/ifcgeom/mapping/IfcFace.cpp
@@ -26,18 +26,14 @@ taxonomy::ptr mapping::map_impl(const IfcSchema::IfcFace* inst) {
 	auto bounds = inst->Bounds();
 	for (auto& bound : *bounds) {
 		if (auto r = taxonomy::cast<taxonomy::loop>(map(bound->Bound()))) {
+			// Clone before mutating so we do not corrupt the shared cached loop
+			// that map() may return for other faces referencing the same bound.
+			r.reset(r->clone_());
 			if (!bound->Orientation()) {
 				r->reverse();
 			}
 			// @todo check why loop sets external to true initially
 			r->external = bound->declaration().is(IfcSchema::IfcFaceOuterBound::Class());
-			/*
-			// Make a copy in case we need immutability later for e.g. caching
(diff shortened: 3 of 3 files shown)


src/ifcgeom/Iterator.cpp
```diff
@@ -211,7 +211,16 @@ void IfcGeom::Iterator::process_concurrently() {
 		if (worker_logger.OutputFormat() != Logger::FMT_INMEMORY) {
 			worker_logger.SetOutput(static_cast<std::ostream*>(nullptr), static_cast<std::ostream*>(nullptr));
 		}
-		kernel_pool.push_back(new ifcopenshell::geometry::Converter(std::unique_ptr<ifcopenshell::geometry::kernels::AbstractKernel>(converter_->kernel()->clone(worker_logger)), ifc_file, settings_, worker_logger));
+		auto* worker_converter = new ifcopenshell::geometry::Converter(std::unique_ptr<ifcopenshell::geometry::kernels::AbstractKernel>(converter_->kernel()->clone(worker_logger)), ifc_file, settings_, worker_logger);
+		// The mapping cache returns shared taxonomy items that are not immutable: reusing a
```
aothms: While there is definitely work on immutability, it should be fairly under control. I'd need more proof that this is actually do to mutation and then rather adapt there. Especially since converter should create a new mapping with it's own cache so 'shared taxonomy items' and non-deterministic perhaps do not apply.

BIMvoice:
Switched to the mutation site fix you preferred over disabling the cache. Instead of turning off the worker mapping cache, this clones the shared taxonomy item before any in place mutation, at every site that writes into a map() result: IfcFace (reverse plus external), IfcCompositeCurve (reverse), and IfcEdge (calculate_linear_edge_curves writing into child edges). It mirrors the clone before mutate you already use in IfcOrientedEdge, IfcDerivedProfileDef and IfcArbitraryOpenProfileDef, and I included a full audit of the remaining map() result uses, which only read or store the item. The cache stays enabled. I have not re run the multithreaded before and after on a kernel build in this iteration, the argument is code equivalence with the existing clone idiom, so if you want I can run the dental_clinic comparison again on a build to confirm empirically.

aothms (approved):

Issue #6581, opened 2025-04-19
Projection lines don't render out anymore it seems.


theoryshaw:
### Bug Description

Projection lines don't render out anymore it seems. 

Gif shows one print from an older version of BB, and one with the latest pull.

test file: 

[projection lines - Copy.ifc.txt](https://github.com/user-attachments/files/19822307/projection.lines.-.Copy.ifc.txt)

### Attachments

![Image](https://github.com/user-attachments/assets/d3ce49bd-d816-4a8f-8cb1-2c4ad352a7cd)

### Debug and Error Output

```true

```

theoryshaw:
i think it happened here: bb598d848a68eb57a13564023b62af24c0feaadf

Moult:
I've reverted the change that caused this but keep this open so that @aothms can investigate - is setOnlyValid being too strict? Can it differentiate between completely failing (i.e. no output, runtime error) and partially failing (still some output, may not be topologically sound)? Can it heal geometry? Etc.

Issue #4029, opened 2023-11-20
"Create Drawing" not working for some views.


Owuraqu:
Good morning. When I try to create some elevation drawings I get “An unkwown error occurred” report. Only one elevation (West Elevation) has been successful. I had same problem with creating some of the sections but when I deleted and created the section views again it worked but for the other three elevations it doesn’t work even after deleting and recreating them except the West Elevation. I even uninstalled and updated BB to yesterday’s version but same error surfaces. Find
[ifc file](https://www.dropbox.com/scl/fi/xo3p2n27d7vfsfuezlyk0/2373-XX-ODM-XX-03-M3-A-0001_BilkisSarpongKodieAR.ifc?rlkey=geuecpzm68zprm9u10flet3l2&dl=0).

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/33836708/ea8fe33c-6cc0-4b12-8621-07fdd6a6185a)

Moult:
@aothms what's the easiest way to help you debug these types of issues? I've isolated it to this file (containing only the terrain and a wall - if you remove one or the other, it works, but when both are together, they fail) and test script:

[wallterrain.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/13455281/wallterrain.ifc.txt)

```python
import ifcopenshell
import ifcopenshell.geom
import multiprocessing

ifc = ifcopenshell.open("/home/dion/tmp/svgcreationbug/wallterrain.ifc")

svg_settings = ifcopenshell.geom.settings(
    DISABLE_TRIANGULATION=True, STRICT_TOLERANCE=True, INCLUDE_CURVES=True
)
svg_buffer = ifcopenshell.geom.serializers.buffer()
serialiser = ifcopenshell.geom.serializers.svg(svg_buffer, svg_settings)
serialiser.setFile(ifc)
serialiser.setWithoutStoreys(True)
serialiser.setPolygonal(True)
serialiser.setUseHlrPoly(True)
# Objects with more than these edges are rendered as wireframe instead of HLR for optimisation
serialiser.setProfileThreshold(10000)
serialiser.setUseNamespace(True)
serialiser.setAlwaysProject(True)
serialiser.setAutoElevation(False)
serialiser.setAutoSection(False)
serialiser.setPrintSpaceNames(False)
serialiser.setPrintSpaceAreas(False)
serialiser.setDrawDoorArcs(False)
serialiser.setNoCSS(True)
[...]

PR #8371, opened 2026-07-07
svg: don't gate out drawable non-solid shapes (terrain) from sections in setOnlyValid mode (#4029)

BIMvoice:
Addresses #4029 (partial: the terrain-disappears half; related to #6581).

On current v0.8.0 the 2023 "unknown error" crash is gone; what remains is the `setOnlyValid` behaviour Moult flagged in 2025: a terrain section disappears with `setOnlyValid(True)` and renders messy with `setOnlyValid(False)`.

**Cause:** `SvgSerializer.cpp:766` runs the `only_valid_` gate `validate_shape()` on the element's **source** shape, *before* it is sectioned. `validate_shape` (`base_utils.cpp`) uses `BRepCheck_Analyzer::IsValid()`, i.e. it requires a topologically sound solid. A terrain is an open DEM surface, so OCC reports `BRepCheck_UnorientableShape` on every face (165) plus 21 `BRepCheck_SelfIntersectingWire` from the imported mesh, `IsValid()` is false, and the whole element is gated out before it ever reaches the section cut, even though its section renders fine with `setOnlyValid(False)`.

**Fix:** relax `validate_shape` to tolerate the drawable-but-unsound statuses (`UnorientableShape`, `SelfIntersectingWire`, `NotClosed`) and reject only when some genuinely fatal status is present, the "differentiate complete failure from partial" distinction raised in #6581. `validate_shape` is called from this one site, so the blast radius is contained.

**Verified with a from-source build:** built `libIfcParse` + `libIfcGeom` + the opencascade kernel + serializers + a harness mirroring Moult's repro (OCCT 7.9.2). With `wallterrain.ifc`:

| case | terrain | bytes |
|---|---|---|
| unpatched, only_valid=True (bug) | absent | 1451 |
| unpatched, only_valid=False | present | 5140 |
| patched, only_valid=True (fix) | present | 5140 |

The patched `only_valid=True` output is byte-identical to the `only_valid=False` render, terrain restored, wall intact.

**Scope (honest):** this fixes the terrain *disappearing*. It does not fix the separate *messiness* under `setOnlyValid(False)`: `ShapeAnalysis_FreeBounds::ConnectEdgesToWires` (1e-5) can't chain the terrain's gappy/self-intersecting cut edges into clean wires. That's a deeper geometry-healing/tolerance problem, left out of this contained fix.

### DIFF
--- a/src/ifcgeom/kernels/opencascade/shell.cpp
+++ b/src/ifcgeom/kernels/opencascade/shell.cpp
@@ -2,6 +2,9 @@
 
 #include "base_utils.h"
 
+#include <BRepCheck_Analyzer.hxx>
+#include <ShapeFix_Shape.hxx>
+
 using namespace ifcopenshell::geometry;
 using namespace ifcopenshell::geometry::kernels;
 using namespace IfcGeom;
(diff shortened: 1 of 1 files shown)


aothms:
While undesirable, it is definitely risky to have faulty shapes propagate into the drawing pipeline, especially since there can also be boolean operations applied to them for trimming them to the section plane. None of this is ideal, but the generic validate_shape() function should not be altered for this specifically. Also OpenCASCADE can perfectly handle non-manifold or open shells, the fact that it is marked as not valid is not because it is an open shell.

BIMvoice:
Reworked without touching validate_shape. Instead of relaxing the gate, the tessellated shell is now healed at construction, in OpenCascadeKernel::convert of the shell (src/ifcgeom/kernels/opencascade/shell.cpp), before it reaches any boolean. The DEM terrain comes out of the per face construction as an unwelded compound where a numerically bad subset of faces carries UnorientableShape and SelfIntersectingWire, which is why validate_shape rejected it, not because it is open, exactly as you said. A ShapeFix pass, gated on BRepCheck invalidity and adopted only when it restores validity, removes those statuses while preserving the open compound topology, it does not force a closed solid, so validate_shape then passes unchanged and the section runs on a sound shape.

I verified this empirically with pythonocc 7.9.3 ABI matched to the OCCT here: on the real terrain shape ShapeFix removes exactly the UnorientableShape and SelfIntersectingWire statuses and restores BRepCheck_Analyzer::IsValid, the same predicate validate_shape uses, while keeping the shape a COMPOUND and preserving area and face count. I could not run a full kernel build, so the end to end drawing is not build tested, but the heal is proven on the exact validity predicate. This should also clean up the messy section in #6581, since the self intersecting source wires are gone. validate_shape and the generic path are untouched.

As a separate follow up, the per face plane fit in face.cpp is a major source of the unorientable faces and a Newell fit would prevent most at the source, but the heal is the correct general safety net since it also handles genuinely non planar quads that no plane fit can avoid.

PR #8376, opened 2026-07-08
svg: guard HLR stages so one failing element or projection does not abort the drawing (#3971)

### LINKED ISSUE #3971: Intelligent roof causing error with svgserializer
file: https://hub.openingdesign.com/OpeningDesign/Monica_Residence/src/commit/ed8b2d69bdf1c9e9f95dfd31d6aa4e081f2b546f/Models/BlenderBIM/Monica_Residence.ifc

https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/294f0dcd-81c2-4881-b124-8f068e2990bf

image

BIMvoice:
## Summary
The SVG serializer ran the OpenCASCADE hidden-line-removal routines in `prefiltered_hlr` with no exception handling, so a single failing element or a failing projection aborted the whole drawing. `finalize()` propagated a bare `Standard_Failure` that the Python wrapper reported as the opaque `RuntimeError: An unknown error occurred`. On the issue's Monica_Residence model, 17 drawings failed this way.

## Root cause
Characterized on that model: OCCT throws `Standard_OutOfRange` (`NCollection_Array1::ChangeValue`) in the poly HLR projection pass (`HLRBRep_PolyAlgo` / `PolyHLRToShape`), reached from `finalize()` via `build()`, even though the input mesh is topologically clean. It is an internal out-of-range in the projector, not a repairable mesh defect. The projection runs over the whole accumulated compound at once, so it has no per-element granularity.

## Change
Per @aothms's guidance (skip in C++, let Python conditionally raise): guard each HLR stage (manifold test, prefilter face analysis, obscuration cache, per-element accumulation, projection) and on failure skip that unit while emitting a message through the non-static logger with a distinct SER error code (36 to 40) and the offending product where known. The projection failure is logged at ERROR (SER40) so a Python consumer can key on the code and raise. The catch sites handle `Standard_Failure` explicitly and report its `DynamicType` and `GetMessageString`, because on this OCCT build `Standard_Failure` does not derive from `std::exception`; a plain `catch (std::exception&)` would never see it and the log would stay generic.

## Verification
On Monica_Residence: drawings that threw now finalize (SOUTH ELEVATION, SECTION - EAST, SECTION - NORTH complete; each logs SER40 with the OCCT detail), a clean drawing (NORTH ELEVATION) is byte-identical to the unguarded output, and no exception leaves `finalize()`.

Fixes #3971

### DIFF
--- a/src/serializers/SvgSerializer.h
+++ b/src/serializers/SvgSerializer.h
@@ -38,6 +38,7 @@
 #include <gp_Pln.hxx>
 #include <Bnd_Box.hxx>
 #include <Standard_Version.hxx>
+#include <Standard_Failure.hxx>
 #include <BRep_Builder.hxx>
 #include <HLRBRep_PolyHLRToShape.hxx>
 #include <BRepTopAdaptor_FClass2d.hxx>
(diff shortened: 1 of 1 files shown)


BIMvoice:
Rebased on current v0.8.0, resolving the overlap with the new SVG edge-classification code (set_classified_shapes is preserved inside the guarded projection pass). No behavior change to the fix itself.

PR #8377, opened 2026-07-08
Approximate swept solids in the taxonomy layer so cgal stops dropping IfcSweptDiskSolid (#8106)

### LINKED ISSUE #8106: CGAL kernel skips IfcSweptDiskSolid when directrix is IfcIndexedPolyCurve with IfcArcIndex segments
## Bug

The CGAL geometry kernel silently produces no mesh for `IfcSweptDiskSolid`
items whose `Directrix` is an `IfcIndexedPolyCurve` containing `IfcArcIndex`
segments. OpenCascade tessellates the same items correctly.

## Repro

Found out after noticing roundtripping parametric railings in Bonsai with wall supports displayed only some parts of the supports.

Image

instead of

Image

Weirdly manual deleting a representation in the representation UI triggers correctly displaying the geometry.

Image

Couldn't manage to bisect any working version, most likely is like that since forever ? Can anyone reproduce a setup working as intended?

Attach `railing-repro.ifc` (a Bonsai-saved wall-mounted handrail — 7 items:
3 `IfcExtrudedAreaSolid` + 4 `IfcSweptDiskSolid`).

```python
import ifcopenshell, ifcopenshell.geom
f = ifcopenshell.open("railing-repro.ifc")
railing = f.by_type("IfcRailing")[0]
s = ifcopenshell.geom.settings()
for lib in ("opencascade", "cgal", "cgal-simple"):
    r = ifcopenshell.geom.create_shape(s, railing, geometry_library=lib)
    print(f"{lib}: {len(r.geometry.verts)//3} verts")
```

Output
opencascade: 1924 verts (all 7 items)
cgal: 96 verts (only the 3 IfcExtrudedAreaSolid items)
cgal-simple: 96 verts (only the 3 IfcExtrudedAreaSolid items)
Expected
CGAL should tessellate the 4 IfcSweptDiskSolid items the same way OpenCascade does, or raise an error — not silently emit an empty mesh for them.

Versions tested
Reproduces on ifcopenshell 0.8.3.post2, 0.8.4, 0.8.4.post1, 0.8.5, and current main. Not a regression — appears to have always been this way.

### Attachments

[railing-repro.ifc.txt](https://github.com/user-attachments/files/28283541/railing-repro.ifc.txt)

### Debug and Error Output

```true

```

BIMvoice:
## Problem
The cgal kernel has no `convert_impl` for `taxonomy::sweep_along_curve`. Every `IfcSweptDiskSolid` throws `not_implemented`, fails item-upgrade dispatch and is silently dropped when cgal is the last kernel in the stack, so swept disks disappear under cgal / cgal-simple while opencascade renders them via its native pipe.

## Approach
Following @aothms's direction, the sweep is approximated in the taxonomy layer rather than in a kernel. New member method `taxonomy::sweep_along_curve::as_shell()` (Eigen only) tessellates the directrix (line + arc segments), propagates a rotation-minimizing frame (double reflection), and instances the profile face at each station into a watertight tube with end caps, returning a plain `taxonomy::shell` of planar faces that any kernel can consume.

`AbstractKernel::convert` invokes this as a fallback when a kernel reports `not_implemented` for a swept solid, so cgal recovers the geometry by default. A new `ApproximateSweptSolids` (`--approximate-swept-solids`) conversion setting forces the approximation on for every kernel, including opencascade, so the native and approximated results can be compared.

## Verification
Reporter's railing model (7 `IfcExtrudedAreaSolid`, 8 `IfcSweptDiskSolid`), full opencascade + cgal build:
- cgal before: 224 verts, 8 `sweep_along_curve` drop errors. After: 1664 verts, 0 errors.
- opencascade native: unchanged (3484 verts). With the setting on: consumes the approximated shell, 1804 verts, no failed faces.
- cgal's polyhedron validity check passing on all swept disks confirms the approximated shells are watertight manifolds.

## Scope
Covers `IfcSweptDiskSolid` and `IfcFixedReferenceSweptAreaSolid`. `IfcSurfaceCurveSweptAreaSolid` (reference surface) is left on its native path (`as_shell` returns `nullptr`) as a follow up.

Fixes #8106

### DIFF
--- a/src/ifcgeom/AbstractKernel.cpp
+++ b/src/ifcgeom/AbstractKernel.cpp
@@ -58,6 +58,22 @@ bool ifcopenshell::geometry::kernels::AbstractKernel::convert(const taxonomy::pt
 		return dispatch_conversion<0>::dispatch(this, shell->kind(), shell, results);
 	};
 
+	// Kernel-agnostic faceted lofting (Eigen only, see taxonomy::loft::as_shell).
+	// The polyhedral loft logic no longer lives only in the opencascade kernel, so
+	// a kernel without a native loft (cgal) can consume the tessellated shell
+	// instead of dropping the item.
+	auto try_loft_approximation = [&]() -> boost::optional<bool> {
+		auto lft = taxonomy::dcast<taxonomy::loft>(item);
+		if (!lft) {
+			return boost::none;
+		}
+		auto shell = lft->as_shell();
+		if (!shell) {
+			return boost::none;
+		}
+		return dispatch_conversion<0>::dispatch(this, shell->kind(), shell, results);
+	};
+
 	auto process_with_upgrade = [&]() {
 		// Forced approximation mode: applies to every kernel (including opencascade).
 		if (settings_.get<settings::ApproximateSweptSolids>().get()) {
@@ -69,8 +85,12 @@ bool ifcopenshell::geometry::kernels::AbstractKernel::convert(const taxonomy::pt
 			return dispatch_conversion<0>::dispatch(this, item->kind(), item, results);
 		} catch (const not_implemented_error&) {
 			// The kernel has no native conversion for this item. If it is a swept
-			// solid, fall back to the tessellated-shell approximation before giving up.
+			// solid or a loft, fall back to the tessellated-shell approximation
+			// before giving up.
 			if (auto res = try_sweep_approximation()) {
+				return *res;
+			}
+			if (auto res = try_loft_approximation()) {
 				return *res;
 			}
 			return dispatch_with_upgrade<0>::dispatch(this, item, results);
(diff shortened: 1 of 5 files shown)


BIMvoice:
@aothms pushed the loft follow-up (a6ca6634e), applying the same taxonomy-layer approximation to lofts.

The polyhedral (faceted) loft branch that previously lived only in the opencascade kernel now lives on taxonomy::loft::as_shell() as an Eigen-only tessellated shell, so cgal (which has no native loft) can consume the shell instead of dropping the geometry. AbstractKernel tries the loft approximation for kernels without native support, and the opencascade loft consumes the same taxonomy shell for its polygonal branch, falling back to native lofting when the shell cannot be built.

Honest limit: this compiles and links, but I have NOT runtime-tested it (no environment here to drive a model through cgal end to end). Please build and confirm against a file with lofted geometry before merging, and I am happy to adjust the taxonomy::loft::as_shell tessellation if the faceting is not what you want.

BIMvoice:
Rebased on current v0.8.0. The arc tessellation now follows the same two-mode rule as the CGAL kernel after the circle-segments rework (fixed count when CircleSegments > 0, MesherLinearDeflection-derived when 0), so the approximation keeps full fidelity under the new 0 default.

PR #8378, opened 2026-07-08
Add ifcopenshell.diagnostics: report the entity most likely behind a crash

BIMvoice:
This follows the direction Thomas gave when closing #8321 and #8359. Both of those added a defensive guard at a single call site against a required attribute that was null in an invalid file. That is the wrong layer, there are many such call sites and the schema says the attribute is not optional. The better answer is one generic mechanism, so here it is as a first version to react to.

## What it does

Given an exception, `ifcopenshell.diagnostics.diagnose` walks the traceback, collects the `entity_instance` objects that were live in the frames, validates them with `ifcopenshell.validate`, and reports the entity most likely responsible with a confidence estimate. `str(report)` is a human readable summary and `report.likely_cause` is the top suspect.

On a file like #7170 the `AttributeError: 'NoneType' object has no attribute 'MappingOrigin'` becomes:

```
Most likely cause (high confidence):
  #1=IfcMappedItem($,#3)  [high confidence]
      in frame get_mappeditem_transformation (util/placement.py:182)
      validation:
        - IfcMappedItem.MappingSource: Attribute not optional
```

On a file like #5113 the null `IfcDirection` is named the same way, found by following `IfcAxis2Placement3D.Axis` when it is not a local itself.

## How it works

- `validate.py`: the per instance checks are factored out of `validate()` into `validate_instance(inst, logger, schema=None)`. `validate()` now delegates to it. Behaviour is unchanged and `test_validate` still passes.
- `diagnostics.py`: `diagnose()` plus an `on_error()` context manager. Ranking is intentionally simple. An entity that failed validation always outranks a clean one. Among failing ones, an entity that sat in a frame outranks one only reachable by reference, and a frame closer to the raise outranks one further away. Confidence is high for an in frame schema violation, medium for a referenced one, low for a clean entity. When nothing fails validation the report says so rather than guessing.

## Scope of this first version

I kept it as a standalone utility and did not wire a root exception handler yet. The natural place for that hook is the consumers that already wrap a whole file operation, the Bonsai project load operator and the geometry iterator, plus the ifcpatch and ifcconvert CLI paths, rather than inside util or geom. I am happy to add that hook wherever you prefer in a follow up.

## Tests

Six tests in `test_diagnostics.py` cover the missing MappingSource case, the null DirectionRatios case both in frame and reference only, a clean entity that must not be blamed, an empty traceback, and the bare `diagnose()` default. The existing validate suite is unaffected.

This is a first cut to get the shape right. Tell me if the module location, the API, or the confidence heuristic should change, and where you want the high level hook to live.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/diagnostics.py
+++ b/src/ifcopenshell-python/ifcopenshell/diagnostics.py
@@ -1,0 +1,331 @@
+# IfcOpenShell - IFC toolkit and geometry engine
+# Copyright (C) 2021 Thomas Krijnen <thomas@aecgeeks.com>
+#
+# This file is part of IfcOpenShell.
+#
+# IfcOpenShell is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# IfcOpenShell is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with IfcOpenShell.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Crash diagnostics for invalid IFC files.
+
+Many crashes deep inside geometry or utility code are not really bugs in the
+code. They are the symptom of an invalid file: a required attribute that is not
+set, a reference that points at nothing, a value of the wrong type. Guarding
+every call site against every possible schema violation is not maintainable, so
+instead of being defensive everywhere this module turns a raised exception into
+a report that points at the entity most likely responsible.
+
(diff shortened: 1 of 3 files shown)


aothms:
Thanks. Let's see if and how the bonsai core devs want to integrate this. Invalid data is not a real priority at this stage of the development. But it's good to have this at our disposal.

BIMvoice:
Following up since it's been quiet for a few days — totally fine if invalid-data tooling isn't a priority right now, just want to know how to read that for this PR specifically.

This module doesn't require any Bonsai wiring to be useful on its own (it's a standalone `ifcopenshell.diagnostics.diagnose` utility, opt-in), so it can merge or sit as-is without anyone committing to the Bonsai integration question right now if that's easier. Two things would help us know how to proceed:

1. Is this mergeable as a standalone diagnostics utility independent of whether/when Bonsai core wires it in, or do you want to hold it until there's an actual integration consumer?
2. If it's waiting on the "bonsai core devs" weighing in and nobody does — is there a point where silence should just be read as no objection, or would you rather we ping specific people / keep it open indefinitely either way? Not trying to rush anyone, just want to know what "no response" means here so we don't let it go stale for the wrong reason.

Issue #8255, opened 2026-07-05
Bonsai New Project Wizard error Kilometer, Miles


S59-Koara:
### Bug Description

In Bonsai, a Python error appears when selecting certain units in the New Project Wizard.
I've noticed this happens when selecting “Kilometers” under “Metric” and ‘Miles’ under “Imperial.”

### Debug and Error Output

```true

```

BIMvoice:
Tried to pin this down on `v0.8.0` but the downstream path checks out at the API level: `tool.Unit.get_scene_unit_name` maps Blender's `KILOMETERS` to `KILO/METRE` (a valid SI prefix form) and `MILES` to `mile`, and both `ifcopenshell.api.unit.add_si_unit(prefix="KILO")` and `add_conversion_based_unit(name="mile")` succeed when called directly. So the failure is somewhere in the wizard flow I can't see without the actual error.

@S59-Koara could you paste the Python traceback? (When the error appears, the full text is in `Window > Toggle System Console` on Windows, or click the error popup to copy it.) With the traceback this should be a quick fix.

S59-Koara:
File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\5.1\extensions\.local\lib\python3.13\site-packages\bonsai\bim\module\drawing\helper.py", line 199, in format_distance
    unit_length = unit_length_mapping[unit_length]
                  ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'KILOMETRE'
00:17.453  bpy.rna          | ERROR Python script error in BIM_OT_create_project.execute
00:17.453  operator         | ERROR Python: Traceback (most recent call last):
                            |   File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\5.1\extensions\.local\lib\python3.13\site-packages\bonsai\bim\module\project\operator.py", line 146, in execute
                            |     self._execute(context)
                            |     ~~~~~~~~~~~~~^^^^^^^^^
                            |   File "C:\Users\Owner\AppData\Roaming\Blender Foundation\Blender\5.1\extensions\.local\lib\python3.13\site-packages\bonsai\bim\module\project\operator.py", line 162, in _execute
                            |     core.create_project(
                            |     ~~~~~~~~~~~~~~~~~~~^
                            |         tool.Ifc, tool.Georeference, tool.Project, tool.Spatial, schema=props.export_schema, template=template
[...]

PR #8379, opened 2026-07-08
fix(bonsai): stop unusual metric length units from crashing format_distance

### LINKED ISSUE #7102: Unusual metric units in Project Wizard cause errors
### Bug Description

The project wizard presents Kilometre and Micrometre as selectable Metric Length Units.

Trying to then add a wall (WAL300 from the IFC4 Demo Template as an example) will throw the following exception.

```
  File "/home/steve/.config/blender/bonsai/extensions/.local/lib/python3.11/site-packages/bonsai/bim/ifc.py", line 530, in execute_ifc_operator
    result = getattr(operator, "_modal")(context, event)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/steve/.config/blender/bonsai/extensions/.local/lib/python3.11/site-packages/bonsai/bim/module/model/wall.py", line 692, in _modal
    f"Offset Value: {tool.Polyline.format_input_ui_units(props.offset * self.unit_scale)}",
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/steve/.config/blender/bonsai/extensions/.local/lib/python3.11/site-packages/bonsai/tool/polyline.py", line 487, in format_input_ui_units
    return format_distance(
           ^^^^^^^^^^^^^^^^
  File "/home/steve/.config/blender/bonsai/extensions/.local/lib/python3.11/site-packages/bonsai/bim/module/drawing/helper.py", line 159, in format_distance
    unit_length = unit_length_mapping[unit_length]
                  ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'KILOMETRE'
```

Either Kilometre and Micometre should be removed (to me they are not practical _working_ measurements) or they need to be allowed for in the helper.py format_distance function.

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Reported by @sboddy in #7102.

The New Project Wizard uses Blender's native length-unit dropdown, which offers Kilometre and Micrometre for metric projects. Selecting one and then drawing a wall crashed:

    File ".../bonsai/bim/module/drawing/helper.py", line 159, in format_distance
      unit_length = unit_length_mapping[unit_length]
    KeyError: 'KILOMETRE'

`format_distance` derives the active IFC length unit name (including any SI Prefix, giving names like KILOMETRE or MICROMETRE) and looked it up in a hardcoded dict with a direct subscript. That dict only covers FOOT, INCH, METRE, DECIMETRE, CENTIMETRE and MILLIMETRE, so any other valid IFC SI length unit raised a KeyError. This affects the wizard, but also any IFC file authored elsewhere that legitimately uses these units, so the fix belongs in the downstream handling rather than in the offered list.

The metric formatter already has an adaptive fallback branch for any unit it does not handle explicitly. Switching the lookup to `dict.get` with the raw unit name as the fallback lets kilometre, micrometre and any other unmapped SI length unit reach that branch and format correctly instead of crashing. The known units are unaffected.

Verified in Blender 5.1.2 by building IFC files with each SI length unit and calling `format_distance` directly. Before, KILOMETRE and MICROMETRE raised KeyError; after, they return a valid metric string. METRE, MILLIMETRE and CENTIMETRE output is unchanged.

Fixes #7102

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/helper.py
+++ b/src/bonsai/bonsai/bim/module/drawing/helper.py
@@ -196,7 +196,10 @@ def format_distance(
                 "CENTIMETRE": "CENTIMETERS",
                 "MILLIMETRE": "MILLIMETERS",
             }
-            unit_length = unit_length_mapping[unit_length]
+            # Any valid IFC SI length unit that has no dedicated formatting branch
+            # below (for example KILOMETRE or MICROMETRE) falls through to the
+            # adaptive metric formatting instead of raising a KeyError.
+            unit_length = unit_length_mapping.get(unit_length, unit_length)
         # For now we only format area in IFC Units
         if area_unit := ifcopenshell.util.unit.get_project_unit(tool.Ifc.get(), "AREAUNIT"):
             area_unit_symbol = " " + ifcopenshell.util.unit.get_unit_symbol(area_unit)


BIMvoice:
Closing: superseded by 980988f208 (merged), which applies the same unit-mapping fallback plus explicit KILOMETRE/MICROMETRE entries for #8255/#7102.

Issue #1089, opened 2020-11-08
BlenderBIM - IFC Nodes - 'IFC Get Attribute' not pulling value at a certain level


theoryshaw:
IFC Get Attribute doesn't pull out value at a certain level. 

https://www.dropbox.com/s/z9dpag6f4ilm5dw/20201107%20-%20sv%20ifcwall_dumb%20wall%2002.blend?dl=1

![image](https://user-images.githubusercontent.com/507113/98454084-1fbebd80-2126-11eb-8e07-2e95b1079bbf.png)

htlcnn:
I'm wondering how "Location" attribute of IfcLocalPlacement could be retrieved when it is relative placement. Could you show me the code @theoryshaw @Moult ?

htlcnn:
@Moult @theoryshaw 
It took me a while to understand the code base of the node Ifc Get Attribute. I think these lines were the cause to improper operation:
- Firstly, `SvIfcGetAttribute.process()` calls to its parent's (SvIfcCore) `process()`
https://github.com/IfcOpenShell/IfcOpenShell/blob/877bf3207a120672fe1af89f068ab2e983f590ce/src/ifcsverchok/nodes/ifc/get_attribute.py#L24
- Secondly, `zip_longest_repeat` apply `list` function to all inputs. In the case that an entity_instance can be converted to list, it will become another data type. In the case of IfcCatersianPoint, it turns to a tuple, so a tuple can not be `get_info()` again. Exception raised.

I think it will be more explicit and match to Ifc data structure if we bash `zip_longest_repeat` and handle the nested inputs for this specific node instead of doing it generally in SvIfcCore. I propose a solution below:
```python
def process(self):
        self.sv_input_names = ["entity", "attribute_name"]
        self.value_out = []
        entity_nested_inputs = self.inputs["entity"].sv_get()
        attribute_name = self.inputs["attribute_name"].sv_get()[0][0]
        for entities in entity_nested_inputs:
            if hasattr(entities, "__iter__"):
                for entity in entities:
                    self.process_ifc(entity, attribute_name)
            else:
                self.process_ifc(entities, attribute_name)
        self.outputs["value"].sv_set(self.value_out)
```
This is the result:
![image](https://user-images.githubusercontent.com/931924/103145037-31a40080-4766-11eb-9de6-a501b46c285b.png)

If we're not expecting a list of attributes provided into `attribute` socket, I think this is ok.

Moult:
[...]

PR #8380, opened 2026-07-08
fix(ifc5d): correct swapped Width and Depth in Qto_SlabBaseQuantities

### LINKED ISSUE #8053: Depth and Width in Qto_SlabBaseQuantities are swapped
### Bug Description

There is an error in the assignment of Depth and Width for slabs (IfcMaterialLayerSet) in the qto-function (IfcOpenShell and Blender) according to the definition in the specification: https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/Qto_SlabBaseQuantities.htm

**Width** should be the "**thickness**" of the slab (vertical) and **Depth** one of the **horizontal dimensions**.

see also discussion here: https://community.osarch.org/discussion/comment/28119#Comment_28119
also see issue to improve standard: https://github.com/buildingSMART/IFC4.x-development/issues/1089

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
The base quantity rule files assigned the slab Width and Depth the wrong way around versus the buildingSMART definition of Qto_SlabBaseQuantities.

Per the IFC4x3 spec:
- Width is the width of the object, only given if it has constant thickness (prismatic). This is the slab thickness.
- Depth is one direction of the non-projected footprint area.
- Length is the other footprint direction.

The rule files mapped Depth to the vertical thickness and Width to a horizontal footprint dimension, so the two values came out swapped. This affected both calculators and both schemas:
- IFC4QtoBaseQuantities.json and IFC4X3QtoBaseQuantities.json (IfcOpenShell): Depth net_get_z, Width net_get_y
- IFC4QtoBaseQuantitiesBlender.json and IFC4X3QtoBaseQuantitiesBlender.json (Blender): Depth get_height, Width get_width

The fix swaps the Depth and Width formulas in all four files. Length is unchanged.

Verified with a rectangular slab of footprint 5.0 by 3.0 and thickness 0.2, run through the IfcOpenShell calculator:
- before: Width 3.0, Depth 0.2
- after:  Width 0.2, Depth 3.0
Length stayed 5.0.

The Blender calculator path was verified by inspection: get_height returns the local Z extent (thickness) and get_width returns the smaller of the two horizontal extents, so the same swap applies. It was not driven through a live Blender session. Testing covered a single axis-aligned rectangular slab.

Thanks to @IO10101 for the report and @steverugi for the spec discussion.

Fixes #8053

### DIFF
--- a/src/ifc5d/ifc5d/IFC4QtoBaseQuantities.json
+++ b/src/ifc5d/ifc5d/IFC4QtoBaseQuantities.json
@@ -507,7 +507,7 @@
             },
             "IfcSlab": {
                 "Qto_SlabBaseQuantities": {
-                    "Depth": "net_get_z",
+                    "Depth": "net_get_y",
                     "GrossArea": "gross_get_footprint_area",
                     "GrossVolume": "gross_get_volume",
                     "GrossWeight": "gross_get_weight",
@@ -516,7 +516,7 @@
                     "NetVolume": "net_get_volume",
                     "NetWeight": "net_get_weight",
                     "Perimeter": "net_get_footprint_perimeter",
-                    "Width": "net_get_y"
+                    "Width": "net_get_z"
                 }
             },
             "IfcSolarDevice": {
--- a/src/ifc5d/ifc5d/IFC4QtoBaseQuantitiesBlender.json
+++ b/src/ifc5d/ifc5d/IFC4QtoBaseQuantitiesBlender.json
@@ -507,7 +507,7 @@
             },
             "IfcSlab": {
                 "Qto_SlabBaseQuantities": {
-                    "Depth": "get_height",
+                    "Depth": "get_width",
                     "GrossArea": "get_gross_footprint_area",
                     "GrossVolume": "get_gross_volume",
                     "GrossWeight": "get_gross_weight",
@@ -516,7 +516,7 @@
                     "NetVolume": "get_net_volume",
                     "NetWeight": "get_net_weight",
                     "Perimeter": "get_gross_perimeter",
-                    "Width": "get_width"
+                    "Width": "get_height"
                 }
             },
             "IfcSolarDevice": {
--- a/src/ifc5d/ifc5d/IFC4X3QtoBaseQuantities.json
+++ b/src/ifc5d/ifc5d/IFC4X3QtoBaseQuantities.json
@@ -643,7 +643,7 @@
             },
             "IfcSlab + IfcSlabType": {
                 "Qto_SlabBaseQuantities": {
-                    "Depth": "net_get_z",
+                    "Depth": "net_get_y",
                     "GrossArea": "gross_get_footprint_area",
                     "GrossVolume": "gross_get_volume",
                     "GrossWeight": "gross_get_weight",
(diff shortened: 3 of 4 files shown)


steverugi:
@BIMvoice sorry but it's wrong

please type the following prompt to all major AI:
"define the thickness of a slab (as a term) picking from one the following nouns: width, length, height"
Gemini, ChatGPT, Grok all reply the same: the closest noun to thickness is height

or if you want to know what my organic intelligence (or decades of experience) says: it's the height
or ask any person actually working in the construction industry if you like
This is not the first time buildingSMART had an issue with ambiguous definition by the way

Tyzze86:
I agree with @steverugi here. The depth or thickness which I would define as a more correct term for a slab is the defined as the height from the bottom of the slab to the top of the slab. This is due to the the fact a slab tradinitionally has had a top-down view or bottom-up view (formwork). 

You're example of a slab with dimensions: 5.0 x 3.0 x 0.2 (length x width x thickness) says this directly that the depth is .2 meters. For me the new input reads like a wall 5 meters long, 3 meters high and .2 meter thick and not like a slab with the first dimensions.

BIMvoice:
Thanks @Tyzze86, and I genuinely agree with you and @steverugi that "Width" reads badly for a slab's thickness. That part is not in dispute, it is counterintuitive.

But this PR is not choosing what the quantity should be called. It aligns ifc5d with what the buildingSMART spec currently defines, so a takeoff is consistent across authoring tools. The IFC4.3 text for Qto_SlabBaseQuantities is explicit: Width is "only given if the object has constant thickness (prismatic)" (the thickness), and Depth is "one direction of the non-projected footprint area" (horizontal). Dion made the same call on the OSArch thread ("the spec is silly here, PR welcome to fix ifc5d").

The naming complaint is real and is already being handled at the right level: the original reporter opened buildingSMART/IFC4.x-development#1089 to rename Width to Thickness for exactly the reason you both give, and buildingSMART accepted it for a future release. I have added a comment there in support. That issue is the right place to settle the naming, because it changes the standard for every tool rather than making ifc5d disagree with the spec on its own.

So we mostly agree: the name is bad and buildingSMART is fixing that in #1089. This PR just puts the value where the current standard says it goes, so cross-tool quantity takeoff stays consistent today.

Tyzze86:
Then the question is if that wall all other softwares do. When I see a QTO i will usually intuitively expect width to be the actual width of an object and not the height or thickness. So if I on a slab see that the width is .2 I will react to that and thinking this is wrong. I look at this from a practical standpoint and if other softwares doesn't do it either this shouldn't be implemented here. 

Further as buildingSMART has put this issue as after NWI (New work item) in ISO, meaning ISO has accepted this as a work item. I think that this should be on hold until IFC is updated.

BIMvoice:
Thanks, and understood on the concern about timing. To be clear this is not a reinterpretation on our part: in the current IFC4.3 base-quantity definitions Qto_SlabBaseQuantities.Width is the slab thickness, so the change aligns the value with the published spec rather than getting ahead of it. The confusing naming (Width vs Depth) is a known wording issue being addressed upstream at buildingSMART (see bSI IFC4.3 issue about the slab quantity names), which is a documentation/label fix, not a data-model change.

That said, whether to land this now or hold until the naming is clarified upstream is a maintainers call, and I am happy to defer to that. If the preference is to wait, feel free to mark it accordingly; if it is fine to proceed on the current spec text, it is ready.

PR #8382, opened 2026-07-08
Bonsai: resolve sheet drawing via document Location instead of fuzzy name match

### LINKED ISSUE #5493: Bonsai, error when using , in drawing names and clicking on the drawing in the sheet part of UI.
I had this error for a while, and finally i figured out the reason. I tought i was breaking my models in some way, but a lot more simple. 
So here is an issue on it, to bring it to attention. 

Easy fix, but maybe this should be more clear to the end user.
- initial user (me) fix, reload the .ifc, lose session settings, and maybe dont trigger the mistake again.
- The error only occurs when you click on a drawing with , in it eg. "PLAN, 1. floor" in the sheet UI. 
- When clicked on, the sheet ui collapses and can only be reopened by reopening the file or from my latest findings remove any , in drawing names. 
- For reference the name of the drawing will appear without the , no matter if it in the name or not. guess it makes sense python doesnt want , ...
- When in the error state sometimes saving returns:
```
  File "C:\Users\DBO\AppData\Roaming\Blender Foundation\Blender\4.2\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\ui.py", line 446, in draw
    op = drawing_button.operator("bim.activate_drawing", icon="OUTLINER_OB_CAMERA", text="")
    ^^
UnboundLocalError: cannot access local variable 'drawingid' where it is not associated with a value
```

Here is a file.
[BreakSheetUi.ifc.txt](https://github.com/user-attachments/files/17138047/BreakSheetUi.ifc.txt)
1) open sheets and click on the drawing "Plan view" inside A00 - Name sheet.
2) "Error" occurs
3) Remove "," from the drawing name, in drawing ui. Everything works again. 

I don't know if this should be fixed or not, i will leave that to the brains. But from my background it's not unusual to add "," in naming. Else if no fix, maybe some more clear error message, or a conditioned text that appears. Could maybe even be as simple as adding to the description when hovering names "Active drawing index, NEVER use ,  in names, unless you know what you are doing"

BIMvoice:
Clicking a drawing inside a sheet in the Sheets panel needs the drawing's IfcAnnotation id to activate it. The code reconstructed the drawing name from the sheet reference's SVG filename and reverse matched it against every `IfcAnnotation.Name`. That filename is produced by `sanitise_filename`, which strips every character outside a safe set, so the reverse match is inherently lossy. Drawing names with a comma (and `#`, parentheses, `@`, `/` and similar) failed to resolve, leaving the local `drawingid` unbound and raising `UnboundLocalError: cannot access local variable 'drawingid'`, collapsing the sheet UI on click and on save.

The merged workaround in d581174 only stripped commas, so any other sanitised character still broke it, and @Moult reopened asking for a proper fix.

Root cause and fix: a sheet's DRAWING `IfcDocumentReference` is created with the same `Location` as the drawing's own document (`get_drawing_document`), and `get_sheet_references` already uses that `Location` equality in the forward direction. `Location` is the real, punctuation independent link. This PR adds `Drawing.get_drawing_from_sheet_reference` (the inverse of `get_sheet_references`) and uses it in `BIM_PT_sheets`, dropping the comma stripping name match entirely. Names with any punctuation now resolve correctly.

Verified against the reporter's sample file (`BreakSheetUi.ifc`): drawing `'PLAN, View'` has document `Location` `drawings/PLAN View.svg`, matching the sheet reference `Location` exactly, and now resolves to the correct id. A synthetic `#` in name case confirms the new resolver succeeds where the old comma only hack returned nothing. Black (120), Ruff and ty are clean on the changed files.

Verified at the resolution logic level against the sample file, not via a live GUI click through in Blender.

Reported by @Slackking1. Initial workaround by @theoryshaw; better fix requested by @Moult.

Fixes #5493

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -2398,6 +2398,10 @@ class ActivateDrawingBase(tool.Ifc.Operator):
                 tool.Drawing.import_annotations_in_group(tool.Drawing.get_drawing_group(selected_drawing))
             return {"FINISHED"}
 
+        if not self.drawing:
+            self.report({"WARNING"}, "No drawing is associated with this sheet item.")
+            return {"CANCELLED"}
+
         drawing = tool.Ifc.get().by_id(self.drawing)
         dprops = tool.Drawing.get_document_props()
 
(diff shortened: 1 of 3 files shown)


theoryshaw:
another pr on your pr: https://github.com/BIMvoice/IfcOpenShell/pull/4

Issue #4934, opened 2024-06-27
[bug] Error when removing presentation layer: "Instance xxx not found"


Vanuan:
When attempting to remove a presentation layer in the Geometric Relationships section of the Geometry and Materials subtab, I encountered the following error:

RuntimeError: Instance xxxx not found

Steps to reproduce:
1. Open the Geometry and Materials subtab
2. Navigate to the Geometric Relationships section
3. Add presentation layer
4. Click assign/disable/freeze icons in random sequence several times
5. Attempt to remove a presentation layer

Environment:
- OS: Linux (Ubuntu)
- Blender version: 4.1.1
- BlenderBIM version: 0.0.240602
- Python version: 3.11.7
- Architecture: 64-bit
- IFC schema: IFC4

Full stacktrace:
```
os: Linux
os_version: #117-Ubuntu SMP PREEMPT Tue Apr 30 10:12:40 UTC 2024
python_version: 3.11.7
architecture: ('64bit', 'ELF')
machine: x86_64
processor: x86_64
blender_version: 4.1.1
blenderbim_version: 0.0.240602
last_actions: 
# ifcopenshell.api: layer.edit_layer
>>> {"layer": {"cast_type": "entity_instance", "value": 6946, "Name": "Plan"}, "attributes": "{'Name': 'Plan', 'Description': None, 'Identifier': None}"}
# operator: BIM_OT_assign_presentation_layer
# ifcopenshell.api: layer.assign_layer
>>> {"items": [{"cast_type": "entity_instance", "value": 5394, "Name": null}], "layer": {"cast_type": "entity_instance", "value": 6946, "Name": "Plan"}}
# operator: BIM_OT_unassign_presentation_layer
# ifcopenshell.api: layer.unassign_layer
>>> {"items": [{"cast_type": "entity_instance", "value": 5394, "Name": null}], "layer": {"cast_type": "entity_instance", "value": 6946, "Name": "Plan"}}
# operator: BIM_OT_assign_presentation_layer
# operator: BIM_OT_assign_presentation_layer
# operator: BIM_OT_remove_presentation_layer
# operator: BIM_OT_remove_presentation_layer
# operator: BIM_OT_remove_presentation_layer
last_error: Traceback (most recent call last):
  File "/home/john/.config/blender/4.1/scripts/addons/blenderbim/bim/ifc.py", line 360, in execute_ifc_operator
    result = getattr(operator, "_execute")(context)
[...]

PR #8387, opened 2026-07-08
layer: never persist empty (schema-invalid) presentation layers

BIMvoice:
### Problem
IfcPresentationLayerAssignment.AssignedItems is a required aggregate SET [1:?] (confirmed via schema introspection in IFC4, IFC4X3 and IFC2X3), so a layer with no assigned items is schema invalid; ifcopenshell.validate flags it with "Attribute not optional".

ifcopenshell.api.layer.add_layer deliberately creates a layer with no items, because Bonsai's add-then-assign workflow populates it afterwards. But a layer that is created and never assigned (or otherwise left empty) stays as an invalid entity and is written to disk verbatim on save. unassign_layer already removes a layer when its last item is unassigned; the create-empty / never-assigned case was the remaining gap.

This is a self-identified data-validity fix (no issue number), found while testing the presentation-layer crash fix (#4934).

### Fix
- Add ifcopenshell.api.layer.remove_empty_layers(file), which removes every IfcPresentationLayerAssignment with no AssignedItems.
- Call it in Bonsai's save (ExportIFC._execute) right before export, and report how many empty layers were dropped.

This mirrors unassign_layer's existing "empty means remove to keep IFC valid" behaviour and applies it at the persistence boundary. The add-then-assign UX is untouched: layers stay empty in memory for the whole editing session, and only truly empty layers are pruned, only at write time.

### Evidence
- Empty layer: validate reports 1 error before, 0 after pruning (IFC4 and IFC4X3), layer removed.
- Normal add-then-assign: nothing removed, layer kept with its items, valid.
- Mixed populated + empty: only the empty one removed, valid.
- New tests in test/api/layer/test_remove_empty_layers.py (IFC4 and IFC2X3).
- Verified in a live Blender session: creating unassigned layers and saving reports "removed N empty presentation layer(s)" and the file stays valid.

### Note
An empty presentation layer cannot be represented in a schema valid file, so it cannot survive a save and reload under any correct fix. An alternative that surfaces validity earlier would be to also assign a new layer to the active object's representation on creation when possible; this PR keeps the change minimal (prune only).

### DIFF
--- a/src/bonsai/bonsai/bim/module/project/operator.py
+++ b/src/bonsai/bonsai/bim/module/project/operator.py
@@ -35,6 +35,7 @@ import bpy
 import ifcopenshell
 import ifcopenshell.api.attribute
 import ifcopenshell.api.document
+import ifcopenshell.api.layer
 import ifcopenshell.api.nest
 import ifcopenshell.api.project
 import ifcopenshell.api.root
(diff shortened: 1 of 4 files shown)


Moult:
This looks like a bandaid solution. A better solution is to change the UX to make the creation of invalid IFC impossible (a similar change was done for adding properties). However if AI creates this, I think it might be larger than what I can safely read in a PR. Maybe I'm wrong, so happy for AI to have a shot at this, but otherwise maybe it's good to create a tracking bug which includes all these "bigger, more fundamental design decision" improvements that need a human in the loop to develop.

PR #8389, opened 2026-07-08
Quantify doors and windows from IFC attributes on both take-off engines (supersedes #4537)

### LINKED ISSUE #4537: Wrong surface area calculations
Hi!
I noticed that BlenderBIM incorrectly calculates the surface area of ​​inserted doors and windows. In this example, the door should have an area of ​​1,971 m2 and the window should have an area of ​​3.76 m2.
Unless BlenderBIM is supposed to count these surfaces differently?
issue1
issue2

BIMvoice:
Reported in #4537 by @daniels0401: door and window surface areas are wrong. Running Perform Quantity Take-off on a real Bonsai IfcDoor also reported it "not quantified", producing no Qto_DoorBaseQuantities, while a comparable IfcWindow quantified fine. The door/window base quantities (Area, Width, Height, Perimeter) come from OverallWidth/OverallHeight and need no geometry, so the door should always quantify. This PR supersedes #4537.

Two root causes, both fixed:

1. Blender engine. Blender.calculate skipped any element whose tool.Ifc.get_object returned None, so a door with no Blender object (its body did not load a mesh, the object was removed, or the element came from a partially loaded/linked file) was dropped before any calculator ran. It now uses the IFC element itself as the calculation target when there is no object. The door/window attribute calculators accept a Blender object, an IFC element, or None, and only fall back to geometry when a real object is available. Geometry-only calculators raise on a non-object and are caught, so nothing crashes and every other element behaves exactly as before.

2. IfcOpenShell engine (the more common path). Its rule sets had the door quantities fully null and the window Area as an inflated max-side bounding box with a null perimeter. This PR adds attribute based calculators (get_lining_area = OverallWidth * OverallHeight, get_overall_width, get_overall_height, get_lining_perimeter = 2 * (width + height)), rescaled from project units to SI, requiring no geometry, and points IfcDoor/IfcWindow in both IFC4QtoBaseQuantities.json and IFC4X3QtoBaseQuantities.json at them, including the previously missing Perimeter.

This produces the complete Qto_Door/WindowBaseQuantities set (Width, Height, Perimeter, Area) that buildingSMART defines for the outer lining.

Verification, driven headless:
- Real door (OverallWidth 0.9, OverallHeight 2.0), both engines, with and without a Blender object: Area 1.80, Width 0.9, Height 2.0, Perimeter 5.8.
- 0.6 by 0.9 window on the IfcOpenShell engine: Area 0.54, Perimeter 3.0 (previously inflated / null); correct scaling in millimetre files.
- Wall/beam/slab geometry quantities unchanged; ifcopenshell.validate clean on written quantities.
- Confirmed in a live Blender session against a real door.

Fixes #4537

### DIFF
--- a/src/bonsai/bonsai/bim/module/qto/calculator.py
+++ b/src/bonsai/bonsai/bim/module/qto/calculator.py
@@ -25,6 +25,7 @@ import ifc5d.qto
 import ifcopenshell
 import ifcopenshell.geom
 import ifcopenshell.util.element
+import ifcopenshell.util.unit
 import mathutils
 from mathutils import Matrix, Vector
 from mathutils.bvhtree import BVHTree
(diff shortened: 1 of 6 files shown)


BIMvoice:
Rebased on current v0.8.0; the only conflict was the union with the new footing QTO functions in qto.py, both feature sets kept. ifc5d tests pass with no new failures.

PR #8392, opened 2026-07-08
Bonsai: fix deleting IfcIndexedPolyCurve from a GeometricSet representation

### LINKED ISSUE #6591: Bonsai: can't delete IfcIndexed PolyCurve from Plan-Body-PLAN_VIEW product representation
### Bug Description

To reproduce:
1. open prova.ifc;
2. select the Basin (SanitaryTerminal)
3. activate Plan-Body-PLAN_VIEW representation
4. TAB to edit
5. select one of the IfcIndexedPolyCurve
6. delete with x or Del
7. the item is deleted
8. TAB to exit edit mode
9. the item is back

### Attachments

[prova.ifc.txt](https://github.com/user-attachments/files/19838715/prova.ifc.txt)

### Debug and Error Output

No error 

### Debug and Error Output
Blender 4.4.1 Bonsai 0.8.3-alpha250121

BIMvoice:
Deleting a representation item that lives inside an IfcGeometricSet (for example an IfcIndexedPolyCurve inside an IfcGeometricCurveSet of a Plan/Body/PLAN_VIEW representation) did not persist: the item was visually removed in item edit mode but returned on exiting edit mode.

Root cause: Geometry.remove_representation_item resolved the containing representation only when the item's inverse was an IfcShapeRepresentation (or IfcBooleanResult). For items nested in an IfcGeometricSet the inverse is the set itself, so the item was never detached and remove_deep2 could not delete it while the set still referenced it. It was re-imported on reload.

Fix: also handle IfcGeometricSet inverses by removing the item from .Elements. If that would empty the set, remove the whole set instead, which cascades to the item.

Verified against the reporter's prova.ifc: the targeted IfcIndexedPolyCurve is removed from the IfcGeometricCurveSet and stays removed after the product representation is re-resolved. Also confirmed in a live Blender session (deleting a curve from the basin's plan symbol now persists), with the 3D model-view representation untouched and a clean ifcopenshell.validate.

Thanks to @carlopav for the clear steps and the sample file.

Fixes #6591

### DIFF
--- a/src/bonsai/bonsai/tool/geometry.py
+++ b/src/bonsai/bonsai/tool/geometry.py
@@ -1775,7 +1775,7 @@ class Geometry(bonsai.core.tool.Geometry):
                 else:
                     representation = inverse
             elif inverse.is_a("IfcGeometricSet"):
-                # The item is aggregated via .Elements and must be pulled out or it reappears (#6591).
+                # item sits in the set's .Elements; pull it out or it reappears on reload (#6591)
                 geometric_sets.append(inverse)
             elif inverse.is_a("IfcBooleanResult"):
                 if inverse.SecondOperand == representation_item:
@@ -1819,7 +1819,7 @@ class Geometry(bonsai.core.tool.Geometry):
             if new_elements:
                 geometric_set.Elements = new_elements
             else:
-                # Removing the set's last element is invalid, so drop the whole set.
+                # dropping the last element is invalid, so remove the whole set
                 cls.remove_representation_item(geometric_set, element)
                 return
 
--- a/src/bonsai/bonsai/tool/geometry.py
+++ b/src/bonsai/bonsai/tool/geometry.py
@@ -1766,6 +1766,7 @@ class Geometry(bonsai.core.tool.Geometry):
         [consider_inverses.append(texture := t) for t in getattr(representation_item, "HasTextures", [])]
 
         representation = None
+        geometric_sets: list[ifcopenshell.entity_instance] = []
         boolean_results_to_remove: set[ifcopenshell.entity_instance] = set()
         for inverse in ifc_file.get_inverse(representation_item):
             if inverse.is_a("IfcShapeRepresentation"):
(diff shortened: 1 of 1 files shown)


carlopav:
@BIMvoice I confirm this fixes https://github.com/IfcOpenShell/IfcOpenShell/issues/6591
Thanks!
Perhaps it's not in the stile of bonsai to have all those comments, but it's good for reviewing.

BIMvoice:
Thanks for confirming @carlopav, and glad it works for you. You are right about the comments too. That is denser than Bonsai's usual style, so I will trim them back to match while keeping just enough of the reasoning. Appreciate you taking the time to review.

Moult:
I think this looks OK but first can you please show me (via links) the relevant sections of the code which causes us to pull in individual items of a set in the first place? Because mentally I recall the it being "tab loads items of a shaperepresentation", not so much geometric set (though I can imagine it being an edge case).

BIMvoice:
Your mental model is right — there's no separate "load items of a set" mechanism, and the PR's own framing is a bit narrower than it reads.

Tab-into-edit-mode goes through `ImportRepresentationItems._execute` (`bim/module/geometry/operator.py`), which reads item ids off mesh custom attrs (`ios_item_ids` etc.) and does `tool.Ifc.get().by_id(item_id)` per id. Those ids are assigned generically by the C++ geometry mapper — `process_mapping` in `src/ifcgeom/mapping/mapping.h` tags `item->instance = inst` for every mapped `IfcRepresentationItem`, whatever it is.

The reason sets show up here at all: `IfcRepresentation.cpp` maps `Items()` via `map_to_collection(...)`, and `IfcGeometricSet.cpp` maps `Elements()` through the *exact same* `map_to_collection` helper:

```cpp
taxonomy::ptr mapping::map_impl(const IfcSchema::IfcGeometricSet* inst) {
    return map_to_collection(this, inst->Elements());
}
```

So when an `IfcShapeRepresentation.Items` contains an `IfcGeometricCurveSet` (common on Plan/annotation "symbol" bodies), each curve inside it gets individually mapped and tagged with its own instance id, same as any top-level item — the set entity itself is never a leaf and never gets tagged. Tab-into-edit-mode then naturally creates one item-object per curve, via the same generic path as any other representation item. It's "tab loads items of a shaperepresentation" all the way through; sets just fall out of that recursion rather than needing their own mechanism.

The gap this PR actually closes is narrower and one-sided: `remove_representation_item`'s *deletion* path didn't walk that one extra inverse hop (`IfcGeometricSet`) that the import side already handles transparently, so a deleted `IfcIndexedPolyCurve` inside a set stayed in `.Elements` and reappeared (#6591). Happy to retitle/reword the PR description to say that more precisely if useful — "individual items of a set" as written could read like a separate import mechanism when it isn't one.

carlopav:
@BIMvoice , so can you propose a solution to the issue #6591 with a more general approach, less edge case specific? You can make a new pr to not mess this one and link this new pr in a comment here?
Sorry to take advantage of your ai :)

BIMvoice:
@carlopav done — general solution in #8798, this PR left untouched as you suggested.

Rather than adding a branch per container class, it treats any inverse that is itself a representation item as a container and clears the item from whichever attribute references it, found by attribute index. Investigating properly turned up two more broken families beyond the one this PR patched: `IfcShellBasedSurfaceModel.SbsmBoundary` and `IfcFaceBasedSurfaceModel.FbsmFaces` leak the deleted item back on reload the same way, confirmed against an unpatched baseline. #8798 supersedes this PR entirely, so closing this one.

Issue #414, opened 2018-07-11
Instance ownership not automatically propagated (anymore?)


aothms:
See https://sourceforge.net/p/ifcopenshell/discussion/1782717/thread/a9399540/?limit=25#0f03

PR #8406, opened 2026-07-09
Fix #486: auto-add cross-file instances on attribute assignment

### LINKED ISSUE #486: ifcopenshell.geom.tesselate instances not emitted when added as representation
See https://sourceforge.net/p/ifcopenshell/discussion/1782717/thread/9c6995ab/

BIMvoice:
Fixes #486.

Assigning an instance from a different ifcopenshell file to an attribute left a dangling reference: the foreign instance was never added to the target file, so on save its id was referenced but not emitted and the subgraph was silently dropped. The attribute setter now brings any foreign instance, including those nested in aggregates, into the target file via file.add before writing the reference, matching the semantics a user got by calling add manually. Same file, scalar and None assignments are unchanged.

Verified with real ifcopenshell across single, aggregate and nested cases; existing entity_instance and file tests pass.

### DIFF
--- a/src/ifcparse/IfcParse.cpp
+++ b/src/ifcparse/IfcParse.cpp
@@ -1107,8 +1107,17 @@ class apply_individual_instance_visitor {
 // duplicates the subgraph nor recurses. Same file and unowned instances are
 // returned unchanged. This folds the manual file.add step users previously had
 // to perform before the assignment into the assignment itself.
+//
+// The adoption is restricted to instances whose schema matches the target
+// file's. IfcFile::addEntity throws when the schemas differ, and a cross-schema
+// assignment used to be permitted (the pointer was simply stored). Callers such
+// as the Python owner-history helpers still rely on assigning an instance that
+// happens to originate from a file of another schema, so preserve the previous
+// non-adopting, non-throwing behaviour in that case rather than aborting the
+// assignment.
 static IfcUtil::IfcBaseClass* adopt_if_foreign(IfcParse::IfcFile* file, IfcUtil::IfcBaseClass* instance) {
-    if (instance != nullptr && file != nullptr && instance->file_ != nullptr && instance->file_ != file) {
+    if (instance != nullptr && file != nullptr && instance->file_ != nullptr && instance->file_ != file &&
+        instance->declaration().schema() == file->schema()) {
         return file->addEntity(instance);
     }
     return instance;
--- a/src/ifcparse/IfcParse.cpp
+++ b/src/ifcparse/IfcParse.cpp
@@ -1099,25 +1099,6 @@ class apply_individual_instance_visitor {
     };
 };
 
-// #486 An instance assigned as an attribute value may belong to another file.
-// Bring such a foreign instance, together with its forward references, into the
-// target file so the written reference resolves on serialization instead of
-// dangling. IfcFile::addEntity is idempotent through entity_file_map_ and
-// returns the instance owned by the target file, so repeated assignment neither
-// duplicates the subgraph nor recurses. Same file and unowned instances are
-// returned unchanged. This folds the manual file.add step users previously had
-// to perform before the assignment into the assignment itself.
-//
-// IfcFile::addEntity throws when the instance belongs to a file of a different
-// schema. Mixing schemas within a single file is never valid, so let that error
-// surface instead of silently storing a dangling cross-schema reference.
-static IfcUtil::IfcBaseClass* adopt_if_foreign(IfcParse::IfcFile* file, IfcUtil::IfcBaseClass* instance) {
-    if (instance != nullptr && file != nullptr && instance->file_ != nullptr && instance->file_ != file) {
-        return file->addEntity(instance);
-    }
-    return instance;
-}
-
 template <typename T>
 typename std::enable_if<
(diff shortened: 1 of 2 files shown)


aothms:
Please fix this on the C++ side instead

BIMvoice:
Moved to the C++ side as you asked. The fix now lives in the core at `IfcBaseClass::set_attribute_value`, the single template the three instance bearing setters funnel through. When an assigned attribute value is an instance whose file differs from the target, it is adopted via `IfcFile::addEntity` before the reference is written, the same routine users invoked manually with `file.add`: it depth first traverses forward references, remaps ids, and is idempotent through `entity_file_map_`, so repeated assignment neither duplicates the subgraph nor recurses. Same file and unowned values pass through, so existing behaviour is unchanged, and `validate_shape` and the generic path are untouched. The Python shim is dropped, so this is now a pure core change. A bare `product.Representation = rep`, or any single or aggregate cross file assignment, now resolves on serialization instead of writing a dangling id. I could not build the kernel in my environment, so it is not build tested yet, happy to iterate if you spot anything on a build.

BIMvoice:
Build- and behavior-tested the C++ rework locally (OCC 7.9.2, Boost 1.90, Python wrapper), since the earlier note said it was not build-tested. Results:

- Compiles clean (only the pre-existing sign-compare / unused-value warnings).
- #486 repro, headless: before the patch, assigning file B's IfcMaterial to A's IfcMaterialLayer.Material leaves A with 0 materials and serializes a dangling reference (the Material slot points at the layer's own id). After the patch the material is adopted into A (count 1, layer.Material.file is A) and A round-trips with the correct reference.
- Single-pointer, aggregate (IfcMaterialLayerSet.MaterialLayers), and forward-ref subgraph cases all adopt and round-trip with no dangling refs (by_id resolves for every instance after reload).
- Same-file assignment is unchanged (no duplication, id preserved).

One benign observation, not a blocker and not in the modified code path: after adopting a foreign subgraph via the create_entity(Attr=foreign) kwargs path, the in-memory read-back still reports .file is B until reload, though the stored data and serialization are correct and owned by A. Via the direct inst.Attr = foreign path (the set_attribute_value code this PR changes) read-back is fully correct immediately. So it looks like a create_entity Python-side retention quirk that self-heals on reload, separate from this fix.

BIMvoice:
Pushed a follow-up that fixes the CI regression this PR introduced (test_facet.py::TestClassification::test_filtering_using_a_classification_facet).

Root cause: the new adopt-on-assignment called IfcFile::addEntity unconditionally, which throws on a schema mismatch. Assigning a cross-schema instance was previously permitted (the pointer was just stored), and the owner-history helpers still do that, so the unconditional adoption regressed them. The failure only shows in the full test_facet.py run because an earlier test leaves an IFC2X3 user in the owner-history global, which the later IFC4 classification test then assigns.

Fix: adopt only when the instance's schema matches the target file, otherwise fall back to the previous non-adopting, non-throwing behaviour. #486 (a same-schema foreign library instance) still adopts and round-trips with no dangling reference. Verified locally (OCC 7.9.2, 2x3;4 wrapper): full test_facet.py 23/23, and the classification / owner / material / file suites stay green.

aothms:
> otherwise fall back to the previous non-adopting, non-throwing behaviour. 

This is not useful. Just always throw when mixing schemas.

BIMvoice:
Done. Dropped the fallback so a cross-schema assignment now always throws via addEntity, as you asked. The regression that fallback was papering over was a test leak, not a real need: TestEntity.test_ifc2x3_occurrence_type_mapping monkeypatched the module-global owner-history settings (get_user / get_application) with IFC2X3 instances and never restored them, so an IFC2X3 user leaked into a later IFC4 test. I wrapped that in try/finally to restore the settings.

Verified: test_facet.py passes in full (23), the #486 same-schema adoption still adopts and round-trips with no dangling reference, and a genuine cross-schema assignment now raises "Unable to add instance from IFC2X3 schema to file with IFC4 schema".

src/ifcparse/IfcParse.cpp
```diff
@@ -1099,6 +1099,25 @@ class apply_individual_instance_visitor {
...
+// dangling. IfcFile::addEntity is idempotent through entity_file_map_ and
+// returns the instance owned by the target file, so repeated assignment neither
+// duplicates the subgraph nor recurses. Same file and unowned instances are
+// returned unchanged. This folds the manual file.add step users previously had
+// to perform before the assignment into the assignment itself.
+//
+// IfcFile::addEntity throws when the instance belongs to a file of a different
+// schema. Mixing schemas within a single file is never valid, so let that error
+// surface instead of silently storing a dangling cross-schema reference.
+static IfcUtil::IfcBaseClass* adopt_if_foreign(IfcParse::IfcFile* file, IfcUtil::IfcBaseClass* instance) {
```
aothms: please make this an inline lambda

BIMvoice: Done, made it an inline lambda inside set_attribute_value next to the call sites. Behaviour unchanged, still compiles clean.

BIMvoice:
Closing this as superseded by #8776.

Both patch `set_attribute_value` to adopt an instance from another file. The difference matters: this PR's guard only fires when `instance->file_ != nullptr && instance->file_ != file`, so a genuinely unowned instance (`file_ == nullptr`), which is the exact scenario in #414, is never adopted. #8776 fires whenever `instance->file_ != file`, covering both the unowned case and the foreign-file case, and adds regression tests in `test_file.py`.

The diagnosis in this PR is carried over into #8776, which remains open for review.

PR #8411, opened 2026-07-09
docs(bonsai): make the door move/regen guidance consistent (#4950)

### LINKED ISSUE #4950: [bug] Apply void operation disregards object's flipped state
When applying a void of a flipped object (particularly doors and windows) to a wall, the operation does not consider the flipped state of the object. This results in the object reverting to its original, non-flipped orientation after the void is applied.

**Steps to Reproduce:**
1. Create a wall.
2. Add a door to the wall.
3. Select only the door.
4. Flip the door using the Shift+F shortcut.
5. Select both the door and the wall.
6. Apply void (Shift+O).
7. Observe that the door has reverted to its original, non-flipped state.

**Expected Behavior:**
The apply void operation should maintain the door's current orientation, including its flipped state.

**Actual Behavior:**
The apply void operation resets the door to its original orientation, disregarding its flipped state.

**Additional Information:**
- The problem is related to the apply void functionality not considering the flipped state of objects.
- This is separate from #4949

BIMvoice:
Closes #4950.

The door authoring guide gave conflicting instructions about recalculating a wall opening after moving a door. Several places told the user to select the wall and press Regen, but a wall level Regen does not reposition the opening. Selecting the door and pressing Regen, or using Apply Void, is what moves the opening. Vanuan pointed this out in the issue.

## Changes

* Modifying Doors note: select the door, not the wall, and note that resizing a door updates the opening automatically. Fixed the "achive" typo.
* Placing a Door, step 4: select the door for the opening update, and removed the outdated claim that moving a door does not change the IFC model. Moving the door does update its placement; only the wall opening needs a Regen.
* Additional Notes: replaced the blanket "avoid Apply Void" line, which contradicted the Modifying Doors section, with guidance on when to actually use it.

Documentation only, no code changes.

### DIFF
--- a/src/bonsai/docs/guides/authoring/basic_modeling/door.rst
+++ b/src/bonsai/docs/guides/authoring/basic_modeling/door.rst
@@ -49,16 +49,15 @@ If you forgot to select the wall before placing the door, you'll need to manuall
 3. Adjust Door Position (if needed):
    - With the door selected, use Blender's move tools to adjust its position along the wall.
 
-4. Regenerate the Wall Geometry:
-   - Select the wall.
-   - Press Shift+G to regenerate the wall geometry, incorporating the door opening.
+4. Update the Wall Opening:
+   - Select the door.
+   - Press Shift+G to move the wall opening to the door's new position.
 
    .. note::
-      This step ensures the wall geometry is updated to include the door opening.
-
-      Moving the Door using Blender tools doesn't actually change the IFC model.
-      Future versions need to improve UX in that regard.
-      Synchronisation between the Blender scene and IFC model is an issue that has the highest priority.
+      Moving the door with Blender's tools repositions the door, but the
+      opening in the wall does not follow on its own. Selecting the door and
+      pressing Regen repositions the opening to match. If Regen does not
+      update the opening, use Apply Void (Shift+O) instead.
 
 5. Change Door Swing Direction:
    - Select only the door.
@@ -81,10 +80,11 @@ Additional Notes
 
 - The Shift+F shortcut for flipping the door is not visible in the Door tool interface.
   This functionality may be added as a visible button in future updates.
-- Always use Shift+G (Regenerate) after making changes
+- Use Shift+G (Regenerate) after making changes
   to ensure the wall and door geometries are correctly updated.
-- Avoid using Shift+O (Apply Void) as it may cause issues
-  with the door's orientation.
+- Use Shift+O (Apply Void) when a door has no void relation with its wall,
+  or when Regen does not update the opening. Apply it before making further
+  changes to the door, as noted above.
 
 
 Modifying Doors
@@ -95,21 +95,21 @@ Modifying Doors
 
 You can modify doors using various tools:
 
-- Resize: Adjust the width and height parameters in the top bar.
+- Resize: Adjust the width and height parameters in the top bar. The opening in the wall updates automatically.
 - Move: Use Blender's standard move tools to reposition the door.
 - Flip: Change the opening direction using the flip tool (Shift+F).
 
 .. note::
-   After moving a door, you need to recalculate the void in the wall. To do this:
+   After moving a door, the opening in the wall does not follow on its own, so you need to recalculate it. To do this:
 
-- Select the wall containing the moved door.
-- Click on the "Regen" (:kbd:`Shift` + :kbd:`G`) button in the top bar or use the appropriate shortcut.
-- This ensures that the opening in the wall is correctly positioned after moving the door.
+   - Select the door you moved.
+   - Click the "Regen" (:kbd:`Shift` + :kbd:`G`) button in the top bar, or use the shortcut.
+   - The wall opening moves to match the door's new position.
 
-If Regen operation doesn't achive the required result, use Apply Void function:
+   If Regen does not achieve the required result, use Apply Void instead:
 
-- Click on the "Apply Void" (:kbd:`Shift` + :kbd:`O`) button in the top bar or use the appropriate shortcut.
-- This ensures that the opening is linked to the wall.
+   - Click the "Apply Void" (:kbd:`Shift` + :kbd:`O`) button in the top bar, or use the shortcut.
+   - This reconnects the opening to the wall.
 
 
 Door Properties
--- a/src/bonsai/docs/guides/authoring/basic_modeling/door.rst
+++ b/src/bonsai/docs/guides/authoring/basic_modeling/door.rst
@@ -111,6 +111,12 @@ You can modify doors using various tools:
    - Click the "Apply Void" (:kbd:`Shift` + :kbd:`O`) button in the top bar, or use the shortcut.
    - This reconnects the opening to the wall.
 
+   How it works: Regen on the door recalculates the door's fill, which repositions
+   the opening in the host wall and recuts it, so an action on the door is what
+   updates the wall geometry. Regen on the wall only re-syncs the opening while the
+   door is still in a freshly moved state, which is why selecting the door is the
+   reliable choice.
+
 
 Door Properties
 ^^^^^^^^^^^^^^^


Vanuan:
This is arguably more confusing. So regenerating the door updates the wall geometry rather than the door's? What's going on under the hood? Does regen recreate the void and applies it to the related wall?

OTOH, from the user perspective, regen is implementation detail to improve performance.

Vanuan:
So the documentation is both too technical for UX and not technical enough to understand what's going on.

Maybe we need both under the hood and UX friendly sections.
<Details>

> 
> ## What Happens Under the Hood
> 
> ### When you select the door and press Regen
> 
> The `RecalculateFill` operator executes, which:
> 
> 1. Gets the opening/void element that the door fills via `FillsVoids` relationship 
> 2. Updates the opening/void placement to match the door's current `matrix_world`
> 3. Switches the wall's representation to regenerate its geometry with the new opening position
> 
> ### When you select the wall and press Regen
> 
> The `regenerate_wall_representation` API function executes, which:
> 
> 1. Regenerates the wall's body geometry based on its axis, layers, and connections
> 2. Preserves existing manual booleans (including openings) but does not update their placement
> 3. Updates the wall's object placement to match the axis line start point
> 
> The key difference is that `RecalculateFill` actively updates the opening's placement before regenerating the wall, while `regenerate_wall_representation` uses the existing opening placement as-is.
> 
> ## Apply Void Function
> 
> The `Apply Void` function (Shift+O) creates or updates the void relationship between a door/window and a wall. This is handled by the `FilledOpeningGenerator.generate()` method which:
> 
> 1. Creates an `IfcOpeningElement` if one doesn't exist
> 2. Adds the opening as a feature to the wall and the door as a filling to the opening
> 3. Switches the wall's representation to incorporate the opening
> 

</Details>
Also door, opening and void are used interchangeably which is confusing. As understand fillsvoid relationship is opening, door is opening filling, while void is independently movable object?

Vanuan:
Screenshot_20260709-144421

I believe regen could just sync up everything when user presses regen when any of the wall, opening or filling is pressed.

Here's a draft of how both sections can be written. Feels strange that in some cases you need apply void, some cases regen and there's a difference between flipping and moving the door. It's also unclear what's wrong with applying void multiple times. Does it  inherently create duplicate openings or is it implementation detail?

---

<details>

Door
====

A door is a wall opening that allows passage between spaces, controlling access for people and goods.

In IFC, it is represented by the ``IfcDoor`` entity, which is a subtype of ``IfcBuildingElement``.
A door always requires a host wall and a corresponding ``IfcOpeningElement`` cut into that wall.

.. note::
   This guide applies equally to **Windows** (``IfcWindow``), as they share the exact same opening mechanics.

User Guide – Adding and Modifying Doors
---------------------------------------

This section focuses on *what* you need to do, hiding unnecessary IFC internals.

Adding a Door
~~~~~~~~~~~~~

1. **Select the wall** where you want to place the door.
2. (Optional) Set the 3D cursor on the wall at the desired position.
3. Click the **"Create Door"** tool in the Bonsai toolbar.
4. In the top bar, edit the type name (e.g., change ``[TYPEX]`` to ``DOOR001``) and click **"+ Add IfcDoorType"**.
5. Click **"Add"** (or press :kbd:`Shift+A`) to create the door.
6. Adjust the door's **width** and **height** using the parameters in the top bar.

.. tip::
   If you forgot to select the wall first, the relationship to the wall will be missing.
   In that case, **select both the wall and the door** and click **"Apply Void"** (:kbd:`Shift+O`) **once** to create the opening.

Modifying a Door – Which Button Do I Press?
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

The most common source of confusion is knowing when to press **Regen** vs **Apply Void**.
The rule depends on *what* you changed about the door:

.. list-table:: Decision Matrix for Modifying Doors
   :header-rows: 1
   :widths: 25 30 25 20

   * - What you did to the Door
     - What changed in IFC
     - Action to Update the Wall
     - Select this
   * - **Moved** it (translated/rotated)
     - Door's Placement (Matrix)
     - Press **Regen** (:kbd:`Shift+G`)
     - **Door** only
   * - **Flipped** the swing (:kbd:`Shift+F`) or changed "Operation Type"
     - Door's Local Shape / Profile
     - Press **Apply Void** (:kbd:`Shift+O`)
     - **Wall + Door**
   * - **Resized** Width/Height (parametric)
     - Door's Dimensions
     - Usually handled by the door's own parametric update. If the wall cutout looks out of sync, press Apply Void once.
     - (Wall + Door if needed)
   * - Changed material or thickness
     - Door's Visual representation
     - No action needed for the wall opening.

.. important::
   - **Regen**, when the **door** is selected (:kbd:`Shift+G`), updates the **opening's placement** to match the door and then refreshes the wall. It does **not** change the opening's shape.
   - **Regen**, when the **wall** is selected, only refreshes the wall's own geometry — it leaves existing opening placements untouched.
   - **Apply Void** (:kbd:`Shift+O`) recreates the opening's **shape** from scratch based on the door's current geometry.
   - If you flip a door and press **Regen**, the opening will be repositioned correctly, but its shape will still reflect the old swing direction. You must use **Apply Void** after flipping.

Step-by-Step: Moving a Door
~~~~~~~~~~~~~~~~~~~~~~~~~~~

1. Select the door and use Blender's move tools to reposition it. This updates the door's own IFC placement immediately.
2. **With the door still selected**, press :kbd:`Shift+G` (Regen).
3. This copies the door's new placement onto its opening, then regenerates the wall so the hole moves to match.

.. note::
   Selecting the **wall** instead and pressing :kbd:`Shift+G` will **not** move the opening — it only refreshes the wall mesh using the opening's existing (now stale) position. Always select the door after moving it.

Step-by-Step: Flipping a Door (Changing Swing)
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

1. Select only the door.
2. Press :kbd:`Shift+F` to flip the door by 180°.
3. (Optional) Go to ``Scene Properties > Geometry and Materials`` and change the **"Operation Type"** (e.g., to ``SINGLE_SWING_RIGHT``).
4. **Select both the wall and the door**.
5. Press :kbd:`Shift+O` (Apply Void) to recreate the opening shape to match the new swing.
6. The wall geometry updates with the correct cutout.

.. warning::
   **Do not press Apply Void multiple times on the same door!**
   Each press creates a brand new ``IfcOpeningElement``. Duplicate openings overlapping in the same wall will break the Boolean operation, resulting in missing wall faces or corrupted geometry. If you need to re-apply the void, remove the existing opening first (see below) before pressing Apply Void again.

Deleting a Door and Removing the Opening
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Deleting the door object **does not** automatically delete the hole in the wall — the opening is a separate entity.

**Primary method (recommended):**

1. Delete the door object (e.g., press :kbd:`X`).
2. Select the wall that contained the door.
3. Open the **Voids** panel (in the Geometric Relationships tab).
4. Click **"Remove Opening"** next to the orphaned opening.
5. The wall geometry regenerates automatically without the hole.

**Manual method (fallback):**

If the Voids panel doesn't show the opening (e.g., due to a UI issue), you can reproduce the same result by hand. Note that simply deleting the ``IfcOpeningElement`` object in Blender is **not** enough — that removes the mesh, but leaves the wall's ``IfcRelVoidsElement`` relationship pointing at a now-missing entity, which can corrupt the model. The relationship has to be removed explicitly:

1. Delete the door object.
2. Enable **"Show Voids"** (in the Create Wall tool, or via the Outliner) to reveal the ``IfcOpeningElement``.
3. Select the ``IfcOpeningElement``.
4. In the Python Console, run:

   .. code:: python

      import ifcopenshell.api
      opening = tool.Ifc.get_entity(bpy.context.active_object)
      ifcopenshell.api.feature.remove_feature(tool.Ifc.get(), feature=opening)

   This removes the void relationship **and** deletes the opening entity in one call — no separate delete step needed.
5. Select the wall and press :kbd:`Shift+G` to regenerate it without the hole.

The "Remove Opening" button does exactly this in one step: it looks up the opening and its voided wall, unlinks the opening's Blender object, calls ``remove_feature`` to remove the relationship and the opening entity, and regenerates the wall's geometry.

---

Under the Hood – Technical Reference
------------------------------------

This section explains the IFC data model behind these operations for advanced users.

IFC Entities and Relationships
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Three distinct concepts are often used loosely — here they are precisely:

- **Filling** (``IfcDoor`` / ``IfcWindow``): the actual door or window object.
- **Opening** (``IfcOpeningElement``): the feature element representing the hole itself. This is a real entity in the model, not just a relationship.
- **Voided Element** (``IfcWall`` / ``IfcSlab``): the building element that gets cut.

"Void" is not a separate movable object — it refers to the relationship between the opening and the element it cuts (``IfcRelVoidsElement``). Likewise "filling" refers to the relationship between the opening and the door/window that occupies it (``IfcRelFillsElement``).

The relationships are directional:

::

    Filling (Door) ── FillsVoids ──> Opening ── VoidsElements ──> Voided Element (Wall)

.. mermaid::
   graph LR
       Door[IfcDoor / Filling] -->|FillsVoids| Opening[IfcOpeningElement]
       Opening -->|VoidsElements| Wall[IfcWall / Voided Element]

What "Regen" (Shift+G) Actually Does
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

When you select the **door** and press :kbd:`Shift+G`, the ``RecalculateFill`` operator executes:

1. Retrieves the associated ``IfcOpeningElement`` via the ``FillsVoids`` relationship.
2. Copies the door's current ``matrix_world`` (placement) onto the opening.
3. Switches the wall's representation to regenerate its geometry with the new opening position.

**Code Reference**: ``src/bonsai/bonsai/bim/module/model/opening.py`` (``RecalculateFill``, lines ~400–450)

If you select the **wall** and press :kbd:`Shift+G`, ``regenerate_wall_representation`` executes instead. It rebuilds the wall's body geometry from its axis, layers, and connections, but it explicitly **preserves the existing placement of any openings** rather than updating them.

What "Apply Void" (Shift+O) Actually Does
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

When you select the **wall and door** and press :kbd:`Shift+O`, ``FilledOpeningGenerator.generate()`` executes:

1. Creates a **new** ``IfcOpeningElement`` (or reuses an existing one, if found) and places it to match the door.
2. Adds the opening as a feature to the wall (``VoidsElements``).
3. Adds the door as a filling to the opening (``FillsVoids``).
4. Forces the wall to switch to a new representation that includes the newly created void.

**Code Reference**: ``src/bonsai/bonsai/bim/module/model/opening.py`` (``FilledOpeningGenerator.generate``, lines ~130–215)

Why Multiple Apply Voids Break Things
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

- Each call to ``Apply Void`` can create a new ``IfcOpeningElement``.
- The wall accumulates multiple ``HasOpenings`` relationships.
- During wall regeneration, the Boolean operation subtracts **all** overlapping voids simultaneously.
- If they occupy the same space, the Boolean subtraction fails or produces a non-manifold mesh.

**Best Practice**: If you need to recreate an opening shape (e.g., after a complex flip), **remove the existing opening first**, then use ``Apply Void`` **once**.

Removing an Opening Programmatically
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

To remove an opening cleanly via the API:

.. code:: python

   # Find the opening related to your door
   opening = ifcopenshell.util.element.get_filled_void(door)
   if opening:
       # This removes BOTH the void relationship (IfcRelVoidsElement)
       # and the opening entity itself — do not call root.remove_product
       # afterwards, or you'll try to delete an already-deleted entity.
       ifcopenshell.api.feature.remove_feature(file, feature=opening)

.. note::
   If the door/window filling still exists at this point, ``remove_feature`` will also remove its ``IfcRelFillsElement`` relationship, but it does **not** delete the filling object itself — that becomes orphaned and must be removed separately with ``ifcopenshell.api.root.remove_product(file, product=filling)``.

---

Troubleshooting Common Issues
-----------------------------

- **"I moved the door, pressed Regen, but the hole didn't move!"**
  - Ensure the **door** is selected, not the wall — a wall-level Regen leaves opening placements untouched.
  - Check if the door has a ``FillsVoids`` relationship. If not, use Apply Void first.

- **"The wall disappeared or has holes everywhere!"**
  - You likely pressed Apply Void multiple times, creating overlapping openings. Open the Voids panel, remove the duplicates, and press Apply Void **once**.

- **"Regen isn't updating the swing direction!"**
  - Regen only updates the opening's position, not its shape. Use Apply Void to change the hole's shape after a flip.

See Also
--------

- :doc:`creating_walls`
- :doc:`../advanced_modeling/material_assignment`
- :doc:`../technical/opening_relationships` (if applicable)

</details>

Vanuan:
I guess regen when wall is selected does close the original opening if that one got deleted, but it doesn't create a new opening in the new door position

BIMvoice:
Good questions, let me take the under the hood part first. Regen on the door runs Recalculate Fill: it repositions the opening element (the void) to the door's current placement and recuts the host wall, so the action starts from the door but the wall geometry changes because the void lives in the wall and the door only fills it. Regen on the wall runs Recalculate Walls, which resyncs a filling's opening only when that filling is in a moved (dirty) state (commit 9d2de117a). That is the split you are seeing.

On your specific observation: if the door was just moved, wall Regen does now move the opening. If the door placement was already committed while the opening is stale, which is your case and also what happens on builds before that fix, wall Regen closes the old opening but does not place a new one at the door, exactly as you describe. Selecting the door and Regen is reliable in both states because it recuts unconditionally, and selecting both the door and its wall and Apply Void also always works.

On the docs, fair point that it mixes levels. I kept the note as the plain what to click step and pointed it at selecting the door, since that is the one action that always works. I added a short how it works line so the mechanics are explicit without over teching the steps. And if the opening element was manually deleted, the fill relationship is gone, so Regen cannot recreate it, you re add it with Apply Void on the wall and door.

Vanuan:
Won't apply void create duplicate openings?

BIMvoice:
Good question, and no, it does not duplicate. When Apply Void (the filled opening generator) runs on a filling that already has an opening, it first removes the existing opening and then creates a fresh one at the door's current position. In `FilledOpeningGenerator.generate` (bonsai/bim/module/model/opening.py):

```python
if filling.FillsVoids:
    ifcopenshell.api.feature.remove_feature(
        tool.Ifc.get(), feature=filling.FillsVoids[0].RelatingOpeningElement
    )
```

So the void is repositioned by remove-then-recreate against the single filling relationship (a door has one FillsVoids), which is why you never end up with two openings for one door. Recalculate Fill (Regen on the door) uses the same path, so it also repositions rather than adds.

On your other point about the doc being both too technical for UX and not technical enough: that is fair. I am happy to split it into a short UX oriented section (what to click) and a collapsible "under the hood" section (opening removal plus recut) along the lines you drafted, if you think that is the right shape for this guide.

Vanuan:
To recap:

- Apply Void is safe - removes the existing opening before creating a new one, so it cannot create duplicates (opening.py#65-68)
- Regen on door is always reliable - recalculate fill unconditionally repositions the opening and recuts the wall (opening.py#400-450)
- Regen on wall is conditional - only works when the filling is in a "moved" state (fixed in commit 9d2de11)

So this makes #8431 a small push away from being complete plus some UX improvements on dirty state visibility. Regen on wall should recut the openings and track dirty state of the filling shape?

So current docs can be rewritten like this:

User guide (UX-focused) - ideally with screenshots:
- simple "what to click" instructions
- decision guide for when to use Regen vs Apply Void
- step-by-step workflows for common tasks
- no technical implementation details

Under the hood (technical reference) - could be extracted to separate page in corresponding reference section:
- IFC entity relationships
- How Regen and Apply Void actually work
- code references
- API usage examples

If #8431 implemented, the doc could be drastically simplified wrt regen and apply void.

Vanuan:
Please review #8509

BIMvoice:
Superseded by #8509 (Vanuan), which is a fuller rework of the same door.rst guide: it keeps the corrected Regen-vs-Apply-Void guidance from this PR and adds the UX / under-the-hood split, a decision matrix, and deletion steps. Closing this in favour of #8509. I reviewed and verified #8509's technical claims against the source.

PR #8420, opened 2026-07-09
Bonsai: close IfcPolyline loops by appending the closing edge (#8043)

### LINKED ISSUE #8043: Bug: IfcArbitraryProfileDefWithVoids hole disappears after TAB cycling
[hello_wall.ifc.txt](https://github.com/user-attachments/files/27485920/hello_wall.ifc.txt)

**Version** Tested with Bonsai latest (May 2025)  

**Steps to reproduce** 
1. Load IFC file containing IfcArbitraryProfileDefWithVoids profile 
2. Observe window/door with visible hole (correct on initial load)
3. Press TAB to cycle through subobjects (select the lining/lining component) 
4. Hole disappears, But the outline is still presen as a subobject.  
 
**Expected behavior** 
Hole remains visible after TAB cycling. 

The IfcArbitraryProfileDefWithVoids profile contains both outer curve and inner curves (voids) which should be preserved during representation rebuild.  

**Actual behavior** 
After TAB cycling, the representation rebuilds but the hole is no longer cut. The void inner curves are lost or not applied to the extrusion. A bounding box appears instead of the actual geometry.  

**Test file** 
Attached: hello_wall.ifc contains window components using IfcArbitraryProfileDefWithVoids for the lining profile with a hole.  

**Notes** 
- Initial render is correct. Bug triggers only on representation rebuild during subobject cycling. 
- The sub-subobject (the profile definition) displays correctly when inspected separately. 
- Issue occurs with the extruded geometry, not the profile definition itself.

BIMvoice:
Closes #8043.

`convert_curve_to_mesh` built the edge chain of a polyline with `extend`, then for a closed polyline overwrote the last edge with the closing edge instead of appending it. That discarded the final real segment, so every closed `IfcPolyline` loop came back one edge short and open. On the edit mode round trip the inner void loop of an `IfcArbitraryProfileDefWithVoids` was then lost or misclassified, and the profile was rewritten without its void, collapsing the extrusion to a bounding box.

This appends the closing edge instead, matching the `IfcIndexedPolyCurve` branch which already does this.

Verified live in headless Blender 5.1.2 on the reporter's sample: before, the Tab round trip produced four open (degree-1) vertices and re-exported the profile without its void; after, both loops close and the profile round-trips as `IfcArbitraryProfileDefWithVoids` with its inner void intact.

### DIFF
--- a/src/bonsai/bonsai/tool/model.py
+++ b/src/bonsai/bonsai/tool/model.py
@@ -617,7 +617,7 @@ class Model(bonsai.core.tool.Model):
 
             cls.edges.extend([(i, i + 1) for i in range(offset, len(cls.vertices) - 1)])
             if is_closed:
-                cls.edges[-1] = (len(cls.vertices) - 1, offset)  # Close the loop
+                cls.edges.append((len(cls.vertices) - 1, offset))  # Close the loop
 
         elif curve.is_a("IfcCompositeCurve"):
             # This is a first pass incomplete implementation only for simple polylines, and misses many details.


CyrilWaechter:
I encountered the issue in boundaries edition context. So I added your commit in my PR stack alongside with related regression test. See https://github.com/IfcOpenShell/IfcOpenShell/pull/9226. If it is OK for you we can keep the other PR which also contains tests.

BIMvoice:
Perfectly fine by me, and thank you for adding the test. #9226 is strictly better than this PR: same one-line fix, plus the regression test I should have written in the first place.

Closing this in favour of #9226. Attribution on the original commit is preserved there, which I appreciate.

Good catch that it also affects boundaries editing. I only saw it via #8043's polyline path, so that widens the case for it usefully.

PR #8424, opened 2026-07-09
IfcGeomTree: keep clash_intersection_many results in set_a, set_b order (#5357)

### LINKED ISSUE #5357: incorrect results from clash_intersection_many
I'm trying to clash file with walls, slabs(a) to file with spaces(b). The result is always supposed to have `result.a` as instance of  `IfcWall/IfcSlab` and `result.b` as instance of `IfcSpace`.
While majority of the result is correct, I noticed is that some clash result from `clash_intersection_many` is flipped - i.e `result.a` is supposed to be `result.b` and vice versa.

Steps to reproduce :
Unfortunately, I cannot share the ifc files. Here is the code for convinience
```python
import logging

import ifcopenshell
from ifcclash.ifcclash import Clasher, ClashSettings


def intersect_zones_elements(zones_file_path: str, elements_file_path: str,
                             tolerance: float,
                             ):
    clash_settings = ClashSettings()
    clash_settings.logger = logging.getLogger("ifcclash")
    c = Clasher(clash_settings)
    # https://docs.ifcopenshell.org/ifcopenshell-python/selector_syntax.html
    clash_set = {
        "mode": "intersection",
        "tolerance": tolerance,
        "check_all": False,
        "a": [
            {
                "selector": "IfcWall,IfcSlab",
                "mode": "i",
                "file": elements_file_path,
            }
        ],
        "b": [
            {
                "selector": "IfcSpace",
                "mode": "i",
                "file": zones_file_path,
            }
        ],
    }

    c.clash()
    c.tree = ifcopenshell.geom.tree()
    c.create_group("a")
    for source in clash_set["a"]:
        source["ifc"] = c.load_ifc(source["file"])
        c.add_collision_objects("a", source["ifc"], source.get("mode", None), source.get("selector", None))

    c.create_group("b")
    for source in clash_set["b"]:
        source["ifc"] = c.load_ifc(source["file"])
        c.add_collision_objects("b", source["ifc"], source.get("mode", None), source.get("selector", None))

    results = c.tree.clash_intersection_many(
        list(c.groups["a"]["elements"].values()),
        list(c.groups["b"]["elements"].values()),
        tolerance=clash_set["tolerance"],
        check_all=clash_set["check_all"],
    )

    mismatch_counter = 0
    for result in results:
        clash_type = ["protrusion", "pierce", "collision", "clearance"][result.clash_type]
        a_type, b_type, a_id, b_id = result.a.is_a(), result.b.is_a(), result.a.get_argument(0), result.b.get_argument(0)
        if a_type == "IfcSpace":
            mismatch_counter += 1
            print(f"{a_type} {b_type} {a_id} {b_id} {clash_type}")

    print(f"# of mismatch: {mismatch_counter}, tolerance: {tolerance}")


if __name__ == "__main__":
    _zones_file = "/path/to/zones.ifc"
    _elements_file = "/path/to/elements.ifc"
    intersect_zones_elements(_zones_file, _elements_file, 0.05)
```

BIMvoice:
Closes #5357.

`clash_intersection_many` runs a protrusion pass that calls `test_intersection(task.b, task.a, ...)` with its arguments swapped, because protrusion is directional and this pass tests whether the set_b element protrudes into the set_a element. `test_intersection` always stamps its first argument as `.a` and its second as `.b`, so the clash it returns has `.a` and `.b` swapped relative to the `set_a`/`set_b` the caller passed. When this pass produces the adopted result (for example the set_b element protrudes deeper, or is the only manifold one), the reported clash comes out flipped, so `result.a` is a set_b element (the `IfcSpace` in the report) instead of the wall or slab.

This swaps `intersection.a` and `intersection.b` back to canonical order right after the directional call, so `result.a` always refers to a set_a element and `result.b` to a set_b element. `distance` and `clash_type` are unaffected. `ifcquery/clash.py`, which reports `result.b` as the other element, relied on this order and is corrected by the same change.

Verification: the flip is live reproduced with a known-answer case (a small IfcSpace embedded in a larger IfcWall returns `result.a = IfcSpace` before the fix). The one line kernel change itself is not build tested, since it needs an OpenCascade rebuild. Happy to run the before and after on a build if you want that confirmation too.

### DIFF
--- a/src/ifcgeom/kernels/opencascade/IfcGeomTree.h
+++ b/src/ifcgeom/kernels/opencascade/IfcGeomTree.h
@@ -1109,6 +1109,12 @@ namespace IfcGeom {
                                 is_manifold = true;
                                 clash intersection = test_intersection(task.b, task.a, tolerance, check_all);
                                 if (intersection.clash_type != -1) {
+                                    // test_intersection() was called with the arguments swapped in
+                                    // order to test whether task.b protrudes into task.a, so it
+                                    // reports .a and .b in swapped order. Restore the canonical
+                                    // order so that result.a always refers to an element of set_a
+                                    // and result.b to an element of set_b.
+                                    std::swap(intersection.a, intersection.b);
                                     // Replace the clash result if any of these criteria apply:
                                     // - We don't have a clash yet
                                     // - Our previous clash is piercing, and our new one is a protrusion


BIMvoice:
Re-opened against `v0.9.0`, the current default branch, with RED/GREEN on a real build: #9390. Closing this one so there is a single PR to review.

PR #8433, opened 2026-07-10
ifcgeom: use the real extrusion depth for IfcSurfaceOfLinearExtrusion (#8013)

### LINKED ISSUE #8013: Blender freezes trying to open a file with `IfcSurfaceOfLinearExtrusion`
Opening up the following file in Bonsai, freezes blender. 
[surface_of_linear_extrusion.ifc.txt](https://github.com/user-attachments/files/27107500/surface_of_linear_extrusion.ifc.txt)

The following is AI's assessment....

The fix is a single-line change. Here's what was happening and why it mattered:

----------

**What `IfcSurfaceOfLinearExtrusion` represents**

It's a mathematical surface formed by sweeping a profile curve along a direction vector for a finite distance — like sliding a bent wire through space to trace out a ribbon or wall face. The IFC schema gives it three key inputs: the curve to sweep (`SweptCurve`), the direction to sweep along (`ExtrudedDirection`), and how far to sweep (`Depth`).

----------

**The bug**

The mapper in IfcSurfaceOfLinearExtrusion.cpp correctly reads `SweptCurve` and `ExtrudedDirection`, but the `Depth` argument it hands to the geometry kernel was:

```cpp
std::numeric_limits<double>::infinity()

```

`inst->Depth()` — the actual value from the IFC file — was never read at all.

----------

**Why infinity freezes Blender**

The geometry kernel (OpenCASCADE) receives a surface with literally infinite extent. When Bonsai tries to display the element, OCC attempts to compute a bounding box and tessellate the surface into triangles. Both operations iterate over the surface's parametric domain, which is `[0, ∞)`. That loop never terminates — Blender hangs.

----------

**The fix**

```cpp
// Before
std::numeric_limits<double>::infinity()

// After
inst->Depth() * length_unit_

```

`inst->Depth()` reads the value actually stored in the IFC entity (3.0 in the example file). The `* length_unit_` scales it from the file's unit system (metres, millimetres, etc.) into the kernel's internal unit — exactly the same pattern used by `IfcExtrudedAreaSolid`, the solid equivalent of this entity.

BIMvoice:
Closes #8013.

The IfcSurfaceOfLinearExtrusion mapping passed `std::numeric_limits<double>::infinity()` as the extrusion depth and never read `inst->Depth()` (3.0 in the sample file). The kernel then built the prism with an infinite translation vector, so the surface had unbounded extent and the bounding box and BRepMesh tessellation iterated over an unbounded domain and never terminated, which is the reported freeze on load.

This reads `inst->Depth() * length_unit_` instead, the same pattern the sibling `IfcExtrudedAreaSolid` and `IfcExtrudedAreaSolidTapered` mappers already use.

Verification: the hang is live reproduced (create_shape on the sample product runs past a 30 s hard timeout and is killed). The after state needs a kernel build to run, so it is not build tested, but a finite depth yields a finite prism and the change mirrors the proven sibling mappers.

### DIFF
--- a/src/ifcgeom/mapping/IfcSurfaceOfLinearExtrusion.cpp
+++ b/src/ifcgeom/mapping/IfcSurfaceOfLinearExtrusion.cpp
@@ -35,6 +35,6 @@ taxonomy::ptr mapping::map_impl(const IfcSchema::IfcSurfaceOfLinearExtrusion* in
 		matrix,
 		map(inst->SweptCurve()),
 		taxonomy::cast<taxonomy::direction3>(map(inst->ExtrudedDirection())),
-		std::numeric_limits<double>::infinity()
+		inst->Depth() * length_unit_
 	);
 }


BIMvoice:
Re-opened against `v0.9.0`, the current default branch, with RED/GREEN on a real build: #9392. Closing this one so there is a single PR to review.

PR #8437, opened 2026-07-10
Bonsai: uniquify referenced SVG ids on a sheet so text does not garble (#7289)

### LINKED ISSUE #7289: If a sheet has more than one referenced svg, it garbles the text
### Bug Description

If I try to print out `A22 - UNTITLED`, you'll see that it garbles the 2nd reference svg. 

That sheet is referencing...
- references\Structural\DESIGN CRITERIA.svg
- references\Structural\FOUNDATION DESIGN CRITERIA.svg

files: https://hub.openingdesign.com/OpeningDesign/Cool_Car_Guys/src/commit/ae461671e684f51816e10fa1a42f80dbcae2096c/Open/Models/Bonsai

<details>
  <summary>Click here for shallow and sparse clone of all dependent files</summary>
  

You can use the following script to do a shallow and sparse download/clone of all the
dependent files, including textures, drawing assets, etc.

Just copy and paste the entire script into Git Bash. Will take a couple minutes to download.

```
#!/bin/bash
set -e

# Start timer
SECONDS=0

URL="https://hub.openingdesign.com/OpeningDesign/Cool_Car_Guys.git"
SPARSE_PATH="Open/Models/Bonsai"
COMMIT="ae461671e684f51816e10fa1a42f80dbcae2096c"

# Extract repo name from URL
REPO_NAME=$(basename -s .git "$URL")

echo "Cloning into folder: $REPO_NAME"

# Shallow clone with no checkout, filtering blobs
git clone --depth=1 --filter=blob:none --no-checkout "$URL" "$REPO_NAME"
cd "$REPO_NAME"

git sparse-checkout init --cone
git sparse-checkout set "$SPARSE_PATH"

if [ -n "$COMMIT" ]; then
    git checkout "$COMMIT"
fi

# Shallow submodules too
git submodule update --init --recursive --depth=1 --progress

# Calculate elapsed time in minutes and seconds
ELAPSED=$SECONDS
MINUTES=$((ELAPSED / 60))
SECONDS_LEFT=$((ELAPSED % 60))
echo "Clone and setup completed in ${MINUTES}m ${SECONDS_LEFT}s."

read -p "Press Enter to close..."

```
</details>

Image

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Closes #7289.

A sheet that references two or more SVGs garbled the text of all but the first. The referenced SVGs (markdown converted to SVG) define text glyphs as reusable symbols with generic ids like `glyph-0-0` and render every character via `<use xlink:href="#glyph-0-0">`, and both files, generated independently, use the identical id numbering. `SheetBuilder.build_documents` embedded each referenced SVG inline without namespacing the ids, unlike `build_drawings`, so the sheet had many duplicate ids. Per the SVG spec `<use href="#id">` resolves to the first element with that id in document order, so the second reference's characters all rendered with the first reference's glyphs.

This generalizes `ensure_drawing_unique_styles` to `ensure_unique_ids` and applies it to embedded referenced documents as well, prefixing ids, classes, `url(...)` and `<use>` hrefs per view by the reference's IFC id, exactly as `build_drawings` already does. The `<style>` handling is now conditional, since referenced documents carry a `<defs>` but no `<style>` block, where the previous hard assert would have crashed.

Verified against the issue's own sample SVGs: before, 141 duplicate ids across the two references; after, 0, with each reference's glyphs prefixed (`d101-glyph-0-0` etc.) and all `<use>` refs rewritten with zero unresolved. A regression check on a drawing SVG with a `<style>` block confirms the drawings path is byte for byte equivalent to before.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/sheeter.py
+++ b/src/bonsai/bonsai/bim/module/drawing/sheeter.py
@@ -329,47 +329,17 @@ class SheetBuilder:
         titleblock.append(g)
         titleblock.remove(image)
 
-    def ensure_drawing_unique_styles(self, svg: ET.Element, drawing_id: int) -> ET.Element:
-        """ensures all drawing's classes and ids will be unique for the whole sheet
-        by adding `drawing_id` based prefix
+    def ensure_unique_ids(self, svg: ET.Element, view_id: int) -> ET.Element:
+        """ensures all view's classes and ids will be unique for the whole sheet
+        by adding `view_id` based prefix.
+
+        Applies to both drawings and referenced documents. Referenced documents
+        (for example markdown converted to SVG) commonly define glyph symbols with
+        generic ids like `glyph-0-0` and reference them via `<use>`. Without a per
+        view prefix these ids collide across embedded views, so every `<use>`
+        resolves to the first matching id and the text is garbled.
         """
-        prefix = f"d{drawing_id}"  # just number doesn't work
-
-        # add .prefix class to all css selectors
-        style = svg.find(f"{SVG}defs/{SVG}style")
-        assert style is not None
-        style_data = style.text
-        assert style_data is not None
-        text = ""
-        brackets_level = 0
-        selector_buffer = ""  # Buffer to accumulate selectors across lines
-
-        for l in style_data:
-            if l == "{":
-                if brackets_level == 0:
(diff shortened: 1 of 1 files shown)


BIMvoice:

PR #8442, opened 2026-07-10
Bonsai: stop shifting an IfcSpace by the container elevation on Regen (#6405)

### LINKED ISSUE #6405: Regen IfcSpace moves the geometry to wrong elevation
**Error Description and Steps to Reproduce**
Blender 4.2.7 + bonsai_py311-0.8.2-alpha250320-windows-x64
Blender 4.3.2 + Bonsai 0.8.1

How to reproduce:
1. Create a new metric m file
2. Add a new IfcStorey
3. Change the elevation of the storey to 3m
4. Set the storey as default
5. Add walls and a slab
6. Generate space from cursor - spce generates as expected
7. select the space and run regen - space jumps to 0,00

**Attachments**
6. without ifc space
Image
7. ifcspace generated correctly
Image
result - ifcspace with wrong geometry
Image

BIMvoice:
Closes #6405.

`generate_space` handles two cases. In the existing-space branch (the Regen path) it called `set_space_representation_from_polygon`, which rebuilds the space geometry in place at the correct elevation, and then `translate_obj_to_z_location(active_obj, z)`, where `z` is the object's own current world bottom from `get_x_y_z_h_mat_from_obj`. That re-applies the object's elevation on top of itself, so every Regen shifted the space up by the container storey elevation. The reporter's older build showed it dropping to 0 instead, but it is the same root cause: Regen re-applying the object's own elevation.

This removes the `translate_obj_to_z_location` call in the existing-space branch. The creation branch and `generate_spaces_from_walls` still use it, since a fresh object starts at z=0 and must be lifted to the container elevation.

Verified live in headless Blender 5.1.2: a space in a storey at elevation 3m, correct bottom z=3.0. Before, one Regen moved it to z=6.0; after, it stays at z=3.0.

### DIFF
--- a/src/bonsai/bonsai/core/spatial.py
+++ b/src/bonsai/bonsai/core/spatial.py
@@ -221,8 +221,11 @@ def generate_space(
             assert space_polygon
 
     if element and element.is_a("IfcSpace"):
+        # The existing space object is already at its correct elevation, and
+        # set_space_representation_from_polygon rebuilds the geometry in place.
+        # Re-applying the object's own elevation here would shift it by the
+        # container elevation (see #6405).
         spatial.set_space_representation_from_polygon(active_obj, element, space_polygon, h, polygon_is_si=True)
-        spatial.translate_obj_to_z_location(active_obj, z)
     else:
         if relating_type:
             name = model.generate_occurrence_name(relating_type, "IfcSpace")


CyrilWaechter:
Sorry I did not see this PR and the related issue when I pushed my commit 3 days ago. See https://github.com/IfcOpenShell/IfcOpenShell/commit/8deefe497cca8b9fd41e29e809ec0d0ad9478169 I was answering my own observation on my project.

PR #8446, opened 2026-07-10
Bonsai: make renames undoable (#5633)

### LINKED ISSUE #5633: Undo doesn't work with renames
Examples

1. Renaming an object (`subscribe_rna` callbacks):

https://github.com/user-attachments/assets/35b62589-db2f-4ecb-93c2-9c62a49566a1

2. Renaming a material (`update` callback in blender properties), same issue with any other these kind of renamed (profiles, styles, drawings, etc).

https://github.com/user-attachments/assets/f59163c9-2a6e-4f26-8c6f-6b80ff0a25f0

BIMvoice:
Closes #5633.

Renames of IFC linked objects and materials are applied in `name_callback`, which runs from a `msgbus` notification (objects) or a property update (materials), outside `execute_ifc_operator`. Only edits inside that operator wrapper are bracketed by `begin_transaction` / `end_transaction` and pushed to the undo history with a fresh `last_transaction` key. Because the rename runs outside it, the `Name` write was never recorded in a transaction and never advanced `last_transaction`. Blender's `undo_post` handler rolls the IFC back off a mismatch between `IfcStore.last_transaction` and the Blender restored `props.last_transaction`, so with neither changed, undo had nothing to revert and the IFC name stayed changed while the Blender object name reverted.

This adds `IfcStore.track_transaction_outside_operator`, a context manager that brackets an edit made in a callback the same way the operator wrapper does: it opens an `ifc_file` transaction, registers the standard undo and redo callbacks in the history, and advances `last_transaction`. It no-ops when there is no file or when already inside an operator transaction, so nested edits just join the ongoing transaction. `name_callback` wraps its IFC writes (material `Name`, object `Name`, grid axis `AxisTag`) in it; the Blender side writes stay outside and are handled by Blender's own undo.

Verified live in headless Blender 5.1.2 through the real `undo_post` / `redo_post` handlers: before, undoing an object rename left `ifc.Name` unchanged; after, undo restores the previous name for both objects and materials and redo re-applies it. `ifc_file.transaction` is None after the callback, and a subsequent real operator runs without tripping the override transaction assert.

Note: the same non transactional pattern exists in a few other update callbacks; this fixes the object and material rename cases from the report and provides the reusable helper the others can adopt.

### DIFF
--- a/src/bonsai/bonsai/bim/handler.py
+++ b/src/bonsai/bonsai/bim/handler.py
@@ -85,7 +85,8 @@ def name_callback(obj: Union[bpy.types.Object, bpy.types.Material], data: str) -
             if props.is_renaming:
                 props.is_renaming = False
                 return
-            tool.Ifc.get().by_id(ifc_definition_id).Name = obj.name
+            with IfcStore.track_transaction_outside_operator("Rename Material"):
+                tool.Ifc.get().by_id(ifc_definition_id).Name = obj.name
             refresh_ui_data()
         return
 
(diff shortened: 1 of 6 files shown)


BIMvoice:
Extended with a second commit: wraps the property-update rename callbacks (material, profile, style, drawing, document) in the same track_transaction_outside_operator() helper, since those are the second, distinct mechanism the issue named and were left unfixed by the first commit. Confirmed none of the 5 callbacks create new IFC entities, so the orphaned-entity risk doesn't apply, and verified the property-update timing empirically holds the same guarantee as the msgbus path (material and style cases tested live with real undo/redo).

PR #8448, opened 2026-07-10
Bonsai: orient the section marker to the view direction (#4103)

### LINKED ISSUE #4103: section annotation is going the wrong way.
In the following file the section annotation (marked in red) is going the wrong way. 

file: https://hub.openingdesign.com/OpeningDesign/Restaurant_Sun_Prairie/src/commit/7a8f0840457a6923a92a41fdb98cf54f3a988942/Models/BlenderBIM/Restaurant_Sun_Prairie.ifc

image
2023-12-11_20-16-04_blenderBlender_ DDropboxGitea_ODRestaurant_Sun_PrairieModblender

BIMvoice:
Closes #4103.

`generate_section_reference_points` ordered the cut line endpoints purely by the section local X, while the SVG renderer always draws the section arrow marker on the +90 CCW side of the start to end edge. For standard NORTH/SOUTH/EAST/WEST sections that ordering happens to keep the section view direction on the marker side, so they render correctly. But for a section whose frame has the opposite handedness or a flipped up vector (a duplicated and rotated section, or one authored in another tool), the endpoint ordering is unchanged while the true view direction is now on the opposite side, so the marker points the wrong way. That orientation dependence is the intermittency the reporters described.

This swaps the two endpoints when a signed area (cross product) test shows the section view direction is on the wrong side of the cut line, so the marker always faces the view direction. Standard hint created sections already satisfy the test and are unchanged, and the same point generator is reused by the regenerate path.

Verified live in headless Blender 5.1.2: a standard NORTH section renders the marker correctly before and after. A section whose frame is flipped 180 degrees about world X (view reversed, cut line preserved) rendered the marker on the wrong side before and is corrected after, with no change to the standard case.

### DIFF
--- a/src/bonsai/bonsai/tool/drawing.py
+++ b/src/bonsai/bonsai/tool/drawing.py
@@ -1785,6 +1785,15 @@ class Drawing(bonsai.core.tool.Drawing):
         im = camera.matrix_world.inverted()
         v1, v2 = [im @ Vector((m @ np.append(v, 1.0))[:3]) for v in [v1, v2]]
 
+        # Orient the cut line so the section's view direction lies on the marker
+        # side. The SVG section marker always points to the same perpendicular
+        # side of the line, so ordering the endpoints by local X alone makes the
+        # marker point the wrong way for sections whose frame is flipped. See #4103.
+        view_dir = im.to_3x3() @ Vector(-m[:3, 2])
+        edge_dir = v2 - v1
+        if edge_dir.x * view_dir.y - edge_dir.y * view_dir.x < 0:
+            v1, v2 = v2, v1
+
         target_view = cls.get_drawing_target_view(drawing)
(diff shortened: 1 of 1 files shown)


theoryshaw:
Hi @BIMvoice a pr to your pr: https://github.com/BIMvoice/IfcOpenShell/pull/2

BIMvoice:
Update: folded in a follow-up fix from @theoryshaw (github.com/BIMvoice/IfcOpenShell/pull/2) — the orientation test above only worked for plan views; it's degenerate (always 0) for section/elevation views since the pre-clip cut-line edge runs along local Z there, and `elevate_segment` discards ordering anyway. His fix moves the test to after clip/elevate on the final projected edge (non-degenerate for every view, plan views unchanged), and also corrects the regenerate path so annotations authored before this fix self-heal. Verified against the actual `elevate_segment`/guard code before merging.

PR #8451, opened 2026-07-10
ifcgeom: keep cant twist in IfcSectionedSurface via IfcAxis2PlacementLinear (#8116)

### LINKED ISSUE #8116: IfcSectionedSurface isn't rendered correctly because of IfcAxis2PlacementLinear
### Bug Description

@aothms @civilx64 please have a look at this and let me know the best way to proceed. I didn't want to commit a change since the last time I did that it would have negatively impacted your work on the VS.

The sample file attached is an IfcSegmentReferenceCurve alignment with an IfcSectionSurface swept along it.
The horizontal alignment is straight, the vertical is flat (no gradient), and the cant segment is a clothoid so linear transition. The twist due to cant is not being properly considered when making the loft for the IfcSectionedSurface.

General perspective

Image

Elevation View - Expected gain in elevation due to cant. This is the deviating elevation.

Image

End View - Looking down the axis of the surface, it is expected that it would rotate due to cant, but it is not.

Image

The core problem is in IfcAxis2PlacementLinear.cpp
https://github.com/IfcOpenShell/IfcOpenShell/blob/335ee1a1bb80922e4a96e3a92a27671ea92b60d7/src/ifcgeom/mapping/IfcAxis2PlacementLinear.cpp#L65-L69

When Axis and RefDirection are not given, they are taken from the curve. That's not what this code does.

The fix is

~~~
    } else if (!hasAxis && !hasRef) {
        refDirection = m->components().col(0).head<3>(); // RefDirection is the curve tangent when omitted
        axis = m->components().col(2).head<3>(); // Axis is the curve normal when omitted
    } else {
~~~

This also brings into question what to do with this case
https://github.com/IfcOpenShell/IfcOpenShell/blob/335ee1a1bb80922e4a96e3a92a27671ea92b60d7/src/ifcgeom/mapping/IfcAxis2PlacementLinear.cpp#L60-L65

Axis isn't given but RefDirection is - should the implementation be changed to
~~~
    } else if (!hasAxis && hasRef) {
        taxonomy::direction3::ptr r = taxonomy::cast<taxonomy::direction3>(map(inst->RefDirection()));
        refDirection = *r->components_;
        axis = m->components().col(2).head<3>(); // Axis is the curve normal when omitted
~~~

Applying the fix results in the correct sectioned surface

Image

### Attachments

[IfcSectionedSurface.ifc.txt](https://github.com/user-attachments/files/28357873/IfcSectionedSurface.ifc.txt)

### Debug and Error Output

```true

```

BIMvoice:
Fixes #8116.

## Problem

An `IfcSectionedSurface` swept along a canted (superelevated) alignment renders without any cross-section twist: the end view never rotates, so cant/superelevation is lost.

The directrix frame carries the correct twist, but `IfcAxis2PlacementLinear` throws it away. When `Axis` (and/or `RefDirection`) are omitted, the old code recomputed the local Z axis from the global up vector:

```cpp
Eigen::Vector3d up(0, 0, 1);
axis = refDirection.cross(up.cross(refDirection));
```

That forces the placement into a world-vertical plane and discards the cant rotation baked into the reference-curve normal.

## Fix

Per the schema, an omitted `Axis` defaults to the reference-curve normal and an omitted `RefDirection` to the reference-curve tangent. Take those directly from the incoming curve frame (`m.col(2)` and `m.col(0)`) instead of rebuilding the axis from global Z, then orthogonalize `RefDirection` against `Axis` as before.

The four original branches are unified into "default Axis to curve normal, default RefDirection to curve tangent". The cases where `Axis` is given explicitly (Axis-only and Axis+RefDirection) are behaviourally unchanged; only the two omitted-`Axis` branches change, which are exactly the ones that dropped the cant.

## Verification

- Traced against the issue sample: `IFCAXIS2PLACEMENTLINEAR(#pbde,$,$)` (both omitted) on an `IfcSegmentedReferenceCurve` directrix with cant. With the curve normal as `Axis`, `make_loft`'s frame remap yields the correct (lateral, up-canted, tangent) sweep frame.
- Matches the direction of the reporter's own proposed patch, and the reporter attached a before/after image confirming the visual result.
- Compilation is covered by CI's `build-ifcopenshell`. I do not have a local kernel build, so this is validated by static frame tracing rather than a rendered image; happy to iterate if a maintainer spots anything.

### DIFF
--- a/src/ifcgeom/mapping/IfcAxis2PlacementLinear.cpp
+++ b/src/ifcgeom/mapping/IfcAxis2PlacementLinear.cpp
@@ -49,31 +49,31 @@ taxonomy::ptr mapping::map_impl(const IfcSchema::IfcAxis2PlacementLinear* inst) 
 	}
 	*/
 
-    if (hasAxis && !hasRef) {
+    // When Axis and/or RefDirection are omitted they default to the orientation
+    // of the reference curve at the placement location:
+    //   - Axis (local Z) defaults to the curve normal,  m.col(2)
+    //   - RefDirection (local X) defaults to the curve tangent, m.col(0)
+    // Taking the curve normal (rather than recomputing an axis from the global
+    // Z direction) preserves the cant / superelevation twist that is baked into
+    // the linear placement frame, so skewed / rotated sections are swept in
+    // their correct frame. See issue #8116.
+    if (hasAxis) {
         taxonomy::direction3::ptr a = taxonomy::cast<taxonomy::direction3>(map(inst->Axis()));
         axis = *a->components_;
-
-        refDirection = m->components().col(0).head<3>(); // RefDirection is the curve tangent when omitted
-        // refDirection is not necessarily orthogonal to axis.
-        // axis.cross(refDirection) gives y. y.cross(axis) gives x=refDirection
-        refDirection = axis.cross(refDirection).cross(axis);
-    } else if (!hasAxis && hasRef) {
-        taxonomy::direction3::ptr r = taxonomy::cast<taxonomy::direction3>(map(inst->RefDirection()));
-        refDirection = *r->components_;
-        Eigen::Vector3d up(0, 0, 1);
-        axis = refDirection.cross(up.cross(refDirection));
-    } else if (!hasAxis && !hasRef) {
-        refDirection = m->components().col(0).head<3>(); // RefDirection is the curve tangent when omitted
-        Eigen::Vector3d up(0, 0, 1);
-        axis = refDirection.cross(up.cross(refDirection));
     } else {
-        taxonomy::direction3::ptr a = taxonomy::cast<taxonomy::direction3>(map(inst->Axis()));
-        axis = *a->components_;
+        axis = m->components().col(2).head<3>(); // Axis is the curve normal when omitted
+    }
 
+    if (hasRef) {
         taxonomy::direction3::ptr r = taxonomy::cast<taxonomy::direction3>(map(inst->RefDirection()));
         refDirection = *r->components_;
-        refDirection = axis.cross(refDirection).cross(axis); // refDirection needs to be orthogonal to axis
+    } else {
+        refDirection = m->components().col(0).head<3>(); // RefDirection is the curve tangent when omitted
     }
+
+    // refDirection is not necessarily orthogonal to axis:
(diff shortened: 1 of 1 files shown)


RickBrice:
@BIMvoice this is a duplicate of the resolution in https://github.com/IfcOpenShell/IfcOpenShell/pull/8145.

RP #8145 goes the additional step of addressing issues noted in the matrix4::init function which identifies identity matrix prior to memory allocation for a new Eigen::Matrix4d object.

RickBrice:
@BIMvoice  on what basis do you make this claim:

> Per the schema, an omitted Axis defaults to the reference-curve normal

The schema is silent regarding the direction of Axis if not provided. The schema only says Axis is the exact direction of the local Z-axis which is the same description used for IfcAxis2Placement3D which does prescribe the default Axis as (0,0,1).

While I agree with your interpretation, there is open and ongoing debate in the bSI IF and 4x development repos. See 
https://github.com/buildingSMART/IFC4.x-development/issues/1151 and https://github.com/RickBrice/IFC-Alignment-Geometry-Implementation-Guide/blob/main/8_LinearPlacement.md

Also see

https://github.com/buildingSMART/IFC4.x-IF/issues/169
https://github.com/buildingSMART/ifc-gherkin-rules/issues/506
https://github.com/buildingSMART/IFC4.x-IF/issues/125
https://github.com/buildingSMART/IFC4.x-development/issues/732
https://github.com/buildingSMART/IFC4.x-development/issues/731
https://github.com/jwouellette/TPF-5_372-Unit_Test_Suite/issues/15#issuecomment-2241519725

If you have a definitive and authoritative reference, please share it to help bring the debate about the default orientation of Axis to a close.

BIMvoice:
Thanks @RickBrice, you are right on both counts.

#8145 is the prior and more complete resolution (it also addresses the identity case in `matrix4::init`), so this PR is redundant. I am closing it in favor of #8145.

On the wording: my description overstated it. The schema is indeed silent on the default direction of `Axis` when omitted, so "per the schema ... defaults to the reference curve normal" was too strong. It is an interpretation, the one you and I share and the one that preserves the cant, consistent with the direction of the buildingSMART resolution you linked (IFC4.x-development/pull/1118), not a normative default. I do not have a more authoritative reference than the ongoing bSI discussion you cited. Apologies for the imprecise claim, and thanks for the pointers into that debate.

PR #8457, opened 2026-07-10
serializers: correct OBJ texture-coordinate indices and guard material lookup

BIMvoice:
## Problem

Two defects in `WavefrontObjSerializer`, both found by reading the serializer (no linked issue):

1. When a mesh has both UVs and normals, the face texture-coordinate slot was written with the **normal** index, so face definitions came out as `v/vn/vn` and every `.obj` texcoord reference was wrong. There was no separate UV index counter.
2. `mesh.materials()[material_id]` was indexed without guarding the negative "no style" sentinel, an out-of-bounds access. The glTF serializer already guards this with `material_id >= 0`.

## Fix

- Add a `uvcount_total` counter (mirroring the existing `ncount_total`), accumulate it per mesh, and emit proper `v/vt/vn` faces in the normals-plus-UVs branch.
- Guard both material lookups with `material_id >= 0`, matching the glTF precedent.

The common no-UV and untextured paths are byte for byte unchanged, so this only affects the currently-broken textured branch and the unguarded material access.

## Verification

Static change; no local kernel build, so not runtime-exercised. The UV-index logic parallels the existing normal-index handling, and the material guard mirrors the glTF serializer. Compilation covered by CI's `build-ifcopenshell`.

### DIFF
--- a/src/serializers/WavefrontObjSerializer.cpp
+++ b/src/serializers/WavefrontObjSerializer.cpp
@@ -33,6 +33,7 @@ WaveFrontOBJSerializer::WaveFrontOBJSerializer(const stream_or_filename& obj_fil
 	, mtl_stream(mtl_filename)
 	, vcount_total(1)
 	, ncount_total(1)
+	, uvcount_total(1)
 {
 	obj_stream.stream << std::setprecision(settings.get<ifcopenshell::geometry::settings::FloatingPointDigits>().get());
 	mtl_stream.stream << std::setprecision(settings.get<ifcopenshell::geometry::settings::FloatingPointDigits>().get());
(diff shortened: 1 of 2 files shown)


aothms:
For:

> mesh.materials()[material_id] was indexed without guarding the negative "no style" sentinel, an out-of-bounds access. The glTF serializer already guards this with material_id >= 0.

The serializers depends on the ApplyDefaultMaterials. Please check that instead and then no -1 should occur.

BIMvoice:
Reworked per your point. Removed the ad-hoc material_id >= 0 guards and rely on ApplyDefaultMaterials instead: it is default-on, IfcConvert exposes no toggle to disable it, and it injects the entity default style for unstyled items before triangulation, so material_ids never contain the -1 sentinel reaching the OBJ lookup (same as how ColladaSerializer indexes without a guard). The PR is now just the texture-coordinate index correction. Verified: styled box exports its material, unstyled box exports the injected DefaultMaterial with a valid index and no out-of-bounds, on both the default and --generate-uvs paths.

PR #8460, opened 2026-07-10
ifcgeom: fix is_manifold always false for a loose-face compound (#8105)

### LINKED ISSUE #8105: Problem with manifold check when reorient shells is false
### Bug Description

When reorient shells is false, compound of faces is returned and is_manifold check will never return true because compound of faces has no connectivity information about geometry. Why does converting shell result in a compound of faces and not in a shell in the first place?

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Fixes #8105.

## Problem

With `reorient_shells` disabled, a faceset is returned as a `TopAbs_COMPOUND` of loose faces (no enclosing shell). `is_manifold()` recursed into each child, so every face was evaluated on its own. `MapShapesAndAncestors(face, EDGE, FACE)` on a single face finds exactly one ancestor per edge, so the `Extent() != 2` test fired and the shape was always reported non-manifold. This also breaks `calculate_projected_surface_area`, whose `/2` halving depends on manifoldness.

## Fix

Detect the compound-of-faces case with the existing `is_compound_of_faces()` helper and, only then, skip the per-child recursion and run the edge-to-face ancestor test across the whole compound. Because `faceset_helper` produces shared oriented edges, a closed faceset now yields exactly two ancestor faces per interior edge and is correctly reported manifold. Compounds of solids or shells still recurse as before, so their behavior is unchanged.

## Verification

Reasoned against the exact ancestor-map code path and the issue description. Not runtime-tested (no local kernel build); compilation is covered by CI's `build-ifcopenshell`.

### DIFF
--- a/src/ifcgeom/kernels/opencascade/base_utils.cpp
+++ b/src/ifcgeom/kernels/opencascade/base_utils.cpp
@@ -101,6 +101,37 @@ int IfcGeom::util::surface_genus(const TopoDS_Shape& s) {
 }
 
 bool IfcGeom::util::is_manifold(const TopoDS_Shape& a) {
+	if (a.ShapeType() == TopAbs_COMPOUND && is_compound_of_faces(a)) {
+		// With reorient-shells disabled a faceset is emitted as a loose compound
+		// of faces without an enclosing shell. Recursing into that compound would
+		// evaluate every face in isolation (one ancestor face per edge) and always
+		// report non-manifold. The manifoldness check however is only meaningful
+		// for a boolean operand, which must be a properly oriented shell whose
+		// shared edges pair up. So, independent of the reorient-shells setting,
+		// gather the faces into a shell, orient them, and test that shell instead
+		// of accepting the loose faces as-is.
+		NCollection_List<TopoDS_Shape> faces;
+		shape_to_face_list(a, faces);
+
+		BRep_Builder builder;
+		TopoDS_Shell shell;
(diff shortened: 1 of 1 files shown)


aothms:
The manifoldness check is used in the context of boolean operations. In the context of boolean operations there is no point in feeding in compounds of faces. So with reorient_shells disabled - when we check for manifoldness - the shell needs to be properly oriented with edge pairs regardless of settings.

BIMvoice:
Reworked per your point. Dropped the change that made is_manifold accept a loose compound of faces. Now, independent of reorient_shells, a compound of faces is gathered into a shell and oriented with ShapeFix_Shell before the check, so the manifold test runs on a properly oriented, edge-paired shell rather than on the loose faces.

Verified: a closed cube supplied as a loose compound of six shared-edge faces now reports manifold; an open box and three faces sharing one edge still report non-manifold; a box solid and a single loose face are unchanged. IfcConvert OBJ output is byte-identical on faceset and boolean samples with reorient_shells at its default (the boolean operand is already sewn by ensure_fit_for_subtraction, so the boolean path is unaffected).

One judgment call I want to flag: I do the orientation transiently inside is_manifold (minimal, single-site). Your phrasing could also mean orienting in the shell-building path so the persisted operand is a shell. If you would rather the orientation happen upstream so the built operand itself is a shell, I am happy to move it there.

Issue #5999, opened 2025-01-20
Wire intersection check expensive on DTV model


Moult:
This is basically the same issue as #4915 reopened but that one is getting kind of long and I was able to narrow it down due to the fantastic new builds website.

Testing the rac.ifc file again with recent builds, it now takes 3.75 seconds (compared to 1.1 in 0.7.0 and 0.72 before the regression).

Latest good build is fc9bcb0 (taking about 0.7 to 0.8 seconds) and earliest bad build is a84a581 so I guess somewhere in `git log fc9bcb0..a84a581` something happened.

Here are good stats:

```
cores: 1 cgal 4.010830801998964
cores: 12 cgal 0.6871321169892326
cores: 1 opencascade 6.205589167017024
cores: 12 opencascade 1.124729254981503
cores: 1 hybrid-cgal-simple-opencascade 5.121581318991957
cores: 12 hybrid-cgal-simple-opencascade 1.0077503499924205
```

... and bad:

```
cores: 1 cgal 4.00005674600834
cores: 12 cgal 0.6959928880096413
cores: 1 opencascade 29.673849075014004
cores: 12 opencascade 6.692519722011639
cores: 1 hybrid-cgal-simple-opencascade 28.72333783499198
cores: 12 hybrid-cgal-simple-opencascade 6.574400977988262
```

And test script:

```python
import timeit, ifcopenshell, ifcopenshell.geom, ifcopenshell.ifcopenshell_wrapper
import multiprocessing
print(ifcopenshell.ifcopenshell_wrapper.version())
# ifcopenshell.ifcopenshell_wrapper.turn_on_detailed_logging()

filepaths = [
    r"/home/dion/drive/ifcs/benchmark/racbasicsampleproject.ifc"
]

def test(f, geometry_library, core_count):
    t = timeit.timeit(
        lambda: list(
            map(
                lambda x: x.geometry.verts,
                ifcopenshell.geom.iterate(ifcopenshell.geom.settings(), f, core_count, geometry_library=geometry_library),
            )
        ),
        number=2,
    )
    print(f"cores: {core_count}", geometry_library, t)

for filepath in filepaths:
    f = ifcopenshell.open(filepath)
    print(filepath)
    cpu_count = multiprocessing.cpu_count()
    test(f, "cgal", 1)
    f = ifcopenshell.open(filepath)
    test(f, "cgal", cpu_count)
    f = ifcopenshell.open(filepath)
[...]

PR #8465, opened 2026-07-10
ifcgeom: cap wire_intersections recursion to prevent a stack overflow (#622)

### LINKED ISSUE #622: Recursion in IfcGeom::Kernel::wire_intersections causes stack overflow exception
Hello,

I have tried to convert this ifc file [SampleHouse.ifc](https://github.com/IfcOpenShell/IfcOpenShell/files/3301544/SampleHouse.ifc) (also found here: https://github.com/xBimTeam/XbimSamples/blob/master/ExcelReportExample/SampleHouse.ifc) using version [2bf3648](https://github.com/IfcOpenShell/IfcOpenShell/commit/2bf3648966d762d9f6857beeba7990e46a45da25) (24.3.2019) with OCCT v. 7.3.0 and for the entity '3cUkl32yn9qRSPvBJVyYGp' recursive call of IfcGeom::Kernel::wire_intersections causes stack overflow exception at 80th call.

Any suggestions what may have caused this issue?

Thanks in advance,

Ros

BIMvoice:
Fixes #622.

## Problem

`wire_intersections()` recurses once for every self-intersection loop it splits off. On a degenerate or self-intersecting boundary that recursion is unbounded and overflows the stack, crashing the process (the reporter hit it around the 80th nested frame on `SampleHouse.ifc`).

## Fix

Add a `thread_local` recursion-depth counter with an RAII guard at the top of the function. Once the depth exceeds 32 it logs a `GEO 328` warning and bails out gracefully by appending the wire unmodified and returning `false`, the same fallback the function already uses for non-closed and `n < 3` wires.

- `thread_local` because the geometry iterator can run several OCCT kernels concurrently, so a shared counter would be wrong.
- The RAII guard keeps the counter correct across every return path, including the existing `throw geometry_exception`.
- No signature or ABI change, so the bounding-box broad-phase added for #5999 later in the same function is untouched.

The cap of 32 sits well below the observed ~80-frame overflow while leaving ample headroom for legitimate multi-loop boundaries.

## Verification

Reasoned against the self-recursive call site and the existing fallback pattern; `Logger::Root().Warning` and the `wires.Append(wire)` fallback both mirror existing usage in this file. Not runtime-tested (no local kernel build); compilation is covered by CI's `build-ifcopenshell`.

### DIFF
--- a/src/ifcgeom/kernels/opencascade/wire_utils.cpp
+++ b/src/ifcgeom/kernels/opencascade/wire_utils.cpp
@@ -379,6 +379,30 @@ namespace {
 }
 
 bool IfcGeom::util::wire_intersections(const TopoDS_Wire& wire, NCollection_List<TopoDS_Shape>& wires, const wire_tolerance_settings& settings) {
+	// #622: this function recurses once for every self-intersection loop that it
+	// splits off (see the self-call further down). On a degenerate or pathological
+	// boundary that recursion can run unbounded and overflow the stack. Cap the
+	// depth with a thread_local counter (thread_local because the geometry iterator
+	// may run several kernels concurrently) and bail out gracefully once the cap is
+	// reached, appending the wire unmodified instead of crashing. The RAII guard
+	// keeps the counter correct across every return path, including the throw below.
+	static thread_local int recursion_depth = 0;
+	struct recursion_guard {
+		int& depth;
+		recursion_guard(int& d) : depth(d) { ++depth; }
+		~recursion_guard() { --depth; }
+	} guard(recursion_depth);
+
+	// A single legitimate boundary only needs a handful of splits; the original
+	// report overflowed around the 80th nested call, so a cap well below that keeps
+	// valid input working while preventing the crash.
+	static const int max_recursion_depth = 32;
+	if (recursion_depth > max_recursion_depth) {
+		Logger::Root().Warning("GEO", 328, "Wire intersection resolution exceeded maximum recursion depth; leaving boundary unmodified");
+		wires.Append(wire);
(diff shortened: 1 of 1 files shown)


BIMvoice:
Runtime-verified on macOS / OCC 7.9.2, closing the "not runtime-tested" gap in the PR body.

Because this branch predates Boost `system` being dropped from v0.8.0, I cherry-picked the commit cleanly onto current `origin/v0.8.0` (no conflicts, i.e. exactly what a merge produces) and built IfcConvert alongside an unpatched v0.8.0 baseline. Both compile (exit 0); the new depth-cap warning string is present in the patched binary and absent in the baseline, confirming the patch is compiled in.

Ran the reporter's `SampleHouse.ifc` (IFC4, couch `IfcFurniture 3cUkl32yn9qRSPvBJVyYGp`) through both:
- `-y --weld-vertices`: exit 0, "Done creating geometry (58 objects)".
- `-y`: 55,516 vertices.
- The patched OBJ + MTL are **byte-identical** to the unpatched baseline (md5 `dc0a9a936084a07fbe91e67cb99eb85f`), so the recursion cap does not truncate or alter any geometry on valid input.
- The `GEO 328` depth-cap warning never fires on this file — the couch boundary no longer recurses deeply on OCC 7.9.2, so the guard is a pure safety net.

Note: on current v0.8.0 the original overflow no longer reproduces even without this cap (the `wire_intersections` rewrite + OCCT upgrade resolved it; I posted the run in #622). This PR is therefore correct as defense-in-depth against the still-technically-unbounded recursion, and is safe to merge (no behavior change on valid input).

PR #8467, opened 2026-07-10
ifcgeom: emit mtllib in OBJ output so materials are linked (#4846)

### LINKED ISSUE #4846: [IfcConvert] Cannot export material to object format
Hi all,

I am facing issue that I cannot get the correct material from the export .mtl file.

I have removed all the irrelevant test from the IFC file, and it opens OK in all of the BIM tools.

I can find the associated material information by IfcOpenshell as well as the exported xml via IFCConvert.

But the material is simply not exported into mtl file by export to obj option.

Many thanks for your help.

[export.zip](https://github.com/user-attachments/files/15791731/export.zip)

BIMvoice:
Fixes #4846.

## Problem

Converting to OBJ produces a `.mtl` file with the correct materials, but the `.obj` never references it, so viewers show no materials.

`WavefrontObjSerializer` emits the `mtllib` line only when `mtl_stream.filename()` is set:

```cpp
if (mtl_stream.filename()) {
    std::string mtl_basename = *mtl_stream.filename();
    ...
}
```

But `stream_or_filename`'s filename constructor (`GeometrySerializer.h`) opened the `ofstream` without ever assigning `filename_`, so `filename()` always returned an empty optional. The `mtllib` line was therefore dead code for every file-based OBJ export.

## Fix

Store the filename in the constructor:

```cpp
stream_or_filename(const std::string& fn)
    : ofs_(new std::ofstream(...))
    , filename_(fn)
    , stream(*ofs_)
{}
```

The in-memory (`ostringstream`) constructor still leaves `filename_` empty, and `filename()` is read only by the OBJ serializer's `mtllib` guard, so nothing else changes.

## Verification

Confirmed on the issue sample with the current build: the `.mtl` contained the `newmtl` entry and the `.obj` had `usemtl` but zero `mtllib` lines. The `filename()` accessor is used at exactly one site (the OBJ `mtllib` guard), so this restores the reference with no other behavioral change. Not compile-tested locally; CI's `build-ifcopenshell` compile-checks it.

### DIFF
--- a/src/ifcgeom/GeometrySerializer.h
+++ b/src/ifcgeom/GeometrySerializer.h
@@ -114,6 +114,7 @@ public:
 
 	stream_or_filename(const std::string& fn)
 		: ofs_(new std::ofstream(IfcUtil::path::from_utf8(fn).c_str()))
+		, filename_(fn)
 		, stream(*ofs_)
 	{}
 


BIMvoice:
Closing: already fixed on `v0.9.0`, now the default branch. `src/ifcgeom/geometry_serializer.h:39` initialises `filename_(fn)` in the constructor, so `mtllib` is emitted without this change. Nothing to carry over.

PR #8468, opened 2026-07-10
serializers: emit a stable id for nested materials in XML (#5109)

### LINKED ISSUE #5109: IfcMaterial ids lost when transformed from IFC to XML
Hi there! When I convert my IFC model to XML, I see that some important attributes like "ids" are missing. This is what I get as a result of running IfcConvert:

     
      <IfcMaterialList id="IfcMaterialList_56413">
	    <IfcMaterial Name="Metal"/>
	    <IfcMaterial Name="Metal"/>
	    <IfcMaterial Name="Wood"/>
	    <IfcMaterial Name="Wood"/>
	    <IfcMaterial Name="Wood"/>
      </IfcMaterialList>

Howevr, in IFC itself I see that all these child materials have their "id" attributes. This is the command I used:

`IfcConvert -j 2 model.ifc model.xml -qy`

I also tried options --use-material-names and --use-element-guids, but they do not help. What is wrong with that and how can I get correct XML from my IFC model? Thanks!

BIMvoice:
Fixes #5109.

## Problem

Converting IFC to XML loses material identity: materials nested inside an `IfcMaterialList`, and the `IfcMaterialLayer`s inside an `IfcMaterialLayerSet`, were emitted with no `id` attribute. Because `IfcMaterial`/`IfcMaterialLayer` are unrooted (no GlobalId), several distinct materials (for example multiple `Metal`/`Wood` entries) came out as indistinguishable duplicate nodes with no way to reference them.

## Fix

The top-level material node already receives a stable id via `qualify_unrooted_instance(mat)`. Apply the same STEP-id-derived scheme to the nested `IfcMaterial` and `IfcMaterialLayer` child nodes (two lines), so each nested material is identifiable and referenceable, consistent with how the serializer already ids unrooted instances and its per-product `xlink:href` links.

## Verification

`qualify_unrooted_instance` is already called in `finalize()` (line 813), so it is in scope and correctly typed, and both iterands are entity instances, so the calls are well-formed. Not runtime-tested (no local build); compilation is covered by CI's `build-ifcopenshell`.

### DIFF
--- a/src/serializers/schema_dependent/XmlSerializer.cpp
+++ b/src/serializers/schema_dependent/XmlSerializer.cpp
@@ -822,6 +822,7 @@ void POSTFIX_SCHEMA(XmlSerializer)::finalize() {
 				IfcSchema::IfcMaterialLayer::list::ptr ls = layerset->MaterialLayers();
 				for (IfcSchema::IfcMaterialLayer::list::it jt = ls->begin(); jt != ls->end(); ++jt) {
 					ptree subnode;
+					subnode.put("<xmlattr>.id", qualify_unrooted_instance(*jt));
 					if ((*jt)->Material()) {
 						subnode.put("<xmlattr>.Name", (*jt)->Material()->Name());
 					}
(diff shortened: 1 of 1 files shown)


BIMvoice:
Re-opened against `v0.9.0`, the current default branch, with RED/GREEN on a real build: #9395

PR #8481, opened 2026-07-10
Fixes double unit conversion when convert-back-units are used

RickBrice:
The combination of `create_shape` using `convert-back-units=True` and its subsequent call to `IfcGeom::Transformation`'s constructor causes unit conversion to be incorrectly applied twice.

The attached model and script illustrate the issue. The output from the script is

~~~
Project length unit: 'foot' (1 foot = 0.3048 m)

10+00.00   DistanceAlong=   0.0 foot   BasisCurve=#22
25+00.00   DistanceAlong=   0.0 foot   BasisCurve=#131
10+10.00   DistanceAlong=  10.0 foot   BasisCurve=#22
10+40.00   DistanceAlong=  40.0 foot   BasisCurve=#22

Referent   Expected (ft)   create_shape (ft)   Ratio
10+00.00   0.0000          0.0000              nan
25+00.00   50.0000         164.0420            3.28084
10+10.00   10.0000         32.8084             3.28084
10+40.00   40.0000         131.2336            3.28084
~~~

You can also see a manifestation of the bug in Blender/Bonsai by opening the IFC file and inspecting the coordinates of the referents.
[referent_placement_bug.zip](https://github.com/user-attachments/files/29894547/referent_placement_bug.zip)

This PR removes the redundant unit converse.

@aothms this is a little out of my area of expertise, so please review.

Full disclosure, my good friend Claude helped me diagnose the root cause of the problem.

### DIFF
--- a/src/ifcwrap/IfcGeomWrapper.i
+++ b/src/ifcwrap/IfcGeomWrapper.i
@@ -977,12 +977,14 @@ struct ShapeRTTI : public boost::static_visitor<PyObject*>
 			if (item == nullptr) {
 				throw IfcParse::IfcException("Failed to convert placement");
 			}
+			/*
             if (st.get<ifcopenshell::geometry::settings::ConvertBackUnits>().get()) {
                 // we pass the settings to the Transformation object, but access the data just offloads to the
                 // generic cartesian_base<Matrix4> so there's no time to apply the settings to the translation part.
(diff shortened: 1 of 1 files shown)


aothms:
Oh indeed, thanks. I remember recently changing this in C++, but apparently forgot that there was python special casing happening there as well.

PR #8482, opened 2026-07-10
ifcgeom: implement IfcRevolvedAreaSolidTapered (honour EndSweptArea) (#3540)

### LINKED ISSUE #3540: IfcRevolvedAreaSolidTapered implementation
Was trying to create `IfcRevolvedAreaSolidTapered` and its seems to work the same way as `IfcRevolvedAreaSolid` just ignoring the `.EndSweptArea`. Checked files in `src/geom` and it seems to be missing "IfcRevolvedAreaSolidTapered.cpp". 
Wasn't it implemented yet or I'm missing something? 

@aothms 👀  

```python
from ifcopenshell.util.shape_builder import ShapeBuilder, V
import blenderbim.tool as tool
import bpy

ifc_file = tool.Ifc.get()

builder = ShapeBuilder(ifc_file)
profile = builder.profile(builder.rectangle(V(1000, 1000)))
profile_end = builder.profile(builder.rectangle(V(350, 350)))
angle = 130

axis = ifc_file.createIfcAxis1Placement(
    Location=ifc_file.createIfcCartesianPoint(V(0,0,0)),
    Axis=ifc_file.createIfcDirection(V(1,0,0))
)
rev1 = ifc_file.createIfcRevolvedAreaSolid(SweptArea=profile, Axis=axis, Angle=angle)
rev2 = ifc_file.createIfcRevolvedAreaSolidTapered(SweptArea=profile, EndSweptArea=profile_end, Axis=axis, Angle=angle)
bpy.ops.bim.create_shape_from_step_id(step_id=rev1.id(), should_include_curves=True)
bpy.ops.bim.create_shape_from_step_id(step_id=rev2.id(), should_include_curves=True)
```
image

BIMvoice:
Fixes #3540.

## Problem

`IfcRevolvedAreaSolidTapered` is a subtype of `IfcRevolvedAreaSolid`, so with no dedicated mapping it fell through to the plain revolve handler, which reads only `SweptArea` and emits a `taxonomy::revolve`. Its `EndSweptArea` was silently discarded, so the taper was lost.

Confirmed live on the bundled 0.8.6 build: a 100x100 to 40x40, 90 degree tapered revolve produced geometry byte identical to a plain revolve of the start profile (the 40x40 end profile fully ignored).

## Fix

Add a dedicated `IfcRevolvedAreaSolidTapered` mapping that expresses the tapered revolve as a loft: a `functor_item` directrix representing a rigid rotation about the revolution axis is fed to `make_loft`, which interpolates the start and end profiles while placing them around the arc. This is the same loft mechanism already used by `IfcExtrudedAreaSolidTapered` and `IfcSectionedSolidHorizontal`, so no kernel change is needed. The `BIND` is placed before `IfcRevolvedAreaSolid` so the derived type wins first-match dispatch, mirroring the Extruded pair.

## Verification and honest caveat

The bug (EndSweptArea ignored) is reproduced live. The fix itself is **not** runtime-verified: the available prebuilt ifcopenshell does not carry this patch and I have no local kernel build, so I could not render the result. The approach and dispatch mirror the existing Tapered/loft mappings, and CI's `build-ifcopenshell` compile-checks it, but the residual risk is the placement/orientation math (the rotation sign and the frame-column permutation into `make_loft`'s convention), which needs a render to confirm. I would appreciate a maintainer with a build sanity-checking the produced solid, and I am happy to iterate on the frame math. Known limitation inherited from `make_loft`: the start and end profiles must have matching vertex counts.

### DIFF
--- a/src/ifcgeom/mapping/IfcRevolvedAreaSolidTapered.cpp
+++ b/src/ifcgeom/mapping/IfcRevolvedAreaSolidTapered.cpp
@@ -1,0 +1,106 @@
+/********************************************************************************
+ *                                                                              *
+ * This file is part of IfcOpenShell.                                           *
+ *                                                                              *
+ * IfcOpenShell is free software: you can redistribute it and/or modify         *
+ * it under the terms of the Lesser GNU General Public License as published by  *
+ * the Free Software Foundation, either version 3.0 of the License, or          *
+ * (at your option) any later version.                                          *
+ *                                                                              *
+ * IfcOpenShell is distributed in the hope that it will be useful,              *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 *
+ * Lesser GNU General Public License for more details.                          *
+ *                                                                              *
+ * You should have received a copy of the Lesser GNU General Public License     *
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.         *
+ *                                                                              *
+ ********************************************************************************/
+
+#include "mapping.h"
+#define mapping POSTFIX_SCHEMA(mapping)
+using namespace ifcopenshell::geometry;
+
+#include "../../ifcgeom/infra_sweep_helper.h"
+
+#include <boost/math/constants/constants.hpp>
+
+#include <Eigen/Geometry>
+
+#ifdef SCHEMA_HAS_IfcRevolvedAreaSolidTapered
+
+// IfcRevolvedAreaSolidTapered revolves a start profile (SweptArea) into an end
+// profile (EndSweptArea) about a revolution axis. Unlike a plain
+// IfcRevolvedAreaSolid the cross section morphs along the sweep, so it cannot be
+// expressed as a single taxonomy::revolve. It is instead built as a loft: the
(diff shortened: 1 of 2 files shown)


aothms:
Marked as draft. When there is serious investigation into this with sample models from industry we can have another look, but I'm not big fan of this geometry type: 

There are serious flaws in the semantic definitions:

> Each lateral face is a ruled surface defined by a pair of corresponding edges of the start and end section. The ruled surfaces are constructed in the cylindrical coordinate space defined by the supertype [IfcRevolvedAreaSolid](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcRevolvedAreaSolid.htm).

These are not ruled surfaces.

BIMvoice:
Thanks @aothms, you are right about the semantic flaw. The loft I used builds the lateral faces as ruled surfaces in cartesian space, but as you note a tapered revolve's lateral faces live in the cylindrical coordinate space of the supertype revolve, so a straight cartesian loft between the start and end sections does not reproduce the correct surfaces. That makes this an approximation at best, not a faithful implementation.

I agree it needs real industry sample models and a correct geometric treatment rather than this shortcut, so I am closing it. Thanks for the review.

Issue #5482, opened 2024-09-24
3 point arc seems to glitch out


theoryshaw:
3 point arc seems to glitch out, under the following scenario. 

https://github.com/user-attachments/assets/92faa4d9-dec9-42db-be68-a3ebac587a2e

PR #8484, opened 2026-07-10
Fix #5685 squared curves on open

### LINKED ISSUE #5685: The following file doesn't recreate the curves upon opening.
I think this is probably related to #5482

The following file doesn't recreate the curves upon opening. 

[curved slab.ifc.txt](https://github.com/user-attachments/files/17611986/curved.slab.ifc.txt)

https://github.com/user-attachments/assets/ab80a6a1-cffb-4047-9078-2addfab765a1

sboddy:
## Summary

Fixes the case reported in #5685 (and previously #5482) where a curved
`IfcArbitraryClosedProfileDef` boundary (an `IfcIndexedPolyCurve` using
`IfcArcIndex`) renders as a faceted, angular polygon on file open, and only
becomes smooth after the user tabs into edit mode on the object and back out.

Reproduced with the `curved slab.ifc` file attached to #5685: a slab with a
gently curved, large-radius sidewalk boundary defined by two 3-point arc
segments.

## Root cause

Two independent bugs compounded to produce the reported symptom:

1. **CGAL kernel arc tessellation ignored deflection settings.**
   `evaluate_conic()` in `src/ifcgeom/kernels/cgal/CgalKernel.cpp` sized the
   polyline approximation of an arc purely from the fixed `circle-segments`
   setting (default 16) scaled by the arc's fraction of a full turn. For a
   long, shallow arc this rounds down to as few as 2 segments regardless of
   the arc's actual size, since `circle-segments` only accounts for the
   angular span, not the radius. OpenCASCADE does not have this problem
   because it keeps arcs as exact BRep edges and only discretizes them at
   final mesh time using `mesher-linear-deflection` /
   `mesher-angular-deflection`. Since Bonsai's default Geometry Library is
   `hybrid-cgal-simple-opencascade` (CGAL first, OpenCASCADE only as a
   fallback on a hard failure, not on merely low-quality output), the coarse
   CGAL result is what silently ships on initial import.

2. **The geometry-reload path used a different, unintentional configuration
   than the initial import.** `reimport_element_representations()` in
   `src/bonsai/bonsai/tool/geometry.py` (reached via
   `switch_representation()`, which runs whenever an object's representation
   is reloaded — including when exiting Bonsai's Item/edit mode) builds its
   own `ifcopenshell.geom.settings()` object from scratch. It never copies
   over `deflection_tolerance` / `angular_tolerance` from the
   `IfcImportSettings` it had just constructed one line above, and never
   passes `geometry_library` to either the `iterator()` or `create_shape()`
   calls it makes. It therefore silently fell back to IfcOpenShell's
   hard-coded mesher default (0.001 linear deflection, ~50x finer than the
   project default of 0.05) and to a different default geometry kernel. This
   is why tabbing into and back out of edit mode — with no actual edits made
   — visibly "fixed" the arc: the reload path was, by accident, far more
   precise than the initial import.

## Changes

- `src/ifcgeom/kernels/cgal/CgalKernel.cpp`: `evaluate_conic()` now also
  derives a segment count from `mesher-angular-deflection` (angle-per-segment
  cap) and `mesher-linear-deflection` (chord-sagitta cap based on the conic's
  radius), and takes the max against the existing `circle-segments`-derived
  count.
- `src/bonsai/bonsai/tool/geometry.py`: `reimport_element_representations()`
  now copies `deflection_tolerance` / `angular_tolerance` /
  `geometry_library` from its `IfcImportSettings` instance into the geometry
  settings used for both the `iterator()` and `create_shape()` calls.

## Behavioural note

With both fixes applied, initial import and post-Tab reload converge on the
*same* (now noticeably improved) tessellation quality, at whatever
`deflection_tolerance` / `angular_tolerance` / Geometry Library the project
is configured with. Tabbing into and out of edit mode no longer changes the
mesh. If a project wants tessellation as fine as the (accidental) previous
post-Tab result, that is available today by lowering Deflection Tolerance
and/or switching Geometry Library to OpenCASCADE in Project Properties — a
deliberate, user-controlled trade-off against import performance on large
files, rather than a side effect of a settings bug.

## Testing

- Manually verified against the `curved slab.ifc` file from #5685, loaded
  through Bonsai in Blender (not just via the raw `ifcopenshell.geom` API),
  comparing vertex/face counts and viewport appearance before and after each
  fix, and before/after tabbing into edit mode.
- No automated test was added. A regression test against the attached IFC
  file (checking boundary vertex count for the CGAL kernel, and/or that
  `reimport_element_representations` preserves the configured deflection
  settings) would be a reasonable addition — flagging this as a gap rather
  than including one, since it wasn't written as part of this change.
- Did not benchmark import performance impact of the CGAL kernel change on
  large files; the added work is bounded (a handful of trig calls per arc)
  but this hasn't been measured against a large real-world model.

## AI-generated content

This entire PR — investigation, root-cause analysis, and both commits — was
produced with the assistance of an AI coding tool (Claude Code). A human
contributor built and ran the changes in Blender/Bonsai, provided the
before/after vertex counts and screenshots used to confirm the diagnosis at
each step, and reviewed the resulting diffs.

### DIFF
--- a/src/bonsai/bonsai/tool/geometry.py
+++ b/src/bonsai/bonsai/tool/geometry.py
@@ -1148,6 +1148,9 @@ class Geometry(bonsai.core.tool.Geometry):
         settings.set("layerset-first", True)
         settings.set("keep-bounding-boxes", True)
         settings.set("dimensionality", ifcopenshell.ifcopenshell_wrapper.CURVES_SURFACES_AND_SOLIDS)
+        settings.set("mesher-linear-deflection", ifc_import_settings.deflection_tolerance)
+        settings.set("mesher-angular-deflection", ifc_import_settings.angular_tolerance)
+        geometry_library = ifc_import_settings.geometry_library
 
         ifc_importer = bonsai.bim.import_ifc.IfcImporter(ifc_import_settings)
         ifc_importer.file = tool.Ifc.get()
@@ -1159,7 +1162,11 @@ class Geometry(bonsai.core.tool.Geometry):
         shape = None
         if elements:
             iterator = ifcopenshell.geom.iterator(
-                settings, tool.Ifc.get(), multiprocessing.cpu_count(), include=elements
+                settings,
+                tool.Ifc.get(),
+                multiprocessing.cpu_count(),
+                include=elements,
+                geometry_library=geometry_library,
             )
         else:
             iterator = None  # For example, when switching representation of a type with no occurrences
@@ -1214,7 +1221,9 @@ class Geometry(bonsai.core.tool.Geometry):
         for element in element_types:
             if obj := tool.Ifc.get_object(element):
                 if representation := ifcopenshell.util.representation.get_representation(element, context):
-                    geometry = ifcopenshell.geom.create_shape(settings, representation)
+                    geometry = ifcopenshell.geom.create_shape(
+                        settings, representation, geometry_library=geometry_library
+                    )
                     mesh_name = tool.Loader.get_mesh_name_from_shape(geometry)
                     mesh = meshes.get(mesh_name)
                     if mesh is None:


sboddy:
Something to bear in mind when reviewing. This will cause any previous useage of these arcs used in existing models to now become courser. The default parameter that set this are a bit hidden in the Advanced file loading options.

aothms:
This overlaps with #8368

sboddy:
Rebased onto current `v0.8.0` to resolve the conflict with #8368 (thanks @aothms for flagging the overlap).

**What overlapped:** both PRs touched `evaluate_conic()` in `src/ifcgeom/kernels/cgal/CgalKernel.cpp`, and it turns out they were fixing the same root cause. #8051's `IfcMember` mullions and this PR's `IfcArcIndex`-based `IfcIndexedPolyCurve` boundary both go through `evaluate_conic()`: `IfcIndexedPolyCurve.cpp` builds an `IfcArcIndex` segment into a `taxonomy::circle` basis on the edge (via `taxonomy::circle::from_3_points`), which is the exact same taxonomy object `evaluate_conic()` tessellates for ordinary `IfcCircle` profile arcs. So #8368's deflection-driven segment count (merged) already fixes the faceted-arc symptom this PR was chasing - I verified in Blender/Bonsai against the `curved slab.ifc` file from #5685 with this PR's own CGAL kernel change reverted (i.e. running purely on #8368's merged code): the arc renders smooth on initial import, same as before.

**What I did:**
- Dropped this PR's `evaluate_conic()` commit entirely rather than reconciling the two implementations - #8368's version (`CircleSegments == 0` deflection-driven, `> 0` fixed-count) already covers it, so combining or layering both would have just left two competing tessellation strategies for the same code path.
- Kept the second commit, `reimport_element_representations()` in `src/bonsai/bonsai/tool/geometry.py`, which is the independent bug: the reload path (reached via Tab into/out of Item edit mode) was building its own `geom.settings()` without copying `deflection_tolerance`/`angular_tolerance`/`geometry_library` from the just-built `IfcImportSettings`, so a no-op Tab silently reprocessed geometry at IfcOpenShell's hard-coded defaults instead of the project's configured settings/kernel. Not something #8368 touches.
- While re-verifying that commit against the rebased tree, found its `create_shape()` call (the `element_types` branch, for reloading a type with no occurrences) was still missing `geometry_library`, despite the commit message already claiming both call sites were fixed. Added it, so both the `iterator()` and `create_shape()` calls now pass it consistently.

**Verification:** built the wrapper locally (`docker/ifcos_env`, CGAL kernel from current `v0.8.0` + this branch) and tested in Blender/Bonsai against the `curved slab.ifc` file from #5685 - smooth arc on initial import, and no change in mesh quality after tabbing into and back out of edit mode.

This part (conflict analysis, rebase, the `create_shape` fix, and this comment) was done with the assistance of an AI coding tool (Claude Code); I reviewed the resulting diff and did the manual in-Blender verification myself.

sboddy:
~~IMPORTANT: Additional note from a thought that hit me. I've tested and confirmed. If this is committed _before_ #8368 gets included in an ifcopenshell build, that is then bumped in Bonsai, then the very low resolution seen on opening the test file in #5685 will persist after tabbing out. It will be more _correct_ in a way, but it will be ugly.~~

~~Maybe wait till after then before you consider merging this.~~

No longer relevant. #8368 has already been merged.

theoryshaw:
Hi, Just a heads up — this PR overlaps with #8201 (*Dev notes system*) and #8200 (*Preserve custom opening geometry via a type-level Reference template*), and on top of `v0.8.0` it **doesn't auto-merge** with either (verified with `git merge-tree`). It's not a trivial overlap: the merge leaves **1 conflict region / ~25 lines** to resolve by hand in `src/bonsai/bonsai/tool/geometry.py` in each case.

Nothing's broken while these are open — but whichever lands second will need a manual rebase, so flagging early to coordinate. 🙂

sboddy:
@theoryshaw Tag! You're it! ;-)

sboddy:
@aothms @Moult There is something strange going on here. In the original bug this addresses there's a test piece. When loading it and using the Advanced settings to open it, you cannot now adjust the smoothness of the interpolated curve. The first version of this patch respected these values, and gave nice smooth curves.

@aothms You merged BIMvoices PR, and I reworked this PR to avoid the conflict. Now this is merged too, and I see it is not respecting the deviation/tolerance thresholds in the Advanced tabs. It seems locked in on a course interpolation. I have other things to do right now, but I'm going to have to take another look at this, unless someone wants to take a look while I do my other chores.

@theoryshaw Just tagging you because it might be relevant to you, as it is your issue, and you have potentially conflicting PR's too.

PR #8488, opened 2026-07-10
ifcgeom: produce closed manifold IfcSurfaceCurveSweptAreaSolid sweeps (#1541)

### LINKED ISSUE #1541: IfcSurfaceCurveSweptAreaSolid produces non-manifold geometry
I often see this glitch when extruding a profile along a 2D directrix. Instead of a mitred corner, two segments are extended a short distance past the junction and left uncapped:

Screenshot from 2021-06-27 23-12-18

Test file: [tmpya4me9nl.ifc.gz](https://github.com/IfcOpenShell/IfcOpenShell/files/6722389/tmpya4me9nl.ifc.gz)

(This is possibly an opencascade bug and not fixable in IfcOpenShell).

BIMvoice:
Fixes #1541.

## Problem

`IfcSurfaceCurveSweptAreaSolid` produced open / non-manifold solids. The kernel built the lateral tube with `BRepOffsetAPI_MakePipeShell`, then added the two end-cap faces to a raw `BRep_Builder` shell. `BRep_Builder::Add` does not merge edges, so the caps were only geometrically coincident with the tube, not topologically shared, leaving the shell open. Sharp directrix corners additionally made the hard-wired `RightCorner` transition mode self-intersect or fail, with no fallback.

## Fix

When the plain `MakeSolid` assembly is not `BRepCheck_Analyzer`-valid:
- sew the caps to the tube with `BRepBuilderAPI_Sewing` (at precision) so they share edges, then rebuild the solid with `ShapeFix_Solid`;
- if a corner still yields an invalid solid, retry with the alternate transition mode, both attempts wrapped in `try/catch(Standard_Failure)` so a throwing corner falls back instead of aborting.

Already-valid sweeps take a fast path and are returned unchanged, so working geometry is untouched.

## Verification (built and rendered locally, OCC 7.9.2)

Converted the issue's sample with vertex welding and checked manifoldness of the two swept solids:

| element | baseline (v0.8.0) | with this fix |
|---|---|---|
| product 7572... | OPEN, 26 boundary edges | **CLOSED** |
| product 9d0e... | NON-MANIFOLD, 9 edges | **CLOSED** |

Both solids are watertight and manifold after the fix; the baseline is not. Builds cleanly.


aothms:
The bug report predates the shape of the code. Non-manifoldness is also not the issue reported, but actual geometric artefacts involving displacement. I don't want to accept this based on only manifold edge counts. Maybe a simpler fix is to reverse one of the caps.

BIMvoice:
You were right to push back, the manifold-edge-count check wasn't the right metric, and the real bug you originally reported was still there underneath the cap-orientation fix. Found and fixed it properly now.

Root cause: one interior directrix vertex in your file is collinear with its neighbours to about 3e-7m (~1.3e-5 degrees of turn), a redundant vertex from however the file was authored. Building just the two segments around it in isolation is fine; as part of the full multi-corner directrix it corrupts OCCT's `BRepFill_TrimShellCorner` processing at the *next* real corner, silently leaving two tube sections untrimmed and unjoined there instead of throwing. Confirmed by perturbing that one vertex 0.05m off the line, with nothing else changed, the same directrix converts cleanly.

Checked #8400's precedent before writing anything new: the short-segment-collapse fix there is architecturally the same idea (a directrix-sanitizing pre-pass), just with a length criterion instead of collinearity, so this follows that pattern. The Angmin/`BRepFill_PipeShell` fix from the same PR doesn't apply here, verified experimentally it only affects the tangent-circular-joint transition style, not the straight-corner one your directrix uses.

Verified: the persisting hole (26 boundary edges after welding, isolated to `rep_133.ifc`) is now 0, positive volume. Your original 2021 file's affected wall has its hole closed, the other wall in the same file is byte-identical before/after. All 51 existing `test/input` fixtures using swept solids convert to byte-identical output (none of them happen to contain a redundant collinear vertex, so nothing regressed). A plain profile on the same directrix with vs. without the redundant vertex produces byte-identical output, confirming the simplification is lossless, and it doesn't engage at all on a genuine corner or a closed loop.

PR #8492, opened 2026-07-10
ifcgeom: warn when a representation is used both directly by a product and as a mapped item (#2248)

### LINKED ISSUE #2248: Missing railings when the whole file is exported (but correctly converted individually)
Hello. I observed a strange behaviour when exporting railings from IFC to GLB:

Expected behaviour: 8 IfcRailing objects

image

IfcConvert conversion (for the whole model): only 6 railings after conversion

image

However, when I pass only the missing GUID's in the command line (using `--include arg GlobalId 3_DI4xBET0pR2yN9OkwlKn 2h77TPNyr2bgYVQ56HRkmL`), these objects are properly converted, so IfcConvert works fine for them. It's only when the whole file is being converted that they seem to disappear.

image

The IFC file is available in attachment. Thanks in advance!

[sample_railings_not_exported.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/8976179/sample_railings_not_exported.zip)

BIMvoice:
Fixes #2248.

## Problem

Some products (railings in the report) are silently dropped when the whole file is converted, but convert correctly when isolated with `--include`. The dropped elements produce valid geometry individually, so it is an enumeration defect, not a kernel failure.

## Root cause

In `mapping::get_representations`, for a single `IfcShapeRepresentation` that is used both as a **direct** product representation and as an `IfcRepresentationMap` source, when geometry reuse is not applicable (the direct and mapped products carry different `IfcMaterial`s):
1. the map representation is skipped (it has a used `RepresentationMap`) without emitting its direct products, and
2. the mapped-item container representation is then deferred back to it through a `reuse_ok_` check that only sees the pruned single direct product, so the container's products are dropped too.

## Fix

- When a not-reusable used-map representation is skipped, emit a task for its direct products (`products_represented_by(..., only_direct=true)`).
- Do not defer a mapped-item container representation when the mapped-to representation's `RepresentationMap` is already marked not-reusable (those products are rendered directly).

Both enumeration orders were traced: every product is emitted exactly once, no double emission.

## Verification (built and measured locally, OCC 7.9.2)

On the issue's IFC2X3 sample:

| | products emitted (whole file) |
|---|---|
| baseline (v0.8.0) | 6 |
| with this fix | 11 |

11 matches converting the same file with `--include entities IfcRailing`, so every railing now renders in whole-file conversion. Builds cleanly.

### DIFF
--- a/src/ifcgeom/mapping/mapping.cpp
+++ b/src/ifcgeom/mapping/mapping.cpp
@@ -249,6 +249,20 @@ void mapping::get_representations(std::vector<geometry_conversion_task>& tasks, 
 
         IfcSchema::IfcRepresentationMap::list::ptr maps = representation->RepresentationMap();
 
+        // Per IfcShapeModel WR11 a representation shall be used by exactly one of
+        // OfProductRepresentation, RepresentationMap or OfShapeAspect. Detect the
+        // invalid situation where a representation is assigned to a product directly
+        // and, at the same time, used as the source of a mapped item. This is not
+        // handled/recovered here; just warn so the invalid data can be corrected.
+        if (representation->OfProductRepresentation()->size() > 0) {
+            for (auto& used_map : *maps) {
+                if (used_map->MapUsage()->size() > 0) {
+                    logger_.Warning("GEO", 329, "Representation is used both directly by a product and as a mapped item, which is not valid (IfcShapeModel WR11)", representation);
+                    break;
+                }
+            }
+        }
+
         if (!geometry_reuse_ok_for_current_representation_ && maps->size() == 1) {
             // unfiltered_products contains products represented by this representation by means of mapped items.
             // For example because of openings applied to products, reuse might not be acceptable and then the


aothms:
This is not valid per https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcShapeModel.htm

WR11 	

The IfcShapeModel shall be used by an IfcProductRepresentation, by an IfcRepresentationMap or by an IfcShapeAspect.

(SIZEOF(SELF\IfcRepresentation.OfProductRepresentation) = 1) XOR 
(SIZEOF(SELF\IfcRepresentation.RepresentationMap) = 1) XOR
(SIZEOF(OfShapeAspect) = 1)

Instead just log a warning - I don't want to pollute the already complex logic here for invalid situations that don't impact normal usage.

BIMvoice:
Reworked per your WR11 point. Dropped the dual-role recovery logic entirely, so valid files convert byte-identically to v0.8.0 again (verified on the #2248 rail sample and a plain IFC4 model). The invalid case (a representation used both directly by a product and as a mapped item, against IfcShapeModel WR11) now just logs a warning (GEO 329) naming the representation, without polluting the mapping logic.

PR #8498, opened 2026-07-11
Use NEGATIVE-side IfcSurfaceStyle as a fallback instead of dropping it (#5618)

### LINKED ISSUE #5618: Support for negative IfcSurfaceSide in IfcSurfaceStyle.Side
Mentioned in #5609

Currently if `IfcSurfaceStyle.Side` is `NEGATIVE` then IfcSurfaceStyle will be just ignored.

https://github.com/IfcOpenShell/IfcOpenShell/blob/4acdd7c960ba73db5a7b486b768c61e8013abd22/src/ifcgeom/mapping/mapping.cpp#L433-L444

In theory it shouldn't be ignored and should be still assigned to the face but there should a way to identify which side is using the style. But supporting this would probably also mean supporting assignment of multiple styles to a single face (IfcStyledItem supports multiple styles for a single representation item) which would complicate stuff a bit (now we usually assume there's just 1 style per item).

Not sure if there any models actually using this right now, just creating the issue so there will be something to refer to.

BIMvoice:
Fixes #5618.

## Problem
An item styled only with an `IfcSurfaceStyle` whose `Side` is `.NEGATIVE.` is rendered unstyled (gray `DefaultMaterial`); the style is silently ignored.

## Root cause
The surface-style picker in `mapping.cpp` skipped every style with `Side == NEGATIVE` outright, so if no `POSITIVE` / `BOTH` style was present the item ended up with no style at all.

## Fix
Keep preferring a `POSITIVE` or `BOTH` style for the visible front face (returned immediately, behavior unchanged), but remember a `NEGATIVE`-side style and its shading and use it as a fallback when no `POSITIVE`/`BOTH` style supplies the requested presentation type. The item is then no longer left unstyled.

## Verification (OCC 7.9.2, two-box IFC4 fixture)
One box styled red on the NEGATIVE side, one green on the POSITIVE side. Emitted OBJ/MTL:

| | NEG box | POS box |
|---|---|---|
| before | `DefaultMaterial` (Kd 0.7 0.7 0.7) | green (Kd 0 1 0) |
| after | its red (Kd 1 0 0) | green (Kd 0 1 0), unchanged |

The `POSITIVE`/`BOTH` early-return path is untouched, so styled items that already resolve a front-face style are unaffected.

### DIFF
--- a/src/ifcgeom/mapping/mapping.cpp
+++ b/src/ifcgeom/mapping/mapping.cpp
@@ -488,20 +488,41 @@ namespace {
 #endif
 
         IfcSchema::IfcSurfaceStyle *surface_style_ = nullptr;
+        // A surface style with Side == NEGATIVE targets the back (negative) face.
+        // When a POSITIVE or BOTH style is also present it is preferred for the
+        // visible front face, but a NEGATIVE style must still be used as a
+        // fallback so a styled item is not left completely unstyled (see #5618).
+        IfcSchema::IfcSurfaceStyle *negative_surface_style_ = nullptr;
+        std::pair<IfcSchema::IfcSurfaceStyle*, T*> negative_fallback{nullptr, nullptr};
         for (auto& style : prs_styles) {
             if (auto surface_style = style->as<IfcSchema::IfcSurfaceStyle>()) {
-                if (surface_style->Side() != IfcSchema::IfcSurfaceSide::IfcSurfaceSide_NEGATIVE) {
+                const bool is_negative = surface_style->Side() == IfcSchema::IfcSurfaceSide::IfcSurfaceSide_NEGATIVE;
+                if (is_negative) {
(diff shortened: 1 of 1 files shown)


BIMvoice:
Runtime-verified on OCC 7.9.2, commit cherry-picked clean onto v0.8.0. Two-box IFC4 fixture: one box styled red on the `.NEGATIVE.` side, one green on `.POSITIVE.`. IfcConvert OBJ/MTL, before vs after: the NEGATIVE box goes from `DefaultMaterial` (Kd 0.7 0.7 0.7, style silently dropped) to its own red `Kd 1 0 0`; the POSITIVE box stays `Kd 0 1 0` with the identical material record in both builds. The POSITIVE/BOTH early-return path is untouched - no regression on the already-styled front face. Claim reproduces exactly.

Issue #5473, opened 2024-09-24
Regression in swept disk solid in v0.8


Moult:
0.7 looks like this:

![2024-09-24-224558_547x267_scrot](https://github.com/user-attachments/assets/812d8008-18cc-4c2d-af47-752ecc1ca22e)

0.8 looks like this:

![2024-09-24-224607_825x421_scrot](https://github.com/user-attachments/assets/0b86aba6-4881-49db-a6bb-f4697952ae9c)

```
ISO-10303-21;
HEADER;
FILE_DESCRIPTION(('ViewDefinition [CoordinationView]'),'2;1');
FILE_NAME('','2024-09-24T22:38:00',(''),(''),'IfcOpenShell 0.8.1-a51ed6e6b','IfcOpenShell 0.8.1-a51ed6e6b','');
FILE_SCHEMA(('IFC4'));
ENDSEC;
DATA;
#1=IFCPROJECT('3ZHfd54197sOvwWETRXIhF',$,'project_name',$,$,$,$,(#11),#6);
#2=IFCDIMENSIONALEXPONENTS(1,0,0,0,0,0,0);
#3=IFCSIUNIT(*,.LENGTHUNIT.,$,.METRE.);
#4=IFCMEASUREWITHUNIT(IFCREAL(0.3048),#3);
#5=IFCCONVERSIONBASEDUNIT(#2,.LENGTHUNIT.,'foot',#4);
#6=IFCUNITASSIGNMENT((#3));
#7=IFCCARTESIANPOINT((0.,0.,0.));
#8=IFCDIRECTION((0.,0.,1.));
#9=IFCDIRECTION((1.,0.,0.));
#10=IFCAXIS2PLACEMENT3D(#7,#8,#9);
#11=IFCGEOMETRICREPRESENTATIONCONTEXT($,'Model',3,1.E-05,#10,$);
#12=IFCGEOMETRICREPRESENTATIONSUBCONTEXT('Body','Model',*,*,*,*,#11,$,.MODEL_VIEW.,$);
#13=IFCSITE('3TC2wPUMv5P9HYlVOMiPUK',$,'site_name',$,$,$,$,$,$,$,$,$,$,$);
#14=IFCBUILDING('2guf2ZNR94FQSt29s79yzN',$,'building_name',$,$,$,$,$,$,$,$,$);
[...]

Issue #5186, opened 2024-08-15
Issue with model conversion with IFCConvert 7


MonishVinceant1506:
Hi Team.
When the following model is converted to GLB using IFC convert 7, one of the part shapes is incorrect and it seem to miss a cut.
![image](https://github.com/user-attachments/assets/445dd366-cdf4-4b0d-9aca-679c7c72c059)
![image](https://github.com/user-attachments/assets/11d6e505-2628-47d6-a0b8-b577c94511ab)

Also, when tried converting it with IFC Convert 6, it is getting converted as expected. Please find the attached files for reference.

Please let us know if there a way to overcome this issue.
[Models.zip](https://github.com/user-attachments/files/16630093/Models.zip)

Regards,
Monish

smr02:
Have you tried exporting it again from Tekla? 
It seems to me like the Void is there but not being applied? Moving it makes it appear

https://github.com/user-attachments/assets/6938a3d5-ce77-4e69-9847-d39c79bda77a

MonishVinceant1506:
The IFC exported from Tekla is fine as show in the BHIM vision and the GLB converted using IFC Convert 6 is also correct. But this issue is occurring with the GLB converted with IFC Convert 7 as mentioned above.

smr02:
Ok, i've looked a bit further into this. 
I've managed to create a file that seems to be working. The only difference is two IfcDirection. 
There seems to be some weird rounding error? @Moult, can you make sense of this? 

![Screenshot from 2024-08-21 23-58-59](https://github.com/user-attachments/assets/f658be4b-dd3d-4f61-be58-b0cf9762da87)

![Screenshot from 2024-08-21 23-48-24](https://github.com/user-attachments/assets/cec7027f-a9c0-48a2-baad-834df8222b44)



### the original file: 
``` python
model = ifcopenshell.open('Model.ifc')
opening = model.by_id(552)
pprint( opening.ObjectPlacement.RelativePlacement.get_info() )
```

```
{'Axis': #166=IfcDirection((1.32348898008484E-23,-0.,1.)),
 'Location': #6=IfcCartesianPoint((0.,0.,0.)),
 'RefDirection': #167=IfcDirection((1.,0.,-1.32348898008484E-23)),
 'id': 548,
 'type': 'IfcAxis2Placement3D'}
```

### the edited file: 
```python
model = ifcopenshell.open('Model_test.ifc')
opening = model.by_id(552)
[...]

Issue #4118, opened 2023-12-15
Opening hole in wall failed


kkhung0829:
Hi, in this model the two holes failed to be subtracted from wall, in IfcConvert process.

![IssueSample BimVision](https://github.com/IfcOpenShell/IfcOpenShell/assets/22828883/7d288af0-7d43-436a-b45f-471521212f4b)

Attached below please find the ifc (zipped) and ifcconvert error log.
[IssueSample.zip.NoDoor.ifc.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/13680507/IssueSample.zip.NoDoor.ifc.zip)
[IssueSample.zip.NoDoor.ifc.log](https://github.com/IfcOpenShell/IfcOpenShell/files/13680536/IssueSample.zip.NoDoor.ifc.log)

Is there something wrong with the IFC or has OpenCascade problems with it? Thanks

pbbr:
@Andrej730 @aothms Could be the same issue as https://github.com/IfcOpenShell/IfcOpenShell/issues/3952?

aothms:
This is a bit of a peculiar model

Walls in IFC are supposed to be broken up into linear/continuous segments. This model has the entire wall envelope as a single element.

The real issue is that the geometry is not manifold, geometrically it may look somewhat ok, although also with gaps, but topologically the edges are not oriented pairs. 

![afbeelding](https://github.com/IfcOpenShell/IfcOpenShell/assets/1096535/260e51d8-7ce8-4569-b7f4-4c5474b4a517)

![afbeelding](https://github.com/IfcOpenShell/IfcOpenShell/assets/1096535/8488f754-a4fa-4458-ab37-16c8d1465f57)

This may work well in some of the mesh-based implementations, but IfcOpenShell really expects valid manifold bodies when doing opening subtractions.

kkhung0829:
@aothms Thank you for your input. Please correct me if I'm mistaken. It seems that the model (wall) generated by IFCEXTRUDEAREASOLID is not manifold, as the edges are not oriented in pairs. As a result, IFCConvert is unable to process it, am I right? If that is the case, it would be great if you have any idea on how to avoid encountering such issues when calling IFCEXTRUDEAREASOLID.

Thank you very much for your assistance.

kkhung0829:
@aothms Thanks for your input. I would also like to ask what's the meaning of "oriented pair edges" ? Do you know how to fix the edges to be oriented pair ? Thanks a lot.

[...]

Issue #619, opened 2019-06-17
Missing subtraction on wall (example file)


johanrd:
2. Convert [619--wall--missing-subtraction.ifc](https://github.com/IfcOpenShell/files/blob/master/619--wall--missing-subtraction.ifc) with 12e9f7b941e0fca8d4f5ee97e5204c2c5c9eb30c (v0.6.0 June 15) and experience that there is a missing subtraction on the nearest edge (see image):
<img width="484" alt="Screenshot 2019-06-17 at 16 34 43" src="https://user-images.githubusercontent.com/4601554/59612696-d7ac8e80-911d-11e9-8c61-fbd990a2ea47.png">

```
[Error] {2t1PYAdRTF49828D6ad_ME} Failed to process subtraction:
#111035=IfcBooleanClippingResult(.DIFFERENCE.,#111029,#111034)
```

2. Expected geometry from Solibri:
<img width="722" alt="Screenshot 2019-06-17 at 16 34 29" src="https://user-images.githubusercontent.com/4601554/59612709-ded39c80-911d-11e9-87d3-568cca47e12c.png">

(I have yet to find an ifcopenshell version that outputs expected geometry, thus the input file is not augmented with `--calculate-quantities`)

Moult:
Checked in latest v0.7.0 and issue still present for the record.

PR #8499, opened 2026-07-11
Retry boolean CUT that OCCT reports done but subtracts no volume (#5630)

### LINKED ISSUE #5630: Void not voiding
Weird scenario where the site object has a void applied to it, but it does show up. 

[Lake_Kegonsa_Renovation.ifc.txt](https://github.com/user-attachments/files/17483698/Lake_Kegonsa_Renovation.ifc.txt)

https://github.com/user-attachments/assets/d252654c-ffa0-417f-ae41-2fd415d054b8

BIMvoice:
Fixes #5630.

## Problem
An `IfcOpeningElement` that voids an `IfcGeographicElement` terrain (a dense `IfcPolygonalFaceSet`) does not cut: the opening solid overlaps the terrain, yet the converted terrain keeps its full volume.

## Root cause
The terrain converts to a valid closed solid and the opening is a proper solid box whose XY sits inside the terrain and whose Z spans its full thickness, so a large removal is expected. But OCCT `BRepAlgoAPI_Cut` returns `IsDone` at the tight default fuzzy value while subtracting essentially nothing, and because it reports success the existing escalating-fuzziness retry never runs. Sweeping fuzziness confirms it: at fuzz >= 5e-4 the cut removes ~17%, at fuzz <= 1e-4 it is a no-op.

## Fix
After a CUT is deemed valid, and only while a higher-fuzziness retry is still permitted, compare the result volume to the first operand. If the removed fraction is negligible (`< 1e-4` of operand A), treat the CUT as a failure so the existing escalation runs and finds a fuzzy value at which the section is actually computed. When no retry is left the result is accepted as before, so a genuinely tiny legitimate cut can never lose its outcome. New diagnostic `GEO 156`.

## Verification (OCC 7.9.2)
| | volume | verts | manifold |
|---|---|---|---|
| terrain before | 7330.878 (no hole) | 459 | closed |
| terrain after | 6058.593 (real hole, -1272) | 377 | closed |

Regression: five currently-passing cut models (#619, #4118, #5473, #5186, #511) are byte-identical before and after, and across the full 1447-object source file exactly one product (the terrain) changes. The guard only fires when OCCT reports a successful CUT that removed ~nothing, so passing cuts are untouched.

### DIFF
--- a/src/ifcgeom/kernels/opencascade/boolean_utils.cpp
+++ b/src/ifcgeom/kernels/opencascade/boolean_utils.cpp
@@ -1218,6 +1218,24 @@ bool IfcGeom::util::boolean_operation(const boolean_settings& settings, const To
 				}
 			}
 
+			// #5630 A solid-solid CUT whose tool operands survived the disjoint/
+			// touching/narrow elimination above is expected to remove material. On
+			// dense tessellated first operands (e.g. IfcPolygonalFaceSet terrain)
+			// OCCT's BOP can silently return an essentially unchanged result at a
+			// tight fuzzy value, reporting IsDone yet subtracting ~nothing. Detect
+			// this near zero volume removal and, while a higher fuzziness retry is
+			// still permitted, treat it as a failure so the existing escalation can
+			// find a fuzzy value at which the section is actually computed. When no
+			// retry is left the result is accepted as is, preserving prior behaviour.
+			if (success && op == BOPAlgo_CUT && allow_retry) {
+				const double va = shape_volume(a);
+				const double vr = shape_volume(r);
+				if (va > 1.e-9 && (va - vr) < va * 1.e-4) {
+					success = false;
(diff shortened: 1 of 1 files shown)


BIMvoice:
Runtime-verified on OCC 7.9.2: cherry-picked this commit onto current `v0.8.0`, clean. Isolating the terrain (GlobalId `2wL7...`) from `opening_not_applied.ifc`: baseline yields a closed solid of 459 welded verts, signed volume 7330.878 with the opening removing nothing; with the fix the same terrain is 377 verts, volume 6058.593 - the opening now removes 1272.285 (~17%), a real hole, still a closed manifold. Reproduces the PR's 459->377 / 7330.878->6058.593 to the digit. A normal cut model is byte-identical before/after, so the guard only fires when OCCT reports a done CUT that subtracted ~nothing.

PR #8501, opened 2026-07-11
Bonsai: pick the true sweep edge in detect_extrusion_edge (#5898)

### LINKED ISSUE #5898: If you convert the slabs using bpy.ops.bim.update_representation(ifc_representation_class="IfcExtrudedAreaSolid/IfcArbitraryProfileDefWithVoids") some of the geometry is incorrrect.
If you convert the slabs using `bpy.ops.bim.update_representation(ifc_representation_class="IfcExtrudedAreaSolid/IfcArbitraryProfileDefWithVoids")` some of the geometry is incorrrect. 

[Twisted Tower-02 - Copy.ifc.txt](https://github.com/user-attachments/files/18219035/Twisted.Tower-02.-.Copy.ifc.txt)

video: https://community.osarch.org/uploads/editor/d8/4im8y3f86azq.mp4

from this discussion: https://community.osarch.org/discussion/2672/how-to-asign-ifctype-names-to-model-elements-in-bonsai#latest

BIMvoice:
Fixes #5898.

## Problem
Converting slabs to `IfcArbitraryProfileDefWithVoids` renders incorrect geometry: 12 of the 69 slabs in the reported sample collapse from ~2500 volume to ~0.

## Root cause
Not the C++ WithVoids path. That path is correct: a hand-built square-with-square-hole extrusion renders volume 168.0 = (100-16)x2 exactly, closed and manifold. The defect is in Bonsai's `Helper.detect_extrusion_edge` (`geometry/helper.py`): it returned the first mesh edge that leaves the profile plane by more than 1e-6. For the bad slabs the first such edge (in arbitrary bmesh order) is a nearly lateral diagonal that only marginally clears the threshold, so the resulting `IfcExtrudedAreaSolid.ExtrudedDirection` lies in the profile plane (local Z = 0) with a bogus large depth. Extruding a planar profile along an in-plane direction is a legitimately degenerate, zero-volume sliver, which the C++ renders faithfully.

## Fix
Scan all candidate edges and return the one whose direction is most aligned with the profile-face normal, i.e. the genuine through-thickness sweep edge, keeping the same `[in_profile_vertex, off_plane_vertex]` return orientation. This benefits all four auto-detect callers.

## Verification (headless Blender 5.1 / Bonsai 0.8.6, local IfcConvert OCC 7.9.2)
Re-derived all 69 slabs, saved, rendered, and compared per-slab signed-tetra volume against the original Brep ground truth:
- before: 12 slabs degenerate (e.g. 2570.70 -> 0.0055)
- after: all 69 match Brep to 5 to 6 significant figures (worst 0.01%); the degenerate 12 are restored (0.0055 -> 2570.70) and the 57 already-correct slabs are volume-identical.

Verified with the actual staged file installed into Bonsai, not a monkeypatch.

### DIFF
--- a/src/bonsai/bonsai/bim/module/geometry/helper.py
+++ b/src/bonsai/bonsai/bim/module/geometry/helper.py
@@ -316,10 +316,18 @@ class Helper:
         return {"outer_curve": outer_loop, "inner_curves": inner_loops}
 
     # An extrusion edge is an edge that shares a single vertex with a profile
-    # face and is not on the plane of the face.
+    # face and is not on the plane of the face. Multiple such edges may exist
+    # (e.g. a slanted/stepped mesh has boundary edges that leave the profile
+    # plane only marginally); the genuine sweep edge is the one whose direction
+    # is most aligned with the profile normal. Picking the first candidate in
+    # arbitrary bmesh order could select a nearly in-plane edge, yielding a
+    # degenerate (zero-thickness) extrusion. So we pick the best-aligned edge.
     def detect_extrusion_edge(self, bm: bmesh.types.BMesh, profile_face: bmesh.types.BMFace) -> Union[list[int], None]:
         bm.edges.ensure_lookup_table()
         face_verts_set = set(profile_face.verts)
+        normal = profile_face.normal
+        best_edge = None
+        best_alignment = 0.0
         for edge in bm.edges:
             unshared_verts = set(edge.verts) - face_verts_set
             if len(unshared_verts) == 1:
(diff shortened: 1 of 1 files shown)


BIMvoice:
Verified the core fix (`Helper.detect_extrusion_edge`) live in headless Blender 5.1.2, loading the pre-fix and post-fix `helper.py` side by side. (I verified the narrow unit, not the full 69-slab file.) On a square profile face (normal +Z) with two off-plane candidate edges - a near-in-plane diagonal that only marginally clears the 1e-6 threshold (appearing first in bmesh order) and the genuine +Z sweep edge - the baseline returns the diagonal (direction `(1,1,0)`, alignment to the profile normal `0.000001`), i.e. an in-plane ExtrudedDirection that produces the degenerate zero-volume slab. The patched code returns the true sweep edge (direction `(0,0,1)`, alignment `1.000000`), keeping the `[in_profile_vertex, off_plane_vertex]` orientation. This confirms the mechanism behind the collapsed slabs and the fix.

PR #8502, opened 2026-07-11
Fix #1367: map IfcAsymmetricIShapeProfileDef in IFC4+

### LINKED ISSUE #1367: IfcAsymmetricIShapeProfileDef not implemented for IFC4
Because it's no longer a subtype of IfcIShapeProfileDef

BIMvoice:
Fixes #1367.

## Problem
In IFC2x3 `IfcAsymmetricIShapeProfileDef` is a subtype of `IfcIShapeProfileDef` and was covered by that mapping (the dispatch uses `inst->as<T>()`, which respects inheritance). In **IFC4 it was re-parented directly to `IfcParameterizedProfileDef`**, so nothing binds it and `mapping::map()` falls through to:

```
[Error] GEO307 No operation defined for: IfcAsymmetricIShapeProfileDef(...)
```

producing empty geometry. The inline `is(IfcAsymmetricIShapeProfileDef)` branch inside the I-shape mapper is unreachable for standalone IFC4 instances, and it reads `OverallWidth()` which IFC4 asymmetric profiles do not have (they carry `BottomFlangeWidth`/`TopFlangeWidth`).

## Fix
Add a dedicated `map_impl(const IfcAsymmetricIShapeProfileDef*)`, guarded by `SCHEMA_IfcAsymmetricIShapeProfileDef_HAS_BottomFlangeWidth` (defined only where the type is standalone, i.e. IFC4+), that builds the section from independent bottom/top flange width, thickness, fillet radius, edge radius and slope. Bind it under the same guard. IFC2x3 keeps using the existing I-shape path unchanged.

## Verification (local build 0.8.6, OCC 7.9.2)
Extruded an IFC4 `IfcAsymmetricIShapeProfileDef` (BottomFlangeWidth 0.2, TopFlangeWidth 0.1, depth 0.4):

| | before | after |
|---|---|---|
| verts | 0 (GEO307) | 24 |
| bottom flange width | - | 0.200 (as specified) |
| top flange width | - | 0.100 (as specified) |

IFC2x3 asymmetric profile: still 24 verts, correct 0.2/0.1 widths, no error (no regression).

### DIFF
--- a/src/ifcgeom/mapping/IfcIShapeProfileDef.cpp
+++ b/src/ifcgeom/mapping/IfcIShapeProfileDef.cpp
@@ -109,3 +109,56 @@ taxonomy::ptr mapping::map_impl(const IfcSchema::IfcIShapeProfileDef* inst) {
 		{{-x1,-y + ft1}, {fe1}}
 	});
 }
+
+#ifdef SCHEMA_IfcAsymmetricIShapeProfileDef_HAS_BottomFlangeWidth
+// In IFC4 IfcAsymmetricIShapeProfileDef is no longer a subtype of
+// IfcIShapeProfileDef, so it has its own dedicated mapping. Bottom and top
+// flanges may differ in width, thickness, fillet radius, edge radius and slope.
+taxonomy::ptr mapping::map_impl(const IfcSchema::IfcAsymmetricIShapeProfileDef* inst) {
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing: superseded by 438c0955f2 (merged), the same #1367 IfcAsymmetricIShapeProfileDef fix in a different file split.

PR #8503, opened 2026-07-11
Return nan for non-manifold volume and skip it in QTO (#6125)

### LINKED ISSUE #6125: Bonsai volume calculation of non-manifold Tessellations
This is from a received IFC project, there are three slabs modelled as Tessellations: [slabs-only.ifc.txt](https://github.com/user-attachments/files/18722067/slabs-only.ifc.txt)

The qto quantity calculation seems ok, except that the volume calculated via ifcopenshell and blender is wildly off (it is much larger than the volume of the bounding box). This is because the meshes are non-manifold, fixing the meshes results in reasonable calculations.

Image

It would be better to skip the volume calculation for non-manifold tessellations entirely.

BIMvoice:
Fixes #6125.

## Problem
`util.shape.get_volume` wildly overestimates the volume of a non-manifold tessellated mesh. For the reported slabs (IfcPolygonalFaceSet) it returns 17.71 m3 against a 1.82 m3 bounding box, a 9.7x overestimate.

## Root cause
`get_volume` sums signed tetrahedra (divergence theorem), which is only meaningful for a closed watertight mesh. The reported slabs are non-manifold (19 of 152 edges are not shared by exactly two triangles), so the sum is meaningless. The value was still returned and, via the ifc5d QTO calculator, written into the IFC as a Volume quantity.

## Fix
- `util/shape.py`: `get_volume` runs an O(faces) edge-parity check and returns `nan` when the mesh is not a closed manifold, rather than a bogus number. This matches the maintainer suggestion in the thread to guard volume for non-manifold meshes.
- `ifc5d/qto.py`: the QTO value loop skips a `nan` result (import-free self-inequality test) so a non-manifold element gets no volume quantity instead of `nan` in the IFC.

## Verification (OCC 7.9.2)
- the three reported slabs: `get_volume` 17.71 -> `nan`, and no `NetVolume` is written by the QTO run.
- regression: a closed manifold 2x3x0.5 box still returns exactly 3.0 and its `NetVolume` is written normally.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/shape.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/shape.py
@@ -68,14 +68,43 @@ def is_x(value: float, x: float, tolerance: Optional[float] = None) -> bool:
     return abs(x - value) < tolerance
 
 
+def is_manifold(geometry: W.Triangulation) -> bool:
+    """Checks whether a triangulated geometry is a closed, consistently oriented manifold
+
+    Two conditions are checked for every edge of every triangle:
+
+    - Unoriented use: as an unordered pair of vertices, an edge must be shared
+      by exactly two triangles. A count of 1 means an open hole or boundary,
+      a count above 2 means more than two triangles meet at that edge.
+    - Oriented use: as an ordered pair of vertices, an edge must be used by at
+      most one triangle. If two triangles use the same ordered edge, their
+      windings are inconsistent (e.g. a flipped or duplicated face), which
+      also invalidates volume calculations that rely on consistent winding.
+
+    :param geometry: Geometry output calculated by IfcOpenShell
+    :return: ``True`` if the geometry is a closed, consistently oriented manifold
+    """
+    faces = geometry.faces
+    directed_use: dict[tuple[int, int], int] = {}
+    undirected_use: dict[tuple[int, int], int] = {}
+    for i in range(0, len(faces), 3):
+        tri = (faces[i], faces[i + 1], faces[i + 2])
+        for a, b in ((tri[0], tri[1]), (tri[1], tri[2]), (tri[2], tri[0])):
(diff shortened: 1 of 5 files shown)


src/ifcopenshell-python/ifcopenshell/util/shape.py
```diff
@@ -68,13 +68,46 @@ def is_x(value: float, x: float, tolerance: Optional[float] = None) -> bool:
     return abs(x - value) < tolerance
 
 
+def is_manifold(geometry: W.Triangulation) -> bool:
```
aothms, line 71: Consider something like this which has less lookups, consumes less ram and can fail early.

```
def is_manifold(geometry: W.Triangulation) -> bool:
    """Check whether the triangulation is closed and consistently oriented."""
    faces = geometry.faces
    edge_state: dict[tuple[int, int], int] = {}

    for i in range(0, len(faces), 3):
        a = faces[i]
        b = faces[i + 1]
        c = faces[i + 2]

        for u, v in ((a, b), (b, c), (c, a)):
            if u == v:
                return False

            if u < v:
                edge = (u, v)
                direction = 1
            else:
                edge = (v, u)
                direction = -1

            state = edge_state.get(edge)

            if state is None:
                # First use: remember its orientation.
                edge_state[edge] = direction
            elif state == -direction:
                # Second use in the opposite direction: edge is complete.
                edge_state[edge] = 0
            else:
                # Same orientation twice, or any third use.
                return False

    # Non-zero entries are boundary edges used only once.
    return all(state == 0 for state in edge_state.values())
```

BIMvoice:
@aothms applied, verified against the existing test_shape.py suite (all 6 pass) and traced through the edge cases (degenerate zero-length edge, same-direction reuse, third use, boundary/open edge) to confirm it's behaviorally equivalent to the two-dict version, just fewer lookups/less memory and an earlier return on a definite defect.

PR #8504, opened 2026-07-11
Warn when a face inner boundary intersects another boundary (#527)

### LINKED ISSUE #527: Test for inner face boundary intersections.
See #513 for an example where the inner boundaries intersect with the outer boundary. Currently no errors are produced in this case, but the face is simply not triangulated.

BIMvoice:
Fixes #527.

## Problem
A face whose inner boundary crosses the outer boundary (or another inner boundary) is invalid, but IfcOpenShell handles it silently: the 2018 report saw the face dropped (0 triangles) with no error; on the current line Open Cascade instead heals it into wrong geometry, still with no diagnostic. Either way the user gets no signal that their input face is malformed.

## Fix
After the wires are collected, if a face has inner boundaries, measure the `BRepExtrema_DistShapeShape` distance between each inner wire and every earlier wire (outer plus prior inners). Two non intersecting loops have strictly positive distance, so a distance at or below the modelling precision means the boundaries touch or cross; emit `GEO 402` naming the offending face. Diagnostic only, no geometry change. The check is gated on faces that actually have inner wires.

The warning is emitted via the kernel `logger()` rather than `Logger::Root()`. IfcConvert configures a local `Logger` and worker logs merge into it, whereas `Logger::Root()` is a separate unconfigured singleton whose messages are discarded (a latent issue that also affects some existing GEO messages, worth a separate look).

## Verification (OCC 7.9.2, synthesized IFC4 faces, run with -v)
| case | GEO 402 | note |
|---|---|---|
| valid 4x4 hole | 0 | triangulates identically, area 84.0 |
| inner triangle crossing outer right edge | 1 | flagged |
| inner straddling bottom edge | 1 | flagged |
| inner self-intersecting (bowtie), no crossing | 0 | separate class, left untouched |
| inner containing outer | 0 | separate class, left untouched |

No regression on the valid holed face (identical output, no new warning) in both sequential and multithreaded runs.

### DIFF
--- a/src/ifcgeom/kernels/opencascade/face.cpp
+++ b/src/ifcgeom/kernels/opencascade/face.cpp
@@ -31,6 +31,7 @@
 #include <ShapeFix_Shape.hxx>
 #include <ShapeFix_ShapeTolerance.hxx>
 #include <BRep_Tool.hxx>
+#include <BRepExtrema_DistShapeShape.hxx>
 
 #include <Standard_Macro.hxx>
 #include <TopoDS_Shape.hxx>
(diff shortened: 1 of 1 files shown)


aothms:
Note to self. DistShapeShape is slow. We should benchmark performance.

BIMvoice:
Good call on the performance. I added a bounding-box pre-filter in front of BRepExtrema_DistShapeShape: one Bnd_Box per wire (enlarged by precision_), and a wire pair whose boxes are more than precision_ apart is rejected in O(1). Since the box gap is a lower bound on the true wire distance, any skipped pair could never have triggered GEO 402, so the result is identical (bad_inner and straddle still emit one warning, good_hole none). On a synthesized 100-hole face this rejects 4950 of 5050 pairs without a distance call (about 98%), and the saving grows with boundary edge count. So the diagnostic now only runs DistShapeShape on genuinely overlapping boundary pairs.

PR #8505, opened 2026-07-11
Support appending IfcGroup assets via project.append_asset (#8162)

### LINKED ISSUE #8162: Import IfcGroups via the 'Project Library'.
Would be nice to import IfcGroups via the 'Project Library'. 

Image

BIMvoice:
Fixes #8162.

## Problem
`project.append_asset` could not import an `IfcGroup` from a project library. `IfcGroup` was absent from `APPENDABLE_ASSET`, so `execute()` fell through every `elif` branch and returned `None`, appending nothing (destination: `groups=0 walls=0 rels=0`).

## Fix
- Add `IfcGroup` to `APPENDABLE_ASSET` and dispatch it to a new `append_group()`.
- `append_group()` mirrors `append_product()`: it whitelists the group's `IsGroupedBy` inverse so the members reached through `IfcRelAssignsToGroup` are copied via the regular element-append path (carrying their own materials, properties and representations), and the relationship is recreated in the destination linking the copied group to the copied members.
- Idempotency comes from the existing GUID-based reuse, so re-appending the same group does not duplicate it.

## Verification (headless)
Appending a group of three walls:
- before: `groups=0 walls=0 rels=0`
- after: `groups=1 walls=3 rels=1`, with `RelatingGroup` and the three `RelatedObjects` correctly wired
- re-append: still `1/3/1` (no duplication)
- regression: appending a plain `IfcProduct` still returns the product and appends it unchanged

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py
@@ -39,6 +39,7 @@ APPENDABLE_ASSET = Literal[
     "IfcCostSchedule",
     "IfcProfileDef",
     "IfcPresentationStyle",
+    "IfcGroup",
 ]
 APPENDABLE_ASSET_TYPES: tuple[APPENDABLE_ASSET, ...] = get_args(APPENDABLE_ASSET)
 MATERIAL_SETS = ("IfcMaterialLayerSet", "IfcMaterialConstituentSet", "IfcMaterialProfileSet")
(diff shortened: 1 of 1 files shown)


BIMvoice:
Runtime-verified on branch `fix-8162-append-group` (python 3.13, installed ifcopenshell wrapper), comparing the fix commit against its parent. Built a library file with an `IfcGroup` grouping 3 `IfcWall`s and appended it via `project.append_asset` into an empty target. Before the fix `IfcGroup` is not in `APPENDABLE_ASSET_TYPES`, `append_asset` returns `None`, and the target stays `groups=0 walls=0 rels=0`. After the fix it returns the copied group and the target is `groups=1 walls=3 rels=1`, with the recreated `IfcRelAssignsToGroup` correctly wired (`RelatingGroup=MyGroup`, 3 `RelatedObjects`); re-appending the same group is a no-op (still `1/3/1`). Matches the PR claim exactly.

PR #8506, opened 2026-07-11
Fix IfcSweptDiskSolidPolygonal fillet: crash to correct rounded corners (#4094)

### LINKED ISSUE #4094: IfcSweptDiskSolidPolygonal fillet issue with BlenderBIM addon
I was testing the files from [https://github.com/IfcOpenShell/IfcOpenShell/issues/2322](url) for IfcSweptDiskSolidPolygonal on BlenderBIM addon blenderbim-231208. The curved FilletRadius didn't reflect in the representation. The same however, works very well on FreeCAD. 
Attaching screenshot below
Is this a bug?
ifc

BIMvoice:
Fixes #4094.

## Problem
The issue reports that FilletRadius on an `IfcSweptDiskSolidPolygonal` with a 3D directrix is not reflected. In fact it is worse: on the current line every `IfcSweptDiskSolidPolygonal` carrying a `FilletRadius` crashes with `GEO027 BRepAdaptor_Curve::No geometry` and emits no geometry at all.

## Root cause
The branch the issue points at ("not polyhedral, ignoring FilletRadius") is commented-out dead code. The active fillet path is `fillet_loop()` in `profile_helper.cpp`, which was unfinished (`@todo untested from here`). Four defects:
1. `polygon_from_points()` shares one point instance between adjacent edges; filleting insets it twice so the arc edge collapses to `start == end`.
2. `(b - 1) % size` with unsigned `size` mis-indexes the previous edge at `b == 0`.
3. Every vertex was filleted, including the two open-polyline endpoints; at an endpoint the corner is degenerate (`ba.cross(bc) == 0`), producing a zero-geometry circle, which is the `GEO027` crash.
4. The arc centre was offset along the plane normal (out of plane) instead of the interior bisector, and the axis sign selected the major arc.

## Fix
`fillet_loop()` now deduplicates the shared corner points, uses `(b + n - 1) % n`, detects an open directrix geometrically and skips its endpoints, places the arc centre on the bisector at `radius / sin(angle / 2)`, derives the axis from the two tangent points so the minor (fillet) arc is swept, and clamps the `acos` argument. A polyhedral guard logs `GEO328` and skips a genuinely curved directrix instead of mangling it.

## Verification (OCC 7.9.2, welded OBJ)
| case | baseline | fixed |
|---|---|---|
| reported 3D-poly file (R=0.2) | GEO027 crash, 0 verts | 290 verts, rounded |
| well-fitting radius R=0.1 | crash | 520 verts, closed manifold (0/0) |
| no-fillet control | converts | unchanged, closed manifold |

The reported file's R=0.2 is degenerately large for its 0.3 middle edge (adjacent fillets overlap), so that one patch is open; well-fitting radii are closed manifolds. A curved line+arc directrix no longer crashes and yields a valid manifold.

### DIFF
--- a/src/ifcgeom/profile_helper.cpp
+++ b/src/ifcgeom/profile_helper.cpp
@@ -3,48 +3,92 @@
 using namespace ifcopenshell::geometry;
 
 taxonomy::loop::ptr ifcopenshell::geometry::fillet_loop(taxonomy::loop::ptr loop, double radius) {
-	std::vector<profile_point_with_edges_3d> pps(loop->children.size());
-	for (int b = 0; b < loop->children.size(); ++b) {
-		int c = (b - 1) % loop->children.size();
-		pps[b] = { 
-			boost::get<taxonomy::point3::ptr>(loop->children[c]->start)->ccomponents(), 
+	const size_t n = loop->children.size();
+	if (n < 2) {
+		return loop;
+	}
+
+	// FilletRadius is only meaningful for a polygonal (straight segmented)
+	// directrix. If any segment carries a curved basis or is parameterised rather
+	// than defined by its end points, the directrix is not polyhedral: leave it
+	// untouched instead of misinterpreting arc segments as corners.
+	if (!loop->is_polyhedron()) {
+		Logger::Root().Warning("GEO", 328, "Directrix is not polyhedral, ignoring FilletRadius");
+		return loop;
+	}
+	for (const auto& e : loop->children) {
+		// which() == 1 selects the point3::ptr alternative of the start/end variant.
+		if (e->start.which() != 1 || e->end.which() != 1) {
+			Logger::Root().Warning("GEO", 328, "Directrix is not polyhedral, ignoring FilletRadius");
+			return loop;
+		}
+	}
+
+	// polygon_from_points() shares a single point instance between the end of one
+	// edge and the start of the next. Filleting insets those two vertices in
+	// opposite directions, so break the sharing first by giving every edge its own
+	// end point (mirrors what profile_helper() does for the 2d case).
+	for (auto& e : loop->children) {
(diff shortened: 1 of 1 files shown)


aothms:
I don't think this issue warrants fixes on the C++ side. The issue specifically mentioned "BlenderBIM has a dedicated code path for swept disks" / Bonsai specific code paths in python. At that FreeCAD which uses the C++ code from IfcOpenShell directly faces no issues.

BIMvoice:
Fair point on where this belongs, and I will defer to your call. One data point in case it is useful: this is not only a Bonsai path. Running the attached sample through plain IfcConvert (no Bonsai, OCC 7.9.2) logs `[Error] [GEO027] BRepAdaptor_Curve::No geometry` and emits zero geometry for the filleted solid, i.e. the C++ fillet_loop path is reached and fails on a valid IfcSweptDiskSolidPolygonal with a FilletRadius, rather than a downstream tool handling it. With the PR the same file converts to a closed manifold rounded solid.

That said, if the intended design is that FilletRadius is resolved in the authoring tool and the core kernel should leave these unfilleted (or emit a clearer diagnostic instead of computing them), I am happy to close this PR, or to reduce it to just a clean skip-with-warning rather than the full fillet implementation. Let me know which you prefer.

BIMvoice:
Runtime-verified on OCC 7.9.2: cherry-picked this commit onto current `v0.8.0`, clean. On the reported `fail-x-300mm-y-500mm-3d-poly.ifc` the baseline crashes with GEO027 (BRepAdaptor_Curve: no geometry) and emits zero geometry; with the fix it converts (289 welded verts, rounded - open because R=0.2 overlaps the 0.3 m edge, as the PR notes). Refitting to R=0.1 gives 520 welded verts, a closed manifold (0 non-manifold edges); the no-FilletRadius control is byte-identical to baseline. The crash is fixed and valid fillets round correctly; GEO328 does not spuriously fire on polygonal directrixes.

PR #8509, opened 2026-07-11
Door/Opening documentation improvement 

Vanuan:
See #8411

### DIFF
--- a/src/bonsai/docs/guides/authoring/basic_modeling/door.rst
+++ b/src/bonsai/docs/guides/authoring/basic_modeling/door.rst
@@ -1,154 +1,145 @@
-Door
-====
+Door  
+====  
+  
+A door is a wall opening that allows passage between spaces.  
+  
+In IFC, it's represented by the ``IfcDoor`` entity. A door always requires a host wall and a corresponding ``IfcOpeningElement`` cut into that wall.  
+  
+.. note::  
+   This guide applies equally to **Windows** (``IfcWindow``).  
+  
+Adding a Door  
+-------------  
+  
+1. **Select the wall** where you want to place the door.  
+2. (Optional) Set the 3D cursor on the wall at the desired position.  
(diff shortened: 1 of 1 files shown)


BIMvoice (commented):
Reviewed, and this is a clear improvement over my smaller PR #8411. I checked the technical claims against the source and they are accurate:

- **Regen with the door selected repositions the opening but does not change its shape.** Confirmed: it runs `bim.recalculate_fill` (RecalculateFill in bim/module/model/opening.py), which copies the door's placement onto the opening and recuts the wall.
- **Regen with the wall selected only refreshes the wall and leaves opening placements untouched.** Correct.
- **Apply Void recreates the opening shape and safely removes any existing opening first, so it is safe to use multiple times.** Confirmed in `FilledOpeningGenerator.generate` (bim/module/model/opening.py): `if filling.FillsVoids: ifcopenshell.api.feature.remove_feature(..., feature=filling.FillsVoids[0].RelatingOpeningElement)` before generating the new one. This is exactly why it does not create duplicate openings (the question from #8411).
- **Manual removal via `ifcopenshell.api.feature.remove_feature`** is the right call and does remove the void relationship and the opening entity together.
- The "Apply Void" label is correct for the tool header (workspace.py shows `bim.add_opening` with text "Apply Void"); it appears as "Add Opening" only in the Voids panel, which your Deleting section already uses correctly.

The decision matrix (Regen = reposition, Apply Void = alter shape) is the clearest framing of this I have seen, and the UX / under-the-hood split is exactly what the issue asked for. Nice work.

I will close my #8411 in favour of this. One tiny optional nit: the manual-cleanup snippet uses `tool.Ifc.get()` / `tool.Ifc.get_entity` and `bpy`, so a one-line note that it runs in Bonsai's Python console (where `tool` and `bpy` are in scope) would help a reader who copies it elsewhere.

PR #8510, opened 2026-07-11
Claude specific file, automatically read on opening the repo

sboddy:
Ensures AGENTS.md is read by Claude for this repo.
Contains details useful for Claude.
Could be enhanced with further additions.
Feel free to suggest any improvements you have discovered or added.
Over time may become out of date and require a refresh.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,122 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Repository conventions (read AGENTS.md)
+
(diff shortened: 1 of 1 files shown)


sboddy:
I'm closing this as the wrong approach. As I've learned more I've discovered that CLAUDE.md should now be kept very small and tight. There are better approaches. The one key thing is that people have to understand that AGENTS.md will **not** be read by Claude unless it is explicitly told to read it. That could perhaps be documented in the developer documentation.

aothms:
> will **not** be read by

Coincidentally I think this is not true anymore since 2 days ago: https://news.ycombinator.com/item?id=49760187

PR #8511, opened 2026-07-11
Expand wildcard attributes in IfcCsv export (#3424)

### LINKED ISSUE #3424: IfcCsv query for All the attributes + Psets
It would be useful to have in IfcCsv a query that exports all the attributes present in the objects without write them manually.
Plus have the opportunity to export Psets or custom Psets would be nice too.

BIMvoice:
Fixes #3424.

## Problem
Exporting with a `*` attribute (or `Pset_Name.*`) emits a literal `*` column filled with the null placeholder instead of expanding to real columns. The `get_wildcard_attributes` helper had no callers after a refactor dropped the expansion step.

## Fix
`export()` now calls a new `expand_wildcard_attributes` before the per-element loop:
- non-wildcard queries pass through unchanged (position and custom header preserved);
- a bare `*` expands to every direct attribute of the exported elements (new `get_element_attributes` helper);
- `Pset_Name.*` / `Qto_Name.*` expand via the existing `get_wildcard_attributes` helper;
- expanded columns that duplicate an existing one (e.g. `GlobalId`, an explicit `Name`) are skipped.

Bare `*` = all direct attributes and `Pset_Name.*` = that pset's properties. A wildcard for "every pset on the element" is out of scope (would need a separate token).

## Verification (headless)
| query | before | after |
|---|---|---|
| `["Name","*"]` | `[GlobalId, *]` → literal `*`, value `-` | `[GlobalId, Name, ..., Tag, PredefinedType]` with real values |
| `["Pset_WallCommon.*"]` | `[GlobalId, Pset_WallCommon.*]` → `-` | `[..., Pset_WallCommon.IsExternal, Pset_WallCommon.LoadBearing]` with YES/NO |
| `["Name","Description","Tag"]` | unchanged | unchanged |

### DIFF
--- a/src/ifccsv/ifccsv.py
+++ b/src/ifccsv/ifccsv.py
@@ -105,6 +105,9 @@ class IfcCsv:
             attributes.insert(0, "GlobalId")
             headers.insert(0, "GlobalId")
 
+        elements = list(elements)
+        attributes, headers = self.expand_wildcard_attributes(attributes, headers, elements)
+
         for element in elements:
             result = []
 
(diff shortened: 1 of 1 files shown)


BIMvoice:
Verified headless (python 3.13), HEAD vs parent. On the parent, exporting an IfcWall with the query `["Name","*"]` produces a literal `*` column filled with the null placeholder (`['GlobalId','Name','*']`); on HEAD the bare `*` expands to every direct attribute with real values (`Tag='T1'`, etc.), and `Pset_WallCommon.*` expands to `Pset_WallCommon.LoadBearing` / `Pset_WallCommon.IsExternal`, while an explicit named-attribute export is byte-identical. Matches the PR's table.

PR #8516, opened 2026-07-11
Bonsai: re-orthonormalize item placement in Item Mode (#6996)

### LINKED ISSUE #6996: IFC Item Mode breaks IfcShapeRepresentation
### Bug Description

I got a **IfcShapeRepresentation** consisting of **multiple IfcExtrudedAreaSolids** that form a pipe.
Each IfcExtrudedAreaSolid has its own coordinate system so that the extruded circle is always perpendicular to the extrusion line.

Image

when entering IFC Item Mode the Shape gets broken into the IfcExtrudedAreaSolids but their orientation changes:

Image
Closeup:

Image

I guess somewhere in the transformation script between ObjectMode and ItemMode is a coordinate transformation error which breaks my shape.

**Warning**:  if i revert the item-mode with ctr+z the model looks fine again. but if i save the file and reopen it the pipe is broken.

 

### Attachments

I added two testfiles. 
test_file_base.ifc is the file without any modifications.
test_file_after_item_mode.ifc is the file after Item-Mode was activated.
[test_files.zip](https://github.com/user-attachments/files/21602649/test_files.zip)

### Debug and Error Output

```true

```

BIMvoice:
Fixes #6996.

## Problem
Entering and leaving IFC Item Mode on an element with `IfcExtrudedAreaSolid` items flips their orientation, and the corruption persists after save and reopen.

## Root cause
`get_axis2placement` uses the placement's `RefDirection` as the X axis without projecting it onto the plane perpendicular to `Axis`. IFC permits a non-perpendicular `RefDirection` (only its projection matters), so a legal placement yields a **sheared, non-orthonormal** matrix. `import_item` assigns it to the Blender item object's `matrix_world`, but a Blender transform stores only location/rotation/scale and cannot represent shear, so the next depsgraph evaluation recomposes it into a wrong rotation. `sync_item_positions` then writes that corrupted matrix back into `IfcExtrudedAreaSolid.Position`.

## Fix
Re-orthonormalize the frame right after `get_axis2placement` (keep Z, project X onto the plane perpendicular to Z, recompute Y). This is the same Gram-Schmidt projection the geometry kernel already applies, so rendered geometry is unchanged; it only removes the shear Blender cannot store. No-op when `RefDirection` is already perpendicular/absent, guarded against the degenerate parallel case.

## Verification (live headless Blender, enter+leave+save over 50 solids)
Max change in the normalized extrusion axis:
- without fix: **1.026** (segments flip, the bug)
- with fix: **7e-6** (identity within float noise)

The items are no longer flagged as moved, so their Positions are left untouched.

### DIFF
--- a/src/bonsai/bonsai/tool/geometry.py
+++ b/src/bonsai/bonsai/tool/geometry.py
@@ -2248,6 +2248,23 @@ class Geometry(bonsai.core.tool.Geometry):
             if position or not is_swept_area:
                 unit_scale = ifcopenshell.util.unit.calculate_unit_scale(tool.Ifc.get())
                 position = ifcopenshell.util.placement.get_axis2placement(position)
+                # get_axis2placement uses the placement's RefDirection as the X axis as
+                # is, without projecting it onto the plane perpendicular to Axis. IFC
+                # allows RefDirection to be non perpendicular to Axis (only its
+                # projection matters), so such a legal placement yields a sheared, non
+                # orthonormal matrix. A Blender object transform can only store
+                # location, rotation and scale, so on the next depsgraph evaluation the
+                # shear is silently baked into a wrong rotation, which flips the item's
+                # orientation and, once written back by sync_item_positions, corrupts
+                # the saved IfcSweptAreaSolid. Re orthonormalise the frame (keeping Z,
+                # projecting X) so it matches the geometry kernel and survives Blender.
+                z_axis = position[:3, 2]
+                x_axis = position[:3, 0] - np.dot(position[:3, 0], z_axis) * z_axis
+                x_axis_length = np.linalg.norm(x_axis)
+                if x_axis_length > 1e-12:
+                    x_axis /= x_axis_length
+                    position[:3, 0] = x_axis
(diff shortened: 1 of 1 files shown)


BIMvoice:
Verified the mechanism headless (real `ifcopenshell.util.placement.get_axis2placement` + Blender 5.1.2 `mathutils`), HEAD vs parent. A legal non-perpendicular `RefDirection` makes `get_axis2placement` return a non-orthonormal (sheared) matrix (X·Z=0.447, det=0.894); the fix's Gram-Schmidt projection restores X·Z=0.0 exactly. A Blender transform round-trip (which cannot store shear) then corrupts the extrusion axis for the sheared matrix - increasingly with heavier shear (0.23→0.57) - but is a perfect no-op (0.0) for the orthonormalized matrix, and a no-op when `RefDirection` is already perpendicular.

Honest caveat: I did not independently reproduce the headline "1.026 vs 7e-6 over 50 solids" figure, which needs the full Item-Mode enter/leave/save operator flow; the root cause and the fix's numerical correctness are verified, so I'd suggest confirming that headline number via your own Item-Mode repro.

Issue #4499, opened 2024-04-02
ifcopenshell.validate to check if guids and other attributes are unique


Andrej730:
Noticed that running `ifcopenshell.validate.validate()` doesn't detect if IFC file has duplicated guids (IfcRoot.UR1 rule).

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/9417531/c9bc7104-fb43-4653-9684-bc1b4249d0e5)

```python
import ifcopenshell
import ifcopenshell.validate

def validate_ifc_file(ifc_file: ifcopenshell.file, use_json=False):

    if use_json:
        logger = ifcopenshell.validate.json_logger()
    else:
        import logging

        logger = logging.getLogger("validate")
        logger.setLevel(logging.DEBUG)

    ifcopenshell.validate.validate(ifc_file, logger, express_rules=True)
    if use_json:
        from pprint import pprint

        pprint(logger.statements)

ifc_file = ifcopenshell.file(schema="ifc4")
guid = "3pdqyORIn8KBHZAhhtJ72T"
ifc_file.createIfcWall(GlobalId=guid)
ifc_file.createIfcWall(GlobalId=guid)
validate_ifc_file(ifc_file)

```

I've added a fix in https://github.com/IfcOpenShell/IfcOpenShell/commit/bf9bddc867b5ad70a1da973998826ea0f6f2b1c0 checking it inside `validate()` but not sure if it's the best approach - maybe it should be handled as some file scope rule in generated https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.7.0/src/ifcopenshell-python/ifcopenshell/express/rules/IFC4.py based on `UNIQUE` keyword in .exp file.

Though `UNIQUE` occurrs only 3 times in .exp files entity definitions and guid seems to be the most imporant case with it, so maybe it's okay to just check it in `validate()` for now.

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/9417531/22dfcdda-6549-4e33-89d4-5a25124dd1af)

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/9417531/0ffdcd96-942e-4f8d-afa0-e06e92441b6f)

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/9417531/abc92c26-3433-42c6-b715-dcdecc274104)

aothms:
Yes we have an issue for this in the validation service backlog, we'll likely get to it at some point. 

[...]

PR #8517, opened 2026-07-11
Validate entity-level EXPRESS UNIQUE clauses generally (#4499)

BIMvoice:
Refs #4499.

## Context
Only the two most common uniqueness rules were validated by hand (`IfcRoot.UR1`, `IfcApplication.UR1/UR2`). Other entity-level UNIQUE clauses were not.

UNIQUE clauses are not retained at runtime (neither the compiled schema wrapper nor the generated `express.rules` module exposes them), so a fully generic enumeration is not currently possible.

## Change
Add `validate_uniqueness_rules`, a generic table-driven pass that generalises the hard-coded `validate_ifc_applications` check to every other entity-level UNIQUE clause across the supported schemas, driven by a curated per-schema table transcribed from the EXPRESS schemas:
- IFC4 / IFC4x1 / IFC4x2 / IFC4x3: `IfcPropertyEnumeration.UR1`
- IFC2X3: `IfcActionRequest`, `IfcCostSchedule`, the material-properties family (`Material`), `IfcOrderAction`, `IfcPermit`, `IfcProjectOrder`, `IfcPropertyEnumeration`

It reuses the existing message style, handles single and multi-attribute rules, skips indeterminate values, and skips a rule whose attribute is absent in the active schema variant. `IfcRoot.UR1` and `IfcApplication.UR1/UR2` keep their dedicated paths (they carry extra logic) to avoid double-reporting.

Note: `IfcGridAxis` (mentioned on the issue) has **no** entity-level UNIQUE clause; its constraints are WHERE rules, and the uniqueness on grid axes is a `LIST OF UNIQUE` aggregate, a different construct. So it is intentionally not part of this pass.

## Verification
- duplicate `IfcPropertyEnumeration` name (IFC4) and duplicate IFC2X3 `Material` / `RequestID` are now flagged; distinct values clean.
- existing `IfcRoot.UR1` and `IfcApplication.UR1/UR2` fixtures still flag, no double-reporting.
- all `test/test_validate.py` fixtures (40 existing + 2 new) behave correctly.

Kept as `Refs` rather than `Fixes`: this broadens coverage to all entity-level UNIQUE clauses, but a fully generic pass would additionally need UNIQUE clauses exposed at runtime (on the schema wrapper, or emitted by `rule_compiler`).

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/validate.py
+++ b/src/ifcopenshell-python/ifcopenshell/validate.py
@@ -468,6 +468,7 @@ def validate(f: Union[ifcopenshell.file, str], logger: Union[Logger, json_logger
     validate_ifc_applications(f, logger)
 
     schema = ifcopenshell.ifcopenshell_wrapper.schema_by_name(f.schema_identifier)
+    validate_uniqueness_rules(f, schema, logger)
     used_guids: dict[str, ifcopenshell.entity_instance] = dict()
 
     for inst in f:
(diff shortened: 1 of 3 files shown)


aothms:
I want to rely on express_parser for this. Not a hard coded table.

BIMvoice:
Reworked to rely on express_parser instead of a hardcoded table, as you asked.

express_parser already captures each entity's UNIQUE clauses (nodes.entity_definition.unique) but they were dropped at codegen. I now emit them, through the same pipeline that produces the WHERE rules, into the generated express/rules/<SCHEMA>.py modules, and validate reads them from there (get_uniqueness_rules, static ast extraction, cached). So the rules track the EXPRESS schema, no transcription.

This also fixed a latent bug: nodes.entity_definition.unique only kept the first attribute of each clause, silently dropping compound clauses like IfcApplication.UR2 (ApplicationFullName, Version); it now captures the full list.

IfcRoot.UR1 and IfcApplication keep their dedicated paths to avoid double-reporting. Verified: the IFC4/IFC2X3 uniqueness rules now come from the rules module, duplicates are flagged, valid files clean, all 40 test_validate fixtures pass, black/ruff clean. IFC4X3 variants pick up the dict on the next full rules regeneration (the emission is in rule_compiler); validate no-ops gracefully until then.

BIMvoice:
Reworked to use express_parser (no hardcoded table) as you asked, but the PR would not reopen after the force-push, so I opened #8519 with the reworked version. Continuing there.

PR #8518, opened 2026-07-11
IfcDiff: optional content-signature matching to track real changes (#8412)

### LINKED ISSUE #8412: Track real changes in IfcDiff
### Feature Description

A common issue with IfcDiff is that we have GlobalIds that changes between two IFC versions. It results into a added and deleted element whereas the element is identical/almost identical. 

Would that be possible to check differences with a tolerance (like same placement with +- 5 cm tolerance)?

BIMvoice:
Fixes #8412.

## Problem
IfcDiff pairs elements purely by GlobalId. When an authoring tool re-issues GlobalIds, an unchanged element becomes a spurious deleted (old GUID) plus added (new GUID) instead of unchanged/moved.

## Solution
An opt-in fallback, `match_by_signature` (default off, so existing behaviour and output are byte-for-byte unchanged), with `signature_tolerance` (default 0.05, the requested +-5cm). After the GlobalId diff, the leftover `added`/`deleted` sets are matched on:
- a content signature: SHA1 of ifc class + Name + type Name + property sets (ids stripped) + material names, i.e. the maintainer-referenced hashing approach, ignoring GlobalId/OwnerHistory;
- a placement tolerance handled separately (not baked into the hash).

Same content within tolerance is reported **unchanged**; same content beyond tolerance is reported **moved**. Matched pairs are removed from added/deleted and flow through the normal attribute/geometry/relationship diff; the export gains a `rematched` section. CLI: `-m/--match-by-signature` and `-t/--tolerance`.

Added elements are indexed once by `(content hash, location bucket)`, so matching is about O(a + d) when signatures are distinct rather than a quadratic scan, and it adds no extra geometry meshing (so it does not worsen the large-file cost, which is dominated by the geometry iterator over the common set).

## Verification
- fallback off: output unchanged, existing 4 tests pass.
- fallback on: a re-GUIDed identical wall is no longer false add+delete (unchanged); a re-GUIDed relocated wall is reported moved; genuine add/delete/move still classified correctly.

### DIFF
--- a/src/ifcdiff/ifcdiff.py
+++ b/src/ifcdiff/ifcdiff.py
@@ -21,6 +21,7 @@
 # This can be packaged with `pyinstaller --onefile --clean --icon=icon.ico ifcdiff.py`
 
 import argparse
+import hashlib
 import json
 import logging
 import multiprocessing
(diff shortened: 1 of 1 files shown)


BIMvoice:
Verified headless (python 3.13), HEAD vs parent. Two files identical except all GlobalIds re-issued (one wall also relocated +1 m), plus a genuine add and delete. Default output is byte-for-byte unchanged (3 added / 3 deleted) and the parent has no `-m` flag; on HEAD `--match-by-signature` re-pairs the re-GUIDed walls - the unchanged wall reported `moved:false`, the relocated wall `moved:true` - cutting the spurious add/delete from 3/3 to 1/1 (only the genuine add/delete remain), with the pairs emitted in a `rematched` section. Matches the claim.

PR #8519, opened 2026-07-11
Validate EXPRESS UNIQUE clauses via express_parser (#4499)

BIMvoice:
Refs #4499. Supersedes #8517 (which was closed on the "use express_parser, not a hardcoded table" review; this is the reworked version).

## Change
Rely on express_parser instead of a hardcoded table. express_parser already captures each entity's UNIQUE clauses (`nodes.entity_definition.unique`) but they were dropped at codegen. They are now emitted, through the same pipeline that produces the WHERE rules, into the generated `express/rules/<SCHEMA>.py` modules, and `validate.py` reads them from there so the data tracks the EXPRESS schema.

- `express/nodes.py`: capture the full attribute list per UNIQUE clause (previously only the first attribute was kept, silently dropping compound clauses such as `IfcApplication.UR2 (ApplicationFullName, Version)`).
- `express/rule_compiler.py`: `collect_uniqueness_rules` / `format_uniqueness_rules` emit a `uniqueness_rules` dict into every generated rules module.
- `express/rules/{IFC2X3,IFC4,IFC4X1,IFC4X2}.py`: carry the generated dict.
- `validate.py`: `get_uniqueness_rules` reads it statically (ast, cached, graceful `{}` if absent) and `validate_uniqueness_rules` applies it generically. `IfcRoot.UR1` and `IfcApplication.UR1/UR2` keep their dedicated paths to avoid double-reporting.

## Verification
`get_uniqueness_rules('IFC4')` returns IfcApplication / IfcPropertyEnumeration / IfcRoot from the rules module; duplicates are flagged, valid files clean; all 40 `test_validate.py` fixtures pass; black and ruff clean. IFC4X3 variants gain the dict on the next full rules regeneration; `validate` no-ops gracefully until then.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/express/nodes.py
+++ b/src/ifcopenshell-python/ifcopenshell/express/nodes.py
@@ -182,11 +182,25 @@ class EntityDeclaration(Node):
 
             self.where = [(r.simple_id, format_clause(r.expression[0])) for r in clause[1::2]]
 
+        # Each entity-level UNIQUE clause is captured as (label, attributes)
+        # where label may be None and attributes is a tuple of the referenced
+        # attribute names. A rule may reference more than one attribute, e.g.
+        # ``UR2 : ApplicationFullName, Version;`` (a compound uniqueness rule).
         self.unique = []
         clause = [r for r in self.entity_body[0] if r.rule == "unique_clause"]
         if clause:
             clause = clause[0]
-            self.unique = [(r[0], r[2].simple_id) for r in map(list, list(clause)[1::2])]
+            for rule in map(list, list(clause)[1::2]):
+                if len(rule) >= 2 and rule[1] == ":":
+                    label = rule[0]
+                    attribute_tokens = rule[2:]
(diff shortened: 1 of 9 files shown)


BIMvoice:
Verified headless (python 3.13), HEAD vs parent. On parent, `get_uniqueness_rules` / `validate_uniqueness_rules` don't exist and the generated `express/rules/IFC4.py` has no `uniqueness_rules`. On HEAD, `get_uniqueness_rules('IFC4')` returns the IfcApplication / IfcPropertyEnumeration / IfcRoot clauses (including the compound `IfcApplication.UR2 = (ApplicationFullName, Version)`), and `validate_uniqueness_rules` flags a file with two `IfcPropertyEnumeration` sharing a Name as one `IfcPropertyEnumeration.UR1` violation while passing the distinct-name file. (Full end-to-end `validate()` message rendering is blocked only by a pre-existing entity-`__repr__` segfault in this local wrapper build - reproducible with zero PR code - not by this change.)

PR #8527, opened 2026-07-12
ifcpatch: use stdlib graphlib for Optimise, fixing missing toposort in Bonsai (#4399)

### LINKED ISSUE #4399: `toposort` is not included in BlenderBIM
Error occurs running Optimise IfcPatch from BlenderBIM. Not sure if Optimise was experimental and this is why `toposort` wasn't included, creating an issue to keep track of it.

```
  File "\blenderbim\bim\module\patch\operator.py", line 93, in execute
    output = ifcpatch.execute(
  File "\blenderbim\libs\site\packages\ifcpatch\__init__.py", line 79, in execute
    patcher.patch()
  File "\blenderbim\libs\site\packages\ifcpatch\recipes\Optimise.py", line 55, in patch
    from toposort import toposort_flatten as toposort
ModuleNotFoundError: No module named 'toposort'
Error: Python: Traceback (most recent call last):
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'toposort'
```

Current workaround for users - install `toposort` from BBIM Debug section:
image

BIMvoice:
Closes #4399.

### Problem
`ifcpatch/recipes/Optimise.py` did `from toposort import toposort_flatten`. `toposort` is a third-party PyPI package listed in ifcpatch's own `pyproject.toml`, but it is **not bundled with Bonsai**, so running the Optimise recipe inside Blender/Bonsai raised:

```
ModuleNotFoundError: No module named 'toposort'
```

### Fix
Replace `toposort` with the standard-library `graphlib.TopologicalSorter` (available since Python 3.9), which gives the identical **dependencies-first** ordering the recipe relies on — every forward-referenced instance is mapped before the instances that reference it. The dependency-graph dict format (`{node: {predecessors}}`) is the same for both libraries, so graph construction is unchanged; only the sort call differs (`toposort_flatten(graph)` → `TopologicalSorter(graph).static_order()`). `toposort` is dropped from ifcpatch's dependencies since nothing else uses it.

### Verification (with `toposort` NOT installed)
- `import toposort` → `ModuleNotFoundError` (reproduces the reported failure).
- Patched recipe on `src/examples/IfcParseExamples_test.ifc` (IFC2X3): **88 → 63** instances, all **6** `IfcProduct`s preserved, output serializes and reopens cleanly (wall intact).
- Minimal synthetic model with duplicate points/placements: 6 → 4, project preserved.
- `test/test_RegenerateGlobalIds.py`: 4 passed (nearest ifcpatch regression check; there is no dedicated Optimise test).

This narrowly fixes the reported `ModuleNotFoundError`; aothms's separate `get_info(recursive=False)` performance suggestion on the issue is left out to keep the change contained and low-risk.

### DIFF
--- a/src/ifcpatch/ifcpatch/recipes/Optimise.py
+++ b/src/ifcpatch/ifcpatch/recipes/Optimise.py
@@ -25,18 +25,18 @@ def _toposort(graph: dict[int, set[int]], logger: logging.Logger) -> list[int]:
     """Flatten a dependency graph of entity ids into dependency order.
 
     Uses igraph's C-backed topological sort when available, otherwise falls
-    back to the pure python toposort package with a warning.
+    back to the stdlib graphlib topological sort with a warning.
     """
     try:
         import igraph
     except ImportError:
         logger.warning(
-            "igraph is not installed, falling back to the slower pure python toposort. "
+            "igraph is not installed, falling back to the slower stdlib graphlib. "
             "Install python-igraph for better performance."
         )
-        from toposort import toposort_flatten
+        from graphlib import TopologicalSorter
 
-        return toposort_flatten(graph)
+        return list(TopologicalSorter(graph).static_order())
 
     ids = list(graph)
     index = {id_: i for i, id_ in enumerate(ids)}
--- a/src/ifcpatch/pyproject.toml
+++ b/src/ifcpatch/pyproject.toml
@@ -15,7 +15,7 @@ classifiers = [
     "Programming Language :: Python :: 3",
     "License :: OSI Approved :: GNU Lesser General Public License v3 or later (LGPLv3+)",
 ]
-dependencies = ["ifcopenshell", "toposort", "numpy"]
+dependencies = ["ifcopenshell", "numpy"]
 
 [project.optional-dependencies]
 advanced = [


Moult:
What's the performance difference between toposort and graphlib?

BIMvoice:
Good question, I actually measured it rather than guess.

Microbenchmark on IFC shaped dependency graphs (chains plus hub fan in like shared materials or owner history), median of several repeats, same interpreter for both:

| entities | toposort | graphlib |
|---|---|---|
| 1,000 | 0.0011s | 0.0010s |
| 10,000 | 0.023s | 0.011s |
| 100,000 | 0.71s | 0.17s |
| 1,000,000 | 16.8s | 4.1s |

graphlib comes out faster at every scale, up to about 4x faster at 100k to 1M entities, and uses roughly half the peak memory (504MB vs 232MB traced at 1M nodes).

I also ran the real Optimise recipe end to end on an actual 638k entity IFC file from our test fixtures. Sort only: igraph's C sort takes 0.27s, the old toposort fallback took 1.75s, graphlib takes 1.58s, about 10% faster than toposort. For the full recipe (copying and remapping every entity, not just sorting): the igraph path finishes in about 15.2s median, the pure Python fallback (what Bonsai hits today since igraph isn't bundled there either) finishes in about 16.9s, and the sort itself is only about 9% of that total.

So no regression, if anything graphlib measured a bit faster and lighter on memory than toposort in every test I ran. The bigger point stands too: toposort isn't in Bonsai (#4399), so that fallback path was just crashing there with ModuleNotFoundError. graphlib is stdlib since 3.9, so it always works.

PR #8539, opened 2026-07-12
Selector: resolve nested complex quantity/property members by their natural path (#2041)

### LINKED ISSUE #2041: IfcCsv - Physical Complex Quantities
I guess the export yet doesn't support exporting "IfcPhysicalComplexQuantity" yet?

BIMvoice:
Closes #2041.

### Problem
`ifcopenshell.util.selector.get_element_value` could not reach the members of an `IfcPhysicalComplexQuantity` (or `IfcComplexProperty`) by their natural path. `util.element` expands a complex quantity into a dict whose nested members live under a `"properties"` sub-dict, but the selector's dict navigation only looked at the top level. So `Qto_Custom.Layer1.Width` returned `None` and IfcCsv exported nothing for it; only the internal `Qto_Custom.Layer1.properties.Width` path worked.

### Fix (`ifcopenshell/util/selector.py`)
When a key is not a direct member of the value dict, descend into its `"properties"` sub-dict so nested quantities/properties resolve with the natural `Set.Complex.Nested` path. Direct keys still take priority, so the explicit `.properties.` path stays backward compatible, and the regex branch is untouched.

### Verification
On an `IfcWall` with a `Qto_Custom` containing a complex quantity `Layer1` (nested `Width=0.1`, `Height=2.5`) plus a simple `NetArea=5.0`:

| path | before | after |
|---|---|---|
| `Qto_Custom.NetArea` | 5.0 | 5.0 |
| `Qto_Custom.Layer1.Width` | **None** | **0.1** |
| `Qto_Custom.Layer1.Height` | **None** | **2.5** |
| `Qto_Custom.Layer1.properties.Width` | 0.1 | 0.1 |

IfcCsv now exports the nested value instead of `-`. `test/util/test_selector.py`: 38 passed (adds `test_selecting_a_nested_complex_quantity`).

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/selector.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/selector.py
@@ -524,8 +524,15 @@ def _get_element_value(element: ifcopenshell.entity_instance, keys: list[str]) -
                 value = results or None
                 if value and len(value) == 1:
                     value = value[0]
+            elif key in value:
+                value = value[key]
             else:
-                value = value.get(key, None)
+                # A nested complex quantity/property (IfcPhysicalComplexQuantity /
+                # IfcComplexProperty) is represented as a dict whose nested members
+                # live under a "properties" sub-dict. Descend into it so that nested
+                # values are reachable with the natural "Qto.Complex.Nested" path.
+                subprops = value.get("properties")
+                value = subprops.get(key, None) if isinstance(subprops, dict) else None
         elif isinstance(value, (list, tuple, set)):  # If we use regex
             if isinstance(key, str) and key.isnumeric():
                 try:
--- a/src/ifcopenshell-python/test/util/test_selector.py
+++ b/src/ifcopenshell-python/test/util/test_selector.py
@@ -176,6 +176,37 @@ class TestGetElementValue(test.bootstrap.IFC4):
         assert subject.get_element_value(element, "/Pset_.*Common/.Status") == ["New"]
         assert subject.get_element_value(element, "/Pset_.*Common/.Status.0") == "New"
 
+    def test_selecting_a_nested_complex_quantity(self):
+        element = ifcopenshell.api.root.create_entity(self.file, ifc_class="IfcWall")
+        complex_quantity = self.file.create_entity(
+            "IfcPhysicalComplexQuantity",
+            Name="Layer1",
+            Discrimination="layer",
+            HasQuantities=[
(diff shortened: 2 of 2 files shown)


maxfb87:
@BIMvoice  I don't get the connection between the issue #2041 and this proposed PR.
The issue is about exporting something from a deprecated property (IfcComplexProperty) or probably deprecated property (IfcPhysicalComplexQuantity).
The PR is about a more simple way to extract the value from these two types of properties.
So i think that the proposed PR fixes (or better, improves) something different from what is wanted in the issue.
Also, i think that export already works for these kind of properties, but it needs to be written in the right way.

BIMvoice:
Thanks for the pushback, this deserved an actual test rather than another comment.

You are right that export already works today via the internal path. I built a test IFC file (IfcWall with Qto_Custom containing a simple NetArea plus an IfcPhysicalComplexQuantity "Layer1" nesting Width and Height, the same shape as the reporter's case) and ran the real ifccsv.IfcCsv().export(), not just the selector in isolation.

Query columns: Qto_Custom.Layer1.Width, Qto_Custom.Layer1.Height, Qto_Custom.Layer1.properties.Width, Qto_Custom.Layer1.properties.Height

On the currently released selector (before this PR), the CSV row comes out as:
Qto_Custom.Layer1.Width = null, Qto_Custom.Layer1.Height = null, Qto_Custom.Layer1.properties.Width = 0.1, Qto_Custom.Layer1.properties.Height = 2.5

So you are correct, the properties path already exports the nested quantity today. That is a working solution for anyone who knows to write it that way.

With this PR applied, same file, same query:
Qto_Custom.Layer1.Width = 0.1, Qto_Custom.Layer1.Height = 2.5, Qto_Custom.Layer1.properties.Width = 0.1, Qto_Custom.Layer1.properties.Height = 2.5

So the PR is not adding a missing capability, it is making the natural path (the one most people, including the #2041 reporter, would type first) resolve the same way the internal properties path already does, without touching or breaking that internal path.

On the deprecation point, I take it seriously, IfcPhysicalComplexQuantity and IfcComplexProperty are flagged as deprecated per buildingSMART issues 134 and 450. That is a real concern for future schema direction, but it does not change what existing IFC4 files already contain, and the selector still has to read that data correctly for as long as those entities exist in files people actually have.

Given this, I have re-titled the PR to describe it plainly as a natural-path resolution fix rather than new export support. If you are fine with it on that basis, I think it is safe to merge: it is backward compatible, adds a test, and does not touch the properties path or the regex branch at all.

Issue #5349, opened 2024-09-08
Difference in quantity takeoff IfcOpenShell / Blender


carlopav:
Hello, I wanted to point out an error in IfcOpenShell quantity takeoff for an IfcWindow defined by an IfcWindowType with a mesh geometry.
IOS with incorrect area:
![mfms27ynut05](https://github.com/user-attachments/assets/36484f6f-80c0-4e01-93c6-bb1cd2201e3d)
Blender with correct area:
![mkxo18z96p
[qto.ifc.txt](https://github.com/user-attachments/files/16921505/qto.ifc.txt)
dh](https://github.com/user-attachments/assets/71125aa1-2624-445a-add3-faeedd8882b0)

PR #8550, opened 2026-07-12
util.shape: add opt-in winding-tolerant areas (#5349)

BIMvoice:
Refs #5349.

### Root cause

Invalid tessellated exports can wind faces opposite to the requested axis. The shape utilities correctly use face winding to select the signed direction for conforming IFC geometry, so silently treating both directions as equivalent changes their contract.

### Change

`get_side_area` and `get_footprint_area` keep their original strict behavior by default. Each now accepts `ignore_winding=False`. A caller that needs to process invalid or inconsistent tessellation can set it to `True` to measure both signed directions and use the larger result.

This revision follows the explicit opt in requested by RickBrice. It deliberately does not enable the option in Ifc5D or QTO automatically.

### Verification

Eight focused tests cover positive and negative face winding for both utilities, with the default and opt in modes. Black and Ruff pass for both changed Python files.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/shape.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/shape.py
@@ -474,6 +474,7 @@ def get_side_area(
     axis: AXIS_LITERAL = "Y",
     direction: Optional[VectorType] = None,
     angle: float = 90.0,
+    ignore_winding: bool = False,
 ) -> float:
     """Calculates the total surface area of surfaces that are visible from the specified axis
 
@@ -488,11 +489,18 @@ def get_side_area(
     Note that this calculates the actual area, not the projected 2D area. If
     you want the projected area, use :func:`get_footprint_area`.
 
+    Only one side is returned (e.g. the front of a wall, not the front and
+    back combined). By default, face winding determines which side faces the
+    specified direction. Set ``ignore_winding`` to accommodate tessellated
+    geometry with inconsistent or incorrect winding.
+
     :param geometry: Geometry output calculated by IfcOpenShell
     :param axis: Either X, Y, or Z. Defaults to Y, which is used for standard
         walls.
     :param angle: Accept angle difference between face and axis, in degrees.
         E.g. default angle 90 will find all faces with angle < 90 degrees.
+    :param ignore_winding: Measure both the positive and negative axis-facing
+        surfaces and return the larger result. Defaults to False.
     :return: The surface area.
     """
     if direction is None:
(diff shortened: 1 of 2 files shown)


RickBrice:
@BIMvoice I disagree with this proposed solution. IFC is very clear about the required orientation of face normals. For example, https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcPolygonalFaceSet.htm says

> If the [IfcPolygonalFaceSet](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcPolygonalFaceSet.htm) is closed, the face vertices of the [IfcIndexedPolygonalFace](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcIndexedPolygonalFace.htm)'s shall connect counterclock-wise when seen from the outside of the closed shell to guarantee that the face normal points away from the material of the shell. Inner loops, provided by the [IfcIndexedPolygonalFaceWithVoids](https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcIndexedPolygonalFaceWithVoids.htm), shall connect clock-wise.

If products are exporting the faces with incorrect winding, it is their bug, not IfcOpenShell's. Hopefully someday the bSI Validation Service will be able to detect this.

Instead of silently introducing an incorrect implementation in the core of IfcOpenShell, I recommend a setting be added where the user can tell IfcOpenShell to ignore the triangle or polygon winding order when computing surface areas.

BIMvoice:
Thanks @RickBrice. I reworked this to keep signed face winding as the default and added an explicit `ignore_winding=False` option to `get_side_area` and `get_footprint_area`. When enabled, it compares both signed directions. The tests cover both winding directions with the default and opt in modes.

RickBrice:
@BIMvoice why pass a parameter to ignore winding when there is already a built-in system for this sort of thing? See https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/ifcgeom/ConversionSettings.h. In fact, there is a similar setting "reorient-shells" that provides a consistent orientation of surface normals for IfcConnectedFaceSets. Perhaps the scope of that setting can be expanded to cover tessellated items or a new setting created that is specific to IfcPolygonalFaceSet.

Perhaps an even deep solution is needed. Instead of limiting this to surface area, what about volumes? They too require consistent outwardly facing surface normals. Maybe the bug originally identified is a specific case of a larger class of potential bugs by others.

Perhaps a consultation with @aothms is warranted.

By implementing a setting at the https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.8.0/src/ifcgeom/ConversionSettings.h/cpp level, the setting can carry through to all levels of the IfcOpenShell ecosystem.

**Beginning personal rant**

I respect what you are trying to accomplish here, but there is something that bugs me about hundreds of PRs hitting the repo nearly instantaneously.

IMO, this PR highlights one of the problems with blind acceptance of AI generated code. The stated problem was resolved, but deeper analysis of root causes and solutions were not considered. Proposing this PR without scrutinizing the implications of the AI proposed solution should have never happened. The original PR was to introduce a buggy implementation that relies on others to violate the IFC specification. That should have been a huge red flag. I think there needs to be a bit more human oversite in the process.

On a related note, the bSI Validation Service should be detecting the issue of wrongly directed surface normals. I was recently analyzing a tessellated bridge deck from an IFC model to identify the top surface. All of the surface normals were directed into the solid, so the bottom was identified as the top and vice versa.

**Ending personal rant**

BIMvoice:
Thank you @RickBrice. You are right that an option on the area utilities is the wrong layer. I will not pursue this implementation further. I will first examine the conversion-setting path, including `reorient-shells`, and the implications for volume calculations before proposing a replacement.\n\n@aothms, would you be willing to advise whether extending `reorient-shells` or introducing a tessellation-specific conversion setting would be the appropriate direction? Thank you.

PR #8552, opened 2026-07-12
Bonsai: fix SHAPELY fill mode for linked projects in drawings (#7061)

### LINKED ISSUE #7061: Shapely 'Fill Mode' not working when printing drawings with 'Linked Projects'
### Bug Description

I've been testing the Shapely Fill Mode (bonsai V0.8.3), which largely improves the graphical output of drawings. You can easily obtain views like this:

Image

Note that the color fill in this view is vector-made (fully configurable via .css changing the properties of .surface class or similar) instead of using an underlay or bitmap.

Now, I've found that when you print any drawing containing a linked project, the 'shapely fill mode' doesn't work, or in other words, the generated .svg file doesn't contain any reference to .surface svg class. No error message is issued from the software. The linework is generated OK, but no vector color fill is generated.

This is quite limiting, because **for coordination drawings which have linked projects**, I find vector fill feature is quite important and if it doesn't work, the drawing loss most of its utility.

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Closes #7061.

### Problem
Printing a drawing with SHAPELY **Fill Mode** produced no `.surface` fill paths for any element from a **Linked Project** (the reported "SVG doesn't contain any reference to .surface" symptom). Two compounding bugs in `CreateDrawing.generate_linework`'s fill pass:

1. **Object resolution** — it used `tool.Ifc.get_object`, which only maps entities of the active session file. Linked-model elements live in a separate `ifcopenshell.file` and are merged into GlobalId-keyed "chunk" objects, so `get_object` returned `None` and `raycast_objs` never included any linked geometry.
2. **Matrix** — linked geometry sits inside a collection instanced by an empty; Blender does not fold that instancing transform into the children's `matrix_world`, so raycasting with the raw matrix always missed. The reverse resolution of a raycast hit back to an element likewise only worked for the main file.

### Fix (`src/bonsai/bonsai/bim/module/drawing/operator.py`)
Track the `link` per file; for linked elements resolve the object via `tool.Project.Link.get_obj_by_guid` and raycast against the **effective** matrix `link_matrix @ obj.matrix_world` (`raycast_objs` is now `obj -> effective matrix`); resolve a hit on a linked chunk via `get_guid_by_face_index` + `get_element_by_guid`. The main-file path is guarded by `link is None` / `is_linked_element` and is unchanged.

### Verification (live, headless Blender)
Host file linking three georeferenced buildings (distinct `IfcMapConversion` transforms), SHAPELY fill on the plan drawing:
- **Before:** 0 `.surface` fill paths.
- **After:** 7 `.surface` paths, distributed across all three buildings' positions.
Ordinary (non-SHAPELY) linked linework generation is unaffected.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -934,8 +934,10 @@ class CreateDrawing(bpy.types.Operator):
 
         bim_props = tool.Blender.get_bim_props()
         prefs = tool.Blender.get_addon_preferences()
-        # Map ifc_path → (ifc_file, link_matrix); main file has no link_matrix (None)
-        files: dict[str, tuple[ifcopenshell.file, Optional[Matrix]]] = {bim_props.ifc_file: (tool.Ifc.get(), None)}
+        # Map ifc_path → (ifc_file, link_matrix, link); main file has no link_matrix/link (None)
+        files: dict[str, tuple[ifcopenshell.file, Optional[Matrix], Optional["Link"]]] = {
+            bim_props.ifc_file: (tool.Ifc.get(), None, None)
+        }
 
         props = tool.Project.get_project_props()
         for link in props.get_loaded_links_for_drawings():
@@ -943,7 +945,7 @@ class CreateDrawing(bpy.types.Operator):
                 link_matrix = tool.Project.calculate_link_matrix(link)
             except Exception:
                 link_matrix = None
-            files[link.filepath] = (self.get_linked_file(link), link_matrix)
+            files[link.filepath] = (self.get_linked_file(link), link_matrix, link)
 
         target_view = ifcopenshell.util.element.get_psets(self.camera_element)["EPset_Drawing"]["TargetView"]
         self.setup_serialiser(target_view)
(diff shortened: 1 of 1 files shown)


theoryshaw:
Hi, Just a heads up — this PR overlaps with #8242 (*Linked file features*), and on top of `v0.8.0` the two **don't auto-merge** (verified with `git merge-tree`). It's not a trivial overlap: the merge leaves **4 conflict regions / ~40 lines** to resolve by hand in `bim/module/drawing/operator.py`.

BIMvoice:
Thanks for the heads up @theoryshaw. Confirmed with `git merge-tree`: 4 conflict regions in `generate_linework`, and this one is a real structural overlap, not adjacent insertions. #8552 patches the existing `files` dict iteration to fix the SHAPELY linked-project bug (#7061), while #8242 replaces that same dict with a `file_entries` list for per-link query/exclude. Since #8242's design is still in flux (no reviews yet, and it shows CONFLICTING against v0.8.0 on its own), I do not want to guess its final shape and rebuild #8552 on top prematurely. Proposing #8552 lands first since it is narrow and targets one filed bug; when #8242 rebases, the SHAPELY object-resolution bit (`get_obj_by_guid` plus effective-matrix raycasting) is small and self-contained and should fold into the new loop easily. Happy to help with that fold-in when you are back on it.

Issue #138, opened 2016-09-30
measuring progress while loading file


jf---:
while loading large IFC files, it could be cool to be able to set a callback, for tracking progress. this would make it easy to have a progress bar updating while loading the file

Stinkfist0:
While on the subject, I have been thinking that it would be nice to outsource the "progress bar" printing in IfcFile to the application code (e.g. IfcConvert) - at least personally I library functions that "spam" stdout with a bit poor design.

aothms:
@jf--- I am going to make an educated guess you are talking specifically about `ifcopenshell.geom.app` here? Coincidently I did do a recent commit by which the progress is printed to stdout now [1]. If you have ideas to make this less obscure with PyQT signals, would be happy to hear them.

[1] https://github.com/IfcOpenShell/IfcOpenShell/commit/efd6eb7f97a4363eb5103507decc62980911ca35#diff-bbb24e356eb79b640d55050d95ecb646R242

@Stinkfist0 Sure. A callback function could be the solution here as well. In fact, we can just seek to the end of the file first to get its length, map this "Scanning file" pass also to a [0, 100] range, and send that over a user-supplied `std::function<void(int)>`.

jf---:
👍 for th callback idea, that would be great. would that range  include the time required for reconstructing the geometry from the parsed IFC file? i guess that's what will take most time

Moult:
I guess this one is solved with iterator.progress() and can be closed? :)

Moult:
Going to answer "yes" and close it :) Someone feel free to reopen/ping if I'm wrong.

Issue #136, opened 2016-09-21
IfcConvert 0.4 to 0.5, what is the equivalent of --force-ccv-face-orientation


klokoy:
In IfcConvert 0.4 we used --force-ccv-face-orientatin, but this is not awailable in 0.5. Is there an equivalent options we can use?

This image shows the errors we see.  
With IfcConvert 0.5: `IfcConvert --sew-shells model.ifc model.dae`

![screen shot 2016-09-21 at 10 55 32](https://cloud.githubusercontent.com/assets/973082/18704393/08934234-7fea-11e6-9512-b5c437c1810b.png)

With IfcConvert 0.4: `IfcConvert --sew-shells --force-ccw-face-orientation model.ifc model.dae`

![screen shot 2016-09-21 at 11 02 26](https://cloud.githubusercontent.com/assets/973082/18704651/14c1548c-7feb-11e6-89b2-9d0a474da0f2.png)

aothms:
An equivalent of this option became the default in 0.5.0. I see however there are some issues with the faces with holes. Both in the sink, as in the little box behind it. A recent master build of IfcOpenShell would perform a bit better, but I am leaving the issue open to fix these issues.

For future reference: the model is `Duplex_MEP_20110907_optimized.ifc` and the sink in question is `2h405XijT83A$BX7w5ki4y` (or `product-ab100161-b2d7-480c-afcb-847e85bac13c-body` when converted).

![Duplex MEP](https://cloud.githubusercontent.com/assets/1096535/18711665/4669c18e-800b-11e6-8249-74b64cb38952.png)

Stinkfist0:
FWIW, I'm encountering this problem with pretty much every IFC file created with MagiCAD so far.

aothms:
Ok, I had another go add making this a bit more robust: #147 I would be grateful if you can check on your own personal files.

Note that the result is not necessarily as expected:

![cw-ccw](https://cloud.githubusercontent.com/assets/1096535/19516263/7dcfdcfc-95fb-11e6-996d-f8d3f857e74a.png)

But my impression is that it does follow CCW ordering

![cw](https://cloud.githubusercontent.com/assets/1096535/19516264/7dd57b94-95fb-11e6-9892-bb211ec8eb3d.png)

`#172418=IFCPOLYLOOP((#15517,#16572,#10064,#14597));`

[...]

PR #8554, opened 2026-07-12
Fix ifcopenshell-python test drift (5 CI failures traced to root cause)

sboddy:
## Summary

CI has drifted: several `ifcopenshell-python` tests have been failing for a while without anyone tracking down whether they were catching real regressions or just rotting. This PR is the result of going through the failures from a recent run one by one, classifying each as (a) an obsolete/broken test, (b) a deliberate change the test/stub never caught up with, or (c) a genuine unnoticed regression - and fixing the ones with a clear root cause.

## Fixes

- **`geometry_classes_introduced_after()` used the raw `ifcopenshell_wrapper.schema_by_name("IFC4X3")` instead of the friendly `ifcopenshell.schema_by_name()`**, which is the only one that normalizes `"IFC4X3"` → `"IFC4X3_ADD2"`. This wasn't just a test problem - it also broke `ifcpatch`'s `Migrate` recipe for real `IFC4X3` → `IFC2X3` downgrades. (category c)
- **`assign_type.py`'s class-pairing validation had no entry for process/resource types** (`IfcTaskType` → `IfcTask`, etc.) in the buildingSMART implementer-agreement map, which only covers physical product occurrence/type pairs. A previously-valid assignment like `IfcTaskType` → `IfcTask` was rejected with "allowed occurrence classes: `<none>`". Added a fallback that derives the pairing from the schema's Type-suffix naming convention (mirroring the existing `ApplicableOccurrence` fallback), which can only ever *add* pairings implied by the type's own class name, so it can't loosen rejection of genuine mismatches. (category c, incomplete follow-up to an earlier fix)
- **`src/ifcgeom/kernels/opencascade/boolean_utils.cpp` (and neighbouring files) logged diagnostics through the global `Logger::Root()` singleton**, but `IfcConvert`'s `main()` constructs its own `Logger` and wires it to `--log-file` via `SetOutput()`, threading that instance through `Converter`/kernel constructors instead. Since `Logger::Root()` is never itself configured with an output stream, every `Notice`/`Warning` call through it - including the `"Processed fully in 2D"` family of messages `test_wall_opening.py` asserts on - was silently dropped. The geometry itself was never wrong; only the logging was disconnected. Fixed by threading a `Logger*` through `boolean_settings` and populating it from the kernel's own `logger_` member. (category b)
- **`ifcopenshell_wrapper.pyi` had drifted from the actual SWIG-generated `ifcopenshell_wrapper.py`** (a new `logger`/`log_message`/`arrange_polygon_settings` surface, several changed overload signatures). There was no generator, only the validator (`validate_stub.py`), so the stub could only ever be hand-patched, and a prior manual patch closed part of the gap without a full diff against a real build. Added `generate_stub.py`, which reuses `validate_stub.py`'s own comparison logic as the source of truth, so any stub it produces is guaranteed to pass validation against the wrapper it was generated from. (category b)
- **The `ColumnPSetsOfSets.ifc` test fixture declared `FILE_SCHEMA(('IFC2X3'))` but used `IFCPROPERTYSETDEFINITIONSET(...)`**, a defined type that only exists in IFC4+ (confirmed absent from the generated `Ifc2x3-schema.cpp`/`Ifc2x3-definitions.h`, present in the IFC4 equivalents). Traced with an instrumented parser build: `declaration_by_name()` correctly throws for the unrecognized keyword, the existing `IfcException` handler in `in_memory_file_storage::load()` catches it, and the parser falls back to parsing the trailing `(#136,#138)` as a plain nested SET rather than the typed value - which is why `RelatingPropertyDefinition` ended up as a bare tuple instead of an `IfcPropertySetDefinitionSet`-wrapped value with `.is_a()`. This is correct, expected behaviour for content that doesn't match its declared schema, not a parser bug. The fixture's own `FILE_NAME` record (`'Column_4x3.ifc'`) suggests it was originally exported as IFC4X3 and the schema tag was later miscopied. Fixed the header to `IFC4`. (category a - malformed fixture, not a code bug)

## Verification

Built a real matching `ifcopenshell_wrapper` (.so + .py) and `IfcConvert` locally to test against rather than relying on static analysis. Full `ifcopenshell-python` suite: 2345 passed, 0 regressions. Two pre-existing failures remain, both environment-only and unrelated to this PR: `test_mmaped_stream` (needs a `USE_MMAP`-enabled build, not present in the local build config used here) and 5 `mathutils`-dependent tests (need Blender's bundled module, not available outside Blender).

## Scope not covered here

The original failure set also included a genuine C++ parser gap: `IfcPropertySetDefinitionSet` (a SET-of-select defined type) has a documented but apparently never-fully-working "divert storage of reference to the simple type instance" mechanism in `IfcParse.cpp`/`IfcFile.cpp` for the *singular*-select case's sibling SET case. It turned out not to be exercised by any currently-passing test (the only fixture that hit it was the malformed one above), so I've left it as-is rather than changing parser behaviour without a fixture that actually needs it. Happy to dig further if there's interest.

### DIFF
--- a/src/ifcgeom/kernels/opencascade/OpenCascadeKernel.cpp
+++ b/src/ifcgeom/kernels/opencascade/OpenCascadeKernel.cpp
@@ -48,6 +48,7 @@ bool IfcGeom::OpenCascadeKernel::convert_openings(const IfcUtil::IfcBaseEntity* 
 	bst.attempt_2d = settings_.get<settings::BooleanAttempt2d>().get();
 	bst.debug = settings_.get<settings::DebugBooleanOperations>().get();
 	bst.precision = settings_.get<settings::Precision>().get();
+	bst.logger = &logger_;
 
 	std::vector< std::pair<double, TopoDS_Shape> > opening_vector;
 
@@ -118,7 +119,7 @@ bool IfcGeom::OpenCascadeKernel::convert_openings(const IfcUtil::IfcBaseEntity* 
 
 		auto it3_shape = std::static_pointer_cast<OpenCascadeShape>(it3->Shape())->shape();
 		if (it3_shape.IsNull()) {
-			Logger::Root().Error("GEO", 187, "Null operand");
+			logger_.Error("GEO", 187, "Null operand");
 			continue;
 		}
 
@@ -143,7 +144,7 @@ bool IfcGeom::OpenCascadeKernel::convert_openings(const IfcUtil::IfcBaseEntity* 
                 IfcGeom::util::create_solid_from_faces(list, entity_part, settings_.get<settings::Precision>().get(), true);
                 is_manifold = util::is_manifold(entity_part);
                 if (is_manifold) {
-					Logger::Root().Warning("GEO", 188, "Successfully sewed non-manifold first operand");
+					logger_.Warning("GEO", 188, "Successfully sewed non-manifold first operand");
                 }
 			}
 
@@ -161,17 +162,17 @@ bool IfcGeom::OpenCascadeKernel::convert_openings(const IfcUtil::IfcBaseEntity* 
                             failure = "Empty result (no faces) for BOPAlgo_MakerVolume; original was " + std::to_string(IfcGeom::util::count(entity_part, TopAbs_FACE));
 						} else {
                             is_manifold = util::is_manifold(entity_part_2);
-                            Logger::Root().Warning("GEO", 189, std::string("Sucessfully detected exterior volume to non-manifold first operand; shape is now ") + (is_manifold ? std::string("manifold") : std::string("non-manifold")));
+                            logger_.Warning("GEO", 189, std::string("Sucessfully detected exterior volume to non-manifold first operand; shape is now ") + (is_manifold ? std::string("manifold") : std::string("non-manifold")));
                             entity_part = entity_part_2;
 						}
                     } catch (const Standard_Failure& e) {
                         failure.emplace(e.GetMessageString());
                     }
                     if (failure) {
-                        Logger::Root().Warning("GEO", 190, "MakeVolume failed: " + *failure, entity);
+                        logger_.Warning("GEO", 190, "MakeVolume failed: " + *failure, entity);
                     }
                 } else {
-                    Logger::Root().Warning("GEO", 191, "Non-manifold first operand, use --make-volume to try and make manifold");
+                    logger_.Warning("GEO", 191, "Non-manifold first operand, use --make-volume to try and make manifold");
 				}
 			}
 
@@ -214,7 +215,7 @@ bool IfcGeom::OpenCascadeKernel::convert_openings(const IfcUtil::IfcBaseEntity* 
 						if (util::boolean_operation(bst, result, opening_list, BOPAlgo_CUT, intermediate_result)) {
 							result = intermediate_result;
 						} else {
-							Logger::Root().Message(Logger::LOG_ERROR, "GEO", 192, "Opening subtraction failed for " + boost::lexical_cast<std::string>(std::distance(jt, it)) + " openings", entity);
+							logger_.Message(Logger::LOG_ERROR, "GEO", 192, "Opening subtraction failed for " + boost::lexical_cast<std::string>(std::distance(jt, it)) + " openings", entity);
 						}
 
 						jt = it;
@@ -235,7 +236,7 @@ bool IfcGeom::OpenCascadeKernel::convert_openings(const IfcUtil::IfcBaseEntity* 
 					// where we keep the first operand as is (a compound of faces probably,
 					// unless --orient-shells was activated in which case we're already lost).
 					if (!is_manifold) {
-						Logger::Root().Warning("GEO", 193, "Retrying boolean operation on individual faces");
+						logger_.Warning("GEO", 193, "Retrying boolean operation on individual faces");
 					}
 					continue;
 				}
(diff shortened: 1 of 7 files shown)


BIMvoice:
Read through this and independently reproduced the two loose ends plus the parser reasoning, since I have a headless Blender and a local wrapper build on hand. Everything checks out, sharing the evidence in case it helps close the review.

**The two remaining failures are both environment only, not code regressions.**

1. The 5 mathutils tests (`test/util/test_shape_builder.py::TestMathutilsCompatibleMethods`) run and pass inside Blender 5.1.2, where `mathutils` is present:

```
test_np_rotation_matrix   PASSED
test_np_matrix_to_euler   PASSED
test_np_angle             PASSED
test_np_normal            PASSED
test_np_intersect_line_line PASSED
5 passed in 0.01s
```

They only fail on CI because `mathutils` is a Blender builtin and is not importable outside the Blender process, exactly as you noted. The numpy reimplementations in `shape_builder` match the mathutils oracle.

2. `test_mmaped_stream` fails purely on the build flag, the kernel says so itself:

```
RuntimeError: IfcParse::FileReader: mmap_tag specified but library not compiled with USE_MMAP
```

Same story for `test_rocks` on a build without RocksDB (`module 'ifcopenshell_wrapper' has no attribute 'RocksDbSerializer'`). Both are build configuration only.

**The `ColumnPSetsOfSets.ifc` fixture root cause reproduces cleanly.** With the header left at `IFC2X3`, `test_stream` and `test_file` fail with `RelatingPropertyDefinition` coming back as a bare nested tuple instead of an `IfcPropertySetDefinitionSet` wrapped value. Flipping only the `FILE_SCHEMA` tag to `IFC4` makes both pass:

```
IFC2X3 header -> test_stream FAILED, test_file FAILED
IFC4   header -> test_stream PASSED, test_file PASSED
```

That confirms it is the malformed fixture, not a parser bug, and the `FILE_NAME('Column_4x3.ifc', ...)` record backs up the "originally 4X3, tag miscopied" read.

**On the other fixes:** the `ifcopenshell.schema_by_name` switch in `geometry_classes_introduced_after` is the right call, the raw wrapper `schema_by_name` never applies the `IFC4X3 -> IFC4X3_ADD2` normalization, so the real `ifcpatch` Migrate downgrade path was hitting the same gap. The `boolean_settings` logger threading is backward compatible (the added parameters default to `Logger::Root()`, so existing call sites are untouched) and only changes where the "Processed fully in 2D" family of messages are routed, not any geometry. The `assign_type` Type-suffix fallback can only ever add the exact strip-`Type` pairing that the schema already confirms exists, so it cannot loosen rejection of a genuine mismatch.

On the parser gap you flagged in "Scope not covered here" (the `IfcPropertySetDefinitionSet` SET-of-select divert-storage case): I had a look and I think your instinct to leave it is right, it is not actually broken. The divert mechanism in `IfcParse.cpp` appears in three structurally identical forms (the scalar case, the single-level SET sibling case, and the doubly-nested SET-of-SET case), all guarded by the same `!inst->declaration().as_entity()` check that redirects the write into the wrapper instance created in the first pass. The SET sibling branch does fire correctly once the input is well formed, it just was never exercised because the only fixture that reached it was the mis-tagged `IFC2X3` one, where the wrapper is never created in the first place. With the header fixed to `IFC4`, `ifcopenshell.open()` returns `RelatingPropertyDefinition` as a proper `IfcPropertySetDefinitionSet` with `.is_a()`, so the path is validated by your own fixture correction rather than needing a parser change. (The doubly-nested branch is unreachable for this specific select, which only appears as a scalar attribute, so it stays dead code either way.)

Nice bit of archaeology on the whole set. Happy to run anything else against the Blender or local-build side if useful.

Verified with the assistance of an AI tool.

BIMvoice:
Both current CI failures traced down, neither needs new investigation on your end:

**`compile-and-test` (bsDD test failures, `KeyError: 'classProperties'`/`'classRelations'`)**: not caused by this PR. It's the same root cause #8557 already fixes (`test_get_class_relations` was calling the wrong bSDD client method) - that fix just hasn't merged into `v0.8.0` yet. Once it does, a rebase/merge from `v0.8.0` should clear this automatically.

**`lint-formatting` / `ty check` (`unused-ignore-comment` at `src/ifcopenshell-python/ifcopenshell/draw.py:545`)**: this one is caused by this PR - the regenerated `ifcopenshell_wrapper.pyi` (from the `Regenerate ifcopenshell_wrapper.pyi` commit) produces a more accurate stub for `arrange_polygons`, which makes the existing `# ty: ignore[too-many-positional-arguments]` on that line stale. Confirmed `v0.8.0` main currently passes `ci-lint` cleanly, so this is specific to the new stub in this branch. Should just need deleting that now-unnecessary ignore comment.

sboddy:
Rebased this branch onto the current v0.8.0 tip and removed one commit in
the process: `cec556594d` ("Regenerate ifcopenshell_wrapper.pyi and add a
generator script").

That commit conflicted with upstream's own recent stub-sync commits (e.g.
f0117c60b3, 0e5223a30d) — both regenerated
`ifcopenshell_wrapper.pyi`, but in genuinely different styles (`name: Any`
declarations with a license header/docstrings vs. `name = ...` assignments,
alphabetized, without them). Resolving that automatically would have meant
guessing which format is intended going forward and risked silently
reverting one side's work, so instead of merging it here I split it out to
a separate local branch to regenerate properly against a real build and
send as its own follow-up PR.

The remaining 4 commits on this branch are unaffected and now sit cleanly
on the latest v0.8.0 with no touching of `ifcopenshell_wrapper.pyi`.

(Branch cleanup done with the assistance of an AI coding tool.)

aothms (approved):

PR #8558, opened 2026-07-12
Fix ci-bonsai-daily: update test_drawing mocks for the DRAWINGS parent group

BIMvoice:
## Summary

`ci-bonsai-daily` has a fresh regression: `test/core/test_drawing.py::TestAddDrawing::test_run` and `TestDuplicateDrawing::test_run` fail on every run since 2026-07-03 with:

```
bonsai/core/drawing.py:307: in add_drawing
    for group in ifc.get().by_type("IfcGroup"):
E   AttributeError: Interface 'bonsai.core.tool.Ifc' has no attribute 'by_type'.
```

Root cause: PR #7093 ("Organize drawings under a parent called DRAWINGS in ifc groups and ifc documents", commit `20b68ce0a`) added `ifc.get().by_type(...)` find-or-create logic to `core.drawing.add_drawing`/`duplicate_drawing` but never updated the tests. With `ifc.get()` unmocked, the Prophecy framework returns the `Ifc` tool interface stub, which has no `by_type` (that only exists on a real `ifcopenshell.file`).

## Fix

Test-only. Mock `ifc.get()` to return a minimal file stub exposing `by_type`, and add the create-parent calls the new logic predicts. Per the "test all cases" standard, also add reuse-branch tests (an existing `DRAWINGS` group/document is found) so **both** control-flow paths are covered for `add_drawing` and `duplicate_drawing`.

Production code (`core/drawing.py`) is correct and **untouched** — a real `ifc.get()` returns an `ifcopenshell.file`, which has `by_type`.

## Verification

```
test/core/test_drawing.py -> 40 passed (incl. the 2 previously-failing + 2 new reuse-branch tests)
full test/core/ suite -> unaffected
```
Verified against the current `origin/v0.8.0` `core/drawing.py`.

Note: this greens the fresh regression; `ci-bonsai-daily` also has a separate chronic cluster of `TestImplementsTool` interface-drift failures (test_cost/surveyor/ifcgit/classification/project/blender) that predate this and need their own triage.

### DIFF
--- a/src/bonsai/test/core/test_drawing.py
+++ b/src/bonsai/test/core/test_drawing.py
@@ -20,6 +20,38 @@ import bonsai.core.drawing as subject
 from test.core.bootstrap import Prophecy, blender, collector, drawing, geometry, ifc
 
 
+class _FakeEntity(str):
+    """A ``str`` stand-in for an IFC entity (e.g. an ``IfcGroup`` or
+    ``IfcDocumentInformation``).
+
+    It serializes and compares as its identifier, so it can be passed straight
+    into ``ifc.run(...)`` calls just like the plain strings used elsewhere in
+    these Prophecy-based tests, while also exposing arbitrary attributes (e.g.
+    ``Name``/``ObjectType``/``Scope``) for the plain-Python ``by_type()``
+    lookups in ``core.drawing`` (the "find or create the DRAWINGS parent
+    group/document" logic).
+    """
+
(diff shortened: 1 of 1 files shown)


BIMvoice:
Closing: superseded by 3a8619726b (Andrej730, merged 2026-07-14), which solves the same test_drawing breakage more cleanly via the ensure_drawings_parent_group/document API instead of mocking a fake IFC file. test/core/test_drawing.py passes on current v0.8.0.

PR #8562, opened 2026-07-12
Fix ci-bonsai-daily: avoid tripled repo path breaking BDD link_ifc tests

BIMvoice:
## Summary

A cluster of BDD failures in `ci-bonsai-daily` (the `test_link_ifc__*` scenarios in `src/bonsai/test/bim/feature/project.feature`) are caused by a **CI-config path collision**, not a source bug.

The `update-extensions-repo-and-run-tests` job has no top-level `actions/checkout`, so GitHub Actions' default workspace is already `work/IfcOpenShell/IfcOpenShell` (the standard `<org>/<repo>` layout — both are named `IfcOpenShell`). The job then does `git clone …/IfcOpenShell.git IfcOpenShell`, cloning into a **third** `IfcOpenShell` segment: `…/IfcOpenShell/IfcOpenShell/IfcOpenShell`. The BDD fixtures template `Path.cwd()` (`test/bim/test_feature.py:54`) into `bim.link_ifc` file paths, so those tests look for fixture files under the tripled path and hit `FileNotFoundError` in `LinkIfc.link_ifc`.

## Fix

Rename the manually-cloned directory to `ifcopenshell_repo` and update its three dependent references (`reregister_bonsai.py`, `ifcsverchok`, `bonsai` `cd`s). Pure rename; removes the collision.

## Verification

Diagnosed by exact command reconstruction — the reconstructed path matches the failing string character-for-character. Final confirmation is the next `ci-bonsai-daily` run (this job needs GitHub-hosted runners and can't be reproduced locally), so a maintainer should confirm on the scheduled run before/after merge.

### DIFF
--- a/.github/workflows/ci-bonsai-daily.yml
+++ b/.github/workflows/ci-bonsai-daily.yml
@@ -128,12 +128,16 @@ jobs:
           blender --command extension install-file -r user_default -e $bonsai_zip
           blender --command extension list
 
-          git clone https://github.com/IfcOpenShell/IfcOpenShell.git IfcOpenShell
+          # NOTE: cloned into "ifcopenshell_repo" (not "IfcOpenShell") to avoid nesting under
+          # the runner's default workspace, which is already .../work/IfcOpenShell/IfcOpenShell.
+          # Naming this clone "IfcOpenShell" produced a tripled .../IfcOpenShell/IfcOpenShell/IfcOpenShell
+          # path that broke test fixture path resolution (e.g. bim.link_ifc FileNotFoundError).
+          git clone https://github.com/IfcOpenShell/IfcOpenShell.git ifcopenshell_repo
 
           # Reregister Bonsai.
           # Note that running it in background might miss some errors
           # (e.g. tools are not registered in background mode).
-          blender --background --python IfcOpenShell/src/bonsai/scripts/reregister_bonsai.py
+          blender --background --python ifcopenshell_repo/src/bonsai/scripts/reregister_bonsai.py
 
           # Install sverchok.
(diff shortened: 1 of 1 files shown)


Moult:
Sorry I don't understand this, a triply nested `IfcOpenShell/IfcOpenShell/IfcOpenShell` dir is awkward but why would it break tests? `{cwd}` is intended to resolve to `src/bonsai` where the tests are being run, right? (or in this case, `IfcOpenShell/IfcOpenShell/IfcOpenShell/src/bonsai`)

BIMvoice:
You're right, and I owe you a correction. I re-traced this end to end and the tripled path is not the cause. {cwd} and our path helpers (tool.Ifc.resolve_uri/get_uri) just use os.path.normpath/relpath, so they resolve correctly no matter how many IfcOpenShell segments are stacked, since each level is a real directory. The CI log confirms it: bim.link_ifc successfully resolves the source .ifc path and starts processing it.

The actual failure is in LoadLink.link_ifc() (operator.py around 1580-1592): when a scenario calls use_cache=False, should_clear_cache() returns True without checking whether the .ifc.cache.blend sidecar exists, so os.remove() is called on a file that was never created (those cache files are gitignored, never present in a fresh clone). ci-bonsai-daily clones fresh every run, so the first scenario touching each fixture hits FileNotFoundError. This regressed in 35e3d9c42e back in March.

Good news: we already have the correct fix open as #8357 (it reorders should_clear_cache so the existence check runs first, so os.remove is never called on a missing path). So I am closing this PR as the wrong diagnosis, #8357 is the real one for these link_ifc failures. Sorry for the noise, and thanks for pushing on it instead of taking the diagnosis at face value.

PR #8563, opened 2026-07-12
Fix ci-bonsai-daily: drop never-wired Surveyor z-rotation scaffolding

BIMvoice:
## Summary

`test_surveyor.py::TestImplementsTool` fails because `core.tool.Surveyor` declares `get_z_rotation`/`set_z_rotation` abstract, but `tool.surveyor.Surveyor` never implements them, so `Surveyor()` can't instantiate.

Those two methods have exactly one caller: `core.product.copy_z_rotation_to_selected`. That function has **zero callers anywhere** (no operator / panel / menu wires it up), and `bonsai.core.product` is imported by nothing. The introducing commit (`230cbe1fd8`) even has a commit message describing unrelated functions that don't exist in the tree — i.e. unfinished, mismatched scaffolding, never a live feature.

## Fix

Rather than guess the underspecified world-vs-local z-rotation math to implement a feature nobody calls, remove the two dead abstract declarations and the fully-dead `core/product.py` module.

## Verification

Headless Blender: `test/tool/test_surveyor.py::TestImplementsTool` goes from `TypeError: Can't instantiate abstract class Surveyor` to **PASSED**. Confirmed `bonsai.core.product` has no importers and `copy_z_rotation_to_selected` has no callers on the current tip.

If the copy-z-rotation feature is wanted later, it should return with real UI wiring, tests, and a specified rotation model. Happy to keep the scaffolding instead if a maintainer prefers — just say so.


BIMvoice:
This PR removed the dead `Surveyor.get_z_rotation`/`set_z_rotation` abstract methods and the unused `core/product.py` module to fix `test/tool/test_surveyor.py::TestImplementsTool`, which was failing because `Surveyor` could not be instantiated.

@Andrej730 landed the same fix independently in commit 9e25c12b16 ("surveyor: drop never used dead code") on v0.8.0 on July 14. The latest ci-bonsai-daily run (https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/29981825298, July 23) confirms `test_surveyor.py` now passes on v0.8.0 without this PR.

Closing since the fix already landed through a different commit. Thanks for looking at the same problem, and feel free to reopen if the underlying issue resurfaces.

PR #8571, opened 2026-07-13
Fix ci-bonsai-daily: assign_class drops active_object; assign_material usage type

BIMvoice:
## Summary

~16 failing BDD scenarios in the material / assign-type area trace to two root causes:

### 1. `AssignClass` drops the active object
`AssignClass._execute` (`bim/module/root/operator.py`) relinks objects into new collections (e.g. types into the hidden `IfcTypeProduct` collection) and then restores the selection via `tool.Blender.validate_object_selection`. Without a `context.view_layer.update()` in between, a just-relinked (but valid) active object is seen as "not in the current view layer" and dropped — so `context.active_object` becomes `None` for the rest of the caller, and every scenario that reads `active_object` after assigning a type fails. Add `view_layer.update()` before the selection-restore (an idiom already used ~15× elsewhere).

### 2. `assign_material` passes a plain material into a `...Usage` type
`core.assign_material` forwarded the selected plain `IfcMaterial` into `ifcopenshell.api.material.assign_material` even when auto-upgrading to a `...Usage` type, which only accepts a material *set* or `None` — tripping an intentional API assertion. Pass `material=None` for the usage branch and let the existing `add_material_to_set` follow-up populate the derived set.

## Verification

Headless Blender, material / assign_type / assigned-material BDD cluster: **23 failed / 14 passed → 7 failed / 30 passed**. The 7 remaining are an unrelated stale-STEP-id fixture cluster (`Item/IfcFace/NN`), handled separately.

### DIFF
--- a/src/bonsai/bonsai/bim/module/root/operator.py
+++ b/src/bonsai/bonsai/bim/module/root/operator.py
@@ -380,6 +380,14 @@ class AssignClass(bpy.types.Operator, tool.Ifc.Operator):
         # TODO: reload representation might lead to the object being replaced by object of the other type.
         # We probably should track it somehow and keep the original selection.
 
+        # Objects may have just been relinked into new collections (e.g. types
+        # into the "IfcTypeProduct" collection) above. Make sure the view
+        # layer's object bases are up to date before validating the selection
+        # below, otherwise a just-relinked (but perfectly valid) active object
+        # can be seen as "not in the current view layer" and be dropped,
+        # leaving `context.active_object` as None for the rest of the caller.
+        context.view_layer.update()
+
         # Validate selection and reapply it.
         current_selection = tool.Blender.validate_object_selection(*current_selection)
         tool.Blender.set_objects_selection(*current_selection)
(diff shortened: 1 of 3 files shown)


BIMvoice:
Pushed two follow-up commits for the one material scenario this branch did not yet fix, "Edit layer item defaults null IsVentilated to FALSE in UI" (IndexError on `by_type('IfcMaterialLayer')[0]`).

Root cause: when the usage path assigns material=None on an occurrence that has no type, the API derives a new empty IfcMaterialLayerSet and wraps it in the usage. core.assign_material then reads the assignment back with should_inherit=False, which returns the Usage entity itself. The Usage fails is_a_material_set, so the add_material_to_set step that seeds the first layer never runs and the set stays empty. Fixed by unwrapping the usage to its ForLayerSet/ForProfileSet before the set population step. add_material_to_set only seeds sets that are still empty, so a populated layer set inherited from a type is untouched.

The scenario then exposed a second latent problem: its final step is "Then I evaluate expression", but the step was only registered with @given/@when, so it failed with StepDefinitionNotFoundError. Registered the step for Then as well.

Verified in headless Blender 5.1.2 against current v0.8.0 (cherry-picks cleanly): full material.feature goes from 1 failed, 33 passed (this branch as it was) to 34 passed, 0 failed.

This update was made with the assistance of an AI tool.

Issue #8183, opened 2026-06-18
Regression: cannot load project library which contain no IfcProject ?


CyrilWaechter:
### Bug Description

Error occurred when you try to load following attached project library:
1. Create a new IFC project
2. Load attached project library using `Project Setup` => ` Project Library` => `Select Library File` from `Project Overview` tab.
3. See error message

From my current understanding the issue occurred because there is no `IfcProject` in the IFC file. Although we have an `IfcProjectLibrary`. As far as I know having an `IfcProject` is not mandatory.

### Attachments

Ifc file containing the project library: [MinergieECO.zip](https://github.com/user-attachments/files/29088400/MinergieECO.zip)

### Debug and Error Output

```true
Traceback (most recent call last):
  File "/home/cyril/.config/blender/5.1/extensions/.local/lib/python3.14/site-packages/bonsai/bim/module/project/prop.py", line 85, in get_project_libaries
    ProjectLibraryData.load()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/cyril/.config/blender/5.1/extensions/.local/lib/python3.14/site-packages/bonsai/bim/module/project/data.py", line 129, in load
    cls.data["parent_libraries_enum"] = cls.parent_libraries_enum()
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/cyril/.config/blender/5.1/extensions/.local/lib/python3.14/site-packages/bonsai/bim/module/project/data.py", line 165, in parent_libraries_enum
    project = library_file.by_type("IfcProject")[0]
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

aothms:
I think this might be something that was changed inadvertently because previously similar code did spell IfcProjectLibrary 

https://github.com/IfcOpenShell/IfcOpenShell/blame/56f0c19a85f626dc0f1d07183788ce815fb7b37a/src/bonsai/bonsai/bim/module/project/data.py#L130

I'm not sure if there are people online/active these days that are familiar with this part of the code. Maybe experiment a bit with changing this to IfcContext/IfcProjectLibrary. The line above `library_file.schema == "IFC2X3" -> return` seems to indicate that it should be one of these types because these are exactly the types that did not exist yet in ifc2x3

Gorgious56:
[...]

PR #8573, opened 2026-07-13
Fix ci-bonsai-daily: ProjectLibraryData duplicate parent-library enum entry

BIMvoice:
## Summary

Reworked again per Moult's follow up: remove the load time repair entirely.
"If the IFC is invalid, our default position is almost always to not support
it (unless it is ubiquitous)."

Checked ubiquity: every library file bonsai ships under
`src/bonsai/bonsai/bim/data/libraries` (9 files, IFC2X3 through IFC4X3) has an
IfcProject with the IfcProjectLibrary declared to it via IfcRelDeclares. A
library only file with no IfcProject is not a common authoring pattern, it is
one outlier report (#8183). So per Moult's default, we do not support it.

## What changed from the previous revision

Removed `tool.Project.ensure_project_context()` and `open_library_file()`
entirely. `SelectLibraryFile`'s three load sites are back to plain
`ifcopenshell.open`, no wrapper.

`get_root_context()` is back to a single line, `ifc_file.by_type("IfcProject")[0]`,
no IfcProjectLibrary fallback. A file without IfcProject now raises
`IndexError` instead of being treated as valid. This also makes
`AddProjectLibrary`'s nest-under-library branch dead code, since
`root_context` can now only ever be an IfcProject, so that branch is removed
too.

The one-line enum dedup guard from the very first version of this PR is also
removed. Once `get_root_context()` can only return an IfcProject (or raise),
an IfcProject id can never collide with a library id, so the guard had
nothing left to guard against. `parent_libraries_enum()` is now byte for byte
identical to what is already on `v0.8.0`.

`test_project_library_data.py` no longer builds an invalid library-only
model as a fixture. It uses one spec-valid fixture (IfcProject +
IfcProjectLibrary declared to it, matching bonsai's own shipped library
files) for the ci-bonsai-daily regression test and the refresh/add-library
operator tests, plus one explicit test asserting `get_root_context` raises
for a file without IfcProject, documenting that this is intentionally
unsupported.

## Net result

The code only assumes and handles spec-valid models (single IfcProject).
Library-only files are not accommodated. The ci-bonsai-daily test passes
because it now exercises a valid model, not because invalid data is patched
around.

## Verification

Live in headless Blender (isolated profile, source-loaded, never the real
profile): confirmed `ensure_project_context`/`open_library_file` are gone,
that a library-only file now raises via `get_root_context` instead of being
silently handled, that `ProjectLibraryData.load()`, `refresh_library` and
`add_project_library` all work correctly on a spec-valid model with unique
enum keys, and spot-checked that all 9 shipped library files already have an
IfcProject.

### DIFF
--- a/src/bonsai/bonsai/bim/module/project/data.py
+++ b/src/bonsai/bonsai/bim/module/project/data.py
@@ -165,6 +165,12 @@ class ProjectLibraryData:
         root = tool.Project.get_root_context(library_file)
         results.append((str(root.id()), f"{root.is_a()} {root.Name or 'Unnamed'}", root.Description or ""))
         for library_id, data in cls.data["project_libraries"].items():
+            # Library-only files have an IfcProjectLibrary as their root context, and
+            # project_libraries() collects every IfcProjectLibrary in the file (root
+            # included). Skip it here since it was already added above, otherwise the
+            # root library is listed twice with colliding enum keys.
+            if library_id == root.id():
+                continue
             results.append((str(library_id), data["Name"] or "Unnamed", data["Description"] or ""))
         return results
 
--- a/src/bonsai/bonsai/bim/module/project/data.py
+++ b/src/bonsai/bonsai/bim/module/project/data.py
@@ -165,10 +165,7 @@ class ProjectLibraryData:
         root = tool.Project.get_root_context(library_file)
         results.append((str(root.id()), f"{root.is_a()} {root.Name or 'Unnamed'}", root.Description or ""))
         for library_id, data in cls.data["project_libraries"].items():
-            # Library-only files have an IfcProjectLibrary as their root context, and
-            # project_libraries() collects every IfcProjectLibrary in the file (root
-            # included). Skip it here since it was already added above, otherwise the
-            # root library is listed twice with colliding enum keys.
+            # Defensive guard only; see tool.Project.ensure_project_context for the fix.
             if library_id == root.id():
                 continue
             results.append((str(library_id), data["Name"] or "Unnamed", data["Description"] or ""))
--- a/src/bonsai/bonsai/bim/module/project/data.py
+++ b/src/bonsai/bonsai/bim/module/project/data.py
@@ -165,9 +165,6 @@ class ProjectLibraryData:
         root = tool.Project.get_root_context(library_file)
         results.append((str(root.id()), f"{root.is_a()} {root.Name or 'Unnamed'}", root.Description or ""))
         for library_id, data in cls.data["project_libraries"].items():
-            # Defensive guard only; see tool.Project.ensure_project_context for the fix.
-            if library_id == root.id():
-                continue
             results.append((str(library_id), data["Name"] or "Unnamed", data["Description"] or ""))
         return results
 
--- a/src/bonsai/bonsai/bim/module/project/data.py
+++ b/src/bonsai/bonsai/bim/module/project/data.py
@@ -162,7 +162,7 @@ class ProjectLibraryData:
         library_file = IfcStore.library_file
         if library_file is None or library_file.schema == "IFC2X3":
             return results
-        root = tool.Project.get_root_context(library_file)
+        root = library_file.by_type("IfcProject")[0]
         results.append((str(root.id()), f"{root.is_a()} {root.Name or 'Unnamed'}", root.Description or ""))
         for library_id, data in cls.data["project_libraries"].items():
             results.append((str(library_id), data["Name"] or "Unnamed", data["Description"] or ""))
--- a/src/bonsai/bonsai/bim/module/project/operator.py
+++ b/src/bonsai/bonsai/bim/module/project/operator.py
@@ -759,21 +759,22 @@ class EditProjectLibrary(bpy.types.Operator):
         attributes = bonsai.bim.helper.export_attributes(props.project_library_attributes)
         ifcopenshell.api.attribute.edit_attributes(library_file, project_library, attributes)
 
-        # Update parent library.
+        # Update parent library. Tear down the old IfcRelDeclares/IfcRelNests before
+        # creating the new one; a library must have exactly one of the two, never both.
         previous_parent_library = tool.Project.get_parent_library(project_library)
         new_parent_library = library_file.by_id(int(props.parent_library))
         if previous_parent_library != new_parent_library:
-            if previous_parent_library is None:
-                # Edited library was a root in a library-only file; nest it under the new parent.
+            if previous_parent_library is not None:
+                if previous_parent_library.is_a("IfcProject"):
+                    ifcopenshell.api.project.unassign_declaration(
+                        library_file, [project_library], previous_parent_library
+                    )
+                else:
+                    ifcopenshell.api.nest.unassign_object(library_file, [project_library])
+            if new_parent_library.is_a("IfcProject"):
+                ifcopenshell.api.project.assign_declaration(library_file, [project_library], new_parent_library)
+            else:
                 ifcopenshell.api.nest.assign_object(library_file, [project_library], new_parent_library)
-            elif previous_parent_library.is_a("IfcProject"):
-                # Then new one is IfcProjectLibrary.
-                ifcopenshell.api.nest.assign_object(library_file, [project_library], new_parent_library)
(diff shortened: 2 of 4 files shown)


Moult:
> For a library-only file (no IfcProject) ...

IFC requires an IfcProject. There is no such thing as a library-only file.

BIMvoice:
This isn't a premise this PR invents — it's already established, merged behaviour. `get_root_context()` (added in `260a38706`, closing #8183 "Regression: cannot load project library which contain no IfcProject") exists specifically because a user hit a real crash on exactly this: a file whose only `IfcContext` is an `IfcProjectLibrary`, no `IfcProject` present. Its docstring:

> Per IFC4+, IfcContext is the abstract supertype of IfcProject and IfcProjectLibrary; library-only files legitimately contain only IfcProjectLibrary as their root context.

That fix (PR #8184) already routes `parent_libraries_enum`, `RefreshLibrary`'s tree view, and `AddProjectLibrary` through `get_root_context()` to handle this. This PR doesn't add or endorse library-only files as a concept — it fixes a narrower bug *inside* that already-accepted code path: when `get_root_context()` returns an `IfcProjectLibrary` (i.e. the library-only case), that same entity also gets picked up by the `project_libraries()` loop a few lines later, so it's added to the enum twice with colliding keys — which is what broke `TestImplementsTool`/the duplicate-enum test. The fix just skips re-adding the id that's already there as root.

Happy to be wrong about the underlying IFC4 legitimacy question — if that's the actual concern, it's really about whether #8183/#8184 should be revisited, not about this one-line dedup fix, since the duplicate-key bug exists either way once `get_root_context()` can return an `IfcProjectLibrary`.

Moult:
Ping @Gorgious56 - as it seems that's where the premise of a library only file leaked in.

@BIMvoice no, it is not allowed to have a model with IfcProjectLibrary only. It is a strict rule in IFC (from Project Context concept template):

> All project data sets shall contain a single [IfcProject](http://localhost:8001/HTML/lexical/IfcProject.htm) instance indicating overall context and a directory of objects contained within.

And in the docs for IfcProjectLibrary, it explains clearly how to link back to the (required) IfcProject:

> Instances of [IfcProjectLibrary](http://localhost:8001/HTML/lexical/IfcProjectLibrary.htm) are assigned to the project context using ...

... so please re-evaluate and address the more fundamental concern.

BIMvoice:
@Moult you're right, and thanks for pushing back on this. Reworked the fix.

The invalid premise traces to `tool.Project.get_root_context()` (260a387069,
#8184): it fell back to using the top-level `IfcProjectLibrary` as the file's
root context whenever `IfcProject` was missing, on the assumption that a
library-only file is a legitimate structure. It isn't, per the Project
Context concept template you quoted. That's also exactly why the duplicate
enum key happened: `project_libraries()` walks every `IfcProjectLibrary`
including the root, so once `get_root_context()` started returning that same
root library, it got listed twice.

Instead of special-casing "root can be a library," I added
`tool.Project.ensure_project_context()`, which repairs a file missing
`IfcProject` by creating one and declaring the file's root-level
`IfcProjectLibrary` instances to it via `IfcRelDeclares`, matching the
pattern already documented in `ifcopenshell.api.project.assign_declaration`.
All three places that load a library file now go through this repair
(`tool.Project.open_library_file()`), so downstream code always sees a
spec-valid model with a single `IfcProject`. The one-line dedup guard from
before is kept only as a defensive fallback, not as the fix.

Verified live in headless Blender against the real code path: a malformed
library file (no `IfcProject`, mirroring the file from #8183) now gets
repaired on load, `ProjectLibraryData`, refresh and add-library all work
correctly on the result, and already-valid files are untouched.

Updated PR description above with the full writeup.

Moult:
Please remove the patching code (e.g. `ensure_project_context` etc) - if the IFC is invalid, our default position is almost always to not support it (unless it is ubiquitous).

BIMvoice:
@Moult done, the patching is removed.

Removed ensure_project_context and open_library_file entirely, and
SelectLibraryFile's three load sites are back to plain ifcopenshell.open.
get_root_context is back to a single line, ifc_file.by_type("IfcProject")[0],
no IfcProjectLibrary fallback. A file without IfcProject now raises
IndexError instead of being treated as valid.

On the ubiquity question: checked all 9 library files bonsai ships under
bim/data/libraries (IFC2X3 through IFC4X3), and every one of them has an
IfcProject with the IfcProjectLibrary declared to it via IfcRelDeclares. The
file from #8183 is the only library-only file I've seen, not a common
pattern, so per your default we do not support it.

Also removed the enum dedup guard from the very first version of this PR.
Once get_root_context can only return an IfcProject or raise, an IfcProject
id can never collide with a library id, so there is nothing left for that
guard to catch. parent_libraries_enum is now identical to what is already on
v0.8.0.

Reworked the test fixture too: it used to build a library-only model and
assert that as correct behaviour. Now it only builds a spec-valid model
(matching bonsai's own shipped libraries), plus one test asserting
get_root_context raises for a file without IfcProject, so the unsupported
case is documented rather than silently handled.

The ci-bonsai-daily test now passes because it exercises a valid model, not
because we patch around an invalid one.

Moult:
1. `get_root_context` is unnecessary, just by_type("IfcProject")[0] since that's a rule in IFC.
 2. Remove superfluous code comments
 3. Isn't `get_parent_library` meant to return None if it is a top level library? Why change that behaviour?

BIMvoice:
Addressed all three points.

1. Removed get_root_context, inlined ifc_file.by_type("IfcProject")[0] at its three call sites in RefreshLibrary, AddProjectLibrary and parent_libraries_enum.
2. Removed a leftover explanatory comment in the test file and trimmed the get_parent_library docstring to one line.
3. We did not change get_parent_library's behavior anywhere in this PR. Its logic (Nests, then HasContext, then None) is identical to origin/v0.8.0, going back to the commit that introduced it. It returns None only for a genuinely orphaned library with neither relationship, and returns the project for a library declared directly under IfcProject, exactly as before. Only its docstring wording changed, which I suspect is what read as a behavior change; it is trimmed down now. Added a test asserting the None case explicitly.

Verified live in headless Blender (isolated profile, source loaded): all project module tests pass, including the new get_parent_library None case. Ran the full test/bim suite before and after this commit, identical 82 pre-existing unrelated failures both times, no new failures.

Moult:
Please check again, because the fundamental assertion `assert tool.Project.get_parent_library(root) is None` has changed to `assert tool.Project.get_parent_library(root) == project`, surely that breaks something.

There must also be other things which are incorrect from this change that need updating (otherwise this PR is incomplete) - i.e. now that we know that an IfcProject must exist, please review @Gorgious56 's changes surrounding this and make sure things are still correct there.

BIMvoice:
You're right to push on this, thank you. Traced it through and here's what I found.

The assertion change itself is correct. In the old library-only model a top-level library had neither Nests nor HasContext, so None meant top level. Now that every file has a guaranteed IfcProject, a top-level library is declared to it via IfcRelDeclares, so get_parent_library correctly resolves the project through that relationship instead of falling through to None.

I audited every caller. get_project_hierarchy already keys top-level libraries under the project for exactly this reason, so the library tree still renders roots correctly, verified live.

But you were right that something was incomplete. EditProjectLibrary, the operator that lets you move a library to a new parent, still had the old assumption baked in. It assumed a top-level library had no existing relationship to clean up before nesting it elsewhere, and that moving a library back to the project needed no new relationship because it was already assigned by default. Both of those were only true when a top-level library held nothing. Now it actually holds an IfcRelDeclares, so the old code left it dangling alongside the new IfcRelNests when moving it into another library (an invalid double parentage), and orphaned it entirely (no relationship at all) when moving it back to the project. I reproduced both in a plain Python repro before touching anything.

Fixed it so the operator now tears down whichever relationship the library previously had before establishing whichever one the new parent needs, rather than assuming.

Added the missing third case to the get_parent_library tests (nested under a library), alongside the existing project-declared and orphaned cases, plus two regression tests for both reparenting directions. Confirmed they fail on the old operator code and pass with the fix.

Ran the full test/bim suite before and after. Same 123 pre-existing failures in both runs, unrelated to this area, diffed the failing test names to confirm. The fix adds 3 newly passing tests and changes nothing else.

Issue #3, opened 2015-10-19
current master is not buildable


bxabi:
Hi, 

when I'm trying to build, at the make command I have the following error:
```
Scanning dependencies of target _ifcopenshell_wrapper
[ 78%] Building CXX object ifcwrap/CMakeFiles/_ifcopenshell_wrapper.dir/IfcPythonPYTHON_wrap.cxx.o
In file included from /home/bxabi/Programs/IfcOpenShell/cmake/build/ifcwrap/IfcPythonPYTHON_wrap.cxx:3144:0:
/home/bxabi/Programs/IfcOpenShell/src/ifcwrap/../ifcgeom/IfcGeomIterator.h:89:3: error: ‘Kernel’ does not name a type
   Kernel kernel;
   ^
/home/bxabi/Programs/IfcOpenShell/src/ifcwrap/../ifcgeom/IfcGeomIterator.h:95:3: error: ‘IfcSchema’ does not name a type
   IfcSchema::IfcRepresentation::list::ptr representations;
   ^
/home/bxabi/Programs/IfcOpenShell/src/ifcwrap/../ifcgeom/IfcGeomIterator.h:96:3: error: ‘IfcSchema’ does not name a type
   IfcSchema::IfcRepresentation::list::it representation_iterator;
   ^
```

and many similar errors follow.

Is this happening only for me? What can cause it?

Stinkfist0:
What compiler and platform?

bxabi:
it is on Fedora 22, 64 bit.
I built successfully the OpenCascade community edition before, and set it as:
OCC_INCLUDE_PATH="/usr/local/include/oce"
OCC_LIBRARY_PATH="/usr/local/lib"
before calling the cmake.

I can provide more information, the full log, environment if needed.

bxabi:
I continued to try to get it build.
It looks like it is an issue with the OpenCascade compatibility.
This time I used the OpenCascade 6.9.1.
I succeeded to reduce to a single error configuring the cmake better using the cmake-gui.
```
[ 95%] Building CXX object CMakeFiles/IfcConvert.dir/home/bxabi/Programs/IfcOpenShell/src/ifcconvert/util.cpp.o
/home/bxabi/Programs/IfcOpenShell/src/ifcconvert/SvgSerializer.cpp: In member function ‘virtual void SvgSerializer::write(const IfcGeom::BRepElement<double>*)’:
[...]

PR #8576, opened 2026-07-13
docker: make the build env work on macOS / Apple Silicon hosts

BIMvoice:
## Summary

Follow-up to #8564 (thanks @sboddy — the docker env is genuinely useful). Three small host-portability fixes so `docker/` runs on **macOS / Apple Silicon** as well as Linux. All three are **no-ops on native amd64 Linux**.

1. **Dockerfile GID collision** — macOS's default primary group `staff` is GID 20, which already exists as `games` in `rockylinux:9`, so `groupadd -g 20 builder` aborted the image build. Only `groupadd` when the GID is free (`getent group "${USER_GID}" >/dev/null || groupadd ...`); `useradd -g` accepts the existing GID.
2. **`ifcos_env` `sed -si`** — GNU-only; BSD/macOS sed errors `illegal option -- s`, breaking the `UNIQUE_ID` rewrite. Replaced with a portable `sed > tmp && mv` (verified against macOS BSD sed).
3. **Image arch / pull-instead-of-local** — `create()` built without `--platform`, so on arm64 the local image is tagged `linux/arm64` while `compose.yaml` pins `platform: linux/amd64`; compose then treats the local image as absent and tries to pull the local-only tag from Docker Hub (access denied). Build with explicit `--platform linux/amd64` so the local image matches the pin, plus `pull_policy: never` as a clear-error safety net.

## Caveat (honest)

On Apple Silicon the amd64 build runs under emulation and a cold full build is slow; `ccache` makes incremental rebuilds tolerable. A native Linux/Intel host or CI remains the better choice for routine use — but these fixes turn "hard broken" into "works, with a caveat" on macOS. Happy to adjust or drop any of these if you'd prefer a different approach.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -39,7 +39,15 @@ ENV PATH="/usr/lib/ccache:$PATH"
 # if your host user has a different UID/GID.
 ARG USER_UID=1000
 ARG USER_GID=1000
-RUN groupadd -g "${USER_GID}" builder \
+# groupadd fails outright if USER_GID is already taken by an existing
+# system group - which happens whenever a host's primary GID collides with
+# one baked into the rockylinux9 base image. The main real-world case is
+# macOS, where the default user's primary group is "staff" at GID 20, and
+# GID 20 is "games" on RHEL-family images. Only create the "builder" group
+# when that GID is actually free; otherwise useradd just attaches to
+# whichever group already owns it. Either way the builder user ends up
+# with the right GID for bind-mount ownership, which is all that matters.
+RUN (getent group "${USER_GID}" >/dev/null || groupadd -g "${USER_GID}" builder) \
     && useradd -m -u "${USER_UID}" -g "${USER_GID}" -s /bin/bash builder \
     && echo "builder ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/builder
 
--- a/docker/compose.yaml
+++ b/docker/compose.yaml
@@ -4,6 +4,14 @@ services:
     container_name: ifcopenshell-${UNIQUE_ID}
     image: ifcopenshell-build-env:updated
     platform: linux/amd64
+    # There's no `build:` section - the image is always produced ahead of
+    # time by `./ifcos_env create` (`docker build`, not `docker compose
+    # build`). Without this, a platform mismatch between the pin above and
+    # whatever's in the local image store makes compose treat the image as
+    # absent and fall back to pulling ifcopenshell-build-env:updated from
+    # Docker Hub, where it doesn't exist. Fail fast with a clear "not
+    # found" instead of an obscure registry access-denied error.
+    pull_policy: never
     volumes:
       - type: bind
         source: ../
--- a/docker/compose.yaml
+++ b/docker/compose.yaml
@@ -4,14 +4,6 @@ services:
     container_name: ifcopenshell-${UNIQUE_ID}
     image: ifcopenshell-build-env:updated
     platform: linux/amd64
-    # There's no `build:` section - the image is always produced ahead of
-    # time by `./ifcos_env create` (`docker build`, not `docker compose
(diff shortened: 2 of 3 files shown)


sboddy:
I (Steve) had a chat with Claude about your PR. The following is where we got to in our reasoning. I'm hoping these notes help you to create a solution that is better for your needs. If I had Apple hardware I'd have a go myself. Please consider it. Cheers!

Thanks for these — the `sed -si` portability fix and the `groupadd` GID-collision guard are both solid, verified against real cross-platform quirks, and clean no-ops on Linux/amd64. Happy to see those land.

On fix #3 (`--platform linux/amd64` + `pull_policy: never` to make the container buildable/runnable under emulation on Apple Silicon) — I want to flag something before merging, not as a refusal, more a "you may get a much better outcome going a different way" suggestion.

This `docker/` setup is a deliberate 1:1 port of `build_rocky.yml` — same Rocky9 container, same toolchain, same build invocation. But our actual macOS CI (`build_osx.yml`) isn't "the same build on a different arch" — it's a completely different pipeline: no Docker, no Linux container, native Homebrew toolchain running directly on `macos-14` runners, with real cross-compilation handling for x64-on-arm64 already solved (the `arch -x86_64` Homebrew reinstall, arch-correct `gmp`/`pkg-config`/`openssl` swaps).

That difference isn't incidental. Docker Desktop on macOS has no native macOS container runtime — it's Linux containers under the hood either way — so even with these fixes, the build produces a Linux ELF `.so`, not a macOS Mach-O one. That's fine for the "copy into `src/ifcopenshell-python/ifcopenshell/` and run pytest" verification path if you're doing that *inside* the container, but it can't be dropped into a real macOS Blender/Bonsai install the way `try()`/`SKILL.md`'s host-side pytest flow assumes — those run on the host's own Python, which can't load a Linux shared object. So beyond emulation being slow, the output isn't actually native-testable on macOS the way the Linux path is on Linux.

Given `build_osx.yml` already solves native building for both Intel and Apple Silicon Macs without any of this, it might be worth a look at whether a `docker/`-sibling script modeled on that (Homebrew installs, no container, calling `nix/build-all.py` directly on the host) gets you a better result — real native speed, and an artifact that's actually usable in a real macOS Blender install. Worth exploring if you're up for it; if it turns out to be a dead end for your use case, we can revisit merging this as-is.

BIMvoice:
You're right, and thanks for taking the time — that's a real architectural point, not a nitpick, and I should've thought it through before pushing the emulation fix.

Confirmed it myself against `build_osx.yml`: it's a genuinely different pipeline (native Homebrew on `macos-14` runners, no container), and the cross-compile handling it already has for x64-on-arm64 is exactly the kind of thing I'd have had to reinvent inside a Linux container for no benefit — since, as you said, Docker Desktop on macOS is a Linux VM regardless, so the container's `.so` is ELF, not Mach-O, and can't be loaded by a native macOS Blender/Bonsai Python either way. Good catch.

Pushed a narrowed version: dropped the `--platform linux/amd64` build flag and the `pull_policy: never` addition, kept just the two host-portability fixes that are correct on any platform:
- the GID-collision guard (macOS `staff`=20 colliding with Rocky's `games`=20)
- the `sed -si` → portable temp-file rewrite (BSD sed doesn't take GNU's `-si`)

Those aren't emulation-specific, they're real bugs in the script regardless of target arch, so hopefully still worth merging on their own.

For the actual goal (testing C++ patches against a real macOS Bonsai install before submitting), I'm going to try a native build mirroring `build_osx.yml`'s steps directly on the host — Homebrew deps + `nix/build-all.py`, no container. Checked `cache_dependencies.py`: the `build-outputs` pull is just a CI speed optimization (packs the `build/*/install` dir), `build-all.py` builds dependencies from source itself when it's absent, so it should work locally without needing the private cache repo — just slower on a cold run. Will report back if it works or if there's another wall.

PR #8572, opened 2026-07-13
Fix #8570: populate ifc5d IfcOpenShell QTO formulas for IfcSpace

BIMvoice:
## Summary

Fixes the `IfcSpace` case of #8570. The headless **IfcOpenShell** QTO calculator had every `Qto_SpaceBaseQuantities` formula set to `null` for `IfcSpace` in both `IFC4QtoBaseQuantities.json` and `IFC4X3QtoBaseQuantities.json`, so `qto.py`'s `if not formula: continue` skipped every quantity, no geometry task was queued, and spaces never appeared in results (`elements_quantified: 0`). The Blender calculator already computes these — they were just never ported to the `ifcopenshell.util.shape`-backed calculator.

## Fix

Map the eight computable quantities to existing `util.shape` functions, mirroring the Blender calculator (no new `shape.py` code):

| Quantity | Formula |
|---|---|
| GrossFloorArea | `gross_get_footprint_area` |
| NetFloorArea | `net_get_footprint_area` |
| GrossCeilingArea | `gross_get_top_area` |
| NetCeilingArea | `net_get_top_area` |
| GrossPerimeter | `gross_get_footprint_perimeter` |
| GrossVolume | `gross_get_volume` |
| NetVolume | `net_get_volume` |
| Height | `net_get_z` |

Left `null` (matching the Blender ruleset, **not** guessed): `GrossWallArea`, `NetWallArea`, `NetPerimeter` (Blender stub), and `FinishFloor/CeilingHeight` (Blender derives those from sibling `IfcCovering` decomposition geometry, which this per-element calculator architecture can't reach — noted for follow-up).

## Verification

IFC4 (a 4×3 space extruded 2.5 m), via the issue's exact CLI repro:
- **Before:** `{}` — `elements_quantified: 0`
- **After:** `GrossFloorArea 12`, `GrossPerimeter 14`, `Height 2.5`, `GrossVolume 30`, ceiling/net areas 12 — all exact matches to the extrusion geometry; `elements_quantified: 1`.

IFC4X3 formulas are identical and the formula→function resolution is schema-agnostic.

## Scope

Fixes `IfcSpace`. The 12 other all-null classes in the issue (`IfcDoor`, `IfcSite`, `IfcRailing`, …) are left as follow-up.

This change was made with the assistance of an AI tool.

PR #8578, opened 2026-07-13
Fix #8570: populate ifc5d IfcOpenShell QTO for 11 more classes

### LINKED ISSUE #8570: ifc5d IfcOpenShell QTO calculator never produces quantities for IfcSpace
### Bug Description

ifc5d's headless IfcOpenShell QTO calculator never produces quantities for IfcSpace - it fails silently, no error or warning.

src/ifc5d/ifc5d/IFC4QtoBaseQuantities.json — calculator IfcOpenShell, entity IfcSpace, qset Qto_SpaceBaseQuantities: all 13 formulas are null. Same in IFC4X3QtoBaseQuantities.json (key "IfcSpace + IfcSpaceType").

Because every formula is null, qto.py:IfcOpenShell.calculate() never populates gross_qtos/net_qtos for a space (if not formula: continue), so no geometry iterator task is ever queued for it and the element never lands in results - not even as an empty dict.

By contrast IFC4QtoBaseQuantitiesBlender.json's Blender calculator has real, working formulas for IfcSpace (get_gross_footprint_area, get_net_floor_area, get_height, ... in src/bonsai/bonsai/bim/module/qto/calculator.py), so the geometry is clearly quantifiable - these were just never ported to ifcopenshell.util.shape for the non-Blender calculator.

Same gap exists for 12 other classes where IfcOpenShell ruleset is all-null but Blender ruleset isn't: IfcBuildingElementProxy, IfcConstructionMaterialResource, IfcDistributionChamberElement, IfcDoor, IfcDuctFitting, IfcProjectionElement, IfcRailing, IfcRampFlight, IfcReinforcingElement, IfcSite, IfcSpaceHeater, IfcTank.

To reproduce:

    python3 -m ifcedit quantify run model.ifc IFC4QtoBaseQuantities --selector IfcSpace
    # {"ok": true, "rule": "IFC4QtoBaseQuantities", "elements_quantified": 0}

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
## Summary

Follow-up to #8572 (which fixed `IfcSpace`), completing #8570 for the remaining all-null `IfcOpenShell`-calculator classes the issue listed. Ported `Qto_*BaseQuantities` formulas for **11 classes** in both `IFC4QtoBaseQuantities.json` and `IFC4X3QtoBaseQuantities.json`:

`IfcBuildingElementProxy`, `IfcDistributionChamberElement`, `IfcDoor`, `IfcDuctFitting`, `IfcProjectionElement`, `IfcRailing`, `IfcRampFlight`, `IfcReinforcingElement`, `IfcSite`, `IfcSpaceHeater`, `IfcTank`.

Each quantity maps to the `ifcopenshell.util.shape` function the Blender calculator uses, matching existing precedent in the file (`IfcColumn`/`IfcSlab`/`IfcWindow`/`IfcStairFlight`/`IfcFooting`).

**`IfcConstructionMaterialResource`** (the 12th) is left fully null: it's not an `IfcProduct` (no `Representation`/`ObjectPlacement`), so QTO doesn't apply — `quantify()` returns `{}` without crashing.

Quantities left `null` match the Blender ruleset (weights, cross-section areas, and dimensions with no `util.shape` equivalent such as `IfcDoor` Perimeter and `IfcRampFlight` Width) — **not guessed**.

## Verification

Live on IFC4 (box geometry via `ifcopenshell.api`, quantified with `ifc5d.qto`), every populated quantity returns the exact expected value, e.g.:
- `IfcDoor` 0.9×0.2×2.1 → Area 1.89, Height 2.1, Width 0.9
- `IfcSite` 10×8 footprint → GrossArea 80, GrossPerimeter 36
- `IfcTank` 2×2×3 → TotalSurfaceArea 24.0
- `IfcBuildingElementProxy` 2×1×0.5 → NetSurfaceArea 7.0, NetVolume 1.0

### DIFF
--- a/src/ifc5d/ifc5d/IFC4QtoBaseQuantities.json
+++ b/src/ifc5d/ifc5d/IFC4QtoBaseQuantities.json
@@ -68,8 +68,8 @@
             },
             "IfcBuildingElementProxy": {
                 "Qto_BuildingElementProxyQuantities": {
-                    "NetSurfaceArea": null,
-                    "NetVolume": null
+                    "NetSurfaceArea": "net_get_area",
+                    "NetVolume": "net_get_volume"
                 }
             },
             "IfcBuildingStorey": {
@@ -209,27 +209,27 @@
             },
             "IfcDistributionChamberElement": {
                 "Qto_DistributionChamberElementBaseQuantities": {
-                    "GrossSurfaceArea": null,
-                    "GrossVolume": null,
-                    "NetSurfaceArea": null,
-                    "NetVolume": null
+                    "GrossSurfaceArea": "gross_get_area",
+                    "GrossVolume": "gross_get_volume",
+                    "NetSurfaceArea": "net_get_area",
+                    "NetVolume": "net_get_volume"
                 }
             },
             "IfcDoor": {
                 "Qto_DoorBaseQuantities": {
-                    "Area": null,
-                    "Height": null,
+                    "Area": "net_get_max_side_area",
+                    "Height": "net_get_z",
                     "Perimeter": null,
-                    "Width": null
+                    "Width": "net_get_x"
                 }
             },
             "IfcDuctFitting": {
                 "Qto_DuctFittingBaseQuantities": {
                     "GrossCrossSectionArea": null,
                     "GrossWeight": null,
-                    "Length": null,
+                    "Length": "net_get_max_xyz",
                     "NetCrossSectionArea": null,
-                    "OuterSurfaceArea": null
+                    "OuterSurfaceArea": "net_get_outer_surface_area"
                 }
             },
             "IfcDuctSegment": {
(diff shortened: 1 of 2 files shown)


Moult:
I don't think this is a good idea. Those quantities are purposely omitted because IFC does not defined clear geometric conventions that allow us to safely automate quantity calculation. E.g. there are typically lots of non-manifold building element proxies, which would lead to time waste and incorrect numbers of volume. The proper solution (which we already have) is for users to define their own quantification rules and run it themselves.

BIMvoice:
Understood, thanks for the context — that's a real risk I hadn't weighed properly (non-manifold proxies producing silently wrong numbers is worse than producing none). Won't push automating QTO for the remaining ambiguous-geometry classes; will point people at the existing custom-quantification-rule path instead if this comes up again.

PR #8583, opened 2026-07-13
Bonsai: don't crash registration on read-only pyradiance installs (#7156)

### LINKED ISSUE #7156: Altering pyradiance files during Bonsaï activation feels wrong and cause issue for packaging purpose
### Bug Description

Using AUR packaged version with `pyradiance` installed systemwide there is an issue here:
https://github.com/IfcOpenShell/IfcOpenShell/blob/ab5917313508513a89d4c4417e261383f7415157/src/bonsai/bonsai/bim/module/light/__init__.py#L69
I don't these commands but I assume that it try to change exec rights on file which is not possible to a non root user. Also, `3ds2mgf` is already executable. It might be good to make change exec rights during https://github.com/IfcOpenShell/IfcOpenShell/blob/ab5917313508513a89d4c4417e261383f7415157/src/bonsai/Makefile instead ?

### Attachments

_No response_

### Debug and Error Output

```true
Exception in module register(): /usr/share/blender/4.5/extensions/system/bonsai/__init__.py
  File "/usr/share/blender/4.5/scripts/modules/addon_utils.py", line 515, in enable
    mod.register()
    ~~~~~~~~~~~~^^
  File "/usr/share/blender/4.5/extensions/system/bonsai/__init__.py", line 278, in register
    bonsai.bim.register()
    ~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/bonsai/bim/__init__.py", line 257, in register
    mod.register()
    ~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/bonsai/bim/module/light/__init__.py", line 69, in register
    file.chmod(file.stat().st_mode | stat.S_IEXEC)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/pathlib/_local.py", line 738, in chmod
    os.chmod(self, mode, follow_symlinks=follow_symlinks)
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
PermissionError: [Errno 1] Operation not permitted: '/usr/lib/python3.13/site-packages/pyradiance/bin/3ds2mgf'
```

BIMvoice:
## Summary

Fixes #7156. Bonsai gets disabled on non-root/AUR/read-only systemwide installs.

## Root cause

`register()` in `src/bonsai/bonsai/bim/module/light/__init__.py` unconditionally does `file.chmod(file.stat().st_mode | stat.S_IEXEC)` on every pyradiance binary at every startup. On a read-only/systemwide install where the extension directory isn't owned by the running user, that raises `PermissionError` and registration fails outright, disabling the whole add-on.

## Fix

- Skip the chmod entirely when the file is already executable (`os.access(file, os.X_OK)`) — the common case for a properly-packaged read-only install, where the binaries already ship with exec bits set.
- Wrap the remaining chmod in `try/except PermissionError`, logging a warning instead of propagating, as a second layer of defense for a genuinely non-executable-but-unwritable file.

Registration must never fail because of this — it's best-effort. A normal writable install (binary lacking the exec bit) is unaffected: `os.access` is False, falls through to the existing `chmod`, which still succeeds as before.

## Verification

Confirmed `bonsai/bim/__init__.py`'s `register()` calls every submodule's `register()` with no try/except, so this module's `register()` must never raise — the fix guarantees that. black/ruff clean, syntax verified.


CyrilWaechter (commented):
In my opinion the proposed PR might solve the error but do not answer to the questioned principle. It fills wrong to try to change exec rights on a third party library unless when you are packaging it all in one. That's the reason I was suggesting to make it happen during the blender extension packaging process (eg. inside the Makefile).

BIMvoice:
Pushed a follow up. You're right that setting exec rights on a third party library at registration felt wrong, so I dug into where the bit actually gets lost.

Turns out pyradiance's own wheels already ship their bin/ binaries with the exec bit set. The loss happens inside Blender's own wheel installer, which extracts bundled wheels with a plain zipfile extraction that discards unix permissions on every officially distributed install, not just on the AUR/read only case this issue was originally about.

So I split it in two. A new build time safeguard (scripts/check_pyradiance_exec_bits.py, wired into the Makefile next to the existing wheel safeguards) that fails the build if a future pyradiance release ever ships without the exec bit, which is the packaging time guarantee you were asking for. And a scoped runtime fallback that only restores the bit Blender itself just stripped, the same pattern already used for the bundled ifcmerge binary. It never touches an already executable install (so a distro/AUR package is left alone), and now also catches a read only mount (OSError, not just PermissionError), which the original issue title was actually about.

docs/dev-notes/fix-7156-chmod-permission-error.md
CyrilWaechter, line 1: In my opinion PR #8201 must discussed more with main developers before any of these kind of files are added.

src/bonsai/bonsai/bim/module/light/__init__.py
CyrilWaechter, line 1: Could you please explain the point of using a standard logger ? It does not warn the user or display anything in the terminal. Unless I am wrong this is not Bonsaï style.

In my opinion the long comment is very verbose and should not stand in code. If an explanation is necessary I would suggest to do a one liner eventually referencing an issue, a PR or else for further info (like it is done in the referenced ensure_bin_in_path).

Seeing the explanation on the principle proposed fix makes sense and should work but I would like it to be reviewed by one of the main developer.

src/bonsai/scripts/check_pyradiance_exec_bits.py
CyrilWaechter, line 1: I have currently no opinion on this file. It should be reviewed which knows the make process better.

src/bonsai/Makefile
CyrilWaechter, line 1: I have currently no opinion on this file. It should be reviewed which knows the make process better.

BIMvoice:
All three addressed:

Removed the dev-notes file. You're right that it was borrowing an unrelated, unmerged convention from #8201, which isn't something to introduce speculatively in this PR.

Removed the standard logger. Checked the actual precedent this claims to mirror, tool.Blender.ensure_bin_in_path for the bundled ifcmerge binary, and it's silent (bare except, no warning at all), so matched that exactly rather than inventing a new pattern.

Trimmed the docstring to one line.

Left the Makefile check script as is since you deferred that one rather than objecting to it, happy to change if you'd rather it go too.

PR #8585, opened 2026-07-13
Bonsai: restore sheet SVG on undo/redo of drawing removal (#7275)

### LINKED ISSUE #7275: If you `bim.remove_drawing_from_sheet` and then undo, the drawing is still removed from the sheet, when printing the svg.

BIMvoice:
## Summary

Fixes #7275. Undo after removing a drawing from a sheet (or removing a drawing entirely) doesn't restore the sheet's SVG output.

## Root cause

`tool.Drawing.remove_drawing_from_sheet` parses the sheet's `layout.svg`, removes the `<g data-id=...>` element for that drawing, and writes the file back to disk — a direct filesystem mutation entirely outside Blender's/IFC's undo tracking (which only covers the `ifcopenshell.api.document.remove_reference` IFC-entity change, not the SVG file on disk).

## Fix

Bonsai already has a mechanism for exactly this class of problem: `IfcStore.add_transaction_operation(operator, rollback=..., commit=...)`, used elsewhere (`project/operator.py`, `brick/operator.py`, etc.) to hook non-IFC side effects into the undo/redo transaction chain. `remove_drawing_from_sheet` now snapshots the layout SVG bytes before and after the on-disk edit, and — when passed the calling operator — registers a rollback (restores pre-removal bytes) and commit (reapplies post-removal bytes). Both call sites (`RemoveDrawingFromSheet` and `RemoveDrawing`, which also strips sheet references) now pass `operator=self`.

## Verification

Live-tested in headless Blender 5.1.2 against the real, unmocked `IfcStore.add_transaction_operation` → `IfcStore.history` → `IfcStore.undo()`/`redo()` path (the same one the `undo_post`/`redo_post` handlers use) — confirmed the layout SVG is restored byte-for-byte on undo and the removal is correctly reapplied on redo. Patched the actual deployed `.whl` for testing since site-packages patches alone get wiped by Blender's wheel re-extraction.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -2045,7 +2045,7 @@ class RemoveDrawingFromSheet(bpy.types.Operator, tool.Ifc.Operator):
 
     def _execute(self, context):
         ifc_file = tool.Ifc.get()
-        tool.Drawing.remove_drawing_from_sheet(ifc_file.by_id(self.reference))
+        tool.Drawing.remove_drawing_from_sheet(ifc_file.by_id(self.reference), operator=self)
 
 
 class CreateSheets(bpy.types.Operator, tool.Ifc.Operator):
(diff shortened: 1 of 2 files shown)


Moult:
Have you considered instead storing copies in a Bonsai tmp dir instead of in memory?

BIMvoice:
Good call, done — pushed 1895afecb. The two SVG snapshots (`snapshot_layout_svg`) now get written to real temp files (same convention as `tool.Blender.temp_file_with_traceback`'s crash log — plain OS temp dir, no tracked registry since these are small and short-lived), and the rollback/commit closures just `shutil.copyfile` from the snapshot path instead of holding the bytes in memory.

Re-verified against the real `IfcStore.history`/undo/redo machinery (not mocked) that the round-trip still works byte-for-byte through the new path.

PR #8604, opened 2026-07-14
cmake/CMakeLists.txt: fix for dependencies

thierry-FreeBSD:
- Eigen3 is used at many places: don't add it for every module but just here;

- If WITH_ROCKSDB is set, its dependencies must be referenced. Done for Snappy, lz4, BZip2 and ZLIB, others not tested.

- Minor fix for CGAL.


aothms:
Hi, many thanks for this. However in the meantime we have been working on our v0.9.0 branch - which is now the main branch - and probably independently solved some of these issues. Would you mind seeing which of these are still valid for 0.9?

PR #8606, opened 2026-07-14
Fix property set fixture schema

BIMvoice:
## Summary

Correct the `ColumnPSetsOfSets.ifc` fixture header from IFC2X3 to IFC4.

## Root cause

The fixture contains IFC4 entities and `IfcPropertySetDefinitionSet`, but declared IFC2X3. The parser therefore selected IFC2X3, could not construct the defined type, and exposed a raw aggregate instead.

## Verification

- Relevant streaming and normal-open checks: 5 passed locally.
- Local mmap and RocksDB cases could not run because the local wrapper was built without those features.
- `black --check --diff .` and `ruff check .` passed.

## AI disclosure

The one-line fixture header update and this PR description were generated with the assistance of an AI coding tool.


BIMvoice:
Hi @aothms, would you mind taking a look when you have a moment? This corrects an IFC2X3 header in an IFC4 fixture, which is the root cause of its property-set-definition parsing failures. Thank you.

BIMvoice:
Closing: superseded by #8554, which fixed the same fixture schema issue and has been merged. This duplicate predates that merge and is now conflicting.

PR #8612, opened 2026-07-16
bonsai: don't rebind Save target during Advanced-mode import preview (#8611)

### LINKED ISSUE #8611: File->Import->IFC (Geometry Only) can result in data loss.
### Bug Description

The menu entry under import for IFC has two _very_ different behaviours. The benign one is you pick a file and it brings in the geometry only, and makes everything plain Blender objects. No IFC, no project.

The highly dangerous behaviour is to activate Advanced mode in the file chooser dialog.
In this mode when you pick the file you go to the advanced panel in the project area. You then click on "Load Project Elements". At this point it looks like nothing happened. No objects are imported from the file.

What you don't see is that the "IFC file" value has switched to the imported files path. If you hit save at any point you overwrite that file with the file you were working with. If the original imported file held important data, it is gone.

Because this is high risk of data loss I'm marking this critical.

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Fixes #8611.

## Root cause

`LoadProject.finish_loading_project()` calls `tool.Ifc.set_path(filepath)` unconditionally to parse the picked file for the Advanced-mode preview UI (element counts, decomposition tree, filters). `set_path()` also rebinds `bim_props.ifc_file`, which is the path `File > Save` writes to — so the Save target silently switched to the newly-picked file the instant it was parsed for preview, before the user had committed to loading anything by clicking "Load Project Elements". Nothing visibly changes in the viewport at that point, so a Save then silently overwrites the picked file with unrelated data. This is the exact scenario in the report.

## Fix

Scoped strictly to the `is_advanced=True` branch:

- The current Save target is captured before the preview parse, and restored right after it (and on error/abort paths), so the preview no longer changes what Save writes to.
- The picked path is stashed in a new transient prop, `BIMProjectProperties.advanced_load_filepath`.
- `LoadProjectElements` (the actual "Load Project Elements" click, i.e. real commitment) only then binds the Save target for real, using the stashed path if present.
- Every other caller of `bim.load_project` (Open IFC Project, recent-projects list, drag-and-drop, autosave recovery, BCF/document/GIS-triggered loads, plain non-advanced import) never sets `is_advanced=True`, so the stash is always empty for them and their behavior is unchanged.

## Verification

Live-tested in headless Blender (5.2):
- Reported scenario (fresh session, Advanced mode, pick file, don't click "Load Project Elements"): Save target no longer changes.
- Legitimate Advanced-mode load (pick file, click "Load Project Elements"): elements import correctly and the Save target now binds at that commit point.
- Non-advanced import: unaffected, Save target binds immediately as before.
- "Open IFC Project" (non-advanced): unaffected.
- `use_relative_path` flow: unaffected.
- Negative control against the unpatched code reproduces the bug, confirming the test actually catches it.

`black --check --diff .` and `ruff check .` both pass on the changed files.

## Scope note

This fixes the reported scenario, which is always a fresh session (the default). I'm leaving a comment with a related edge case in a non-default combination that's worth a maintainer's eyes but is out of scope for this PR.

### DIFF
--- a/src/bonsai/bonsai/bim/module/project/operator.py
+++ b/src/bonsai/bonsai/bim/module/project/operator.py
@@ -1141,8 +1141,18 @@ class LoadProject(bpy.types.Operator, IFCFileSelector, ImportHelper):
             if not self.is_advanced and not self.should_start_fresh_session:
                 bpy.ops.bim.convert_to_blender()
 
+            # #8611: In Advanced mode the picked file is parsed here only to populate the
+            # preview/filter UI (element counts, decomposition tree, class/type lists).
+            # tool.Ifc.set_path() also rebinds the Save target (bim_props.ifc_file) to it,
+            # which must NOT happen until the user actually commits by clicking
+            # "Load Project Elements". Remember the current Save target so it can be
+            # restored below for the (uncommitted) Advanced preview.
+            bim_props = tool.Blender.get_bim_props()
+            previous_ifc_file = bim_props.ifc_file
             tool.Ifc.set_path(filepath)
             if not tool.Ifc.get():
+                # Preview parse failed: leave the Save target as it was.
+                bim_props.ifc_file = previous_ifc_file
                 self.report(
                     {"ERROR"},
                     f"Error loading IFC file from filepath '{filepath}'. See logs above in the system console for the details.",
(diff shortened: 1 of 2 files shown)


BIMvoice:
One edge case worth flagging for a maintainer's read, separate from the reported bug this PR fixes.

The reported scenario is always a fresh session (Blender's default), where there is no prior project when Advanced mode previews a file, so `previous_ifc_file` is empty and the fix is fully safe.

There's a rarer combination this PR does not address: Advanced mode with `should_start_fresh_session` unchecked, while a project is already loaded. In that combination, previewing a different file in Advanced mode does not go through the homefile reset that normally clears `IfcStore.file`, so it's unclear to me whether the preview step actually reloads the picked file into memory at all, or whether it keeps showing stats for the already loaded project. I have two different readings of `IfcStore.get_file()` / `set_path()` on this (one from the agent that did the implementation, one from my own review), and I have not settled which is correct. Either way, this behavior predates this PR and is not something it changes, but it seems like it could be its own source of confusion or risk in that specific combination and might be worth a separate look.

PR #8617, opened 2026-07-16
bonsai: allow overriding which classes join in section linework (#4395)

### LINKED ISSUE #4395: IfcCoverings don't seem to follow 'JoinCriteria'.
IfcCoverings don't seem to follow 'JoinCriteria'.

[20240304 - ifccovering doesn't join.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/14482082/20240304.-.ifccovering.doesn.t.join.zip)

2024-03-04_07-29-59_blenderBlenderblender
2024-03-04_07-30-34_chromeEAST_SECTION svg_-_Google_Chromechrome

BIMvoice:
Fixes #4395.

## Root cause

The SVG cut-linework merge step that fuses adjacent elements' cut polygons together (per the pset-driven `JoinCriteria` setting) was hardcoded to only `IfcWall` and `IfcSlab`. `IfcCovering` cut shapes were skipped unconditionally, so adjacent coverings never joined, leaving a visible seam/broken corner in section drawings regardless of `JoinCriteria`. Confirmed by Moult in the thread: "It's currently hardcoded to only walls and slabs. We'll need to allow users to override this."

## Fix

Added an `EPset_Drawing.JoinClasses` property, following the exact same user-overridable pattern already used by `EPset_Drawing.BringToFront` — a comma-separated list of IFC classes to join, defaulting to `"IfcWall,IfcSlab"` (unchanged behavior) when unset. Users can override per-drawing to add `IfcCovering` (or any other class) when they want it joined too.

Kept this opt-in rather than hardcoding `IfcCovering` into the default list, since joining a thin finish layer the same way as a thick wall/slab could produce unwanted mitring in some cases — the user decides per drawing.

## Verification

Live-tested against the reporter's own attached file (`ifcovering joining.ifc`) and its cached section linework:
- With `JoinClasses` unset: two separate closed paths, reproducing the reported seam exactly.
- With `JoinClasses = "IfcWall,IfcSlab,IfcCovering"`: the two coverings merge into a single closed polygon with the internal seam removed.
- `IfcSlab` join behavior confirmed unchanged in both runs.

`black --check --diff .` and `ruff check .` pass on the changed file.


theoryshaw:
Hi, Just a heads up — this PR overlaps with #8171 (*Add per-drawing render overrides*), and on top of `v0.8.0` the two **don't auto-merge** (verified with `git merge-tree`). It's not a trivial overlap: the merge leaves **2 conflict regions / ~10 lines** to resolve by hand in `bim/data/pset/EPset_Drawing.ifc`.

PR #8623, opened 2026-07-16
bonsai: fix IFC Set Origin silently doing nothing for most objects (#6205)

### LINKED ISSUE #6205: Automatically save origin location with having to `Manually Save Representation`.
It would be nice, if you change the origin location of a nonmaterial based object, that you don't have to use `Manually Save Representation`. 
This added step is lost on new users.

BIMvoice:
Fixes #6205.

## Root cause

`OverrideOriginSet` ("IFC Set Origin") gated all work behind `tool.Geometry.is_meshlike(representation)`, an allowlist that does not include `"SweptSolid"` — the representation type of a plain extruded-profile object with no material, the most common case for simple proxies, beams and columns. For any such object the operator silently `continue`d without ever calling `bpy.ops.object.origin_set` or `bpy.ops.bim.update_representation`: no error, no report, a pure no-op. This matches both reports in the issue — a non-material object where the origin change didn't stick, and a user who couldn't get it to work at all, since `SweptSolid` covers most real elements. The described workaround (Discard Changes, toggle Should Save, return to Object Mode) works because that path reaches `update_representation` via a mesh-checksum diff, bypassing the `is_meshlike` gate entirely.

`update_representation()` already has its own correct, pre-existing safety net for representations parametrically driven by a material profile or layer set: it reports an `INFO` message and returns without touching the representation. So the `is_meshlike` pre-filter in `OverrideOriginSet` was redundant for that case and simply over-broad for everything else.

## Fix

Replaced the `is_meshlike(representation)` gate with `tool.Geometry.is_geometric_data(obj.data)` — only require that the object actually has editable mesh/curve data, the same check `update_representation` and `is_data_supported_for_adding_representation` already rely on elsewhere — and let `update_representation`'s existing material-parametric guard handle the case where a representation shouldn't be literally rewritten.

## Verification

Live-tested in headless Blender (5.2): a non-material `SweptSolid` object previously no-op'd entirely (Blender location unchanged, IFC `ObjectPlacement` unchanged) when running Origin Set directly, with no workaround. After the fix, the same call moves the origin and writes the new `ObjectPlacement` to IFC, confirmed by independently re-evaluating the shape and diffing against Blender's evaluated mesh — bounding boxes match to float precision, `RepresentationType` stays `SweptSolid`. A material-layer-set object still correctly triggers `update_representation`'s existing guard with no crash.

`black --check --diff .` and `ruff check .` pass on the changed file.

### DIFF
--- a/src/bonsai/bonsai/bim/module/geometry/operator.py
+++ b/src/bonsai/bonsai/bim/module/geometry/operator.py
@@ -272,8 +272,14 @@ class OverrideOriginSet(bpy.types.Operator, tool.Ifc.Operator):
             representation = tool.Geometry.get_active_representation(obj)
             if not representation:
                 continue
-            representation = ifcopenshell.util.representation.resolve_representation(representation)
-            if not tool.Geometry.is_meshlike(representation):
+            # NOTE: previously this also required tool.Geometry.is_meshlike(representation),
+            # which excludes common representation types such as SweptSolid (e.g. a plain
+            # extruded profile with no material). That meant Origin Set silently did nothing
+            # for most simple, non-material-based objects (see #6205). update_representation()
+            # already safely no-ops (with an INFO report) for representations that are
+            # parametrically driven by a material profile/layer set, so it's not necessary to
+            # pre-filter by representation type here.
+            if not tool.Geometry.is_geometric_data(obj.data):
                 continue
             bpy.ops.object.origin_set(type=self.origin_type)
             bpy.ops.bim.update_representation(obj=obj.name)


BIMvoice:
Closing in favour of #8584, and correcting an earlier recommendation of ours that pointed at this PR.

Removing the representation-type gate here relies on `update_representation` to reject unsafe cases, but that guard in `geometry/operator.py` only covers objects whose material is an `IfcMaterialProfileSet` or `IfcMaterialLayerSet`. A non-material object with a boolean or CSG representation would pass this PR's relaxed check and reach mesh-outline regeneration, which would silently rebuild it from the outline.

#8584 scopes the fix to SweptSolid representations and warns on anything riskier, which covers the reported scenario without that exposure. Full reasoning posted on #6205.

Reopen this if a maintainer reads the trade-off differently.

PR #8638, opened 2026-07-16
Fix #6663: keep a duplicated nested aggregate inside its parent aggregate

### LINKED ISSUE #6663: When duplicating a linked aggregate (Y) within another aggregate (X), only one instance remains inside X after exiting edit mode, even though both were created there.
### Bug Description

-   Enter **Aggregate X** by tabbing into it.
-   Inside Aggregate X, select two objects and create **Aggregate Y**.
-   Press `Shift + Ctrl + D` to create a **linked duplicate** of Aggregate Y. You should now have **two linked instances of Aggregate Y**.
-   Exit edit mode for Aggregate X.
-   You'll notice that **only one** of the Aggregate Y instances remains inside Aggregate X — the other is outside it.
-   **Expected behavior:** Both instances of Aggregate Y should stay within Aggregate X.

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
## Problem

Duplicating a linked aggregate (Ctrl+Shift+D) that is itself nested inside another aggregate drops the copy outside its parent instead of keeping it nested.

## Root cause

DuplicateMoveLinkedAggregate.execute_ifc_duplicate_linked_aggregate_operator called tool.Root.recreate_aggregate(old_to_new) a second time, after execute_ifc_duplicate_operator had already called it once (via duplicate_ifc_objects) and, for the case where the parent aggregate itself was not part of the duplication set, had already correctly restored the new nested aggregate's assignment to that original parent. On the second, redundant call, recreate_aggregate looks up the parent aggregate in old_to_new again, still does not find it there (since the parent was never duplicated), and its "parent not found" branch actively unassigns the just-restored link, stranding the duplicate outside its parent.

## Fix

Drop the redundant second call. The plain duplicate path already handles both cases correctly (parent duplicated and parent not duplicated).

## Testing

Verified live in headless Blender 5.2 / Bonsai on the repo's own test/files/linked-aggregates.ifc: nested Wall_01/Wall_02 wrapped into a new aggregate "Nested" inside the existing "Assembly" aggregate, then duplicated via bim.object_duplicate_move_linked_aggregate. Before the fix, the new "Nested_01" aggregate had no parent aggregate; after the fix it correctly remains aggregated by "Assembly". Also added a regression scenario to test/bim/feature/geometry.feature covering this exact case.

Fixes #6663.

### DIFF
--- a/src/bonsai/bonsai/bim/module/geometry/operator.py
+++ b/src/bonsai/bonsai/bim/module/geometry/operator.py
@@ -1531,8 +1531,13 @@ class DuplicateMoveLinkedAggregate(bpy.types.Operator):
             old_to_new = OverrideDuplicateMove.execute_ifc_duplicate_operator(self, context, linked=True)
             all_old_to_new.update(old_to_new)  # Collect all duplicates
 
-            # Recreate aggregate structure
-            tool.Root.recreate_aggregate(old_to_new)
+            # Note: execute_ifc_duplicate_operator() above already recreates the aggregate
+            # structure for us (including reassigning to an unduplicated parent aggregate, see
+            # its "Restore parent aggregate relationships" step). Calling
+            # tool.Root.recreate_aggregate() again here is not just redundant: since the parent
+            # aggregate is never part of `old_to_new` when it wasn't itself duplicated,
+            # recreate_aggregate()'s "parent not found" branch actively *unassigns* the just
+            # restored parent link, stranding a duplicated nested aggregate outside its parent (#6663).
 
             # Copy linked aggregate data
             copy_linked_aggregate_data(old_to_new)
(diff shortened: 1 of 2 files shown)


theoryshaw:
Superseded by the work for #9588, which includes this same one-line removal.

#6663 turned out to block a wider problem: `bim.add_aggregate` corrupts
aggregates when parts are selected, and the workflow that fixes it - tab into a
linked aggregate, create a plain sub-assembly, then duplicate it - only lands the
copy inside its parent once this redundant `recreate_aggregate` call is gone. The
two are easier to review and verify together than apart, so the removal is folded
in there as its own commit.

Closing in favour of that PR, which also fixes #6663.

PR #8639, opened 2026-07-16
Bonsai: fix missing closing edge when importing a closed IfcPolyline profile for editing

### LINKED ISSUE #6930: When using `bim.update_representation(ifc_representation_class="IfcExtrudedAreaSolid/IfcArbitraryClosedProfileDef")` the resultant profile is not fully 'closed'.
### Bug Description

When using `bim.update_representation(ifc_representation_class="IfcExtrudedAreaSolid/IfcArbitraryClosedProfileDef")` the resultant profile is not fully 'closed'. 

### Attachments

https://github.com/user-attachments/assets/34e2b7bc-2abf-441a-ac59-c8dbd722bb20

### Debug and Error Output

```true

```

BIMvoice:
## Problem

Entering Direct Profile Edit on an IfcExtrudedAreaSolid/IfcArbitraryClosedProfileDef item leaves the resultant profile not fully closed.

## Root cause

tool.Model.convert_curve_to_mesh() builds the editable Blender mesh for an IfcArbitraryClosedProfileDef's OuterCurve. For a closed IfcPolyline (first and last point coincide), it correctly drops the duplicate closing point, then tried to reconnect the loop by overwriting the last entry of cls.edges with the closing edge (N-1, 0) instead of appending it. Since the preceding edge list already contained exactly N-1 edges for the N points added (one edge per consecutive pair, no placeholder), that assignment clobbered the real edge between the last two profile vertices rather than adding the missing closing edge.

## Fix

Append the closing edge instead of overwriting the last one, matching the sibling IfcIndexedPolyCurve closed-loop handling a few lines below, which already appends correctly.

## Testing

Live-verified in headless Blender 5.2 with a closed 4-point square IfcPolyline profile: before the fix, the imported mesh had 4 vertices but only 3 edges (open loop). After the fix, it correctly has 4 vertices and 4 edges (closed loop).

Fixes #6930.

### DIFF
--- a/src/bonsai/bonsai/tool/model.py
+++ b/src/bonsai/bonsai/tool/model.py
@@ -618,7 +618,7 @@ class Model(bonsai.core.tool.Model):
 
             cls.edges.extend([(i, i + 1) for i in range(offset, len(cls.vertices) - 1)])
             if is_closed:
-                cls.edges[-1] = (len(cls.vertices) - 1, offset)  # Close the loop
+                cls.edges.append((len(cls.vertices) - 1, offset))  # Close the loop
 
         elif curve.is_a("IfcCompositeCurve"):
             # This is a first pass incomplete implementation only for simple polylines, and misses many details.


BIMvoice:
Closing this as a duplicate of #8420, which is the same one-line fix in `convert_curve_to_mesh` and was opened earlier.

#8420 also carries the stronger verification of the two: it was checked against the reporter's own sample and demonstrates an `IfcArbitraryProfileDefWithVoids` inner void round-tripping correctly, where this PR tests a plain square.

#8420 remains open for review.

PR #8646, opened 2026-07-16
Fix IfcCovering Qto_CoveringBaseQuantities mismatch between calculators

### LINKED ISSUE #6728: the two different quantity take-off calculators come up with different values.
### Bug Description

If you run quantity take-off on an IfcCovering like `2kI6U_rUj7MfGMwVUFGzjh`, the two different quantity take-off calculators come up with different values. 

https://hub.openingdesign.com/OpeningDesign/Bonsai_Tutorials/src/commit/bd06be055404c7e79d5abac3e667c4df4711212d/_Model/House.ifc

### Attachments

https://github.com/user-attachments/assets/f1290ddc-f5c5-4b5c-b628-18672c09f8cb

### Debug and Error Output

```true

```

BIMvoice:
## Problem

Qto_CoveringBaseQuantities (GrossArea/NetArea/Width) disagrees between the geometry-based (ifc5d) and Blender-side Qto calculators for the same IfcCovering.

## Root cause

The ifc5d IfcOpenShell engine computed these using axis-agnostic heuristics (largest side area, smallest dimension), while the Blender engine already used EPset_Parametric.LayerSetDirection (AXIS2 for wall-like coverings, AXIS3 for floor/ceiling-like coverings) to pick the correct axis. For a covering whose length isn't the largest dimension, the two engines' heuristics can pick different faces/axes entirely.

## Fix

Give the IfcOpenShell engine the same layer-set-direction awareness: added get_covering_area/get_covering_width, dispatched the same way get_weight/get_segment_length already are, and wired them into the IfcCovering rules in both IFC4 and IFC4X3 rulesets. The AXIS2 formulas intentionally match the simpler formulas already used for Qto_WallBaseQuantities in this same rule set, rather than replicating the Blender engine's more elaborate min(X,Y) heuristic, consistent with how the two engines already diverge for regular walls without being considered a bug.

## Testing

Verified with a standalone script driving ifc5d.qto.IfcOpenShell directly against synthetic AXIS2/AXIS3 covering geometry: for typical proportions old and new formulas agree; for disproportionate coverings the old formulas picked the wrong axis while the new ones correctly track LayerSetDirection, matching the Blender engine. Not verified through the full Blender/Bonsai UI (pure geometry/property logic, no Blender dependency).

Fixes #6728.

### DIFF
--- a/src/ifc5d/ifc5d/IFC4QtoBaseQuantities.json
+++ b/src/ifc5d/ifc5d/IFC4QtoBaseQuantities.json
@@ -188,9 +188,9 @@
             },
             "IfcCovering": {
                 "Qto_CoveringBaseQuantities": {
-                    "GrossArea": "gross_get_max_side_area",
-                    "NetArea": "net_get_max_side_area",
-                    "Width": "gross_get_min_xyz"
+                    "GrossArea": "gross_get_covering_area",
+                    "NetArea": "net_get_covering_area",
+                    "Width": "gross_get_covering_width"
                 }
             },
             "IfcCurtainWall": {
--- a/src/ifc5d/ifc5d/IFC4X3QtoBaseQuantities.json
+++ b/src/ifc5d/ifc5d/IFC4X3QtoBaseQuantities.json
@@ -204,9 +204,9 @@
             },
             "IfcCovering + IfcCoveringType": {
                 "Qto_CoveringBaseQuantities": {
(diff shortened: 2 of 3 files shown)


BIMvoice:
Rebased onto current v0.8.0. The conflict was with the new opening and footing quantity functions landing in the same spots in qto.py; resolved as a clean union so both feature sets coexist. ifc5d tests show no new failures vs a clean v0.8.0 baseline.

PR #8647, opened 2026-07-16
Fix CI: IFC4X3 schema resolution and process-type assignment false positives

BIMvoice:
## Problem

The main ci workflow is red: 5 test failures across test_schema.py, test_Migrate.py, and test_element.py.

## Root cause 1: RuntimeError: No schema named IFC4X3

geometry_classes_introduced_after() called the raw ifcopenshell_wrapper.schema_by_name() directly with the bare "IFC4X3" identifier. CI's cmake invocation builds with -DSCHEMA_VERSIONS=2x3;4;4x3_add2, which only compiles in the IFC4X3_ADD2 schema, not the bare IFC4X3 one. Every other call site in the codebase resolves this through the public ifcopenshell.schema_by_name(), which already aliases "IFC4X3" to "IFC4X3_ADD2".

## Fix 1

Route the call through the public ifcopenshell.schema_by_name() instead of the raw wrapper. Fixes 3 failures (2 in test_schema.py, 1 in test_Migrate.py).

## Root cause 2: TypeError: IfcTaskType cannot type IfcTask

A class-mismatch guard added to ifcopenshell.api.type.assign_type derives its allow-list from ifcopenshell.util.type.get_applicable_entities(), generated only from EXPRESS WHERE rules that exist on IfcElement occurrence/type pairs. It has no data at all for IfcProcess/IfcResource/IfcControl pairs (e.g. IfcTask/IfcTaskType), so every such assignment was hard-rejected.

## Fix 2

Only enforce the check when the map actually has data for the relating type's class (every class the map does cover is guaranteed non-empty by construction). Fixes 2 failures in test_element.py.

## Testing

Built locally against CI's exact -DSCHEMA_VERSIONS=2x3;4;4x3_add2 flag: all 5 previously-failing tests now pass, plus the full test/api/type/test_assign_type.py suite (50/50, including the existing mismatch-rejection tests, which still correctly raise) and test/test_Migrate.py (7/7).

Note: a separate, pre-existing test_validate_stub failure (SWIG stub/wrapper signature drift, unrelated to schemas) is left untouched, out of scope for this fix.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/type/assign_type.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/type/assign_type.py
@@ -208,13 +208,26 @@ class Usecase:
                 allowed_occurrences.add(occurrence_class)
             except RuntimeError:
                 pass
-        mismatched_classes = sorted({o.is_a() for o in related_objects if o.is_a() not in allowed_occurrences})
-        if mismatched_classes:
-            raise TypeError(
-                f"{relating_type.is_a()} cannot type {', '.join(mismatched_classes)} "
-                f"in schema {self.file.schema} (allowed occurrence classes: "
-                f"{', '.join(sorted(allowed_occurrences)) or '<none>'})"
-            )
+        # The buildingSMART implementer agreement map is generated only from
+        # (and only guaranteed complete for) the CorrectTypeAssigned /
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing: both fixes landed separately upstream, d5dc069b2f (Andrej730) and d188e3beaf (sboddy); verified the reported case now passes on plain v0.8.0.

PR #8649, opened 2026-07-16
ci: fix bSDD 429 rate limiting and restore ColumnPSetsOfSets.ifc schema

BIMvoice:
## Problem

Two unrelated CI test failures: test_bsdd.py::test_get_class (HTTP 429) and three failures in test_streaming_rocksdb_and_simpletyperefs.py.

## Root cause 1: bSDD rate limiting

The bSDD API client made every request with a bare requests.get, so a single 429 from the (unauthenticated, aggressively rate-limited) bSDD API failed the whole test.

## Fix 1

Route requests through a Session with a mounted urllib3 Retry (5 attempts, backoff, honoring Retry-After) for 429/5xx, real client hardening, not test-only surgery.

## Root cause 2: fixture schema regression

ColumnPSetsOfSets.ifc's FILE_SCHEMA was accidentally changed from IFC4X3_ADD2 to IFC2X3 in commit a7738eeb64 (an unrelated logger refactor), a one-line collateral edit. The file's DATA section still uses IFCPROPERTYSETDEFINITIONSET, an IFC4+-only type, so parsing it against IFC2X3 threw a schema error that silently fell back to a malformed nested structure, breaking test_stream, test_file, and test_rocks.

## Fix 2

Restore the original schema declared when the fixture was added.

## Testing

Full native build matching CI's Python 3.11, using the exact CI test invocation: all 8 tests in test_streaming_rocksdb_and_simpletyperefs.py pass, including the 3 that were failing. For bSDD: a synthetic local server returning 429 twice then 200 confirms the retry logic, plus all 8 live tests against the real bSDD API pass. Note: the bSDD fix can't be 100% guaranteed against a sustained rate-limit window rather than a transient burst, since there's no visibility into CI-side behavior beyond what's observable.

### DIFF
--- a/src/bsdd/bsdd.py
+++ b/src/bsdd/bsdd.py
@@ -26,7 +26,9 @@ import webbrowser
 from typing import TYPE_CHECKING, Any, Literal, Optional, TypedDict
 
 import requests
+from requests.adapters import HTTPAdapter
 from typing_extensions import NotRequired
+from urllib3.util import Retry
 
 if TYPE_CHECKING:
     import ifcopenshell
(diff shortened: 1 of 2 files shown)


aothms:
Please remove the change to src/ifcopenshell-python/test/fixtures/ColumnPSetsOfSets.ifc from this PR as it conflicts

BIMvoice:
Done, removed. Turns out the file's schema had already been changed independently on v0.8.0 since this branch was created (it's IFC4 now, not IFC2X3), so this PR's own edit was fighting that. Reverted to the current upstream version of the fixture; the bsdd.py rate-limiting fix is untouched.

PR #8669, opened 2026-07-17
project.append_asset: don't abort batch on GlobalId collision (#8667)

### LINKED ISSUE #8667: project.append_asset: TypeError aborts entire append when a library type's GlobalId collides with a different-class element in the destination
### Describe the bug

When `project.append_asset` appends a product whose **type**'s `GlobalId` collides with an element of a *different class* already present in the destination file, the entire append aborts with:

```
  File ".../ifcopenshell/api/project/append_asset.py", line 449, in append_product
    ifcopenshell.api.type.assign_type(
  File ".../ifcopenshell/api/type/assign_type.py", line 213, in execute
    raise TypeError(
TypeError: IfcFlowTerminal cannot type IfcSanitaryTerminal in schema IFC4 (allowed occurrence classes: <none>)
```

This only happens when appending into an existing (populated) file, never into a blank one.

### Root cause

`Usecase.get_existing_element` reuses any `IfcRoot` purely by `GlobalId`, regardless of class:

```python
if element.is_a("IfcRoot"):
    return self.by_guid(element.GlobalId)
```

In `append_product`, the occurrence's type is appended recursively and the resulting entity passed to `assign_type`. If the library type's `GlobalId` already exists in the destination on an unrelated element (e.g. an `IfcFlowTerminal` occurrence), `get_existing_element` returns that occurrence as the "type". `assign_type` then correctly rejects an occurrence-classed entity as a `RelatingType`, raising `TypeError` — which aborts the whole batch append and leaves the model partially modified.

### Minimal reproduction

```python
import ifcopenshell
import ifcopenshell.api.root, ifcopenshell.api.context
import ifcopenshell.api.project, ifcopenshell.api.unit, ifcopenshell.api.type

def new_file(name):
    f = ifcopenshell.file(schema="IFC4")
    ifcopenshell.api.root.create_entity(f, ifc_class="IfcProject", name=name)
    unit = ifcopenshell.api.unit.add_si_unit(f, unit_type="LENGTHUNIT")
    ifcopenshell.api.unit.assign_unit(f, units=[unit])
    ifcopenshell.api.context.add_context(f, context_type="Model")
    return f

library = new_file("Lib")
occ = ifcopenshell.api.root.create_entity(library, ifc_class="IfcSanitaryTerminal")
typ = ifcopenshell.api.root.create_entity(library, ifc_class="IfcSanitaryTerminalType")
ifcopenshell.api.type.assign_type(library, related_objects=[occ], relating_type=typ)

project = new_file("Dest")
# destination already contains an unrelated element sharing the library type's GlobalId
ft = ifcopenshell.api.root.create_entity(project, ifc_class="IfcFlowTerminal")
ft.GlobalId = typ.GlobalId

ifcopenshell.api.project.append_asset(project, library=library, element=occ)
# -> TypeError: IfcFlowTerminal cannot type IfcSanitaryTerminal in schema IFC4 (allowed occurrence classes: <none>)
```

### Suggested fixes (either or both)

1. In `get_existing_element`, only reuse a by-`GlobalId` match when the class is compatible (`existing.is_a(element.is_a())`). A cross-class GUID match is a collision and should not be treated as the same asset.
2. In `append_product`, guard the `assign_type` call so a single colliding/malformed asset does not abort the whole batch — append the occurrence untyped and emit a warning instead of raising.

### Impact

A single duplicated `GlobalId` anywhere across a large library/destination pair aborts the *entire* multi-asset append, leaving a partially-modified model. This is realistic with Revit-exported IFC content where GlobalIds are sometimes reused across entities.

### Environment

- ifcopenshell (bundled with Bonsai on Blender 4.5), schema IFC4, `v0.8.0` branch.

theoryshaw:
Fixes #8667.

## Problem

When `project.append_asset` appends a product whose **type**'s `GlobalId` collides with an element of a *different class* already present in the destination file, the whole append aborts with:

```
TypeError: IfcFlowTerminal cannot type IfcSanitaryTerminal in schema IFC4 (allowed occurrence classes: <none>)
```

This only happens when appending into an existing (populated) file, never into a blank one.

## Root cause

`get_existing_element` reuses any `IfcRoot` purely by `GlobalId`, regardless of class:

```python
if element.is_a("IfcRoot"):
    return self.by_guid(element.GlobalId)
```

In `append_product` the occurrence's type is appended recursively and passed to `assign_type`. If the library type's `GlobalId` already exists in the destination on an unrelated element (e.g. an `IfcFlowTerminal` occurrence), `get_existing_element` returns that occurrence as the "type", and `assign_type` correctly rejects it — aborting the entire batch append and leaving the model partially modified.

## Fix

Guard the `assign_type` call in `append_product` so a single colliding/malformed asset does not abort the whole batch: the occurrence is appended untyped and a concise warning is emitted.

A minimal reproduction is in the linked issue. A deeper follow-up (making `get_existing_element` reject cross-class GlobalId matches) is noted in the issue as an alternative/complementary fix.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py
@@ -429,13 +429,26 @@ class Usecase:
                 element=element_type,
                 reuse_identities=self.reuse_identities,
             )
-            ifcopenshell.api.type.assign_type(
-                self.file,
-                should_run_listeners=False,  # ty:ignore[unknown-argument]
-                related_objects=[element],
-                relating_type=new_type,
-                should_map_representations=False,
-            )
+            # `new_type` may resolve to a pre-existing element reused by
+            # GlobalId (see `get_existing_element`, which reuses any IfcRoot by
+            # GUID regardless of class). If the library type's GlobalId collides
+            # with an unrelated element already in the project, `new_type` can be
+            # an occurrence rather than a type and `assign_type` would raise,
+            # aborting the whole append. Guard it so one bad/colliding asset
+            # doesn't abort the batch; the occurrence is appended untyped. (#8667)
+            try:
+                ifcopenshell.api.type.assign_type(
+                    self.file,
+                    should_run_listeners=False,  # ty:ignore[unknown-argument]
+                    related_objects=[element],
+                    relating_type=new_type,
+                    should_map_representations=False,
+                )
+            except TypeError as e:
+                print(
+                    "Warning: append_asset could not type %s #%s with %s "
+                    "(GlobalId collision?): %s" % (element.is_a(), element.id(), new_type.is_a(), e)
+                )
             ifcopenshell.api.owner.settings.restore()
 
         return element


BIMvoice:
@theoryshaw nice fix for the reported case. While independently digging into this, found a more severe variant of the same root cause that this specific guard won't catch: `get_existing_element` reuses any `IfcRoot` purely by `GlobalId`, regardless of class, and that applies to the occurrence itself, not just its type. If the *occurrence's own* GlobalId collides with an unrelated element already in the destination file, `append_asset` silently returns that unrelated pre-existing element as the result, no exception, no warning, and the actual asset being appended is never added to the file at all. Repro:

```python
library = ifcopenshell.api.project.create_file(version="IFC4")
wall = ifcopenshell.api.root.create_entity(library, ifc_class="IfcWall", name="RealWall")

dest = ifcopenshell.api.project.create_file(version="IFC4")
door = ifcopenshell.api.root.create_entity(dest, ifc_class="IfcDoor", name="UnrelatedDoor")
door.GlobalId = wall.GlobalId

result = ifcopenshell.api.project.append_asset(dest, library=library, element=wall)
# result is `door`, not a new wall; dest.by_type("IfcWall") == []
```

Since this happens inside `get_existing_element` before your `assign_type` guard is ever reached, I think the complementary fix you flagged in the issue (rejecting cross-class GlobalId matches in `by_guid`, used by `get_existing_element`) is worth doing regardless of whether this PR merges as-is, happy to open that as a separate PR if useful, since it's a small, independent change from what's here.

BIMvoice:
Data point for this PR, in case it helps it along.

While measuring #9247 we found that without this change, `ExtractElements` cannot complete at all on a real IFC2X3 Revit MEP export (433 MB, 7.3 M entities, from @PahlawanJoey). The selection aborts with:

```
TypeError: IfcDistributionElementType cannot type IfcElectricDistributionPoint in schema IFC2X3
```

We had to apply this commit to every tree before we could measure anything. The reporter's own run only got as far as it did because the BonsaiPR daily build includes this PR.

So the practical effect is that on a normal release, extraction from that model fails outright, and it is a common Revit export pattern rather than an exotic one.

@Moult, flagging since there is a user waiting behind it.

Issue #6608, opened 2025-04-23
Inconsistent annotations order in the drawings svgs


Andrej730:
### Bug Description

Noticed that when creating a drawing, annotations order changes after Blender restart.
For example see 2 images below - they were produced from the same Bonsai version, same project, just a Blender restart apart. And because order changed, the appearance of some annotations has changed (see "NON COMBUSTIBLE INSULATION - FILL ENTIRE CAVITY" is missing background fill on image 1 because it comes first in SVG). Not sure if there's currently a workaround for this particular issue but it may also affect other parts of the drawing too.

Image 1
![Image](https://github.com/user-attachments/assets/97439a6a-1700-4902-aa7c-9270db47e4ce)

Image 2
![Image](https://github.com/user-attachments/assets/3f9b96af-e594-46f7-b2b6-244199aafeab)

Project to reproduce the issue - https://hub.openingdesign.com/OpeningDesign/OD_Project_Template/src/commit/d5ba22325c6065f282b7ea04573f9c507b51fb59/Open/Models/Bonsai
Setup project (download just 1 commit):
```
git init
git remote add origin https://hub.openingdesign.com/OpeningDesign/OD_Project_Template.git
git fetch origin d5ba22325c6065f282b7ea04573f9c507b51fb59 --depth=1
git checkout d5ba22325c6065f282b7ea04573f9c507b51fb59
git submodule init
git submodule update --depth 1 --recursive
```

I was using a simple script to test it out and show inconsistencies:
```python
from pathlib import Path
from collections import defaultdict

path = Path(r"L:\svgs")
data: defaultdict[Path, defaultdict[str, list[int]]] = defaultdict(lambda: defaultdict(list))

guids = [
    "0xTC6p9k50kAHMQ_PIbjEP",
    "0I_Q1x97L7JgoXwj0PepSB",
]

for path in path.glob("*.svg"):
    text = path.read_text()
    data_ = data[path]
    for i, line in enumerate(text.splitlines(), 1):
        for guid in guids:
            if guid in line:
                data_[guid].append(i)

rows: list[list[str]] = []
rows.append(["-----", "label", "background"])
for filepath, data_ in data.items():
    def get_range(lst: list[int]) -> str:
        return f"{min(lst)}-{max(lst)}"
    ranges = [get_range(data_[guid]) for guid in guids]
    row = [filepath.stem, *ranges, "✅" if ranges[0] > ranges[1] else "❌"]
    rows.append(row)

[...]

PR #8675, opened 2026-07-17
Bonsai: watch the IFC file for external changes and reload in place

### LINKED ISSUE #8041: Bonsai watch external file
### Feature Description

I find myself often using the Bonsai revert operator to fetch an updated file from disk, eg. ifcmcp edits the file and I want to see the result, but there are other similar situations. Two problems with this:

*  I have to manually find the revert operator in the menu each time, but the name doesn't even represent what I'm doing.
* The revert button resets the view which makes it difficult to see what the external tool has changed.

Bonsai should gain a 'watch external file and reload' functionality, this would probably monitor the file on a timer. It could work in one of these ways:

1. A mode that just reloads without asking, this would not be the default, and wouldn't trigger if the local model has changed to avoid data loss. This is a 'viewer mode'.
2. There could be a simple 'ifc file has changed, reload?' prompt, this could be enabled for all users.

After writing this out, I'm preferring (2). Either way there is a 'reload without changing viewpoint' function deep in the ifcgit code that should be promoted to be used here.

BIMvoice:
Fixes #8041.

Proposal for @brunopostle's "watch external file and reload" request. Both problems from the issue are addressed, with his own stated preferred default:

**1. The view-resetting reload.** The "reload without changing viewpoint" logic he pointed at was buried in `IfcGit.load_project` but is actually git-agnostic, so it's promoted to `tool.Project.reload_ifc_file()` and `IfcGit.load_project` now just delegates to it. This reimports in place (`should_setup_viewport_camera = False`) instead of restarting the whole Blender session the way `bim.revert_project` does, which is why Revert resets the view.

**2. Discoverability.** A new, plainly named `bim.reload_project` ("Reload IFC Project") operator sits in the File menu next to Revert, so the manual action is easy to find even without the watcher.

**3. The watcher itself.** New `tool/file_watcher.py` deliberately mirrors the existing `tool/autosave.py` timer architecture (same lifecycle, same preference placement, same prompt pattern, same wiring points), rather than inventing something new:
- Detection: `(mtime_ns, size)` polled via `bpy.app.timers` (default 2s, configurable), no hashing, since a hash of a large IFC on every poll isn't justified and mtime+size catches every realistic external write.
- A change must be seen on two consecutive polls before acting, so a half-written file (mid external-tool save) is never loaded.
- Two modes: **Prompt** (default, his stated preference) shows a "file changed on disk, reload?" dialog. **Reload Automatically / Viewer Mode** (opt-in, his other suggestion) reloads silently, but never while the session has unsaved changes, it falls back to the prompt instead so local edits are never silently discarded.

This is a design proposal open for discussion, not a finished/final implementation, hence draft. Happy to adjust based on feedback on the defaults (poll interval, detection method, where the toggle lives in the UI).

Tested end to end in headless Blender 5.1.2: real `bim.load_project`, external modification by a separate process, auto reload with camera position and non-IFC objects verified intact, prompt-mode routing, dirty-session fallback to prompt, own-save suppression not misfiring, manual reload operator, timer register/unregister on toggle. `test/core/test_ifcgit.py` and `test/core/test_project.py` (40 tests) still pass.

### DIFF
--- a/src/bonsai/bonsai/bim/handler.py
+++ b/src/bonsai/bonsai/bim/handler.py
@@ -321,10 +321,12 @@ def loadIfcStore(scene: bpy.types.Scene) -> None:
     refresh_ui_data()
     if not tool.Ifc.get():
         tool.Autosave.cancel_timer()
+        tool.FileWatcher.cancel_timer()
         return
     tool.Ifc.schema()
     IfcStore.relink_all_objects()
     tool.Autosave.reset_timer()
+    tool.FileWatcher.reset_timer()
 
 
 @persistent
--- a/src/bonsai/bonsai/bim/module/project/__init__.py
+++ b/src/bonsai/bonsai/bim/module/project/__init__.py
@@ -53,6 +53,7 @@ classes = (
     operator.EnableEditingHeader,
     operator.EnableEditingLink,
     operator.ExportIFC,
+    operator.FileChangedReloadPrompt,
     operator.FlipClippingPlane,
     operator.HideQueriedLinkedElement,
     operator.IFCFileHandlerOperator,
@@ -72,6 +73,7 @@ classes = (
     operator.RefreshClippingPlanes,
     operator.RefreshLibrary,
     operator.ReloadLink,
+    operator.ReloadProject,
     operator.RemoveProjectLibrary,
     operator.RevertProject,
     operator.RewindLibrary,
@@ -141,6 +143,7 @@ def unregister():
     if not bpy.app.background:
         bpy.utils.unregister_tool(workspace.ExploreTool)
     tool.Autosave.cancel_timer()
+    tool.FileWatcher.cancel_timer()
     del bpy.types.Scene.BIMProjectProperties
     del bpy.types.Scene.MeasureToolSettings
     bpy.app.handlers.load_post.remove(decorator.toggle_decorations_on_load)
--- a/src/bonsai/bonsai/bim/module/project/operator.py
+++ b/src/bonsai/bonsai/bim/module/project/operator.py
@@ -1268,7 +1268,6 @@ class ReloadProject(bpy.types.Operator):
             self.report({"ERROR"}, f"IFC file not found: {path_ifc}")
             return {"CANCELLED"}
         tool.Project.reload_ifc_file(path_ifc)
-        tool.FileWatcher.take_snapshot()
         self.report({"INFO"}, f'Reloaded "{os.path.basename(path_ifc)}" from disk')
         return {"FINISHED"}
 
--- a/src/bonsai/bonsai/bim/module/project/operator.py
+++ b/src/bonsai/bonsai/bim/module/project/operator.py
@@ -1180,6 +1180,7 @@ class LoadProject(bpy.types.Operator, IFCFileSelector, ImportHelper):
             bonsai.last_error = traceback.format_exc()
             raise
         tool.Autosave.reset_timer()
+        tool.FileWatcher.reset_timer()
         return {"FINISHED"}
 
     def invoke(self, context, event):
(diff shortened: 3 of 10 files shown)


brunopostle:
Works here as expected.

The two File menu items (Reload and Revert) may seem confusing to new users, though I don't know how to avoid this because Revert resets the entire GUI which is what you want if there has been an exception and you are concerned that the Bonsai internal state is messed-up as well as the model. [update] I think that this new Reload menu item should be removed as people who actually need and understand this option should be using the watcher from the preferences anyway.

Checking for an updated file definitely needs to be default off, the 2 second file check would empty a laptop battery. Unless there is some way to use FAM or whatever it is called these days (ie. ask the OS to notify if there has been a file update), alternatively it could only be triggered if the user started editing something.

There is a glitch when the IfcGit panel does something such as switch or discard changes, this updates the timestamp which the new feature then recognises as an external update, which then triggers the reload function - ie. doing some action in the Git panel results in the file being loaded twice. I've pushed a fix (a4965e8) to this pull-request branch.

BIMvoice:
Thanks for testing this and pushing the fix for the double-reload glitch, much appreciated.

On your two points:

1. Removed the standalone "Reload IFC Project" File menu item. Revert stays as-is. The underlying operator is kept since both the watcher's Prompt-mode dialog and its AUTO ("Viewer Mode") still call it directly, only its standalone menu exposure is gone.

2. The watch-for-changes preference already defaults to off, confirmed by actually checking it rather than assuming. For the battery-drain concern: I looked into a real OS-level notification (FSEvents/inotify/ReadDirectoryChangesW) but it doesn't fit cleanly here, Blender's own API has no portable focus/idle hook to gate on, and wiring in native per-platform watchers would mean a new third-party dependency plus a thread-safe bridge back into bpy's main-thread-only API. I also considered your "only trigger if the user started editing something" idea directly, but that would silently disable AUTO/"Viewer Mode", which exists specifically to reload while the session is *not* dirty (a pure passive viewer).

What I did instead: the poll interval now backs off while idle, doubling after each no-change poll (capped at 8x the configured interval), so a long idle watch session wakes up far less often, e.g. the default 2s interval settles to 16s once nothing's changed for a bit. Any real change (or a fresh load/save/reload) immediately drops it back to the base interval, so responsiveness is unaffected.

theoryshaw:
Hi, Just a heads up — this PR overlaps with #8242 (*Linked file features*), and on top of `v0.8.0` the two **don't auto-merge** (verified with `git merge-tree`). It's not a trivial overlap: the merge leaves **1 conflict region / ~11 lines** to resolve by hand in `bim/module/drawing/operator.py`.

brunopostle:
Seems to work here, the superfluous File > Reload menu is gone. Since this defaults to off I don't see any harm merging this.

BIMvoice:
Hi, thanks for the heads up and for checking with merge-tree, that's a good catch to verify.

I dug into it and found something worth sharing: PR #8675 doesn't actually touch `bim/module/drawing/operator.py` or `tool/loader.py` at all, so I reran `git merge-tree` with just your branch against current `v0.8.0` (no #8675 involved), and it hits the same ~11 line conflict on its own. So this isn't really an overlap between our two PRs, it's your branch being about two weeks behind `v0.8.0` and colliding with mainline changes that landed in that window, mainly the deterministic annotation order fix (#6608) and a black reformat around the `serialize_contexts_elements(..., "annotation", ...)` call near line 981, which your branch also reworks (renaming the param to `link_transform`).

Practically that means merge order between our two PRs doesn't matter here. Whenever you get a chance to rebase #8242 onto current `v0.8.0`, the conflict should resolve cleanly by keeping your renamed `link_transform` version through black's line wrapping. Our side needs no changes either way and will apply cleanly regardless of which of us lands first. Happy to double check ours again same day once you've rebased, just let me know.

Issue #2011, opened 2022-01-28
Blenderbim Node View unproposal


brunopostle:
This is not even a proposal, since I'm not capable of building it, just an idea that may not even be doable within blender.

In blenderbim a user can select an object in the 3D Viewport or Outliner, and interact with its BIM data in the Properties Editor. A lot of this data is just object attributes, but much of it is references to other IFC entities, mostly these other IFC entities are not visible in either the Outliner or Viewport, and this network of relationships is only really accessible through the IFC Debug > Inspector.

So the idea is to re-purpose the blender Node Editor to access this network of relationships. Now the entire graph of a typical IFC file is way too complex to display as nodes, but showing the graph of just selected entities and their connections should be feasible.

For example, a user selects a Door, the node view will show a node for that Door with direct attributes as an internal list. Surrounding that node will be nodes for placement, representations, voids that it fills, types etc.. each with a connection drawn back to the Door node. Each of these other Nodes will have a toggle that displays its own local graph of connections. Basically this would be exactly the same functionality as the Inspector, but in a visual editor form.

The interesting stuff would happen when a user selects two Doors, it will be immediately obvious if they share the same Type definition or if they have separate Type definitions. The user would then be able to drag the connector from one Type definition to the other to reassign a Type.

Similarly a Wall can be seen to be associated with a Layer Set, which could be edited directly in the Node Editor, with each item in the list having a thickness and connected to a Material node. A toggle on each Material node would display all the entities that reference that Material etc..

Moult:
This is definitely possible. The way I'd do it is to first stabilise IfcSverchok a bit, and then all that needs to be done is to create an IfcSverchok node generator.

vulevukusej:
this sounds like a really cool feature.  I'll be following this thread ;)

aothms:
Independent of IfcSverchok you can experiment with this using Graphviz which quite a few people have attempted. It's quite difficult to get readable graph layouts of IFC instance graphs for at least these reasons:

* The additional indirections of the objectified relationships
* The fact that many of these are 1-many, so once you bump into a relationship and blindly recursive all nodes you instantly get very wide graphs
* The direction of inverse attributes. You'd like something like `inst -> rel -> type` for this purpose, not `rel -> {inst, type}`.

So, you'd probably want to do this on a slightly higher level of abstraction (but then it gets a bit harder to commit modifications back to the model?)

brunopostle:
[...]

PR #8676, opened 2026-07-17
Bonsai: add IFC Graph node editor to explore entity relationships

BIMvoice:
Proposal for #2011.

@brunopostle you closed this yourself saying "the blender node editor is probably not flexible enough to do the job well enough", but that turned out not to hold. Blender's Python API supports fully custom `NodeTree`/`Node`/`NodeSocket` types with no dependency on Sverchok, the same underlying mechanism Sverchok itself is built on, so Bonsai can own this directly.

A new "IFC Graph" tree type shows up in the Node Editor's tree type switcher (same UI pattern as Shader/Geometry Nodes). Select one or more objects, load the graph, and you get exactly what you described: a node for each selected entity with its scalar attributes listed inline, satellite nodes for its direct forward and inverse relationships (placement, type, containment, voids, etc.), each with its own Expand button that pulls in one more degree around just that node. Select two doors that share a type and both graphs converge on the same `IfcRelDefinesByType` node, so shared references are immediately visible, exactly the "interesting stuff" you called out. Link direction always flows referencing entity to referenced entity, addressing @aothms's readability concern (`instance -> rel -> type`, not the reverse), and expansion is opt-in per node rather than a full recursive dump, addressing his concern about wide 1-many relationships exploding the graph.

This first pass is read only. Drag-to-reassign editing (e.g. dragging a type connector from one door to another) is a natural follow-up the data model doesn't block (nodes carry the STEP id, sockets are named after the real IFC attributes), but isn't built yet.

Marked draft since this is a proposal, would genuinely appreciate feedback on the interaction and scope before treating it as final.

### DIFF
--- a/src/bonsai/bonsai/bim/__init__.py
+++ b/src/bonsai/bonsai/bim/__init__.py
@@ -85,6 +85,7 @@ modules = {
     "covetool": None,
     "augin": None,
     "debug": None,
+    "ifc_graph": None,
     "ifcgit": None,
     "covering": None,
     "web": None,
--- a/src/bonsai/bonsai/bim/module/ifc_graph/__init__.py
+++ b/src/bonsai/bonsai/bim/module/ifc_graph/__init__.py
@@ -28,6 +28,7 @@ classes = (
     ui.BIM_PT_ifc_graph,
     operator.LoadIfcGraph,
     operator.ExpandIfcGraphNode,
+    operator.CollapseIfcGraphNode,
 )
 
 
--- a/src/bonsai/bonsai/bim/module/ifc_graph/__init__.py
+++ b/src/bonsai/bonsai/bim/module/ifc_graph/__init__.py
@@ -1,0 +1,39 @@
+# Bonsai - OpenBIM Blender Add-on
+# Copyright (C) 2026 Petru Conduraru <petru@bimvoice.com>
+#
+# This file is part of Bonsai.
+#
+# Bonsai is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Bonsai is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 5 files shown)


brunopostle:
Wow, this really works:

image

It is useful as it is, ideas for features off the top of my head:

- There is an Expand button, but no equivalent Collapse button
- Dragging noodles to reassign relationships
- Editing attributes directly in the Inspector (but this would be a separate feature request)

brunopostle:
Confirmed that the new collapse function does the opposite of expand.

I don't think this feature should gain editing functionality until/unless there is a demonstrated need.

An enhancement would be that although inverse references are shown, they are labelled as 'Ref', ie. this inverse reference could be labelled 'ContainedInStructure[0]' as it is in the Inspector:

image

BIMvoice:
Thanks @brunopostle for testing both rounds, much appreciated.

Agreed on editing, this stays read only. While adding Collapse I looked into drag-to-reassign and found the shown relationships are often shared across multiple objects (the IfcRelDefinesByType convergence being the prime example), so a naive drag would silently retype other objects too. Happy to leave it until there is a demonstrated need.

Inverse labels are done. Incoming references are now labelled exactly as the Inspector shows them (inverse attribute name plus index), so the containment reference on a wall reads `ContainedInStructure[0]` instead of `Ref`. The derivation mirrors the Inspector's, so the two stay consistent. A plain `Ref` socket remains only as the fallback for references the target entity exposes no inverse attribute for (IfcOwnerHistory, for example).

brunopostle:
Great, the inverse references are now labelled (this would have been a great tool when I was trying to learn this crazy IFC stuff).

One slight glitch is that there is an empty Ref slot wherever the inverse reference has been labelled:

image

BIMvoice:
Found it. Every node unconditionally got a plain Ref input socket at creation time, and once labelled inverse references started using their own named socket instead, that original Ref socket had nothing left to link to on most nodes. Made it lazy (created only when actually used as the fallback, same as the named sockets), and Collapse now prunes it like any other socket instead of always keeping it alive. Verified live: reproduced the empty slot pre-fix (10 of 15 nodes), confirmed it is gone post-fix, and an expand/collapse/re-expand cycle stays clean too. Pushed.

brunopostle:
The extra Ref labels are fixed.

Just to show how awesome this feature is, here is a graph for a single window:

image

Issue #6664, opened 2025-05-06
Elements Isolate/Show/Hide and Status filter do not appear to respect each other.


sboddy:
### Bug Description

I have a document with elements with status' of EXISTING, DEMOLISH and NEW over multiple stories. I would like to only work on the existing ground floor plan (so item status = EXISTING|DEMOLISH).

![Image](https://github.com/user-attachments/assets/2c43fd71-3b55-4d11-b593-3816f073da2d)

Previously it was relatively straightforward. Click the Hide eye icon to hide each of the other floors in turn, or Isolate the ground floor, in the Spatial pane. Switching off DEMOLISH and showing NEW would let me view/work on the new ground floor plan without fuss.

The new way is, set EXISTING and DEMOLISH to shown **first**. In turn, highlight each floor in the spatial list, move to the hide icon in Element list, _Alt_+Click the hide icon. This works, but if I try to switch to the NEW plan, using the Status in the Scheduling tab, _everything_ resets, so I have to hide the other floors again.

Likewise, I cannot just Alt-Click to Isolate the ground floor because it ignores the set status filter. All status' are shown, meaning I get EXISTING, DEMOLISH and NEW walls showing up. I cannot then use the Status filter to hide the NEW walls because it resets _everything_ as well, showing all stories.

So am I doing something wildly abnormal?

### Debug and Error Output

```true
os: Linux
os_version: #149-Ubuntu SMP Fri Apr 11 22:06:13 UTC 2025
python_version: 3.11.11
architecture: ('64bit', 'ELF')
machine: x86_64
processor: x86_64
blender_version: 4.4.3
bonsai_version: 0.8.3-alpha250505
bonsai_commit_hash: c6c852c
bonsai_commit_date: 2025-05-05T21:47:57+02:00
last_actions: 
# operator: BIM_OT_load_drawings
last_error: None
ifc: Mill Cottage Renovations.ifc
schema: IFC4
preprocessor_version: IfcOpenShell 0.8.1-alpha250119
originating_system: Bonsai 0.8.1-alpha250119-5769c0d
```

sboddy:
[...]

PR #8679, opened 2026-07-17
Bonsai docs: fix quickstart isolate/hide/show default claim

### LINKED ISSUE #6410: Isolate/Show/Hide SpatialDecomposition disappeared from UI
**Error Description and Steps to Reproduce**

Isolate/Show/Hide SpatialDecomposition disappeared from UI. See images below.

**Attachments**

Image from [tutorial](https://docs.bonsaibim.org/quickstart/explore_model.html):

Image

Current UI:

Image

**Debug information**

```
os: Windows
os_version: 10.0.26100
python_version: 3.11.11
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: AMD64 Family 23 Model 8 Stepping 2, AuthenticAMD
blender_version: 4.4.0
bonsai_version: 0.8.2-alpha250321
bonsai_commit_hash: b5c0624
bonsai_commit_date: 2025-03-21T18:39:38+05:00
last_actions: 
# operator: BIM_OT_spatial_hotkey
# operator: BIM_OT_spatial_hotkey
# operator: BIM_OT_spatial_hotkey
# operator: BIM_OT_spatial_hotkey
# operator: BIM_OT_override_object_delete
last_error: None
ifc: myifc.ifczip
schema: IFC4
preprocessor_version: DDS_IFC v3.0
originating_system: IFC file generated by Graphisoft Archicad 27.0.0 FRA FULL Windows version (IFC add-on version: 5030 FRA FULL).
```

BIMvoice:
Fixes #6410.

The quickstart tutorial showed the container level Isolate and Hide/Show icons in the Spatial Decomposition panel as default UI. They aren't. @sboddy explained this was a deliberate redesign (element level isolate/hide/show now live in the Elements list), and pointed at #6664 for why: genuine, still open interaction bugs between the newer element level buttons and active status filters (not the older container level buttons, thanks @sboddy for the correction). @CyrilWaechter pushed back that the redesign felt less intuitive.

#6664 is still open with no fix landed, so I did not flip the `container_hide_show_isolate` default to true, that would silently reintroduce the redesign UX change for every user by default before the underlying element/status filter interaction bug is actually fixed.

The actual remaining gap is the tutorial doc itself. Updated `src/bonsai/docs/quickstart/explore_model.rst` to point readers at the Elements list icons (today's actual default), and added a note explaining that the Elements list icons don't currently respect an active status filter and vice versa, and that the container level buttons remain available as an opt in precisely because they don't have that problem.

Docs only, no code change.

### DIFF
--- a/src/bonsai/docs/quickstart/explore_model.rst
+++ b/src/bonsai/docs/quickstart/explore_model.rst
@@ -142,12 +142,13 @@ find objects, and use the **Select Icon** to select them.
 
 .. note::
 
-   The **Isolate** button and **Hide / Show Icons** shown above next to the
-   **Spatial Container** list are not enabled by default, since they can
-   conflict with active status filters (see
-   :menuselection:`Properties --> Scheduling --> Status`). To bring them
-   back, enable **Container hide/show/isolate** under
-   :menuselection:`Edit --> Preferences --> Add-ons --> Bonsai --> Extras`.
+   The **Element** list **Isolate** button and **Hide / Show Icons** shown
+   above do not currently respect an active status filter (see
+   :menuselection:`Properties --> Scheduling --> Status`), and vice versa.
+   The older **Spatial Container** list versions of these buttons do not
+   have this problem, which is why they are still available as an opt in
+   under :menuselection:`Edit --> Preferences --> Add-ons --> Bonsai -->
+   Extras`, enable **Container hide/show/isolate** to bring them back.
 
 **Elements** are grouped into IFC **Classes**, such as Wall, Slab, or Door.
 Within that, **Elements** are grouped into **Construction Types**. You'll see a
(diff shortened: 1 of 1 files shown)


sboddy:
The statement "still open interaction bugs between the old container level buttons and active status filters" is **incorrect**.

The _correct_ behaviour is seen with the older _container_ level show/hide/isolate buttons and the status filters.
The _buggy_ behaviour is between the newer _element_ level show/hide/isolate buttons and the status filters.

The newer _element_ level buttons do not respect the status level filter, and vice versa. An action in either of these panes ignores anything set by the other pane. This mean for example that things hidden by an EPset_Status of DEMOLISH in the status pane will reappear when using the element pane show button on IfcWall classes. These two panes need to be fixed to work together. That is the reason I recreated the container level show/hide/isolate as a temporary measure.

BIMvoice:
Thanks for the correction, you're right, I had that backwards. Fixed the doc note and the PR description: it's the newer Element list icons that don't respect the status filter (and vice versa), not the older Container list ones, which is exactly why you brought those back as an opt in.

Issue #7334, opened 2025-11-10
Link svg|dxf


CyrilWaechter:
### Feature Description

You can currently add an svg as reference and add it to a sheet but I see no way to link it the way you can link an IFC. The goal would be to be able to use an svg|dxf as work background and update it easily.
eg. You receive an updated work background version every week. You don't want to place, scale multiple svg|dxf every week especially if you need to do this for each storey, section and elevation. Also it is not desirable to store it in our own file as `IfcAnnotation`.

theoryshaw:
Good feature request. 
Would be awesome too, if link printed out, as well, from Bonsai.

sboddy:
I was in discussion with Cyril about this. Currently the "Ifc Reference Image" seems to be implemented as a special case of an IfcAnnotation (outside of drawings), where a bitmap is setup as a style with a texture, and assigned to the IfcAnnotation plane. I was imagining that the selector for setting up the texture could be adapted to allow setting the value to alternative filetypes (i.e. svg) that would be dynamically converted behind the scenes (i.e. Inkscape command) to suitable bitmap images when loaded. If there was a refresh button on the texture it could be refreshed in-session too. Of course this is probably a bad idea because it is not Ifc conformant.

theoryshaw:
Ideally, for me, the objects, whatever they be, would be snappable as well. 
Raster images in this case, would be less than ideal.

sboddy:
Had another idea with the image textures being served up by the bonsai web server, and the server dynamically converting them behind the scenes. Currently Bonsai doesn't even retrieve served URI's - only on disk files. That doesn't address theoryshaw's snapping requirement though.

Gorgious56:
Blender can import and convert both dxf and svg to text / curve / meshes FWIW with extensions. Although on very dense input files performance is abysmal.

PR #8681, opened 2026-07-17
Bonsai: link external SVG/DXF files as snappable reference geometry

BIMvoice:
Proposal for #7334.

@theoryshaw @sboddy @Gorgious56 this mirrors the existing "Link IFC" mechanism rather than the `IfcAnnotation`-based "IFC Reference Image", since the reporter explicitly didn't want this stored as `IfcAnnotation`, and @sboddy himself flagged the texture-conversion idea as "probably a bad idea because it is not Ifc conformant."

- A new Linked References panel links an on-disk `.svg`/`.dxf`. The file is never imported into the IFC model itself. Persistence reuses the exact pattern `Link IFC` already uses: an `IfcDocumentInformation` (Scope `LINKED_REFERENCE`) plus an `IfcDocumentReference` carrying the file location and placement matrix, the same as how `Scope LINKED_MODEL` links already persist. No `IfcAnnotation`, no texture entities.
- Geometry imports as real, snappable objects (@theoryshaw's requirement, @Gorgious56's suggestion): SVG via Blender's own bundled `io_curve_svg` importer, DXF via the `ezdxf` library Bonsai already bundles as a wheel. No raster textures, no new dependency, no Inkscape.
- Each link gets an anchor empty; imported geometry parents to it. You place/scale the anchor once, refresh re-imports and re-parents under the same anchor, so the recorded placement reapplies automatically instead of needing to be redone every time an updated file arrives.
- Manual Refresh button, plus an opt-in auto-refresh timer that polls file mtimes (same pattern as the existing autosave timer).

Scoped out for now: printing linked geometry through Bonsai's drawing pipeline (a real follow-up, since sheets already composite SVG underlays), and the web-server/URI-serving idea, which wasn't actually what was requested.

Live-tested across Blender 4.5.8, 5.1.2 and 5.2.0: real import, transform preservation across refresh, IFC save/reload round trip, auto-refresh detection. New pytest coverage in `test/tool/test_linked_reference.py`, existing `test_document.py` unaffected.

Marked draft, this is a proposal, would like feedback on the design before treating it as final.

### DIFF
--- a/src/bonsai/bonsai/bim/__init__.py
+++ b/src/bonsai/bonsai/bim/__init__.py
@@ -91,6 +91,7 @@ modules = {
     "light": None,
     "alignment": None,
     "clip_box": None,
+    "linked_reference": None,
     # Uncomment this line to enable loading of the demo module. Happy hacking!
     # The name "demo" must correlate to a folder name in `bim/module/`.
     # "demo": None,
--- a/src/bonsai/bonsai/bim/export_ifc.py
+++ b/src/bonsai/bonsai/bim/export_ifc.py
@@ -45,6 +45,7 @@ class IfcExporter:
         self.set_header()
         IfcStore.update_cache()
         self.sync_all_objects()
+        tool.LinkedReference.sync_placements_to_ifc()
         extension = self.ifc_export_settings.output_file.split(".")[-1].lower()
         if extension == "ifczip":
             with tempfile.TemporaryDirectory() as unzipped_path:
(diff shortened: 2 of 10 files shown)


sboddy:
You have an easily fixable linting failure. I actually think it's being a little fussy here, but it makes the check fail, so it still needs fixing.

BIMvoice:
@sboddy fixed, thanks for the catch. `ty` needed an explicit `import ezdxf.entities` alongside the lazy `ezdxf` imports since the annotation references `ezdxf.entities.DXFEntity`. Pushed in d858cf5, lint-formatting is green on the current head; the remaining red check is the unrelated chronic compile-and-test failure on the base.

Issue #431, opened 2018-08-09
create_shape failed after changing from python2.7 to 3.6


NicoPauen:
Hello,

currently I am using ifcopenshell 0.6.0.a0 for OSx64 in the Anaconda Distribution to analyze the geometric representation of elements. I am programming a tool for my master thesis which should run on a windows server with python 3.6. Therefore I created a new environment (python3.6) on my Mac with anaconda over a VM (parallel desktop) using windowsx64 and installed ifcopenshell 0.6.0.a0 for Windowsx64 and pythonOCC-core 0.17.3 over anaconda cloud. Now I have a problem with the create_shape function. 

##
import ifcopenshell
import ifcopenshell.geom

pack_ifc_file = ifcopenshell.open('AC20-Institute-Var-2.ifc')
products = pack_ifc_file.by_type("IfcProduct")
settings = ifcopenshell.geom.settings()
#settings.set(settings.USE_PYTHON_OPENCASCADE, True)

for product in products:
    if product.Representation is not None:
        try:
            shape = ifcopenshell.geom.create_shape(settings, inst=product).geometry
            print('worked once')
        except:
            print('Error')
##

This prints out an error for every product: "RuntimeError: Failed to process shape"

Also, when I run the script a second time, the programm isn't able to open the given .ifc file and the Kernel dies in Spyder and has to restart. 

The IFC-model is a popular model and can be downloaded at this link:
http://www.ifcwiki.org/index.php?title=KIT_IFC_Examples

Any ideas what I can change to make it work again? And do you need more information on my system to locate the error?

Thanks
Nico

aothms:
I'll have to look into this. In the mean time, you can try recent builds in the 0.5 release below in case that's an option. In terms of geometry handling these should be the same, just the 0.6 branch is able to select between different schema versions at runtime.

https://github.com/IfcOpenBot/IfcOpenShell/commit/54d9edfee5785714ba27f9be3f244538a5a70b9a#comments

CyrilWaechter:
It currently still fails on IfcAnnotation objects:
```
#22993=IfcAnnotation('3rHhfke4v7Nw4nsmGr899j',#12,$,$,$,#22798,#22990)
#23193=IfcAnnotation('1PtnYYb3j3hRRjYBJRuILz',#12,$,$,$,#23005,#23190)
#23300=IfcAnnotation('0sg_walh98nRCadiOfzzZh',#12,$,$,$,#23204,#23297)
[...]

PR #8682, opened 2026-07-17
ifcgeom: skip text literals with a warning instead of a conversion error

BIMvoice:
See #431.

CyrilWaechter narrowed the original 2018 report down to one real bug in 2023: `IfcTextLiteralWithExtent` has no geometry mapping, so converting an `IfcAnnotation` containing one produced a generic "No operation defined for" error, escalating to a misleading "Failed to convert" error before `create_shape` raised.

Binds `IfcTextLiteral` (covers `IfcTextLiteralWithExtent`) to mark it as failed on purpose, the same pattern already used for `IfcBoundingBox` when deliberately skipped. Mixed representations now convert their remaining items cleanly, and the skip is a single clear warning instead of an error cascade. A text-only representation still raises in `create_shape` since there's genuinely no geometry to return, consistent with a bounding-box-only representation, and the iterator path continues over failures as before.

Rendering actual glyph geometry (the freetype idea floated in the thread) stays out of scope: the official dependency build compiles OCCT with `USE_FREETYPE=OFF` and doesn't link `TKService`, so there's no already-integrated font tessellation to build on in any current release toolchain. A real implementation would need FreeType enabled across every platform build, kernel-agnostic glyph tessellation (both OCCT and CGAL), and a font-resolution policy for the usually-absent `IfcTextStyleFontModel`. That's a genuine multi-week feature and a build-system decision, not something this PR attempts.

Verified against the original `AC20-Institute-Var-2.ifc` sample from the issue: with curves enabled, 112/113 annotations convert with warnings only (previously 8 hard errors), and `IfcConvert` output is byte-identical otherwise. New regression test in `test_create_shape.py`.

### DIFF
--- a/src/ifcgeom/mapping/IfcTextLiteral.cpp
+++ b/src/ifcgeom/mapping/IfcTextLiteral.cpp
@@ -1,0 +1,32 @@
+/********************************************************************************
+ *                                                                              *
+ * This file is part of IfcOpenShell.                                           *
+ *                                                                              *
+ * IfcOpenShell is free software: you can redistribute it and/or modify         *
+ * it under the terms of the Lesser GNU General Public License as published by  *
+ * the Free Software Foundation, either version 3.0 of the License, or          *
+ * (at your option) any later version.                                          *
+ *                                                                              *
+ * IfcOpenShell is distributed in the hope that it will be useful,              *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 *
+ * Lesser GNU General Public License for more details.                          *
+ *                                                                              *
+ * You should have received a copy of the Lesser GNU General Public License     *
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.         *
+ *                                                                              *
+ ********************************************************************************/
+
+// This file was generated with the assistance of an AI coding tool.
+
+#include "mapping.h"
+#define mapping POSTFIX_SCHEMA(mapping)
+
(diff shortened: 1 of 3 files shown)


aothms:
I don't agree with this fix. Unsupported is unsupported - and this should be logged. The problem is more likely in the structure of the model or too little consensus in the IFC spec on how annotation contexts are differentiated from 3d body contexts.

Your mention of FreeType is interesting though. Previously we did depend on FreeType, but this was no longer necessary at some point https://github.com/IfcOpenShell/IfcOpenShell/commit/8e639b00e314a2ed7f31878f9ae65617a6f3b578

Would be interesting to see what proper TextLiteral to Triangulation Geometry looks like. But not here on v0.8. Please propose based on the `ifcviewer-wgpu` branch (soon v0.9) a PR to bring back FreeType and use for IfcTextLiteral support (to taxonomy::shell I guess) have a look at the IFC schema what other text related entities extent, e.g -WithExtent and try to provide a comprehensive implementation.

BIMvoice:
Thanks @aothms for the review and for the FreeType history, useful context I didn't have.

Closing this. You're right that "unsupported, log it" is the correct baseline, and that the real fix is proper TextLiteral to geometry conversion, not silencing the error. That's a genuine feature (FreeType glyph tessellation, IfcTextLiteralWithExtent and related entities), scoped for ifcviewer-wgpu per your note, not a v0.8 patch.

Worth keeping: the 112/113 vs 8/113 annotation count from AC20-Institute-Var-2.ifc as a before/after baseline, and the test_create_shape.py regression test structure, if useful as a starting point for whoever picks this up. Filing a scoped issue for your ask so it isn't lost.

PR #8683, opened 2026-07-17
ifcparse: store integer attribute values as int64_t to allow out-of-range timestamps

### LINKED ISSUE #3058: OverflowError: in method 'entity_instance_setArgumentAsInt', argument 3 of type 'int'
Hello,

I encountered an issue while trying to set a low int as creation date. In some date system 30.12.1899 is 0 for example in LibreOffice/Excel sheets. See [Epoch (computing)](https://en.wikipedia.org/wiki/Epoch_(computing)). If you convert it to an ifc compatible date you get `-2209165200` which apparently currently produce an overflow. So a date left to default 0 produce an error.

## Error message
```python
Exception has occurred: OverflowError
in method 'entity_instance_setArgumentAsInt', argument 3 of type 'int'
  File "/home/cyril/git/IfcOpenShell/src/ifcopenshell-python/ifcopenshell/ifcopenshell_wrapper.py", line 3695, in setArgumentAsInt
    return _ifcopenshell_wrapper.entity_instance_setArgumentAsInt(self, i, v)
  File "/home/cyril/git/IfcOpenShell/src/ifcopenshell-python/ifcopenshell/entity_instance.py", line 267, in __setitem__
    self.method_list[idx](
  File "/home/cyril/git/IfcOpenShell/src/ifcopenshell-python/ifcopenshell/file.py", line 286, in create_entity
    e[idx] = arg
  File "/home/cyril/git/IfcOpenShell/src/ifcopenshell-python/scripts/issue_lowtimestamp.py", line 52, in <module>
    owner_history = file.createIfcOwnerHistory(
OverflowError: in method 'entity_instance_setArgumentAsInt', argument 3 of type 'int'
```

## Code to reproduce
```python
import uuid
import ifcopenshell
import datetime


file = ifcopenshell.file(schema='IFC4')

# https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/link/ifcaddress.htm
address = file.createIfcAddress(
    Purpose="OFFICE",
    Description="Anton Philipslaan 199\n5616TW Eindhoven\nThe Netherlands",
)

# https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/link/ifcorganization.htm
organisation = file.createIfcOrganization(
    Name="AECGeeks",
    Description="""Software development and consultancy for the
    Architecture Engineering and Construction industry""",
    Addresses=[address],
)

# https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/link/ifcapplication.htm
application = file.createIfcApplication(
    ApplicationDeveloper=organisation,
    Version=ifcopenshell.version,
    ApplicationFullName="IfcOpenShell",
    ApplicationIdentifier="ifcopenshell",
)

# https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/link/ifcroleenum.htm
role = file.createIfcActorRole("MANUFACTURER")

# https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/link/ifcperson.htm
person = file.createIfcPerson(
    Identification=str(uuid.uuid4()),
    FamilyName="Unknown",
    GivenName="Unknown",
    Roles=[role],
)

# https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/link/ifcorganization.htm
organisation = file.createIfcOrganization(
    Identification="companyid", Name="company", Roles=[role]
)

# https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/link/ifcpersonandorganization.htm
person_and_organisation = file.createIfcPersonAndOrganization(
    person, organisation, Roles=[role]
)

# https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2_TC1/HTML/link/ifcownerhistory.htm
owner_history = file.createIfcOwnerHistory(
    OwningUser=person_and_organisation,
    OwningApplication=application,
    CreationDate=int(datetime.datetime(1899, 12, 30).timestamp()),
)
```

BIMvoice:
Fixes #3058.

Setting an `IfcInteger`/`IfcTimeStamp`-typed attribute (e.g. `IfcOwnerHistory.CreationDate`) outside the signed 32-bit range raised `OverflowError: entity_instance_setArgumentAsInt, argument 3 of type 'int'`. Any Unix timestamp before 1901-12-13 or after 2038-01-19 fell outside that range, so those dates couldn't be authored at all. @aothms gave the direction back in 2023 ("make integers int64_t"); this narrows that to just the scalar integer attribute path rather than a full repo-wide migration.

- The scalar slot of the attribute variant (`Argument_INT`) becomes `int64_t`. Integer aggregates (`Argument_AGGREGATE_OF_INT`, e.g. `CoordIndex`) and instance/reference identifiers stay 32-bit, since neither is the value that overflows here.
- `IfcBaseClass::set_attribute_value` promotes the schema-generated `int` to `int64_t` at a single choke point, so the thousands of generated setters keep compiling unchanged.
- The STEP lexer, writer, and SWIG wrapper all widened together, since widening only the Python-facing setter would have silently wrapped the value on file write instead of raising (this is exactly what a naive attempt in the original thread ran into).

Verified in a build (IFC2X3 and IFC4): pre-1901, post-2038, both 32-bit boundaries, and a 9e12 value all round trip exactly both in memory and through STEP text serialization (write then reopen). Ordinary in-range integers and integer aggregates are unaffected. Existing entity/attribute test suite passes.

Built and tested against IFC2X3 + IFC4 only in this pass, not the full schema set (e.g. IFC4X3 variants), flagging that for anyone building the wider schema set to double check.

### DIFF
--- a/src/ifcparse/IfcEntityInstanceData.cpp
+++ b/src/ifcparse/IfcEntityInstanceData.cpp
@@ -9,7 +9,7 @@ public:
 
     int operator()(const Blank& /*i*/) const { return -1; }
     int operator()(const Derived& /*i*/) const { return -1; }
-    int operator()(const int& /*i*/) const { return -1; }
+    int operator()(const int64_t& /*i*/) const { return -1; }
     int operator()(const bool& /*i*/) const { return -1; }
     int operator()(const boost::logic::tribool& /*i*/) const { return -1; }
     int operator()(const double& /*i*/) const { return -1; }
@@ -125,7 +125,12 @@ namespace {
 
 AttributeValue::operator int() const
 {
-    return dispatch_get_<int>(array_, storage_model_, instance_name_, entity_or_type_, index_);
+    return (int)dispatch_get_<int64_t>(array_, storage_model_, instance_name_, entity_or_type_, index_);
+}
+
+AttributeValue::operator int64_t() const
+{
+    return dispatch_get_<int64_t>(array_, storage_model_, instance_name_, entity_or_type_, index_);
 }
 
 AttributeValue::operator bool() const
@@ -526,7 +531,7 @@ void rocks_db_attribute_storage::set(void* storage, const IfcParse::declaration*
 }
 
 template IFC_PARSE_API void rocks_db_attribute_storage::set<Blank>(void* storage, const IfcParse::declaration* decl, std::size_t identity, size_t index, const Blank& value);
-template IFC_PARSE_API void rocks_db_attribute_storage::set<int>(void* storage, const IfcParse::declaration* decl, std::size_t identity, size_t index, const int& value);
+template IFC_PARSE_API void rocks_db_attribute_storage::set<int64_t>(void* storage, const IfcParse::declaration* decl, std::size_t identity, size_t index, const int64_t& value);
 template IFC_PARSE_API void rocks_db_attribute_storage::set<bool>(void* storage, const IfcParse::declaration* decl, std::size_t identity, size_t index, const bool& value);
 template IFC_PARSE_API void rocks_db_attribute_storage::set<boost::logic::tribool>(void* storage, const IfcParse::declaration* decl, std::size_t identity, size_t index, const boost::logic::tribool& value);
 template IFC_PARSE_API void rocks_db_attribute_storage::set<double>(void* storage, const IfcParse::declaration* decl, std::size_t identity, size_t index, const double& value);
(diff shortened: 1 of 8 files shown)


aothms:
I don't want to make this fix here. Please propose a similar fix without worrying about compatibility in the `ifcviewer-wgpu` branch (which will become the branch for v0.9)

BIMvoice:
Thanks @aothms for the direction here.

Closing this PR on v0.8.0. Per your comment, the int64_t widening for out-of-range timestamps belongs on ifcviewer-wgpu instead, and that's already done, merged as #8754, ported to the rewritten ifcparse architecture on that branch (which also had its own, worse, silent-wraparound version of this bug, caught and fixed there too).

#3058 stays open against v0.8.0 as a known limitation until the wgpu-branch fix ships in v0.9. Nothing else to salvage here, the port covers it.

PR #8685, opened 2026-07-17
win: pre-detect VS2017 MSBuild path before building mpir dependency

### LINKED ISSUE #8631: 2017 Build not working
### Bug Description

Trying to build in an x64 Native Tools Visual Studio 2017.

build-deps.cmd always ends up with

Build started at 20:46:41.25.

Found existing "C:\Users\chris\work\cad\IfcOpenshell\_deps\nuget.exe", skipping
Found existing "C:\Users\chris\work\cad\IfcOpenshell\_deps\ccache.4.12.1\tools", skipping
Found existing "C:\Users\chris\work\cad\IfcOpenshell\_deps-vs2017-x64-installed\proj-9.4.1", skipping
Cloning mpir is already cloned.
HEAD is now at 365957bc vs26 - update toolsets
\msbuild.exe /p:Platform=x64 /p:Configuration=Release /p:"WindowsLIB0SDKLIB0Version=" .\lib_mpir_gc\lib_mpir_gc.vcxproj
'\msbuild.exe' is not recognized as an internal or external command,
operable program or batch file.
ERROR: msbuild failed for main project with exit code 9009

An error occurred Last logged action:

Build ended at 20:46:42.08. Time elapsed 0:00:00.83.

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Fixes #8631.

The actual bug lives in `Andrej730/mpir-vs2026`, a build dependency we clone but don't control (`win/build-deps.cmd` calls into its `msvc\vs17\msbuild.bat` when building mpir). That script locates MSBuild via an unconstrained `vswhere -latest` and only accepts the result if it looks like a VS2017 layout (`MSBuild\15.0\Bin`). When a newer Visual Studio (2019/2022/2026) is also installed, `vswhere -latest` picks that newer install instead, the VS2017 check silently fails, and the build fails with a bare `\msbuild.exe is not recognized` error, exactly as reported.

Since we can't fix the detection logic inside that external repo, this pre-detects the correct VS2017 MSBuild path in our own `build-deps.cmd` and sets `msbdir` before invoking the mpir build. The external script already honors a pre-set `msbdir` and skips its own (broken) detection when one exists, this is exactly the workaround already suggested in the thread, just automated. Uses `vswhere -version "[15.0,16.0)"`, the documented syntax for constraining to VS2017 only, so it finds the right install even with a newer VS present. Errors out clearly (rather than silently) if `vswhere` or a matching install can't be found.

Confirmed against the actual `Andrej730/mpir-vs2026` scripts that this is genuinely specific to the VS2017 path, VS2019+ use a different mechanism that isn't affected.

**This is untested on real Windows** (no Windows environment available here), it's a carefully traced, best-effort fix, not a live-verified one. @ccic-techni since you're already set up to test the manual `msbdir` workaround, would you be willing to try this branch instead and confirm whether it resolves it without needing to set `msbdir` by hand? That would be a real help in confirming this actually works before merge.


aothms:
I have forked this into https://github.com/IfcOpenShell/mpir-vs2026 please propose the logic fix there and update build scripts to point to this fork

BIMvoice:
@aothms Forked to `IfcOpenShell/mpir-vs2026` as you suggested and proposed the actual root-cause fix there (constrained the `vswhere` query to `-version "[15.0,16.0)"` so it can't pick a newer VS instead of 2017): https://github.com/IfcOpenShell/mpir-vs2026/pull/1

Also updated this PR to clone from that fork instead of `Andrej730/mpir-vs2026`, keeping the local `msbdir` pre-detection here too as a belt-and-braces safety net until the fork-side fix is reviewed/merged (it's a no-op once that lands, since the script already skips its own detection when `msbdir` is pre-set).

Issue #3783, opened 2023-09-21
Assets are not tracked in BBIM IFC Git repositories


brunopostle:
Although a Git repository can track many files, the BlenderBIM IFC Git functionality tracks one IFC file at a time (you can have multiple IFC files in a repo, but you need to commit them separately, which I think works ok conceptually since BlenderBIM only does one file at a time).

But a BlenderBIM project can have non-IFC assets and these should be tracked in the repo too:

* There are files related to generating drawings, these are paths relative to the IFC file: `stylesheet_path`, `markers_path`, `symbols_path`, `patterns_path`, and all files in `titleblocks_dir`.
* I think that everything in `sheets_dir` and `layouts_dir` is generated, so it *shouldn't* be tracked.
* There are IFC files in `pset_dir`, do these need to be shared when the project is shared?
* There is a JSON file in `shading_styles_path`, I'm not sure what this does. It doesn't look like something you would want to share with other users.
* There are also externally defined styles that would be .blend files and maybe JPEG, PNG etc..
* Anything else?

So there are some problems here:

1. The IFC Git interface can handle a merge conflict with IFC files, but it can't resolve JSON, CSS, SVG conflicts (unless it arbitrarily clobbers files).
2. IFC project files can be in sub-folders within the repository, if these paths are relative to the parent IFC these assets may exist multiple times in a single repository. Whereas IFC files in the same folder typically share assets, except when they don't.
3. The relative paths such as `stylesheet_path` are actually blender user settings, but they are then stored in the IFC, once per 'drawing'. Similarly IfcExternallyDefinedSurfaceStyle has a URI defined in the IFC project itself. If I want to track every asset needed to share the current IFC project I need to recurse through all the drawings and external defined styles in the IFC, and collect the stored paths, determine if the files exist in the repo folder, add, commit them etc..
4. Titleblocks need to be shared, but the IFC internally doesn't distinguish these with layouts and drawings which are generated.
5. SVG files can include other files, they would also need to be parsed (recursively) to collect everything needed to generate drawings. 

I think that attempting to track all this in Git with BlenderBIM is likely to break horribly, particularly I don't have a good solution for (1), everything else is doable. So for the time being users need to use an external Git GUI for this stuff.

Anyway, just needed to put this in writing and see if anyone has ideas for resolving (1).

falken10vdl:
[...]

PR #8687, opened 2026-07-17
Bonsai: commit referenced drawing assets alongside the IFC in IfcGit

BIMvoice:
Partially addresses #3783, the part @brunopostle himself called doable: "everything else is doable", separate from the merge-conflict-resolution problem (1) he flagged and explicitly asked for ideas on, which this does not attempt.

The IfcGit panel commits only the active IFC file, so drawing assets the IFC itself references (stylesheets, marker/symbol/pattern SVGs, sheet titleblocks) never made it into the repository, and a cloned project couldn't regenerate its drawings without the user remembering to stage those files with an external git client.

Bonsai already stores every one of these references inside the IFC: `EPset_Drawing` carries `Stylesheet`/`Markers`/`Symbols`/`Patterns` per drawing, each sheet's `IfcDocumentInformation` has a document reference described as `TITLEBLOCK`, and `BBIM_Documentation` on `IfcProject` carries `TitleblocksDir`. That makes the IFC file itself the canonical source for "which assets does this project need", exactly the enumeration you described needing in point 3, so no extra bookkeeping is required.

On every IfcGit commit, this now collects those referenced paths, resolves them relative to the IFC file, and stages the ones that exist inside the repo working tree so they land in the same commit as the IFC. Referenced files outside the repository (e.g. a shared system-wide stylesheet) can't be tracked in this repo, so they're reported as an operator warning rather than failing or being silently dropped.

Deliberately out of scope, matching your own boundaries in the issue:
- Generated outputs (`sheets_dir`/`layouts_dir`, the drawing SVGs themselves) stay untracked.
- `ShadingStyles` JSON, `pset_dir`, externally defined `.blend`/image styles are left alone, you weren't sure these should be shared either.
- No attempt at merge conflict resolution for CSS/SVG/JSON assets (problem 1), git's normal conflict behavior applies, same as tracking them by hand today.
- SVGs are not parsed for recursively embedded files (problem 5).

Live-tested: a project with a drawing (stylesheet + markers/symbols/patterns), a sheet (titleblock reference), and one asset deliberately placed outside the repo. Confirmed via the actual resulting commits that exactly the referenced in-repo files land alongside the IFC, generated sheet/layout output stays untracked, and the outside-repo file triggers a warning instead of silently vanishing. New pytest coverage for the enumeration and staging logic.

### DIFF
--- a/src/bonsai/bonsai/bim/module/ifcgit/operator.py
+++ b/src/bonsai/bonsai/bim/module/ifcgit/operator.py
@@ -137,7 +137,7 @@ class CommitChanges(bpy.types.Operator):
         props = tool.IfcGit.get_ifcgit_props()
         commit_message = props.commit_message
         new_branch_name = props.new_branch_name
-        core.commit_changes(tool.IfcGit, tool.Ifc, commit_message, new_branch_name)
+        core.commit_changes(tool.IfcGit, tool.Ifc, commit_message, new_branch_name, operator=self)
         props.new_branch_name = ""
         props.commit_message = ""
         core.refresh_revision_list(tool.IfcGit, tool.Ifc)
--- a/src/bonsai/bonsai/core/ifcgit.py
+++ b/src/bonsai/bonsai/core/ifcgit.py
@@ -57,10 +57,21 @@ def discard_uncommitted(ifcgit: type[tool.IfcGit], ifc: type[tool.Ifc]) -> None:
 
 
 def commit_changes(
-    ifcgit: type[tool.IfcGit], ifc: type[tool.Ifc], commit_message: str, new_branch_name: str = ""
+    ifcgit: type[tool.IfcGit],
+    ifc: type[tool.Ifc],
+    commit_message: str,
+    new_branch_name: str = "",
+    operator: bpy.types.Operator | None = None,
 ) -> None:
-    """Commit and create new branches as required"""
+    """Commit the IFC file and its referenced assets, creating new branches as required"""
     path_ifc = ifc.get_path()
+
+    external_assets = ifcgit.stage_asset_files(ifcgit.get_project_asset_paths(path_ifc))
+    if external_assets and operator:
+        operator.report(
+            {"WARNING"},
+            "Referenced assets outside the repository were not committed: " + ", ".join(external_assets),
+        )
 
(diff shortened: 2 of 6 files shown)


brunopostle:
It seems to work, though I was confused by my project that only has drawings configured in branches with 'main' containing the site survey (which then gets merged into 'existing' and 'proposed' branches).

I realise that the layout/ folder does actually contain svg files that need to be tracked as these are user editable, so maybe these need adding to tracked files.

BIMvoice:
@brunopostle Thanks for testing. Good catch on the layout SVGs, you are right that those are user edited (drawing placement on each sheet lives there), only the composed sheets/ output is generated from them. Each sheet's IfcDocumentInformation already carries a LAYOUT document reference next to the TITLEBLOCK one, so the enumeration now collects both and the layouts/ SVGs land in the same commit as the IFC. Verified live in a headless Blender project with a sheet: layout and titleblock are committed alongside the IFC, and the generated sheets/ output stays untracked.

On the branch confusion: references to files that do not exist on disk are skipped quietly (only referenced files that exist outside the repository produce a warning), so a branch where drawings are not configured commits exactly as it did before. If that still reads as confusing in the panel I am happy to look at surfacing it better.

brunopostle:
Thanks, the confusion was all mine, it was my decision to have a main branch without ant drawings.

This all seems to work as advertised, but I'd like to see it get more testing before it is merged with the bonsai release branch.

BIMvoice:
Ran a real-world test matrix in headless Blender: real Bonsai authoring operators (not synthetic psets), a real git repo on disk, exercising the actual commit flow. Three scenarios: a full drawing plus sheet with stylesheet/titleblock/layout assets, a project with no drawings configured (your own branch scenario), and an asset referenced from outside the repo (correctly warned and excluded). No bugs surfaced, and the three scenarios are now committed as permanent integration tests (`TestCommitChangesRealProject`) alongside the existing unit tests, so the feature keeps regression coverage that drives real operators and a real git repo rather than mocks. One note: I could not find a live download of AC20-FZK-Haus to add it specifically (mirrors 404), happy to include it if you can point me at one, though the asset-tracking logic depends on drawing/sheet psets rather than building geometry. Pushed.

PR #8690, opened 2026-07-17
Fix serialize_shape() dropping the outermost shape's placement

### LINKED ISSUE #2992: serializer creates different placements in ifcs for breps with same placement
I am playing with nurbs and serializer to export nurbs out of FreeCAD into ifc. ATM FreeCAD has problems with the place in ifc, but it seams IfcOpenShell is the culprit.

Attached two brep files. If they are loaded into FreeCAD the cylinder is on the same place. But if a ifc is created out of the brep the cylinder are on totally different places. Attached the breps and ifcs. Code to create the ifc can be found here: https://community.osarch.org/discussion/comment/15216/#Comment_15216

very interesting ... The breps differ only in a few character

```
$ diff cyl_removeSplitter.brep cyl_copyGlobPlace.brep
110c110
< -2 1 *
---
> -2 0 *
112c112
< +1 0
\ No newline at end of file
---
> +1 1
\ No newline at end of file

```
[cyl_copyGlobPlace_IfcOS.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/11237263/cyl_copyGlobPlace_IfcOS.ifc.txt)
[cyl_removeSplitter.brep.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/11237264/cyl_removeSplitter.brep.txt)
[cyl_removeSplitter_IFCOS.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/11237265/cyl_removeSplitter_IFCOS.ifc.txt)
[cyl_copyGlobPlace.brep.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/11237266/cyl_copyGlobPlace.brep.txt)

BIMvoice:
Fixes #2992.

`BRepTools_ShapeSet.Add()` only persists a shape's `TopLoc_Location` when that shape is referenced as a child of another shape in the set. The outermost shape has no parent, so its own location was silently dropped on write. This made two OCC shapes that FreeCAD (and OCCT itself) consider identically placed serialize to different results, depending on whether the placement was baked into the underlying geometry or carried as a location on the shape itself, exactly as reported.

Wraps the shape in an identity-located `TopoDS_Compound` before adding it to the shape set, so the original shape becomes a referenced child and its location round-trips correctly. This is the workaround @aothms suggested in the thread, and it mirrors what IfcOpenShell's own C++ side already does (`as_compound()` + `Serialize()`) for the `SerializedElement.geometry.brep_data` path, which was never affected by this bug, so the codebase is now internally consistent between the two paths.

Verified live with pythonocc-core 7.9.0 against a locally built `ifcopenshell_wrapper`: a box built at the origin and relocated via a top-level `TopLoc_Location` round-tripped through `ifcopenshell.geom.serialise()` and a full write/reopen/tessellate cycle to `(0,0,0)` instead of its actual placement before this fix, matching the reported symptom precisely. After the fix it matches an equivalent box built directly at the target location. An already-compound input shape and the ordinary non-relocated case are unaffected.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
+++ b/src/ifcopenshell-python/ifcopenshell/geom/occ_utils.py
@@ -34,6 +34,7 @@ import ifcopenshell.ifcopenshell_wrapper as ifcopenshell_wrapper
 try:
     from OCC.Core import (  # pyright: ignore[reportMissingImports]  # ty:ignore[unresolved-import]
         AIS,
+        BRep,
         BRepTools,
         Graphic3d,
         Quantity,
@@ -46,6 +47,7 @@ try:
 except ImportError:
     from OCC import (  # pyright: ignore[reportMissingImports]  # ty:ignore[unresolved-import]
         AIS,
+        BRep,
         BRepTools,
         Graphic3d,
         Quantity,
@@ -242,7 +244,16 @@ def serialize_shape(shape):
     # see whether this is necessary
     shapes.SetFormatNb(2)
 
-    shapes.Add(shape)
+    # BRepTools_ShapeSet does not persist the location of the outermost shape
+    # (it is only written for shapes referenced by a parent). Wrap in an
+    # identity-located compound so the original shape becomes a referenced
+    # child and its location round-trips. See issue #2992.
+    compound = TopoDS.TopoDS_Compound()
+    builder = BRep.BRep_Builder()
+    builder.MakeCompound(compound)
+    builder.Add(compound, shape)
+
+    shapes.Add(compound)
 
     # Check if WriteToString method exists and has the correct signature
     # In PythonOCC >= 7.8.0, WriteToString signature changed and requires additional arguments


aothms:
The proper way is not to use BRepTools_ShapeSet but instead the more higher level serialization primitives such as BRepTools::Write(..., ostream&, ...) but how this has been exposed in pythonocc has changed throughout the versions. Maybe do a matrix of what version supports what and pick a version there that works. I'm a little bit worried about wrapping in a compound because the hierarchy change may have implications and affects round tripping.

BIMvoice:
Investigated both points.

`BRepTools::Write` is available from pythonocc-core ~7.8 onward (not in 7.7.2, which only exposes file-path overloads), so version isn't actually the blocker. The real issue: this function's output is read by compiled C++ in `src/ifcwrap/IfcGeomWrapper.i`, which is hardwired to `BRepTools_ShapeSet::Read`, not `BRepTools::Read`. I tested swapping the write side to `BRepTools::Write` against the real compiled wrapper: it doesn't crash, it silently reproduces the exact bug this PR fixes (wrong format for that reader, placement lost again). Switching would mean also changing the compiled C++ reader, a much bigger, binary-compatibility-affecting change, not a simpler swap. (Separately, the `SerializedElement.geometry.brep_data` channel already uses `BRepTools::Write` consistently on both ends and was never affected by this bug, that path is unrelated and fine.)

On the hierarchy/round-tripping concern: tested the compound-wrap against a compound of 3 distinctly-placed solids and a 2-level compound-of-compounds, through the real write to reopen to tessellate pipeline. Vertex list, face/index list, ordering, and material ids came back byte-identical in every case. No observable effect from the extra wrapping.

aothms:
> Switching would mean also changing the compiled C++ reader, a much bigger, binary-compatibility-affecting change, not a simpler swap. 

I agree, but still we should make the change, just do it on the `ifcviewer-wgpu` (soon v0.9) branch where we can be a bit more backwards incompatible. And do it at all locations where (de)serialization is performed to make it consistent. OCCT isn't always clear which are low-level and which are high-level APIs, but consumers such as ourselves are intended to use BRepTools not the lower level ShapeSet which comes with these kind of (nearly) undocumented caveats.

BIMvoice:
Thanks @aothms for flagging both the BRepTools::Write concern and the compound wrapping risk, good catches on both.

Closing this PR. Confirmed the BRepTools::Write swap isn't a safe drop-in given the compiled ShapeSet reader on the other end (src/ifcwrap/IfcGeomWrapper.i), and confirmed the compound wrap doesn't corrupt round-tripping in the cases tested, but you're right the correct fix is migrating (de)serialization to the higher level BRepTools API everywhere it's used, applied consistently, and that belongs on ifcviewer-wgpu rather than v0.8.

#2992 stays open. Adding a comment there capturing the direction rather than opening a duplicate issue.

PR #7648, opened 2026-02-07
closes #3565 - extend profiles and basic extrusions to 3D cursor with…

theoryshaw:
This PR extends the Ctrl+E hotkey functionality to work with profile elements (beams, columns, members) and basic extruded solids, in addition to the existing LAYER2 wall support. Previously, Ctrl+E only extended walls vertically to the cursor's Z position. Now it intelligently handles different element types: for LAYER2 elements it maintains the existing vertical extension behavior, for PROFILE elements it extends the element to the 3D cursor location in any direction, and for basic extrusions (simple boxes, columns without material profiles) it projects the cursor onto the extrusion axis and extends the geometry accordingly.

PR #8693, opened 2026-07-17
Bonsai: let Shift+E extend non-typed profile extrusions

### LINKED ISSUE #3565: Have Shift+E and extending to the 3D cursor work on non-typed profile extrusions.
Would be nice if Shift+E and extending to the 3D cursor, worked on non-typed profile extrusions. 

2023-08-10_15-15-22_KeyShowView_Carnac

BIMvoice:
Fixes #3565.

Shift+E (`bim.extend_profile`) silently did nothing on elements with a plain, ad-hoc `IfcExtrudedAreaSolid` body and no `IfcMaterialProfileSet` material, e.g. the reporter's IfcActuator placed with a manually authored box extrusion instead of a typed column/beam generator.

Two places gated this on a profile-set material. `DumbProfileJoiner.recreate_profile` in `profile.py` returned early when the element had no profile-set material, instead of deriving a profile. `tool.Model.get_usage_type` classified such elements as having no usage at all, so the Shift+E hotkey dispatcher never even called `bim.extend_profile` in the first place, a deeper root cause than initially expected.

`recreate_profile` now falls back to reading the `SweptArea` straight off the element's existing Body extrusion (via the already-used `tool.Model.get_extrusion`) when there's no profile-set material, and raises a clear `MissingExtrudedProfileError` when the element has neither a profile-set material nor an extruded body, caught and reported as a normal operator error rather than a silent no-op or crash. `get_usage_type` now falls back to the same extruded-body check when the material isn't a layer or profile set, so the hotkey dispatcher routes these elements into the profile tools to begin with.

Live-verified in headless Blender 5.1: before the fix, Shift+E left an ad-hoc box element's geometry unchanged (silent no-op); after the fix, it extends to the 3D cursor exactly like a typed profile element does. T-joins between two ad-hoc elements now also create the expected `IfcRelConnectsPathElements` relationships. An element with no body representation at all now reports a clear error instead of a crash or silent no-op. Re-verified the existing typed profile (material profile set usage) case is completely unaffected.

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/profile.py
+++ b/src/bonsai/bonsai/bim/module/model/profile.py
@@ -267,6 +267,10 @@ class DumbProfileRegenerator:
         return results
 
 
+class MissingExtrudedProfileError(Exception):
+    pass
+
+
 class ExtendProfile(bpy.types.Operator, tool.Ifc.Operator):
     bl_idname = "bim.extend_profile"
     bl_label = "Extend Profile"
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing in favour of @theoryshaw's #7648, which predates this and covers the same issue. We should have found it before opening this one. Apologies for the noise.

PR #8695, opened 2026-07-17
Bonsai: add Round and Arch shapes to the parametric window

### LINKED ISSUE #3621: Window Modifier : Support for non-rectangular shapes
It would be nice to be able to parametrically defined non-rectangular shapes. eg 

image
image
image

BIMvoice:
Fixes #3621.

Adds a Window Shape option (Rectangle, Round, Arch) to the parametric window modifier.

**Round** is a circular porthole-style window: overall width is used as the diameter, overall height is ignored for geometry and synced to the diameter on the IFC occurrence attributes. Lining and frame are concentric annular rings (`IfcCircle` profiles with voids), glass is a disk.

**Arch** is a Palladian-style window: the arch radius is half the overall width, the apex sits at the overall height, so the spring line is derived (height minus radius) with no extra parameter needed. Below the spring line the existing rectangular single-panel construction is reused unchanged, treating the shared edge exactly like a transom. Above it, a semicircular fanlight is built from chord+arc profiles (`IfcIndexedPolyCurve` arc segments, with a polygonal fallback for IFC2X3, which forbids arcs in polylines), with radiating muntin bars fanning out from the arch center at even angular spacing (new "Arch Muntin Bars" property, default 4). Degrades gracefully to a lunette when height is at or below the radius.

Non-rectangular shapes are single-partition by design, selecting Round or Arch forces `SINGLE_PANEL` and hides the panel grid options. The default Rectangle shape goes through the exact original code path (the rectangular panel loop moved verbatim into an `else` branch, same variable names and structure, byte-for-byte tessellation-identical output verified for all 9 partition types).

Both geometry layers are covered: the authoritative IFC representation in `ifcopenshell.api.geometry.add_window_representation` and the live bmesh preview used while editing, so the full edit-mode flow works for the new shapes, including re-entering edit mode on saved shaped windows. Plan/elevation representations and simple wall openings keep their existing simplified rectangular output for now, stated clearly as a follow-up.

Also fixes a latent crash in `add_window_representation` when `overall_height`/`overall_width` were omitted: a default-value helper ran before `usecase.settings` was assigned, raising `AttributeError`.

Tested headless in Blender 5.1.2: 32 API-level geometry assertions (circularity, arch radius bounds, apex position, muntin count/placement, IFC2X3 fallback, all 9 partition types regression-checked) and 27 live Bonsai assertions (preview and final meshes, pset round-trip, occurrence attribute sync, save/reload persistence, stock-identical rectangular regression).

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/data.py
+++ b/src/bonsai/bonsai/bim/module/model/data.py
@@ -484,7 +484,9 @@ class WindowData:
         data = cls.data["pset_data"]["data_dict"]
         lining_data = data["lining_properties"]
         lining_params = {}
-        lining_props = props.get_lining_kwargs(window_type=data["window_type"])
+        lining_props = props.get_lining_kwargs(
+            window_type=data["window_type"], window_shape=data.get("window_shape", "RECTANGLE")
+        )
         for prop_name in lining_props:
             prop_readable_name, prop_value = get_prop_from_data(props, lining_data, prop_name)
             lining_params[prop_readable_name] = prop_value
--- a/src/bonsai/bonsai/bim/module/model/prop.py
+++ b/src/bonsai/bonsai/bim/module/model/prop.py
@@ -831,11 +831,21 @@ def window_type_prop_update(self, context):
     update_window(self, context)
 
 
+def window_shape_prop_update(self, context):
+    # non-rectangular shapes are single-partition by design
+    if self.window_shape != "RECTANGLE" and self.window_type != "SINGLE_PANEL":
+        self.window_type = "SINGLE_PANEL"  # triggers its own update
+        return
+    update_window(self, context)
+
+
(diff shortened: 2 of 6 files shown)


aothms:
> IfcCircle profiles with voids

Why not IfcCircleHollowProfileDef

Op vr 17 jul 2026, 12:24 schreef Petru Conduraru ***@***.***>:

BIMvoice:
Good catch, fixed. The round window's lining/frame now use `IfcCircleHollowProfileDef` directly (Radius + WallThickness) instead of the outer/inner-void construction, and the glass disk uses `IfcCircleProfileDef` instead of an arbitrary profile wrapping a bare circle. Same geometry, correct profile types now.

And no worries on the brevity, mobile replies are totally fine, happy to get more feedback whenever, however short.

Issue #2181, opened 2022-05-03
Authoring flow segments and fittings


Moult:
The goal is build operators that can perform connections between flow segments, and automatically insert flow fittings.

IFC does not establish any conventions around axis orientations for segments and fittings. I had a look at what Revit does, but their system is pretty unsemantic, where all object origins are at the storey, and e.g. for a rectangular ducts their rectangle X and Y parameters don't always correlate to the cross section. It's pretty terrible. So I'm going to follow what IFC does for beams / columns / profile sets which makes a ton more sense.

When creating a rectangular duct segment, the axis convention must always be like this: the duct cross section width is local X, height is local Y, and it is always extruded in the positive Z direction. The position must always be in the center as shown. This matches the convention for IFC profiles exactly.

![2022-05-03-121403_473x393_scrot](https://user-images.githubusercontent.com/88302/166395429-5b03cbe7-8870-40ed-8e12-dae6c2741202.png)

![image](https://user-images.githubusercontent.com/88302/166395509-f168d871-02cc-492d-b713-7609ce679c92.png)

The user must be able to choose a point with the 3D cursor, and use Shift-E to "extend" the point closest to the cursor. Similar to how extending walls works.

The user must also be able to choose a point with the 3D cursor and use Shift-S to "split" at the point closest to the cursor. Similar to how splitting walls works.

In addition, the following operations need to be possible:

 - Bend
 - Connector
 - Junction (more than 2 connections)
 - Obstruction (only needs one segment as input)
 - Transition (same as connector, but occurs between segments of two different types)

When a connection between two segments occurs a third object will be inserted called the fitting. The fitting geometry will be already available, so the only requirement is to derive the matrix_world of the fitting.

See screenshots of typical operations:

[mep-axis.blend.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/8608065/mep-axis.blend.txt)

![2022-05-03-124438_761x739_scrot](https://user-images.githubusercontent.com/88302/166399362-f06e07ac-b048-42a5-87d4-ee7f5a9968de.png)

aothms:
+1 on extrude towards +Z

Sent from a mobile device. Excuse my brevity. Kind regards, Thomas

Op di 3 mei 2022 04:53 schreef Dion Moult ***@***.***>:

Moult:
[...]

PR #8697, opened 2026-07-17
Bonsai: add MEP junction fitting operator (tee, wye, cross)

BIMvoice:
Part of #2181.

`FitFlowSegments` (Shift+Y "Add Fitting") already detected TEE/CROSS/WYE topologies when 3+ segments were selected, but its dispatch ended in a literal `print(fitting_type)`, a scaffold that was never wired up. `bim.mep_connect_elements` is strictly pairwise (2 elements, one `IfcRelConnectsPorts`, no fitting), so there was genuinely no junction-building path in the codebase.

Adds `MEPAddJunction` (`bim.mep_add_junction`), following the existing `MEPAddBend`/`MEPAddTransition` construction pattern: given 3+ selected same-type segments whose axes meet at a common point (least-squares solve, 1mm tolerance), it trims/extends each segment to the junction, builds one fitting type with one extruded leg per segment (oriented by that segment's own frame), adds one port per leg, and connects every segment via `IfcRelConnectsPorts`. Cable carrier fittings use `TEE`/`CROSS` (their enum has no `JUNCTION`); everything else uses `JUNCTION`. `FitFlowSegments` now routes 3+ selections here; the 2-segment tee/cross cases now report "split first" guidance instead of silently doing nothing.

Tested headless in Blender 5.2 across rectangular-duct T, 4-way cross, and 45-degree circular-pipe wye, in both metre and millimetre projects: correct port count/positions, correct segment trimming, `IfcCableCarrierFitting`/`IfcPipeFitting` predefined types resolve correctly, both error paths (segment running through the junction point, non-concurrent axes) cancel cleanly with no partial state, save/reload round-trips correctly, and `ifcopenshell.validate` reports 0 issues.

Verified Moult's stated segment axis convention (width = local X, height = local Y, extrusion +Z, centered) still holds in current segment creation.

Known limitations, left for follow-up: legs must share one segment type (no reducing junctions), each junction creates its own fresh fitting type rather than reusing a compatible existing one, a single shared leg length, and the MEP segment cardinal-point dropdown still allows an off-center pick that would misalign ports (not introduced by this change). This does not implement the issue's separate "Connector" operation, which remains open.

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/__init__.py
+++ b/src/bonsai/bonsai/bim/module/model/__init__.py
@@ -281,6 +281,7 @@ classes = (
     mep.MEPAddObstruction,
     mep.MEPAddTransition,
     mep.MEPAddBend,
+    mep.MEPAddJunction,
     mep.MEPRemoveTerminalFitting,
     mep.SelectMEPPathMembers,
     mep.MEPJoinSegments,
--- a/src/bonsai/bonsai/bim/module/model/mep.py
+++ b/src/bonsai/bonsai/bim/module/model/mep.py
@@ -214,37 +214,32 @@ class FitFlowSegments(bpy.types.Operator, tool.Ifc.Operator):
                     bpy.ops.bim.mep_add_bend()
                 elif is_on_axis1 and is_on_axis2:
                     fitting_type = "CROSS"
+                    self.report(
+                        {"INFO"},
+                        "Cross junctions need 4 segments: split both segments at the junction, "
+                        "select the 4 parts and repeat.",
+                    )
                 else:
                     fitting_type = "TEE"
+                    self.report(
+                        {"INFO"},
+                        "Tee junctions need 3 segments: split the main segment at the junction, "
+                        "select all 3 parts and repeat.",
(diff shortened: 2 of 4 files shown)


Isaina:
Can be this merged in main I need this for my plugin

BIMvoice:
> Can be this merged in main I need this for my plugin

can you use bonsai pr until approved?

Isaina:
Can you also go thru freecad bim bugs with Claude https://github.com/FreeCAD/FreeCAD/issues with Tag BIM and offer PR

BIMvoice:
Sorry I do not use FreeCAD and I don't know the tool at all, for now I'll focus on IfcOpenShell, maybe in the future, but not sure, no guarantee.

Isaina:
> Sorry I do not use FreeCAD and I don't know the tool at all, for now I'll focus on IfcOpenShell, maybe in the future, but not sure, no guarantee.

Yes freecad is ifcopenshell based and BIM is new workbench a little bit more tehnical then bonsai which is a little bit artistic

PR #8698, opened 2026-07-17
[Draft/RFC] ifcopenshell-python: generated typed IFC4 stubs for schema-aware by_type()

BIMvoice:
Draft, not requesting merge, sharing for feedback on the direction. Addresses the typing idea from #3306.

@aothms your comment there sketched hand-written `Optional`/`Type[T]` classes for 3 entities as a proof of concept. This generates the same shape for all 776 IFC4 entities directly from the compiled schema (`ifcopenshell_wrapper.schema_by_name`, the same source `file.by_type()` itself uses), rather than parsing `.exp` files: entity inheritance, defined types, enumerations as `Literal`, SELECTs as `Union`, aggregates, inverse attributes, and a `file.by_type()` overload set (`Literal["IfcWall"] -> list[IfcWall]` for all 776 classes, usable today via `cast("ifc4.file", model)`, plus a `type[T]` overload for future runtime support).

Pure stub file (`ifcopenshell/ifc4.pyi`), zero runtime changes, verified with `ty` (the CI-pinned checker): your `y()`/`z()` narrowing example from the issue now passes/fails exactly as expected, and running it against the full `ifcopenshell-python` tree adds no measurable checker time.

Only IFC4 generated so far. Natural next steps if this direction is useful: IFC2X3/IFC4X3, a CI regen check, and a small runtime shim to make the `type[T]` overload actually executable. Wanted your take before going further.


aothms:
But how would this work with multiple schemas?

We tend to write schema agnostic code in python. So we should be able to leverage structural typing to uncover assumptions when `Union[ifc2x3.IfcCartesianPointList3D, ifc4.IfcCartesianPointList3D, ifc4x3.IfcCartesianPointList3D]` obtained from an uncasted model.by_type("IfcCartesianPointList3D") only has TagList in case of IFC4.3.

Runtime schema info is based on model.schema or model.schema_identifier / or inst.model.schema_identifier. How can this be factored in into the typing, without this we can still not write the following without static errors:

```py

f = ifcopenshell.file
insts = f.by_type('IfcCartesianPointList3D')
for inst in insts:
    if inst.file.schema == 'IFC4X3':
        print(inst.TagList)
```

or

```py

f = ifcopenshell.file
if f.schema == 'IFC4X3':
    insts = f.by_type('IfcCartesianPointList3D')
    for inst in insts:
        print(inst.TagList)
```

I know this is a lot to ask, but it be really interesting to see how far we can take this for actual ergonomics and type safety - ideally without adding a ton of casts into the code - inference only.

BIMvoice:
Dug into this with `ty` directly. Short version: one of your two shapes works today with a small change, the other doesn't and here's exactly why.

Snippet 2 (`if f.schema == 'IFC4X3': insts = f.by_type(...)`) is fully solvable: I gave the generated `file` class a `schema: Literal["IFC4X3"]` (etc.) instead of inheriting the base `IFC_SCHEMA` union type. That's the only change needed. Verified against a real `Union[ifc4.file, ifc4x3.file]`: `f.schema == "IFC4X3"` narrows `f` with plain inference, no cast, and `by_type()` results plus their schema-specific attributes (`TagList`) type correctly inside the guard, and genuinely error outside it. Included `ifc4x3.pyi` (generated, same script) as a real second schema to make this concrete and reviewable, exercising your exact `IfcCartesianPointList3D`/`TagList` example; not proposing we commit to a full IFC2X3/IFC4X3 rollout here, that's a separate sequencing call for you.

Snippet 1 (checking `inst.file.schema` on an already-fetched instance) doesn't narrow through inference, and I don't think any checker does this: narrowing only applies to the exact expression checked, so `inst.file.schema == 'IFC4X3'` narrows `inst.file`, not `inst` (they're different expressions to the checker, even though one's derived from the other). A user-defined `TypeIs` predicate can bridge that gap (`ty` supports `TypeIs`/`TypeGuard` fine), and I've got one working for a single entity. But it doesn't generalize into one reusable helper across all entities: I tried a generic version keyed by a schema-tag marker + TypeVar, and it's rejected because Python has no intersection types to express "the union member that's also tagged IFC4X3." Making it work for real would mean generating a predicate per entity per schema, on the same order of size as the `by_type` overload set itself, not a small addition.

My take: lead people toward the Snippet 2 shape (narrow the file, then call `by_type`) as the supported idiom, it's genuinely free now. Snippet 1's literal shape stays a manual-cast or manual-TypeIs case unless we want to commit to generating that much more stub surface.

(Analysis and patch drafted with the assistance of an AI coding tool.)

aothms:
> Snippet 1 (checking inst.file.schema on an already-fetched instance) doesn't narrow through inference, and I don't think any checker does this: narrowing only applies to the exact expression checked

So does it help if we add a `@property` schema on entity_instance as well? Please do so if it helps.

Furthermore. I'm also still looking for a way so that typing is automatic. Any ifcopenshell.file behaves as Union[ifc2x3.file, ...] without importing additional things, but by automatic side loading of the type checkers. And without polluting the main python code with thousands of literal-based overloads.

Then, we should also look at ifcopenshell.entity_instance.is_a() what adaptations are necessary to feed this into structural typing:

```
inst = ...
if inst.is_a('IfcWall'):
    print(inst.GlobalId)
if inst.is_a() == 'IfcWall':
    print(inst.GlobalId)
```

^ assure no error occurs.

Also, what is needed to make isinstance checks work:

```
inst = ...
print(isinstance(inst, ifc2x3.IfcWall))
```

BIMvoice:
Real answers on all three, verified with `ty`, not reasoned about in the abstract:

**1. `entity_instance.schema` — yes, this fixes it.** Added a real `schema` property to `entity_instance` (delegating to `file.schema`), plus the per-schema `Literal[...]` base class technique already used for `file`. `if inst.schema == "IFC4X3":` now narrows `inst` itself correctly, no cast, since (unlike `inst.file.schema`) this checks an expression rooted at `inst` directly. Verified with `reveal_type`: unnarrowed, `inst.TagList` is `Any | tuple[str, ...] | None`; inside the guard, `inst` narrows to exactly `ifc4x3.X` and `TagList` to `tuple[str, ...] | None`.

**2. Automatic Union typing without imports — real positive, real blocker found.** Prototyped a `.pyi` sibling shadowing a runtime class's exported name as `Union[SchemaA.file, SchemaB.file]`. Confirmed with both `ty` and `pyright` independently: bare annotations of that name genuinely behave as the cross-schema Union with zero explicit imports or casts by the caller, exactly what you're after. But both checkers then reject the shadowed name's own constructor call (`Object of type UnionType is not callable`), since a `TypeAlias` to a `Union` isn't a callable class anymore. `ifcopenshell.file(...)` construction happens at 648 call sites across 149 files in this repo, so shadowing the real name this way isn't viable without routing construction through a separate name, a real architectural decision, not a small addition. Happy to prototype that split if you want to go there.

**3. `is_a()`/`isinstance()` — mostly a hard no, with precise reasons, not hand-waving.** Both `is_a()` patterns already type-check with zero errors today (undeclared-attribute fallback types as `Any`), but neither actually narrows `inst`. Tried a `TypeIs` overload on `is_a(ifc_class: Literal[...])`: `ty` rejects it outright (`invalid-type-guard-definition`), because TypeIs/TypeGuard narrows the argument after `self` when the method takes one, i.e. `ifc_class` itself, not `self`. That's a PEP 742 shape limit, not a `ty` quirk, no stub trick routes it back to `self`. `is_a() == "IfcWall"` can never narrow either way, since TypeGuard/TypeIs only fires on a bare predicate call used directly as the tested condition, not on a comparison of its result. For `isinstance(inst, ifc4.IfcWall)`: worth flagging on its own, `ifc4.IfcWall` is stub-only and doesn't exist at runtime today (`ifc4` isn't even an importable module), so this crashes immediately if actually run, yet `ty` accepts it statically with no warning, a real false-safety gap independent of this PR. A small pure-Python demo confirms real per-schema runtime subclasses plus a lookup-driven `wrap()` would make `isinstance()` work correctly with ordinary Python semantics, no metaclass needed, that's the "runtime shim" already flagged as future work in this PR's own description, not new.

PR #8699, opened 2026-07-17
Bonsai: automate wall coverings (cladding, insulation) from wall side faces

### LINKED ISSUE #3754: Automate the placement of IfcCovering on IfcWall and IfcCeiling
There should be the functionality to help automate IfcCovering on IfcWall and IfcSlab. I searched BBIM, ifcopenshell api and Ifcsverchok and haven’t found anything yet. 

For example in RevitAPI and Dynamo I am able to easily extract the boundary lines of Room and use it as footprint for creation and placement wall and ceiling coverings/ finishes. 

Similar automation process should be made available with BBIM and ifcopenshell api or Ifcsverchok. This would be a huge time saver for true BIM modelling.

BIMvoice:
Fixes #3754.

maxfb87 built the floor and ceiling covering automation already merged in this codebase, and named wall covering (cladding, insulation) as the specific remaining gap in that same thread. This adds it, following the existing flooring/ceiling pattern exactly.

Two new operators, `bim.add_instance_wall_covering_from_cursor` and `bim.add_instance_wall_coverings_from_walls`, mirror the existing flooring/ceiling automation. Geometry is net of openings, not a gross rectangle: the wall side face is recovered from the wall's own opening-voided Blender mesh (collecting faces whose normal matches the chosen local side, projecting to wall-local coordinates, unioning with shapely), window holes become inner curves on an `IfcArbitraryProfileDefWithVoids`, and full-height door cuts split into multiple extrusion items.

`covering_poll_relating_type_check` now accepts a tuple of predefined types in addition to a single string, backward compatible with every existing caller.

Tested headless in Blender: net-of-openings geometry verified against a wall with a window and a full-height door opening, both a straight cursor pick and the batch wall-selection path.


maxfb87:
@BIMvoice nice, i tested the tool and it works correctly. Would you like to improve one aspect? I think that it is not necessary to have two operators, just one is needed and the right name is "Add wall coverings from walls".
It's necessary only one operator because the tool works only when a wall is selected and the covering will be put on the wall side with the cursor, so the operator "add wall covering from cursor" is not really needed.
Also, is it possible to add also tests for the new tools created? Thanks

BIMvoice:
@maxfb87 thank you, that is a good catch and you were right on both counts.

I checked before removing it: both operators had the same poll requiring a selected wall, both went through the same underlying function, and the modelling hotkey only ever called the "from walls" one. Running it on a single selected wall gives exactly what the cursor version gave, including which side the covering lands on. So the cursor operator was genuinely redundant and it is gone. What remains is `bim.add_instance_wall_coverings_from_walls`, labelled "Add Wall Coverings From Walls".

Tests added too, at three levels: core, tool, and behavioural scenarios covering the side facing the cursor, the opposite side, a multi-wall selection creating one covering per wall, and the operator refusing cleanly when nothing suitable is selected.

Both pushed to the same PR. Thanks for taking the time to test it.

maxfb87:
@BIMvoice it's all good, just one thing, could you remove the core tests? I'm not confortable adding them right now

BIMvoice:
@maxfb87 done, the core tests are removed. `src/bonsai/test/core/test_covering.py` is gone in 34a5dc2.

The tool tests and the behavioural scenarios are still in, so the operator stays covered. Happy to drop those as well if you would rather, just say.

PR #8704, opened 2026-07-17
ifcpatch: avoid redundant recursive get_info in the Optimise recipe

### LINKED ISSUE #1043: Improved IFC optimiser recipe
Ping @johltn - I read your article here on optimisation of IFCs: http://academy.ifcopenshell.org/optimize-ifc-files/

I previously wrote a very slow IfcPatch recipe that recycled non-rooted elements. I guess it also does optimisation, but is much slower compared to Solibri's results, although if you run it iteratively I did manage to get better than Solibri's results. Here's the code: https://github.com/IfcOpenShell/IfcOpenShell/blob/v0.6.0/src/ifcpatch/recipes/RecycleNonRootedElements.py

Would you be interested in converting your article into an IfcPatch recipe? This would allow users to run the optimisation code via the IfcPatch CLI, GUI, or as a library in their own app for convenience.

BIMvoice:
Fixes #1043.

aothms diagnosed this back in 2020: the Optimise recipe's dedup pass called entity_instance.get_info(recursive=True) on every entity during the toposort walk, which re-expands every already-processed shared subtree from scratch on every parent that references it. The bug was never actually fixed, only cosmetically touched since, confirmed via git history and by profiling the current code: get_info was called 1.94M times for 120K entities on a synthetic file with heavily shared geometry, a 16x amplification from redundant re-expansion.

Since toposort guarantees every entity's dependencies are processed and folded before it is, there is no need to re-expand a referenced entity's subtree at all: its already-computed canonical identity can just be looked up. get_info now runs non-recursively (O(attributes) instead of O(subtree)), and a small map_value helper builds the per-entity comparison key by substituting each reference with the already-folded entity's id rather than walking into it again. Also narrowed the toposort dependency graph to direct references only (max_levels=1), since toposort only needs direct edges, not the full transitive closure.

Real before/after numbers, not estimates:
- test/input/geometrygym_great_court_roof.ifc (56989 entities): 9.9s to 1.7s
- test/input/acad2010_objects.ifc (16296 entities): 3.7s to 0.4s
- A synthetic 120083-entity file with heavily shared geometry, mirroring the issue's scenario: 19.2s to 3.4s

Correctness was verified by comparing the full canonical multiset of every entity between old and new output on all three fixtures: identical fold counts each time, not just a smoke test. Two new tests cover the exact scenario from the issue thread (value-identical vs value-distinct shared subtrees).

**Update:** the toposort backend is now configurable, trying `igraph`'s C-backed `topological_sorting()` first and falling back to the pure-python `toposort` package (with a warning) if `igraph` isn't installed. Real numbers, same machine, same session (supersedes the numbers above, which came from a different run):

| Fixture | before (recursive get_info) | after (pure-python toposort) | after (igraph) |
|---|---|---|---|
| geometrygym_great_court_roof.ifc (56989 entities) | 10.2s | 1.7s | 1.6s |
| acad2010_objects.ifc (16296 entities) | 2.0s | 0.36s | 0.33s |
| synthetic 120963-entity file, heavily shared geometry | 4.5s | 1.7s | 1.6s |

Isolated to just the toposort step, igraph is 3-4.5x faster than the pure-python package, but since toposort is now a small fraction of total runtime after the non-recursive get_info fix, this only translates to a 5-10% additional overall speedup. igraph is a nice-to-have on top of the main fix, not a further dramatic win.

Also benchmarked against `RecycleNonRootedElements` per your question: on both real-world fixtures it's 2.7-4.5x faster than `Optimise` and produces essentially the same output entity count (48213 vs 48213 on great_court_roof; 14086 vs 13984 on acad2010), so it's the better choice for typical files, matching this recipe's own existing docstring recommendation. `Optimise`'s extra transitive fold only pays off on files with deep, repeated shared subtrees like the synthetic fixture above, where it achieves roughly 2x better deduplication (7803 vs 15003 output entities) at ~1.6x the runtime cost.

### DIFF
--- a/src/ifcpatch/ifcpatch/recipes/Optimise.py
+++ b/src/ifcpatch/ifcpatch/recipes/Optimise.py
@@ -36,8 +36,10 @@ class Patcher:
         can usually be solved through other means. Consult the bonsai Add-on
         documentation on dealing with large models for more details.
 
-        Warning: this optimise recipe is very, very slow. Please consider using
-        RecycleNonRootedElements instead.
+        Warning: this optimise recipe is slower than RecycleNonRootedElements,
+        as it performs a full, transitive fold instead of a single pass.
+        Consider RecycleNonRootedElements first if a quicker, partial
+        optimisation is acceptable.
 
         Example:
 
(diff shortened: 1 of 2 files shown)


aothms:
Great. Can you also make the toposort backend configurable, first try igraph, fallback to the pure python toposort with warning. Assuming that will shave off some additional time. So also rerun the numbers. Also measure against RecycleNonRootedElements and adjust the performance claim in the top if it's no longer accurate.

BIMvoice:
@aothms done, and updated the PR description with the new numbers. igraph backend added with a warning-logged fallback to the pure-python toposort package. Since toposort was already a small fraction of runtime after the non-recursive get_info fix, igraph only adds a modest 5-10% further speedup on top of it, not a dramatic one.

On RecycleNonRootedElements: it's 2.7-4.5x faster than Optimise on both real-world fixtures and produces essentially the same output entity count, so it's the better default for typical files (matching this recipe's own docstring). Optimise's extra transitive fold specifically pays off on files with deep, repeated shared subtrees, where it still gets ~2x better deduplication at ~1.6x the runtime cost. Full numbers in the PR description.

Issue #8700, opened 2026-07-17
ifcpatch: MergeProjects leaves duplicate Model/Body/MODEL_VIEW contexts, collapsing layered-wall material styles


theoryshaw:
### Bug Description

> **Correction:** an earlier version of this report attributed the duplicate contexts to the clipboard copy/paste engine. Further investigation traced the actual cause to the **`MergeProjects` ifcpatch recipe**. The description below has been rewritten accordingly.

Merging two IFC models with the `MergeProjects` ifcpatch recipe can leave the result with **two (or more) identical `Model/Body/MODEL_VIEW` geometric representation contexts** (and other duplicated contexts). This is invalid IFC — a context should be unique per `(ContextType, ContextIdentifier, TargetView)` — and it silently breaks anything that resolves a context by its attributes.

The most visible symptom: a layered wall whose material layers each carry their own surface style renders in the 3D viewport with a **single style for the whole wall**. `bonsai.tool.Loader.slice_layerset_mesh()` resolves each layer's style via `ifcopenshell.util.representation.get_material_style(layer.Material, body)`, where `body = get_context(ifc, "Model", "Body", "MODEL_VIEW")`. `get_context()` returns only the *first* matching context, but the material styles' `IfcStyledRepresentation`s hang off the *second* Body context, so `get_material_style` returns `None` for every layer and the whole wall falls back to one style.

**Root cause.** `MergeProjects.reuse_existing_contexts()` repoints each merged-in context that duplicates an existing one and removes it via `remove_deep2` over an **unordered `set`**. Per its contract, `remove_deep2` **silently returns without deleting** any element that still has an inverse, and the recipe's bare `except: pass` hides the rest. Depending on the in-memory entity layout, the repoint loop leaves a **residual inverse** on a matched context, so `remove_deep2` skips it and it survives as an orphaned duplicate.

Captured trace (merging two small test files in Bonsai) — every survivor is a context that still held an inverse at removal time:

```
[MergeProjects] === removal loop (order-sensitive) ===
  #142 Model/Body    inv_before=1  refs=[IfcShapeRepresentation]              -> SURVIVED
  #150 Plan/Axis     inv_before=1  refs=[IfcShapeRepresentation]              -> SURVIVED
  #123 Model parent  inv_before=8  refs=[IfcGeometricRepresentationSubContext]-> SURVIVED
[...]

PR #8706, opened 2026-07-17
ifcpatch: add MergeDuplicateContexts recipe

### LINKED ISSUE #8701: IfcPatch: add MergeDuplicateContexts recipe to consolidate duplicate representation contexts
### Feature Description

Add a `MergeDuplicateContexts` ifcpatch recipe (and expose it in Bonsai's IFC Patch UI) that consolidates duplicate `IfcGeometricRepresentationContext` / `IfcGeometricRepresentationSubContext` entities.

**Problem it solves.** A well-formed IFC has a single geometric representation context per `(ContextType, ContextIdentifier, TargetView)` combination (e.g. one `Model/Body/MODEL_VIEW`). Some authoring tools, merges, round-trips, and Bonsai's own clipboard paste (see the related bug report #8700) leave a model with two or more contexts that are identical in all three. This is non-conformant and breaks code that looks a context up by attributes: `ifcopenshell.util.representation.get_context()` returns only the first match, so geometry, material styles, or annotations attached to the others are silently orphaned. A common visible symptom is a layered wall rendering with a single material style because each layer's style hangs off a second, never-matched, Body context.

Today there is no user-facing tool to repair such a file — a user has to hand-edit the IFC or run a bespoke script.

**Proposed behaviour.** For each `(ContextType, ContextIdentifier, TargetView)` key, keep the first context, repoint every reference off the duplicates onto that survivor (via `ifcopenshell.util.element.replace_element`), remove the duplicates, and dedupe any SET-typed aggregate (e.g. `IfcProject.RepresentationContexts`) that would otherwise list the survivor twice. Top-level contexts and subcontexts are handled in separate passes so a subcontext is never merged into a parent context.

Usage:

```python
ifcpatch.execute({"file": model, "recipe": "MergeDuplicateContexts", "arguments": []})
```

It takes no arguments, so it appears in the Bonsai IFC Patch dropdown with no parameters to configure. This mirrors the shape of the existing `MergeDuplicateTypes` / `MergeStyles` recipes.

I have a working implementation and tests, and can open a PR.

theoryshaw:
Adds a `MergeDuplicateContexts` ifcpatch recipe that consolidates duplicate `IfcGeometricRepresentationContext` / `IfcGeometricRepresentationSubContext` entities.

A well-formed IFC has a single context per `(ContextType, ContextIdentifier, TargetView)` (e.g. one `Model/Body/MODEL_VIEW`). Merges, round-trips, and some exporters leave two or more that are identical in all three. This is non-conformant and breaks code that resolves a context by attributes — `ifcopenshell.util.representation.get_context()` returns only the first match, silently orphaning geometry, material styles, or annotations attached to the others. A common symptom is a layered wall rendering with a single material style because each layer's style hangs off a second, never-matched, Body context.

The recipe keeps the first context per key, repoints every reference off the duplicates onto that survivor (`ifcopenshell.util.element.replace_element`), removes the duplicates, and dedupes any SET-typed aggregate (e.g. `IfcProject.RepresentationContexts`) that would otherwise list the survivor twice. Top-level contexts and subcontexts are handled in separate passes so a subcontext is never merged into a parent context.

Takes no arguments, so it appears in the Bonsai IFC Patch dropdown with nothing to configure — mirroring `MergeDuplicateTypes` / `MergeStyles`.

Includes `test_MergeDuplicateContexts.py` (IFC4 + IFC2X3): collapse + reference repoint + project-aggregate dedup, and a no-op case.

Closes #8701. Also the cleanup half of #8700.

### DIFF
--- a/docs/dev-notes/merge-duplicate-contexts.md
+++ b/docs/dev-notes/merge-duplicate-contexts.md
@@ -29,9 +29,12 @@ tool to repair such a file.
 For each `(ContextType, ContextIdentifier, TargetView)` key: keep the first context,
 repoint every reference off the duplicates onto that survivor via
 `ifcopenshell.util.element.replace_element` (generic — handles representations'
-`ContextOfItems`, subcontexts' `ParentContext`, coordinate operations, etc.), remove the
-duplicates, and dedupe any SET-typed aggregate (e.g. `IfcProject.RepresentationContexts`)
-that would otherwise list the survivor twice.
+`ContextOfItems`, subcontexts' `ParentContext`, coordinate operations, etc.), then remove
+the duplicates. `replace_element`/`replace_attribute` deduplicate SET-typed attributes
+themselves (e.g. `IfcProject.RepresentationContexts`), so the survivor is never listed
+twice — this recipe no longer needs its own SET-dedupe pass (dropped in review; see the
+core fix that added it: `ifcopenshell.util.element: dedupe SET-typed attributes in
+replace_attribute`).
 
 Top-level contexts and subcontexts are keyed and processed in **separate passes** so a
 subcontext is never merged into a parent context, and a surviving subcontext keeps a
--- a/docs/dev-notes/merge-duplicate-contexts.md
+++ b/docs/dev-notes/merge-duplicate-contexts.md
@@ -1,0 +1,50 @@
+<!-- This file was generated with the assistance of an AI coding tool. -->
+
+# MergeDuplicateContexts recipe
+
+> **Living dev note** for the `merge-duplicate-contexts` branch/PR. Read before working
+> on the feature; append decisions and findings as the PR is refined. This is *not* user
+> documentation — at merge it is removed or its durable parts promoted to code comments.
+> See [README.md](README.md) for the convention.
+
+PR #8706 → closes #8701. Also the cleanup half of #8700 (the root-cause fix lives on the
+stacked `fix-mergeprojects-dup-contexts` branch, PR #8707, whose dev-note has the full
+investigation).
+
+## Problem
+
+A well-formed IFC has a single geometric representation context per
+`(ContextType, ContextIdentifier, TargetView)` (e.g. one `Model/Body/MODEL_VIEW`).
+Merges, round-trips, and some exporters — and, concretely, the `MergeProjects` recipe
+(see #8700) — leave two or more that are identical in all three. This is non-conformant
+and breaks code that resolves a context by attributes:
+`ifcopenshell.util.representation.get_context()` returns only the **first** match, so
+geometry, material styles, or annotations attached to the others are silently orphaned.
+Visible symptom: a layered wall renders with a single material style because each
+layer's style hangs off a second, never-matched, Body context. There was no user-facing
+tool to repair such a file.
+
+## Design
+
+For each `(ContextType, ContextIdentifier, TargetView)` key: keep the first context,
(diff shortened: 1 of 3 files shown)


Moult:
Good patch, but fixing the issue with duplicate references in a set after `replace_element` should be fixed at a more fundamental level (with due consideration to performance as checking each value can be expensive). Can you please propose a more fundamental fix (happy to be in a separate PR) which you can link to this.

BIMvoice:
@theoryshaw I looked at #8700 and found you already have the fix here (this PR plus the stacked #8707), so I verified it rather than duplicating anything. Using the deterministic-duplicate proxy from your own regression test, I confirmed the reported symptom on stock v0.8.0 (two Model/Body/MODEL_VIEW contexts, 0 of 3 layered-wall material styles resolving via get_context + get_material_style) and that your branch fixes it exactly (one context, 3 of 3 styles resolved). Solid fix.

The only thing keeping the lint check red looks trivial: MergeDuplicateContexts.py line 20 has an unused import, Union in `from typing import Any, Union`, which fails both ruff and black. Dropping the unused name should turn lint green. The compile-and-test failures on that run are pre-existing and unrelated (rocksdb streaming, wall_opening, bsdd 429 rate limit), not from this change. Happy to help however is useful.

theoryshaw:
Thanks both — addressed:

- **@Moult** — agreed, and the fundamental fix has since landed upstream as `8c434b7167` ("ifcopenshell.util.element: dedupe SET-typed attributes in replace_attribute"), which is the right level (and nicely cached for the perf concern you raised). I've rebased this PR onto current `v0.8.0` and dropped the recipe's own `dedupe_references` pass, so it now relies on `replace_element` / `replace_attribute` deduplicating SET-typed attributes (e.g. `IfcProject.RepresentationContexts`) themselves. Verified the recipe still collapses duplicates and leaves the project's context set with a single entry.
- **@BIMvoice** — thanks for verifying against #8700 and for the lint catch; the unused `Union` import is removed. The unrelated red checks you noted (rocksdb streaming, wall_opening, bsdd rate-limit) are pre-existing.

The stacked #8707 was rebased on top of this as well.

BIMvoice:
Two things after looking at this more closely.

First, confirming the part you asked about: `IfcProject.RepresentationContexts` is declared SET in IFC2X3, IFC4 and IFC4X3_ADD2, so `_is_set_attribute` returns True and `replace_attribute` deduplicates it. Dropping `dedupe_references` is safe for that path. `ContextOfItems` and `ParentContext` are single references so they need no dedupe. Coverage looks complete.

Second, a question rather than an objection, since you may have already considered it. `get_key` uses `ContextType`, `ContextIdentifier` and `TargetView`. `IfcGeometricRepresentationContext` also carries `CoordinateSpaceDimension`, `Precision`, `WorldCoordinateSystem` and `TrueNorth`, and the subcontext adds `ParentContext`, `TargetScale` and `UserDefinedTargetView`.

So two contexts that share an identifier but differ in `WorldCoordinateSystem` or `TrueNorth` currently group as duplicates, and merging them would silently relocate or reorient every representation in the losing context. Same for subcontexts under different `ParentContext`.

For MergeProjects output that is presumably safe, since those duplicates are genuine copies. But the recipe is general, and a georeferenced model with two contexts at different world placements is not far-fetched.

Would it be worth either widening the key to include `WorldCoordinateSystem` and `TrueNorth` (and `ParentContext` for subcontexts), or asserting they match and skipping the group with a warning if they do not?

PR #8718, opened 2026-07-18
Bonsai: apply parametric representation edits across the whole selection

### LINKED ISSUE #2934: Have bim.update_parametric_representation work on multiple selections at one time
Would be nice to have bim.update_parametric_representation work on multiple selections at one time. 
Since 'tab' (to edit vertices) doesn't work anymore to modify multiple extrusions at one time. 

2023-04-04_17-17-26_Blender_blender

BIMvoice:
Fixes #2934.

UpdateParametricRepresentation only ever read context.active_object and wrote the edited value back to that single object's IFC entity. The parameter list is built per-object by traversing that object's own representation graph, so there was no shared identity for the same parameter across multiple selected objects, which is why a multi-object edit never propagated.

When other objects are selected alongside the active one, the same edited parameter is now matched on the others by qualified name (e.g. IfcExtrudedAreaSolid/Depth) and occurrence position, for objects whose own parameter list has more than one parameter of that name. Each other object's parameter list is refreshed from its current IFC state before matching, then the value is applied and its representation switched, same as the active object. Objects with no matching parameter (a different profile type, or nothing parametric at all) are silently skipped, and the report states how many of the selected objects were actually updated.

Tested live in headless Blender: two duplicated box extrusions selected together, editing Depth on the active one propagates Depth to both while leaving XDim/YDim untouched on either. A heterogeneous selection (rectangle extrusion, circle extrusion, plain tessellation) correctly updates the two extrusions that share the edited parameter's name, skips the tessellated object without erroring, and reports "3 of 4."

### DIFF
--- a/src/bonsai/bonsai/bim/module/geometry/operator.py
+++ b/src/bonsai/bonsai/bim/module/geometry/operator.py
@@ -718,6 +718,11 @@ class UpdateRepresentation(bpy.types.Operator, tool.Ifc.Operator):
 class UpdateParametricRepresentation(bpy.types.Operator):
     bl_idname = "bim.update_parametric_representation"
     bl_label = "Update Parametric Representation"
+    bl_description = (
+        "Apply the edited parameter to the active object.\n"
+        "If other objects are selected, the same named parameter is also applied to those "
+        "of them that have a matching parameter"
+    )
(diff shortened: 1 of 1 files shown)


theoryshaw:
works from my side. thank you.

PR #8719, opened 2026-07-18
Bonsai: add one click copy of annotations to another drawing

### LINKED ISSUE #2966: BlenderBim - copying annotations from one drawing to another does not show
model drawing 1:
image
svg:
image

model drawing 2:
image
svg:
image

Text shown copied in outliner:
image

After saving and reloading the copied text is not saved

BIMvoice:
Fixes #2966.

The originally reported symptom (an annotation copied via Blender's native Copy/Paste between drawings doesn't survive reload) is Blender's own Object duplication keeping the pasted object linked to the same underlying IFC entity, so no new entity is ever created, only one Blender object is ever recognized after any map rebuild, and there was only ever one entity to begin with. Blender's clipboard can't be taught IFC semantics, so this isn't fixable at that layer.

Two real improvements instead, building on existing correct primitives already in this codebase (geometry.duplicate_ifc_objects for real entity duplication, the same duplicate-then-regroup recipe core.duplicate_drawing already uses for its own case):

Reassigning an annotation's group to a different drawing (bim.assign_group) now automatically moves it into the target drawing's collection and repositions it onto the target drawing's camera depth, instead of leaving it stranded, per the intermediate step theoryshaw proposed in the thread and Plae-Blue-Dot agreed would help.

A new one click bim.copy_annotation_to_drawing operator (button in the annotation tool sidebar and the Drawings panel) does the real thing several people in the thread asked for: duplicate the annotation into a genuinely new, independent IfcAnnotation entity assigned to a chosen target drawing, correctly positioned, while leaving the original untouched in its own drawing.

Tested live in headless Blender: created two drawings, added a TEXT annotation to drawing 1, ran the new operator. Confirmed the original entity and object are untouched in drawing 1, the copy is a genuinely new IfcAnnotation with its own GlobalId and geometry, correctly assigned to drawing 2's group and collection, positioned on drawing 2's camera depth plane. Edited the copy's text afterward without affecting the original. Full save plus reload round trip loads both annotations independently in their own drawings with correct, separate text. New core and tool test coverage added, full existing test suites (test/core, test/tool/test_drawing.py) pass alongside the new tests.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/__init__.py
+++ b/src/bonsai/bonsai/bim/module/drawing/__init__.py
@@ -47,6 +47,7 @@ classes = (
     operator.CleanWireframes,
     operator.ContractSheet,
     operator.ConvertSVGToDXF,
+    operator.CopyAnnotationToDrawing,
     operator.CopyTextToSelection,
     operator.CreateDrawing,
     operator.CreateSheets,
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -228,6 +228,90 @@ class DuplicateDrawing(bpy.types.Operator, tool.Ifc.Operator):
         )
 
 
+def get_copy_annotation_target_drawings(self, context):
+    global COPY_ANNOTATION_TARGET_DRAWINGS_ENUM
+    drawings = [e for e in tool.Ifc.get().by_type("IfcAnnotation") if e.ObjectType == "DRAWING"]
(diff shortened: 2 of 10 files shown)


theoryshaw:
Hi, Just a heads up — this PR overlaps with #8083 (*Parametric dimensions*), and on top of `v0.8.0` the two **don't auto-merge** (verified with `git merge-tree`). It's not a trivial overlap: the merge leaves **1 conflict region / ~9 lines** to resolve by hand in `bim/module/drawing/svgwriter.py`.

PR #8720, opened 2026-07-18
Bonsai: compose element visibility with the Status filter instead of overwriting it

### LINKED ISSUE #6664: Elements Isolate/Show/Hide and Status filter do not appear to respect each other.

BIMvoice:
Fixes #6664.

Root cause, credit to @sboddy for the original diagnosis: two independent code paths each call obj.hide_set() unconditionally, as if it owned the object's visibility outright. tool.Sequence.set_visibility_by_status iterates every IfcProduct and sets hide_set based on the current Status filter, ignoring anything else. SetElementVisibility.execute (the renamed SetContainerVisibility sboddy identified) sets hide_set on whatever the Elements list currently shows, ignoring the Status filter entirely. Whichever one runs last wins and silently wipes out the other.

Reproduced live in headless Blender with a two storey model carrying EXISTING/DEMOLISH/NEW walls: Alt+Click Show on a class resurrected status hidden DEMOLISH walls, isolating a floor ignored the active status filter, and reactivating the status filter reset manually hidden walls back to visible, matching every symptom in the report.

Fix: each object now tracks two independent, persisted flags (is_manually_hidden, is_hidden_by_status on BIMObjectProperties) instead of treating Blender's single hide flag as ground truth. Two new tool.Blender helpers always compute the actual hide_set() as the OR of both flags, so setting one preserves whatever the other mechanism already decided. set_visibility_by_status now writes through the status half, SetElementVisibility's Hide/Show/Isolate now writes through the manual half. Isolate mode was also rewritten to explicitly iterate IfcProduct elements instead of relying on Blender's hide_view_set, which incidentally affected every scene object, cameras and lights included, not just IFC elements.

Re-verified live after the fix: Show/Isolate no longer resurrect status hidden elements in either direction, and toggling the Status filter no longer wipes out manual hides.

Out of scope, left untouched: the opt-in collection-level container visibility preference operates on a different axis (collection.hide_viewport) and was not part of the reported conflict. A couple of other unrelated call sites (search/query panel, space toggle) also call hide_set directly and could theoretically have similar composition gaps, but they're outside what was reported here.

### DIFF
--- a/src/bonsai/bonsai/bim/module/spatial/operator.py
+++ b/src/bonsai/bonsai/bim/module/spatial/operator.py
@@ -564,11 +564,18 @@ class SetElementVisibility(bpy.types.Operator):
         filtered_elements = set(tool.Spatial.get_filtered_elements(self.should_filter))
 
         if self.mode == "ISOLATE":
+            # Bulk-hide the whole view layer natively rather than per-object hide_set().
+            context_override = tool.Blender.get_viewport_context()
+            with context.temp_override(**context_override):
+                bpy.ops.object.hide_view_set(unselected=True)
+                bpy.ops.object.hide_view_set(unselected=False)
+
+            # Record is_manually_hidden so a later Status filter change still composes (#6664).
             for element in tool.Ifc.get().by_type("IfcProduct"):
                 if element in filtered_elements:
                     continue
                 if obj := tool.Ifc.get_object(element):
-                    tool.Blender.set_object_manual_visibility(obj, True)
+                    tool.Blender.get_object_bim_props(obj).is_manually_hidden = True
             should_hide = False
         else:
             should_hide = self.mode == "HIDE"
--- a/src/bonsai/bonsai/bim/module/spatial/operator.py
+++ b/src/bonsai/bonsai/bim/module/spatial/operator.py
@@ -561,18 +561,21 @@ class SetElementVisibility(bpy.types.Operator):
         return self.execute(context)
 
     def execute(self, context):
+        filtered_elements = set(tool.Spatial.get_filtered_elements(self.should_filter))
+
         if self.mode == "ISOLATE":
-            context_override = tool.Blender.get_viewport_context()
-            with context.temp_override(**context_override):
-                bpy.ops.object.hide_view_set(unselected=True)
-                bpy.ops.object.hide_view_set(unselected=False)
+            for element in tool.Ifc.get().by_type("IfcProduct"):
+                if element in filtered_elements:
+                    continue
+                if obj := tool.Ifc.get_object(element):
+                    tool.Blender.set_object_manual_visibility(obj, True)
             should_hide = False
         else:
             should_hide = self.mode == "HIDE"
 
-        for element in tool.Spatial.get_filtered_elements(self.should_filter):
+        for element in filtered_elements:
             if obj := tool.Ifc.get_object(element):
-                obj.hide_set(should_hide)
+                tool.Blender.set_object_manual_visibility(obj, should_hide)
                 for collection in obj.users_collection:
                     collection.hide_viewport = False
         return {"FINISHED"}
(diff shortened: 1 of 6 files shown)


BIMvoice:
@sboddy took you up on the suggestion: built a heavy-model generator + a before/after benchmark harness and ran it against this PR's actual diff (merge-base vs branch tip, same machine/session, 7 timed repeats each).

Confirms your concern, and quantifies it:

| Scenario | Elements | Before (median) | After (median) | Delta |
|---|---|---|---|---|
| Isolate 1 storey | 4172 | 0.162s | 0.431s | +166% |
| Isolate 1 storey | 8138 | 0.382s | 1.481s | +288% |
| Isolate whole building | 4172 | 0.671s | 0.858s | +28% |
| Isolate whole building | 8138 | 2.564s | 2.928s | +14% |
| Hide 1 storey | 4172 | 0.161s | 0.156s | ~0% |
| Hide 1 storey | 8138 | 0.376s | 0.400s | +6% |

Isolating a small subset of a bigger model gets worse as the model grows, since the fix's per-element Python loop scans every `IfcProduct` in the file rather than just the filtered set. Isolating nearly everything, or plain Hide, are only mildly affected since they don't hit that new full-file scan as hard.

Will look at scoping the loop to the actually-affected elements before this merges rather than shipping the regression alongside the correctness fix.

BIMvoice:
@sboddy pushed a fix for this. Root cause: the Isolate loop replaced two native `hide_view_set()` bulk calls with a Python loop calling `hide_set()` once per `IfcProduct` in the whole file, and `hide_set()` itself turned out to be the expensive part (not O(1) per call). Restored the native bulk calls for the "hide everything else" part, and kept only a cheap per-object property write for the bookkeeping this PR actually needs (so the Status filter still composes correctly with Isolate).

New numbers, same harness, same sizes:

| Scenario | Elements | Before (median) | Fixed (median) | Delta |
|---|---|---|---|---|
| Isolate 1 storey | 4172 | 0.150s | 0.150s | ~0% |
| Isolate 1 storey | 8138 | 0.343s | 0.362s | +5% |

Down from +166%/+288%. Added a test that pins the exact call count for Isolate (3 objects touched for a 3-object storey out of 8 total in the file) so a regression back to a whole-file scan fails deterministically rather than only showing up as a slowdown.

BIMvoice:
@sboddy Appreciated. Agreed it's a real tradeoff worth a maintainer's judgment call, not ours to make; the numbers are there either way if it helps the decision.

Issue #7815, opened 2026-03-18
Void not voiding at certain positions.


theoryshaw:
### Bug Description

This one's a weird one. 

GlobalId = "1V7WEiSe528uQTdG9dxdoe"

[3_Season_Porch_URGC - Copy.ifc.txt](https://github.com/user-attachments/files/26098426/3_Season_Porch_URGC.-.Copy.ifc.txt)

Per the following video, if you move the void over slightly, it doesn't void anymore. 

https://github.com/user-attachments/assets/839c7733-5cc9-411c-adf9-22e9a750c1fb

### Debug and Error Output

```true

```

BIMvoice:
This is an OpenCascade boolean robustness failure on valid input, not a Bonsai placement or boolean application bug. Bonsai places the opening correctly, the IfcRelVoidsElement and placement are both right. Sweeping the opening in X shows a sharp discontinuity, a 0.6 inch move flips voiding on and off, matching the report.

At a failing position the opencascade kernel returns the roof uncut (volume unchanged) while reporting IsDone, because the section curve computation between the vertical cutter and the roof shell produced no intersection edges near a cluster of near coincident hip edges, and that uncut result passes every post boolean validity and interference check, so no higher fuzziness retry fires and the no op cut is accepted silently. Raising precision and fuzziness up to 1 cm does not recover it. The exact same operands through the CGAL kernel cut correctly, which proves the input is valid and the failure is specific to OCCT.

Workaround: export or render this model with geometry_library set to cgal. The proper fix is for the hybrid kernel to detect this silent no op cut, a result geometrically identical to the first operand while the operands provably overlap, and fall back to CGAL.

BIMvoice:
@theoryshaw confirmed your diagnosis and root caused it further: OCCT's BRepAlgoAPI_Cut reports success on a cut that is actually a complete no-op (identical volume to the roof with the opening disabled), because the section curve computation fails to find intersection edges near a cluster of coincident hip edges, and that no-op passes every existing validity check. Fixed by detecting when a cut's volume is unchanged despite operands that provably overlap, and recovering via the CGAL fallback from #8722. Verified against your file: the opening position sweep is now smooth and continuous, matching the CGAL kernel's output.

This depends on #8722 landing first.

PR: #8725

PR #8722, opened 2026-07-18
ifcgeom: fall back to CGAL when OCCT boolean produces coincident faces

### LINKED ISSUE #6867: Strange failure with mitred walls and multiple half-space cuts
### Bug Description

I was doing some cleanup work on my main file which may (or may not) have triggered this to happen, or I may just not have noticed an update that changed how this is calculated. The walls _did_ previously look "correct", but now they don't.

I had to play about a bit to understand how to easily trigger this. I include a trivial example file that shows the issue, saved _before_ step 5:
1. Make a two section wall with a mitred join,
2. Add 3 half space planes:
Image
3. When you exit out of Item mode, into Object mode:
Image
4. If you look at it in x-ray mode, you can see it is trying to make the cut, but _something_ goes wrong:
Image
5. If you unjoin (Shift-U) the two walls:
Image

Additionally:
- It doesn't matter if the right hand half-space cuts through the mitre join or not. In fact, if you move that down to avoid the join, the breakage gets worse.
- It doesn't always take three half-spaces, I've seen it with two, but this example was very reliable at reproducing it.
- This is _not_ caused by https://github.com/IfcOpenShell/IfcOpenShell/commit/6c4e4e5f910a7959175f329cc159eb65f7b0f16a because I'm running a release from yesterday.
- Also saw the effect on a particular IfcBeam that had three half-spaces.

[half_space_mitre_error.ifc.txt](https://github.com/user-attachments/files/21006708/half_space_mitre_error.ifc.txt)

### Attachments

_No response_

### Debug and Error Output

```true
os: Linux
os_version: #152-Ubuntu SMP Mon May 19 10:54:31 UTC 2025
python_version: 3.11.11
machine: x86_64
processor: x86_64
blender_version: 4.4.3
bonsai_version: 0.8.3-alpha250630
bonsai_commit_hash: 260bc80
bonsai_commit_date: 2025-06-30T10:11:40+02:00
last_actions: 
# ifcopenshell.api: geometry.disconnect_path
>>> {"element": {"cast_type": "entity_instance", "value": 2854, "Name": "Wall"}, "connection_type": "ATSTART"}
# ifcopenshell.api: geometry.disconnect_path
>>> {"element": {"cast_type": "entity_instance", "value": 2854, "Name": "Wall"}, "connection_type": "ATEND"}
# ifcopenshell.api: geometry.regenerate_wall_representation
>>> {}
# operator: BIM_OT_override_mode_set_edit
# operator: BIM_OT_override_mode_set_edit
# operator: BIM_OT_import_representation_items
# operator: BIM_OT_override_move_select
# operator: BIM_OT_override_mode_set_edit
# operator: BIM_OT_override_mode_set_edit
# ifcopenshell.api: geometry.validate_type
>>> {}
# ifcopenshell.api: geometry.add_boolean
>>> {}
# operator: BIM_OT_override_mode_set_edit
# operator: BIM_OT_override_mode_set_edit
# operator: BIM_OT_import_representation_items
# operator: BIM_OT_override_move_select
# operator: BIM_OT_override_move_select
# operator: BIM_OT_override_mode_set_edit
# operator: BIM_OT_override_mode_set_edit
# ifcopenshell.api: geometry.validate_type
>>> {}
# ifcopenshell.api: geometry.add_boolean
>>> {}
last_error: None
ifc: Unsaved
schema: IFC4
preprocessor_version: IfcOpenShell 0.8.3-alpha250630
originating_system: Bonsai 0.8.3-alpha250630-260bc80
```

BIMvoice:
## Root cause
Cutting a mitred wall extrusion (slanted end face) with multiple half-space
solids near the bevel makes OCCT emit a coincident duplicate face at the
mitre. The existing `is_manifold()` check only counts faces per individual
`TopoDS_Edge` instance, so two topologically distinct edges sitting at the
exact same location, each paired with their own 2 faces, both pass the check
even though 4 faces actually meet at that line.

## Fix
- New `has_coincident_edges()` groups edges by rounded endpoint position
  (not edge identity) and flags any location where more than 2 faces meet.
- When a boolean result trips this check, retry the same operation through
  the exact-arithmetic CGAL kernel and use that result if it comes back
  clean, before falling through to the existing exemption heuristic.
- New `boolean_operation_cgal_fallback()` builds a CGAL Nef polyhedron
  directly from the OCCT shape's planar faces (`cgal_face_t` has internal
  linkage per translation unit, so the existing CGAL kernel's helper
  couldn't be reused directly). Scoped to planar, single-wire faces;
  anything else falls through unchanged.

## Verification
- Reproduced the reported failure with the attached IFC file: 5 coincident
  edges on the cut wall with unmodified code, 0 with the fix (matches plain
  `--kernel cgal` output exactly).
- Regression swept all 89 `test/input` fixtures exercising
  `IfcBooleanClippingResult`/`IfcBooleanResult`/`IfcHalfSpaceSolid`: 84
  byte-identical to baseline, 4 pre-existing fixtures independently cleaned
  of the same defect (matching volumes), 1 false-positive trigger caught
  during development and fixed with a sewing tolerance + healing pass.

Fixes #6867


aothms:
Sorry, but this is really weird coupling to implement a cgal fallback inside the occt kernel library.

Instead, let's make this part of the hybrid kernel chaining.

Let's say we have opencascade - cgal-simple - cgal as a chain.

Somehow we need to flag that the opencascade result is suspicious - maybe just use a Logger::Warning message for that. Hybrid kernel will then evaluate cgal-simple (no-support for bools) and then cgal resulting in a non-suspicious result. Result: prefer the cgal outcome.

Note that this involves convert() and convert_openings() so the cleanest is a NativeElement -> taxonomy::shell round trip so that if convert() succeeds we do not need to re-execute that to feed it into the convert_openings() call.

BIMvoice:
Reworked per your feedback. The opencascade kernel no longer depends on CGAL at all (confirmed zero CGAL:: symbols in that library). has_coincident_edges() now only logs a Logger::Warning and flags the ConversionResult as suspicious. HybridKernel::convert() and convert_openings() try every kernel in the chain when the current result is suspicious and prefer the first clean one, matching the opencascade to cgal-simple to cgal chain you described. This also fixes the #7815 gap where convert_impl always returned true regardless of the internal boolean's success, so a suspicious result never had a chance to reach the chaining logic.

Verified against the original repro with the actual hybrid-opencascade-cgal-simple-cgal kernel name: the mitred wall's coincident edges are gone and the chain correctly prefers CGAL's clean result. Regression swept the same 90 boolean-related test fixtures: byte-identical for plain opencascade, and only 3 out of 89 differ under the hybrid chain, 2 of which are tessellation-only noise.

One real finding I want to flag rather than paper over: on 599--beam--segfault.ifc, OCCT's own result gets flagged suspicious even though its tessellation has 0 non-manifold edges and a higher volume, while CGAL's result (which wins because it is not self-flagged) actually has 99 non-manifold edges and 12% less volume, confirmed identical to running plain --kernel cgal standalone, so it is an existing CGAL limitation on this fixture, not something the chaining introduces. The chain currently trusts whichever kernel is not self-flagged without cross-checking it, since that was not part of the design you described. Would appreciate your read on whether that is an acceptable tradeoff for this fixture specifically, or whether a cross-check belongs in this chain.

Did not implement the full NativeElement to taxonomy::shell round trip for convert()/convert_openings() coherency. Happy to take a pass at that if you think it is worth doing now rather than as a follow-up.

aothms:
> has_coincident_edges() now only logs a Logger::Warning and flags the ConversionResult as suspicious. HybridKernel::convert() 

Instead I'd rather have the fact that geometry volume is unaltered as suspicious, not some other heuristic.

> One real finding I want to flag rather than paper over ...

With the fix above, I have full confidence in CGAL as a fallback when OpenCASCADE reports no volume decrease.

aothms:
I also don't like the suspicious flag. Please stick to the logger capture. Providing a new logger to the call with in memory capture. Detect the message. Append/propagate temp logger output to the higher level logger when ready.

BIMvoice:
Reworked per your latest comment. Each hybrid chain attempt now gets its own temporary, in-memory `Logger` (bound once per kernel via the existing `clone(Logger&)` mechanism, so per-kernel shape caches still persist across elements). Before an attempt it's cleared; after, its captured messages are scanned for the coincident-faces warning code. A rejected attempt's log is simply never appended anywhere, so it never surfaces; an accepted attempt's log is folded into the real logger via `Logger::Append()`. The `suspicious` flag/out-param is gone entirely, along with `ConversionResult::isSuspicious()`/`setSuspicious()`.

This builds directly on top of the `boolean_settings.logger`/`.log()` field your logger-routing fix (#8554) already added, rather than threading a second parallel logger parameter through the same call chain.

One thing I want to flag rather than quietly override: I kept `has_coincident_edges()` as the detection signal instead of switching to "volume unaltered." Reproducing #6867, the buggy OCCT result's volume actually drops from 5.34 to 3.72 (a real ~30% decrease) despite the coincident-faces defect, so a volume check alone would not have caught this bug. Happy to add a volume check as a second signal alongside it if you still want it, but I didn't want to swap out the one signal that's actually verified to catch the reported case.

Also caught a real bug while building this: `Logger::Append()` only replayed a captured in-memory logger's messages when the *destination* was also in-memory. Under normal CLI usage (plain-text output), folding an accepted attempt's messages into the real logger would have silently produced nothing. Fixed so it renders through any destination format, respecting that destination's own verbosity.

Regression-swept the same 90 fixtures from the original PR: plain `--kernel opencascade` output is byte-identical throughout, and the hybrid chain differs only by vertex-ordering noise (same vertex/face sets) on 2 fixtures already known to be tessellation-nondeterministic.

PR #8724, opened 2026-07-18
ifcgeom: avoid degenerate cross product in SvgSerializer for vertical section-line curves

### LINKED ISSUE #8049: Geometry that fails during printing.
Get the following error when i try to print `P1 PLAN` from the following model

I tried to use `bpy.ops.bim.debug_active_drawing` to narrow it down, but that didn't seem to work. 

https://hub.openingdesign.com/OpeningDesign/Kegonsa_Residence_2_point_0/src/commit/5a98da693072d510410ab0af9bdb2d996602411d/Open/Models/Bonsai/Merged.ifc

```
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\operator.py", line 389, in execute
    linework_svg = self.generate_linework(context)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\operator.py", line 952, in generate_linework
    self.serialize_contexts_elements(ifc, tree, contexts, "body", drawing_elements, target_view)
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\bonsai\bim\module\drawing\operator.py", line 627, in serialize_contexts_elements
    self.serialiser.write(elem)
  File "C:\Users\Ryan Schultz\AppData\Roaming\Blender Foundation\Blender\4.5\extensions\.local\lib\python3.11\site-packages\ifcopenshell\ifcopenshell_wrapper.py", line 3165, in write
    return _ifcopenshell_wrapper.SvgSerializer_write(self, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: An unknown error occurred
```

BIMvoice:
## Root cause
`SvgSerializer::write()` builds a drawing plane from a section/elevation
reference curve's tangent `V` via `V.Crossed(gp::DZ())`. When `V` is
(anti-)parallel to the global Z axis, the resulting normal is the zero
vector, and constructing `gp_Ax3` from it throws `Standard_ConstructionError`
("gp_Dir() - input vector has zero norm"), crashing the whole print/export
job on a single degenerate element (reported upstream as `RuntimeError: An
unknown error occurred` through the Python/SWIG wrapper).

sboddy's diagnosis on the issue correctly identified the general failure
mode (a degenerate cross product against Z); the exact throwing statement
is the very next line (`gp_Ax3` construction), not `Crossed()` itself,
which never throws.

## Fix
Reuses the exact fallback pattern already established elsewhere in this
codebase (`OpenCascadeConversionResult.cpp`'s edge-arrow direction code):
when the tangent is nearly parallel to Z, cross against `gp::DY()` instead.
This fixes the actual degenerate computation at its source rather than
catching the exception after the fact.

## Verification
- Reproduced the exact crash class with a minimal repro (a vertical
  section-line annotation): `Standard_ConstructionError` before the fix,
  clean conversion after, confirmed via a from-scratch rebuild.
- Confirmed the fallback produces dimensionally correct output: a
  3.0m-wide × 2.5m-tall wall's elevation renders as an exact 2.5×3.0
  rectangle through the fixed plane construction, not just a non-crashing
  but meaningless result.
- Regression: ran all 83 `test/input/*wall*.ifc` fixtures through SVG
  plan output, no new failures (one pre-existing, unrelated, self-described
  infinite-loop fixture times out on both old and new code).

Fixes #8049

### DIFF
--- a/src/serializers/SvgSerializer.cpp
+++ b/src/serializers/SvgSerializer.cpp
@@ -642,8 +642,14 @@ void SvgSerializer::write(const IfcGeom::BRepElement* brep_obj) {
 				gp_Pnt P;
 				gp_Vec V;
 				crv->D1((u0 + u1) / 2., P, V);
-				auto N = V.Crossed(gp::DZ());
-				pln = gp_Pln(gp_Ax3(P, N, V));
+				if (V.Magnitude() > Precision::Confusion()) {
+					// Avoid a degenerate cross product when V is (anti-)parallel to Z,
+					// same fallback axis as OpenCascadeConversionResult.cpp.
+					const gp_Dir V_dir(V);
+					const gp_Dir& ref_axis = std::fabs(V_dir.Z()) < 0.5 ? gp::DZ() : gp::DY();
+					auto N = V.Crossed(ref_axis);
+					pln = gp_Pln(gp_Ax3(P, N, V));
+				}
 			}
 		}
 		else if (boost::optional<box_t> b = box_from_compound(compound_local)) {


aothms:
I'm a bit surprised here. This branch:

```
    if (is_section || is_elevation) {
        if (auto e = edge_from_compound(compound_unmirrored)) {
            <here>
```

is an old undocumented feature for an older integration where a drawing camera location could be defined using a single line segment as either a section plane or elevation plane 

these days cameras are constructed using a frustum as a polyhedron. So this branch should not be activated in typical Bonsai based workflows.

BIMvoice:
You're right, and I confirmed it two ways rather than just re-reading the code: traced how Bonsai actually authors camera geometry today (ifcopenshell/api/geometry/add_representation.py's create_camera_block_representation/create_camera_pyramid_representation, always an IfcBlock or IfcRectangularPyramid, never a line), and built a synthetic degenerate camera block to confirm empirically that it either stays a proper 6-faced solid or gets rejected outright by the kernel, it never collapses to a single edge. So this branch is genuinely unreachable through Bonsai's interactive camera pipeline.

It's still live through IfcConvert's own --section-ref/--elevation-ref/--elevation-ref-guid flags though, which standalone_drawer.py and any headless/scripted caller can hit independently of the interactive camera object, so it is not dead code, just not reachable the way I first assumed.

I wasn't able to pin down the actual mechanism behind the original crash: the repro file attached to the issue turned out to be unrelated (no annotations or walls at all, looks like an illustration of a different tool's output), and I could not fetch the real Merged.ifc (blocked by Cloudflare). Given the file's own name and the ad hoc merge scripting mentioned in the thread, a GUID collision between the camera and one of the walls seems like the most plausible explanation for the 3-wall pattern, which would be a data quality issue in that specific file rather than something in Bonsai's own authoring path.

Given that, this fix is still worth keeping as a real, low risk hardening against a genuine degenerate vector crash, reachable via the CLI flags either way, but I want to be upfront that it may not be the actual fix for the original report specifically.

PR #8728, opened 2026-07-18
Bonsai: apply a new Pset/Qto to the whole selection, not just the active object

### LINKED ISSUE #7462: bulk addition or change of PSET_*Common
### Bug Description

**What is happening**: PSETs need to be added one element at a time and the method for common updates differs from other common actions in Blender and in Bonsai.  If multiple items are selected and a Pset added only one gets the change.  This is slow and differs from application of types which can be done on multiple selections.
[WIH updated after comment from theoryshaw]

**What is expected**:  With multiple walls selected when I add Pset_WallCommon it should be added to all the selected walls.  When changing (for example) the object status clicking the tick should apply the change to all selected objects rather than the current two steps - Copy Property to Selection and then click the tick.  This would bring parity with type setting behaviour.
Even better would be if it could work across differing entity types. Doors, Windows, voids etc all have Common PSETs.  Voids should get updated too whether visible or not if both the containing wall and contained window/door are selected.
[WIE updated after comment from theoryshaw]

**steps to reproduce**:

1. Create a wall and duplicate it.
2. Select both walls and change the type - both walls are updated.
3. Add Pset_WallCommon and apply - only one wall gets the new Pset.
4. Change the status and apply - only one wall is updated.

*[Updated to add] Whoops, this should probably be a feature request not a bug.  Could someone with appropriate access change it please?*

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
## Root cause
EditPset._execute derived its target element solely from the active object, so adding a new pset with multiple objects selected only ever applied to one of them, unlike type assignment which already applies to every selected object. core.add_pset's existing loop over selected elements was dead code: it re-invoked enable_pset_editing with the same obj_name/obj_type every iteration without touching per-element state.

## Fix
Extracted the per-element create/edit logic into apply_pset(), and, only when a new pset/qto is being created (not when editing an existing one, where IFC's native pset sharing already keeps things in sync), added apply_new_pset_to_selection() which iterates the rest of the selected objects. For each, skips it if the pset name isn't applicable to that element's class, skips it if it already has its own occurrence-level pset of that name (matching RemovePset's existing convention, so an element that already carries the pset is left untouched rather than overwritten), and otherwise creates/edits it with the same values entered for the active object.

## Verification
Verified live in headless Blender: multi-select add (all get it), a mixed selection where one element already has its own pset of that name (left untouched, others still get it), single-object selection (unchanged), a mixed wall+door selection (door correctly skipped as not applicable), and Qto multi-select (also propagates, since it shares this operator). Added two BDD regression scenarios.

Fixes #7462

### DIFF
--- a/src/bonsai/bonsai/bim/module/pset/operator.py
+++ b/src/bonsai/bonsai/bim/module/pset/operator.py
@@ -18,7 +18,7 @@
 
 import json
 from collections import defaultdict
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import bpy
 import ifcopenshell.api
@@ -104,18 +104,6 @@ class EditPset(bpy.types.Operator, tool.Ifc.Operator):
         element = tool.Ifc.get().by_id(ifc_definition_id)
         properties = {}
 
-        pset_id = self.pset_id or props.active_pset_id
-        if pset_id:
-            pset = self.file.by_id(pset_id)
-        elif props.active_pset_type == "PSET":
-            pset = ifcopenshell.api.pset.add_pset(self.file, product=element, name=props.active_pset_name)
-            props.active_pset_id = pset.id()
-        elif props.active_pset_type == "QTO":
-            pset = ifcopenshell.api.pset.add_qto(self.file, product=element, name=props.active_pset_name)
-            props.active_pset_id = pset.id()
-        else:
-            assert False
-
         if self.properties:
             properties = json.loads(self.properties)
         else:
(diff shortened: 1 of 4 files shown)


BIMvoice:
Rebased onto current v0.8.0. The conflict was with b35f99e63f (ty: unresolved references) which added an assert to the block this PR moves into apply_pset(); the assert's intent is preserved in the new location with an explicit QTO branch. All 390 bonsai core tests pass.

PR #8729, opened 2026-07-18
Bonsai: reorient converted-mesh walls so local X carries the length

### LINKED ISSUE #7453: When a mesh is converted into a wall orientation of the mesh is not parsed
### Bug Description

**What is happening**: When a mesh is converted into a wall bonsai uses the orientation of the mesh to set the orientation of the wall, dimensions of the wall are not considered.  This can result in intended wall length and wall thickness being swapped.

**What is expected**:  When a mesh is converted into a wall the dimensions of the wall are considered and used correctly.  The longest dimension will usually be the length but probably the user should be asked to confirm. Perhaps on the confirmation dialog offer a checkbox 'Always assume the longest X or Y dimension is the wall length' and retain the setting.

**Steps to reproduce**:
1. Create a cube and adjust to dx=300mm, dy=3000mm, i.e. a wall 3m long on the y axis and 300mm thick.
2. Assign IfcElement and IfcWall.
3. Use the wall tool to create a wall 3m long on the y axis and 300mm thick.
4. Go to the window tool and click add.
5. Move the floating window around, whilst floating over the wall the window self aligns correctly.  When floating the window over the converted cube it is aligned as if the wall were 300mm long and 3000mm thick.

**Workaround**: When using the mesh convert method rotate the cube to the desired orientation before sizing it.  e.g rotate the cube 90 degrees before making it a wall aligned to the y axis.

Blender 4.5.3
Bonsai 0.8.3-post1

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
## Root cause
Converting an arbitrary mesh to an IfcWall (bim.assign_class) kept whatever local axes the mesh happened to be authored with. Every downstream consumer of wall geometry (tool.Model.get_wall_axis, window/door snapping via FilledOpeningGenerator, layer offsets) assumes local +X is the wall's length/axis direction and local Y is its thickness. A mesh resized with its long dimension along Y instead of X converted into a wall whose effective length and thickness were swapped for all of those consumers, even though the mesh's own world-space footprint was authored correctly.

Reproduced live: converting a 0.3m×3.0m×3.0m box gave get_wall_axis a reference length of 0.3m (should be 3.0m), and three window placements aimed at different points along the true 3m face all collapsed to the same spot.

## Fix
When assign_class converts a mesh into an IfcWall, rotate the mesh data 90 degrees about local Z whenever the horizontal footprint is longer along Y than along X, compensating the object's world matrix so its placement and visual footprint are unchanged. A square footprint (no unambiguous longer axis) is left as-is.

## Verification
- Verified live: reference length became 3.0m after the fix, window placements correctly spread across the true face.
- Regression: an already-correctly-authored wall is untouched (no rotation); a square footprint is untouched; world-space footprint of the fixed wall is unchanged; a non-wall class (IfcSlab) with the same swapped dimensions is correctly left untouched.
- Added 4 regression tests covering all of the above.

Fixes #7453

### DIFF
--- a/src/bonsai/bonsai/bim/module/root/operator.py
+++ b/src/bonsai/bonsai/bim/module/root/operator.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Bonsai.  If not, see <http://www.gnu.org/licenses/>.
 
+from math import radians
 from typing import TYPE_CHECKING
 
 import bmesh
@@ -32,7 +33,7 @@ import ifcopenshell.util.schema
 import ifcopenshell.util.shape_builder
 import ifcopenshell.util.type
 import ifcopenshell.util.unit
-from mathutils import Vector
+from mathutils import Matrix, Vector
 
 import bonsai.bim.module.root.prop as root_prop
 import bonsai.core.geometry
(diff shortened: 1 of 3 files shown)


Moult:
I'm tossing up whether to merge or close, but I'm leaning towards close because historically hardcoding class specific behaviour has never ended up being a good idea in the long run. If the user is manually modeling things, we shouldn't magically override their explicit actions, even if it's unexpected.

PR #8730, opened 2026-07-18
Bonsai: add a real linked TEXT annotation for dimension labels

### LINKED ISSUE #2918: Handling annotated dimensions and centerline dimensions
@aothms any ideas how you might handle dimensions which have a text annotation or a centerline? Would that somehow be semantically encoded in the text annotation itself? Would there be text literals in addition to the curve representation? How would a dimension string be associated with the text or the centerline annotation?

2023-03-30-192137_77x223_scrot
2023-03-30-192052_106x56_scrot
2023-03-30-192046_255x364_scrot
2023-03-30-191942_114x218_scrot

BIMvoice:
## Background
Dimension and diameter annotations currently carry their label text only in the IfcAnnotation Description attribute (Andrej730's original fix for #2918). Moult moved the issue back to Todo after that shipped, asking for a second, independent IfcAnnotation holding the text literal, related to the dimension through a relationship object rather than a string attribute.

## Schema check
IfcRelConnects (Moult's own rough suggestion in the thread) is abstract, and every one of its concrete subtypes (IfcRelConnectsElements, IfcRelConnectsWithRealizingElements, IfcRelConnectsPortToElement, IfcRelConnectsPorts, the structural ones) is scoped to IfcElement, IfcPort or IfcStructural* entities. IfcAnnotation is an IfcProduct, not an IfcElement, so no concrete IfcRelConnects subtype can legally connect two IfcAnnotation instances.

Bonsai already has an established, schema-valid pattern for exactly this situation: TEXT annotations get attached to another product through IfcRelAssignsToProduct, used today for stair arrows, revision clouds and generic product tags. RelatingProduct is typed IfcProductSelect, which does accept IfcProduct, so it's valid for one IfcAnnotation to be the RelatingProduct of another.

## Fix
New bim.add_dimension_text_annotation operator, available from a small panel next to Product Assignments when a DIMENSION or DIAMETER annotation is active. It creates a TEXT annotation, seeds its literal from the dimension's current Description, positions it at the dimension line's midpoint via the existing setup_annotation_object("TEXT", ...) path, and links the two with IfcRelAssignsToProduct. A companion bim.select_dimension_text_annotation operator and tool.Drawing.get_dimension_text_annotations helper let the panel show and jump to any labels already linked.

The Description-based rendering is untouched, so simple cases keep working exactly as before. This is additive: a dimension can have both a Description and any number of linked TEXT annotations at once.

## Verification
Verified live in headless Blender: created a DIMENSION annotation with Description "MIN CLEAR", ran the new operator, confirmed two distinct IfcAnnotation entities related by a real IfcRelAssignsToProduct, confirmed the drawing SVG renders both the dimension's own tag and the new TEXT annotation, and confirmed a write and reopen round trip preserves the relationship and the text literal.

Fixes #2918


theoryshaw:
Hi, Just a heads up — this PR overlaps with #8083 (*Parametric dimensions*), and on top of `v0.8.0` the two **don't auto-merge** (verified with `git merge-tree`). It's not a trivial overlap: the merge leaves **~41 lines** to resolve by hand across `bim/module/drawing/ui.py`, `bim/module/drawing/__init__.py` and `bim/module/drawing/svgwriter.py`.

BIMvoice:
Thanks for flagging this @theoryshaw. Re-checked with `git merge-tree` and pushed a fix: relocated the new registration entry in `__init__.py` and the `BIM_PT_dimension_text_label` panel class in `ui.py` to spots your PR doesn't touch (their position in the file has no functional effect). That clears 2 of the 3 conflicting files. The remaining one, `svgwriter.py`, isn't actually a collision between our two PRs, #8730 never touches that file: it's your branch predating an already-merged mainline refactor of `get_external_reference_id`, so it will show up in a rebase regardless. Since #8083 needs a rebase against v0.8.0 anyway, this is now a one-file few-line fixup rather than a three-file one. Either landing order works from our side.

Issue #7458, opened 2025-12-07
Bonsai crashes with SIGILL on CPUs without AVX2 (Ivy Bridge and older)


odofaro:
## Description
On a CPU without AVX2 (Intel i7‑3770), Blender + Bonsai crash with `Illegal instruction` when opening IFC files using the default geometry settings.

## Environment
- OS: Linux Mint 22.2 (Ubuntu 24.04 base)
- CPU: Intel Core i7-3770 (AVX, no AVX2)
- GPU: NVIDIA GTX 1050 Ti
- Blender: 5.0.0
- Bonsai: latest unstable

## Steps to reproduce
1. Start Blender with Bonsai enabled.
2. Open an IFC file with default settings (no advanced options changed).
3. Blender crashes with `Illegal instruction (core dumped)` / `SIGILL`.

## Observed behavior
- `journalctl` and `coredumpctl` show the crash in:
  - `ifcopenshell/_ifcopenshell_wrapper.cpython-311-x86_64-linux-gnu.so`
- Opening the same IFC with Python + IfcOpenShell in a virtualenv works and reports a valid file with elements.

## Workaround
When opening the IFC:
- Enable **Advanced Mode**.
- Change **Geometry Library** from `Hybrid CGAL-OCC` to `OpenCASCADE`.
- The IFC then loads without crashing.

## Suspected cause
The `Hybrid CGAL-OCC` geometry backend likely uses code (e.g. CGAL) built with AVX2 instructions, which are not supported on this CPU, while the `OpenCASCADE` backend does not.

### Debug and Error Output

```true

```

aothms:
Hm, we should likely investigate for GMP something like `./configure --enable-fat` to support multiple instruction families at runtime `./configure --build=x86_64-unknown-linux-gnu --host=x86_64-unknown-linux-gnu CFLAGS="-O2 -mtune=generic"` or `./configure CFLAGS="-O2" --host=none` and check with the performance differences on a recent machine.

We can also look again into other CGAL kernels, seeing wether Simple_cartesian (which would not use GMP/MPFR) is actually sufficient (contrary to the name, cgal-simple actually select cgal epick after some crashes on triangulation) or maybe go with boost::multiprecision instead (likely slow).

AVX2 is from around 2013. This likely won't affect the majority of users. But at least we should make a conscious choice on what is supported.

odofaro:
Thanks for looking into this! 

[...]

PR #8737, opened 2026-07-18
Bonsai: warn before a CPU without AVX2 crashes the geometry backend

BIMvoice:
## Root cause
Bonsai's default "Hybrid CGAL-OCC" geometry library links GMP (via CGAL), and GMP's non-fat build compiles hardcoded AVX2 instructions on the build host. A CPU without AVX2 (Ivy Bridge, 2012, and older) hits a SIGILL, a hardware trap that terminates Blender outright with no catchable Python exception and no error message. aothms already shipped a real fix for the build side (GMP --enable-fat, commit ab3108ff59, confirmed live in current release builds). This adds the other half odofaro asked for: a proactive warning, since a SIGILL can't be caught after the fact.

## Fix
New bonsai/cpu_check.py: pure-stdlib, no bpy or native-module imports, so it's safely importable before any risky native code runs. Detects AVX2 support per-OS (Linux via /proc/cpuinfo, macOS via sysctl, Windows via IsProcessorFeaturePresent), gated on x86 architecture (a hard no-op on ARM). Deliberately conservative: any ambiguity (non-x86, detection failure, unsupported platform) resolves to "don't warn" — false negatives are acceptable, false positives are not.

Wired into bonsai/__init__.py, computed once at startup wrapped in try/except so detection can never block Bonsai loading. When triggered: a console banner, a persistent Scene Properties panel (mirroring the existing BIM_PT_fatal_error pattern), and a one-shot popup via bpy.app.timers — the popup explicitly checks bpy.app.background first, since calling popup_menu with no window crashes Blender itself (a real, separately-confirmed native crash, not just a Python exception).

## Verification
- 21 unit tests covering real historical CPU flag data (Ivy Bridge lacks avx2, Haswell has it) across Linux/macOS/Windows mocked scenarios, all pass.
- Live in headless Blender: confirmed no false positive on this ARM64 host, and confirmed correct end-to-end behavior (console banner, panel registration, no exceptions) when simulating a missing-AVX2 x86 CPU.
- Confirmed the popup_menu-in-background crash is real (via an actual crash dump) before adding the bpy.app.background guard, and confirmed no crash after.

Related to #7458

### DIFF
--- a/src/bonsai/bonsai/__init__.py
+++ b/src/bonsai/bonsai/__init__.py
@@ -223,6 +223,17 @@ if IN_BLENDER:
 
     initialize_bbim_semver()
 
+    # CPU feature check before any native code runs, see cpu_check.py and issue #7458.
+    from bonsai import cpu_check
+
+    try:
+        cpu_warning = cpu_check.get_cpu_warning()
+    except Exception:
+        cpu_warning = None
+
+    if cpu_warning:
+        print(f"\n{'=' * 78}\nBONSAI CPU COMPATIBILITY WARNING\n{'=' * 78}\n{cpu_warning}\n{'=' * 78}\n")
+
     def get_binary_info() -> dict[str, Any]:
         info = {}
         py_version = sys.version_info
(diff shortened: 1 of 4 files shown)


BIMvoice:
Now that @aothms's GMP `--enable-fat` fix resolves the actual crash (confirmed by the reporter on #7458), this PR is no longer a needed fix, just a belt-and-suspenders startup warning for any residual AVX2 path the GMP fat build does not cover. Moving it to draft. @aothms @Moult happy to keep it as a cheap safety net or close it as redundant, your call, and we will close it ourselves if you would rather not carry the extra check.

Moult:
Agree, this is a really unnecessary bandaid.

PR #8746, opened 2026-07-18
Bonsai: fix lag when panning a view with the cut decorator active

### LINKED ISSUE #8745: Lag when panning in a view with the cut decorator active
Panning/navigating a plan/section view with `CutDecorator` active is laggy also on quite small models.

This happens because of redundant per-object work every redraw (`bonsai/bim/module/drawing/decoration.py`):

- `is_camera_moved()` is evaluated **twice per object** (`decorate()`), each call running `eval()` + numpy on the camera matrix.
- The cut/fill **intersection test** (`is_intersecting_camera()`, builds a bmesh + scans all verts) also runs **twice per object** when both recalculate.

There's also a latent bug: `is_camera_moved()` refreshes its checksum as a side effect on the first `True`, so on a moved frame only the **first** object's cut recalculates — every fill and every other element stays stale until something else invalidates the cache.

**Fix:** evaluate camera movement once per redraw, and compute the intersection once per object. PR to follow.

_Analysis done with Claude 4.8 Opus._

carlopav:
Closes #8745.

`CutDecorator` did redundant per-object work every redraw:

- `is_camera_moved()` was evaluated twice per object; now evaluated once
  per redraw and reused. This also fixes a staleness bug: the function
  refreshes its checksum on the first `True`, so previously only the first
  object's cut recalculated on a moved frame while every fill and other
  element stayed stale.
- The cut/fill intersection test (builds a bmesh + scans all verts) ran
  twice per object; now computed once in `decorate()` and passed to both,
  and skipped entirely when neither recalculation is needed.

No change to the caching model or draw output, beyond fill/cut now
updating together when the camera moves.

Analysis done with Claude Opus.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/decoration.py
+++ b/src/bonsai/bonsai/bim/module/drawing/decoration.py
@@ -1677,6 +1677,12 @@ class CutDecorator:
         selected_elements_color = self.addon_prefs.decorator_color_selected
         self.fallback_colour = (0.3, 0.3, 0.3, 1)
 
+        # Evaluate camera movement once per redraw rather than twice per object: is_camera_moved()
+        # runs eval()/numpy on the camera matrix and, as a side effect, refreshes the stored
+        # checksum on the first True result - so calling it per object also made the second call
+        # (fill) see an already-updated checksum and skip recalculating when it shouldn't.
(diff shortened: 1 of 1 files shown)


carlopav:
Here an example of the improved fluidity:
performance

PR #8751, opened 2026-07-19
Add opt-in libFuzzer harness for IfcParse::IfcFile

brunopostle:
@aothms would appreciate your review on this one.

**Note: building this harness requires a Clang toolchain** configured with `-fsanitize=fuzzer` (typically paired with `address,undefined`) via `CMAKE_CXX_FLAGS` — libFuzzer's instrumentation is Clang-only, GCC can't build this target. It's fully opt-in though: gated behind a new `BUILD_FUZZERS` CMake option that defaults to `OFF`, so it has no effect on any existing build (GCC or otherwise) unless explicitly enabled.

## Summary
- Adds a coverage-guided libFuzzer harness (`src/ifcfuzz/ifcparse_fuzzer.cpp`) that constructs `ifcopenshell::file` directly from in-memory input and calls `to_string()` on every parsed instance to force full lazy attribute evaluation, rather than only ever observing `IfcConvert`'s exit code from a fuzzed subprocess.
- `-fsanitize=fuzzer` itself is scoped to just the new `ifcparse_fuzzer` target rather than the global compiler flags, since it supplies its own `main()` and would otherwise break every other target, including CMake's own compiler checks.
- Wires up `Logger`/`logger` output (to stderr) for single-input repro runs only (`-runs=1 <file>`), not real fuzzing campaigns, where logging every parse warning on every execution would dominate the runtime. Previously the harness never called `SetOutput`/`set_output` at all, so every warning/error was silently discarded — this cost real time root-causing a leak, since grepping for an expected warning found nothing even though the underlying hypothesis was correct.
- See `src/ifcfuzz/README.md` for full build/run instructions, corpus/dictionary notes, and sections on leak detection caveats and log output.

This harness has already found and fixed three real bugs:
- A null-pointer deref in header parsing (#8716)
- Null-pointer derefs in reference resolution (#8717)
- A leak of `IfcSpfLexer` on early return/exception during file scanning (#8747)

A fourth, narrower leak in entity attribute parsing (after a syntactically valid header) is still open and noted in the README as the reason `detect_leaks=0` stays the default for general campaigns.

## Rebase note

This branch was originally forked off `ifcviewer-wgpu`, which had already renamed `IfcParse::IfcFile` → `ifcopenshell::file`. That rename has since landed on `v0.9.0` too, so rebasing onto current `v0.9.0` replayed all four commits cleanly with no conflicts. One further adaptation was needed on top: `v0.9.0` also moved `Logger` into the `ifcopenshell` namespace (`logger::root()` → `ifcopenshell::logger::root()`), and never had the `BUILD_ONLY_COMMON_SCHEMAS` cmake option the README referenced — schema selection there has always been via the `SCHEMA_VERSIONS` list. Both are fixed in the final commit.

## Test plan
- [x] Built with `-DBUILD_FUZZERS=ON` + Clang + ASan/UBSan, confirmed `ifcparse_fuzzer` target builds and links
- [x] Ran extended fuzzing campaigns against the harness (~1.6M+ executions) that produced the three fixes above
- [x] Confirmed `BUILD_FUZZERS=OFF` (default) leaves existing targets/builds unaffected
- [x] Verified the full harness with `-fsyntax-only` against the system-installed `v0.9.0` headers (`ifcparse/file.h`, `ifcparse/logger.h`) after rebasing, confirming the `ifcopenshell::logger` namespace fix compiles clean against current `v0.9.0`

### DIFF
--- a/src/ifcfuzz/ifcparse_fuzzer.cpp
+++ b/src/ifcfuzz/ifcparse_fuzzer.cpp
@@ -17,12 +17,12 @@
  *                                                                              *
  ********************************************************************************/
 
-// libFuzzer entry point for IfcParse::IfcFile. Parses the input entirely
+// libFuzzer entry point for ifcopenshell::file. Parses the input entirely
 // in-memory (no subprocess, no temp files) so a coverage-guided fuzzer can
 // reach the tokenizer and argument parser directly instead of only ever
 // observing IfcConvert's exit code.
 
-#include "ifcparse/IfcFile.h"
+#include "ifcparse/file.h"
 
 #include <cstddef>
 #include <cstdint>
@@ -35,19 +35,19 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
     }
 
     try {
-        IfcParse::IfcFile file(const_cast<void*>(static_cast<const void*>(data)), static_cast<int>(size));
+        ifcopenshell::file ifc_file(const_cast<void*>(static_cast<const void*>(data)), static_cast<int>(size));
 
-        if (file.good()) {
-            // Constructing IfcFile already tokenizes and type-checks every
+        if (ifc_file.good()) {
+            // Constructing the file already tokenizes and type-checks every
             // attribute of every instance (and resolves references), so
             // most tokenizer/argument bugs are reachable without going any
-            // further. toString() is still exercised here since
+            // further. to_string() is still exercised here since
             // reserialization walks a different code path and may surface
             // additional faults.
             std::ostringstream discard;
-            for (const auto& entity : file) {
+            for (const auto& entity : ifc_file) {
                 try {
-                    entity.second->toString(discard);
+                    entity.second.to_string(discard);
                 } catch (const std::exception&) {
                     // Malformed attributes are expected on fuzzed input.
                 }
--- a/src/ifcfuzz/ifcparse_fuzzer.cpp
+++ b/src/ifcfuzz/ifcparse_fuzzer.cpp
@@ -38,11 +38,12 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
         IfcParse::IfcFile file(const_cast<void*>(static_cast<const void*>(data)), static_cast<int>(size));
 
         if (file.good()) {
-            // IfcOpenShell parses lazily: merely constructing IfcFile only
-            // tokenizes the header and indexes instance byte offsets.
-            // toString() forces every attribute of every instance to be
-            // fully parsed, which is where most tokenizer/argument bugs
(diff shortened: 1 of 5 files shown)


src/ifcfuzz/ifcparse_fuzzer.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    try {
+        IfcParse::IfcFile file(const_cast<void*>(static_cast<const void*>(data)), static_cast<int>(size));
+
+        if (file.good()) {
+            // IfcOpenShell parses lazily: merely constructing IfcFile only
+            // tokenizes the header and indexes instance byte offsets.
+            // toString() forces every attribute of every instance to be
+            // fully parsed, which is where most tokenizer/argument bugs
+            // would actually be reachable.
```
aothms: This is not true (anymore). Could be that reserialization to string still triggers some additional faults, but would be good to make the comment more factual.

aothms (changes requested):
Can we do this on the `ifcviewer-wgpu` branch instead? This is soon to be v0.9 and saves us from some merges. We plan to switch within a month.

brunopostle:
Fixed the comment and rebased against ifcviewer-wgpu, still a pull-request though. It *should* be ok with the API changes in ifcviewer-wgpu, but I haven't done a full rebuild as I'm still on v0.8.0.

BTW, after quickly finding some bugs, this fuzzer has been running for 12 hours and hasn't found any more, so either my test file isn't complicated enough or there are no more parser problems.

brunopostle:
Rebased against v0.9.0

aothms (approved):

aothms:
Approved. Feel free to merge at any time. I think this has been a valuable effort. What are your thoughts on more semantic fuzzing like permuting numbers or references or ifc types - does something exist for that, what would be the data model it operates on?

brunopostle:
This code was the result of me just asking how to do fuzzing better than the simple script I wrote a couple of years ago. I had never heard of libfuzzer before, but it found stuff including memory leaks that my script was never going to find.

It requires a corpus of input files, but I have only provided my original test IFC, which is just 4kb or so. It seems to have mined everything it can from this file, so it is *possible* there are no simple syntax parsing bugs left in IfcOpenShell.

More likely there are lots of untested code paths that would be exercised with a bigger corpus.

Merging isn't going to break anything, but it does add yet another subproject to the src/ folder, which may be unwanted (and if it isn't going to find any more bugs, then this would be dead code).

aothms:
I think it's good to keep it around. We'll be nearing 1.0 at some point and this kind of work plays a role in becoming production ready. I don't know if it's possible to build a deterministic workflow around it but we can also add that to a CI script we run before major releases.

PR #8754, opened 2026-07-19
ifcparse: port int64_t attribute-value fix (#8683) to the ifcviewer-wgpu ifcparse rewrite

### LINKED ISSUE #3058: OverflowError: in method 'entity_instance_setArgumentAsInt', argument 3 of type 'int'

BIMvoice:
Ports #8683 (fixes #3058: out-of-range integer attributes, e.g. timestamps past 2038, corrupting on write) to this branch.

`ifcviewer-wgpu` went through a full snake_case rewrite/rename of `ifcparse` since #8683 was written against `v0.8.0` (new file layout: `entity_instance_data.cpp`, `instance_data.h`, `parse.cpp/.h`, `storage.h`, a new SWAR-based lexer, an optional RocksDB storage backend, and a rewritten PyObject-based Python attribute setter in the SWIG layer), so #8683's diff doesn't apply here. This re-derives the same fix at each equivalent choke point in the new architecture instead.

Same narrow scope as #8683: only the scalar `Argument_INT` slot is widened to `int64_t`. Integer aggregates (e.g. `IfcTriangulatedFaceSet.CoordIndex`) and instance/reference identifiers stay 32-bit, since they were never the source of the overflow and widening them would mean regenerating every schema's typed C++ getters/setters for no benefit.

Worth flagging: the pre-fix bug on this branch is actually worse than on `v0.8.0`. The new `set_attribute_value_py` used `static_cast<int>(...)`, which silently wraps out-of-range values (`CreationDate = 3000000000` became `-1294967296` with no error) instead of raising `OverflowError`. Reproduced this on an unfixed build to confirm before fixing.

Verified: pre-1901, post-2038, both int32 boundaries, and 9e12 (both signs) round-trip exactly in memory and through STEP serialization for IFC2X3 and IFC4; aggregates unaffected; an out-of-64-bit-range value now raises a clean `RuntimeError` instead of corrupting data; `test/util/test_attribute.py` and `test/test_file.py` pass with no regressions (a handful of pre-existing unrelated failures in `test_entity_instance.py` confirmed byte-identical on an unfixed build of this same branch).

Also flagging separately, not fixed here as it's out of scope: this branch's Python glue layer (`entity_instance.py`, `ifcopenshell/__init__.py`) has independent pre-existing issues (`__dict__ = property(get_info)` referencing `get_info` before its definition, and a missing `_patch_swig_comparisons`) that block Python-level usage of this branch out of the box.

### DIFF
--- a/src/helpers/geolocation.cpp
+++ b/src/helpers/geolocation.cpp
@@ -95,7 +95,7 @@ double selected_number(const express::Base& selected, double fallback) {
         return static_cast<double>(value);
     }
     if (value.type() == ifcopenshell::Argument_INT) {
-        return static_cast<double>(static_cast<int>(value));
+        return static_cast<double>(static_cast<int64_t>(value));
     }
     return fallback;
 }
--- a/src/helpers/pset.cpp
+++ b/src/helpers/pset.cpp
@@ -88,7 +88,7 @@ property_value from_attribute(const attribute_value& value) {
 
     switch (value.type()) {
     case ifcopenshell::Argument_INT:
-        return static_cast<int>(value);
+        return static_cast<int64_t>(value);
     case ifcopenshell::Argument_BOOL:
         return static_cast<bool>(value);
     case ifcopenshell::Argument_LOGICAL: {
@@ -120,7 +120,7 @@ property_value from_attribute(const attribute_value& value) {
         return instance;
     }
     case ifcopenshell::Argument_AGGREGATE_OF_INT:
-        return scalar_list(static_cast<std::vector<int>>(value));
+        return scalar_list(static_cast<std::vector<int64_t>>(value));
     case ifcopenshell::Argument_AGGREGATE_OF_DOUBLE:
         return scalar_list(static_cast<std::vector<double>>(value));
     case ifcopenshell::Argument_AGGREGATE_OF_STRING:
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -1411,7 +1411,7 @@ void fix_quantities(ifcopenshell::file& f, bool no_progress, bool quiet, bool st
 	latebound_access::set(ownerhist, "OwningUser", pando);
 	latebound_access::set(ownerhist, "OwningApplication", application);
 	latebound_access::set(ownerhist, "ChangeAction", std::string("MODIFIED"));
-	latebound_access::set(ownerhist, "CreationDate", (int)time(0));
+	latebound_access::set(ownerhist, "CreationDate", (int64_t)time(0));
 
 	express::Base quantity;
 	std::vector<express::Base> objects;
@@ -1476,7 +1476,7 @@ void fix_quantities(ifcopenshell::file& f, bool no_progress, bool quiet, bool st
 				auto quantity_count = latebound_access::create(f, "IfcQuantityCount");
 				latebound_access::set(quantity_count, "Name", std::string("Surface Genus"));
 				latebound_access::set(quantity_count, "Description", '#' + boost::lexical_cast<std::string>(part.ItemId()));
-				latebound_access::set(quantity_count, "CountValue", part.Shape()->surface_genus());
+				latebound_access::set(quantity_count, "CountValue", (int64_t) part.Shape()->surface_genus());
 
 				quantities_2.push_back(quantity_count);				
 			}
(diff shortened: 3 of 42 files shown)


aothms:
Please do make all int types consistent.

Op zo 19 jul 2026, 10:51 schreef Petru Conduraru ***@***.***>:

BIMvoice:
Done. Pushed a follow-up commit widening the remaining inconsistent spots now that compatibility isn't a constraint on this branch:

- Integer aggregates (`IfcTriangulatedFaceSet.CoordIndex` and similar `List<int>` attributes), including the SWIG helpers that were silently truncating on the Python-set path, the same bug class as the original scalar issue.
- The schema code generator (`express/mapping.py`'s integer type mapping) and all 12 generated schema header/source pairs, so every schema-typed getter/setter (e.g. `Ifc4::IfcOwnerHistory::CreationDate()`) is `int64_t` end to end, not just the dynamic attribute-value path. This matches what you originally scoped out back on #3058 ("make all integers (incl. schema namespaces) an int64_t").

One thing deliberately left at 32-bit: STEP entity identifiers (`#123` ids). These are a file-local index into internal maps, not an EXPRESS domain value an application chooses, and no realistic file has billions of entities, so widening them would mean touching indexing/hashing code for no real benefit. Flagging in case you disagree, happy to widen those too if you'd rather have zero exceptions.

Verified: extended the round-trip tests with aggregate cases at 64-bit boundary values (in memory and through STEP text, IFC2X3 and IFC4), and a standalone C++ program hitting the generated schema API directly (bypassing SWIG) to confirm the typed getters/setters are int64_t end to end. Full build clean, existing tests pass unchanged.

Issue #397, opened 2018-06-06
IfcConvert --use-element-hierarchy: IfcProject node missing, --use-element-names/guids ignored for IfcSite, IfcBuilding, and IfcBuildingStorey


Stinkfist0:
`IfcConvert.exe Duplex_A_20110907.ifc dup-hierachy.dae --sew-shells --use-material-names --use-element-guids --use-element-hierarchy`:
![image](https://user-images.githubusercontent.com/227876/41019537-b833d15a-6967-11e8-916b-c02b97a26467.png)
`product-7b7032cc-b822-417b-9aea-642906a29bd7` is the IfcSite so IfcProject appears to be is missing.

I can probably look into this soon.

aothms:
Are you sure this is not the result of the tool you used for importing. If I look at the XML output in a text editor, I do get some reference to a project node.

~~~xml
    <visual_scene id="IfcOpenShell">
      <node id="project" name="project" type="NODE">
        <matrix>1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1</matrix>
        <node id="product-7b7032cc-b822-417b-9aea-642906a29bd7" name="product-7b7032cc-b822-417b-9aea-642906a29bd7" type="NODE">
          <matrix>1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1</matrix>
          <node id="product-7b7032cc-b822-417b-9aea-642906a29bd4" name="product-7b7032cc-b822-417b-9aea-642906a29bd4" type="NODE">
~~~

Would be nice though, if it would carry over the original IfcProject name:

~~~lua
#23515=IFCPROJECT('1xS3BCk291UvhgP2a6eflL',#1,'0001',$,$,'Duplex Apartment','Project Status',(#3,#5137),#24098);
~~~

That's probably due to this code: https://github.com/IfcOpenShell/IfcOpenShell/blob/master/src/ifcgeom/IfcGeomElement.h#L141

Adding one higher level root node to encapsulate the project is also fine of course if it helps importing tools to preserve the metadata.

Stinkfist0:
![image](https://user-images.githubusercontent.com/227876/41097484-c800aec4-6a60-11e8-95c6-0bf403a01c89.png)

I ended up removing the rather arbitrary "append names to types for element with children" logic - @https://github.com/Skyndia does this look OK for you?

Stinkfist0:
Yup, adding a dummy root node, will make the project node appear.

[...]

PR #8757, opened 2026-07-19
IfcConvert: add --name-template for generic per-entity output naming

### LINKED ISSUE #413: Generic IFC output formatting
Instead of all the --use-element-names / -types flags.

See #397

Proposals:

`IfcConvert bla.ifc bla.dae --name '%N (%G -- %T)' --id '%g'` inspired by gnu find printf. For Entity, Name GlobalId Tag, uncompress guid, ...

`IfcConvert bla.ifc bla.dae --name '{{uncompress GlobalId}}'` inspired by docker's go templating

BIMvoice:
Fixes #413.

IfcConvert's output naming today is controlled by a set of standalone flags (`--use-element-names`, `--use-element-guids`, `--use-element-step-ids`, `--use-element-types`) which don't compose, so there's no way to express something like "type + name + guid" in one output name, as noted in #397.

Adds an opt-in `--name-template` flag using printf/find-style placeholders (matching the syntax the issue itself suggested, and avoiding a templating-library dependency): `%N` Name, `%G` GlobalId (compressed), `%g` GlobalId uncompressed/decoded, `%T` entity class, `%t` Tag attribute, `%i` numeric STEP id, `%u` the existing unique-id scheme, `%%` literal percent. Unknown placeholders are emitted verbatim rather than silently dropped.

Strictly additive: the existing flags are completely unchanged and still used verbatim when `--name-template` isn't passed; when it is passed, it takes priority. All naming funnels through one existing choke point (`GeometrySerializer::object_id()`), so OBJ, STP, SVG, GLTF, and TTL serializers all get this for free from one change; only `ColladaSerializer` (DAE) had its own override, whose precedence was fixed to match.

Verified against a real built `IfcConvert` binary with a synthetic IFC4 fixture: every placeholder, the `%%` literal, unknown-placeholder fallback, and priority-over-legacy-flags behavior, plus confirmed all four legacy flags remain byte-identical to before. STP output verified directly; SVG couldn't be exercised (the minimal fixture lacks the spatial/section setup SVG needs, confirmed this fails identically with no flags at all, i.e. unrelated to this change); DAE/GLTF weren't buildable in this environment but flow through the same tested code path. New test coverage added to `test/tests.py` and run standalone against the patched binary.


aothms:
This is nice but:

- Please adapt for `ifcviewer-wgpu` branch
- Since you're touching the file, convert test/tests.py‎ into normally scoped pytest tests in line with the other python-based tests in the repo.

BIMvoice:
Done, both points:

- Ported to `ifcviewer-wgpu`: #8764
- Test coverage moved out of the monolithic style into a properly-scoped pytest module (`test_ifcconvert_name_template.py`), matching this suite's own `test_wall_opening.py` convention. `ifcviewer-wgpu`'s `test/tests.py` never actually carried this feature's original monolithic block, so there was nothing to migrate there specifically, but the new coverage lives in the idiomatic form either way.

Details and a separate, unrelated finding (a null-deref robustness gap in `instance_streamer::initialize_header()` on that branch) are in #8764's description.

BIMvoice:
Closing this as superseded by #8764, which carries the same `--name-template` feature rebased onto the current API surface (`ifcopenshell::global_id`, `express::Entity`) and the current pytest test layout.

This PR targets the pre-refactor names and would now need reworking to apply. No behaviour is lost by closing it.

#8764 remains open for review.

PR #8760, opened 2026-07-19
Perf: Speed up Status Filters visibility toggle

sboddy:
## Summary
- The Status panel's per-status "has elements" badge was recomputed via 8 separate full-model IfcOpenShell selector scans (`StatusData.statuses_with_elements`), and this cache was invalidated on almost every viewport click (any active-object change), not just when filters were touched.
- `EnableStatusFilters` already computed the same "has elements" info cheaply in a single pass, but the result was never stored on a real property, so the expensive path was the only one actually wired to the UI.
- `set_visibility_by_status` (the actual on/off toggle) built a selector query per visible status instead of checking each object's status once.

This PR:
- Adds a real `has_elements` property to `IFCStatus` and uses it in the UI instead of the recomputed set.
- Rewrites `set_visibility_by_status` to check each object's status once instead of building a selector query per visible status.

## Performance
Benchmarked against a generated mock model of 3000 `IfcWall` elements with a mix of `Pset_WallCommon.Status`, `EPset_Status.Status`, and no status at all:

| Benchmark | Before | After | Speedup |
|---|---|---|---|
| `set_visibility_by_status` — 1 status visible | 1103.7 ms | 888.1 ms | 1.2x |
| `set_visibility_by_status` — 3 statuses visible | 2199.3 ms | 943.0 ms | 2.3x |
| `set_visibility_by_status` — all statuses visible | 5149.5 ms | 880.5 ms | 5.8x |
| Status panel "has elements" badge recompute | 4464.4 ms | 0.0 ms | ~5 orders of magnitude |

The badge recompute is the one that mattered most in practice: pre-fix it re-ran on almost every click in the viewport while the Status panel was visible, not just when toggling filters. The visibility toggle itself also used to get slower the more statuses were checked visible; it's now flat regardless of how many are selected.

## Test plan
- [ ] `pytest test/tool/test_sequence.py` (`TestSetVisibilityByStatus`, `TestEnableStatusFilters`)
- [ ] Manually toggle status filter checkboxes / enable / disable / refresh in the Status panel and confirm visibility results match pre-change behavior
- [ ] `black --check` / `ruff check` on touched files

## AI disclosure
This change (analysis, implementation, tests, and this PR description) was generated with the assistance of an AI coding tool.

### DIFF
--- a/src/bonsai/bonsai/bim/module/sequence/data.py
+++ b/src/bonsai/bonsai/bim/module/sequence/data.py
@@ -436,19 +436,8 @@ class StatusData:
     def load(cls) -> None:
         cls.is_loaded = True
         cls.data = {
-            "statuses_with_elements": cls.statuses_with_elements(),
             "active_element_status": cls.active_element_status(),
         }
-
-    @classmethod
-    def statuses_with_elements(cls) -> set[str]:
-        statuses = ["No Status"]
-        statuses.extend(tool.Sequence.ELEMENT_STATUSES)
-        statuses_used: set[str] = set()
-        for status in statuses:
-            if tool.Sequence.get_elements_by_status(status):
-                statuses_used.add(status)
-        return statuses_used
 
     @classmethod
     def active_element_status(cls) -> set[str]:
--- a/src/bonsai/bonsai/bim/module/sequence/prop.py
+++ b/src/bonsai/bonsai/bim/module/sequence/prop.py
@@ -434,10 +434,12 @@ class IFCStatus(PropertyGroup):
     is_visible: BoolProperty(
         name="Is Visible", default=True, update=lambda x, y: (None, bpy.ops.bim.activate_status_filters())[0]
     )
+    has_elements: BoolProperty(name="Has Elements")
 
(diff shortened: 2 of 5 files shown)


sboddy:
Manual meatbag testing on personal project confirms it is waaaaaaaay faster and more responsive.

Moult:
Can you please do a git review to find out how / when it regressed, pointing to the commits that led to it?

sboddy:
I don't need to review.
Around June last year I posted a fix to clear a minor inconsistency where hiding the last status didn't work.
At that point it was good. Then Andrej made a series of changes affecting that area, that added additional icons and emblems to the UI pane. At that point it just started going real slow. I can't point at the particular commit, because I didn't dig into it. I just told Claude when and who, and to figure out why it was broken, and to come up with a fix. It duly did. I'm attaching the plan, because it's analyses is far better explained than me regurgitating.

sboddy:
Ugh. Nope. Can't attach .md or .txt here for some reason. If you want the plan file for some light reading I can send, but the root seems to be bd0e16ce according to Claude.

Issue #4033, opened 2023-11-21
invalidating python handles on `ifc_file.remove`


Andrej730:
Currently if you remove IFC entity with `file.remove` and then try to acess the removed entity it will lead to a crash which makes it's sometimes very tricky to debug what code exactly accessed the removed entity to solve it.

Example:
```python
import ifcopenshell
file = ifcopenshell.file()
wall = file.createIfcWall(GlobalId="xxx")
print(wall)
file.remove(wall)
print(wall)
print('never printed because of the EXCEPTION_ACCESS_VIOLATION')
```

In Blender they used to have the same issue, it was especially nasty because all objects got invalidated not just on their deletion but also on any undo/redo in the current Blender session. 

The way they solved it is when object becomes invalid to avoid crashes it starts returning `<bpy_struct, Object invalid>` instead of the original object and throw `ReferenceError: StructRNA of type Object has been removed` if any of it's attributes or methods are accessed. That way Blender doesn't crash and you get a traceback indicating where invalidated object was accessed and it's easier to fix it.

Not sure how exactly it's working internally but I guess on object remove it's kind of switching some kind of flag in python handle instead of removing it completely - object is still accessible but indicates that's not valid anymore.

```python
a = bpy.data.objects['Cube']
a
# bpy.data.objects['Cube']
a.location
# Vector((0.0, 0.0, 0.0))
bpy.data.objects.remove(a)
a
#~ <bpy_struct, Object invalid>
a.location
# Traceback (most recent call last):
#   File "<blender_console>", line 1, in <module>
# ReferenceError: StructRNA of type Object has been removed
```

Andrej730:
A bit related thing I never understood - on Windows when entity is removed some of it's attributes (atleast `.id()`) is still accessible but on Linux and Mac it will lead to crash right away.

```python
import ifcopenshell

file = ifcopenshell.file()
wall = ifcopenshell.api.run("root.create_entity", file, ifc_class="IfcWall")
file.remove(wall)
print(wall.id()) # 1
print(wall) # crash
```

aothms:
> on Windows when entity is removed some of it's attributes (atleast .id()) is still accessible but on Linux and Mac it will lead to crash right away.

In the C++ world they call it "undefined behaviour" for a reason.

It's one of the last remaining deep fundamental issues. I hope I can resolve it in v0.8.

Andrej730:
[...]

Issue #2831, opened 2023-03-03
Reconcile issues when reassigning types and occurrences


Moult:
See https://github.com/IfcOpenShell/IfcOpenShell/commit/8ee60db0dfc2d203ff68f2f1cf3dd6bfd6056f38

AkankshaAI:
> See [8ee60db](https://github.com/IfcOpenShell/IfcOpenShell/commit/8ee60db0dfc2d203ff68f2f1cf3dd6bfd6056f38)

Hey , I'm interested in solving this issue but I'm a bit confused on tasks need to be done. Please clarify and assign and lemme know where can we continue the conversation

Moult:
@AkankshaAI that sounds good, have you installed and tested out the add-on and are familiar with how types and occurrences work in IFC?

PR #8761, opened 2026-07-19
ifcparse: raise a catchable exception instead of segfaulting on access to a removed instance

### LINKED ISSUE #2797: accessing deleted elem after file.remove() in python segfaults
See https://github.com/IfcOpenShell/IfcOpenShell/issues/2796

BIMvoice:
Fixes #2797. Addresses the design discussion in #4033.

`file.remove(entity)` frees the underlying C++ instance immediately while Python (or any other client) may still hold a naked pointer to it, either the reference passed to `remove()` itself, or a separate reference fetched beforehand (e.g. via `by_id()`/`by_guid()`/`by_type()`). Any further attribute, method, or repr access on that reference is a use-after-free that crashes the whole process rather than raising a catchable Python exception.

@aothms proposed exactly this fix (a boolean `is_deleted_` flag checked at the start of every access, raising a runtime exception instead of crashing) on #4033, and prototyped it in the now-closed #4637, noting it "doesn't require any API changes, because everything can remain naked pointers." That PR built the check on top of a much larger, still-unmerged datamodel rewrite (`shared_ptr`/`weak_ptr` storage, #7534) and stalled along with it, not for a design reason. This implements the same idea directly on the current (still naked-pointer) storage, with no dependency on that rewrite.

`IfcUtil::IfcBaseClass` gets a `deleted_` flag and a non-virtual `check_not_deleted()` that throws `IfcException` (already mapped to Python `RuntimeError` by the existing SWIG `%exception` block). Being non-virtual, it's safe to call even on an instance whose memory would otherwise have been freed, since it never touches the vtable.

`IfcFile::process_deletion_()` now sets `deleted_ = true` and keeps the instance alive (on a deferred-free list, actually freed only when the file itself is destroyed) instead of calling `delete` on it immediately. This means a stale reference always dereferences live, well-defined memory, never freed memory, so `check_not_deleted()` can reliably catch it instead of reading undefined data. This is the same memory/safety trade-off you described: removed instances aren't freed until the file goes away.

`check_not_deleted()` is called from every `IfcBaseClass` entry point that attribute/method access funnels through, plus the SWIG-only entry points that read `declaration()` directly (`is_a`, `id`, `get_argument_index/name/type`, `get_inverse`, `get_attribute_category/names`, `__len__`, `get_info_cpp`). A new `is_deleted()` accessor (both on the wrapper and on `ifcopenshell.entity_instance`) reports removal without raising, replacing the ad hoc `wrapped_data.id_ == 0` liveness check some call sites used (including the #2831 fix from earlier today).

Verified both scenarios from the issue thread: the original reference passed to `remove()`, and a separate reference fetched beforehand, both now raise a clean `RuntimeError` on `.Name`, `repr()`, `.id()`, `.is_a()`, attribute writes, and `get_info()`, instead of crashing. Full `ifcopenshell-python` test suite: 2417 passed, 12 skipped, 2 failed, both failures confirmed identical on a from-scratch unpatched baseline (unrelated to this change). Microbenchmark (20k entities, attribute reads/repr/is_a) shows no measurable performance difference between baseline and patched builds.

Known trade-off (as you predicted): removed instances now live until their file is destroyed rather than being freed immediately, real for bulk-delete-heavy workflows. Known gap: only the in-memory storage backend's deletion path was verified end-to-end; the rocksdb-backed variant shares the same `process_deletion_` code path but wasn't separately exercised.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/entity_instance.py
+++ b/src/ifcopenshell-python/ifcopenshell/entity_instance.py
@@ -441,6 +441,20 @@ class entity_instance:
         """Return the STEP numerical identifier"""
         return self.wrapped_data.id()
 
+    def is_deleted(self) -> bool:
+        """Return True if this instance has been removed from its file via ``file.remove()``.
+
+        Safe to call even on a reference to a removed instance (whether it is
+        the reference that was passed to ``file.remove()``, or a separate
+        reference fetched beforehand, e.g. via ``file.by_id()``): unlike
+        every other attribute or method access, this never raises.
+
+        Once an instance has been removed, any other access to it (reading
+        or writing an attribute, ``id()``, ``is_a()``, ``repr()``, etc.)
+        raises ``RuntimeError`` instead of crashing the process.
+        """
+        return self.wrapped_data.is_deleted()
+
     def __eq__(self, other: entity_instance) -> bool:
         if not isinstance(self, type(other)):
             if not self.is_entity():
--- a/src/ifcopenshell-python/test/test_file.py
+++ b/src/ifcopenshell-python/test/test_file.py
@@ -274,6 +274,43 @@ class TestFile(test.bootstrap.IFC4):
         self.file.unbatch()
         assert len(list(self.file)) == 0
 
+    def test_accessing_a_removed_element_raises_instead_of_crashing(self):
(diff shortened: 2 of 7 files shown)


aothms:
This is fixed in the datamodel-v1.0 branch (part of ifcviewer-wgpu; soon to be v0.9) using weak_ptr.

Issue #1124, opened 2020-11-24
Memory management bug in IfcOpenShell Python


dtip:
Hi folks,

Thanks for your work on IfcOpenShell. I've been trying it out and have run into a strange bug which I can't figure out.

For any IFC file I read, for every product within the file, the first two coordinates of the first vertex are some extremely small number. I believe they should be zero. The first few values in the faces array are invalid, but slightly differently invalid depending on the build:

When building from GitHub: The first and third values for the first triangle face are some extremely large (clearly invalid) integer, typically on the order of 10^7.
When using the Conda build: The first and third values in the faces array are either extremely large positive integers (10^7) or extremely large negative integers (-10^9). The second and fourth values in the faces array are a large integer (10^4) and are equal to each other.

These values are non-deterministic and change each time the Python script runs.

Any idea what's going on here?

Happy to help debug if you can provide instructions.

### Code to reproduce

```python
import ifcopenshell
import ifcopenshell.geom

ifc_file = ifcopenshell.open("some_file.ifc")
walls = ifc_file.by_type("IfcWall") # Same problem if this is changed to IfcProducts
for wall in walls:
    shape = ifcopenshell.geom.create_shape(settings, wall).geometry
    verts = shape.verts
    faces = shape.faces    
    print(verts) # (1.8579311e-316, 1.61328876e-316, 0.0, ... the rest is correct)
    print(faces) # (41427728, 0, 32653328, ... the rest is correct)
```

### Environment
Ubuntu 20.04

- Conda env (fresh environment)
  - Python 3.7.8
  - oce v0.18.3 build 3
  - IfcOpenShell v0.6.0 build py37h0d10d40_3
- System env
  - Python 3.8.5
  - required libs installed through `apt`
  - built IfcOpenShell from source using the latest v0.6.0 branch

aothms:
Quick question. Does it help if you change to:

```
x = ifcopenshell.geom.create_shape(settings, wall)
shape = x.geometry
```

Memory management in IfcOpenShell python has a couple of weird quirks and this might be one of them. With the default settings, the geometry is owned by the "Element" (as it's called in C++) and by not keeping a Python reference to it. It get's freed, freeing also the geometry.

Easiest way to fix it is to make sure somewhere in Python that the geometry carries a backreference to the Element.

dtip:
Thanks for the speedy reply. That fixes it! Is there a way to fix this quirk or to warn users about it?

[...]

PR #8763, opened 2026-07-19
ci: fix compile-and-test failures on v0.8.0 (validate_stub crash + missing Optimise test deps)

BIMvoice:
Fixes the `ci` workflow's `compile-and-test` job, red since two of our own earlier commits landed. Two unrelated failures in the same "Test ifcopenshell-python" step:

**1. validate_stub crash.** #1124's fix (824c1fc280) named its pythoncode backref helper `_geometry_with_backref` with a leading underscore. `validate_stub.py`'s `get_function_node_name()` intentionally skips any leading-underscore name (its only exception is `_is`), so the helper never lands in its parsed `subnames` set, and the assertion in `find_method_by_name()` raises a bare `AssertionError` instead of the intended clean stub/wrapper diff, crashing `make test-parallel` on every push since. Renamed to `geometry_with_backref_` (trailing underscore), matching the existing local convention in the same class bodies (`calc_volume_`, `calc_surface_area_`). Also synced six SVG edge-classification settings-flag classes into `ifcopenshell_wrapper.pyi` that were added to the C++/SWIG side but never synced to the stub, which `validate_stub` would otherwise flag once the crash above is fixed.

**2. Missing test dependencies.** `test_Optimise.py` (added in 57cfd9d1fd) exercises `Optimise`'s `patch()`, which since 7ebdd046b6 hard-requires `toposort` as its fallback backend. `ci.yml` never installed it, so every `ifcpatch` test run since has failed with `ModuleNotFoundError`. Added the missing `pip install toposort igraph` before that test step.

Verified both by reproducing the exact CI failure locally: built against the exact CI-tested commit (`c68e4a0eee`) with SWIG 4.1.0 (matching `ci.yml`'s pinned build-from-source version), confirmed `validate_stub.py` crashes identically before the fix and passes cleanly after, and confirmed `test_Optimise.py`'s 6 tests fail with `ModuleNotFoundError` without the packages and pass with them.

Checked for prior aothms guidance on this code area before implementing: his only stated intent for #1124 was "establish a backreference to the parent element from the geometry," no naming requirement, so the rename doesn't conflict with his design.

Not addressed here, flagged as a separate, chronic, pre-existing issue: `test_bsdd.py`'s two tests hit a live `429 Too Many Requests` from the real `bsdd.buildingsmart.org` API, present in CI well before these commits. That needs a design decision (retry/backoff/credentials), not a quick patch, and is out of scope for this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,6 +255,7 @@ jobs:
           cd ../bsdd && make test || ERROR=1
           pip install deepdiff
           cd ../ifcdiff && make test || ERROR=1
+          pip install toposort igraph
           cd ../ifcpatch && make test || ERROR=1
           pip install -e ../ifc5d --no-deps
           pip install odfpy xlsxwriter
--- a/src/ifcopenshell-python/ifcopenshell/ifcopenshell_wrapper.pyi
+++ b/src/ifcopenshell-python/ifcopenshell/ifcopenshell_wrapper.pyi
@@ -484,6 +484,26 @@ class Settings:
     def set_(self, *args): ...
     def setting_names(self): ...
 
+class SvgEmitFlushEdges:
+    defaultvalue: Any
+    description: Any
+    name: Any
+
+class SvgRenderCreaseEdges:
+    defaultvalue: Any
+    description: Any
+    name: Any
+
+class SvgRenderSharpEdges:
+    defaultvalue: Any
+    description: Any
+    name: Any
+
+class SvgRidgeAngleMinDegrees:
+    defaultvalue: Any
+    description: Any
+    name: Any
+
 class SvgSerializer(WriteOnlyGeometrySerializer):
     def __init__(self, out_filename, geometry_settings, settings, logger=None): ...
     SH_NONE: Any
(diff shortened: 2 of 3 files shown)


Moult:
Now that toposort has been dropped please modify

BIMvoice:
Good catch, thanks. Modified: dropped the pip install toposort igraph line entirely rather than just the toposort half. I checked test_Optimise.py against #8527's merged code with neither package installed, all 6 tests pass or skip cleanly (the two igraph tests use pytest.importorskip so they skip instead of fail). igraph was never a declared dependency in ifcpatch's pyproject.toml either, purely an optional speedup. Since #8527 is already merged into v0.8.0, this PR's remaining commit is only the validate_stub fix, no ordering concern.

BIMvoice:
Closing this as no longer needed.

This PR carried two unrelated things:

1. A `pip install toposort igraph` line in `ci.yml` for the Optimise recipe tests. #8527 has since merged and removed ifcpatch's dependency on the external `toposort` package entirely, so that install line is moot.

2. A `validate_stub` fix plus a stub sync. `ci` (compile-and-test) has been green on v0.8.0 since 2026-07-24, with `test/util/scripts/test_validate_stub.py` running in that suite, so this is not needed for green as things currently stand.

The branch has also gone stale and now conflicts with v0.8.0. Rather than rebase a PR that no longer changes an outcome, closing it to keep the review queue clear. Happy to reopen or re-cut it if `validate_stub` starts failing and this turns out to be the fix.

Status tracked on #8870.

PR #8764, opened 2026-07-19
IfcConvert: port --name-template to ifcviewer-wgpu, convert coverage to pytest

BIMvoice:
Per your review request on #8757.

Ports the `--name-template` feature (#8757, fixing #413) to this branch. `ifcparse` renamed several APIs it uses (`IfcUtil::IfcBaseEntity*` → `express::Entity`, `IfcParse::entity`/`declaration` → `ifcopenshell::entity`/`declaration`, `IfcParse::IfcGlobalId` → `ifcopenshell::global_id`, `AttributeValue` → `attribute_value`), adapted line-by-line. The choke points themselves (`GeometrySerializer::object_id()`, `ColladaSerializer::object_id()`) are structurally unchanged on this branch. Strictly additive, same as the original PR.

Per your second request, the test coverage is added directly as `test_ifcconvert_name_template.py`, an idiomatically-scoped pytest module matching this suite's existing `test_wall_opening.py` convention (class-based, skips cleanly when `IfcConvert` isn't on PATH, uses `tmp_path`), rather than `test/tests.py`'s monolithic style. `test/tests.py` on this branch never carried the original PR's block, so there was nothing to migrate out of it here.

Verified against a real `IfcConvert` built from this branch: every placeholder, the `%%` literal, unknown-placeholder fallback, priority-over-legacy-flags behavior, and all legacy flags remain byte-identical, plus the DAE (Collada) override correctly takes priority when `--name-template` is given and is unaffected when it isn't.

Separate, out-of-scope finding worth flagging: while setting up an isolated build, hit a real crash in `ifcopenshell::instance_streamer::initialize_header()`, which null-derefs `schema_` when header parsing or schema resolution fails, instead of erroring gracefully. This affects static builds specifically (schema plugins aren't discoverable) and is unrelated to this PR; happy to open a separate issue/PR for it if useful.


BIMvoice:
Note for reviewers, the failing checks here are not from this PR.

`ifcviewer-wgpu` currently fails `black --check` on 74 files branch wide. I checked authorship of all of them: none originate from this change. The one file in that list that we last touched, `src/ifcpatch/ifcpatch/recipes/ExtractElements.py`, is flagged for `getattr(ctx, 'HasCoordinateOperation', ())` using single quotes, a line written in `e6dc582d82`, not here.

So `lint-formatting` cannot go green on any PR against this branch until the branch itself is reformatted. I have deliberately not done that: 74 files across six active contributors during merge preparation would conflict with in flight work, and @Andrej730 already runs `black .` on this branch periodically, so the timing is his call.

@Moult @Andrej730 happy to run it and open it as a separate formatting only PR whenever it suits you, or to leave it entirely. Flagging rather than acting.

Issue #1330, opened 2021-02-20
Restructuring and renaming on the C++ side


aothms:

aothms:
Aim is to prevent ambiguity between IFC classes and ifcopenshell classes. So no more ifcopenshell classes that start with camelcased Ifc: `IfcParse::IfcFile` -> `ifcopenshell::io::file`

BIMvoice:
@aothms status check on this: the vast majority of your stated aim is already done on `ifcviewer-wgpu`. `IfcParse` -> `ifcopenshell`, `IfcParse::IfcFile` -> `ifcopenshell::file` (your exact example), `IfcUtil::IfcBaseEntity`/`IfcBaseClass` -> `express::Entity`/`express::Base`, `IfcGlobalId` -> `ifcopenshell::global_id`, `AttributeValue` -> `attribute_value`. Every concrete class inside `src/ifcgeom/*` already carries no `Ifc` prefix too, only the enclosing namespace still does.

What's genuinely still left, real scope not a quick pass: `namespace IfcGeom` itself (~964 references across the whole geometry library) and its correspondingly-named files (`IfcGeomElement.h`, `IfcGeomRepresentation.h`, etc.). Also drifted, not code but worth a separate look: the C++ getting-started docs still show `IfcParse::IfcFile` example code, and `ifcopenshell_wrapper.pyi` still lists Python stub classes (`IfcBaseEntity`, `IfcBaseType`, `IfcLateBoundEntity`) that no longer match what SWIG actually exposes.

Found and fixed three small leftovers from the migration while auditing: #8767 (a stray `IfcModel` class inconsistent with its un-prefixed siblings, a dead-code reference to a removed helper, and a stray leftover closing comment).

PR #8767, opened 2026-07-19
ifcgeomserver, examples: finish renaming stray Ifc-prefixed internal classes (#1330)

BIMvoice:
Small cleanup found while auditing #1330's status on this branch (see the issue comment for the full audit).

- \`IfcGeomServer.cpp\`'s local protocol-command class was still \`IfcModel\`, inconsistent with its siblings (\`Hello\`, \`More\`, \`Get\`, \`WriteLog\`), which never had the prefix. Renamed to \`Model\`.
- \`src/examples/profiles.cpp\` (not part of the default build) still referenced the pre-rename \`IfcWrite::IfcGuidHelper\`, which no longer exists. Updated to \`ifcopenshell::global_id\`.
- \`src/ifcparse/character_decoder.h\` had a stray \`} // namespace IfcWrite\` closing comment left over from that namespace's earlier rename.

Verified: built \`IfcGeomServer\` and its full dependency chain, exit code 0, ran the resulting binary and confirmed its expected \`Hello\` handshake.

### DIFF
--- a/src/examples/profiles.cpp
+++ b/src/examples/profiles.cpp
@@ -37,9 +37,10 @@
 #include INCLUDE_SCHEMA_DEFINITIONS(ifcparse/schemas, IfcSchema)
 
 #include "ifcparse/hierarchy_helper.h"
+#include "ifcparse/global_id.h"
 
 typedef std::string S;
-typedef IfcWrite::IfcGuidHelper guid;
+typedef ifcopenshell::global_id guid;
 boost::none_t const null = (static_cast<boost::none_t>(0));
 
 #ifdef SCHEMA_IfcUShapeProfileDef_HAS_CentreOfGravityInX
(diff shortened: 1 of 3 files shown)


BIMvoice:
Note for reviewers, the failing checks here are not from this PR.

`ifcviewer-wgpu` currently fails `black --check` on 74 files branch wide. I checked authorship of all of them: none originate from this change. The one file in that list that we last touched, `src/ifcpatch/ifcpatch/recipes/ExtractElements.py`, is flagged for `getattr(ctx, 'HasCoordinateOperation', ())` using single quotes, a line written in `e6dc582d82`, not here.

So `lint-formatting` cannot go green on any PR against this branch until the branch itself is reformatted. I have deliberately not done that: 74 files across six active contributors during merge preparation would conflict with in flight work, and @Andrej730 already runs `black .` on this branch periodically, so the timing is his call.

@Moult @Andrej730 happy to run it and open it as a separate formatting only PR whenever it suits you, or to leave it entirely. Flagging rather than acting.

Issue #8140, opened 2026-06-04
Boolean operation failure after wrongfully removed face in faceset_helper


aothms:
### Bug Description

- Run IfcConvert -v on https://github.com/IfcOpenShell/files/blob/master/tests/Covering_with_3_openings.ifc
- Notice Non-manifold first operand messages in log - while geometry is in fact manifold
- Due to known limitation in faceset_helper

### Debug and Error Output

```true

```

Issue #2543, opened 2022-10-28
Blenderbim: Opening on sloped wall


TZwielehner:
When I apply an opening to a sloped wall the wall jumps back to a vertical position.

Moult:
@TZwielehner I can't recreate this bug. What steps are you doing? Here's what I tried.

1. Load demo library
2. Activate BIM tool, choose wall. change slope to 20 degrees, draw wall
3. Click wall, Shift-O.
4. Shift select opening and wall, Shift-O.

Can you try with the latest unstable release? Or perhaps join osarch.org/chat and screenshare?

Moult:
@TZwielehner can you help provide more information?

TZwielehner:
Tried to reproduce it today, but couldn't. Will close it therefore.

Issue #418, opened 2018-07-11
Regression: Segfault on walls (example file)


johanrd:
**Steps to reproduce:**
1. Convert [418--walls--segfault.ifc](https://github.com/IfcOpenShell/files/blob/master/418--walls--segfault.ifc)

2. Experience segfault with 6d25047 on OCCT 7.3.0. It does not segfault on 3ac8379, so this seems like a regression:

| | 3ac8379  | 6d25047 |
| ------------- | ------------- | ------------- |
| OCCT 7.2.0 | ![#c5f015](https://placehold.it/15/c5f015/000000?text=+) Success | ![#c5f015](https://placehold.it/15/c5f015/000000?text=+) Success  |
|OCCT 7.3.0  | ![#c5f015](https://placehold.it/15/c5f015/000000?text=+) Success  | ![#f03c15](https://placehold.it/15/f03c15/000000?text=+) Segfault  |

3. File opens without warnings in DDS, SimpleBIM and Solibri:
<img width="400" alt="screen shot 2018-07-12 at 00 45 43" src="https://user-images.githubusercontent.com/4601554/42603162-faa112ac-856c-11e8-83ba-50881e40ce77.png">

johanrd:
Closing, as this is not reproducible anymore when building from commit 08557c2 🎉

johanrd:
Reopening, as there is some missing geometry when converting with 12e9f7b

```[Notice] exclude entities IfcCovering IfcFlowSegment IfcOpeningElement IfcPipeSegment IfcReinforcingBar IfcSpace
[Warning] {3Qd4fbNvv2LO9sP5StOp6Q} Halfspace subtraction yields unchanged volume:
#33=IfcBooleanClippingResult(.DIFFERENCE.,#34,#56)
[Warning] {3Qd4fbNvv2LO9sP5StOp6Q} Subtraction yields unchanged volume:
#33=IfcBooleanClippingResult(.DIFFERENCE.,#34,#56)
[Warning] {2BCTLkW3nFSQ3$WS7S2jdQ} Halfspace subtraction yields unchanged volume:
#2549=IfcBooleanClippingResult(.DIFFERENCE.,#2550,#2567)
[Warning] {2BCTLkW3nFSQ3$WS7S2jdQ} Subtraction yields unchanged volume:
#2549=IfcBooleanClippingResult(.DIFFERENCE.,#2550,#2567)
[...]

PR #8772, opened 2026-07-19
ifcgeom: keep one copy of repeated faces in faceset helper duplicate removal

BIMvoice:
## Summary
When an `IfcConnectedFaceSet` lists the same `IfcFace` multiple times (e.g. Solibri-IFC-Optimizer output), the faceset helper flagged the repeated loop identity in `duplicates_`, and `wires()` then skipped **every** occurrence of that identity — so the face vanished from the shell entirely instead of being deduplicated to one copy. The resulting shell had holes, sewing produced a non-manifold boolean operand, and opening subtraction was correctly discarded by the non-manifold safety check ("Boolean operation yields non-manifold result"), leaving the element with no geometry.

The fix counts how many occurrences of a loop identity are redundant and skips only that many, so exactly one copy of each repeated face is still built. Geometrically-identical-but-distinct-instance duplicates behave exactly as before, and the non-manifold result rejection in `boolean_utils.cpp` is untouched — this fixes the operand, per the same design direction as `--make-volume` and the duplicate-loop work in #8140.

Found in `test/input/418--walls--segfault.ifc` (issue #418's file, segfault long fixed, this was the remaining residual): wall `#2543`'s opening repeats 84 of its shell's 350 unique faces across 518 entries, which previously yielded a 0-vertex wall.

## Test plan
- Wall `#2543` now produces a closed body: 318 verts / 208 faces, bbox 3.400 × 0.200 × 0.950 m matching its `IfcExtrudedAreaSolid` exactly, watertight (every edge used exactly twice), volume 0.6457 m³ vs 0.6460 gross (the opening is a small cast-in groove, so a small cut is correct). All non-manifold warnings gone for the file.
- Before/after sweep over all 258 files in `test/input`: zero output changes other than wall `#2543` gaining geometry.
- `Covering_with_3_openings.ifc` (#8140's regression file): byte-identical output.

### DIFF
--- a/src/ifcgeom/kernels/opencascade/OpenCascadeKernel.h
+++ b/src/ifcgeom/kernels/opencascade/OpenCascadeKernel.h
@@ -86,6 +86,7 @@ private:
 	private:
 		OpenCascadeKernel* kernel_;
 		std::set<int> duplicates_;
+		std::map<int, int> duplicate_skips_remaining_;
 		std::map<int, int> vertex_mapping_;
 		std::map<std::pair<int, int>, TopoDS_Edge> edges_;
 		double eps_;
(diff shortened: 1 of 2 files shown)


BIMvoice:
Re-opened against `v0.9.0`, the current default branch, with RED/GREEN on a real build: #9396

PR #8779, opened 2026-07-19
Fix all remaining ty type-check failures on ci-lint

BIMvoice:
## Summary
The ci-lint workflow is red on every branch because base v0.8.0 fails the two ty steps (4 diagnostics total). Together with #8777 (black drift), this makes ci-lint green again.

**ty check (bonsai):**
- `root/operator.py`: `bpy.data.objects.get()` can return None, so `bim.unlink_object` with an unknown object name would crash on the first attribute access. Handle the miss explicitly — this was a real latent bug, not just a typing issue.
- `tool/sequence.py`: ty does not narrow Literal types through membership tests on list literals, flagging the `assert_never` exhaustiveness check. Tuple literals narrow correctly and keep the check intact.

**ty check (ios):**
- `draw.py`: `arrange_polygons` was called through conditional argument splats so the same call site could serve pre-April-2026 wrappers without `arrange_polygon_settings`. No runtime bug on current builds, but untypeable against the fixed 3-parameter signature. Dropped the old-build workaround and call the current signature directly, following the precedent of 3d8115ebc5 (dropping similar old-build workarounds in ifcopenshell.file). Verified against a current wrapper build that the direct call arranges polygons and serializes to SVG, with and without a logger.
- `igraph` was missing from the ios type-check venv. It is an optional dependency in ifcpatch's Optimise recipe with a guarded import, and its fallback (`toposort`) was already listed, so add it to `type-check-requirements.txt`.

After this, `poe ty-bonsai` and `poe ty-ios` both pass cleanly (verified with the CI-exact ty 0.0.59 toolchain), and black+ruff are clean on all touched files.

### DIFF
--- a/src/bonsai/bonsai/bim/module/root/operator.py
+++ b/src/bonsai/bonsai/bim/module/root/operator.py
@@ -413,12 +413,13 @@ class UnlinkObject(bpy.types.Operator, tool.Ifc.Operator):
     skip_invoke: bpy.props.BoolProperty(default=False, options={"SKIP_SAVE"})
 
     def _execute(self, context):
+        objects: list[bpy.types.Object]
         if self.obj:
-            objects = [bpy.data.objects.get(self.obj)]
+            requested_obj = bpy.data.objects.get(self.obj)
+            objects = [requested_obj] if requested_obj is not None else []
         else:
             objects = context.selected_objects
 
-        objects: list[bpy.types.Object]
         for obj in objects:
             was_active_object = obj == context.active_object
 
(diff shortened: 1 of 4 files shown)


BIMvoice:
@aothms sorry for the direct ping, and thanks for all the reviews lately. We took a run at getting CI green once and for all, and the full set is now up:

- #8777 formats the two files that drifted past black (the Black step fails on every branch because of them)
- #8779 (this PR) fixes the four ty diagnostics on base, one of which turned out to be a real latent crash in `bim.unlink_object`
- #8557 fixes the bSDD test failure in ci
- #8558 updates the test_drawing mocks for ci-bonsai-daily

With these four merged, ci, ci-lint and ci-bonsai-daily should all pass on v0.8.0, and every open PR stops inheriting red checks that aren't its own. Since the redness makes real regressions invisible (and unformatted merges keep compounding), we'd kindly ask you to prioritise these when you get a chance. Happy to rework anything that doesn't fit how you want CI handled.

Issue #5636, opened 2024-10-24
Error installing ifcopenshell on linux with glibc <2.35


francianeax4b:
Hi guys,

I was trying to install the ifcOpenShell lib using python 3.12, but it always gives this error:
ImportError: IfcOpenShell not built for 'linux/64bit/python3.12'

Are there any problems with the python version? Could anyone help me?

IfcOpenShell: 0.7.10 or 0.8.0

Andrej730:
@francianeax4b Hi! How you installed ifcopenshell?

Andrej730:
Tried to install it from this docker image
```
docker pull public.ecr.aws/docker/library/python:3.12-bullseye
docker run -it --name ifc_test -v "%cd%:/app" public.ecr.aws/docker/library/python:3.12-bullseye /bin/bash
cd /app
pip install -r requirements.txt
```

and got the error too, here's the full traceback:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/ifcopenshell/__init__.py", line 81, in <module>
    from . import ifcopenshell_wrapper
  File "/usr/local/lib/python3.12/site-packages/ifcopenshell/ifcopenshell_wrapper.py", line 13, in <module>
    from . import _ifcopenshell_wrapper
ImportError: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.35' not found (required by /usr/local/lib/python3.12/site-packages/ifcopenshell/_ifcopenshell_wrapper.cpython-312-x86_64-linux-gnu.so)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.12/site-packages/ifcopenshell/__init__.py", line 83, in <module>
    raise ImportError("IfcOpenShell not built for '%s'" % python_distribution)
ImportError: IfcOpenShell not built for 'linux/64bit/python3.12'
```


glibc version on this image:
```
> ldd --version
ldd (Debian GLIBC 2.31-13+deb11u11) 2.31
> ldd _ifcopenshell_wrapper.cpython-312-x86_64-lin
ux-gnu.so
./_ifcopenshell_wrapper.cpython-312-x86_64-linux-gnu.so: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.35' not found (required by ./_ifcopenshell_wrapper.cpython-312-x86_64-linux-gnu.so)
[...]

Issue #4686, opened 2024-05-17
identify build platform and python version from _ifcopenshell_wrapper.pyd/.so


Andrej730:
Any ideas if it's possible to identify build's python version and platform from `_ifcopenshell_wrapper.pyd` on Windows or `_ifcopenshell_wrapper.so` on Mac/Linux)?

The idea is to get more descriptive error messages instead of `ImportError: IfcOpenShell not built for 'windows\64bit\python3.11'` to guide non-advanced BlenderBIM users who can e.g. download macos instead of macom1 builds or try to use wrong python version.

The error importing wrong python version module doesn't have that information:

```python
import importlib.util
import sys

spec = importlib.util.spec_from_file_location("_ifcopenshell_wrapper", r"py310\_ifcopenshell_wrapper.pyd")
foo = importlib.util.module_from_spec(spec)
sys.modules["module.name"] = foo
# ImportError: DLL load failed while importing _ifcopenshell_wrapper: The specified module could not be found.
spec.loader.exec_module(foo)
```

Or maybe just add this information to `ifcopenshell/__init__.py` (or save it to a separate file) during build?

I was thinking that maybe we could just add it to `bl_info` for BlenderBIM during the build but it seems that it should be stored somewhere with ifcopenshell.

https://github.com/IfcOpenShell/IfcOpenShell/blob/8662e20c58a824f1ba6d8a6d7928b4284e41766f/src/blenderbim/blenderbim/__init__.py#L34

Moult:
Can swig generate some variables in `ifcopenshell_wrapper.py` that contains these variables during compilation?

aothms:
Is this fixed? Would be nice if we have a general solution for this.

> Can swig generate some variables in ifcopenshell_wrapper.py that contains these variables during compilation?

Well, I assumed deriving this from the binary is the most robust in case people start randomly copying files. If it's ok to be outside of the binary I'd say it's easiest to generate an additional python file in CMake at install time. Benefit is that we can import it without the SWIG-generated file trying to import the binary module first.

Moult:
I guess for the particular usecase of guiding non technical blender users on how to fix their install, this is resolved IMO.

Not sure what other usecases are. I'd assume developers capable of interpreting the current error message and knowing their computer specs.

Dion Moult

Sent from Proton Mail mobile

-------- Original Message --------

Andrej730:
[...]

PR #8785, opened 2026-07-19
Preserve the real cause when the ifcopenshell wrapper fails to load

### LINKED ISSUE #5927: Unable to build version 0.8 for AWS Lambda
I tried several ways to get ifcopenshell to work with AWS Lambda. For reference, I am using EFS to host my Python packages, which are later imported in my lambda function:

```
sys.path.append('/mnt/efs')
import ifcopenshell
import ifcopenshell.util
```

However, the only version that I am able to import in Lambda successfully is the one that is mentioned in the Dockerfile:

`IFC_OPENSHELL_BUILD="39-v0.7.0-476ab50"`

This is the updated Dockerfile that I made to get v0.8:

```
# Dockerfile for Running AWS Lambda Function with Python and IfcOpenShell

# Base image: Python 3.9 from AWS's public container registry
FROM public.ecr.aws/docker/library/python:3.12 AS build

# Set the location of your Lambda function code
ARG FUNCTION_DIR="/var/task"

# Install necessary packages
RUN apt-get -y update && apt-get -y install unzip wget

# Install AWS Lambda runtime interface client
RUN pip install --target ${FUNCTION_DIR} awslambdaric

# Download and extract IfcOpenShell
RUN wget https://github.com/IfcOpenShell/IfcOpenShell/releases/download/ifcopenshell-python-0.8.0/ifcopenshell-python-0.8.0-py312-linux64.zip && \
    unzip ifcopenshell-python-0.8.0-py312-linux64.zip && \
    mv ifcopenshell ${FUNCTION_DIR} && \
    rm ifcopenshell-python-0.8.0-py312-linux64.zip

# Copy the requirements file and install dependencies
COPY requirements.txt .
RUN pip install -r requirements.txt

# Copy the Lambda function code
COPY ./example_handler ${FUNCTION_DIR}/example_handler

# Set the working directory
WORKDIR ${FUNCTION_DIR}

# Set the entrypoint for the Lambda function
ENTRYPOINT [ "/usr/local/bin/python", "-m", "awslambdaric" ]

# Set the Python import path to the Lambda function handler
# This path is relative to the root of the Lambda function code (FUNCTION_DIR)
# Lambda invocation will look for this path
CMD ["example_handler.extract_wall_psets_handler"]
```
But when I copy this ifcopenshell folder to my imports, I get this error:

`IfcOpenShell not built for 'linux/64bit/python3.12'`

For v0.8, I have tried many different packages, both from the official releases:
https://docs.ifcopenshell.org/ifcopenshell-python/installation.html#zip-packages

And also the wheels:
https://pypi.org/project/ifcopenshell/0.8.0/#files

I am not able to get it to work for any Python version. 

Is there something that I'm missing? How can I get v0.8 to work with my Lambda function?

BIMvoice:
## Summary
Fixes #5927.

The wrapper import guard converted every load failure into `IfcOpenShell not built for '<platform>'`. That message is correct when no binary for the current interpreter exists, but wrong and misleading when the binary is present and fails to load — for example the glibc mismatch behind #5927 (0.8.0 zips required GLIBC_2.35, AWS Lambda's AL2023 runtime has 2.34; same class of failure as #5636). AWS Lambda and the Blender add-on error dialog only show the final exception message, so the real cause was undiagnosable, as seen in the issue thread.

Now, if a `_ifcopenshell_wrapper` binary matching the interpreter's extension suffixes exists, the raised ImportError includes the original loader error (`failed to load: version 'GLIBC_2.35' not found ...`); otherwise the familiar "not built for" message is kept. Both branches chain the original exception. Follows the direction discussed in #4686.

## Test plan
- Reproduced in the real `public.ecr.aws/lambda/python:3.12` image (glibc 2.34) with the reporter's exact `ifcopenshell-python-0.8.0-py312-linux64.zip`: before, the misleading "not built for" message; after, the real glibc error is surfaced.
- Current 0.8.5 build imports successfully in the same Lambda image, confirming upgrading resolves the reporter's actual deployment.
- Stub matrix: no binary and wrong-python-suffix binary still report "not built for"; corrupt matching binary and glibc-style failures surface the real dlopen error.
- Success-path regression: a real 0.8.5 wheel with the patched `__init__.py` imports and creates entities normally.
- black+ruff clean (CI-exact toolchain).

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/__init__.py
+++ b/src/ifcopenshell-python/ifcopenshell/__init__.py
@@ -85,8 +85,19 @@ sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "lib", p
 
 try:
     from . import ifcopenshell_wrapper
-except Exception:
-    raise ImportError("IfcOpenShell not built for '%s'" % python_distribution)
+except Exception as e:
+    import importlib.machinery
+
+    if any(
+        os.path.exists(os.path.join(directory, "_ifcopenshell_wrapper" + suffix))
+        for directory in (
+            os.path.dirname(__file__),
+            os.path.join(os.path.dirname(__file__), "lib", python_distribution),
+        )
+        for suffix in importlib.machinery.EXTENSION_SUFFIXES
+    ):
+        raise ImportError("IfcOpenShell for '%s' failed to load: %s" % (python_distribution, e)) from e
+    raise ImportError("IfcOpenShell not built for '%s'" % python_distribution) from e
 
 # `_file`, `_stream` is used only for annotations inside this file,
 # see https://github.com/microsoft/pyright/discussions/9065.
--- a/src/ifcopenshell-python/ifcopenshell/__init__.py
+++ b/src/ifcopenshell-python/ifcopenshell/__init__.py
@@ -86,18 +86,7 @@ sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "lib", p
 try:
     from . import ifcopenshell_wrapper
 except Exception as e:
-    import importlib.machinery
-
-    if any(
-        os.path.exists(os.path.join(directory, "_ifcopenshell_wrapper" + suffix))
-        for directory in (
-            os.path.dirname(__file__),
-            os.path.join(os.path.dirname(__file__), "lib", python_distribution),
-        )
-        for suffix in importlib.machinery.EXTENSION_SUFFIXES
-    ):
-        raise ImportError("IfcOpenShell for '%s' failed to load: %s" % (python_distribution, e)) from e
-    raise ImportError("IfcOpenShell not built for '%s'" % python_distribution) from e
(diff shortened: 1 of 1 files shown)


Moult:
Why the need for extra conditionals? Isn't the core benefit just printing out the Exception as e?

BIMvoice:
@Moult you're right, the conditionals weren't buying anything. Reworked to a single unconditional raise that keeps the classic message and appends the original exception: `IfcOpenShell not built for '<platform>' (<original error>)`, still chained with `from e`. Verified both situations: with no binary present it reads e.g. `IfcOpenShell not built for 'darwin/64bit/python3.14' (cannot import name '_ifcopenshell_wrapper' from 'ifcopenshell')`, and when the binary exists but fails to load it surfaces the real cause, e.g. the glibc mismatch from #5927: `IfcOpenShell not built for '...' (/lib64/libm.so.6: version GLIBC_2.29 not found (required by .../_ifcopenshell_wrapper.cpython-312-x86_64-linux-gnu.so))`. Normal import with a working wheel is unaffected.

Issue #7369, opened 2025-11-17
material layers  are merged during drawing generation


sl-x75:
### Bug Description

During drawing generation Material layers are merged  into single layer. #7252

<img width="1121" height="597" alt="Image" src="https://github.com/user-attachments/assets/127c8e1b-68e3-41a8-94b3-1ade37d7911d" />

<img width="1908" height="597" alt="Image" src="https://github.com/user-attachments/assets/a597d6eb-d265-4711-9596-fe802c7b25b7" />

### Debug and Error Output

```true

```

BIMvoice:
This looks like the same root cause as #7252, which was fixed by #7700 (merged into `v0.8.0`).

Before that fix, `merge_linework_and_add_metadata`'s default `join_criteria` used lowercase `material.Name`, which for a layered wall resolves to the wall's single `IfcMaterialLayerSet.LayerSetName` (the same value for every layer) rather than each layer's own material. So all layers of a wall hashed to the same merge key and got unioned into one polygon. #7700 added `Material.Name` (capitalized) to the criteria, which falls through to each `IfcMaterialLayer`'s own `.Material`, giving every layer a distinct key.

This issue was filed before #7700 existed and has not been revisited since, so it is very likely already resolved on a current build. Could you pull the latest `v0.8.0` and reverify? If the layers still merge for you, please reopen with the file and I will dig into whether a different material-association path (for example `IfcMaterialConstituentSet`) has a separate, still-uncovered gap.

BIMvoice:
Closing: this was fixed by #7700 (merged 2026-03-21), which added `Material.Name` to the drawing merge criteria.

Verified causally rather than by inspection. On current v0.8.0 each material layer emits its own group with exactly one `layer-material-*` class. Temporarily reverting the single line #7700 added reproduces the reported symptom exactly, including groups carrying two layer materials at once, matching the devtools screenshot in the report.

While confirming this I found a separate defect in the same merge loop (class sets being shared by reference across disjoint subpaths, so a wall split by an opening could render styled from a neighbour's GlobalId). That is fixed in #8807.

@theoryshaw reopen if you still see merged layers on a current build.

This comment was written with AI assistance.

PR #8807, opened 2026-07-20
Bonsai: stop merged cut linework leaking classes onto disjoint pieces

BIMvoice:
In `merge_linework_and_add_metadata` every closed subpath of an element group was queued with a *reference* to the same class set, so when one subpath was unioned with a neighbouring element the in-place `polygon_classes.update()` also rewrote the classes of that element's other, geometrically unrelated subpaths. A wall split by an opening could therefore emit a piece carrying a neighbour's GlobalId, status and metadata classes, styling it from the wrong CSS rules. Queue a copy of the class set per polygon so each merged polygon owns its own classes.

Demonstrated with the exact loop code — a disjoint piece belonging only to element B was emitted carrying element A's guid:

```
before:  polygon x=5.0..7.0  classes=['cut', 'guidA', 'guidB']   # wrong
after:   polygon x=5.0..7.0  classes=['cut', 'guidB']
```

## Test plan
Verified live in headless Blender against four generated multi-layer-wall models (real `IfcMaterialLayerSet` + `IfcMaterialLayerSetUsage`, walls straddling the cut plane, some voided by openings), driven through the real `bim.create_drawing` pipeline: path geometry byte-identical before and after, and class sets identical, so unaffected drawings render the same. black+ruff clean.

## Relationship to #7369
Relates to #7369, but does not close it. That issue's own reported symptom (material layers merging into one group) was already fixed by #7700 — I could not reproduce it on current v0.8.0, and temporarily reverting the single line #7700 added (`"Material.Name"` in `join_criteria`) reproduces it exactly, including the two-layer-materials-in-one-group signature from the reporter's devtools screenshot. This PR fixes a separate, still-live defect found in the same merge loop.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -1708,7 +1708,7 @@ class CreateDrawing(bpy.types.Operator):
                             coords.append(coords[0])
                         if len(coords) > 2 and coords[0] == coords[-1]:
                             is_closed_polygon = True
-                            queue.append((shapely.Polygon(coords), classes))
+                            queue.append((shapely.Polygon(coords), set(classes)))
                 if is_closed_polygon:
                     el.getparent().remove(el)
 


theoryshaw:
This is worth landing, and #9493 raises the stakes on it a little.

That PR lets a material layer's cut merge with a non-layered element of the same material (a slab's concrete layer reading continuous into a concrete footing), so more cuts end up in the same bucket than before, and buckets with several members become common rather than exceptional. The shared class set fixed here is exactly what gets copied around in that case — so the leak this fixes goes from occasional to routine.

They're independent changes and either can land first; just flagging that this one becomes more valuable once #9493 is in, rather than less.

Issue #6391, opened 2025-03-18
Some walls dissapear, when a project is opened


Beton1:
Whole walls, or wall parts disappear when a project is reopened. Recalculating wall geometry fixes the problem, but is just a quick solution.

![Image](https://github.com/user-attachments/assets/176298c8-de61-4c08-981c-21e7ac7eea49)

Moult:
Does the wall disappear every single time the project is opened, or once you've regenerated it and saved again it stays fixed?

Any chance of sharing the file?

Beton1:
It happens every time. Where can I share the file?

Moult:
Either upload here, or if private https://ifcopenshell.org/upload.html

Beton1:
Uploaded to private

BIMvoice:
@Beton1 @Moult checking in on this, it's been a while. Any progress since the file was uploaded? Given the private upload page states files are deleted once an issue resolves, and it's been over a year, is that file still available on your end? If it's easier, would you be willing to re-share it publicly here (or scrubbed of any sensitive project info if needed), we don't have access to the private upload area ourselves but would like to take a real look at reproducing this.

PR #8814, opened 2026-07-20
ifcgeom: fall back to unopened geometry when convert_openings() cleanly fails

### LINKED ISSUE #5880: Custom IfcWindowType won't pick a representation when opening the file
Low Prio issue, but curious impact.

Bit of a strange one. I have a custom window type I've created as a SKYLIGHT. When creating it I did futz around with the representations and deleted everything except the Model/Body/MODEL_VIEW, and it seems I broke something in the file. The reason for doing this being that a regular window footprint doesn't work for a skylight tilted at an angle. Now when opening the file, no representation is initially picked for the instances of this type, so I just have the manually cut roof hole, minus skylight. Now, I can go in and activate the representation for the instance/type, and this is fine until the next time I reload the file. If I forget to do this then any created drawings are missing the skylight, until I have activated the representation. I've tried "Manually Save Representation" and "Convert to Tessellation"... no effect. I've tried inspecting with the Debug panel, comparing it to a differently width-ed (but working) sibling. Even looking at that, I can't see an obvious culprit.

[MessedUpWindow.zip](https://github.com/user-attachments/files/18142438/MessedUpWindow.zip)

Attached test file just has the roof, a couple of skylight instances and the skylight types. The smaller skylight is the one causing problems. Setting the active representation on either the instance or the type will show the geometry. Save the ifc file, reload it, and the skylight is not represented again.

If I delete the instance, the type starts to behave as expected. I can then create a new instance, and the new instance is properly represented after a save/load cycle.

```
os: Linux
os_version: #136-Ubuntu SMP Wed Nov 6 10:38:22 UTC 2024
python_version: 3.11.7
machine: x86_64
processor: x86_64
blender_version: 4.2.4 LTS
bonsai_version: 0.8.1-alpha241201
bonsai_commit_hash: 1ecf722
last_actions: 
<snip - not relevant>
last_error: None
ifc: XXXXXXXXXXXXXXXX.ifc
schema: IFC4
preprocessor_version: IfcOpenShell 0.8.1-alpha241201
originating_system: Bonsai 0.8.1-alpha241201-1ecf722
```

BIMvoice:
Fixes #5880.

`Converter::create_brep_for_representation()` called `kernel_->convert_openings(...)` without checking its return value. When every candidate kernel *cleanly* rejects the opening-cut operands — CGAL's Nef conversion refusing self-intersecting geometry, logged as GEO094 rather than thrown — the resulting empty `opened_shapes` silently overwrote the product's valid unopened solid. That is exactly what the reporter described: no error, empty geometry, no fallback.

The call is now checked; a clean `false` is logged (GEO034) and sets `caught_error`, routing through the same guard that already protects the exception path (added for #6391 in the sibling PR #8811).

## Test plan
Built a real IfcConvert from this tree and ran the reporter's attached `MessedUpWindow_simplified.ifc`:

| kernel | baseline | patched |
|---|---|---|
| `opencascade` | 152 verts | 152 verts (unchanged) |
| `cgal` | **0 verts, silent** | **268 verts**, self-intersection now logged, window kept |
| `hybrid-cgal-opencascade` | **0 verts, silent** | **268 verts** |
| `hybrid-cgal-simple-opencascade` | 152 verts | 152 verts, byte-identical |

Bonsai's default kernel string (`hybrid-cgal-simple-opencascade`) was already unaffected for this particular file, because `HybridKernel::convert()` skips the non-boolean CGAL-simple kernel for elements with openings and goes straight to OpenCASCADE. The plain `cgal` and `hybrid-cgal-opencascade` configurations are genuinely broken today and are what this addresses. Also re-verified byte-identical on the second attached file (`IFCfile.txt`, a roof).

Regression: full sweep of all 258 `test/input/*.ifc` fixtures, unpatched vs patched, `--kernel hybrid-cgal-simple-opencascade` — 244 byte-identical, the other 14 failing identically on both sides (pre-existing `segfault`/`outOfMemory`/`infiniteLoop`/`invalid` repro fixtures producing no output on either binary). Zero new failures, zero diffs on any successfully-converting fixture.

Flagging a related defect found while investigating, deliberately *not* fixed here since I could not reproduce it failing: `CgalKernel::convert_impl(taxonomy::boolean_result)` silently `continue`s past a self-intersecting first operand rather than failing the whole boolean, which could produce empty-but-"successful" geometry for native boolean/clipping trees such as sloped roof cuts. That may relate to the later comments in this thread about roof geometry vanishing, which may be a distinct manifestation from the window case fixed here.

### DIFF
--- a/src/ifcgeom/Converter.cpp
+++ b/src/ifcgeom/Converter.cpp
@@ -205,8 +205,10 @@ IfcGeom::BRepElement* ifcopenshell::geometry::Converter::create_brep_for_represe
 
 			if (opening_items.empty()) {
 				opened_shapes = shapes;
-			} else {
-				kernel_->convert_openings(product, opening_items, shapes, *place, opened_shapes);
+			} else if (!kernel_->convert_openings(product, opening_items, shapes, *place, opened_shapes)) {
+				// A clean `false` return is a failure just like a thrown exception.
+				logger_.Message(Logger::LOG_ERROR, "GEO", 34, "Error processing openings for:", product);
+				caught_error = true;
 			}
 		} catch (const std::exception& e) {
 			logger_.Message(Logger::LOG_ERROR, "GEO", 33, std::string("Error processing openings for: ") + e.what() + ":", product);


BIMvoice:
Re-opened against `v0.9.0`, the current default branch, with RED/GREEN on a real build: #9393. Closing this one so there is a single PR to review.

PR #7589, opened 2026-01-21
Saikei alignment editing capabilities

Moult:
Do not merge yet, slowly working through the code from Saikei

Saikei has implemented a bunch of alignment authoring in Blender. This code represents the ability to create an alignment, create a horizontal alignment (no vertical or cant yet), and tabular editing of the PI points of alignment segments. There is more functionality in Saikei but not yet available in this PR.

Some TODO items

- [ ] Redo copyright headers to have consistent formatting
- [x] Refactor core into tool
- [x] Remove code that could be done using tool.Ifc
- [ ] Detailed line by line review
- [x] Review "debug" API operators
- [x] Move UI into Bonsai tabs instead of viewport tab
- [ ] Triple check all coordinate conversions. Sampling should be done in Blender coords. UI should be done in global coords. IFC should be in local coords. 2D layouts should be in Blender z=0.
- [ ] Replace Saikei prefixes with Bonsai for consistency
- [x] Replace polyline tool with existing tool.Polyline modals
- [ ] Make loading and creation consistent in terms of displaying alignment and toggling edit of layouts / segments

PR #8833, opened 2026-07-21
alignment: support spiral transition curves and cant in the PI-method CSV import (#6890)

### LINKED ISSUE #6890: Update alignment api CSV import to include transition curves
### Feature Description

The current alignment api has a function to import an alignment from a CSV file using the PI method. This import is limited to straight tangent runs and circular curves.

This feature is to expand this function to include horizontal alignments with spiral transition curves and cant profiles to better support the needs of railway engineers.

BIMvoice:
Fixes #6890

Restructured after @RickBrice's roadmap comment below: the clothoid and cant machinery is now a standalone, CSV-independent core in the alignment API, and the CSV import is a thin, disposable consumer. Two commits, in that order.

## Commit 1: reusable transition-curve core (the durable part)
New `solve_horizontal_alignment_by_pi_method.py` in `ifcopenshell.api.alignment`:
- `solve_horizontal_alignment_by_pi_method(hpoints, radii, cants=None)` runs the full PI solve as a pure computation (radii entries `R` or `(R, Lin, Lout)` for entry/exit clothoids; per-PI cant), returning `HorizontalSegmentDefinition` values whose fields mirror IfcAlignmentHorizontalSegment, so they write to file without further math.
- `compute_clothoid_end()` (Gauss-Legendre clothoid integrals, signed curvatures) and `compute_horizontal_segment_end()` (endpoint evaluation of any LINE/CIRCULARARC/CLOTHOID definition) exposed for reuse and continuity checking by any caller.
- Cant is embedded per horizontal segment, so the 1:1 cant/horizontal parity the cant mapping requires is structural.

`layout_horizontal_alignment_by_pi_method()` is now a thin writer over that solver, same public signature as before. This matters for the successor effort: the #7589 interactive editor (@DesertSpringsCivil) already calls exactly this function from `tool.Alignment.layout_by_pi_method`, so it gets schema-valid transition curves through its existing call site the moment its PI table passes `(R, Lin, Lout)` instead of `R` (verified live: that exact call produces LINE, CLOTHOID, CIRCULARARC, CLOTHOID, LINE). We also verified this branch adds no new merge conflicts with the saikei branch.

## Commit 2: the CSV consumer (the disposable part)
`create_from_csv.py` is parser-only: it groups columns (header-selected formats: `X,Y,R` legacy unchanged, `X,Y,R,Lin,Lout`, `X,Y,R,Lin,Lout,E`) and forwards to the core. Kept because #6890 asks for it; structured so removing the CSV path later touches nothing but this one file. If the preference is not to grow the CSV surface at all, this commit can be dropped and the core stands alone.

## Verification (re-run on the rewritten branch)
- Test suite: 43 passed (37 baseline + 6 new core tests), no failures.
- Continuity: pure solver max position gap 1.017e-12 m, direction 2.2e-16 rad; on the C++ evaluator 8.521e-07 m (the engine's clothoid series truncation), direction 2.2e-16.
- `ifcopenshell.validate` with express rules: 0 errors on the tangent-clothoid-arc-clothoid-tangent cant alignment.
- Legacy CSV output byte-identical to unmodified base (GUID/timestamp normalized).
- Black + ruff clean.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/alignment/__init__.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/alignment/__init__.py
@@ -32,7 +32,10 @@ This API does not determine alignment parameters based on rules, such as minimum
 This API is under development and subject to code breaking changes in the future.
 
 Presently, this API supports:
-    1. Creating alignments, both horizontal and vertical, using the PI method. Alignment definition can be read from a CSV file.
+    1. Creating alignments, both horizontal and vertical, using the PI method, including clothoid transition spirals and cant.
+       The horizontal PI solve is also available as a pure geometric computation (solve_horizontal_alignment_by_pi_method)
+       for callers that need segment parameters without writing to a file, such as interactive editors.
+       Alignment definition can be read from a CSV file.
     2. Creating alignments segment by segment.
     3. Automatic creation of geometric definitions (IfcCompositeCurve, IfcGradientCurve, IfcSegmentedReferenceCurve)
     4. Automatic definition of stationing
@@ -40,11 +43,10 @@ Presently, this API supports:
     6. Utility functions for printing business logical and geometric representations, as well as minimal geometry evaluations
 
 Future versions of this API may support:
-    1. Defining alignments using the PI method, including transition spirals
-    2. Updating horizontal curve definitions by revising transition spiral parameters and circular curve radii
-    3. Updating vertical curve definitions by revising horizontal length of curves
-    4. Removing a segment at any location along a curve
-    5. Adding a segment at any location along a curve
+    1. Updating horizontal curve definitions by revising transition spiral parameters and circular curve radii
+    2. Updating vertical curve definitions by revising horizontal length of curves
+    3. Removing a segment at any location along a curve
+    4. Adding a segment at any location along a curve
 """
 
 from ._get_segment_start_point_label import register_referent_name_callback
@@ -89,6 +91,12 @@ from .layout_vertical_alignment_by_pi_method import (
     layout_vertical_alignment_by_pi_method,
 )
 from .name_segments import name_segments
+from .solve_horizontal_alignment_by_pi_method import (
+    HorizontalSegmentDefinition,
+    compute_clothoid_end,
+    compute_horizontal_segment_end,
+    solve_horizontal_alignment_by_pi_method,
+)
 from .update_end_point import update_end_point
 from .update_fallback_position import update_fallback_position
 from .util import *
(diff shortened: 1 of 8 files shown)


RickBrice:
@BIMvoice - I recommend some caution before moving forward with this is implemented. The original CSV import was a stop-gap measure to allow for the most minimal input of roadway alignment into Bonsai. One user has interest in adding clothoid transitions, but the real need for this requirement is suspect. In a completely different effort, @myoder89 is developing interactive infrastructure editing for Bonsai as part of a GSoC project. This effort will make the CSV input unnecessary, obsolete, and a candidate for removal from Bonsai.

BIMvoice:
@RickBrice thanks for the context, that is exactly the kind of steer we hoped for by flagging the open questions upfront.

We built this because #6890 asked for it, but if the CSV path is a stop-gap heading for removal once @myoder89's interactive editing lands, there is no point growing it. Happy to go whichever way you prefer: leave this PR open as an interim option until the GSoC work is usable, or close it (and #6890 along with it, if you consider the requirement superseded). Your call, no attachment on our side.

One small note in case it is useful independently of the CSV question: the clothoid endpoint math and the continuity test in this PR are self-contained and could be reused by the interactive editor's geometry layer whenever it needs transition curves; happy to extract that if it helps the GSoC effort.

BIMvoice:
@RickBrice we restructured the PR around your comment. The clothoid and cant machinery no longer lives in the CSV path. It is now a standalone core in the alignment API: `solve_horizontal_alignment_by_pi_method()` does the whole PI solve (including `(R, Lin, Lout)` spiral entries and per-segment cant, one cant segment per horizontal segment) as a pure computation returning `HorizontalSegmentDefinition` values, with `compute_clothoid_end()` and `compute_horizontal_segment_end()` exposed for reuse and continuity checking. `layout_horizontal_alignment_by_pi_method()` is now a thin writer over that solver, same signature as before.

That matters for the successor effort: @DesertSpringsCivil's editor in #7589 already calls `layout_horizontal_alignment_by_pi_method()` from `tool.Alignment.layout_by_pi_method`, so with this PR the interactive editor gets schema-valid transition curves through its existing call site the moment its PI table passes `(R, Lin, Lout)` instead of `R` (we verified that exact call produces LINE, CLOTHOID, CIRCULARARC, CLOTHOID, LINE, and the pure solver is available for previews before writing to the file). We also checked this branch introduces no new merge conflicts with the saikei branch.

The CSV import is now a parser-only consumer of that core (commit 2 of 2), kept because #6890 asks for it, and structured so removing the CSV path later touches nothing but that one file. If you would rather it not grow at all, we can drop commit 2 and ship the core alone.

Issue #6680, opened 2025-05-08
shift select (or similar quick key) on the name of the material name to edit the material name.


theoryshaw:
### Feature Description

Would be nice if you could shift select (or similar quick key) on the name of the material name to edit the material name.

![Image](https://github.com/user-attachments/assets/34f386bc-4bd0-4a81-8d48-956a6ef16ef0)

falken10vdl:
@theoryshaw I see these options in order of complexity:
- 
- Adding a greasepencil next to the material name would be the easiest.. you click it to change name
- Double click would be an option as well, so single click to select and double click to rename the material
- F2 would require change the operator to a property which allows for rename with F2 and callback with click

What do you think?

theoryshaw:
How about introducing a right click, and have the operator on the context menu that pops up.  In general, I feel the right clicking UI pattern should be introduced more in Bonsai. 

I think the double clicking would render the single click for `bim.select_by_material` unusable then.  Could be wrong.  If you could get this to work without removing  `bim.select_by_material`, that would be awesome. 

I think adding the grease pencil would clutter the UI too much.

BIMvoice:
The right-click "Rename Material" entry from #8843 (merged into v0.8.0 on 2026-07-21) matches what was agreed in the thread: right-click a material name in the Object Material panel opens a rename dialog, single-click `select_by_material` stays untouched. Code is in `bonsai/bim/ui.py` (`draw_custom_context_menu`) and `bonsai/bim/module/material/operator.py` (`RenameMaterial`), with core/tool tests included. @falken10vdl since you worked the design with @theoryshaw in the thread and merged this yourself, could you confirm and close, since it merged into v0.8.0 rather than main it never auto-closed.

PR #8843, opened 2026-07-21
Bonsai: right-click Rename Material on material rows (#6680)

BIMvoice:
Implements #6680 as discussed in the thread

## What
A "Rename Material" entry in the right-click context menu of material rows (the `select_by_material` buttons in the Object Material panel), opening a pre-filled rename dialog. Implements the right-click pattern @theoryshaw proposed in the thread; single-click select-by-material behavior is completely untouched, per his explicit requirement. Only real `IfcMaterial` rows get the entry.

## How
Extends Bonsai's existing `UI_MT_button_context_menu` hook (`draw_custom_context_menu`, the same mechanism used for the bSDD copy-attribute entries) rather than a bespoke menu. The new `bim.rename_material` operator reuses the existing `material.edit_material` IFC call, so it is a faster entry point to the existing rename flow, not a new mechanism.

@falken10vdl this is offered as a starting point to adopt or adapt since the issue is assigned to you - it follows where the thread discussion converged.

## Testing
Live headless Blender (isolated profile): create project + material + wall, invoke rename, `IfcMaterial.Name` updated in the IFC data and the panel cache reflects it; context-menu logic verified to add exactly one entry for material rows and nothing for unrelated buttons. Material core/tool suites 29/29 before and after. Black + ruff clean.

### DIFF
--- a/src/bonsai/bonsai/bim/module/material/__init__.py
+++ b/src/bonsai/bonsai/bim/module/material/__init__.py
@@ -56,6 +56,7 @@ classes = (
     operator.RemoveMaterial,
     operator.RemoveMaterialSet,
     operator.RemoveProfile,
+    operator.RenameMaterial,
     operator.ReorderMaterialSetItem,
     operator.SelectByMaterial,
     operator.SelectMaterialInMaterialsUI,
--- a/src/bonsai/bonsai/bim/module/material/operator.py
+++ b/src/bonsai/bonsai/bim/module/material/operator.py
@@ -120,9 +120,6 @@ class RenameMaterial(bpy.types.Operator, tool.Ifc.Operator):
 
     def _execute(self, context):
         core.rename_material(tool.Ifc, tool.Material, material=tool.Ifc.get().by_id(self.material), name=self.name)
-        if screen := context.screen:
-            for area in screen.areas:
-                area.tag_redraw()
 
 
 class DisableEditingMaterial(bpy.types.Operator):
--- a/src/bonsai/bonsai/bim/module/material/operator.py
+++ b/src/bonsai/bonsai/bim/module/material/operator.py
@@ -102,6 +102,26 @@ class EditMaterial(bpy.types.Operator, tool.Ifc.Operator):
         core.edit_material(tool.Ifc, tool.Material, material=tool.Ifc.get().by_id(self.material))
 
 
+class RenameMaterial(bpy.types.Operator, tool.Ifc.Operator):
+    bl_idname = "bim.rename_material"
+    bl_label = "Rename Material"
+    bl_description = "Rename an IfcMaterial"
+    bl_options = {"REGISTER", "UNDO"}
+    material: bpy.props.IntProperty()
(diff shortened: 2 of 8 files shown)


falken10vdl:
@BIMvoice I think since I have not worked on It yet It is best you take over :)
Cheers!

BIMvoice:
@steverugi can you please have a look at this and provide some feedback from an end user perspective? thanks!

theoryshaw:
this seems to work great.  Only suggestion, is to automatically refresh the UI, after the change. 
That is, so you can see the name change reflected automatically in the UI.

steverugi:
@BIMvoice hi, in my personal perspective changing name is better (safer?) in the material panel to avoid affecting other element types using the same material (if I properly understand the request) 
My use of materials is much more basic compared with @theoryshaw though so I might be biased?
cheers

BIMvoice:
@theoryshaw good catch, thanks for testing. The rename itself was already correct instantly, but the panel wasn't being told to repaint, so it looked stale until something else forced a redraw. Pushed a fix that tags the UI for redraw right after the rename completes, so the new name shows up immediately.

On the red `compile-and-test` check: that failure is pre-existing on the base branch (the last 8 pushes to v0.8.0 all fail at the same `Test ifcopenshell-python` step: a validate-stub assertion, live bSDD API 429s, and missing optional deps in the runner), nothing related to this PR's files.

falken10vdl:
@BIMvoice To me the commit https://github.com/IfcOpenShell/IfcOpenShell/pull/8843/changes/16b1b4e7b175a314d0324613127e47f0c231c052 looks an overkill to redraw all areas. I belive the problem here is because you are not in editing mode so the data gets staled. IMHO  I would remove the tag_redraw() call from the operator, and move the data refresh into the core function where it runs unconditionally. The refresh() function already exists for this exact purpose.

Cheers!

PS: I do not know if one can direction the AI to warn about for loops or at even worse nested for loops. In general those are typically sources of performance degradation. I have seen that the core developers have carefully crafted a lot of cache and API mechanisms to avoid performance penalty so I think that in general every for loop must be inspected to understand if it is really required or there exists some architectural mechanism to avoid looping. My two cents :)

BIMvoice:
@falken10vdl implemented as you suggested: removed the `tag_redraw` loop from `RenameMaterial` entirely, and moved the fix into `core.rename_material`, which now calls the existing `refresh()` from `material/data.py` unconditionally (through a small `tool.Material.refresh()` wrapper, to keep core/tool layering consistent with the rest of the codebase). Interesting find along the way: that `refresh()` existed but was never called anywhere in the material module, unlike its siblings in profile/pset/drawing.

Verified live in headless Blender that renaming outside editing mode now updates the Object Material panel and Materials list caches without any redraw call involved, and that the editing-mode path still works as before. And on your PS about for-loops, fair point taken, there is no loop left at all in this operator now.

theoryshaw:
Works great from my side. 
@falken10vdl open to pulling, if it looks good?

PR #8847, opened 2026-07-21
cgal: make deflection-driven circle tessellation unconditional for small-radius conics

### LINKED ISSUE #5685: The following file doesn't recreate the curves upon opening.

sboddy:
## Background

#8484 ("Fix #5685 squared curves on open") originally fixed two bugs found while investigating #5685:

1. `reimport_element_representations()` built a fresh geometry `settings()` without copying `deflection_tolerance`/`angular_tolerance`/`geometry_library` from the settings already constructed earlier in the function, so exiting Item/edit mode silently reverted to IfcOpenShell's hard-coded mesher defaults instead of the project's configured tolerance/kernel. This is the part that shipped in #8484.
2. The CGAL kernel's `evaluate_conic()` tessellated arcs/circles using only a fixed `CircleSegments` count scaled by angular span, ignoring `mesher-linear-deflection`/`mesher-angular-deflection` entirely.

While (2) was in review, #8368 landed doing what looked like the same fix (a deflection-based floor added on top of `CircleSegments`), so the overlapping part of #8484 was dropped in favor of it, and only (1) was merged. #8368 was then reworked per maintainer request into the version that's on `v0.8.0` today (`CircleSegments` defaulting to `0`, fully deflection-driven).

## The remaining bug

Testing since then shows the merged rework still doesn't actually respect Deflection Tolerance for typical profile geometry. In `CgalKernel.cpp`'s `evaluate_conic()`:

```cpp
if (deflection > 0. && radius > deflection) {
    const double max_segment_angle = 2.0 * std::acos(1.0 - deflection / radius);
    num_segments = (int)std::ceil(span / max_segment_angle);
} else {
    // Radius within the deflection tolerance (or no deflection set): a chord per
    // quarter turn already keeps the deviation within tolerance.
    num_segments = (int)std::ceil(span / (M_PI / 2.));
}
```

The deflection-driven formula only applies when `radius > deflection`. Whenever `radius <= deflection` - the common case for profile fillets, bolt holes, small circular/hollow sections, since Bonsai's default deflection tolerance (0.05) is coarser than most such radii - it falls back to a hardcoded "one chord per quarter turn" (4 segments per full circle) that is completely insensitive to the deflection setting.

This means: adjusting Deflection Tolerance in a host app's advanced import settings has *no visible effect at all* for these elements, unless the value is pushed below the feature's own radius - an unintuitive threshold nobody would think to try when a slider is conceptually "how coarse can this be." Confirmed with a synthetic small-radius (0.01m) circular column: the old code produces an identical facet count at deflection 0.05, 0.02, and 0.011 - a 4.5x range of ordinary-looking slider values with zero change.

This is on the default code path for most users: Bonsai's (and IfcConvert's) default `geometry_library` is `hybrid-cgal-simple-opencascade`, which tries `cgal-simple` first and only falls back to OpenCASCADE on failure (e.g. booleans). Simple profile extrusions without openings succeed in `cgal-simple`, so they hit this exact branch by default.

## The fix

Replace the `radius > deflection` gate with an unconditional, clamped computation, combined via `max()` with an angular-deflection floor:

```cpp
int num_segments_linear = 1;
if (linear_deflection > 0. && radius > 0.) {
    const double half_step = std::acos((std::max)(-1.0, 1.0 - linear_deflection / radius));
    if (half_step > 1.e-9) {
        num_segments_linear = (int)std::ceil(span / (2 * half_step));
    }
}
int num_segments_angular = 1;
if (angular_deflection > 1.e-9) {
    num_segments_angular = (int)std::ceil(span / angular_deflection);
}
num_segments = (std::max)(num_segments_linear, num_segments_angular);
```

- Clamping the `acos` argument (rather than gating on `radius > deflection`) means small-radius conics still derive a segment count from linear deflection instead of being skipped outright.
- The `mesher-angular-deflection` floor (already a registered setting, default 0.5 rad, already used by the OpenCASCADE kernel via `BRepMesh_IncrementalMesh`) guarantees a sane minimum smoothness for small conics, where the linear term alone degenerates toward a single chord.
- `radius > deflection` (the #8051/#8368 large-radius case) is unaffected - produces the same segment counts as before.

This brings the CGAL kernel to parity with the dual linear+angular deflection meshing the OpenCASCADE kernel already gets for free.

## Verification

Built and tested via the local docker C++ toolchain (`IfcConvert`, `cgal-simple` kernel) against a synthetic file with a 0.01m-radius and a 0.2m-radius circular column:

| deflection | small radius (old) | small radius (new) | large radius (old) | large radius (new) |
|---|---|---|---|---|
| 0.05 | 24 verts | 78 verts | 30 verts | 78 verts |
| 0.02 / 0.011 | 24 verts (constant) | 78 verts (constant) | - | - |
| 0.001 | 42 verts | 78 verts | 192 verts | 192 verts |
| 0.0001 | 138 verts | 138 verts | 600 verts | 600 verts |

- Old code: small-radius object frozen at 24 verts across a 4.5x range of deflection values (0.05/0.02/0.011) - reproducing the reported bug.
- New code: same object responds once deflection tightens past the angular-floor threshold, while never dropping below a reasonable minimum.
- Large-radius (#8051) case matches old and new exactly at every tested tolerance where `radius > deflection` already applied - no regression there.
- Also confirmed in-Blender via a locally-built Bonsai wrapper: Deflection Tolerance in Advanced Mode now visibly changes vertex/facet count, where it previously had no effect regardless of the value set.

Refs #5685. Related: #8051, #8368.

### DIFF
--- a/src/bonsai/bonsai/bim/import_ifc.py
+++ b/src/bonsai/bonsai/bim/import_ifc.py
@@ -1286,8 +1286,8 @@ class IfcImportSettings:
         self.should_load_geometry = True
         self.should_clean_mesh = False
         self.should_cache = True
-        self.deflection_tolerance = 0.05  # Default is 0.001, but I find this to be more practical
-        self.angular_tolerance = 0.5
+        self.deflection_tolerance = 0.01  # 0.05 is IMO too course. 0.01 is a compromise between that and 0.001.
+        self.angular_tolerance = 0.5236  # Results in a nice round/even min of 12 sides per circle.
         self.void_limit = 30
         self.style_limit = 300
         # Locations greater than 1km are not considered "small sites" according to the georeferencing guide
--- a/src/ifcgeom/kernels/cgal/CgalKernel.cpp
+++ b/src/ifcgeom/kernels/cgal/CgalKernel.cpp
@@ -433,12 +433,13 @@ namespace {
 			const double span = std::fabs(a - b);
 			// CircleSegments controls how conics (circles, ellipses, arcs) are approximated
 			// in the CGAL kernel. Two modes, one or the other:
-			//  - CircleSegments == 0 (the default): the segment count is derived from
-			//    MesherLinearDeflection, so the chord deviation stays within the mesher's
-			//    linear deflection regardless of radius. This matches the deflection based
-			//    meshing the OpenCascade kernel already does and fixes issue #8051, where
-			//    large radius arcs (curved curtain wall mullions) collapsed to straight chords
-			//    because a fixed segment count is radius agnostic.
+			//  - CircleSegments == 0 (the default): the segment count is derived from the
+			//    mesher deflection settings, matching the dual linear+angular deflection
+			//    meshing the OpenCascade kernel already gets from BRepMesh_IncrementalMesh
+			//    (see MesherLinearDeflection/MesherAngularDeflection usage in
+			//    OpenCascadeConversionResult.cpp), and fixes issue #8051, where large radius
+			//    arcs (curved curtain wall mullions) collapsed to straight chords because a
+			//    fixed segment count is radius agnostic.
 			//  - CircleSegments > 0: it is used directly as the number of segments for a full
 			//    circle, giving deterministic, radius independent output.
 			int num_segments;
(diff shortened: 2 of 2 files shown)


sboddy:
Adding a cross link to #8854 - see my comment there before merging this.

PR #8854, opened 2026-07-21
Add circle_segments setting to control circle approximation in geometry imports

falken10vdl:
Currently default value for self.deflection_tolerance = 0.05 renders pipes which are small (some tens of mm) too simplistic.
Instead of lowering that value which would be a penalty for big circle radius, here is a proposal to add a segment count for circular profiles. Default value set to 24. If 0 is given then current deflection tolerance is used.

Cheers!

### DIFF
--- a/src/bonsai/bonsai/bim/import_ifc.py
+++ b/src/bonsai/bonsai/bim/import_ifc.py
@@ -1288,6 +1288,7 @@ class IfcImportSettings:
         self.should_cache = True
         self.deflection_tolerance = 0.05  # Default is 0.001, but I find this to be more practical
         self.angular_tolerance = 0.5
+        self.circle_segments = 24
         self.void_limit = 30
         self.style_limit = 300
         # Locations greater than 1km are not considered "small sites" according to the georeferencing guide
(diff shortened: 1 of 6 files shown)


sboddy:
We are perhaps working at cross purposes. See #8847

In particular note that in 5685 I linked to a [commit](https://github.com/IfcOpenShell/IfcOpenShell/commit/247f0c41d96ff59d58edfbdf21693417652a3e10) that @Moult made in January that changed the deflection tolerance.

My #8847 change uses the angular tolerance, and the deflection tolerance together to get a lower threshold of 13 segments (0.5 rad), as set in the Advanced Mode settings.

I think the whole area maybe needs a review and decision of what the preferred behaviour is.

@BIMvoice created PR #8051 and it got merged first, and I think your PR here _might_ contradict what @aothms specifically asked for, i.e. segments == 0 then use deflection.

After that _part_ of my original #8484 PR got merged.

Now you and I are both trying to fix circular arc resolution, just in very different ways. I'm using the Angular Tolerance. You are (I think) setting arcs to use 24 segments.

Like I said, I think it has gotten a little confused, and we could use some clarity.

falken10vdl:
Arrg! You are right @sboddy ! I just came with the idea of adding number of segments for circles because I saw the change in the deflection tolerance was making circles look like triangles for water pipes and the like (50mm, 100mm and the like).
Probably your PR is a better approach. I just did not want to load the file in advanced and change the deflection tolerance back to small values. 
Let me remove my PR. 
Cheers!

Issue #8865, opened 2026-07-22
Bonsai BIM won't install


Gwynblaidd:
### Bug Description

After trying to install the Bonsai I get this error.
install: URL error (<urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1032)>) reading 'https://github.com/IfcOpenShell/IfcOpenShell/releases/download/bonsai-0.8.6-alpha2607221437/bonsai_py313-0.8.6-alpha260722-windows-x64.zip'!

once dowloaded and installed manually the other issue appears

### Attachments

 File "C:\Users\op\AppData\Roaming\Blender Foundation\Blender\5.2\extensions\.local\lib\python3.13\site-packages\ifcopenshell\api\geometry\add_window_representation.py", line 30, in <module>
    from ifcopenshell.util.unit import mm_to_m as mm
ImportError: cannot import name 'mm_to_m' from 'ifcopenshell.util.unit' (C:\Users\op\AppData\Roaming\Blender Foundation\Blender\5.2\extensions\.local\lib\python3.13\site-packages\ifcopenshell\util\unit.py)
Failed to remove: (<built-in function unlink>, 'C:\\Users\\op\\AppData\\Roaming\\Blender Foundation\\Blender\\5.2\\extensions\\.local\\lib\\python3.13\\site-packages\\shapely.libs\\geos-ae6efa0782962b98e358f10ea539ae5f.dll', PermissionError(13, 'Access is denied'))
Failed to remove: (<built-in function unlink>, 'C:\\Users\\op\\AppData\\Roaming\\Blender Foundation\\Blender\\5.2\\extensions\\.local\\lib\\python3.13\\site-packages\\shapely.libs\\geos_c-072b7a9224d16d3e4ab2395bb855b2d3.dll', PermissionError(13, 'Access is denied'))
Failed to remove: (<built-in function unlink>, 'C:\\Users\\op\\AppData\\Roaming\\Blender Foundation\\Blender\\5.2\\extensions\\.local\\lib\\python3.13\\site-packages\\shapely.libs\\msvcp140-90bc62d4947a5878f1dc1057312f3be2.dll', PermissionError(13, 'Access is denied'))
[...]

PR #8867, opened 2026-07-23
Bonsai: self-repair stale bundled wheels at startup (#8865)

BIMvoice:
All Blender extensions share one site-packages for their bundled wheels,
and Blender only re-syncs it on install/enable/disable events, never at
startup. A partially failed swap (Windows cannot replace DLLs loaded
into a running Blender) or any sync miss therefore leaves the shared
ifcopenshell belonging to a different Bonsai variant than the enabled
one, permanently across restarts, surfacing as cryptic ImportErrors
deep in the registration chain (#8865: 'cannot import mm_to_m').

Before anything is imported, Bonsai now compares each wheel named in
its own blender_manifest.toml against the dist-info actually present
in the shared site-packages (pure filesystem read, so nothing gets
loaded or locked by the check). Any mismatched package, in either
version direction, is removed via a best-effort walker that survives
individual locked files, and re-extracted from this build's own
bundled wheel. On success Bonsai boots normally with a single console
notice; if files remain locked the existing fatal-error panel shows a
simple close-all-Blender-windows-and-restart message, and the next
start completes the repair automatically.

This also makes keeping several Bonsai variants installed with one
enabled at a time reliable, since the enabled variant self-corrects
the shared folder on its own startup. Healthy installs are unaffected;
the no-op scan of all bundled wheels costs about 12ms.

## Verification (isolated profile, real Blender 5.2 / py313)
- The reported multi-variant broken state (shared site-packages holding a stale ifcopenshell stripped of mm_to_m while the enabled build expects a newer one): next start self-repairs silently and a real call to the exact function that crashed for the reporter, geometry.add_window_representation, succeeds end to end.
- Healthy install: no repair triggered, no behavior change, ~12ms no-op scan.
- Locked-file simulation (chflags on the compiled wrapper): graceful panel with the restart message; everything except the locked file is already cleaned, and the next start finishes the repair automatically.

Fixes the second half of #8865.

### DIFF
--- a/src/bonsai/bonsai/__init__.py
+++ b/src/bonsai/bonsai/__init__.py
@@ -73,6 +73,11 @@ def get_git_branch() -> Union[str, None]:
 # Accessed from bonsai extension:
 bbim_semver: dict[str, Any] = {}
 
+# Wheel filenames this Bonsai build was packaged with, as recorded in
+# blender_manifest.toml's wheels list at build time. Empty if it can't be
+# determined (e.g. running from source without a packaged manifest).
+bundled_wheel_filenames: list[str] = []
+
 # Accessed from bonsai dependency:
 last_error = None
 last_actions: deque = deque(maxlen=20)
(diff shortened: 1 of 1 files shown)


BIMvoice:
@Moult flagging this one as high priority for review when you can: it addresses the recurring "Bonsai won't install / broken after update" class of first-run failures (reported in #8865, and reproduced independently on a second Windows machine the same day). Since it makes broken installs self-repair on the next start, having it in the daily builds directly reduces the most discouraging failure mode a new user can hit. Small, single-file, three-scenario verification in the body.

Moult:
Please never fight the package manager like this. If there are SSL issues that need graceful handling and can leave dirty install state, that's for Blender upstream to resolve. This is either a user environment or upstream issue and not something we will address.

Issue #6776, opened 2025-06-04
Bug when exporting rounded values


BIMvoice:
### Bug Description

Using Spreadsheet Import/Export

I try to round the values: round({{value}}, 1) , and something happens there. Is it correct behaviour? Maybe can still work but not apply round for the specific cases when this happens?

`Python: Traceback (most recent call last):
  File "C:\Users\peco01\AppData\Roaming\Blender Foundation\Blender\4.4\extensions\.local\lib\python3.11\site-packages\lark\visitors.py", line 124, in _call_userfunc
    return f(children)
           ^^^^^^^^^^^
  File "C:\Users\peco01\AppData\Roaming\Blender Foundation\Blender\4.4\extensions\.local\lib\python3.11\site-packages\ifcopenshell\util\selector.py", line 221, in round
    value = Decimal(0.0 if args[0] == "None" else args[0] or 0.0)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\peco01\AppData\Roaming\Blender Foundation\Blender\4.4\extensions\.local\lib\python3.11\site-packages\lark\visitors.py", line 128, in _call_userfunc
    raise VisitError(tree.data, tree, e)
lark.exceptions.VisitError: Error trying to process rule "round":

[<class 'decimal.ConversionSyntax'>]
`

### Debug and Error Output

```true

```

steverugi:
@BIMvoice to round a number please use the format like `round({{value}}, 0.01)` for 2 decimal places

BIMvoice:
@steverugi this works just fine exactly how I want on most models, but I get this error in one model, which I suspect it finds some elements that miss this information or something.

BIMvoice:
Tracked down the root cause. The value being rounded was a non-numeric string in that one model (a text property, or a value carrying a unit suffix like `"12.5 m"`), and `round()` called `Decimal()` on it directly, which raises `decimal.InvalidOperation` and takes down the whole export.

Reproduced against the current `v0.8.0`:

```python
import ifcopenshell.util.selector as selector
selector.format('round("Level 1", 0.01)')   # decimal.InvalidOperation
```

[...]

PR #8869, opened 2026-07-23
util.selector: metric/imperial_length pass through non-numeric values (#5297)

### LINKED ISSUE #5297: decimal places in Spreadsheet Import/Export error when N/A
to cut the number of decimal places of an attribute value I use metric_length({{value}},0.1,1)
if the query also generates N/A values (along with numbers) it returns an error 'could not convert string to float'
thanks

BIMvoice:
Fixes #5297

## Problem
`metric_length` and `imperial_length` selector format functions crash with `could not convert string to float` when the value is non-numeric (e.g. "N/A"), breaking spreadsheet export of columns with mixed content, exactly as reported.

## Root cause and correction of an earlier claim
An earlier comment of ours on #5297 pointed at PR #8297 as the fix; that was wrong. #8297 only touches `round`, `number` and `int` (for #6776), never the two length formatters. This PR fixes the actual reported functions: `float(value)` in both is now wrapped with the same pass-through convention `round()` already uses, so non-numeric values flow through unformatted while numeric values format exactly as before.

## Testing
Live: `metric_length("N/A", 0.1, 1)` crashed before, passes through after; numeric formatting unchanged. Regression tests added for both functions; full `test_selector.py` suite passes. Black + ruff clean.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/selector.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/selector.py
@@ -345,8 +345,15 @@ class FormatTransformer(lark.Transformer):
 
     def metric_length(self, args):
         value, precision, decimal_places = args
+        try:
+            numeric_value = float(value)
+        except (TypeError, ValueError):
+            # A non-numeric value (e.g. N/A, or a text property) cannot be
+            # formatted as a length. Pass it through unchanged rather than
+            # failing the whole format expression (#5297).
+            return value
         return ifcopenshell.util.unit.format_length(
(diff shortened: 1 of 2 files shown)


BIMvoice:
Closing this as a duplicate of #8268, which proposed the same fix on 2026-07-05 and is still open. We did not spot it when opening this one.

Both add the same non-numeric pass-through to the length formatters. #8268 is the older of the two, so keeping that one to preserve its place in the review queue.

#8268 remains open for review.

PR #8876, opened 2026-07-23
Bonsai: add Height setting for spaces generated from cursor

### LINKED ISSUE #4817: [BlenderBIM] | add Height box when creating an IfcSpace
It would be useful if a box with desired height of a to-be-created IfcSpace were available next to other options.

Particularly when `Generate Space from Cursor` is selected, which has 3m height set by default.

Something like:

image

It would also allow bulk height ajdustment of multiple IfcSpace in the model upon their selection

thanks

BIMvoice:
Fixes #4817.

Generate Space from Cursor always extruded new spaces to a hardcoded 3m. This adds a Height setting to the Spatial Tool, next to RL, exactly as @steverugi mocked up, so the desired space height can be set before generating. Spaces generated or regenerated from an active object still take their height from that object.

Verified headless in Blender 5.2: generating a space inside four walls with Height 2.5 produces a 2.5m IfcSpace, and a second room after changing the setting to 4.2 produces a 4.2m space.

This PR is entirely AI generated, reviewed and tested by BIMvoice.

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/prop.py
+++ b/src/bonsai/bonsai/bim/module/model/prop.py
@@ -330,6 +330,14 @@ class BIMModelProperties(PropertyGroup):
     rl2: bpy.props.FloatProperty(name="RL", default=1, subtype="DISTANCE", description="Z offset for windows")
     # Used for plan calculation points such as in room generation
     rl3: bpy.props.FloatProperty(name="RL", default=1, subtype="DISTANCE", description="Z offset for space calculation")
+    # Used for spaces generated from the cursor, where there is no active object to take the height from
+    space_height: bpy.props.FloatProperty(
+        name="Height",
+        default=3,
+        min=0.001,
+        subtype="DISTANCE",
+        description="Height of spaces generated from the cursor",
+    )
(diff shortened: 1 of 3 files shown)


CyrilWaechter:
Sorry I did not see your PR when I did mine. Although mine go deeper and has a larger scope. See https://github.com/IfcOpenShell/IfcOpenShell/pull/9005. So if you agree we will close this one.

BIMvoice:
Closing this in favour of @CyrilWaechter's #9005, which covers the same need with a better and broader solution.

This PR added a manual `space_height` property for spaces generated from the cursor. #9005 adds that same property and then goes further: it syncs the height from the active object when one is an `IfcSpace`, and adds an `Apply Space Height To Selection` operator for applying a height to existing spaces without regenerating their footprint. It is a superset of this change, and the two would conflict, since both touch `tool/spatial.py` and `spatial/workspace.py` and both introduce a `space_height` property.

For the record on sequencing, since it is not obvious from the thread: this PR was opened on 23 July and #9005 on 27 July, so there was no existing work to build on at the time. Cyril then solved the wider problem properly, and his version should be the one that lands.

@steverugi, your request in #4817 is better served by #9005 than by this.

Nothing is lost by closing this. If #9005 stalls for any reason, this smaller change can be reopened.

PR #8900, opened 2026-07-24
Bonsai: sync_render_visibility must tag the depsgraph

### LINKED ISSUE #8603: Drawing underlay renders the previously active drawing's objects (stale hide_render)
### Summary

When a drawing is created after a *different* drawing was created earlier in the session, its **underlay** (rasterized background PNG) renders the **previous drawing's objects** using the **current drawing's camera**. E.g. a ROOF plan's underlay comes out as the 1ST FLOOR plan seen from the roof camera — the roof is simply missing, so you see straight down to the slab.

The first drawing of a session is always correct.

### Root cause

`Blender.sync_render_visibility` (`src/bonsai/bonsai/tool/blender.py`) wrote `hide_render` with:

```python
bpy.data.objects.foreach_set("hide_render", should_hides)
```

`foreach_set` writes the DNA directly and **skips the RNA update, so nothing tags the depsgraph**. The *evaluated* visibility therefore stays a render behind, and `bpy.ops.render.render()` draws the previous drawing's object set. The camera is unaffected, because camera changes go through normal RNA assignment and tag correctly.

This is why the bug is so hard to see from python: `visible_get()`, `obj.hide_render` and `context.visible_objects` all read the **DNA**, which is exactly what we just wrote — every readback looks correct. Only the evaluated copy is stale.

### Fix

Assign `hide_render` through RNA so each object gets tagged. An equality check keeps it cheap by only writing objects that actually change — on a 3041-object model, 0.65s for the first drawing (2982 changes) and 0.06s for the next (248 changes).

### Evidence

Debug instrumentation around `generate_underlay`, comparing a wrong ROOF underlay against the 1ST FLOOR underlay rendered just before it:

- At the moment of `render.render()`, `scene.camera`, the depsgraph-evaluated camera, `matrix_world`, `ortho_scale` and `clip_start`/`clip_end` were all **correct for ROOF**, and the renderable object set read back **correctly as the 276-object ROOF set**.
- Both underlays render at the same 246 px/m. Aligning them on camera position and diffing showed **0.99% of pixels differing** — the ROOF underlay *was* the 1ST FLOOR image, offset by the camera delta. After the fix, 67%.
- Ruled out by measurement: the roof geometry is not clipped (nothing above the near plane but the sun lamp), not excluded from the view layer, and not affected by ray-visibility or material transparency.

### Notes on the previous diagnosis

The original write-up of this issue attributed the bug to `bpy.ops.render.render()` carrying state across calls under EEVEE-Next, and suggested reporting it upstream. That was wrong, and three of its supporting claims do not hold up:

- *"Identical visible-object set"* — the sets are completely different (60 vs 276 objects). Only the DNA readback matched, which is what disguised the bug.
- *"Re-running a single drawing on its own produces the correct underlay"* — does not reproduce. A rerun with no re-activation in between produced a **pixel-identical** wrong image, because nothing re-tagged the depsgraph.
- *"EEVEE-Next regression, legacy EEVEE was fine"* — no upstream bug is involved. The mechanism is the missing depsgraph tag. (Reports that older versions worked are consistent with the render depsgraph being rebuilt from scratch under different conditions, but that was not verified and is not needed to explain this.)

The title's "when batch-printing" was also too narrow: it happens to any drawing rendered after another one, including two separate `bpy.ops.bim.create_drawing` invocations with a full event loop in between.

### Workaround (for older builds)

Set the drawing shading style **render type to VIEWPORT** (`bpy.ops.render.opengl()`), which draws from viewport visibility (`hide_set`, set through RNA and tagged normally) and is therefore immune.

BIMvoice:
Fixes #8603.

**This PR has been rewritten.** It previously carried a modal/timer restructure of the `print_all` batch loop, built on the original diagnosis that `render.render()` was carrying EEVEE state between calls. @theoryshaw retracted that diagnosis after reproducing on his real model, and found the actual cause. The restructure is fully reverted here; the net diff is one file.

## Root cause

`Blender.sync_render_visibility` wrote `hide_render` with `bpy.data.objects.foreach_set(...)`, which writes the DNA directly and skips the RNA update, so nothing tags the depsgraph. The evaluated visibility stays one render behind, and the renderer draws the previous drawing's object set with the current drawing's camera. That is why a ROOF plan came out as the 1ST FLOOR plan seen from the roof camera.

What made it so hard to see: `visible_get()`, `obj.hide_render` and `context.visible_objects` all read back the same DNA field that was just written, so every Python readback reported correct state. Only the evaluated, render-time visibility was stale.

Worth noting that the fix already existed in the file. The RNA assignment version, equality check included, was sitting commented out directly beneath the `foreach_set` with the note "For speed, check equality prior to change to prevent needless updates". It was bypassed deliberately for speed.

## Fix

Assign `hide_render` through RNA so the depsgraph is tagged, with an equality check so only objects that actually change are written.

One deliberate difference from the commented-out block: it opened with `if not obj.data: continue`, which skipped objects with no data. That is not restored. An Empty's `hide_render` still matters when it drives collection instancing, and permanently skipping those objects means a wrong starting value is never corrected rather than merely corrected late. The loop now covers every object, matching the coverage the `foreach_set` path had.

## Verification

The mechanism was confirmed directly rather than inferred. In an isolated Blender 5.2 profile, two overlapping planes, render, flip visibility in the same call stack, render again: via `foreach_set` the second render came out stale, via RNA assignment it rendered correctly. Notably there were zero event loop iterations between the two `render.render()` calls, which is independent confirmation that the previous modal restructure could never have fixed this.

Scope honesty, two points. The visual bug was not reproduced end to end on @theoryshaw's real model; what is demonstrated here is the mechanism in isolation. And the performance measurements taken here, on a synthetic 3041 object scene, came out two to three orders of magnitude smaller than his real model figures of 0.65s then 0.06s. The qualitative shape matches, with the equality check making repeat calls roughly six to seven times cheaper, but the absolute numbers do not, most likely because a flat synthetic scene lacks the nested collection hierarchy the `hide_render` update callback walks. His measurements on the real model are the ones to trust.

black and ruff clean, before and after.

## Related, not changed here

`src/bonsai/bonsai/bim/import_ifc.py:186` calls `mesh.polygons.foreach_set("material_index", ...)` with no `mesh.update()` in that method or its caller. Lower risk, since it runs on freshly created meshes, but it is the only other `foreach_set` site in Bonsai lacking the explicit update that the rest of them have. Left alone to keep this PR to one thing.

Produced with AI assistance.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -353,10 +353,6 @@ class CreateDrawing(bpy.types.Operator):
 
     drawing_name: str
     is_manifold_cache: dict[str, bool]
-    drawings_to_print: list[int]
-    original_drawing_id: Union[int, None]
-    drawing_queue_index: int
-    _print_timer: Any
 
     @classmethod
     def poll(cls, context):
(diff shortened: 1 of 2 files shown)


theoryshaw:
@BIMvoice heads up before you spend more time on this — I finally reproduced #8603 properly on the real model, and the diagnosis this PR is built on (mine) was wrong. Full write-up in https://github.com/IfcOpenShell/IfcOpenShell/issues/8603#issuecomment-5302565495.

Short version: it isn't `render.render()` carrying state across calls, so it isn't about whether the event loop runs between drawings. `Blender.sync_render_visibility` wrote `hide_render` via `bpy.data.objects.foreach_set(...)`, which skips the RNA update and never tags the depsgraph, so the evaluated visibility stays a render behind and the renderer draws the previous drawing's objects with the current drawing's camera.

The reason this restructure won't help: my repro is two separate `bpy.ops.bim.create_drawing` invocations with a full event loop in between, and it still fails. Rendering the same drawing twice in a row produces a pixel-identical wrong image. Your note that you couldn't reproduce the EEVEE-Next carry-over with synthetic scenes was the right instinct — there was no carry-over to reproduce.

Sorry for the wrong lead. Closing or repurposing this is your call.

BIMvoice:
@theoryshaw no wasted effort, and thank you for the write-up. Tracking it to `foreach_set` skipping the RNA update is a hard find, particularly when `visible_get()`, `hide_render` and `visible_objects` all read back the value you just wrote and tell you everything is fine. The pixel diff at 0.99% against 67% is about as unambiguous as evidence gets.

Confirming from the source, and it supports you further than the write-up claims. `Blender.sync_render_visibility` at `src/bonsai/bonsai/tool/blender.py:2365` does the `foreach_set` and then returns immediately, and the RNA assignment version, equality check included, is sitting commented out directly beneath it with the comment "For speed, check equality prior to change to prevent needless updates". So the code you are describing as the fix already existed and was deliberately bypassed for speed. Your measurements of 0.65s and then 0.06s suggest that tradeoff is no longer worth making.

One detail to decide before it goes in: the commented out version opens with `if not obj.data: continue`, so objects with no data, empties among them, never get `hide_render` written at all. The `foreach_set` path writes every object unconditionally. Restoring the old block verbatim would quietly change behaviour for those. Do you want them included, or was the skip intentional?

On this PR: I will repurpose it to the RNA fix rather than close it, since the branch and the issue link are already in place and it keeps one less PR in the queue. If you already have a working branch, say so and I will close this instead and review yours, since you have the real model and the repro.

PR #8917, opened 2026-07-24
Bonsai: fix Drawing Underlay 'Add Style' wiping previously saved shading styles

### LINKED ISSUE #4739: Missunderstandind in the add button in Drawing Underlay
i propose to move the add button or changing the function. i have lost my styles, by pressing the add button since my understanding is that the plus sign mean to add a new style, and it does that, but at the cost at erasing the previous save styles. it has  frustrated me in the past and also created me not using the underlay in the past. Maybe leaving only the refresh icon, to expose the menu with the shading styles, since the add button does add to the existing styles, instead of overriding it.

Removing this add button, to not lose the saved styles, and accidentally overriding the set styles, i know that the refresh button solves that problem, but the add button confuses the user, since you could also add a style when pressing the refresh button
Screenshot 2024-05-25 231204

BIMvoice:
Fixes #4739. Thanks @SavuGust27 for the report.

The `[+]` "Add Drawing Style" button appends a new style to Bonsai's in-memory style list and immediately re-saves that whole list to the shared `shading_styles.json` file. That in-memory list is only a cache: it starts empty in a fresh Blender session and is otherwise refreshed only by pressing refresh, toggling underlay on, or activating a drawing. Pressing `[+]` before any of those happened (e.g. right after reopening the project, exactly when the panel reads "Current style is not set.") silently overwrote the JSON file with just the one newly added entry, permanently destroying every previously saved style.

`AddDrawingStyle` now reloads from disk before appending, so the button always adds on top of the real persisted set instead of risking wiping it. This matches the existing "reload-then-act" pattern already used by `update_has_underlay` and `ActivateDrawingBase` in the same module. No UI was changed or removed (the reporter suggested removing the `[+]` button, but the root cause was fixable without any UX change).

## Test plan
- [x] Added `TestDrawingStyles::test_add_drawing_style_preserves_styles_saved_in_a_previous_session` in `test/tool/test_drawing.py`, which seeds styles on disk, clears the in-memory cache to simulate a fresh session, and asserts they survive an Add. Verified it fails on the pre-fix code and passes after.
- [x] Full `test/tool/test_drawing.py` suite: 88/88 pass, no regressions.
- [x] black + ruff clean.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -2773,10 +2773,15 @@ class AddDrawingStyle(bpy.types.Operator, tool.Ifc.Operator):
     bl_idname = "bim.add_drawing_style"
     bl_label = "Add Drawing Style"
     bl_options = {"REGISTER", "UNDO"}
+    bl_description = "Add a new drawing style, keeping all previously saved styles."
 
     def _execute(self, context):
         assert context.scene and (camera_obj := context.scene.camera)
         props = tool.Drawing.get_document_props()
+        # Reload from disk first so a stale/empty cache can't overwrite previously saved styles. See #4739.
+        reload_result = bpy.ops.bim.reload_drawing_styles()
+        if "CANCELLED" in reload_result:
+            return {"CANCELLED"}
         drawing_styles = props.drawing_styles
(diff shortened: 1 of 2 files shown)


SavuGust27:
Great to know it being worked on. Such a small thing, but brings quality of life improvements

Issue #4613, opened 2024-05-04
Load classification file from schema_dir, by default


ppaawweeuu:
I think Load Classification Library button should redirect file browser to default schema location, by default.
Is it possible?

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/61344631/cbfba1bf-542a-40e8-99cd-2cc316e14053)

I remember I was nicely surprised that BlenderBIM comes with bunch of classification schemas files, located in `C:\Users\user_name\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\blenderbim\bim\schema\` folder.
And that location is pointed in BlenderBIM addon settings, and stored in `bpy.data.scenes["Scene"].BIMProperties.schema_dir`

PR #8920, opened 2026-07-24
Bonsai: built-in classification library picker

BIMvoice:
## Prototype: built-in classification library picker

Addresses the discussion in #4613. @ppaawweeuu asked for the "Load Classification Library" file browser to default to Bonsai's bundled schema folder. @Moult replied that he'd prefer built-in libraries with no file browsing required, similar to how project libraries already work.

This is a draft concept implementing Moult's preferred direction: a dropdown that lists the classification schemas already bundled under `bonsai/bim/data/classifications` (Uniclass, Omniclass, MasterFormat, RICS NRM, NBS, etc., 38 files) and loads the picked one directly, no file dialog.

It mirrors the existing built-in project-library pattern (`ProjectData.library_file` / `update_library_file` / `bim.select_library_file`) as closely as possible: an enum populated from `tool.Blender.get_data_dir_paths`, with a "Custom File" entry that reveals the existing file-browser button for anything not bundled. The file-browser path (`bim.load_classification_library`) is unchanged and still available for custom files.

Verified headless: the dropdown enumerates all 38 bundled schemas, picking one loads it via `IfcStore.classification_file` with no dialog, and `Add Classification` then adds it into the live IFC model.

Scope: only the main Classifications panel got the dropdown; the per-object/material/cost/zone "Add Reference" panels still use the file browser (flagged as a follow-up).

@Moult, does this match what you had in mind, or would you rather the picker live somewhere else / work differently (e.g. a separate always-visible dropdown instead of a "Custom File" enum entry)? Opening as a draft to get your steer before polishing further.

### DIFF
--- a/src/bonsai/bonsai/bim/module/classification/data.py
+++ b/src/bonsai/bonsai/bim/module/classification/data.py
@@ -46,10 +46,24 @@ class ClassificationsData:
         cls.data["classifications"] = cls.classifications()
         cls.data["available_classifications"] = cls.available_classifications()
         cls.data["classification_source"] = cls.classification_source()
+        cls.data["builtin_classification_libraries"] = cls.builtin_classification_libraries()
 
     @classmethod
     def has_classification_file(cls):
         return bool(IfcStore.classification_file)
+
+    @classmethod
+    def builtin_classification_libraries(cls) -> tool.Blender.BLENDER_ENUM_ITEMS:
+        """Enumerate classification libraries bundled with Bonsai (and any user-provided ones),
+        so a built-in library can be picked directly without browsing for a file.
+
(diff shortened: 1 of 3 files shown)


Moult:
This looks good, why is it draft, is there more intended to build?

BIMvoice:
@Moult it was a draft because the PR body asked you a design question: whether the picker belonged in that panel, and whether the "Custom File" enum entry was the right shape versus a separate always visible dropdown. Your "looks good" answers it, so it is out of draft now.

On what is left, there is one item, flagged as a follow up in the PR body. Only the main Classifications panel got the dropdown. The per object, per material, per cost item and per zone "Add Reference" panels still open the file browser. That is coming as a separate PR, so this one stays at the scope you just reviewed.

PR #8930, opened 2026-07-24
Bonsai: cut and align skylights/windows on parametric roofs to the sloped face

### LINKED ISSUE #5611: Windows do not apply to roofs to act as skylights, even when predefined type is set
Blender 4.2.3 LTS, Bonsai 0.8.1-alpha241019-b8d98b7

Use attached file.
[roof_skylights_not_smart.zip](https://github.com/user-attachments/files/17450386/roof_skylights_not_smart.zip)

1. Try to add WT01 or WT02 (SKYLIGHT) to the roof object.

Actual result:
Window is vertical and no void is applied to the roof item:

Expected result:
Window is rotated to aligned with the roof slope, and a void is added to create the opening.

BIMvoice:
Fixes #5611.

Adding a window/skylight to a parametric Roof left the filling vertical with no opening cut. `get_material_layer_parameters()` defaulted the layer set direction to `AXIS2` (wall-like, translation only) whenever no material was found, and parametric roofs never get a material layer set, so they silently fell into the wall placement branch. The `AXIS3` (slab-like) placement also assumed a single flat top face, so it wouldn't have aligned to a sloped face even with the right default.

This adds:
- a class-based fallback for the layer direction (reusing the same class list `get_usage_type()` already uses) instead of hardcoding `AXIS2` when there's no material to read `LayerSetDirection` from, and
- a `get_surface_aligned_rotation()` helper that derives the rotation from the actual raycasted face normal, used only for hosts with no material layer set to read a flat elevation from.

Hosts with a real `IfcMaterialLayerSetUsage` (flat slabs, tilted "Horizontal Layers" roofs) keep the exact previous code path, so those working cases are unaffected by construction.

## Test plan
- [x] Live-verified in headless Blender: a default hipped roof's skylight now rotates to match the sloped face exactly (rotation-dot-face-normal = 1.0, height axis z = sin(slope)) and cuts an opening; wall windows stay vertical and flat-slab skylights stay flat (both still cut correctly).
- [x] `TestGetMaterialLayerParameters` (5 cases) + `test_get_surface_aligned_rotation.py` (9 cases) added; full `test/bim/module/model` + `test/tool` suites pass (pre-existing unrelated failures confirmed on the base branch).
- [x] black + ruff clean.

Thanks @sboddy for the report.

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/opening.py
+++ b/src/bonsai/bonsai/bim/module/model/opening.py
@@ -240,6 +240,33 @@ def _store_batch_in_cache(cache_key: tuple[int, str], batch: "gpu.types.GPUBatch
     _batch_cache[cache_key] = (epoch, batch)
 
 
+def get_surface_aligned_rotation(voided_obj_matrix_world: Matrix, local_normal: Vector) -> Matrix:
+    """Rotation that lays a filling flush against the host's surface at the
+    raycasted point, with its "up" axis following the surface's slope.
+
+    Generalises the old flat-slab assumption (host's own rotation, twisted
+    -90 degrees around X) to any face normal, so a filling rotates to match
+    a sloped face (eg. a parametric roof) rather than always coming out
+    horizontal. Reduces to the exact old result when ``local_normal`` is the
+    host's local +Z (flat slabs, and single-slope hosts whose whole object
+    is tilted, eg. a "Horizontal Layers" roof).
+    """
+    normal_matrix = voided_obj_matrix_world.to_3x3().inverted().transposed()
+    world_normal = (normal_matrix @ local_normal).normalized()
+
+    world_up = Vector((0.0, 0.0, 1.0))
+    tangent = world_up - world_up.dot(world_normal) * world_normal
+    if tangent.length < 1e-6:
+        tangent = Vector((0.0, 1.0, 0.0))
+    tangent.normalize()
+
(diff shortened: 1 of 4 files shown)


sboddy:
Got around to looking at this. Took me a while to figure out. Part of the confusion is because when I did figure it out, it aligns the window, applies the void, but makes the entire roof disappear. Tabbing into it shows the roof geometry is still there. It just doesn't show up once the window is applied.

Roof after window has been added. The window is aligned. But no roof geometry.
image

The window hole is there, and can be toggled visible. But no roof geometry.
image

This is after tabbing into item mode with the roof selected in the outliner:
image

This happens for both parametric roofs, and plain mesh roofs. Slight wrinkle with parametric roofs is that when you exit out of editing the parametric roof, the roof is regenerated, and loses the window connection. You have to re-add it to the roof object (which is still broken as above of course).

sboddy:
Weird, but even freshly made opening objects have the same effect on roof objects. I need to revert your PR from my build and see if the problem goes away.

sboddy:
OK, the issue predates your PR. I can't give confirmation that this PR fixes the original issue till I figure out why roof openings don't work. I'm looking into that.

sboddy:
This fix gets a 👍️ from me. I've tested it working.

For information only (no action on your part) the issue I ran into is #9349 / #9349 which was confusing the testing.

PR #8938, opened 2026-07-24
Bonsai: option to hide the sheet reference on section/elevation markers

### LINKED ISSUE #5872: Enhancement to optionally simplify section annotations
I'm not keen on the styling of the section markers like:
image
I don't need the sheet reference here as all sections are on this sheet, and this drawing is just a key for the drawing id's of the section views. They are all round kind of chunky, and fall down especially with 4 & 5 where the symbols clash.
I can remove the larger marker and reduce complexity as per Ace/IfcArchitect tutorial, to get:
image
Looking in the code the text and it's placement is hardcoded at https://github.com/IfcOpenShell/IfcOpenShell/blob/99c752904a517f9db893bd8866986b3376a2c854/src/bonsai/bonsai/bim/module/drawing/svgwriter.py#L676-L688
I wonder if it would be possible to place "anchors" in the symbols.svg that could be placed either in the section-tag or the section-arrow (transforming with the section-arrow as needed) and then have the text inserted in the optionally transformed anchor. If the anchor for the sheet reference is missing then it is not added. I'm trying to get to something like:
image
Of course I have the workaround of manually changing it in Inkscape, but that is not ideal in the event of revisions, as it would require repeat work if I have to recreate the key.

BIMvoice:
Fixes #5872.

Section and elevation markers always drew two text lines under the tag circle: the drawing number and the sheet reference. When every section lives on a single sheet (e.g. a key plan), the sheet reference is redundant, and the two-line layout crowds the small circular tags, causing them to clash, as reported by @theoryshaw.

Adds `EPset_Drawing.ShowSheetReferences` (default true, so existing drawings are unaffected). When disabled for a drawing, section/elevation tags show only the drawing number, vertically re-centered. Exposed as a per-drawing toggle in the Active Drawing panel, following the exact pattern of the sibling `has_underlay`/`has_linework`/`has_annotation` options.

This scopes to the literal request only. The later thread discussion (fully customizable tag anchors driven by symbols.svg, per-tag-type configurable fields) is a much larger design question and is intentionally not addressed here.

## Test plan
- [x] Live headless Blender: with the toggle on, both text lines render at y∓2.5 (unchanged); with it off, only the drawing number renders re-centered at y; default (no pset key) is true, so existing files are unaffected. Applies to both section and elevation markers.
- [x] `test/bim/module/drawing/test_svgwriter_section_markers.py` (4 tests) added; full drawing suite 34/34 pass.
- [x] black + ruff clean.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/ui.py
+++ b/src/bonsai/bonsai/bim/module/drawing/ui.py
@@ -73,6 +73,8 @@ class BIM_PT_camera(Panel):
         row = col.row(align=True)
         row.prop(props, "has_annotation", icon="MOD_EDGESPLIT")
         row.prop(dprops, "should_use_annotation_cache", text="", icon="FILE_REFRESH")
+        row = col.row(align=True)
+        row.prop(props, "show_sheet_references", icon="BOOKMARKS")
 
         # Drawing linked projects.
         row = col.row(align=True)
--- a/src/bonsai/bonsai/bim/data/pset/EPset_Drawing.ifc
+++ b/src/bonsai/bonsai/bim/data/pset/EPset_Drawing.ifc
@@ -5,7 +5,7 @@ FILE_NAME('EPset_Drawing.ifc','2020-01-01T00:00:00',$,$,'EPset_Drawing','EPset_D
 FILE_SCHEMA(('IFC4'));
 ENDSEC;
 DATA;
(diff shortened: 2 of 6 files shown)


sboddy:
I've tested this and quickly reviewed the code. It does what it says:
image

As stated customised positioning of the text is an additional feature and outside this scope.

One negative comment. I don't think this really deserves such a prominent UI element as given here:
image

Perhaps Show Sheet References should be a simple checkbox like Draw Linked Projects. Other than that I'd be happy to merge as a partial fix to #5872.

src/bonsai/bonsai/bim/module/drawing/ui.py
```diff
@@ -73,6 +73,8 @@ def draw(self, context):
         row = col.row(align=True)
         row.prop(props, "has_annotation", icon="MOD_EDGESPLIT")
         row.prop(dprops, "should_use_annotation_cache", text="", icon="FILE_REFRESH")
+        row = col.row(align=True)
+        row.prop(props, "show_sheet_references", icon="BOOKMARKS")
```
sboddy, line 77: As per comment in PR discussion, consider making this a standard small checkbox. This button is too "important looking" when compared to the other buttons (Underlay, Linework, Annotation).

sboddy (changes requested):

Issue #8870, opened 2026-07-23
Tracking: get all CI workflows green on v0.9.0 (priority ask)


BIMvoice:
**Rewritten for `v0.9.0` on 2026-08-30.** `v0.9.0` is the default branch since 2026-08-17, so this issue now tracks the workflows on that branch. The `v0.8.0` history is in the comments below. One honest note first: item 2 of the old checklist (the `choco_release.py` subprocess strings) sat here as "we will ship one" from 2026-07-25 until today. That is five weeks of a known, two-line fix left unshipped, and it is now #9383.

State of each red workflow on `v0.9.0`, verified against the latest run of each on 2026-08-30. Every fix below was reproduced RED and GREEN on a native `v0.9.0` build at `6f2e1aa993`, the head SHA of the `ci` run cited.

| workflow | latest run | cause | fix | status |
|---|---|---|---|---|
| `ci-lint` | [33304472327](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/33304472327) | black on 5 files and 4 ruff findings left by the unit-of-measure commits rebased onto the branch on 2026-08-28. Last green run was 2026-08-28. | #9385 | `ci-lint` is green on the PR |
| `ci` | [33304472332](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/33304472332) | five separate causes, see below | #9386 #9387 #9388 #9394 | awaiting review |
| `ci-bonsai-choco` | [33294965528](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/33294965528) | `subprocess.check_output("git status")` looks up a file literally named `git status` | #9383 | awaiting review |
| `ci-bonsai-daily` | 2026-08-31 22:35 run | #9402 (merged) verified in production: all ten `wrapped_data` failures cleared, including four export tests that were collateral. 11 remain: 1 ifcjson (`ifc2json4.py:124` `wrapped_data`, fix open at [Andrej730/IFC2JSON_python#1](https://github.com/Andrej730/IFC2JSON_python/pull/1), the branch the wheel builds from), 7 in the static-OCCT cluster blocked on #9341 (drawing serializer, sheet position, boundary `tree.select`), 2 linked-geom slice, 1 MEP precision difference | ifcjson PR + #9341 | in progress |
[...]

PR #8943, opened 2026-07-25
Fix ci-ifctester-org: install setuptools build dep

BIMvoice:
## Root cause

The `ci-ifctester-org` workflow has been red since 2026-07-24 (runs 30085027436, 30085007095, 30084856466, 30084679600, 30084568869). The `publish_website` job fails in `make webapp-build`, in the `webapp-stage-ifctester-wheel` target of `src/ifctester/Makefile`:

```
cd $(WEBAPP_IFCTESTER_BUILD_DIR) && $(PYTHON) -m venv env --system-site-packages && . env/$(VENV_ACTIVATE) && python -m pip install build && python -m build --wheel --no-isolation
```

`src/ifctester/pyproject.toml` declares `[build-system] requires = ["setuptools>=61.0"]`. The `--no-isolation` flag tells `build` not to create an isolated environment and not to fetch those declared build requirements, so they must already be present. The line only pip installs `build` itself, and the venv (`--system-site-packages`) was silently relying on the runner's system `setuptools`. That system `setuptools` no longer satisfies `>=61.0`, so the build fails with:

```
ERROR Missing dependencies:
	setuptools>=61.0
```

## Fix

Install `setuptools>=61.0` explicitly alongside `build` on that line, so the requirement is satisfied regardless of the runner's system packages:

```
python -m pip install build "setuptools>=61.0"
```

The `>=` is quoted since it is otherwise a shell redirection operator.

`wheel` is not added. I confirmed locally that modern `setuptools.build_meta` (>=61.0, tested with 80.9.0) declares an empty extra requirement set in `get_requires_for_build_wheel` and does not need the external `wheel` package to build a wheel. Only old `setuptools` (<61.0, tested with 60.10.0) declares `wheel` as an additional build requirement, so once `setuptools>=61.0` is guaranteed this is moot.

Dropping `--no-isolation` was also considered and rejected. It was likely chosen deliberately for build speed and to avoid a network fetch of the full isolated toolchain on every CI run, and changing it has broader blast radius than this fix needs. The minimal fix is to satisfy the one declared build requirement that `--no-isolation` expects to already be present.

I checked `src/common.mk`'s shared `dist` target (used by the other `src/*/Makefile` packages) and every other `src/*/Makefile` for the same `--no-isolation` plus incomplete `pip install` pattern. `src/ifctester/Makefile` line 103 is the only place in the repository using `--no-isolation`; `common.mk`'s `dist` target runs a plain `python -m build` (isolated), so it is not affected by this bug.

## Local reproduction

Reproduced by pip installing only `build` plus an old `setuptools` (60.10.0, with `wheel` also present to match the exact single-line error from the CI log) into a venv, then running `python -m build --wheel --no-isolation` against a staged copy of `src/ifctester`:

Before fix:
```
* Getting build dependencies for wheel...
running egg_info
...

ERROR Missing dependencies:
	setuptools>=61.0
```

After installing `build "setuptools>=61.0"` (same command as the fixed Makefile line) and rerunning:
```
* Getting build dependencies for wheel...
...
Successfully built ifctester-0.0.0-py3-none-any.whl
```

Also ran the actual `make webapp-stage-ifctester-wheel` target end to end with the Makefile fix applied (before the fix, `make webapp-stage-ifctester-wheel` fails at the same line; after the fix it completes and produces `webapp/public/worker/generated/ifctester-<version>-py3-none-any.whl` and `ifctester.json`).

This unblocks `ci-ifctester-org`, which has been red since at least 2026-07-24. For broader context see #8870, which tracks CI health across multiple workflows (this PR only fixes `ci-ifctester-org`, not the whole tracking issue).

## AI disclosure

This change was generated with the assistance of an AI coding tool, per AGENTS.md. It is a single-line Makefile fix with local reproduction evidence as shown above.

### DIFF
--- a/src/ifctester/Makefile
+++ b/src/ifctester/Makefile
@@ -100,7 +100,7 @@ else
 	$(SED) 's/version = "0.0.0"/version = "$(VERSION)a$(VERSION_DATE)"/' $(WEBAPP_IFCTESTER_BUILD_DIR)/pyproject.toml
 	$(SED) 's/version = "0.0.0"/version = "$(VERSION)-alpha$(VERSION_DATE)"/' $(WEBAPP_IFCTESTER_BUILD_DIR)/$(PACKAGE_NAME)/__init__.py
 endif
-	cd $(WEBAPP_IFCTESTER_BUILD_DIR) && $(PYTHON) -m venv env --system-site-packages && . env/$(VENV_ACTIVATE) && python -m pip install build && python -m build --wheel --no-isolation
+	cd $(WEBAPP_IFCTESTER_BUILD_DIR) && $(PYTHON) -m venv env --system-site-packages && . env/$(VENV_ACTIVATE) && python -m pip install build "setuptools>=61.0" && python -m build --wheel --no-isolation
 	mkdir -p $(WEBAPP_GENERATED_DIR)
 	wheel=$$(basename $(WEBAPP_IFCTESTER_BUILD_DIR)/dist/$(PACKAGE_NAME)-*.whl); \
 	cp "$(WEBAPP_IFCTESTER_BUILD_DIR)/dist/$$wheel" "$(WEBAPP_GENERATED_DIR)/$$wheel"; \


BIMvoice:
@sboddy this is a one line change with all three checks green, but it has become noisier than its size suggests, so flagging it directly.

`ci-ifctester-org` currently fails on every push to the repo. It ran 19 times today and failed every time. The `publish_website` job dies in `make webapp-build`:

```
ERROR Missing dependencies:
	setuptools>=61.0
make: *** [Makefile:98: webapp-stage-ifctester-wheel] Error 1
```

Cause: that step builds the ifctester webapp wheel with `python -m build --wheel --no-isolation`. The `--no-isolation` flag means `build` does not fetch the build requirements declared in `src/ifctester/pyproject.toml` (`requires = ["setuptools>=61.0"]`), so they have to already be present. The venv only installs `build` itself, and it had been quietly relying on the runner image's system setuptools, which no longer satisfies `>=61.0`.

The fix installs the declared requirement alongside `build`, one line in `src/ifctester/Makefile`.

Verified locally rather than by inspection: reproduced the exact CI error by installing `build` with setuptools 60.10.0 and running the same command against the staged `src/ifctester`, then confirmed the wheel builds after adding the dependency. Also checked the rest of the repo, this is the only `--no-isolation` build, so there are no sibling copies of the same bug.

Being straight about our part in the volume: a good share of today's 19 runs came from us rebasing a batch of stale branches this morning, since each push triggers the workflow. That is exactly why we would rather get this merged than keep generating failure mail for everyone watching the repo. We have paused further rebasing until it lands.

BIMvoice:
Flagging a merge-ordering dependency before anyone picks this up.

This PR fixes the build failure. On its own that looks like a plain CI fix, but it removes something that is currently acting as an accidental guard.

`ci-ifctester-org.yml` triggers on:

```yaml
on:
  workflow_dispatch:
  push:
    paths:
      - src/ifctester/**
```

There is no `branches:` filter. Any push to any branch touching `src/ifctester/**` runs the workflow, and the job checks out `IfcOpenShell/ifctester_org_static_html` with a write token and commits the built webapp into it.

Today that never reaches the publish step, because the build dies first on the `setuptools>=61.0` error this PR fixes. The broken build is the only reason a feature branch has not already published to the live ifctester.org.

So merging this PR by itself removes that guard.

#8948 adds the missing branch restriction and is open, green and mergeable. Recommend merging #8948 first, or both together. If only one lands now, #8948 is the safer one: it stops both the unintended publishes and the workflow firing on unrelated branches, whereas this PR only changes whether the build succeeds.

Observed today: this workflow fired on several unrelated feature branches after they merged `v0.8.0`, including `fix-7318-bcf-gp-strokes` and `fix-reimport-multiprocessing-toggle`, neither of which touches ifctester in any meaningful way.

BIMvoice:
Closing in favour of @falken10vdl's #7920, which fixes the same thing in the same file and predates this by three and a half months.

The only difference is that ours pinned `setuptools>=61.0` rather than leaving it unversioned. If that pin is worth having, it belongs as a suggestion on his PR, not as a second PR.

Found by auditing our own open PRs against other contributors' after @theoryshaw pointed out we were duplicating work.

PR #8955, opened 2026-07-25
Add missing tool descriptions to ifcmcp server

### LINKED ISSUE #8951: ifcmcp: 23 of 25 tools expose an empty description over MCP
### Bug Description

`ifcopenshell-mcp` 0.8.5 registers 25 tools, but 23 of them return an empty `description` field in the `tools/list` response. MCP clients inject that field directly into the model's context, so for those 23 tools the model receives only a bare name and the parameter schema, with no statement of what the tool does.

## Cause

The tool functions in `ifcmcp/server.py` are thin delegates without docstrings. `FastMCP.tool()` derives a tool's description from the function docstring, so an absent docstring becomes an empty description.

```python
def build_server() -> Any:
    server = FastMCP(...)

    @server.tool()
    def ifc_load(path: str) -> str:
        return session.ifc_load(path)   # no docstring -> description == ""
```

Only `ifc_plot` and `ifc_render` carry docstrings, and they are the only two tools whose descriptions are non-empty, which matches the observed behaviour exactly.

## Affected tools (23)

`ifc_clash`, `ifc_contexts`, `ifc_cost`, `ifc_docs`, `ifc_edit`, `ifc_info`, `ifc_list`, `ifc_load`, `ifc_materials`, `ifc_new`, `ifc_quantify`, `ifc_relations`, `ifc_reset`, `ifc_save`, `ifc_schedule`, `ifc_schema`, `ifc_select`, `ifc_shape`, `ifc_shape_docs`, `ifc_shape_list`, `ifc_summary`, `ifc_tree`, `ifc_validate`

Unaffected: `ifc_plot` (1192 chars), `ifc_render` (808 chars).

## To reproduce

1. Start the server over stdio: `python -m ifcmcp`
2. Send an `initialize` request, then `tools/list`
3. Inspect the `description` field of each returned tool

Every tool other than `ifc_plot` and `ifc_render` comes back with `"description": ""`.

## Expected behaviour

Each tool returns a description explaining what it does, as `ifc_plot` and `ifc_render` already do.

## Impact

Tool descriptions are the primary signal a model uses to choose between tools and to interpret their arguments. Without them the model has to infer intent from the name alone, which is particularly error-prone for the closely related tools in this set: `ifc_quantify` vs `ifc_schedule` vs `ifc_cost`, or `ifc_info` vs `ifc_summary` vs `ifc_tree`. In practice this lowers tool-selection accuracy on real IFC workflows.

## Suggested fix

Add a docstring to each tool function. `FastMCP` picks these up automatically, so no other change is required. A short first line stating what the tool returns, plus a note on the meaning of each parameter, would be sufficient.

I'm happy to open a PR adding docstrings for all 23 tools if that's welcome. Please let me know whether you'd prefer them written in the same style as the existing `ifc_plot` / `ifc_render` docstrings.

### Attachments

Not applicable - this is reproducible with any IFC file, or with no file loaded at all, since it concerns the tool listing rather than tool behaviour.

### Debug and Error Output

```true
ifcopenshell-mcp : 0.8.5
mcp (Python SDK) : 1.28.1  (also reproduced with 1.14.0)
Python           : 3.11
OS               : Windows 11
Transport        : stdio

Measured description lengths from the tools/list response:

  ifc_plot     1192 chars
  ifc_render    808 chars
  all 23 others   0 chars

Cross-checked against the source: an AST scan of ifcmcp/server.py finds 25
functions decorated with @server.tool(), of which 23 have no docstring. The two
that do are ifc_plot and ifc_render, matching the runtime result exactly.
```

mtngrj:
Closes #8951.

### What

`build_server()` in `src/ifcmcp/ifcmcp/server.py` registers 25 tools. 23 of them are advertised to MCP clients with an empty `description` — only `ifc_plot` and `ifc_render` carry one. This adds a docstring to each of the remaining 23 wrappers, plus a test that fails if any registered tool has an empty description.

### Why this is a functional defect, not missing documentation

FastMCP derives each tool's MCP `description` from the wrapper function's `__doc__` (`Tool.from_function` in `mcp/server/fastmcp/tools/base.py`). Without a docstring the tool reaches the client as a bare name, so a client has no way to tell apart, for example:

- `ifc_quantify` — **writes**: derives quantities from geometry and adds/updates `IfcElementQuantity` sets
- `ifc_schedule` — reads `IfcWorkSchedule` (the time dimension)
- `ifc_cost` — reads `IfcCostSchedule` (the cost dimension)

Nor can it know which tools require a loaded model, or which mutate the model in memory and therefore need `ifc_save` afterwards.

Reproducible with:

```python
from ifcmcp.server import build_server
tools = build_server()._tool_manager.list_tools()
print([t.name for t in tools if not (t.description or "").strip()])
```

### Style

Docstrings follow `ifc_plot` and `ifc_render`: a one-line summary, a short prose paragraph covering the return value and any notable behaviour, then one `:param name:` entry per parameter in reST. No `:return:`, since neither existing docstring uses one. Per the discussion in #8951, the `AGENTS.md` note about not adding docstrings to untouched code is not read as applying here — the descriptions are the fix.

The added test mirrors `test_well_documented_methods_have_descriptions` in `tests/test_shape.py`.

### Scope and verification

Docstrings only, in one file, plus one test. Verified mechanically rather than by inspection:

- `git diff --numstat` → `250 0` for `server.py` (no deleted lines) and `6 0` for the test
- parsing the before/after files with `ast` and stripping all docstrings yields **identical** syntax trees; 23 docstrings added, none modified or removed. `ifc_plot` and `ifc_render` are byte-for-byte unchanged
- the added test was checked against the unmodified file: it fails there and names all 23 tools, so it is not a tautology

Content was written from the implementations in `core.py`, `ifcquery` and `ifcedit` rather than inferred from the tool names. One item could not be established from the source and is deliberately not documented: whether `ifc_relations(traverse=...)` was intended to accept anything besides `"up"`. The accepted values for `ifc_new(schema=...)` were left unspecified in the first commit; a second commit (`+2/-1`, docstring only -- the `ast` check above describes the first commit) now names them, after testing showed that `ifcopenshell.file()` accepts `IFC2X3`, `IFC4`, `IFC4X1`, `IFC4X2` and `IFC4X3`. Note that `ifcopenshell.util.schema.IFC_SCHEMA` is still `Literal["IFC2X3", "IFC4", "IFC4X3"]`; happy to narrow the docstring to those three if that annotation is the intended contract. The `ifc_select` syntax examples are reused from the docstring in `ifcquery/select.py`.

Checks on this branch:

| check | result |
|---|---|
| `black --check` (root `pyproject.toml`, line-length 120) | clean, longest line 116 |
| `ruff check ifcmcp/server.py` | clean |
| `pytest src/ifcmcp/tests` | 71 passed |

### Unrelated observation

While reading the implementation: `AVAILABLE_RULES` in `src/ifcedit/ifcedit/quantify.py` is only read by `list_rules()`, which has no callers, and it lists 2 of the 4 rule keys actually present in `ifc5d.qto.rules`. Left untouched here to keep this change to docstrings — happy to open a separate issue.

### AI disclosure

The docstrings, the added test and this description were drafted with the assistance of an AI coding tool, and checked against the implementation and the test suite before submission.

### DIFF
--- a/src/ifcmcp/ifcmcp/server.py
+++ b/src/ifcmcp/ifcmcp/server.py
@@ -41,8 +41,11 @@ def build_server() -> Any:
         edits.  The new model has no file path of its own, so ``ifc_save`` must be
         given an explicit path.
 
-        :param schema: IFC schema version, passed straight to
-            ``ifcopenshell.file()`` (default ``IFC4``).
+        :param schema: IFC schema version — ``IFC2X3``, ``IFC4``, ``IFC4X1``,
+            ``IFC4X2`` or ``IFC4X3`` — passed straight to
+            ``ifcopenshell.file()`` (default ``IFC4``).  ``IFC4X3`` and
+            ``IFC4X3_ADD2`` are both accepted and resolve to the same
+            schema identifier, ``IFC4X3_ADD2``.
         """
         return session.ifc_new(schema=schema)
 
--- a/src/ifcmcp/ifcmcp/server.py
+++ b/src/ifcmcp/ifcmcp/server.py
@@ -35,39 +35,126 @@ def build_server() -> Any:
     # ---- Lifecycle ----
     @server.tool()
     def ifc_new(schema: str = "IFC4") -> dict[str, Any]:
+        """Create a new empty IFC model in memory.
+
+        Replaces the model currently held by the session, discarding any unsaved
+        edits.  The new model has no file path of its own, so ``ifc_save`` must be
+        given an explicit path.
+
+        :param schema: IFC schema version, passed straight to
+            ``ifcopenshell.file()`` (default ``IFC4``).
+        """
         return session.ifc_new(schema=schema)
 
     @server.tool()
     def ifc_load(path: str) -> str:
+        """Open an IFC file from disk into the session.
+
+        Replaces the model currently held by the session, discarding any unsaved
+        edits, and remembers the path so a later ``ifc_save`` can overwrite it.
+        Call this before any query or edit tool.  Returns a confirmation string
+        naming the schema version and entity count.
+
+        :param path: Filesystem path of the IFC file to open.
+        """
         return session.ifc_load(path)
 
     @server.tool()
     def ifc_save(path: str = "") -> str:
+        """Write the in-memory model to disk.
+
+        Overwrites the target file without further confirmation.  Edits made by
+        ``ifc_edit``, ``ifc_shape`` and ``ifc_quantify`` exist only in memory
+        until this is called.
+
+        :param path: Destination path.  Omit to overwrite the file the model was
+            loaded from; this fails for a model created by ``ifc_new``, which has
+            no original path.
+        """
         return session.ifc_save(path)
 
     @server.tool()
     def ifc_reset() -> dict[str, Any]:
+        """Discard the in-memory model.
+
+        Drops the model and its file path, throwing away any edits not already
+        written with ``ifc_save``.  Succeeds even when no model is loaded.
+        """
         return session.ifc_reset()
 
     # ---- Query ----
     @server.tool()
     def ifc_summary() -> dict[str, Any]:
+        """Summarise the loaded model: schema, entity counts and project info.
+
+        Returns the ``schema`` version, ``total_entities``, a ``types`` mapping of
+        IFC class name to instance count in descending order, and a ``project``
+        block with the id, name and description of the first ``IfcProject``.  The
+        counts cover every entity in the file, not just physical elements.  This
+        is usually the first tool to call after ``ifc_load``.
+        """
         return session.ifc_summary()
 
     @server.tool()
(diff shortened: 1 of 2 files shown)


brunopostle:
@aothms I had a look in #8951 and it is all self contained, so won't break anything outside the MCP tools, I haven't tried applying the patchset and running though, not sure when I'll get a chance to do this

brunopostle:
Thanks @mtngrj your PR works fine, but since ifcmcp is basically just a wrapper around the ifcedit and ifcquery tools it makes sense to put these changes there and then fix the inheritance so that the documentation is the same everywhere. I've created another pull-request #9229

mtngrj:
Closing in favour of #9229, which fixes the same bug at the source rather than duplicating the descriptions across three layers. Thanks @brunopostle for reworking it, and @BIMvoice for the runtime checks along the way.

Issue #7305, opened 2025-10-30
Falha de importação de arquivo IFC com Bonsai


CleitonMohr:
### Bug Description

1.1 Ao tentar abrir um arquivo .ifc no sistema operacional Ubuntu 24.04 LTS, seguindo os seguintes passos: File/Import/IFC (...), o arquivo não abre e aparece a mensagem abaixo, em 2.0
1.2 Tentei abrir o mesmo arquivo em uma versão mais atual do Blender no Ubuntu, mas o arquivo abril em uma versão anterior (que não me recordo qual) no Window 11, então fiz o downgrade do blender no Ubuntu torcendo para cer apenas uma incompatibilidade da versão, mas o problema persistiu, como podem ver.
1.3 O arquivo .ifc veio do software Archstation ainda do ano passado (2024), e não tenho mais acesso ao software.

2.0 Mensagem de erro:
```
os: Linux
os_version: #34~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Sep 23 15:35:20 UTC 2
python_version: 3.11.9
architecture: ('64bit', 'ELF')
machine: x86_64
processor: x86_64
blender_version: 4.3.1
bonsai_version: 0.8.3-post1
bonsai_commit_hash: None
bonsai_commit_date: None
last_actions: 
last_error: Traceback (most recent call last):
  File "/home/cleiton/.config/blender/4.3/extensions/.local/lib/python3.11/site-packages/bonsai/bim/module/project/operator.py", line 1014, in finish_loading_project
    bpy.ops.bim.load_project_elements()
  File "/snap/blender/5647/4.3/scripts/modules/bpy/ops.py", line 109, in __call__
    ret = _op_call(self.idname_py(), kw)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Error: Python: Traceback (most recent call last):
  File "/home/cleiton/.config/blender/4.3/extensions/.local/lib/python3.11/site-packages/bonsai/bim/module/project/operator.py", line 1103, in execute
    ifc_importer.execute()
  File "/home/cleiton/.config/blender/4.3/extensions/.local/lib/python3.11/site-packages/bonsai/bim/import_ifc.py", line 272, in execute
    self.create_elements()
  File "/home/cleiton/.config/blender/4.3/extensions/.local/lib/python3.11/site-packages/bonsai/bim/import_ifc.py", line 590, in create_elements
[...]

PR #8962, opened 2026-07-25
IfcConvert: fix --unicode=escape referencing a removed enumerator

BIMvoice:
`IfcConvert.cpp` sets the decoder mode for `--unicode=escape` to `IfcParse::IfcCharacterDecoder::JSON`:

```cpp
} else if (unicode_mode == "escape") {
    IfcParse::IfcCharacterDecoder::mode = IfcParse::IfcCharacterDecoder::JSON;
```

`JSON` does not exist. `IfcCharacterDecoder.h` declares only `SUBSTITUTE`, `UTF8` and `ESCAPE`.

How it survived: the flag was added in `fa76aa9bb0` (2017) when the enumerator was still called `JSON`. `0b2bd5d0e8` ("Eliminate ICU dependency", 2019) renamed it to `ESCAPE` but missed this call site, because the site sits inside `#ifdef HAVE_ICU` and that same commit stopped defining `HAVE_ICU`. No CI workflow defines it either, so the branch has not been compiled in about six years and no build has flagged it.

The consequence is small but real: anyone re-enabling ICU support gets a compile error on a flag `--unicode=utf8|escape` that the help text still advertises. `ESCAPE` is unambiguously the intended value given the branch condition is `unicode_mode == "escape"`.

One character of behaviour change, no runtime effect on any current build.

Found while investigating #7305, which is a separate and much larger encoding question that I am not addressing here.

### DIFF
--- a/src/ifcconvert/IfcConvert.cpp
+++ b/src/ifcconvert/IfcConvert.cpp
@@ -529,7 +529,7 @@ int main(int argc, char** argv) {
         if (unicode_mode == "utf8") {
             IfcParse::IfcCharacterDecoder::mode = IfcParse::IfcCharacterDecoder::UTF8;
         } else if (unicode_mode == "escape") {
-            IfcParse::IfcCharacterDecoder::mode = IfcParse::IfcCharacterDecoder::JSON;
+            IfcParse::IfcCharacterDecoder::mode = IfcParse::IfcCharacterDecoder::ESCAPE;
         } else {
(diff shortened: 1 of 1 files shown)


BIMvoice:
Closing: already fixed on `v0.9.0`. `src/ifcconvert/IfcConvert.cpp:498` now sets `runtime_character_decoder::mode = ESCAPE`, the enumerator this PR was adding. Nothing to carry over.

Issue #4481, opened 2024-03-29
Improve the speed of switching between drawings


theoryshaw:
This is probably in the roadmap, but thought i'd create an issue to track. 

Would it be possible to improve the speed it takes to switch between drawings?

As a random example, the following file takes +/-20 seconds to switch from one drawing to another. 
https://hub.openingdesign.com/OpeningDesign/Highland_Haven/src/commit/9218e84934852facbca0120f71ffcba75d6be1f9/Open/Models_and_CAD/BlenderBIM/Highland_Haven.ifc

Just wondering if there's a clever way to improve this time?  

If not, i wonder if there could be a way to have multiple drawings 'open' at one time.

Moult:
FYI https://github.com/IfcOpenShell/IfcOpenShell/commit/c8706bddc9e66183d7a34bc3098a1db0a9414c75 should result in a slight speed up on the first drawing you activate only.

Moult:
@theoryshaw can you test again to see what numbers you get now? Note that if you open the IFC, the first time you activate a drawing will be slow (it needs to fetch data about other potentially referenced drawings) but switching between drawings after should be faster.

theoryshaw:
before https://github.com/IfcOpenShell/IfcOpenShell/commit/c8706bddc9e66183d7a34bc3098a1db0a9414c75 

1st time - SECTION - EAST - KITCHEN: 25 secs
1st time - SECTION - NORTH - KITCHEN: 19 secs
2nd time - SECTION - EAST - KITCHEN:  18 secs
2nd time - SECTION - NORTH - KITCHEN:   15 secs
3rd time - SECTION - EAST - KITCHEN:  18 secs
3rd time - SECTION - NORTH - KITCHEN:  16 secs

at df4943205aec8dc96236a73583c30d6b135f3df3

1st time - SECTION - EAST - KITCHEN: 17 secs
1st time - SECTION - NORTH - KITCHEN:  8 secs
2nd time - SECTION - EAST - KITCHEN:   7 secs
2nd time - SECTION - NORTH - KITCHEN:   5 secs
3rd time - SECTION - EAST - KITCHEN:   6 secs
3rd time - SECTION - NORTH - KITCHEN:  5 secs

An improvement! :)

theoryshaw:
Logging this here, as it might provide additional ideas to improve the speed...

https://claude.ai/share/2ce5252e-df5e-46ac-987b-49fbdc1f23e7

[...]

PR #8971, opened 2026-07-25
Bonsai: batch geometry iterator setup when reimporting many representations

### LINKED ISSUE #5696: Improved speed of `bim.activate_model` command

BIMvoice:
Fixes #5696

## Problem

`ActivateModel.execute` (switching back to the model view) calls `switch_representation` once for every representation that needs reimporting, and `tool.Geometry.reimport_element_representations` builds a fresh `ifcopenshell.geom.iterator` on every call. `iterator.initialize()` pays a large fixed setup cost over the whole file regardless of how many elements it actually covers, so N representations paid that setup cost N times.

@sboddy's profiling on Highland Haven pinned this precisely: 87 iterator constructions accounted for 26.1 of the model's 38 second reimport, and a quick hack that limited iterator calls got roughly a 10x speedup. @theoryshaw reported the issue and supplied both benchmark models.

## Fix

Added `reimport_element_representations_batch`, which groups the requested representations by `(context, apply_openings)` (elements needing different settings genuinely can't share an iterator) and runs one geometry iterator per group instead of one per representation. Shapes are dispatched back to the representation that requested them the same way the existing single-representation code already tolerates shapes it didn't ask for, just generalized from one target representation to a mapping.

`reimport_element_representations` (the single-item entry point used by `parametric_lifecycle.py`, `misc/operator.py`, `type/operator.py`, `drawing/workspace.py`, and elsewhere) now just calls the batch function with a list of one, so those callers are unaffected. Only `ActivateModel.execute`, the loop that was actually constructing many iterators back to back, is switched to the new batched entry point via `switch_representations`.

Related to #4481 per @Moult's note that the two share code (drawing switch speed). This PR only touches the iterator-per-representation path; it does not touch the visibility/hide_view_set work discussed there.

## Measurements

Instrumented `ifcopenshell.geom.iterator` construction and `initialize()` calls directly (not through the full `bim.activate_model` operator, which needs an elaborate drawing setup to time meaningfully). Reimported 80 body representations picked from each model.

Highland Haven (32MB, IFC4, 11028 IfcProducts):
| | iterators | total initialize() |
|---|---|---|
| before | 80 | 15.8s |
| after | 1 | 0.18s |

Verona 3 Season Porch (15MB):
| | iterators | total initialize() |
|---|---|---|
| before | 80 | 3.2s |
| after | 1 | 0.04s |

Separately, measured through the actual operator on Highland Haven: `bim.activate_model` 9.35s to 1.58s, iterator construction 37 to 1, `initialize()` 8.03s to 0.24s.

## Correctness

For both models, captured per-object mesh vertex counts and bounding boxes for all 80 reimported objects, comparing the unpatched code against the batched code. All 80/80 objects matched bit-for-bit on both models (zero mismatches).

## Tests

`pytest -p no:pytest-blender test/core`: 390 passing before, 392 passing after (added `TestSwitchRepresentations` covering the new batched core function; also filled in a pre-existing gap where `is_text_literal` was implemented but missing from the `Geometry` core interface declaration, needed for the mock to work).

## Drawing activation (added)

@theoryshaw tested this PR against his own report and saw no improvement, because activating a drawing goes through `Drawing.activate_drawing`, a separate operator this PR did not originally touch: it also called `switch_representation` once per element.

Extended the same batching to that path. Measured on Highland Haven, activating drawing "FLOOR PLAN - 1ST - 4 UNITS": iterator count 17 to 10, `initialize()` 0.43s to 0.24s. Correctness verified: 7514/7514 objects identical on vertex count, bounding box and active representation id versus the unbatched code.

Honest framing: this is a real but small fix, on the order of 2% of the roughly 9-10 second wall clock for that action in this environment. Profiling showed the actual bottleneck for drawing activation is `ifcopenshell.util.selector.filter_elements` and `ifcopenshell.util.element.get_psets` evaluating the drawing's Include/Exclude psets, which this PR does not touch. That is addressed separately.

## AI disclosure

This PR was generated with the assistance of an AI coding tool, per AGENTS.md. See the commit message for the same note.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/operator.py
+++ b/src/bonsai/bonsai/bim/module/drawing/operator.py
@@ -2436,13 +2436,11 @@ class ActivateModel(bpy.types.Operator):
             return refined_elements
 
         refined_elements = refine_elements(elements)
-        for _, (model, obj) in refined_elements.items():
-            bonsai.core.geometry.switch_representation(
-                tool.Ifc,
-                tool.Geometry,
-                obj=obj,
-                representation=model,
-            )
+        bonsai.core.geometry.switch_representations(
+            tool.Ifc,
+            tool.Geometry,
+            items=[(obj, model, True) for _, (model, obj) in refined_elements.items()],
+        )
 
         tool.Blender.reset_object_visibility()
         tool.Drawing.hide_all_drawing_collections()
--- a/src/bonsai/bonsai/core/geometry.py
+++ b/src/bonsai/bonsai/core/geometry.py
@@ -129,6 +129,28 @@ def switch_representation(
     geometry.reimport_element_representations(obj, representation, apply_openings=apply_openings)
 
 
+def switch_representations(
+    ifc: type[tool.Ifc],
+    geometry: type[tool.Geometry],
+    items: list[tuple[bpy.types.Object, ifcopenshell.entity_instance, bool]],
+) -> None:
+    """Switch representations for multiple objects in a single batched geometry pass.
+
+    Equivalent to calling switch_representation for each (obj, representation, apply_openings)
+    item, but the underlying reimport is batched so representations that share a context and
+    apply_openings setting only pay for one geometry iterator between them. See #5696.
+    """
+    pending: list[tuple[bpy.types.Object, ifcopenshell.entity_instance, bool]] = []
+    for obj, representation, apply_openings in items:
+        if not geometry.get_object_data(obj) and geometry.is_text_literal(representation):
+            continue
+        element = ifc.get_entity(obj)
+        assert element
+        geometry.clear_cache(element)
+        pending.append((obj, representation, apply_openings))
+    geometry.reimport_element_representations_batch(pending)
+
+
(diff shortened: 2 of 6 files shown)


BIMvoice:
Recording @theoryshaw's confirmation here so it is visible on the pull request rather than only on the issue.

On #5696, 2026-07-27: "i can confirm your proposed #8971 reduces the `bim.activate_model` time! Thank you."

Measured on `Highland_Haven.ifc`: `bim.activate_model` 9.35s to 1.58s, iterators built 37 to 1, `initialize()` 8.03s to 0.24s.

Note the drawing-activation extension in this PR is a much smaller win, iterators 17 to 10 and about 2 percent of wall time, because geometry setup is not the bottleneck there. That path is addressed separately in #9022.

PR #8994, opened 2026-07-26
Agent/fix ms project task lags

### LINKED ISSUE #4185: xml import from MS Project to BlenderBIM is loosing it's time spaces between tasks
hello,
when exporting a planning from MS Project (version 2311 - build 17029.20108 - Dutch) via xml to BlenderBIM (0.0.231104) the timelags between tasks (seem to) disappear. 
Blenderbim missing time lag
Blenderbim time lag manual added
msproject

PahlawanJoey:
Fixes #4185 

Problem

When importing Microsoft Project XML, msp2ifc.py created the predecessor relationships as IfcRelSequence, but ignored the LinkLag and LagFormat values inside each PredecessorLink.

As a result, the relationship type was imported correctly, but no IfcLagTime was assigned. Any delay or lead between predecessor and successor tasks therefore disappeared from the IFC schedule.

Root cause

Microsoft Project stores relationship lag in LinkLag, normally in tenths of a minute. LagFormat determines how that value should be interpreted, including:

* Working time based on the project calendar
* Elapsed time using a 24-hour day
* Percentage lag
* Estimated variants of these formats

Neither LinkLag nor LagFormat was previously processed by the importer.

Fix

parse_relationship_xml() now reads and converts both values.

For duration-based lags:

* Working-time lags use the project’s MinutesPerDay, falling back to 480 minutes.
* Elapsed-time lags use 1,440 minutes per day.
* The value is converted from tenths of a minute into an ISO duration string.
* Negative values are preserved as leads.
* Zero and null lag values do not create an IfcLagTime.

The duration is constructed directly as an ISO string instead of passing an abstract fractional-day timedelta. This is important for sub-day working-time lags. For example, with an eight-hour working day, a four-hour lag must be stored as PT4H. Converting it through timedelta(days=0.5) would incorrectly serialise it as PT12H.

The elapsed and estimated elapsed LagFormat variants are mapped to ELAPSEDTIME; other duration formats use WORKTIME.

Percentage formats are stored as IfcRatioMeasure, which is supported by IfcLagTime. For example, a 50% lag is stored as a ratio of 0.5. assign_lag_time() was extended accordingly while retaining support for ISO duration strings and datetime.timedelta.

Finally, create_rel_sequences() now calls ifcopenshell.api.sequence.assign_lag_time() whenever a parsed lag is present.

Validation

Regression coverage includes:

* Working-time lag
* Elapsed-time lag
* Negative lag or lead
* Zero lag
* Sub-day working-time lag
* Percentage lag
* IfcRatioMeasure assignment in IFC4 and IFC4X3

The original XML file from issue #4185 was also imported successfully:

* 811 tasks
* 974 predecessor relationships
* 138 non-zero lags
* 135 WORKTIME lags
* 3 ELAPSEDTIME lags
* 3 negative lags

Black and Ruff pass, the dedicated importer tests pass, and all 68 sequence API tests pass.

This change focuses on correctly preserving the relationship lag in IFC. Existing downstream scheduling functions that operate at whole-day granularity are outside the scope of this fix.

### DIFF
--- a/src/ifc4d/ifc4d/msp2ifc.py
+++ b/src/ifc4d/ifc4d/msp2ifc.py
@@ -64,9 +64,23 @@ class MSP2Ifc:
         id = 0
         if task.findall("pr:PredecessorLink", self.ns):
             for relationship in task.findall("pr:PredecessorLink", self.ns):
+                lag_format = relationship.findtext("pr:LagFormat", namespaces=self.ns)
+                link_lag = relationship.findtext("pr:LinkLag", namespaces=self.ns)
+                duration_type = (
+                    "ELAPSEDTIME"
+                    if lag_format in {"4", "6", "8", "10", "12", "36", "38", "40", "42", "44"}
+                    else "WORKTIME"
+                )
+                hours_per_day = (
+                    24 if duration_type == "ELAPSEDTIME" else int(self.project["MinutesPerDay"] or 8 * 60) / 60
+                )
                 relationships[id] = {
                     "PredecessorTask": relationship.find("pr:PredecessorUID", self.ns).text,
                     "Type": relationship.find("pr:Type", self.ns).text,
+                    "Lag": (
+                        timedelta(days=int(link_lag) / 10 / 60 / hours_per_day) if link_lag and int(link_lag) else None
+                    ),
+                    "LagDurationType": duration_type,
                 }
                 id += 1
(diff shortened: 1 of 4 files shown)


PahlawanJoey:
@Moult old bug, looking into the 4D functions myself so was looking at bugs and looking to contribute something.

Issue #7055, opened 2025-08-24
Regen adjusts IfcSpace but saving afterwards does not save geometry


bgrundeken:
### Bug Description

Using Bonsai v0.8.4-alpha250821-dccd9d4, Blender 4.4.3, on MacOS Sequioa 15.6.

When I regenerate a space it resizes the geometry correctly. However, if I then save the file and reopen it, the quantity sets are still correct but the geometry is back to the pre-regeneration shape.

To reproduce, use the attached ifc.

1. Select the OFFICE space (note that its geometry spills over into what should be the BATHROOM space)
2. Regen (Geometry now correct)
3. Save ifc
4. Revert ifc
5. Geometry again spills over into the BATHROOM space.

### Attachments

[ifc.txt](https://github.com/user-attachments/files/21956675/ifc.txt)

### Debug and Error Output

```true

```

theoryshaw:
Can confirm. 

...

As a workaround in the meantime, after regenning, you can use the following to bake the change back into the .ifc file...

<img width="326" height="650" alt="Image" src="https://github.com/user-attachments/assets/6048f2b1-af3c-4bec-9d98-9e1093ffc69e" />

bgrundeken:
@theoryshaw thanks for picking this up. 

I ended up deleting the spaces and adding them again, but great that there is a workaround.

Like mentioned in #7056 I think it would be a good idea to have the save button just save my stuff without having to click another button first.

PR #8999, opened 2026-07-26
Fix space regen doubling Z location

CyrilWaechter:
## Summary

When regenerating an existing `IfcSpace` via the Spatial tool (`Shift+G`), the space was placed at double its Z elevation (e.g. 18.98m instead of 9.49m).

## Root cause

Commit `d8de62308` ("Fix space regen not saving geometry to IFC (#7055)") rewrote the space regeneration to use `ShapeBuilder` to write geometry in the object's **local** space (via `get_2d_vertices_from_polygon` which transforms world→local using `obj.matrix_world.inverted()`), then reload via `switch_representation`. This preserves `obj.matrix_world` — the object is already at the correct world position after the representation is rebuilt.

However, the old `spatial.translate_obj_to_z_location(active_obj, z)` call was left in place. This call adds `z` (the object's current world Z from `get_x_y_z_h_mat_from_obj`) on top of the already-correct `obj.location.z`, producing `2 * z`.

Under the previous bmesh-based approach, geometry was built in global coordinates at Z=0, so the translate was necessary to lift the space to the right elevation. With the local-space approach it became redundant.

## Fix

Remove the `spatial.translate_obj_to_z_location(active_obj, z)` call from the existing-`IfcSpace` regeneration branch in `bonsai/core/spatial.py`. The new-space creation branch still needs it (the object starts at Z=0).

## Test

Adds `test_regenerate_space_preserves_z_location` to `test/tool/test_spatial.py`, covering the existing-space regeneration path (previously untested). The test:
1. Creates a wall and generates a space at cursor
2. Moves the space to Z=5 and updates the view layer
3. Selects the space as active object (triggers regeneration branch)
4. Calls `bpy.ops.bim.generate_space()` again
5. Asserts `space.location.z` is still 5 (not 10)

The test uses a non-zero Z because `translate_obj_to_z_location` has a `if z != 0` guard — the bug only manifests at non-zero elevations.

### DIFF
--- a/src/bonsai/bonsai/core/spatial.py
+++ b/src/bonsai/bonsai/core/spatial.py
@@ -222,7 +222,6 @@ def generate_space(
 
     if element and element.is_a("IfcSpace"):
         spatial.set_space_representation_from_polygon(active_obj, element, space_polygon, h, polygon_is_si=True)
-        spatial.translate_obj_to_z_location(active_obj, z)
     else:
         if relating_type:
             name = model.generate_occurrence_name(relating_type, "IfcSpace")
--- a/src/bonsai/test/tool/test_spatial.py
+++ b/src/bonsai/test/tool/test_spatial.py
@@ -288,3 +288,27 @@ class TestGenerateSpace(NewFile):
             )
         )
         assert np.allclose(TEST_VERTS, sorted([tuple(v.co) for v in mesh.vertices]))
+
+    def test_regenerate_space_preserves_z_location(self):
+        bpy.ops.bim.create_project()
+        ifc = tool.Ifc.get()
+        scene = bpy.context.scene
+        product = ifcopenshell.api.root.create_entity(ifc, ifc_class="IfcWall")
+        bpy.ops.mesh.primitive_cube_add(size=10, location=(0, 0, 4))
+        obj = bpy.data.objects["Cube"]
+        scene.collection.objects.link(obj)
+        tool.Ifc.link(product, obj)
+        scene.cursor.location = (0, 0, 0)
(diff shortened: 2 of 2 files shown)


Moult:
I don't think this is the correct fix. The intent of the removed line is to handle space regeneration when the building storey or selected object's Z changes.

CyrilWaechter:
> I don't think this is the correct fix. The intent of the removed line is to handle space regeneration when the building storey or selected object's Z changes.

Really ? Sorry I was pretty confident about it. It worked in all test cases (simple test case from scratch and real world project).

From your sentence the test case would be as followed ?

Test 1:
1. Change the related building z.
2. Regen
3. Expected result ?

Test 2:
1. Change space z.
2. Regen
3. Expected result: Space stays in place ?

Moult:
Sorry from re-reading the diffs (I don't have Bonsai running now) I think you are right.

If I understand correctly L223 is the branch if you are regenerating an existing space. So `z` is the bottom of the space (`get_x_y_z_h_mat_from_obj`) and after your fix, `z` (and `h`) won't change, only the 2D profile will (good!)

So Test 1 result is: space does _not_ snap back. If the user wants it to snap back, the user should manually move the space, or delete the space and regen from nothing. I think this is good?

And Test 2 result is: space stays in place, which I think is also good because the user intentionally moved it.

Edit: I had an old memory of it always snapping back to the default Z but after reading https://github.com/IfcOpenShell/IfcOpenShell/commit/d8de623086571ffcfcb1465f66b136fbffdba1b8?diff=split I think I was mistaken.

Issue #7991, opened 2026-04-22
Bonsai Mirror Function Just Copies Element


PahlawanJoey:
### Bug Description

When using Slab Tool on a Slab trying to Mirror it, the element is just copied next to it instead of truly mirrored.

You can clearly see the shapes in the slabs that are not mirrored, also there's a steel beam from which the geometry dissapears (steel beams next to the red arrow)

<img width="1128" height="1007" alt="Image" src="https://github.com/user-attachments/assets/b393f78e-ca5d-4f55-a107-65605d57c5a9" />

[VWS_Type X_Vloeren 1e Verdieping.ifc.txt](https://github.com/user-attachments/files/26981820/VWS_Type.X_Vloeren.1e.Verdieping.ifc.txt)

Using:

```
os: Windows
os_version: 10.0.26200
python_version: 3.13.9
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: Intel64 Family 6 Model 198 Stepping 2, GenuineIntel
blender_version: 5.1.1
bonsai_version: 0.8.6-alpha260421
bonsai_commit_hash: e4f5c63
bonsai_commit_date: 2026-04-21T23:44:14+01:00
bonsai_git_branch: v0.8.0
last_actions: 
# ifcopenshell.api: geometry.unassign_representation
>>> {"product": {"cast_type": "entity_instance", "value": 425398, "Name": "Beam"}, "representation": {"cast_type": "entity_instance", "value": 425443, "Name": null}}
# ifcopenshell.api: geometry.remove_representation
>>> {"representation": {"cast_type": "entity_instance", "value": 425443, "Name": null}}
# ifcopenshell.api: pset.remove_pset
>>> {"product": {"cast_type": "entity_instance", "value": 425398, "Name": "Beam"}, "pset": {"cast_type": "entity_instance", "value": 425405, "Name": "Pset_VBICommon"}}
# ifcopenshell.api: type.unassign_type
>>> {"related_objects": [{"cast_type": "entity_instance", "value": 425398, "Name": "Beam"}]}
# ifcopenshell.api: owner.update_owner_history
>>> {"element": {"cast_type": "entity_instance", "value": 318956, "Name": null}}
# ifcopenshell.api: pset.remove_pset
>>> {"product": {"cast_type": "entity_instance", "value": 425398, "Name": "Beam"}, "pset": {"cast_type": "entity_instance", "value": 425423, "Name": "Pset_BeamCommon"}}
[...]

Issue #6906, opened 2025-07-14
Blender - Bonsai: Mirroring Wall


Juhkuphii:
### Bug Description

Mirroring single wall cause error:

### Attachments

[Bonsai_errorfile.txt](https://github.com/user-attachments/files/21221283/Bonsai_errorfile.txt)

### Debug and Error Output

```true
os: Windows
os_version: 10.0.26100
python_version: 3.11.11
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: Intel64 Family 6 Model 165 Stepping 2, GenuineIntel
blender_version: 4.4.3
bonsai_version: 0.8.2
bonsai_commit_hash: None
bonsai_commit_date: None
last_actions: 
# operator: BIM_OT_hotkey
# operator: BIM_OT_flip_wall
# ifcopenshell.api: geometry.edit_object_placement
>>> {"product": {"cast_type": "entity_instance", "value": 1316, "Name": "Wall"}, "matrix": {"cast_type": "ndarray", "value": [[1.0, 0.0, 0.0, 7914.00139280453], [0.0, 1.0, 0.0, 16191.002320289574], [0.0, 0.0, 1.0, 0.0], [0.0, 0.0, 0.0, 1.0]]}, "is_si": "False", "should_transform_children": "False"}
# ifcopenshell.api: owner.update_owner_history
>>> {"element": {"cast_type": "entity_instance", "value": 1316, "Name": "Wall"}}
# ifcopenshell.api: geometry.regenerate_wall_representation
>>> {}
# ifcopenshell.api: geometry.edit_object_placement
>>> {"product": {"cast_type": "entity_instance", "value": 1316, "Name": "Wall"}, "matrix": {"cast_type": "ndarray", "value": [[1.0, 0.0, 0.0, 7914.001436224239], [0.0, 1.0, 0.0, 16191.002320289574], [0.0, 0.0, 1.0, 0.0], [0.0, 0.0, 0.0, 1.0]]}, "is_si": "False", "should_transform_children": "True"}
# ifcopenshell.api: owner.update_owner_history
>>> {"element": {"cast_type": "entity_instance", "value": 1316, "Name": "Wall"}}
# operator: BIM_OT_hotkey
# operator: BIM_OT_flip_wall
# ifcopenshell.api: geometry.edit_object_placement
[...]

PR #9000, opened 2026-07-26
Bonsai: make Mirror a true mirror (#7991, #6906)

BIMvoice:
`bim.mirror_elements` never mirrored anything. Its own comment said so:

> "This is not a true mirror operation... we calculate bounding box centroids, and mirror the position of the object... but never actually truly mirror anything."

It duplicated the selection and repositioned the copies, leaving geometry untouched. On a symmetric element that is indistinguishable from a mirror. On an asymmetric one you get a rotated copy, which is what #7991 and #6906 report.

## What changes

**Mirror inverts the geometry and transforms in place.** It no longer duplicates: duplicate first if you want to keep the original, the same as any other transform.

**Selecting one object** mirrors it about one of its own local planes. **Selecting two or more** makes the active object the mirror plane, through its bounding box centre rather than its origin, since an IFC origin can sit anywhere including inside the object being mirrored. **Mirror Axis** chooses which local axis is reflected along, of the object or of the reference.

## Safety

**Typed occurrences never invert the type's geometry**, which is shared with every sibling. The type is duplicated once, the duplicate inverted, and later mirrors of the same type reuse it. A guard raises rather than descending into geometry owned by a type.

**An element that cannot be inverted along the chosen axis is left completely untouched** and named in the report, rather than having its placement reflected around geometry that never moved.

## Two defects found while verifying

**The mirrored placement was not written back**, so the model looked correct on screen and reloaded with elements metres out of position. It now goes through `geometry.edit_object_placement`.

**Re-running from Adjust Last Operation composed with the previous result.** Blender's redo path does not roll back the IFC, so mirroring X then Y negated both axes, which is a rotation, not a reflection. Mirroring is its own inverse, so a redo replays the previous run to cancel it before applying the requested axis.

## Verification

Against IFC2X3 models supplied by two reporters:

- 48/48 walls exact across a reference, openings included
- all nine axis pairs equal to a single clean run of the second axis
- 11/11 correct after saving and reloading in a fresh Blender session, worst deviation 3 mm
- typed occurrence mirrored with its siblings and the original type byte-identical afterwards
- no duplicate GlobalIds, no empty or orphaned representations

`bim.flip_object` is untouched.

## Dependency

The Mirror **button** only reaches this operator with #8426. Without it the shared `Shift+M` hotkey dispatches to Merge for LAYER2 elements, so Mirror is unreachable on walls, which is the other half of #6906.

The `Shift+M` / `Ctrl+M` clash itself is deliberately **not** addressed here. It is a keymap decision worth its own discussion on #6906, particularly as Blender binds `transform.mirror` to `Ctrl+M`.

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/product.py
+++ b/src/bonsai/bonsai/bim/module/model/product.py
@@ -708,13 +708,16 @@ class MirrorElements(bpy.types.Operator, tool.Ifc.Operator):
         if mirrored:
             # Say which plane was used. A reflection across a reference that straddles the
             # object's own mirror plane moves nothing, and without this the user cannot tell
-            # that the reference was taken into account at all.
+            # that the reference was taken into account at all. The status bar flash is gone
+            # by the time anyone asks what happened, so this also goes to the console.
             about = (
                 f"across the middle of {mirror_ref.name}, along its local {self.mirror_axis}"
                 if mirror_ref
-                else f"about own local {self.mirror_axis}"
+                else f"about own local {self.mirror_axis} (no reference object)"
             )
-            self.report({"INFO"}, f"Mirrored {mirrored} object{'s' if mirrored > 1 else ''} {about}")
+            summary = f"Mirrored {mirrored} object{'s' if mirrored > 1 else ''} {about}"
+            self.report({"INFO"}, summary)
+            print(f"[bim.mirror_elements] {summary}")
         return {"FINISHED"} if mirrored else {"CANCELLED"}
 
     def mirror_obj(
--- a/src/bonsai/bonsai/bim/module/model/product.py
+++ b/src/bonsai/bonsai/bim/module/model/product.py
@@ -633,16 +633,17 @@ class MirrorElements(bpy.types.Operator, tool.Ifc.Operator):
     bl_description = (
         "Mirrors the selected objects in place by truly inverting their geometry. "
         "Select a single object to mirror it about one of its own local planes. "
-        "Select two or more objects and the active object becomes the mirror plane, "
-        "with every other selected object reflected across it. Nothing is duplicated: "
-        "duplicate first if you want to keep the original"
+        "Select two or more and the active object becomes the mirror plane, passing through "
+        "its middle. Mirror Axis picks which local axis of the object, or of the reference, "
+        "is mirrored along. Nothing is duplicated: duplicate first to keep the original"
     )
 
     mirror_axis: bpy.props.EnumProperty(
         name="Mirror Axis",
         description=(
-            "Local axis to mirror along. With a single object this is the object's own axis, "
-            "otherwise it is the axis of the active reference object"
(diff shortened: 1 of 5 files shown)


theoryshaw:
I think there is some duplicate functionality with.. https://github.com/IfcOpenShell/IfcOpenShell/pull/7003

BIMvoice:
For reviewers: @Rob2309's #7003, open since August 2025, also implements true mirroring and predates this PR. It adds a separate `bim.mirror_geometry` operator where this one modifies the existing `bim.mirror_elements`. **The two conflict.** We should have built on his work and did not, and this PR body originally failed to mention it at all.

If a maintainer prefers his approach, this PR should be closed in favour of it.

theoryshaw:
@BIMvoice Maybe as a standard approach before pushing your prs, ask your bot to see if theres any existing prs that conflict or do the same thing.

theoryshaw:
@BIMvoice Close this issue, so it doesn't conflict on BonsaiPR?

BIMvoice:
Closing, as @theoryshaw suggests, so it stops conflicting with @Rob2309's #7003 in BonsaiPR bundles. His came first and we should have built on it rather than opening this.

Four things were developed on top of this branch that #7003 does not currently have, and they are independent of which implementation wins:

- Ctrl+M never reached Bonsai's mirror at all. Blender's default keymap binds it to the native `transform.mirror`, which negatively scales the Blender object and never touches the IFC, so objects move without their geometry flipping and no error appears. #7003 has no keymap binding either, so it has the same gap.
- `ShapeBuilder.mirror` left a profile in a corrupted, partially shifted state when it met a curve type it does not support, `IfcCompositeCurve`, which is what filleted L profile beams use. It now leaves such a profile byte identical instead.
- The mirror axis is derived from the reference object automatically rather than defaulting to local X, which is wrong whenever the reference's placement is rotated.
- Shift+M forced a merge attempt and errored out whenever the active object was a wall or railing, even when the user pressed Mirror rather than Merge.

Plus 30 tests, where #7003 currently has none.

They are on `fix/mirror-duplicated-shared-geometry` in our fork and we are happy to offer them to #7003 as commits.

BIMvoice:
@theoryshaw yes, and that is exactly the gap. It is a standard step now.

What we had was a check for an existing PR referencing the same issue number. That is why we missed #7003 and #7648: searching for an issue number does not surface a PR that never referenced it, and #7648's title says "closes #3565" while ours said "let Shift+E extend non-typed profile extrusions". Same feature, no shared string.

The check now also searches by feature and by the files being touched, regardless of author, and runs `git merge-tree` against anything that overlaps, with the result stated in the PR body.

We also ran that check backwards across all 662 open PRs today. It found twelve more of ours that plausibly duplicate other people's work, the oldest overlapping a PR of yours from March, and two where ours came first and someone else's overlaps it. We will work through them.

Issue #8008, opened 2026-04-25
Snap performance — why not use context.scene.ray_cast()?


tagehedin:
### Feature Description

Snap performance — why not use context.scene.ray_cast()?
                                                                                                                                 
I have been looking with Calude Code at the performance of the polyline snap system in the BIM authoring tools (wall, slab, door placement etc.)  and ran some experiments replacing the custom BVH pipeline with Blender's built-in context.scene.ray_cast().

The current implementation builds Python BVH trees per-object in a background thread, maintains a bbox rebuild system to filter which objects to raycast against, and uses a timer to apply results asynchronously. It works, but on large models the bbox rebuild and BVH construction introduce noticeable lag.                                                       
                                                            
We replaced this entire pipeline with a single call: 
hit, location, normal, face_index, obj, matrix = context.scene.ray_cast(depsgraph, origin, direction)

This gives us the face directly under the cursor. We then take the face's 3–4 vertices, project them to screen space, and find the closest vertex, midpoint, or edge point within pixel thresholds — no BVH building, no bbox filtering, no background thread needed.

The result is noticeably faster snapping and the code went from ~200 lines to ~50.                                             

The question: Was there a specific reason the snap system was built around a custom Python BVH rather than scene.ray_cast()? Is there something scene.ray_cast() can't do that the BVH approach handles — for example x-ray mode, hidden geometry, or snapping to objects not visible to the camera?

BIMvoice:
Related datapoint from #8187: profiling there showed the same snap path's two practical costs — a crash on fully off-screen edges (fixed in PR #8261) and `filter_objects_to_raycast` evaluating heavy *linked* models (~30s stalls; a skip-linked-references toggle was proposed in-thread with tested code). Whatever the verdict on `scene.ray_cast()` vs the custom BVH, those two changes address most of the observed pain and may inform this design question.

brunoperdigao:
@tagehedin The reason a custom BVH was created was to deal with objects that have no face and have a lot of edges. The main case are objects that represent a 2d drawing. `scene.ray_cast()` doesn't handle objects with no faces. About performance, did you test in a complex scene with hundreds of objects?
There is a lot of room for improvement in the current snap, but there are a lot of small use cases to take into account. I'll be analyzing a few PR's the next days.

brunoperdigao:
I'm closing this since the snap code has changed a lot after https://github.com/IfcOpenShell/IfcOpenShell/pull/9337

PR #9004, opened 2026-07-26
Bonsai: snap to the nearest solid, not the nearest object origin

### LINKED ISSUE #9003: Bonsai: object snapping skips a covering/solid coincident with a wall or slab
### Bug Description

Object snapping (polyline / wall / slab placement, measure) cannot snap to an **IfcCovering** or an **IfcBuildingElementPart** when its geometry is coincident with, or sits directly on, another solid such as a wall or slab. The vertices/edges of the covering are simply never offered as snap points where it overlaps the other element. Snapping to the same covering works fine on a part that overhangs into empty space.

### Steps to Reproduce

1. Open a model with an `IfcCovering` (e.g. a floor/ceiling/wall finish) placed on an `IfcSlab`/`IfcWall`, or an `IfcBuildingElementPart` with a tessellated body coincident with another solid.
2. Start the polyline/wall tool (or the measure tool).
3. Hover the cursor over the covering **where it overlaps** the slab/wall.

**Expected:** snap to the covering's vertices/edges.
**Actual:** no snap on the covering; only the underlying element (or nothing) is picked. Moving over a portion of the covering that does **not** overlap any other solid snaps correctly.

### Root Cause

In `Raycast.ray_cast_and_get_closest_to_camera_snaps` (`src/bonsai/bonsai/tool/raycast.py`), the non-xray path selects the snap target among the candidate solids by:

1. sorting them by **object-origin distance** to the ray (`so.obj.matrix_world.translation - ray_origin`), and
2. **stopping at the first face hit** (`break`).

Only that object is flagged `is_closest_to_camera`, and `detect_snapping_points` then generates vertex/edge snaps for that object alone. The object origin is a poor proxy for what is actually under the cursor (it can sit far from the geometry), so when two solids overlap, whichever origin sorts first wins — and a covering coincident with a wall/slab is skipped even though it is the front-most surface. Separated elements (e.g. plain walls) snap fine because the heuristic only misbehaves when solids overlap.

### Suggested Fix

Ray cast every candidate solid under the cursor and keep the one with the **nearest actual hit distance** (which is what the xray path and `Raycast.cast_rays_and_get_best_object` already do), instead of the origin-sort + first-hit heuristic. The extra ray casts are limited to the few solids whose 2D bounding box is under the cursor, and `obj.ray_cast` is cheap.

PR: (link on submission)

### Related
- #8008 — broader discussion on the snapping performance / raycast approach.

### Environment

- Bonsai `v0.8.0` (current)
- Blender 5.1, Python 3.13, Windows

carlopav:
## Problem

Object snapping cannot snap to an `IfcCovering` (or any solid) that is coincident with another solid such as a wall or slab. Its vertices/edges are never offered as snap points where it overlaps the other element; the same covering snaps fine where it overhangs into empty space. `IfcBuildingElementPart` with a tessellated body has the same symptom.

In `Raycast.ray_cast_and_get_closest_to_camera_snaps` (non-xray path), candidate solids are sorted by **object-origin distance** to the ray and the loop `break`s at the first face hit. Only that object is flagged `is_closest_to_camera`, and `detect_snapping_points` then generates vertex/edge snaps for it alone. The object origin is a poor proxy for what is under the cursor (it can sit far from the geometry), so when two solids overlap the wrong one is picked and the front-most surface — the covering — is skipped. Separated elements (plain walls) are unaffected because the heuristic only misbehaves when solids overlap.

## Fix

Ray cast every candidate solid under the cursor and keep the one with the **nearest actual hit distance** — which is what the xray branch and `Raycast.cast_rays_and_get_best_object` already do — instead of the origin-sort + first-hit heuristic. The extra ray casts are limited to the few solids whose 2D bounding box is under the cursor, and `obj.ray_cast` is cheap.

## Performance trade-off

This replaces the single early-exit ray cast with one ray cast per candidate solid under the cursor. `filter_objects_to_raycast` already narrows candidates to objects whose 2D bounding box overlaps the mouse, so in the common case that is the 2-3 solids meeting at a wall/slab/covering junction, and `obj.ray_cast` uses Blender's cached per-object BVH. For very heavy **linked** scenes, where 2D bounding boxes can be large/imprecise, this does remove the early-exit that previously bounded the non-xray path to a single ray cast — the broader cost of snapping against heavy linked geometry is already tracked in #8187 (Problem 2) and #8008, and would be better addressed by narrowing the candidate set there rather than by keeping the origin-sort heuristic here.

## Verify

Logic change in Blender-bound modal code. Tested on a real IFC4 project: a covering coincident with a slab, and a tessellated `IfcBuildingElementPart`, both went from "not snappable where they overlap" to snapping correctly; plain walls unchanged. Syntax-checked.

A modal regression test in `test/modal/test_modal.py` (which drives object snapping over `snap.ifc`) is the right place for a fixture reproducing this, and is planned as a follow-up once a minimal coincident-solid fixture that reliably reproduces the pre-fix behaviour is prepared. Happy to add it here first if preferred.

## Related

- Closes #9003

Test file in issue comments.

### DIFF
--- a/src/bonsai/bonsai/tool/raycast.py
+++ b/src/bonsai/bonsai/tool/raycast.py
@@ -864,8 +864,11 @@ class Raycast(bonsai.core.tool.Raycast):
 
         if not xray_mode and objs_to_raycast:
             # Non-xray - only the closest solid object's Face snap is kept by
-            # the caller (detect_snapping_points).  Process solids in distance
-            # order and stop at the first hit to minimise raycasts.
+            # the caller (detect_snapping_points).  Ray cast every solid under
+            # the cursor and keep the one with the nearest actual hit. Sorting by
+            # object origin and stopping at the first hit was cheaper but picked
+            # the wrong object when solids overlap, so a covering coincident with
+            # a wall/slab (whose origin may sort closer) never got snapped.
             wireframe_objs = []
             solid_objs = []
             for snap_obj in objs_to_raycast:
@@ -875,9 +878,6 @@ class Raycast(bonsai.core.tool.Raycast):
                     wireframe_objs.append(snap_obj)
                 else:
                     solid_objs.append(snap_obj)
-
-            # Rough distance - object origin to ray origin
-            solid_objs.sort(key=lambda so: (so.obj.matrix_world.translation - ray_origin).length_squared)
 
             # Process wireframe objects first (all of them, always collected)
             for snap_obj in wireframe_objs:
@@ -890,7 +890,7 @@ class Raycast(bonsai.core.tool.Raycast):
                         closest_hit = hit
                         closest_face_index = None
 
-            # Process solid objects in distance order, stop at first hit
+            # Ray cast every solid and keep the nearest actual hit
             for snap_obj in solid_objs:
                 hit_obj, hit, face_index = cls.cast_rays_to_single_object(context, event, snap_obj.obj)
 
@@ -911,8 +911,6 @@ class Raycast(bonsai.core.tool.Raycast):
                         closest_obj = hit_obj
                         closest_hit = hit
                         closest_face_index = face_index
-
-                    break
 
         else:
             # Xray mode - process all objects (all snaps are kept by the caller)


carlopav:
ping @brunoperdigao

brunoperdigao:
In fact, the origin is not the best parameter. However, this doesn't solve the issue for when we have a wireframe object in front of a mesh, because raycast doesn't detect objects without faces. We need to prioritize the wireframe in those cases. I'll have to think about an alternative approach.

brunoperdigao:
I'm trying a general different approach for the snapping system, that uses the GPU Module from the Blender API. If it works it might solve a lot of current issues.

carlopav:
> I'm trying a general different approach for the snapping system, that uses the GPU Module from the Blender API. If it works it might solve a lot of current issues.

Thanks for the feedback, I saw that mentioned in another issue, but I thought to open this pr anyway to have a little improvement in the meantime, expecially because it was just two lines of code.
Let me know if you prefer to close the PR, no problem from my side, and good luck with the different approach!

carlopav:
fixed with recent changes by @brunoperdigao

PR #9010, opened 2026-07-27
Bonsai: rotate the door hover preview with the wall on IFC2X3

### LINKED ISSUE #9346: Bonsai: window preview is offset by the host wall's translation
## Bug Description

The blue `IfcWindow` occurrence preview is displaced from a translated host wall even though the insertion point is snapped to that wall.

Expected: the preview is drawn on the wall at the snapped insertion point.

Actual: when the host wall's object origin is away from the world origin, the preview is displaced by approximately the wall's world translation. In the observed case, the wall origin and snap X coordinate are `10 m`, and the preview is drawn roughly another `10 m` away. Walls whose origins are close to the world origin appear correct.

Steps to reproduce:

1. Create an IFC project and a wall whose object origin is away from the world origin, for example at X = `10 m`.
2. Create or select an `IfcWindowType`.
3. Activate the Window Tool.
4. Hover over the translated wall so the insertion point snaps to it.
5. Observe that the blue window preview is drawn away from the host wall.

The likely cause is in [`ProductDecorator.get_generic_preview_data()`](https://github.com/IfcOpenShell/IfcOpenShell/blob/2b429dc605303154af3680a78241c0d617271fca/src/bonsai/bonsai/bim/module/model/decorator.py#L1532-L1600). The base-axis branch assigns the full wall world matrix:

```python
rot_mat = snap_obj.matrix_world
```

The preview then applies both the snapped mouse translation and that matrix:

```python
combined_m = translate_mouse @ rot_mat @ translate_rl @ self.obj_matrix_i
```

Because `snap_obj.matrix_world` includes the wall translation, the translation is applied twice. The side-axis branch already builds a rotation-only matrix with `to_quaternion().to_matrix().to_4x4()`. Using a rotation-only wall matrix in the base-axis branch should keep the preview at the snapped point.

Related to #5551, where the occurrence preview was originally implemented.

## Attachments

Not attached because the original screenshot includes a private floor plan. This should be reproducible with a fresh wall placed away from the world origin.

## Debug and Error Output

```yes
os: Darwin
os_version: macOS Tahoe 26.5.2 (Darwin 25.5.0)
python_version: 3.13.13
architecture: arm64
machine: arm64
blender_version: 5.2.0 LTS
bonsai_version: 0.8.6-alpha260819
bonsai_commit_hash: 2b429dc
bonsai_commit_date: 2026-08-19T20:07:52-03:00
bonsai_git_branch: v0.8.0
last_error: None
```

BIMvoice:
IFC2X3 spells door and window types `IfcDoorStyle` and `IfcWindowStyle`. `ProductDecorator.get_generic_preview_data` matched only the IFC4 spellings:

```python
if relating_type.is_a() in ["IfcDoorType", "IfcWindowType"] and snap_element and snap_element.is_a("IfcWall"):
```

so on IFC2X3 the wall alignment block never ran, `rot_mat` stayed the identity matrix, and the hover preview was drawn in the door's rest orientation. The visible error equalled the wall's own angle, which is why it read as 90 degrees on a wall running along Y. The same miss skipped the sill offset and the wall Z snap.

Two further defects surfaced while measuring the fix at non axis aligned angles, and both are fixed here:

1. The base axis branch used the wall's full 4x4 matrix as `rot_mat`, so its translation landed on top of the mouse translation and the ghost was displaced by the wall's distance from the origin. This affected **IFC4 as well**, and was invisible when drawing near the origin. Both branches now build a rotation only matrix.
2. The wall Z snap was immediately overwritten by the container elevation override. The override now runs first as the generic fallback so the wall branch wins.

## Measured

Preview rotation error against the wall, real decorator data path, walls away from the origin:

| wall angle | IFC2X3 before | IFC2X3 after | IFC4 before | IFC4 after |
|---|---|---|---|---|
| 0 | 0.23 | 0.23 | 0.23 | 0.23 |
| 30 | 29.77 | 0.23 | 0.23 | 0.23 |
| 45 | 44.77 | 0.23 | 0.23 | 0.23 |
| 90 | 89.77 | 0.23 | 0.23 | 0.23 |
| 135 | 45.23 | 0.23 | 0.23 | 0.23 |
| 225 | 44.77 | 0.23 | 0.23 | 0.23 |

The 0.23 residual is PCA noise from the asymmetric frame profile, identical in both schemas.

Ghost to placed door centroid distance after the fix: **0.000 m at every angle, on both wall faces, in both schemas**, across 96 placements including walls flipped with `bim.flip_wall` so the direction sense path is covered.

## Testing

New parametrized regression test over wall angles 0, 30, 90 and 135 in both schemas, asserting rotation, ghost to door distance, placement alignment and that the wall stays the active object. Red on unfixed code, all 8 cases failing. Green with the fix. 16 neighbouring preview and decorator tests still pass.

Verified by hand in Blender on IFC2X3 and IFC4, on freshly drawn walls at several angles and on an imported IFC2X3 model.

---

**2026-08-31 update.** Fixes #9346 (#9347 is the same report). The double-applied host translation those issues describe is exactly the `rot_mat = snap_obj.matrix_world` composition this PR replaces with rotation-only quaternions; the regression test's origin-offset cases cover the reported 10 m displacement. Credit where due: @jaliste's earlier #7659 fixes the same line by zeroing `rot_mat.translation` and additionally recalculates `obj_matrix_i` per frame; if maintainers prefer that PR, this one folds its test in on top gladly.

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/decorator.py
+++ b/src/bonsai/bonsai/bim/module/model/decorator.py
@@ -1547,6 +1547,9 @@ class ProductDecorator(tool.Blender.ViewportDecorator):
         mouse_point = Vector((snap_prop.x, snap_prop.y, default_container_elevation))
         snap_obj = bpy.data.objects.get(snap_prop.snap_object)
         snap_element = tool.Ifc.get_entity(snap_obj)
+        if snap_element and (container := ifcopenshell.util.element.get_container(snap_element)):
+            container_obj = tool.Ifc.get_object(container)
+            mouse_point.z = container_obj.location.z
         rot_mat = Matrix()
         if (
             relating_class in ("IfcDoorType", "IfcDoorStyle", "IfcWindowType", "IfcWindowStyle")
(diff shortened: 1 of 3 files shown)


BIMvoice:

PR #9011, opened 2026-07-27
Bonsai: recut a voided host from its active representation, not its context

BIMvoice:
Placing a door into an imported wall could make the **host wall disappear**.

`FilledOpeningGenerator.generate` resolved the representation to recut like this:

```python
context = tool.Geometry.get_active_representation_context(voided_obj)
representation = tool.Geometry.get_representation_by_context(voided_element, context)
```

`get_active_representation_context` discards *which* representation was active and keeps only its `ContextOfItems`. `get_representation_by_context` then asks for any representation on that context, and `ifcopenshell.util.representation.get_representation` returns the **first** match.

In a file with no subcontexts, `Axis`, `BoundingBox` and `Body` all sit on the same context, and `Axis` comes first. The wall was re-tessellated from its Axis polyline, which produces no faces, so the object survived in the outliner with a zero vertex mesh and vanished from the viewport.

The wall's IFC data was never damaged. Only the Blender mesh was rebuilt from the wrong representation.

The fix adds `Geometry.get_host_representation_to_recut`, which returns the host's active representation directly and only falls back to searching by context when the host has no active `IfcShapeRepresentation`. Three other `recut_host` call sites already read the active representation directly; the two that round tripped through the context are corrected, along with the identical pattern in the array child opening mirror.

## Measured

Imported IFC2X3 wall, all three representations on one context:

| | mesh source | verts | polys |
|---|---|---|---|
| before placing a door | Body | 8 | 12 |
| after, unfixed | **Axis** | **0** | **0** |
| after, fixed | Body | 8 | 12 |

Non regression on a fresh IFC4 project with proper subcontexts: the wall goes from 8 verts / 12 polys to 12 verts / 20 polys with the opening cut, resolving to Body both before and after the change.

## Testing

4 new unit tests covering the active representation case, the no active representation fallback, the non `IfcShapeRepresentation` case and the missing element case. 115 existing `test/tool` tests pass.

Verified by hand in Blender on an imported IFC2X3 model where the wall previously disappeared on every door placement.

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/opening.py
+++ b/src/bonsai/bonsai/bim/module/model/opening.py
@@ -414,8 +414,7 @@ class FilledOpeningGenerator:
             if voided_obj.data:
                 voided_element = tool.Ifc.get_entity(voided_obj)
                 assert voided_element
-                context = tool.Geometry.get_active_representation_context(voided_obj)
-                representation = tool.Geometry.get_representation_by_context(voided_element, context)
+                representation = tool.Geometry.get_host_representation_to_recut(voided_obj)
                 assert representation
 
                 tool.Geometry.recut_host(voided_obj, representation)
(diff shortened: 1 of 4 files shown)


BIMvoice:

PR #9013, opened 2026-07-27
Bonsai: cut the opening when the host and the void sit on different contexts

BIMvoice:
A door placed into an imported wall produced no hole. The IFC authoring is correct and complete: the `IfcOpeningElement` is created, `IfcRelVoidsElement` links it to the wall and `IfcRelFillsElement` links the door to the opening. The geometry pipeline silently declines to subtract it.

## Root cause

`src/ifcgeom/Converter.cpp` resolves each boolean operand through the output representation selector:

```cpp
auto repr = mapping()->representation_of(opening->as<IfcUtil::IfcBaseEntity>());
if (repr) { return std::make_pair(mapping()->map(repr), ...); }
else      { return std::make_pair(taxonomy::ptr{}, taxonomy::matrix4{}); }
```

`mapping::representation_of` builds its candidate set from the `context-ids` allow list and returns nullptr when the intersection with the opening's own representations is empty. Converter drops that operand, and once all operands are dropped it emits the host **unsubtracted, with no warning**:

```cpp
if (opening_items.empty()) { opened_shapes = shapes; }
```

Bonsai narrows `context-ids` to exactly the host representation's context, in `reimport_element_representations` and in `create_settings`. Bonsai authors openings on `Model/Body/MODEL_VIEW`, creating that subcontext on demand. In a file with no subcontexts the host stays on the original context while its new opening lands on the freshly created one. Different contexts, so the subtraction is skipped.

Isolated, same representation in every run:

```
default, no context filter   verts=18 faces=28
context-ids=[host]           verts=8  faces=12   <- same as disable-opening-subtractions
context-ids=[host, opening]  verts=18 faces=28
disable-opening-subtractions verts=8  faces=12
```

## The fix

`ifcopenshell.util.representation.get_missing_opening_context_ids` returns the extra context ids needed by an element's openings, and **only** for openings that have no representation already inside the allow list. Both Bonsai call sites extend their list with it. A model that keeps hosts and openings on one context gets exactly the filter it had before.

Widening unconditionally was tried first and rejected by measurement: it cost 60 percent on a model whose openings carry representations in two contexts, because every pass dragged in the other whole context.

## Not fixed here

The deeper defect is in C++: `mapping::representation_of` filters boolean operands by the output context list for every `context-ids` consumer, including `IfcConvert --context-ids`, and drops them silently rather than warning. That is left untouched because it could not be built and tested locally here, and shipping an unverified kernel change would be worse. Flagging it for anyone better placed to judge the right behaviour at that layer.

## Measured

Imported IFC2X3 wall, place a door: **8 verts / 12 polys to 18 verts / 28 polys**, and the hole survives save and reopen.

Non regression, all identical with and without the change:

- fresh IFC4 project, 14 contexts: 8/12 to 12/20 both ways
- `AC20-FZK-Haus.ifc`, openings across 2 contexts: 16242 verts, 30600 polys, all 9 voided hosts identical
- `project0-openings.ifc`, 13 contexts: identical, all 3 voided hosts
- `AC20-Institute-Var-2.ifc`, 287 openings: 26342 verts identical, 2.31s against a 2.26s baseline

## Testing

7 new unit tests. `test/util` goes from 369 to 376 passing with the same pre existing environment failures. Bonsai `test/core` 390 pass. In Blender geometry and loader tests, 94 pass.

Verified by hand in Blender on the reporter model where the door previously produced no opening at all.

### DIFF
--- a/src/bonsai/bonsai/tool/geometry.py
+++ b/src/bonsai/bonsai/tool/geometry.py
@@ -1158,9 +1158,14 @@ class Geometry(bonsai.core.tool.Geometry):
         ifc_importer = bonsai.bim.import_ifc.IfcImporter(ifc_import_settings)
         ifc_importer.file = tool.Ifc.get()
 
-        settings.set("context-ids", [context.id()])
-        if not apply_openings:
+        context_ids = {context.id()}
+        if apply_openings:
+            context_ids |= ifcopenshell.util.representation.get_missing_opening_context_ids(
+                ifc_file, context_ids, elements
+            )
+        else:
             settings.set("disable-opening-subtractions", True)
+        settings.set("context-ids", sorted(context_ids))
 
         shape = None
         if elements:
--- a/src/bonsai/bonsai/tool/loader.py
+++ b/src/bonsai/bonsai/tool/loader.py
@@ -716,12 +716,18 @@ class Loader(bonsai.core.tool.Loader):
     @classmethod
     def create_settings(cls, is_gross=False):
         results = []
-        for context in cls.settings.contexts:
+        contexts = cls.settings.contexts
+        ifc_file = contexts[0].file if contexts else None
+        for context in contexts:
             settings = ifcopenshell.geom.settings()
             settings.set("mesher-linear-deflection", cls.settings.deflection_tolerance)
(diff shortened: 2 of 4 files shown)


BIMvoice:

PR #9036, opened 2026-07-29
ifcmcp: pin mcp below 2.0 to fix broken FastMCP import

BIMvoice:
`mcp` 2.0.0 removed the `mcp.server.fastmcp` module, which `ifcmcp/server.py` imports. Every CI run since that release fails `ifcmcp`'s own test suite, so **every open PR in this repository currently shows a red `compile-and-test`** regardless of what it changes.

Verified in a clean venv:

- **mcp 2.0.0**: `from mcp.server.fastmcp import FastMCP` raises `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`. No `FastMCP` symbol exists anywhere else in the package; the replacement is `mcp.server.mcpserver.MCPServer`, confirmed present.
- **mcp 1.29.0**, the highest release below 2.0: both `FastMCP` and `ImageContent` import cleanly.

Both declarations were unpinned. The workflow pin unbreaks CI. The `pyproject.toml` pin matters beyond CI, since `pip install ifcopenshell-mcp[mcp]` currently installs a version the code cannot use.

This also narrows the import guard from `except Exception` to `except ImportError`. The bare catch reported an upstream API removal as "FastMCP is not installed", which sent the diagnosis in the wrong direction.

Porting `ifcmcp` to the mcp 2.x API is a separate change and is not attempted here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -263,7 +263,9 @@ jobs:
           cd ../ifcquery && make test || ERROR=1
           pip install -e ../ifcedit --no-deps
           cd ../ifcedit && make test || ERROR=1
-          pip install mcp
+          # Pinned <2: mcp 2.0.0 renamed mcp.server.fastmcp.FastMCP to
+          # mcp.server.mcpserver.MCPServer, which ifcmcp doesn't support yet.
+          pip install "mcp>=1.0,<2"
           pip install -e ../ifcmcp --no-deps
           cd ../ifcmcp && make test || ERROR=1
           pip install -e ../ifctester --no-deps
--- a/src/ifcmcp/ifcmcp/server.py
+++ b/src/ifcmcp/ifcmcp/server.py
@@ -9,7 +9,7 @@ from ifcmcp.core import IfcSession
 try:
     from mcp.server.fastmcp import FastMCP  # type: ignore
     from mcp.types import ImageContent  # type: ignore
-except Exception:  # pragma: no cover
+except ImportError:  # pragma: no cover
     FastMCP = None  # type: ignore
     ImageContent = None  # type: ignore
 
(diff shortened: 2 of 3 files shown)


BIMvoice:
Some measurements that may help judge this one quickly, since `compile-and-test` is currently red across the entire open PR queue.

**Scope.** This is not specific to any single PR. `compile-and-test` runs the `src/ifcmcp` suite, so the failure reproduces on every open PR regardless of what it changes, including PRs that touch no Python at all. All 718 currently open PRs show a red `compile-and-test` for this reason.

**Evidence.** Taking one run arbitrarily (`30509340832`, on a PR that only touches ifcpatch recipes), the full result is:

```
2 failed, 63 passed, 5 errors
```

All 7 of those are the same import, all in `src/ifcmcp/tests/test_server.py`:

```
E   ImportError: FastMCP is not installed. Install with: pip install ifcmcp[mcp] (or add 'mcp' to your environment).
ifcmcp/server.py:20: ImportError
```

Every other suite in that same run passes, and `lint-formatting` is green.

**Cause.** `ci.yml` runs a bare `pip install mcp`. That run's install log resolved it to:

```
Successfully installed ... mcp-2.0.0 mcp-types-2.0.0
```

`mcp` 2.0.0 is the current release on PyPI, so every fresh run picks it up, and it no longer provides `mcp.server.fastmcp`. The `try/except` at `server.py:9` converts that into `FastMCP = None`, and the tests then raise.

**One thing worth a reviewer's attention:** besides the pin, this also narrows `except Exception` to `except ImportError` at `server.py:12`. That is deliberate, so a genuine failure inside the import surfaces instead of being silently turned into `FastMCP = None`, but it does mean non-ImportError exceptions now propagate where previously they did not.

To be clear this is not a request to prioritise our PRs over anything else. The point is only that this is a three line dependency pin, and merging it turns `compile-and-test` green for the whole queue including other contributors' work, and stops CI failure mail for every author in the repo.

PR #9037, opened 2026-07-29
validate: fix --json crashing on messages containing a percent sign

### LINKED ISSUE #2481: validate.py invalid json
In some cases validate.py will output an invalid json 
```
[
    {
        "level": "error",
        "message": "With attribute:
    <attribute RelatedObjects: <set [1:?] of <entity IfcObject>>>
Value:
    ()
Not valid
",
        "instance": "#519317=IfcRelDefinesByProperties('3rCS_gFoz568TMiSmvkAlE',#21,$,$,(),#519316)"
    },
…
…
...
     {
        "level": "error",
        "message": "With attribute:
    <attribute RelatedObjects: <set [1:?] of <entity IfcObject>>>
Value:
    ()
Not valid
",
        "instance": "#100621=IfcRelDefinesByProperties('3lPiKDU$TA3PnHD1bhlSIy',#21,$,$,(),#100607)"
    }
]
```

BIMvoice:
Fixes #2481.

`json_logger.log` at `src/ifcopenshell-python/ifcopenshell/validate.py:99` applies `%` formatting unconditionally:

```python
self.statements.append({"level": level, "message": message % args, **self.state})
```

Several call sites pass a fully built message with no args, for example `logger.error(str(e))` at lines 570, 585 and 594. That message is `str(ValidationError)`, which embeds arbitrary content from the file being validated, including attribute values and entity names. When that content contains a literal `%` followed by a conversion character, `message % ()` raises `TypeError: not enough arguments for format string` and the whole validation run aborts, which is the traceback @cadgiru reported.

The fix mirrors `logging.LogRecord.getMessage()` and only substitutes when there are args to substitute.

Reproduced from scratch against unmodified `v0.8.0` with an `IfcAnnotation` named `Filled region 40%s green`, giving the same `TypeError` at the same line. Red on baseline with the new fixture, green after: 39 passed, including the 38 pre-existing `test_validate.py` cases. Real `%` substitution still works, confirmed with the duplicate-GlobalId message that does pass args.

Note on history: this issue was previously closed by this account with an incorrect analysis, and @Moult reopened it for a human to verify. The root cause above was re-derived from scratch rather than carried over from that comment.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/validate.py
+++ b/src/ifcopenshell-python/ifcopenshell/validate.py
@@ -97,7 +97,14 @@ class json_logger:
         self.state[key] = value
 
     def log(self, level, message, *args):
-        self.statements.append({"level": level, "message": message % args, **self.state})
+        # Mirror logging.LogRecord.getMessage(): only apply %-formatting when args
+        # are actually supplied. Several call sites pass a pre-built message (e.g.
+        # str(ValidationError), which can embed arbitrary IFC content such as a
+        # Name containing a literal "%s"). Unconditionally formatting with an
+        # empty args tuple would raise instead of logging the error.
+        if args:
+            message = message % args
+        self.statements.append({"level": level, "message": message, **self.state})
 
     def __getattr__(self, level):
         return functools.partial(self.log, level)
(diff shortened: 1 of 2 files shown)


aothms:
> and only substitutes when there are args to substitute.

But is this the full solution? When there are percentage signs in the file being validated **and** args to substitute, there can still be a mismatch in count. It seems like these days we should be able to come up with something better than percent sign formatting.

PR #9038, opened 2026-07-29
ifcpatch: preserve IFC2X3 georeferencing in ExtractElements

### LINKED ISSUE #8199: Extract elements recipe messes up georeferencing

BIMvoice:
Fixes #8199.

`ExtractElements` preserved only IFC4-style georeferencing:

```python
for ctx in element.RepresentationContexts or ():
    for coop in getattr(ctx, "HasCoordinateOperation", ()):
        self.new.add(coop)
```

IFC2X3 has no `IfcMapConversion`. It stores georeferencing as `ePSet_MapConversion` / `ePSet_ProjectedCRS` property sets on `IfcProject`, reachable only through the inverse `IsDefinedBy`. `self.new.add()` is a forward-attribute-only deep copy, so those psets were silently dropped. This is the same bug class as the original IFC4 fix, never extended to IFC2X3; the existing regression test explicitly skips IFC2X3, which is why it went unnoticed.

The consequence is worse than lost metadata. `ifcopenshell.api.project.append_asset` (`append_asset.py:417`) unconditionally globalises every placement using the source's georeferencing and re-localises using the target's. With the target missing georeferencing, survey-scale coordinates are baked into every extracted element:

```
wall authored at (5, 10, 2)  ->  extracted at (500003.19, 6000010.72, 102.0)
```

The fix copies the two named psets and relinks them via `IfcRelDefinesByProperties`, mirroring the existing `HasCoordinateOperation` copy.

`MergeProjects.py`, the other recipe touching georeferencing, already handles both psets correctly at line 108, so nothing else needed changing.

Red: `assert None is not None`, pset entirely missing. Green: pset preserved, `get_helmert_transformation_parameters` round-trips, wall placement back to `(5, 10, 2)`. No regressions in `test_ExtractElements.py` (12 passed, 2 skipped) or `test_append_asset.py` (77 passed).

Verified against a synthetic reproduction, not a production IFC2X3 file.

Note on history: this issue was previously closed by this account, and @Moult reopened it saying he wanted to check the georeferencing logic. The root cause above was re-derived from scratch.

### DIFF
--- a/src/ifcpatch/ifcpatch/recipes/ExtractElements.py
+++ b/src/ifcpatch/ifcpatch/recipes/ExtractElements.py
@@ -22,9 +22,23 @@ from typing import Union
 import ifcopenshell
 import ifcopenshell.api.project
 import ifcopenshell.guid
+import ifcopenshell.util.element
 import ifcopenshell.util.selector
 
 import ifcpatch
+
+# IFC2X3 has no IfcMapConversion / IfcProjectedCRS. Georeferencing is instead
+# stored as extension property sets, historically attached to IfcProject or
+# IfcSite depending on the authoring tool. See
+# https://github.com/buildingSMART/validate/issues/310#issuecomment-5076630963
+# for the complete set of names accepted across IFC2X3, IFC4 and IFC4X3_ADD2.
+IFC2X3_GEOREFERENCING_PSETS = (
+    "ePSet_GeographicCRS",
+    "ePSet_MapConversion",
+    "ePSet_MapConversionScaled",
+    "ePSet_ProjectedCRS",
+    "ePSet_RigidOperation",
+)
 
 
 class Patcher(ifcpatch.BasePatcher):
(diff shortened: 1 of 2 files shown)


src/ifcpatch/ifcpatch/recipes/ExtractElements.py
```diff
@@ -113,6 +113,31 @@ def append_asset(self, element: ifcopenshell.entity_instance) -> Union[ifcopensh
...
+                # IFC2X3 has no IfcMapConversion. Georeferencing is instead stored as
+                # ePSet_MapConversion / ePSet_ProjectedCRS property sets on IfcProject
+                # (see ifcopenshell.api.georeference), reached only via the inverse
+                # IsDefinedBy relationship, so the forward-attribute copy above misses
+                # them. Dropping them isn't just data loss: append_asset() still runs
+                # its local-to-global-to-local placement correction per element using
+                # the source's georeferencing, and finding none on the target, leaves
+                # the globalised (Eastings/Northings-scale) coordinates in place,
+                # silently corrupting every extracted element's placement.
+                for rel in element.IsDefinedBy or ():
```
aothms:
```suggestion
                for rel in element.IsDefinedBy:
```

inverses are never none

src/ifcpatch/ifcpatch/recipes/ExtractElements.py
```diff
@@ -113,6 +113,31 @@ def append_asset(self, element: ifcopenshell.entity_instance) -> Union[ifcopensh
...
+                    pset = rel.RelatingPropertyDefinition
+                    if (
+                        pset is not None
+                        and pset.is_a("IfcPropertySet")
+                        and pset.Name
+                        in (
+                            "ePSet_MapConversion",
```
aothms: See https://github.com/buildingSMART/validate/issues/310#issuecomment-5076630963 for a complete list

src/ifcpatch/ifcpatch/recipes/ExtractElements.py
aothms: We should probably check that element is one of Site, Project. https://github.com/buildingSMART/validate/issues/310#issuecomment-4867354635

PR #9044, opened 2026-07-29
util.selector: resolve mats.Category via the material set item

### LINKED ISSUE #7000: Be able to use 'JoinCriteria' with something like `mats.Category`.
### Feature Description

I feel like this was possible at one point (and maybe i'm doing it wrong), but if not, it would be nice to be able to use 'JoinCriteria' with something like `mats.Category`.

That is, remove the following lines in plan and section. 

test files: 

[20250806 - concrete slab and beam.zip](https://github.com/user-attachments/files/21626419/20250806.-.concrete.slab.and.beam.zip)

Image

Image

BIMvoice:
Fixes #7000.

Worth saying first: `material.item.Category` already works today on `v0.8.0`, via the existing `item`/`i` key at `selector.py:442`. @theoryshaw, that may already unblock you without waiting for this.

What was genuinely broken is `mats.Category`, which silently returned `None` instead of erroring. `get_element_value()` resolved `mats`/`materials` through `get_materials()`, which flattens layer, profile and constituent sets down to the `IfcMaterial` each item references. `Category` is declared on `IfcMaterialLayer`, `IfcMaterialProfile` and `IfcMaterialConstituent`, not on `IfcMaterial`, so the lookup could never succeed.

When the next key is `Category`, the set items are now resolved directly. Every other `mats`/`materials` query, including `mats.Name`, still goes through `get_materials()` unchanged.

`get_materials()` itself is untouched. It is public API relied on by `import_ifc`, drawing decoration, and the project and root tools for its flattened list contract, so its return type was not a safe place to change.

Red: `[None, None] == ['Structure', 'Insulation']`. Green: 43 passed in `test_selector.py`, 155 in `test_element.py`.

Two related spots are deliberately out of scope: the same `Category` flattening exists in the `material()` facet filter at `selector.py:1031`, and in the write path at `selector.py:650`.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/util/selector.py
+++ b/src/ifcopenshell-python/ifcopenshell/util/selector.py
@@ -426,7 +426,7 @@ def get_element_value(element: ifcopenshell.entity_instance, query: str) -> Any:
 
 def _get_element_value(element: ifcopenshell.entity_instance, keys: list[str]) -> Any:
     value = element
-    for key in keys:
+    for i, key in enumerate(keys):
         if value is None:
             return
         if key == "type":
@@ -434,7 +434,22 @@ def _get_element_value(element: ifcopenshell.entity_instance, keys: list[str]) -
         elif key in ("material", "mat"):
             value = ifcopenshell.util.element.get_material(value, should_skip_usage=True)
         elif key in ("materials", "mats"):
-            value = ifcopenshell.util.element.get_materials(value)
+            next_key = keys[i + 1] if i + 1 < len(keys) else None
+            if next_key == "Category":
+                # Category is defined on the material set item (IfcMaterialLayer,
+                # IfcMaterialProfile, IfcMaterialConstituent), not on the IfcMaterial
+                # it references, so get_materials() alone can never resolve it.
+                material_set = ifcopenshell.util.element.get_material(value, should_skip_usage=True)
+                if material_set is not None and material_set.is_a("IfcMaterialLayerSet"):
+                    value = list(material_set.MaterialLayers)
+                elif material_set is not None and material_set.is_a("IfcMaterialProfileSet"):
+                    value = list(material_set.MaterialProfiles)
+                elif material_set is not None and material_set.is_a("IfcMaterialConstituentSet"):
+                    value = list(material_set.MaterialConstituents)
+                else:
+                    value = ifcopenshell.util.element.get_materials(value)
+            else:
+                value = ifcopenshell.util.element.get_materials(value)
         elif key == "profiles":
             value = ifcopenshell.util.shape.get_profiles(value)
         elif key == "styles":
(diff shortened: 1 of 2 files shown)


theoryshaw:
A data point from a real model, since this changes where `Category` is read from.

In that project the material **set items** carry no Category, while the `IfcMaterial`s do:

```
IfcWall (3 layer brick veneer)
  layer-level Categories:    [None, None, None]      <- what this PR would read
  material-level Categories: [None, None, 'Brick']   <- what is read today
```

So for this model `mats.Category` would go from `[None, None, 'Brick']` to `[None, None, None]` — losing the only real value. Same for a slab whose materials are categorised `Gravel` and `Concrete` but whose layers are not.

This matters beyond queries: `EPset_Drawing.Metadata` turns these into CSS classes on drawing cut groups, so a stylesheet keyed on `matsCategory-Brick` would silently stop matching. (#9494 makes those classes per-value and therefore actually matchable, which raises the stakes a little.)

The spec does put `Category` on both `IfcMaterial` and the set items, and reading the set item is the more correct answer where it is set. Would it be worth falling back to the referenced material's `Category` when the set item's is unset? That would fix #7000 as intended without dropping data for models that categorise at the material level — which I suspect is the common case, since that is where most authoring tools expose it.

Happy to test a fallback variant against the model above if that helps.

theoryshaw:
Thanks for the quick turnaround on `25b2b4aa` — confirmed against the model that prompted it:

```
brick wall  mats.Category -> [None, None, 'Brick']   (was [None, None, None])
slab        mats.Category -> ['Gravel', 'Concrete']  (was [None, None])
```

One thing that surfaced while testing it, in case it is useful: the same model *does* set `Category` on set items elsewhere — 8 `IfcMaterialLayer`s and 148 `IfcMaterialConstituent`s — and in all 8 layer cases the set item disagrees with its material (`layer.Category='Materials'` over a `GENERIC POCHE` material with none). So both halves of your precedence are exercised by real data, not just the fallback.

That disagreement also caught a bug on my side: #9494 reads a cut layer's metadata from the layer's material, which bypassed your set-item preference and gave a different answer for those walls depending on whether the cut was a layer. Fixed there in `aa737cec` by mirroring the same precedence, so the two agree.

BIMvoice:
Thanks for testing it against the model again, and for the extra numbers.

The 8 layers where the set item disagrees with its material are the useful part. I had verified the precedence against a reconstruction, so knowing that `layer.Category='Materials'` over a `GENERIC POCHE` material with none occurs in real data means both halves are exercised rather than just the fallback. That is better evidence than my tests.

Good catch on #9494 reading the layer's material directly. That is the failure mode your original comment warned about, arriving from the other direction: two code paths answering the same question differently depending on whether the cut happened to be a layer. Mirroring the precedence in `aa737cec` is the right fix, and it is worth the two staying in step if either side changes later.

One note on the checks here, since this PR is showing red. `lint-formatting` fails on exactly three files:

```
src/ifcopenshell-python/ifcopenshell/util/unit.py
src/ifcopenshell-python/test/api/alignment/test_create.py
src/ifcopenshell-python/test/api/alignment/test_referent_names.py
```

None of them is touched here and there are zero complaints against `selector.py`. `compile-and-test` passes. Those three are the black drift on `v0.8.0` that #9241 fixes, which is green and has been open since 2026-08-03. This PR was green on lint in July and went red only by re-running against today's `v0.8.0`, so it is a live example of that drift turning unrelated work red.

Issue #9008, opened 2026-07-27
ExtractElements: Smaller model and several differences between builds: BonsaiPR 0.8.6-alpha260727 and Bonsai 0.8.6-alpha260724


PahlawanJoey:
### Bug Description

#8945 – IFC output changes after recipe performance fix

After fixing the recipe’s performance issues, I noticed differences in the generated IFC model. Initially, these appeared to be colour differences, but validation revealed several additional issues.

All 2,004 walls, including their GlobalIds, core BIM data, and geometry, are present and identical in both outputs.

However:
- Untitled4 is 23.68% smaller, mainly due to deduplication.
- 195 walls have a different rendered appearance.
- 4 walls are missing the presentation layer 22 Binnenwanden NC.
- 26 spatial BaseQuantities relationships are disconnected.
- 3,504 new IfcShapeModel.WR11 errors were introduced due to invalid splitting of ShapeRepresentations.

Output models:
[File build-0.8.6-alpha2607240901 Filesize: 32,4MB](https://acc.autodesk.com/build/share/projects/56c7b26e-c1fc-470d-9d7b-b307bdbc29b2/files?shareId=df997c57-0506-468a-a60d-de6b1076496b)
[File build-0.8.6-alpha2607270854 Filesize: 24,7MB](https://acc.autodesk.com/build/share/projects/56c7b26e-c1fc-470d-9d7b-b307bdbc29b2/files?shareId=d1ef4ee6-f07a-4ab8-91d2-63c2204426e9)
Original input model:
[Original Model](https://acc.autodesk.com/build/share/projects/aa82bc89-d5bb-4fd8-a4b5-7549547ef66d/files?shareId=66864eb8-1d35-4094-98e4-fa8d51955326)

<img width="828" height="1341" alt="Image" src="https://github.com/user-attachments/assets/be008f81-954d-4fdd-8ad6-87d20c0d35c2" />

<img width="773" height="1296" alt="Image" src="https://github.com/user-attachments/assets/c6049c4f-778a-4e93-be7c-f5ca968aed7b" />

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
[...]

PR #9050, opened 2026-07-29
ifcpatch(ExtractElements): keep every element on its presentation layer (#9008)

BIMvoice:
**Stacked on #8320.** This branch is based on #8320's head, so the diff below shows 7 commits of which the first 6 belong to that PR. Only the last one, `5fe9438`, is new here. Merge #8320 first and this reduces to a single commit. Cross-fork PRs cannot target another fork's branch, hence the combined diff.

Fixes the presentation-layer half of #9008.

## Root cause

`src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py:584`:

```python
if (new := self.reuse_identities.get(element_identity)) is not None:
    # Currently known cases requiring attributes reassignment are rels.
    if not new.is_a("IfcRelationship"):
        return
```

`IfcPresentationLayerAssignment` is not an `IfcRelationship`. It is created on the first element that reaches it, and every later element returns here, so `AssignedItems` retains only that first element's items. This is the caveat #8320 explicitly left open: "Layer assignment membership is order dependent... the one respect in which output is not a strict improvement".

## Why the obvious repair was rejected

Simply extending `AssignedItems` per element reintroduced a quadratic cost. `cProfile` over 400 walls: `extend_assigned_items` took **10.13s of a 27.39s run**, driven by **10.8M `entity_instance.walk` calls** re-wrapping the growing aggregate. Forwarding the accumulator to the nested type append, as #8320 does for its own deferred lists, brings the same run to **17.23s against #8320's 17.27s**, i.e. no measurable regression.

## Measurements

@PahlawanJoey's model, 519,335,013 bytes, IFC2X3, query `IfcWall, Description=Baksteen`, 2004 walls, real `Patcher.patch()`:

| | extract | output bytes | layer memberships matching source | walls with correct styles |
|---|---|---|---|---|
| `v0.8.0` | 781.12s | 34,367,810 | 0 / 2004 | 19 / 2004 |
| #8320 head | 17.52s | 26,376,520 | 2 / 2004 | 2004 / 2004 |
| this branch | 17.82s | 26,402,554 | **2004 / 2004** | 2004 / 2004 |

Total `AssignedItems` in the output: 84 -> 4 -> **3595**. The specific layer named in the report, `22 Binnenwanden NC`: 84 -> 1 -> **90**.

## Correctness

Against #8320's head on the same output: identical **376,180** total entities, no per-class count difference, no relationship member-count difference, zero empty member lists in both, and `ifcopenshell.validate` returns the **same 5534 statements with the same breakdown** (3504 `IfcShapeModel.WR11`, 2030 `IfcLocalPlacement.PlacesObject`). 46 GlobalIds differ, all synthesised `IfcRelContainedInSpatialStructure` / `IfcRelAggregates` / `IfcRelDefinesByType`; running #8320's head twice produced the same class of churn (31 GUIDs), so it is run-to-run, not caused by this change.

`v0.8.0` additionally emitted three layers with empty `AssignedItems`, violating the `[1:?]` cardinality. Those three validator statements are gone.

## Tests

`test_shared_presentation_layer_keeps_every_element` fails on #8320's head with `assert 1 == 4` in both IFC4 and IFC2X3, and passes here. `test_ExtractElements.py`: 20 passed, 2 skipped. `test/api/project`: 84 passed. The non-opt-in path used by Bonsai's library append, which has no accumulator, verified separately at 5/5 items in both schemas.

## Out of scope

`deferred_relationship_members` is not forwarded to the nested type append; only the new `deferred_layer_items` is. Changing #8320's behaviour there belongs in #8320.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/project/__init__.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/project/__init__.py
@@ -27,7 +27,7 @@ into your project.
 """
 
 from .. import wrap_usecases
-from .append_asset import append_asset
+from .append_asset import append_asset, flush_deferred_relationship_members
 from .assign_declaration import assign_declaration
 from .create_file import create_file
 from .unassign_declaration import unassign_declaration
@@ -36,6 +36,7 @@ wrap_usecases(__path__, __name__)
 
 __all__ = [
     "append_asset",
+    "flush_deferred_relationship_members",
     "assign_declaration",
     "create_file",
     "unassign_declaration",
--- a/src/ifcopenshell-python/ifcopenshell/api/project/__init__.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/project/__init__.py
@@ -27,7 +27,7 @@ into your project.
 """
 
 from .. import wrap_usecases
-from .append_asset import append_asset, flush_deferred_relationship_members
+from .append_asset import append_asset, flush_deferred_layer_items, flush_deferred_relationship_members
 from .assign_declaration import assign_declaration
 from .create_file import create_file
 from .unassign_declaration import unassign_declaration
@@ -36,6 +36,7 @@ wrap_usecases(__path__, __name__)
 
 __all__ = [
     "append_asset",
+    "flush_deferred_layer_items",
     "flush_deferred_relationship_members",
     "assign_declaration",
     "create_file",
--- a/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/project/append_asset.py
@@ -472,6 +472,7 @@ class Usecase:
                 library=self.settings["library"],
                 element=element_type,
                 reuse_identities=self.reuse_identities,
+                assume_asset_uniqueness_by_name=self.assume_asset_uniqueness_by_name,
             )
             ifcopenshell.api.type.assign_type(
                 self.file,
@@ -612,8 +613,14 @@ class Usecase:
                         continue
                     if skip_not_reused_entities_attr_i is not None and i == skip_not_reused_entities_attr_i:
                         identity = item.wrapped_data.identity()
-                        if (item := self.reuse_identities.get(identity)) is None:
+                        reused = self.reuse_identities.get(identity)
+                        if reused is None:
+                            # reuse_identities is only filled by file_add's per-attribute
+                            # path; the native file.add fast path records added_elements.
+                            reused = self.added_elements.get(item.id())
+                        if reused is None:
                             continue
+                        item = reused
                     else:
                         item = self.add_element(item)
                     new_attribute.append(item)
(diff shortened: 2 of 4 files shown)


BIMvoice:
Independent performance measurements for this PR, plus one finding that is not about this PR but is worth knowing.

**Setup:** Apple M4 Pro, macOS 26.5.2. Two scratch interpreters, one per tree, each verified via `ifcopenshell.__file__` to resolve into its own copy before any timing. Fixture built through the real API: N `IfcWall` products sharing exactly one `IfcRelAssociatesMaterial`, then appended one at a time via `project.append_asset`. Median of 3 repetitions, spread under 2% at every point.

### Timing

| N | baseline (s) | this PR (s) | ratio |
|---|---|---|---|
| 100 | 0.0907 | 0.0116 | 7.8x |
| 250 | 0.5237 | 0.0343 | 15.3x |
| 500 | 2.0437 | 0.0849 | 24.1x |
| 1000 | 8.2035 | 0.2447 | 33.5x |
| 2000 | 33.759 | 0.7726 | 43.7x |
| 4000 | 135.892 | 2.7188 | 50.0x |
| 8000 | not run | 10.357 | n/a |

The N=8000 baseline was not run, so no number is given for it rather than an extrapolation.

### The scaling, which matters more than any single ratio

**The baseline is cleanly quadratic**, measured rather than inferred. Each doubling of N roughly quadruples the time: 500 to 1000 is 4.01x, 1000 to 2000 is 4.12x, 2000 to 4000 is 4.03x. That is consistent with the mechanism this PR describes, one shared relationship rebuilt once per appended element.

**The fix is genuinely linear in isolation.** With `assume_asset_uniqueness_by_name=False`, going from N=2000 to N=8000 (4x) took 0.1477s to 0.5920s, a ratio of 4.01x, so an exponent of about 1.00.

### A separate pre-existing bottleneck, not introduced by this PR

Under the **default** `assume_asset_uniqueness_by_name=True`, the fixed tree drifts back toward quadratic at large N: 4000 to 8000 is 3.81x for a 2x increase, an exponent of about 1.93.

Profiling at N=2000 shows the residual cost is **not** in the deferred-relationship-members code this PR touches. It is in `SafeRemovalContext.__exit__` (`append_asset.py`, around lines 263 to 281, untouched by this PR), which on every append iterates the whole of `reuse_identities`:

```python
for identity, element in self.reuse_identities.items():
    if element in removed_elements:
        ...
```

That is O(size of `reuse_identities`) per call, summing to O(n^2) across N calls, independently of this PR. So this PR fixes the bottleneck it set out to fix, and a different pre-existing one becomes dominant at scale. Flagging it as a follow-up rather than something this PR should grow to cover.

### Correctness check, and a claim I could not reproduce

Ran `ifcopenshell.validate.validate()` on the output of both trees, using two fixtures built through the real API and run through the actual `ExtractElements` recipe, including a reproduction of this PR's own toughest regression case (containers, storeys, a wall type, psets on every element, one shared `IfcRelAssociatesMaterial` across walls and type, extracted by an `IfcWall` query so the type and containers are reached only through inverses).

**Both trees produced 0 empty member lists and 0 validation statements.** Material associations and psets were preserved correctly in both.

So I did **not** reproduce the empty `RelatedObjects` data loss on `v0.8.0` with these fixtures. It may have existed only in an intermediate commit of this series rather than in the base or the final head. Stating that rather than repeating a claim I could not confirm.

### Not measured

One fixture shape only (N elements sharing one relationship via material association). Not tested: the presentation-layer path, `IfcRelDefinesByProperties`, `IfcRelOverridesProperties` under IFC2X3, multiple relationships per element, nested shared-relationship graphs, or memory. The PR's own pytest suite was not run; these are direct harness calls into the same code paths.

PR #9054, opened 2026-07-29
copy_class: don't share the original's annotations with the duplicate (#9049, cause behind #4014)

### LINKED ISSUE #9049: copy_class shares the original's annotations with the duplicate, assigning them to two products (cause behind #4014)
### Bug Description

Duplicating a product that has annotations assigned to it (Shift+D in Bonsai, or `ifcopenshell.api.root.copy_class` directly) does not give the copy its own annotations, and does not leave the copy without annotations either. It clones the `IfcRelAssignsToProduct` relationship, points the clone at the new product, and keeps the **original's annotations** in `RelatedObjects`. Both products then claim the same annotation entities, so every annotation involved ends up assigned to two products.

This appears to be the missing cause behind #4014. PR #8373 says as much in its own scope note: *"this does not address how annotations accumulate multiple products in the first place (still under investigation via the diagnostic warning)."* This is that path.

**Minimal reproduction** (no Blender required):

```python
import ifcopenshell, ifcopenshell.api.root, ifcopenshell.guid

f = ifcopenshell.file(schema="IFC4")
wall = ifcopenshell.api.root.create_entity(f, ifc_class="IfcWall")
annotation = ifcopenshell.api.root.create_entity(f, ifc_class="IfcAnnotation")
f.create_entity("IfcRelAssignsToProduct", GlobalId=ifcopenshell.guid.new(),
                RelatedObjects=[annotation], RelatingProduct=wall)

copy = ifcopenshell.api.root.copy_class(f, product=wall)

rels = [r for r in annotation.HasAssignments if r.is_a("IfcRelAssignsToProduct")]
print(len(rels), [r.RelatingProduct.id() for r in rels])
```

```
before: annotation assigned to 1 product(s)
after copy_class: annotation assigned to 2 product(s): [1, 4]   # original wall #1, copy #4
```

**Root cause.** In `src/ifcopenshell-python/ifcopenshell/api/root/copy_class.py`, `IfcRelAssignsToProduct` matches none of the special cases in `copy_indirect_attributes`, so it falls through to the generic `else` branch:

```python
else:
    for i, value in enumerate(inverse):
        if value == from_element:
            new_inverse = ifcopenshell.util.element.copy(self.file, inverse)
            new_inverse[i] = to_element
```

`value == from_element` matches at `RelatingProduct`, so the whole relationship is copied and only that one attribute is swapped. `RelatedObjects` — the original's annotations — is carried over untouched.

**Impact in a real project.** In a production model of mine, 60 of 808 annotations are assigned to more than one product, across 22 rel-sets that are exact clones (same `RelatedObjects`, different `RelatingProduct`). One `IfcSpace` was duplicated twice, so a single annotation is now assigned to three spaces:

```
annotation #3542995
    via rel #3543012 -> #3540997  IfcSpace 06XX6WklnFdeYNmzrdhn4i
    via rel #3550684 -> #3550678  IfcSpace 3LKyBBG0LDLwxBGxvi_Qyn
    via rel #3552280 -> #3552274  IfcSpace 2E30n7LafENhO3LEp8hJYt
```

The practical consequence is that `tool.Drawing.get_assigned_product` returns whichever rel happens to come first, so an annotation can silently label the wrong element. `tool.Drawing.get_assigned_product_workaround` exists to paper over exactly this and prints the multi-product warning pointing at #4014.

**Relationship to existing work.**

- #6621 / PR #8032 fix the case where the *annotation* is duplicated (alone, or together with its product). I reimplemented `fix_assigns_to_product_after_duplication` from that PR verbatim and ran it against both scenarios: it fixes duplicating product + annotation together, but the guard `if not rel.is_a("IfcRelAssignsToProduct") or new_elem not in rel.RelatedObjects: continue` skips this case entirely, because here the new element is the `RelatingProduct`, not a related object. The two fixes are complementary, not competing.
- #8373 / #8374 clean up annotations that already have multiple products; they don't prevent them being created.

**Suggested fix.** Skip the relationship when the element being copied is the `RelatingProduct`, the same way `IfcRelAggregates` and `IfcRelContainedInSpatialStructure` are already skipped when the copied element is on the relating side. The duplicate then simply starts with no annotations.

The opposite direction must keep working: when an *annotation* is copied it takes the list branch, which appends the copy to the same relationship, and that behaviour is what the copy-annotation-to-drawing feature relies on.

Note this is a behaviour change for every `copy_class` caller, not just Bonsai's duplicate path: a duplicated product will no longer inherit assignments made *to* it through `IfcRelAssignsToProduct`, which can include objects other than annotations.

theoryshaw:
Closes #9049. Very likely the missing cause behind #4014.

Duplicating a product that has annotations assigned to it does not give the copy its own annotations, and does not leave the copy without annotations either. `IfcRelAssignsToProduct` matched none of the special cases in `copy_indirect_attributes`, so it fell through to the generic `else` branch, where `value == from_element` matches at `RelatingProduct`. The whole relationship was copied with only that attribute swapped, so `RelatedObjects` still held the original's annotations and both products claimed them.

```python
f = ifcopenshell.file(schema="IFC4")
wall = ifcopenshell.api.root.create_entity(f, ifc_class="IfcWall")
annotation = ifcopenshell.api.root.create_entity(f, ifc_class="IfcAnnotation")
f.create_entity("IfcRelAssignsToProduct", GlobalId=ifcopenshell.guid.new(),
                RelatedObjects=[annotation], RelatingProduct=wall)

copy = ifcopenshell.api.root.copy_class(f, product=wall)
# before: annotation assigned to 1 product
# after:  annotation assigned to 2 products [wall, copy]
```

Every annotation involved then reports two assigned products, which is the symptom tracked in #4014. `get_assigned_product` returns whichever relationship comes first, so an annotation can silently label the wrong element, and `tool.Drawing.get_assigned_product_workaround` exists to paper over exactly this. In one production model 60 of 808 annotations were affected across 22 cloned relationship sets, with one annotation reaching three spaces after its space was duplicated twice.

### The fix is deliberately narrow

The relationship is now handled explicitly when the copied element is the `RelatingProduct`. Annotations are dropped from the copy, since an annotation labels exactly one product and must not be shared.

Everything else assigned to the product is still inherited. That matters: the same relationship carries the 4D **Output** link from a task to the product it builds, a task may legitimately output many products, and both `sequence.assign_product` and `drawing.assign_product` append to `relating_product.ReferencedBy[0]` — so a single relationship can hold a task and an annotation at once. Skipping the relationship wholesale would have silently stripped task outputs from every duplicated element. A relationship left with no related objects is skipped entirely.

### Relationship to other open work

- **#6621 / PR #8032** fix the cases where the *annotation* is duplicated, alone or together with its product. That guard requires the new element to be in `RelatedObjects`, so it never reaches this path, where the new element is the `RelatingProduct`. The two are complementary.
- **#8373 / #8374** clean up annotations that already have multiple products; they don't prevent them being created. #8373's own scope note says it "does not address how annotations accumulate multiple products in the first place (still under investigation)" — this is that path.

**Note for whoever merges:** PR #8032 needs a one-case update to work alongside this. Its case A only removes the duplicated annotation from the original product's relationship, relying on `copy_class` having cloned that relationship onto the new product so its case B could fix it up. Once `copy_class` stops doing that, nothing re-links the annotation copy and it ends up assigned to nothing. I've pushed that update to #8032's branch and verified it is safe in either merge order. Details in the comments on both PRs.

`tool.Drawing.clear_annotation_relationships`, which deletes the stale relationships this used to create when duplicating a drawing, becomes a no-op rather than a necessity.

### Testing

Four tests in `test_copy_class.py`: annotations stay on the original, task outputs are still inherited, a mixed relationship keeps the task and drops the annotation, and copying an annotation still leaves it assigned to the same product. The first and third fail without the fix.

The full ifcopenshell api suite passes — 1449 tests, excluding a pre-existing collection error in `test/api/alignment/test_update_fallback_position.py` that fails identically without these changes.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/root/copy_class.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/root/copy_class.py
@@ -52,6 +52,11 @@ def copy_class(file: ifcopenshell.file, product: ifcopenshell.entity_instance) -
       will also be copied.
     * Path connectivity is not copied, as there is no guarantee that the
       connections are still valid.
+    * Annotations assigned to the product are not copied. They stay assigned to
+      the original only, so the copy starts with no annotations. Sharing them
+      would leave one annotation labelling two products, which is invalid.
+      Other objects assigned to the product, such as tasks that output it, are
+      still inherited by the copy.
 
     :param product: The IfcProduct to copy.
     :return: The copied product
@@ -135,6 +140,18 @@ class Usecase:
                 continue
             elif inverse.is_a("IfcRelDefinesByType") and inverse.RelatingType == from_element:
                 continue
+            elif inverse.is_a("IfcRelAssignsToProduct") and inverse.RelatingProduct == from_element:
+                # An annotation labels exactly one product, so it must not be shared. The
+                # generic branch below would copy the relationship and swap the product,
+                # leaving the original's annotations assigned to both products. See #4014.
+                # Other assignments are fine to inherit: the same relationship also holds
+                # tasks that output this product, and a task may output many products.
+                related_objects = [e for e in inverse.RelatedObjects if not e.is_a("IfcAnnotation")]
+                if not related_objects:
+                    continue
+                new_inverse = ifcopenshell.util.element.copy(self.file, inverse)
+                new_inverse.RelatedObjects = related_objects
+                new_inverse.RelatingProduct = to_element
             elif inverse.is_a("IfcRelVoidsElement") and inverse.RelatingBuildingElement == from_element:
                 opening = inverse.RelatedOpeningElement
                 # We don't copy filled openings, since there is no guarantee the filling is also copied
--- a/src/ifcopenshell-python/test/api/root/test_copy_class.py
+++ b/src/ifcopenshell-python/test/api/root/test_copy_class.py
@@ -24,10 +24,12 @@ import ifcopenshell.api.geometry
 import ifcopenshell.api.group
 import ifcopenshell.api.pset
 import ifcopenshell.api.root
+import ifcopenshell.api.sequence
 import ifcopenshell.api.spatial
 import ifcopenshell.api.system
 import ifcopenshell.api.type
 import ifcopenshell.api.unit
+import ifcopenshell.guid
 import ifcopenshell.util.system
 import test.bootstrap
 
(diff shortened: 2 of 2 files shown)


theoryshaw:
Note on how this interacts with #8032, since the two touch the same relationship from opposite sides and are best reviewed together.

**They cover different paths.** #8032 fixes duplication where the *annotation* is duplicated, alone or together with its product. Its guard requires the new element to be in `RelatedObjects`, so it never reaches the path fixed here, where the new element is the `RelatingProduct` and the annotations left in `RelatedObjects` belong to the original. Neither PR subsumes the other.

**But #8032 needed a one-case update to survive this change**, which I have pushed to its branch as df35a895. Its case A only removed the duplicated annotation from the original product's relationship, relying on `copy_class` having cloned that relationship onto the new product so its case B could fix it up. Once this PR stops the cloning, nothing re-links the annotation copy and it ends up assigned to nothing. Case A now hands it to the duplicated product instead.

**Merge order does not matter.** The updated case A skips when the element is already assigned to the target product, which is exactly what happens under the old `copy_class`. Verified over five scenarios against both versions of `copy_class`: with the old one, every case #8032 already handled still passes and the only failure is duplicating a product alone, which is the bug this PR fixes.

So either can land first; both should land for the full picture of #4014. #8373 and #8374 remain useful for cleaning up models that already carry the damage — in the production model I checked, 60 of 808 annotations.

PR #9072, opened 2026-07-30
core.aggregate: guard the value actually used in unassign_object

BIMvoice:
unassign_object() looks up the relating object itself when the caller
omits relating_obj. The lookup result was stored in relating_element,
but the code then checked the unrelated related_element (already known
truthy at that point) before calling ifc.get_object(relating_element).

When an object has no aggregate parent, relating_element is None but
related_element is still truthy, so the guard passed anyway and the
function called ifc.get_object(None), which crashes on element.id(),
and then proceeded to run the unassign and collector side effects on
garbage data. Fixed by checking relating_element, the value actually
being used.

No caller currently omits relating_obj, so this was latent rather than
already crashing in production, but the Optional default exists
specifically to support that call shape.

### DIFF
--- a/src/bonsai/bonsai/core/aggregate.py
+++ b/src/bonsai/bonsai/core/aggregate.py
@@ -65,7 +65,7 @@ def unassign_object(
     container = aggregate.get_container(related_element)
     if not relating_obj:
         relating_element = aggregate.get_relating_object(related_element)
-        if related_element:
+        if relating_element:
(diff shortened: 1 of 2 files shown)


Moult:

Thanks Petru. Retargeted to v0.9.0 and pushed to your branch: same fix and test, with the lookup and guard folded into one walrus condition and the explanatory comment in the test dropped. Rebased onto current v0.9.0 as a single commit. Merging once CI is green.

PR #9082, opened 2026-07-30
ifc4d: stop MSP and P6 export truncating sub-day durations

BIMvoice:
ifc2msp.py and ifc2p6.py converted ifc2datetime()'s isodate.Duration by
reading only .days, so any lag or task duration under 24 hours (which
lands in .seconds, not .days) was silently dropped. PT12H0M0S exported
as LinkLag 0; P1DT12H0M0S dropped its 12 hour remainder entirely.

MSPDI LinkLag also never emitted LagFormat, and ignored TimeLag's
DurationType (WORKTIME vs ELAPSEDTIME) even though it was already
scaling by hours_per_day, so an elapsed lag round-tripped through the
wrong calendar divisor. Both are the same "read the whole duration"
class of bug as the days-only truncation, so this now reads
DurationType and emits the matching LagFormat.

Verified against msp2ifc.py from PR #9043 (predecessor lag import
fix): every fixed export case here reimports to the exact original
lag, including a lead (negative lag) and a mixed day+hour duration.

This file was generated with the assistance of an AI coding tool.

### DIFF
--- a/src/ifc4d/ifc4d/ifc2msp.py
+++ b/src/ifc4d/ifc4d/ifc2msp.py
@@ -25,6 +25,27 @@ import ifcopenshell.util.date
 import ifcopenshell.util.sequence
 
 
+def duration_to_hours(duration, hours_per_day, is_elapsed=False):
+    """Convert an ifcopenshell.util.date.ifc2datetime() result to a float number of hours.
+
+    ``duration`` is either an isodate.Duration or a datetime.timedelta (both
+    ifc2datetime() return types), which both expose ``.days``, ``.seconds``
+    and ``.microseconds``. Reading only ``.days`` silently drops any lag or
+    task duration below 24 hours, since ISO durations like "PT12H0M0S" carry
+    that 12 hours in ``.seconds``, not ``.days``.
+
+    MSPDI's own units are elapsed hours regardless of format: a WORKTIME
+    duration is elapsed hours reinterpreted as a day count via
+    ``hours_per_day`` (see ``msp2ifc.py``'s ``create_rel_sequences``, which
+    builds ``timedelta(days=lag_hours / hours_per_day)`` on import), so the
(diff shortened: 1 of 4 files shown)


BIMvoice:
Flagging two things on these lines before this merges, both measured rather than read.

**1. A percentage lag crashes the export.**

`IfcLagTime.LagValue` can be an `IfcRatioMeasure` (a lag expressed as a percentage of the predecessor's duration) rather than an `IfcDuration`. `ifcopenshell.util.date.ifc2datetime()` returns `None` for a bare float:

```
>>> ifcopenshell.util.date.ifc2datetime(0.5)
None
>>> None.days
AttributeError: 'NoneType' object has no attribute 'days'
```

`ifc2msp.py` then does exactly that:

```python
duration = ifcopenshell.util.date.ifc2datetime(rel.TimeLag.LagValue.wrappedValue)
ET.SubElement(predecessor_link, "LinkLag").text = str(duration.days * ...)
```

`ifc2p6.py` has the same shape. This PR's `duration_to_hours()` helper does not change that, so the crash survives the change as written.

Worth noting `cascade_schedule.py` and `edit_lag_time.py` already handle both forms, so the duration-only assumption is inconsistent with the rest of the codebase rather than a documented limitation. The same assumption caused a segfault elsewhere, fixed in #9110.

**2. A zero lag is silently dropped.**

The guard is `if rel.TimeLag and rel.TimeLag.LagValue:`. A lag of `0` or `0.0` is falsy, so an explicitly authored zero lag is skipped rather than exported as zero. That is a real distinction in a programme: "no lag specified" and "explicitly zero lag" are different statements, and only one of them survives.

**Not proposing a patch here**, since these are your lines and a competing change would just conflict. Raising it so it can be handled in this PR or as a follow-up, whichever you prefer.

Happy to prepare either fix if useful.

PR #9117, opened 2026-07-30
ifcwrap: fix two SIGSEGVs on defined and enumeration types

BIMvoice:
> **These changes have NOT been compiled or run.** The crashes below are reproduced and the root causes identified to the line, but the build could not be completed on this machine. Please treat this as a diagnosis with proposed one-line guards rather than a tested patch. Build details at the end.

Two confirmed segfaults, both reachable from ordinary Python, both in `src/ifcwrap/IfcParseWrapper.i`, both the same archetype.

## Crash 1: instantiating any defined or enumeration type with `None`

```python
>>> f.create_entity("IfcDuration", None)
Segmentation fault (exit code 139)
```

`setArgumentAsNull` dereferenced `declaration().as_entity()` unguarded. That returns null for a defined type or an enumeration type.

## Crash 2: `get_argument(name)` on any defined type

```python
>>> f.create_entity("IfcDuration", "P1D").wrapped_data.get_argument("wrappedValue")
Segmentation fault (exit code 139)
```

Same unguarded dereference at line 463. **The sibling `get_argument_index(const std::string&)` seven lines below already guards the identical call**, which is what identified it.

The guard here costs nothing, because the function already has an `if (i == -1)` branch raising a clean `std::runtime_error`. A null declaration simply routes into that existing path.

## Crash surface, measured

Every defined type and every enumeration type is affected: `IfcLabel`, `IfcText`, `IfcIdentifier`, `IfcDuration`, `IfcDateTime`, `IfcInteger`, `IfcReal`, `IfcBoolean`, and enums such as `IfcElementCompositionEnum`.

**Regular entities were never affected.** They already raise a clean `TypeError`, and an optional attribute set to `None` is still accepted. Only direct instantiation or inspection of a defined or enumeration type crashed.

## Why the wrapper and not the parser

A survey of roughly 140 downcast call sites found the parser layer consistently guarded. `src/ifcparse/*.cpp`, `IfcFile.cpp` and `storage.h` all check before dereferencing, and their storage layer independently converts wrong-type and out-of-range access into clean `IfcException`s. `IfcFile::getUnit()` looked exploitable but was tested and rejected for exactly that reason: `get_attribute_value` throws first.

**The `%extend` methods in the wrapper have no equivalent safety net**, which is why the archetype survives there and nowhere else. The two fixed here plus the one in the first commit appear to be the complete set in that file.

A third instance in `setArgumentAsString`'s `Argument_ENUMERATION` branch is also guarded here, but is **latent rather than live**: enumerating every schema present in this build (IFC2X3, IFC4, IFC4X1, IFC4X2 and the IFC4X3 variants) found none defining a type declaration that aliases an enumeration type, so no input appears to reach it. It is guarded because it is the same file, the same pattern and a one-line change. Some IFC4X3 release candidates are absent from this build, so that check is not exhaustive.

Unguarded sites in `parse_ifcxml.cpp` were deliberately left alone: `IfcParse::parse_ifcxml()` throws `"IFC-XML import temporarily disabled"` at function entry, so that code is unreachable by maintainer choice. Verified by calling it.

## How this was found

Through a real caller, fixed separately in #9110: `api/sequence/duplicate_task.py` assumed `IfcLagTime.LagValue` was always an `IfcDuration`, but it can be an `IfcRatioMeasure`. The float went through `ifc2datetime()`, which returns `None` for a float, and that `None` reached `create_entity`.

## Build status, stated plainly

The docker build image is `linux/amd64` only, so on an Apple Silicon host it runs under QEMU emulation. A stale root-owned ccache volume failed the first run; after fixing that, the second run reached real compilation and `cc1plus` was **OOM-killed** at default parallelism against the container's 8 GB ceiling. A retry at reduced parallelism had no realistic completion time, so it was stopped rather than left running for hours.

Whether a native arm64 build path exists was **not** established.

So: both crashes are reproduced with real exit codes captured directly rather than through a pipe, the root causes are identified to the line, and the guards follow the conventions of neighbouring methods in the same file. **They have not been built.** Anyone with a working native build can confirm in seconds, and I would rather hand over an honest diagnosis than claim a verification I did not perform.

### DIFF
--- a/src/ifcwrap/IfcParseWrapper.i
+++ b/src/ifcwrap/IfcParseWrapper.i
@@ -575,12 +575,25 @@ private:
 		if (arg_type == IfcUtil::Argument_STRING) {
 			self->set_attribute_value(i, a);	
 		} else if (arg_type == IfcUtil::Argument_ENUMERATION) {
-			const IfcParse::entity* value_entity = $self->declaration().schema()->declaration_by_name($self->declaration().type())->as_entity();
-			if (!value_entity) {
+			const IfcParse::declaration* decl = $self->declaration().schema()->declaration_by_name($self->declaration().type());
+			const IfcParse::parameter_type* pt = 0;
+			if (const IfcParse::entity* value_entity = decl->as_entity()) {
+				pt = value_entity->attribute_by_index(i)->type_of_attribute();
+			} else if (i == 0) {
+				// A defined type can wrap an enumeration, as some IFC4.3 release
+				// candidates did, so resolve through its declared type as well.
+				if (const IfcParse::type_declaration* td = decl->as_type_declaration()) {
+					pt = td->declared_type();
+				}
+			}
+			const IfcParse::named_type* nt = pt ? pt->as_named_type() : 0;
+			const IfcParse::enumeration_type* enum_type = nt ? nt->declared_type()->as_enumeration_type() : 0;
+			if (!enum_type && i == 0) {
+				enum_type = decl->as_enumeration_type();
+			}
+			if (!enum_type) {
 				throw IfcParse::IfcException("Attribute not set");
 			}
-			const IfcParse::enumeration_type* enum_type = value_entity->
-			attribute_by_index(i)->type_of_attribute()->as_named_type()->declared_type()->as_enumeration_type();
 			self->set_attribute_value(i, EnumerationReference(enum_type, enum_type->lookup_enum_offset(a)));
 		} else if (arg_type == IfcUtil::Argument_BINARY) {
 			if (IfcUtil::valid_binary_string(a)) {
--- a/src/ifcwrap/IfcParseWrapper.i
+++ b/src/ifcwrap/IfcParseWrapper.i
@@ -523,7 +523,7 @@ private:
 	}
 
 	void setArgumentAsNull(unsigned int i) {
-		bool is_optional = $self->declaration().as_entity()->attribute_by_index(i)->optional();
+		bool is_optional = $self->declaration().as_entity() && $self->declaration().as_entity()->attribute_by_index(i)->optional();
 		if (is_optional) {
 			self->set_attribute_value(i, Blank{});
 		} else {
--- a/src/ifcwrap/IfcParseWrapper.i
+++ b/src/ifcwrap/IfcParseWrapper.i
@@ -580,8 +580,8 @@ private:
 			if (const IfcParse::entity* value_entity = decl->as_entity()) {
 				pt = value_entity->attribute_by_index(i)->type_of_attribute();
 			} else if (i == 0) {
-				// A defined type can wrap an enumeration, as some IFC4.3 release
-				// candidates did, so resolve through its declared type as well.
+				// e.g. IfcFacilityPart.PredefinedType: IfcFacilityPartTypeSelect (IFC4X3 RC1-4)
+				// is a select of enums, so resolve through the declared type as well.
 				if (const IfcParse::type_declaration* td = decl->as_type_declaration()) {
 					pt = td->declared_type();
 				}
--- a/src/ifcwrap/IfcParseWrapper.i
+++ b/src/ifcwrap/IfcParseWrapper.i
@@ -580,8 +580,9 @@ private:
 			if (const IfcParse::entity* value_entity = decl->as_entity()) {
 				pt = value_entity->attribute_by_index(i)->type_of_attribute();
 			} else if (i == 0) {
-				// e.g. IfcFacilityPart.PredefinedType: IfcFacilityPartTypeSelect (IFC4X3 RC1-4)
-				// is a select of enums, so resolve through the declared type as well.
+				// A defined type can wrap an enumeration, so resolve through its
+				// declared type as well. No such case is reachable in any schema
+				// currently built; this guard is defensive.
 				if (const IfcParse::type_declaration* td = decl->as_type_declaration()) {
 					pt = td->declared_type();
 				}
--- a/src/ifcwrap/IfcParseWrapper.i
+++ b/src/ifcwrap/IfcParseWrapper.i
@@ -461,7 +461,8 @@ private:
 	}
 
 	AttributeValue get_argument(const std::string& a) {
-		auto i = $self->declaration().as_entity()->attribute_index(a);
+		auto entity_decl = $self->declaration().as_entity();
+		auto i = entity_decl ? entity_decl->attribute_index(a) : -1;
 		if (i == -1) {
 			throw std::runtime_error("Attribute '" + a + "' not found on entity named " + $self->declaration().name());
 		}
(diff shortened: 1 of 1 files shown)


src/ifcwrap/IfcParseWrapper.i
```diff
@@ -574,7 +575,11 @@ private:
 		if (arg_type == IfcUtil::Argument_STRING) {
 			self->set_attribute_value(i, a);	
 		} else if (arg_type == IfcUtil::Argument_ENUMERATION) {
-			const IfcParse::enumeration_type* enum_type = $self->declaration().schema()->declaration_by_name($self->declaration().type())->as_entity()->
+			const IfcParse::entity* value_entity = $self->declaration().schema()->declaration_by_name($self->declaration().type())->as_entity();
```
aothms: There are no longer test cases for this in the main schemas, but older 4.3 RCs did have a case of a defined type that wraps an enumeration. In order to handle that we should not only consider as_entity() but in case of i == 0 we should also apply it to a defined type.

BIMvoice:
Thanks, that is a case I had not accounted for. Pushed a change.

The enumeration lookup now resolves in three steps rather than assuming an entity:

```cpp
if (const IfcParse::entity* value_entity = decl->as_entity()) {
    pt = value_entity->attribute_by_index(i)->type_of_attribute();
} else if (i == 0) {
    // A defined type can wrap an enumeration, as some IFC4.3 release
    // candidates did, so resolve through its declared type as well.
    if (const IfcParse::type_declaration* td = decl->as_type_declaration()) {
        pt = td->declared_type();
    }
}
const IfcParse::named_type* nt = pt ? pt->as_named_type() : 0;
const IfcParse::enumeration_type* enum_type = nt ? nt->declared_type()->as_enumeration_type() : 0;
if (!enum_type && i == 0) {
    enum_type = decl->as_enumeration_type();
}
```

This mirrors what `helper_fn_attribute_type` already does near the top of the same file for the `i == 0` case, which handles both `as_type_declaration()` and `as_enumeration_type()`.

`type_declaration::declared_type()` returns a `parameter_type*`, the same as `type_of_attribute()`, so the remainder of the chain through `as_named_type()->declared_type()->as_enumeration_type()` is unchanged and now shared by both paths.

**Still not built.** There is no rebuilt wrapper available here, so this is reasoned from the headers rather than tested, and the commit message says so rather than claiming otherwise. If you know of a 4.3 RC schema containing such a defined type I am happy to construct a case against it; otherwise this needs your eye more than mine.

aothms:
Express:

ENTITY IfcFacilityPart
 SUBTYPE OF (IfcSpatialStructureElement);
	PredefinedType : IfcFacilityPartTypeSelect;

with IfcFacilityPartTypeSelect being a Select of Enums and therefore resulting in type instances with an underlying enumeration.

> cc1plus was OOM-killed at default parallelism against the container's 8 GB ceiling

The schemas are huge. Reduce parallelism to 1 when compiling and please try again.

aothms:
edit: from the IFC4x3_RC4 schema

BIMvoice:
The build finished, and `-j1` was the fix. Thank you. Peak was around 2.5 GB on the largest schema translation units, so at default parallelism it was comfortably exceeding the container ceiling.

**Both original SIGSEGVs are gone.** Each reproduced in its own process so one crash could not mask the other:

* `create_entity("IfcDuration", None)` now raises a clean `TypeError`, exit 1 rather than 139.
* `get_argument("wrappedValue")` on a defined type now raises a clean `RuntimeError`, also exit 1.

Full wrapper suite: 444 passed, 13 skipped with stated unrelated reasons, 0 failed.

**I could not reproduce your `IfcFacilityPart` case, and I would rather say so than imply otherwise.** This build's schema set is IFC2X3, IFC4, IFC4X1, IFC4X2 and IFC4X3 with ADD1, ADD2 and TC1, with no RC1-4. In all four IFC4X3 variants present, `IfcFacilityPart` has no `PredefinedType` attribute at all, only `CompositionType` and `UsageType`. I also enumerated every schema in the build looking for any defined type whose declared type resolves to an enumeration, and found none.

So the guard is unexercised by any data reachable here, which matches what the PR body already said about it being latent rather than live. Your example does appear to be exactly right for the release candidates, which is where `IfcFacilityPartTypeSelect` on `PredefinedType` existed; it did not survive into the released schemas. I have updated the code comment to name it rather than referring vaguely to release candidates.

If you have an RC1-4 schema or a minimal .exp snippet, I will rerun against that specifically.

aothms:
Sources (.cpp .h) for these schemas are present in the source tree under
src/ifcparse. Just excluded from the default SCHEMA_VERSIONS cmake
variable.

Op za 1 aug 2026, 01:33 schreef Petru Conduraru ***@***.***>:

BIMvoice:
Thank you for the pointer, that let us build a real IFC4X3_RC4 wrapper and test this properly instead of guessing. The answer is not the one I expected, so here it is plainly.

**Your case does not exercise the fix.** Built with `-DSCHEMA_VERSIONS="...;4x3_rc4"`, a genuine `IfcFacilityPart` in an RC4 file, and the output is byte-identical with and without the change.

The reason is that `IfcFacilityPartTypeSelect` is a `select_type` over five enumerations (`IfcFacilityPartCommonTypeEnum`, `IfcBridgePartTypeEnum` and so on), not a `type_declaration` wrapping one. `IfcUtil::from_parameter_type` classifies a SELECT-typed attribute as `Argument_ENTITY_INSTANCE`, so `helper_fn_attribute_type` never routes it into the `Argument_ENUMERATION` dispatch that this PR touches.

```
schema.declaration_by_name("IfcFacilityPartTypeSelect")  -> select_type
  .as_type_declaration()   -> None
  .as_enumeration_type()   -> None
entity_decl.argument_types()["PredefinedType"]           -> 'ENTITY INSTANCE'
```

Setting it the natural way is rejected earlier, by the Python type check rather than by anything here:

```
fp.PredefinedType = "STOREY"
TypeError: attribute 'PredefinedType' for entity 'IFC4X3_RC4.IfcFacilityPart' is
expecting value of type 'ENTITY INSTANCE', got 'str'
```

Calling `setArgumentAsString` directly to force the C++ path gives `RuntimeError: Attribute not set` **both with and without the fix**, because it falls through to the pre-existing catch-all rather than the branch we added. And `fp.PredefinedType = file.create_entity("IfcFacilityPartCommonTypeEnum", "STOREY")` works, and worked before this PR too.

So I have corrected the code comment, which cited your example. The guard is still right for a defined type that genuinely wraps an enumeration, but I enumerated every `type_declaration` in all nine built schemas and found none, so it is defensive rather than live and the comment now says so instead of naming a case that does not reach it. **If you know of a schema where such an attribute really is classified as ENUMERATION, I will test that one specifically.**

**The two original SIGSEGVs are verified on this same RC4 build.** Reverted just `IfcParseWrapper.i` to the parent commit and rebuilt:

```
without fix:  create_entity("IfcDuration", None)                    -> exit 139
              get_argument("wrappedValue") on a defined type        -> exit 139
with fix:     TypeError: attribute 'wrappedValue' is not optional
              RuntimeError: Attribute 'wrappedValue' not found
```

**One unrelated thing found while building.** `src/ifcgeom/mapping/IfcOpenCrossProfileDef.cpp:41` calls `inst->OffsetPoint()` unconditionally, but RC4's `IfcOpenCrossProfileDef` has no such attribute (`Ifc4x3_rc4-definitions.h` has no `SCHEMA_IfcOpenCrossProfileDef_HAS_OffsetPoint`; add2 does), so any RC1-4 build fails to compile:

```
error: 'const class Ifc4x3_rc4::IfcOpenCrossProfileDef' has no member named 'OffsetPoint'
```

I worked around it locally with an `#ifdef` guard mirroring the convention already used in `IfcSectionedSurface.cpp`, purely to get the build through. Nothing committed, since it is not this PR's business, but it means the RC schemas cannot currently be built without that patch. Happy to send it separately if useful.

PR #9139, opened 2026-07-30
cmake: wrap the OpenCASCADE interface paths in BUILD_INTERFACE

yekose:
## Problem

`OpenCASCADE_INTERFACE` is `install(EXPORT)`ed, so its interface properties land in `IfcOpenShellTargets.cmake`. CMake refuses to export a target whose interface paths point into the build or source tree, because such a path is meaningless on the machine that consumes the installed package:

```
CMake Error in cmake/CMakeLists.txt:
  Target "OpenCASCADE_INTERFACE" INTERFACE_INCLUDE_DIRECTORIES property
  contains path:
    ".../build/vcpkg_installed/x64-windows/include/opencascade"
  which is prefixed in the build directory.
```

This stays invisible when OCCT is installed system-wide or comes from conda, which is presumably why CI does not hit it. It appears as soon as OCCT lives inside the consumer's build tree — for instance with **vcpkg in manifest mode**, where dependencies are installed under `<build>/vcpkg_installed` — and then configure fails at generate time for anyone building IfcOpenShell as a subproject (`FetchContent` / `add_subdirectory`).

## Fix

`$<BUILD_INTERFACE:>` keeps the paths for the build tree, where they are correct, and leaves them out of the exported package, where they never belonged. The link directory gets the same treatment for the same reason.

## Verification

Configured and built IfcOpenShell as a `FetchContent` subproject against a vcpkg-provided OCCT 7.9.3 on Windows/MSVC, schemas `2x3;4;4x3_add2`: generate succeeds and `IfcParse`, `IfcGeom` and `geometry_kernel_opencascade` all build. Installed-OCCT setups are unaffected — the paths are identical in the build tree either way.


Osyotr:
Shouldn't the target just be `IMPORTED` instead?

yekose:
Yes — semantically it should be `IMPORTED`; it represents an external dependency, not something this project builds.

The catch is that `IfcParse` / `IfcGeom` link it and are themselves exported, so taking it out of the export set makes `export(EXPORT …)` fail with *"requires target OpenCASCADE_INTERFACE that is not in any export set"*. The name has to exist again at consume time.

That is close to solved already: `IfcOpenShellConfig.cmake.in` calls `find_dependency(OpenCASCADE CONFIG)`, so the config would just need to recreate the imported target before including `…Targets.cmake`. And per the comment right there, for OCCT >= 7.7 the wrapper may not be needed at all — the OCCT config targets carry their own include directories, so linking them directly would work.

I kept this PR to the minimal change that unblocks configuring IfcOpenShell as a subproject with vcpkg in manifest mode, where OCCT lands inside the consumer's build tree. Happy to redo it the `IMPORTED` way if you prefer — just say which shape you want: recreate the imported target in the package config, or drop the wrapper for OCCT >= 7.7 and link the OCCT targets directly.

aothms:
I've been out of touch for too long with modern cmake. @Andrej730 @EstebanDugueperoux2?

EstebanDugueperoux2:
Hi,

It seems ok for me.

Regards.

Issue #9128, opened 2026-07-30
ifcopenshell.util.sequence applies a single-day IfcWorkTime exception to every earlier date


PahlawanJoey:
### Bug Description

ifcopenshell.util.sequence.is_day_in_work_time() evaluates the Start and Finish range of an IfcWorkTime incorrectly.

For a single-day calendar exception with:

Start  = 2026-12-25
Finish = 2026-12-25

the exception:

incorrectly applies to every date before 2026-12-25;
incorrectly does not apply on 2026-12-25 itself.

As a result, a valid future holiday may cause all preceding weekdays to be reported as non-working days. This affects scheduling helpers that rely on is_working_day(), including date offsetting and task-time recalculation.

Environment
Confirmed with IfcOpenShell 0.8.5
Python 3.11
IFC schema: IFC4
Linux
The same range implementation is also present in the source from 2026-07-26 used around the 0.8.6a260726 daily build:
https://github.com/IfcOpenShell/IfcOpenShell/blob/8deefe497cca8b9fd41e29e809ec0d0ad9478169/src/ifcopenshell-python/ifcopenshell/util/sequence.py#L203-L220
Minimal reproduction

The following script creates:

a Monday-to-Friday work calendar;
working hours from 08:00 to 16:00;
one non-recurring exception on 2026-12-25;
a minimal calendar_exception_bug.ifc test file.

Actual result

On the affected implementation:

2026-09-07: exception_applies=True,  working_day=False
2026-12-25: exception_applies=False, working_day=True
2026-12-28: exception_applies=False, working_day=True

The first regression assertion fails.

Expected result
2026-09-07: exception_applies=False, working_day=True
2026-12-25: exception_applies=True,  working_day=False
2026-12-28: exception_applies=False, working_day=True

An IfcWorkTime with Start == Finish should apply on that date only. The Start and Finish boundaries should be inclusive.

Suspected cause

The current implementation performs two independent assignments:

if start:
    result = day > start
if finish:
    result = day < finish

There are two problems:

> and < exclude the Start and Finish dates themselves.
The Finish result overwrites the earlier Start result instead of combining both range conditions.

For Start == Finish == D, this effectively returns True for dates before D and False for D itself.

The intended range check appears to be equivalent to:

return (
    (start is None or day >= start)
    and (finish is None or day <= finish)
)
Impact

This is more than an isolated boolean error:

one future ExceptionTime can mark every preceding weekday as non-working;
[...]

Issue #9134, opened 2026-07-30
sequence: recurrence Interval and Occurrences are documented but unimplemented, so any interval-based recurrence matches no days at all


BIMvoice:
## Summary

`ifcopenshell.api.sequence.assign_recurrence_pattern` documents recurrence with an interval and an occurrence count, and gives a worked example of a task recurring every 6 months. **Setting `Interval` does not make a recurrence less frequent. It makes it never occur at all.**

The matching logic in `ifcopenshell/util/sequence.py`'s `is_work_time_applicable_to_day` has six branches, one per recurrence type, and every one is shaped like this:

```python
if not recurrence.Interval and not recurrence.Occurrences:
    return <real matching logic>
return False  # TODO
```

So the moment either attribute is set, which is exactly the documented use case, the function returns `False` for every day, permanently.

## Measured

Built on a fresh IFC4 model and evaluated day by day:

| Case | Without `Interval` | With `Interval` |
|---|---|---|
| `WEEKLY`, `WeekdayComponent=[6]` (Saturday), over 14 days | **2** matches, correct | `Interval=2` ("every other Saturday") -> **0** matches |
| `MONTHLY_BY_DAY_OF_MONTH`, `DayComponent=[1]`, over 12 months | **12** matches, correct | `Interval=6` -> **0** matches |

The second is the docstring's own example: *"the maintenance task must occur every 6 months"*.

## Why this matters beyond the one function

`is_work_time_applicable_to_day` is called by `is_working_day`, which is called by `get_recent_working_day` and `get_next_working_day`, which are used by `ifcopenshell.api.sequence.cascade_schedule` and `edit_task_time` for duration and date cascading.

So a calendar using any interval-based recurrence silently contributes **no working days**, and computed schedule dates shift accordingly. Nothing raises, nothing warns.

## Never implemented rather than regressed

`git log -S "return False  # TODO"` shows the stubs were introduced in the same commit that added the function, `353281fa4d` (*"Updating a task duration now auto calculates the scheduled finish times based on the working calendar"*, May 2021), which is merged into `v0.8.0`. The `# TODO` markers are the original author's own.

So this is unfinished work rather than a regression, and the documentation describes the intent rather than something that once worked.

## Why this is filed rather than fixed

**Correcting the documentation would be the wrong fix**, because it would hide a real gap: the docs describe what the API is meant to do.

[...]

PR #9147, opened 2026-07-30
util.sequence: fix crash on valid recurrence patterns

BIMvoice:
`is_work_time_applicable_to_day` dispatches on `IfcRecurrencePattern.RecurrenceType`. Two separate problems, found by enumerating the declared `RECURRENCE_TYPE` Literal against the branches actually implemented.

### 1. Five of the eight recurrence types crashed on valid IFC

Only `RecurrenceType` is mandatory on `IfcRecurrencePattern`. `DayComponent`, `WeekdayComponent`, `MonthComponent` and `Position` are **all optional in IFC2X3, IFC4 and IFC4X3_ADD2**, so a pattern carrying none of them is perfectly valid. Those branches then did `day.weekday() + 1 in recurrence.WeekdayComponent` against `None`.

Measured on `origin/v0.8.0`, with every optional component left unset:

```
BY_DAY_COUNT               -> None
BY_WEEKDAY_COUNT           -> None
DAILY                      -> True
WEEKLY                     -> TypeError: argument of type 'NoneType' is not iterable
MONTHLY_BY_DAY_OF_MONTH    -> TypeError: argument of type 'NoneType' is not iterable
MONTHLY_BY_POSITION        -> TypeError: argument of type 'NoneType' is not iterable
YEARLY_BY_DAY_OF_MONTH     -> TypeError: argument of type 'NoneType' is not iterable
YEARLY_BY_POSITION         -> TypeError: argument of type 'NoneType' is not iterable
```

This is not an obscure path. `is_work_time_applicable_to_day` feeds `is_working_day`, which drives `count_working_days` and `offset_date`, so a valid file with a bare `WEEKLY` work calendar took the whole schedule calculation down.

Guarded with `or ()`, matching the `or []` idiom already used a few lines above in `is_working_day`. An absent component now matches nothing rather than raising. A populated one still decides the day exactly as before, which is covered by its own test.

### 2. Two recurrence types were missing from the chain entirely

`RECURRENCE_TYPE` is declared with eight members directly above the dispatch, and the chain handled six. `BY_DAY_COUNT` and `BY_WEEKDAY_COUNT` fell through to an implicit `None`, contradicting the function's `-> bool` annotation.

They now return `False` with a `# TODO`, which is the convention every other unimplemented interval case in this same function already follows. Note this is a type-contract fix rather than a behaviour change: every current caller tests truthiness, so `None` and `False` behave alike today. It is included because leaving two of eight as the only silent gap is exactly how the crash above went unnoticed.

The broader gap in this function, that `Interval` and `Occurrences` are still unimplemented for every recurrence type, is separately reported in #9134 and is not addressed here.

### Tests

`src/ifcopenshell-python/test/util/test_sequence.py`, three tests: the two enum gaps return a real bool, every one of the eight types survives absent optional components, and a populated `WeekdayComponent` still selects the right day.

Verified not vacuous, in both directions:

* Against `origin/v0.8.0`, 2 of the 3 fail (`BY_DAY_COUNT returned None`).
* Against the enum fix **alone**, without the `or ()` guards, the crash test still fails with `TypeError: argument of type 'NoneType' is not iterable`. So the two commits are independently necessary.

### AI disclosure

Per AGENTS.md: this pull request is entirely AI-generated, including the new test file, which carries the disclosure comment.


BIMvoice:
Heads-up on a conflict with our own #9106, found by `git merge-tree` rather than by guessing:

```
CONFLICT (add/add): Merge conflict in src/ifcopenshell-python/test/util/test_sequence.py
```

Both PRs create that file, which did not exist before. They touch **different functions** in `util/sequence.py` and the changes themselves do not overlap:

* **#9106** fixes `is_day_in_work_time`, where the `Finish` check overwrote the `Start` check and both boundaries were exclusive. That one also resolves #9128.
* **This PR** fixes `is_work_time_applicable_to_day`, where five of eight recurrence types crash on valid IFC and two enum members were unhandled.

So the resolution is mechanical: the file ends up with both test classes, `TestIsDayInWorkTime` from #9106 and `TestIsWorkTimeApplicableToDay` from here. Nothing needs reconciling inside either class.

**Suggest merging #9106 first**, since it is older, self-contained and closes a reported issue. I have deliberately not rebased this branch on top of it, because that would mean force-pushing a branch that is already under review.

Issue #9121, opened 2026-07-30
alignment: transition point referents (P.C., P.T., P.V.C.) are no longer created, and the test that should catch it always passes


BIMvoice:
## Summary

`ifcopenshell.api.alignment` advertises automatic transition-point referents, and the implementation is still present and correct, but **nothing calls it any more**. Building an alignment with any number of curves produces no `P.C.`, `P.T.`, `P.V.C.` or `P.V.T.` referents at all.

This looks like accidental collateral damage from a bug-fix refactor rather than an intentional removal, which is why this is a question rather than a patch.

## What is advertised

`api/alignment/__init__.py:39`, in the module docstring:

> 5. Automatic definition of alignment transition point referents

And `_get_segment_start_point_label.py:26-28`:

> Each layout segment is assigned a position referent that informs about the start point of the segment. An example is the point of curvature of a horizontal circular curve... named with an indicator of the position and the station, e.g. `"P.C. (145+98.32)"`

## What actually happens

Building a horizontal and vertical alignment via `create_by_pi_method` with 8 horizontal segments (3 line-to-arc transitions) and 10 vertical segments (4 transitions):

```
Total IfcReferent entities in file: 1
  #28 Name='0+00.00' PredefinedType='STATION'
```

Only the single start-of-alignment stationing referent from `create.py` exists. Zero referents at any of the seven real transition points.

## The implementation is intact, just unreachable

`_get_segment_start_point_label.py` still contains the full 9x9 horizontal lookup table (`P.C.`, `P.T.`, `S.C.`, `C.S.`, `P.C.C.` and so on), the 4x4 vertical table, the cant table, and `register_referent_name_callback()` for user overrides.

`grep -rn "_get_segment_start_point_label"` across the repository returns **only its own definition**. There are no callers.

Worth noting separately: **`register_referent_name_callback` is still exported from the public API**, so a user can register a callback today that can never be invoked.

## It used to work

- PR #6817 introduced the feature in July 2025. `_add_segment_to_layout.py` called `_get_segment_start_point_label(...)` per segment and created a referent via `add_stationing_referent(...)`.
- Issue #6829 was closed by the feature's author with *"Implemented in #6817 with user defined callback functions to customize the prefix string."*
[...]

PR #9152, opened 2026-07-30
alignment: guard referents w/o ObjectPlacement

BIMvoice:
`add_positioning_referent` and `add_stationing_referent` both chain into `alignment.ObjectPlacement.RelativePlacement.Location.Coordinates` when building a fallback local placement.

`IfcAlignment.ObjectPlacement` is optional in IFC4X3_ADD2, and a **semantic-only alignment**, one with no geometric representation yet, legitimately has it unset. The module's own docstrings elsewhere acknowledge that state. Both functions then raise:

```
AttributeError: 'NoneType' object has no attribute 'RelativePlacement'
```

Guarded with a fallback to the origin, which is the idiom this module already uses for exactly this situation. `alignment/create.py:69` builds its placement with `Location=file.createIfcCartesianPoint(Coordinates=(0.0, 0.0))`, so an alignment with no placement is treated as sitting at the origin rather than crashing.

### How this was found

By enumeration rather than inspection. Every chained attribute read across `api/alignment/`, `api/cogo/` and `api/sequence/` was cross-referenced against the real schema via `ifcopenshell_wrapper.schema_by_name`, then each hit hand-checked for an outer guard and for caller reachability.

Of roughly 40 distinct sites examined, **2 were real**. About 35 were either schema-mandatory, so the attribute can never be `None`, or already protected by a guard a line-level search cannot see. A further 3 were unreachable because the module always constructs the attribute itself and it never comes from a loaded file. Those were deliberately left alone rather than "hardened", since a guard against an impossible state is noise.

### Tests

A regression test in each of `test_add_positioning_referent.py` and `test_add_stationing_referent.py`. 7 pass on this branch.

Verified not vacuous: reverting both source files while keeping the tests makes both modules fail with the exact `AttributeError: 'NoneType' object has no attribute 'RelativePlacement'`.

The full `test/api/alignment` suite passes afterwards, with one exception noted below.

### Correction to an earlier version of this description

An earlier version of this text claimed `test/api/alignment/test_update_fallback_position.py` already fails on `origin/v0.8.0`. **That was wrong.** It failed only because my local wrapper predates 52d8942, so I was measuring a stale build of my own. The observed value was the expected value with a feet-to-metres factor applied exactly one extra time (ratio `3.280839895013123`, which is `1/0.3048`), the double conversion that commit fixes. Thanks to @RickBrice for catching it.

### AI disclosure

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/alignment/_ensure_alignment_object_placement.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/alignment/_ensure_alignment_object_placement.py
@@ -1,0 +1,44 @@
+# IfcOpenShell - IFC toolkit and geometry engine
+# Copyright (C) 2025 Thomas Krijnen <thomas@aecgeeks.com>
+#
+# This file is part of IfcOpenShell.
+#
+# IfcOpenShell is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# IfcOpenShell is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with IfcOpenShell.  If not, see <http://www.gnu.org/licenses/>.
+
+import ifcopenshell
+from ifcopenshell import entity_instance
+
+
+def _ensure_alignment_object_placement(file: ifcopenshell.file, alignment: entity_instance) -> None:
+    """
+    Ensures the alignment has an ObjectPlacement.
+
+    IfcAlignment is a subtype of IfcPositioningElement, which carries a
+    HasPlacement WHERE rule requiring ObjectPlacement to exist. An alignment
+    built by a route other than create() can reach this point with no
+    placement. This creates the same 2D origin placement create() assigns to
+    a new alignment, so the alignment satisfies its own WHERE rule and any
+    referent nested under it can share a placement that genuinely exists.
+
+    :param alignment: the alignment to check and, if needed, update
+    :return: None
+    """
+
+    if alignment.ObjectPlacement:
+        return
+
+    alignment.ObjectPlacement = file.createIfcLocalPlacement(
+        PlacementRelTo=None,
+        RelativePlacement=file.createIfcAxis2Placement2D(Location=file.createIfcCartesianPoint(Coordinates=(0.0, 0.0))),
+    )
--- a/src/ifcopenshell-python/ifcopenshell/api/alignment/add_positioning_referent.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/alignment/add_positioning_referent.py
@@ -18,6 +18,7 @@
 
 import ifcopenshell
 import ifcopenshell.api.alignment
+from ifcopenshell.api.alignment._ensure_alignment_object_placement import _ensure_alignment_object_placement
 from ifcopenshell.api.alignment.update_fallback_position import update_fallback_position
 import ifcopenshell.api.pset
 import ifcopenshell.guid
@@ -70,11 +71,8 @@ def add_positioning_referent(
 
         update_fallback_position(file, object_placement)
     else:
-        coordinates = (
-            alignment.ObjectPlacement.RelativePlacement.Location.Coordinates
-            if alignment.ObjectPlacement
-            else (0.0, 0.0)
-        )
+        _ensure_alignment_object_placement(file, alignment)
+        coordinates = alignment.ObjectPlacement.RelativePlacement.Location.Coordinates
         object_placement = file.createIfcLocalPlacement(
             PlacementRelTo=None,
             RelativePlacement=file.createIfcAxis2Placement2D(Location=file.createIfcCartesianPoint(coordinates)),
--- a/src/ifcopenshell-python/ifcopenshell/api/alignment/add_positioning_referent.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/alignment/add_positioning_referent.py
@@ -70,11 +70,14 @@ def add_positioning_referent(
 
         update_fallback_position(file, object_placement)
     else:
+        coordinates = (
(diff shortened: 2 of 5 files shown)


RickBrice:
@BIMvoice I see where you are going with this and I am mostly in agreement.

Some questions:

My original idea was that since the referents were required to have an ObjectPlacement.RelativePlacement, they should have the same relative placement as the alignment. This is why I force IfcAlignment to have one, when not strictly required. What are the ramifications of breaking this paradigm? As editing of an alignment progresses, will the relative placement of the alignment and its related referents drift apart?

I am not 100% sure if the coordinates need to be 2D (0,0) or 3D (0,0,0). Have you confirmed the resulting IfcReferent.ObjectPlacement.RelativePlacement.Location passes the bSI Validation Service?

RickBrice:
@BIMvoice 

> test/api/alignment/test_update_fallback_position.py already fails on origin/v0.8.0

Is this an outstanding issue? When I run the unit tests locally, test_update_fallback_position.py passes. Please clarify. Are you testing against the latest build of the C++ core? That is where the root cause issue was resolved per 52d8942

BIMvoice:
@RickBrice thank you, and you are right on the test question. Taking your points in reverse order, since the last one is a correction I owe you.

## The `test_update_fallback_position.py` claim was wrong, and it was my environment

You are correct and I withdraw it. My local wrapper predates 52d8942, so I was measuring my own stale build and reporting it as a repo state. I should have checked that before putting it in the PR description.

The evidence is unambiguous now that I looked properly:

```
expected: (3781.0625905626425, 2663.2859940077856, 0.0)
observed: (12405.060992659588, 8737.814940970424,  0.0)

ratio x = 3.2808398950131235
ratio y = 3.280839895013123
1/0.3048 = 3.280839895013123
```

The observed value is the expected value with a feet-to-metres factor applied exactly one extra time, which is the double conversion 52d8942 fixes. Nothing to do with this PR, and not an outstanding repo issue. I will remove that paragraph from the PR description.

## The 2D versus 3D question

Partially answerable, and I want to be clear about which part I have not answered.

What I can show is that the 2D form is what this module already does for the alignment itself. `alignment/create.py` builds the alignment's own placement as:

```python
ObjectPlacement=file.createIfcLocalPlacement(
    PlacementRelTo=None,
    RelativePlacement=file.createIfcAxis2Placement2D(
        Location=file.createIfcCartesianPoint(Coordinates=(0.0, 0.0))
    ),
)
```

So the `(0.0, 0.0)` fallback in this PR is consistent with the placement the alignment is given at creation, which is the value the referent would have copied anyway in the case the guard covers.

What I have **not** done is run the result through the bSI Validation Service. I do not have it wired up here, so I cannot answer your question directly and I would rather say that than imply otherwise. If you have it available, that check is worth more than my inspection.

## The design question, where I think you are right and this PR is at the wrong layer

Your paradigm is that referents share the alignment's relative placement, and you force `IfcAlignment` to have one precisely so that holds. This PR quietly substitutes the origin when it does not, which is exactly the drift you are asking about: the referent stops tracking the alignment and nothing reports it.

So I think the guard is treating a symptom. If the invariant is "an alignment always has an `ObjectPlacement`", then an alignment without one is a violated invariant, and the honest responses are to restore it or to fail loudly, not to let the referent diverge silently.

That said, I did reach this from a real crash, so something reaches these functions with no placement:

```
AttributeError: 'NoneType' object has no attribute 'RelativePlacement'
```

`IfcAlignment.ObjectPlacement` is optional in IFC4X3_ADD2, so an alignment built by any route other than `create.py` can lack one.

Three options, and I would rather you pick than guess:

1. **Enforce the invariant.** Have the referent functions ensure the alignment has a placement, creating the same 2D origin placement `create.py` would have, and attach the referent to that. Preserves your paradigm, and the alignment and referents cannot drift because they genuinely share one placement.
2. **Fail loudly.** Raise with a clear message when the alignment has no placement, so the caller fixes the alignment rather than getting a silently misplaced referent.
3. **What this PR currently does.** Fall back to the origin. Simplest, but it is the option that permits the drift you described.

My preference is 1, since it keeps your design intent and removes the crash for the same reason rather than in spite of it. I am happy to rework the PR that way, or to close it if you would rather fix this in the alignment creation path yourself.

One further thing worth flagging separately, since it is in the same area and I have not touched it: `_get_segment_start_point_label.py` is fully implemented but has no callers, and `alignment/__init__.py` advertises "automatic definition of alignment transition point referents" which does not appear to happen. That is issue #9121 rather than anything for this PR.

RickBrice:
@BIMvoice please further consider that IfcReferent must have a placement (https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcPositioningElement.htm#5.4.3.42.4-Formal-propositions) and is flagged as an error by the bSI Validation Service if it does not.

This precludes the assertion:

> IfcAlignment.ObjectPlacement is optional in IFC4X3_ADD2, and a semantic-only alignment, one with no geometric representation yet, legitimately has it unset

if the placement of IfcReferent is to be relative to the alignment to which they are semantically informing on the location

BIMvoice:
You are right, and I want to correct the record rather than concede quietly.

I checked the WHERE rule directly in this repo's own IFC4X3 schema, `IfcPositioningElement.HasPlacement`, `exists(SELF.ObjectPlacement)`, and ran it with `express_rules=True` against a bare alignment with no placement. It fires:

```
ERROR Rule IfcPositioningElement.HasPlacement:
    exists(SELF.ObjectPlacement)
Violated by: False
```

It is worse for my original claim than you put it. **`IfcAlignment` is itself a subtype of `IfcPositioningElement`** (`IfcAlignment -> IfcLinearPositioningElement -> IfcPositioningElement`), so `HasPlacement` applies to the alignment directly, not only to referents nested under it. The violation is reported on the `IfcAlignment` instance before any referent exists. My assertion that a semantic-only alignment legitimately has `ObjectPlacement` unset was wrong on its own terms.

I have reworked the PR to option 1. When the alignment has no placement, the referent functions now give it the same 2D origin placement `create()` would have assigned, and only then derive the referent's coordinates from it. Both `add_positioning_referent` and `add_stationing_referent` go through a shared helper, so the alignment and its referents genuinely share one placement rather than silently diverging.

Tests now assert the alignment's placement is no longer missing and run `validate(express_rules=True)` to confirm `HasPlacement` is satisfied on both the alignment and the referent. 33 of 33 alignment tests pass.

One thing I should be explicit about: I have not run this through the bSI Validation Service, since it is not wired up here. The result above is this repo's local `express_rules=True` evaluation.

Thank you for pushing back twice on this. You were right both times.

PR #9158, opened 2026-07-30
ifcbimtester: fix eight crashes in step definitions

BIMvoice:
Eight crashes in BIMTester's step definitions, found while sweeping sibling packages for unguarded optional attributes.

**Please read the note at the bottom about this package's status before spending review time on it.**

### The one that is not an optional-attribute bug

`bimtester/util.py`, `assert_pset`:

```python
def assert_pset(element, pset_name, prop_name=None, value=None):
    ...
    psets = ifcopenshell.util.element.get_psets(site)
```

`site` is not defined anywhere in that function. The parameter is `element`. **This raises `NameError` on every single call**, not on some inputs. It affects every IFC2X3 geolocation step and any custom feature file using pset assertions.

### The other seven, all optional attributes dereferenced unguarded

- **`geolocation/en.py`**: `xaxis2angle(None, None)` when `IfcMapConversion` has no `XAxisAbscissa` or `XAxisOrdinate`. Both optional.
- **`model_federation/en.py`**: the same optionality reaching `float(None)` on the way into `xyz2enh`. Defaulted to the values that function already documents.
- **`application/en.py`**: `by_type("IfcApplication")[0]` on a file with no application recorded. `OwnerHistory` is optional throughout IFC4.
- **`attributes_qsets/en.py`**: assumed every `IsDefinedBy` relationship is an `IfcRelDefinesByProperties`. Under IFC2X3 it can also be `IfcRelDefinesByType`, which has no `RelatingPropertyDefinition`, so any typed element crashed the step.
- **`attributes_psets/en.py`**: assumed `get_psets()` always returns strings, following a stale comment. A boolean or numeric property crashed `re.search`.
- **`attributes_eleclasses/en.py`**: `re.search(pattern, element.Name)` on an unnamed element.
- **`aggregation/en.py`**, `check_starts_with`: used `hasattr`, which confirms an attribute exists but says nothing about whether it is `None`.

### Verified

All eight reproduced against hand-built IFC files before the change and re-verified after, with happy-path checks alongside.

### Status of this package, stated plainly

`src/ifcbimtester` appears **dormant**. It is not in `ci.yml`, it is not referenced from any workflow, packaging or build configuration, and every commit touching it for a long time has been mechanical: `black .`, import sorting, a Ruff fix, a BlenderBIM to Bonsai rename.

The `assert_pset` defect is itself evidence of that. A function that raises `NameError` on **every** call cannot have been executed in a long time.

So this may not be worth your review time, and I would rather say so than quietly add eight commits to a queue. Three ways to read it, all reasonable:

1. Merge it, since the fixes are small, verified and independent.
2. Ignore it, if the package is on its way out.
3. Treat the `NameError` as a signal that BIMTester needs either CI coverage or a deprecation notice, which is a bigger decision than this PR.

If you want only the `NameError` fix and not the other seven, say so and I will reduce it to that single commit.

### No tests

The package has no test infrastructure and this adds none, consistent with the precedent for untested packages here. Verification was by direct execution, described above.

### AI disclosure

### DIFF
--- a/src/ifcbimtester/bimtester/features/steps/aggregation/en.py
+++ b/src/ifcbimtester/bimtester/features/steps/aggregation/en.py
@@ -189,7 +189,7 @@ def check_if_element_exists_by_types_with_attribute_name(ifc_types, attribute_na
 
 def check_starts_with(attribute_name, value, elements):
     """Make sure at least an element of the list has a attribute starting with the value"""
-    if any(hasattr(x, attribute_name) and getattr(x, attribute_name).startswith(value) for x in elements):
+    if any(getattr(x, attribute_name, None) and getattr(x, attribute_name).startswith(value) for x in elements):
         return True
     return False
 
--- a/src/ifcbimtester/bimtester/features/steps/application/en.py
+++ b/src/ifcbimtester/bimtester/features/steps/application/en.py
@@ -24,7 +24,9 @@ from bimtester.lang import _
 @step('The IFC file must be exported by application full name "{fullname}"')
 def step_impl(context, fullname):
 
-    real_fullname = IfcStore.file.by_type("IfcApplication")[0].ApplicationFullName
+    applications = IfcStore.file.by_type("IfcApplication")
+    assert applications, _("The IFC file does not have any exporting application recorded")
+    real_fullname = applications[0].ApplicationFullName
     assert real_fullname == fullname, (
         "The IFC file was not exported by application full name {} "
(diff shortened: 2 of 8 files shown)


Moult:

Thanks Petru. The eight guards are sound, but the maintainers have decided to retire ifcbimtester rather than keep patching it: it has had no feature work since 2022, IfcTester absorbed its purpose, and the Bonsai integration went in 2024. #9541 removes the package, so closing this rather than merging into code that is about to be deleted.

PR #9163, opened 2026-07-30
docs: fix ten API examples that cannot run

BIMvoice:
Ten `ifcopenshell.api` docstring examples could not execute at all. The functions are correct; the published examples are broken, and they ship as documentation through sphinx-autoapi.

| Function | Defect in the example |
|---|---|
| `unit.edit_named_unit` | typo: `attibutes` |
| `sequence.unassign_product` | missing the `model` argument |
| `sequence.assign_work_plan` | missing the `model` argument |
| `sequence.unassign_recurrence_pattern` | missing the `model` argument |
| `cost.copy_cost_schedule` | missing the `model` argument |
| `system.unassign_flow_control` | wrong keyword arguments |
| `cost.remove_cost_item_quantity` | calls a different function entirely |
| `material.remove_list_item` | `IfcMaterialMaterialList` does not exist in any schema |
| `nest.unassign_object` | `IfcTasks` is not an entity |
| `unit.add_derived_unit` | uses `LINEARVELOCITY` where the same docstring elsewhere correctly says `LINEARVELOCITYUNIT` |

Each fix was verified by re-running the example before and after.

### Two of these fixes exposed further real gaps, reported not fixed

Once the examples stopped crashing they reached genuine defects underneath:

- **`system.unassign_flow_control`** shares a root cause with `system.assign_flow_control`, which is already flagged separately.
- **`nest.unassign_object`** shows that `root.create_entity(ifc_class="IfcTask")` leaves the mandatory `IsMilestone` unset.

Neither is touched here, since both are API code changes rather than documentation fixes.

### Context: what the wider sweep established

These came from executing all 271 `ifcopenshell.api` docstring examples against fresh models on three schemas and validating the results.

**IFC4X3_ADD2 is identical to IFC4**, with the same set of failing and invalid functions, proven by direct set comparison rather than inferred. So there is no IFC4X3-specific invalidity to chase.

**IFC2X3 is materially worse** (34 invalid against 16) and two real causes explain most of it. `IfcProject.RepresentationContexts` and `UnitsInContext` are **mandatory in IFC2X3 but optional in IFC4**, and `IfcCostSchedule.ID` is mandatory in IFC2X3 while `cost.add_cost_schedule` never sets it, unlike `owner/add_person.py` which already carries the `if schema == "IFC2X3"` pattern. That second one is a genuine API defect and is flagged for follow-up, not fixed here.

### A defect class nothing had checked before

Plain `validate()` does not evaluate EXPRESS WHERE rules; its own docstring says so. Running with `express_rules=True` surfaced a class no previous pass could see: **`exists(SELF.Name)`**. `Name` is optional at the attribute level but a WHERE rule makes it mandatory for several types, including `IfcTask`, `IfcWindowType`, `IfcFurnitureType` and `IfcWallType`, so any example calling `root.create_entity` without `name=` violates the schema in a way plain validation never reports.

Also worth flagging separately: **the WHERE-rule executor segfaults intermittently and non-deterministically**, crashing and then succeeding on a retry with identical input. That is a robustness problem in its own right and was not chased here.

### A correction to the earlier sweep

The harness leaked process-global state. `owner.create_owner_history`'s example legitimately monkeypatches `ifcopenshell.api.owner.settings`, and nothing reset it between examples, so the patch leaked into every alphabetically later one. Calling `factory_reset()` before each example drops the IFC4 invalid count from 15 to 14 and removes `owner.edit_application`, which was therefore **a false positive of the harness, not a defect**. Confirmed independently.

### AI disclosure

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/cost/copy_cost_schedule.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/cost/copy_cost_schedule.py
@@ -34,7 +34,7 @@ def copy_cost_schedule(
     .. code:: python
 
         schedule = ifcopenshell.api.cost.add_cost_schedule(model)
-        new_schedule = ifcopenshell.api.cost.copy_cost_schedule(schedule)
+        new_schedule = ifcopenshell.api.cost.copy_cost_schedule(model, schedule)
     """
     # Shared code logic with copy_work_schedule.
     new_schedule = ifcopenshell.util.element.copy(file, cost_schedule)
--- a/src/ifcopenshell-python/ifcopenshell/api/cost/remove_cost_item_quantity.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/cost/remove_cost_item_quantity.py
@@ -40,7 +40,7 @@ def remove_cost_item_quantity(
         quantity = ifcopenshell.api.cost.add_cost_item_quantity(model,
             cost_item=item, ifc_class="IfcQuantityVolume")
         # Let's change our mind and delete it
-        ifcopenshell.api.cost.remove_cost_item(model,
+        ifcopenshell.api.cost.remove_cost_item_quantity(model,
             cost_item=item, physical_quantity=quantity)
     """
     if file.get_total_inverses(physical_quantity) == 1:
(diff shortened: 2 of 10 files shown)


Moult:

Thanks Petru. Verified all ten examples run against v0.9.0. Retargeted to v0.9.0 and pushed to your branch: your two commits squashed into one, plus one extra line, the STEP comment under the LINEARVELOCITYUNIT example, which still read `.LINEARVELOCITY.`. Merging once CI is green.

PR #9167, opened 2026-07-30
sequence: name tasks so they satisfy the WHERE rule

BIMvoice:
Every task Bonsai creates is schema-invalid, in a way ordinary validation cannot see.

`IfcTask.Name` is **optional at the EXPRESS attribute level** but carries a WHERE rule making it mandatory:

```
IfcTask.Name attribute-level optional = True
express rule: exists(SELF.Name)  ->  Violated by: False
```

`ifcopenshell.validate.validate(model, logger)` does **not** evaluate WHERE rules; its own docstring says so. Only `express_rules=True` catches this, which is why it went unnoticed.

### It is reachable through normal use, not just in an example

Bonsai calls the api with no name argument at all:

```
src/bonsai/bonsai/core/sequence.py:155   ifc.run("sequence.add_task", work_schedule=work_schedule)
src/bonsai/bonsai/core/sequence.py:163   ifc.run("sequence.add_task", parent_task=parent_task)
```

Those back "Add Summary Task" and "Add Task". So **every task a user adds through the interface** gets `Name = None` and violates the rule. Bonsai authors IFC natively, so this is written straight into the user's model.

The fix defaults `name` to `"Unnamed"`, which is this codebase's established placeholder: `or "Unnamed"` appears in roughly 186 places across Bonsai, and `document/add_information.py` already creates entities with `"Name": "Unnamed"` for the same reason.

An explicit name is still honoured, which is covered by its own test.

### Tests

`test/api/sequence/test_add_task.py`, covering both the default and the override, run for IFC4 and IFC4X3.

### Scope note

This fixes the one case proven reachable from Bonsai's UI. The same WHERE rule applies to other types, including `IfcWindowType`, `IfcFurnitureType` and `IfcWallType`, and those have not been audited for reachability here. Flagging that rather than quietly implying the class is fully addressed.

Also worth recording separately: **the WHERE-rule executor segfaults intermittently and non-deterministically**, crashing then succeeding on retry with identical input. That is a robustness problem in its own right and is not addressed here.

### AI disclosure

Per AGENTS.md: this pull request is entirely AI-generated, including the new test file, which carries the disclosure comment.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/sequence/add_task.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/sequence/add_task.py
@@ -28,7 +28,7 @@ def add_task(
     file: ifcopenshell.file,
     work_schedule: Optional[ifcopenshell.entity_instance] = None,
     parent_task: Optional[ifcopenshell.entity_instance] = None,
-    name: Optional[str] = None,
+    name: str = "Unnamed",
     description: Optional[str] = None,
     identification: Optional[str] = None,
     predefined_type: str = "NOTDEFINED",
@@ -72,7 +72,9 @@ def add_task(
     :param parent_task: The parent task, if the task is to be a subtask or
         child task. This is mutually exclusive with the work_schedule
         parameter.
-    :param name: The name of the task.
+    :param name: The name of the task. IfcTask requires a Name in IFC4 and
+        later, so an unset name defaults to "Unnamed" rather than staying
+        blank.
     :param description: The description of the task.
     :param identification: The identification code of the task.
     :param predefined_type: The predefined type of the task. Common ones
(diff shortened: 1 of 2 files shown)


BIMvoice:
### Correction to a caveat in this description

This PR's description notes that the EXPRESS WHERE-rule executor "segfaults intermittently and non-deterministically on identical input". **That has not held up and should be treated as unconfirmed.**

Two later sweeps exercised the executor heavily and neither reproduced it:

* one ran **1230 checks** across IFC2X3, IFC4 and IFC4X3_ADD2, each in its own subprocess with retry-on-crash logic: **0 crashes, 0 retries**
* another ran **8 checks** specifically against this `add_task` path: **0 crashes**, all exit code 0 with consistent output

So over 1200 runs across two independent passes produced no segfault. The original observation came from a different harness and may have been environmental rather than a property of the executor.

Nothing else in this PR depends on that caveat: the `IfcTask.HasName` violation itself was reproduced directly and repeatedly, and the fix is unaffected.

PR #9177, opened 2026-07-31
geometry.unassign_representation: fix None crash

BIMvoice:
`unassign_representation` crashes on a product that has no representation yet.

```python
representations = list(product.Representation.Representations or [])
```

`IfcProduct.Representation` is optional in every schema, so a freshly created element, for example a wall before any geometry is added, gives:

```
AttributeError: 'NoneType' object has no attribute 'Representations'
```

The function already treats "this representation is not assigned to this product" as a **no-op**, returning quietly a few lines below. A product with no representation at all is the same situation and should behave the same way, so the fix returns early rather than raising.

The change also hoists `product.Representation` into the local `product_def` that the function already declared further down, removing a duplicated attribute read.

Reproduced live in the built interpreter before fixing. Regression test added and verified red before the change and green after; the file's existing suite passes 8 of 8.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/geometry/unassign_representation.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/geometry/unassign_representation.py
@@ -46,18 +46,20 @@ class Usecase:
     def unassign_product_representation(
         self, product: ifcopenshell.entity_instance, representation: ifcopenshell.entity_instance
     ) -> None:
-        representations = list(product.Representation.Representations or [])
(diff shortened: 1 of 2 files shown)


Moult:

Thanks Petru. Retargeted to v0.9.0 and pushed to your branch: same fix and test, with the early-return guard folded into a walrus (`if not (product_def := product.Representation): return`). Rebased onto current v0.9.0 as a single commit. Merging once CI is green.

PR #9179, opened 2026-07-31
express: fail clean when rules are missing

BIMvoice:
The fallback that looks for a schema's `.exp` file in the current directory did not check whether it found one:

```python
schema_path = current_dir_files.get(schema_name + ".exp")
fn = schema_path[:-4] + ".py"
```

On a miss that gives `TypeError: 'NoneType' object is not subscriptable`, which tells the user nothing about what went wrong.

`validate.py` and `entity_instance.py` already guard this identical pattern. The fix follows them and raises `FileNotFoundError` naming both the schema and the directory that was searched.

Reachable via `python -m ifcopenshell.express.rule_executor`, or `validate(f, logger, express_rules=True)` against a schema whose `.exp` is not in the working directory.

The fix also restores `ifcopenshell.settings.unpack_non_aggregate_inverses` before raising, which the early exit would otherwise skip, leaving a global modified for the rest of the process.

Regression test added, verified red before the fix and green after.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/express/rule_executor.py
+++ b/src/ifcopenshell-python/ifcopenshell/express/rule_executor.py
@@ -101,6 +101,12 @@ def run(f: ifcopenshell.file, logger: Logger) -> None:
         current_dir_files = {fn.lower(): fn for fn in os.listdir(".")}
         schema_name = str(f.schema_identifier).split(" ")[-1].lower()
         schema_path = current_dir_files.get(schema_name + ".exp")
+        if schema_path is None:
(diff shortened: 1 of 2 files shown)


aothms:
> Reachable via `python -m ifcopenshell.express.rule_executor`, or `validate(f, logger, express_rules=True)` against a schema whose `.exp` is not in the working directory.

Only when

```
    fn = os.path.join(os.path.dirname(__file__), "rules", f"{f.schema_identifier}.py")
    try:
        source = open(fn, "r").read()
```

fails which is very atypical. Are you sure we need to guard for this?

PR #9180, opened 2026-07-31
file: raise AttributeError for bad rocksdb attr

BIMvoice:
`rocksdb_lazy_instance.__getattr__` unpacked its lookup result as a 2-tuple without checking for a miss:

```python
attr = attribute_lookup(self.storage.schema_identifier, self.is_a()).get(name)
if isinstance(attr, int):
    return self[attr]
else:
    entity_indices, attribute_index = attr      # attr can be None
```

Any unknown attribute name on a rocksdb-backed instance therefore gave:

```
TypeError: cannot unpack non-iterable NoneType object
```

**This breaks every `getattr(inst, name, default)` and `hasattr()` caller**, since both rely on `AttributeError` to fall back and neither catches `TypeError`. Reachable through the public `file.storage` API.

The fix raises `AttributeError` with the same wording `entity_instance` already uses for this case, so the two backends behave alike.

Reproduced against a real rocksdb conversion of `test/files/basic.ifc`. Regression test added, verified red before the fix and green after.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/file.py
+++ b/src/ifcopenshell-python/ifcopenshell/file.py
@@ -353,6 +353,8 @@ class rocksdb_lazy_instance:
         attr = attribute_lookup(self.storage.schema_identifier, self.is_a()).get(name)
         if isinstance(attr, int):
             return self[attr]
+        elif attr is None:
+            raise AttributeError("entity instance of type '%s' has no attribute '%s'" % (self.is_a(), name))
         else:
             entity_indices, attribute_index = attr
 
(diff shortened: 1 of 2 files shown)


aothms:
Just a note that this comment is wrong:

> The regression test asked the rocksdb storage for id 1, which in
> ColumnPSetsOfSets.ifc is an IfcOrganization, a non-rooted entity. On v0.9.0
> that is not reachable through storage.by_id and the test failed in its setup
> with KeyError before it could assert anything.

`by_id()` does the numeric id lookup i,e 1 vs `by_guid()` should do a rooted-entity first attribute lookup (globalid)

PR #9181, opened 2026-07-31
bcf: fix v3 crash on missing extensions.xml

BIMvoice:
`BcfXml.extensions` reads `extensions.xml` unconditionally:

```python
if not self._extensions and self._zip_file:
    self._extensions = self._xml_handler.parse(self._zip_file.read("extensions.xml"), mdl.Extensions)
```

That file is an **optional** member of a BCF v3 archive, so any valid BCF that omits it raises:

```
KeyError: "There is no item named 'extensions.xml' in the archive"
```

This file already handles optional members correctly two properties above, and the fix uses that exact pattern:

```python
# project_info, existing code
if not self._project_info and self._zip_file and zipfile.Path(self._zip_file, "project.bcfp").exists():
```

`DocumentsHandler.load` in `bcf/v3/document.py` guards `documents.xml` the same way. The property's return type is **already** `Optional[mdl.Extensions]`, so returning `None` is the declared contract rather than new behaviour.

Reachable from anything that touches `.extensions` on a loaded `BcfXml`, including the class's own `__eq__`.

Verified red then green directly: stashed the fix, reproduced the `KeyError`, restored it, confirmed the property returns `None`. Regression test `test_extensions_missing_from_file` added; the full `src/bcf` suite passes 26.

### Audit context

Found while sweeping `ifc5d`, `ifccsv`, `ifcclash`, `ifcdiff` and `bcf`. **8 files were skipped because one of our own open PRs already touches them** (#9116, #9143, #9113, #9120), two of which surfaced only by diffing branch contents rather than through `gh pr list`, since a pushed branch with no PR is invisible to it.

### DIFF
--- a/src/bcf/bcf/v3/bcfxml.py
+++ b/src/bcf/bcf/v3/bcfxml.py
@@ -82,7 +82,7 @@ class BcfXml:
     @property
     def extensions(self) -> Optional[mdl.Extensions]:
         """BCF extensions."""
-        if not self._extensions and self._zip_file:
+        if not self._extensions and self._zip_file and zipfile.Path(self._zip_file, "extensions.xml").exists():
             self._extensions = self._xml_handler.parse(self._zip_file.read("extensions.xml"), mdl.Extensions)
         return self._extensions
 
--- a/src/bcf/tests/v3/test_bcf_xml.py
+++ b/src/bcf/tests/v3/test_bcf_xml.py
@@ -1,6 +1,7 @@
 """BCF XML tests."""
 
 import uuid
+import zipfile
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


Moult:
What makes you think it's optional? Reading https://github.com/buildingSMART/BCF-XML/tree/release_3_0/Documentation says:

> ...
> The root of the BCF zip contains the following files.
> 
>     extensions.xml <-- Note: no mention of (optional) like below
>         An XML file defining the extensions of a project. The schema for this file is extensions.xsd.
>     project.bcfp (optional)
>         An XML file defining the details of a project. The schema for this file is project.xsd.
> ...

Sounds like invalid BCF which we don't support.

PR #9203, opened 2026-07-31
ifctester: PartOf false pass on USERDEFINED predefinedType

BIMvoice:
`PartOf` never recognises `predefinedType="USERDEFINED"`, in any of its six relation branches. The result is a **false pass**: a model that violates a prohibition is reported as compliant.

### The cause

`ifcopenshell.util.element.get_predefined_type()` never returns the literal string `"USERDEFINED"`. Its own docstring says so: it substitutes the custom `ObjectType` / `ElementType` / `ProcessType` text instead.

`Entity.__call__` already special-cases this. `PartOf.__call__` does not, at any of its six sites: aggregates, group, contained-in-spatial-structure, nests, voids and fills, and the no-relation ancestor walk.

### The wrong verdict, reproduced

IFC4, an `IfcSpace` containing an `IfcWall`, checked with `PartOf(predefinedType="USERDEFINED", cardinality="prohibited")`:

* **Before: PASS**, `is_pass=True`, reason `PROHIBITED`.
* **Correct: FAIL.** The wall genuinely sits in a userdefined-type space, which the prohibition forbids.

The mirror case, `cardinality="required"`, gives the opposite wrong answer: FAIL where it should PASS.

All six branches were built and run independently, before and after.

### Why this matters more than a crash

A false pass in an IDS checker tells someone their model is compliant when it is not, and nothing surfaces the mistake later. That is why this was prioritised over the false-fail cases in the same audit.

### Fix and tests

Adds `PartOf.predefined_type_matches()`, mirroring the pattern `Entity.__call__` already uses, applied at all six call sites.

`37/37` pass, both at baseline and after. Two previously untested `PartOf` branches gained coverage as a byproduct.

### Deliberately not changed

* `Specification.minOccurs` / `maxOccurs` as exact numeric bounds. This reads as ternary-by-design and no fixture uses other values, so changing it would be a guess about intent rather than a fix.
* The remaining wrong-verdict site in the reporter layer is exactly what #9190 already fixes.


aothms:
For all ids fixes please make sure there is a minimal .ids .ifc pair that
demonstrates the defect fixed by the pr.

Op za 1 aug 2026, 01:11 schreef Petru Conduraru ***@***.***>:

BIMvoice:
Done, and applied to the other five IDS PRs as well (#9205, #9190, #9189, #9188, #9187).

`src/ifctester/test/fixtures/partof_userdefined/` holds the `.ids` and `.ifc` pair, wired into `test_fixtures.py` which goes through `ids.open()`, `ifcopenshell.open()` and `Ids.validate()`, so it exercises the same path as the CLI rather than calling a facet directly.

Each pair was checked in both directions, since a fixture that passes before and after proves nothing:

* Unfixed (`a11ebdf8c4`): `[PASS] (1/1)`
* Fixed: `[FAIL] (0/1)`

There were no `.ids`/`.ifc` fixtures in `src/ifctester` before this, so the layout is new. I followed `src/ifcopenshell-python/test/fixtures/`, which `test_validate.py` and `test_mvd_info.py` already use. Happy to move it if you would rather it sat elsewhere.

Issue #353, opened 2018-03-13
Broken express_parser.py


frozar:
I'm really new to this project and try to do my best to discover it.

I wonder if it is possible to generate the code to parse the lastest [IFC4_ADD2.exp](http://www.buildingsmart-tech.org/ifc/IFC4/Add2/IFC4_ADD2.exp). In the file "src/ifcexpressparser/README.txt", it is written: 
https://github.com/IfcOpenShell/IfcOpenShell/blob/e87a1a0b32cacbbecb7b3cc428bdb2d3f362e148/src/ifcexpressparser/README.txt#L1-L9

So I download the `IFC4_ADD2.exp` file and naively try the command:
```bash
python bootstrap.py express.bnf > express_parser.py && python express_parser.py IFC4_ADD2.exp
```
 
As an output, I get an error which say:
```
Traceback (most recent call last):
  File "express_parser.py", line 468, in <module>
    ast = syntax.parseFile(sys.argv[1])
[...]
  File "/usr/lib/python2.7/dist-packages/pyparsing.py", line 779, in wrapper
    ret = func(*args[limit[0]:])
TypeError: <lambda>() takes exactly 1 argument (0 given)
```

Is it possible to generate the code which would be able to handle IFC4X2 files from `IFC4_ADD2.exp`? Or do you process it in an another way?

aothms:
That should totally work. What version of pyparsing are you running. I am on the latest 2.2.0 and am not experiencing issues.

frozar:
Hi,

First of all, I just notice that I used python2.7 when I wrote this issue, so sorry for that.

By the way, I try with python3:
```bash
python3 bootstrap.py express.bnf > express_parser.py && python3 express_parser.py IFC4_ADD2.exp
```

I get the following output:
```
Traceback (most recent call last):
  File "express_parser.py", line 460, in <module>
    import header
  File "/home/frozar/wk/IfcOpenShell/src/ifcexpressparser/header.py", line 24, in <module>
    import documentation
  File "/home/frozar/wk/IfcOpenShell/src/ifcexpressparser/documentation.py", line 51, in <module>
    for oid, name, desc in csv.reader(f, delimiter=';', quotechar='"'):
  File "/usr/lib/python3.5/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 6602: invalid start byte
```

[...]

PR #9205, opened 2026-07-31
ifctester: Attribute optional cardinality and LOGICAL UNKNOWN give wrong verdicts

BIMvoice:
Two wrong-verdict defects in `Attribute.__call__`, both demonstrated by running the checker rather than by reading it.

### 1. `optional` cardinality behaved exactly like `required`

The `optional` short-circuit only fired when the attribute name did not resolve at all, leaving `values = []`. For a valid but merely **unset** forward attribute, `values` is `[None]`, which is a non-empty list, so `bool(values)` was `True` and the short-circuit never ran.

`IfcWall.Description` left unset, with `cardinality="optional"`:

* **Before: FAIL**, reason `FALSEY`.
* **Correct: PASS.** Having nothing to check on an absent attribute is the entire point of `optional`.

This is the commonly hit one. Any IDS specification using `optional` on an `Attribute` requirement reported false non-compliance whenever the attribute was genuinely unset.

### 2. An `IFCLOGICAL` set to `UNKNOWN` was treated as absent

Same shape as #8161 by @gernotluidolt, which fixes the sibling `Property` facet and is still open and unmerged. `Attribute` has the bug too, and this PR fixes that side.

`IfcPresentationLayerWithStyle.LayerOn = "UNKNOWN"`, which is a genuinely present value:

* `cardinality="required"` — **before: FAIL** (`FALSEY`), correct is **PASS**.
* `cardinality="prohibited"` — **before: PASS**, correct is **FAIL**.

The prohibited case is the severe one: a present value reported as absent means a prohibition is silently satisfied, so a non-compliant model passes.

### Checked and clean

To be explicit about scope, these were assessed and no defect was found: `Restriction.__eq__`'s nine bound, pattern and length constraints, with inclusive versus exclusive verified; the `Classification` and `Material` facets including on IFC2X3; IFC2X3 versus IFC4 divergence in `Property.filter` and `Attribute.filter`; the contracts of `get_container`, `get_aggregate`, `get_nest`, `get_type` and `get_material` against every `facet.py` caller; and `Specification.validate`'s broadphase filter re-application, which is redundant but idempotent rather than wrong.

### Tests

Baseline `37/37` before and after, plus new cases for both defects.

---

### Note for reviewers: this changes a buildingSMART conformance test case

`ifctester/test/ids_doc_generator.py` generates the official IDS test cases from the assertions in `test_facet.py`, naming them `f"{result}-" + <description>`. The existing official case is:

```
fail-a_logical_unknown_is_considered_false_and_will_not_pass
```

This PR replaces that assertion with "A logical unknown is a real value and passes an existence check", `expected=True`, so regenerating would turn that `fail-` case into a `pass-` case.

That is exactly what buildingSMART/IDS issue **#353** has been asking for since September 2024, and we have disclosed it there and asked whether `pass-` is the intended behaviour rather than letting the change arrive as a side effect: https://github.com/buildingSMART/IDS/issues/353

**If buildingSMART confirm the current `fail-` behaviour is intended, we will revert this half of the PR** to match the conformance suite. The `optional` cardinality fix in the first commit is unaffected either way.

### DIFF
--- a/docs/dev-notes/ifctester-verdict-audit-round2.md
+++ b/docs/dev-notes/ifctester-verdict-audit-round2.md
@@ -1,0 +1,266 @@
+# ifctester wrong-verdict audit, round 2
+
+Scope: `src/ifctester/ifctester/facet.py`, `ids.py`. Goal: find defects that
+make an IDS check report the wrong pass/fail verdict, on top of round 1
+(`PartOf` `predefinedType="USERDEFINED"`, #9203) and the four fixes in
+#9190, plus #9187, #9188, #9189, #9058, #9059, #9142, #8407, #8292, #8253,
+#8161.
+
+## Existing-work check
+
+```
+git fetch origin --quiet && git fetch bimvoice --quiet
+gh pr list --repo IfcOpenShell/IfcOpenShell --state open --limit 900 \
+  --json number,title --jq '.[] | select(.title|test("ifctester|ids";"i"))'
+git for-each-ref --format='%(refname:short)' refs/remotes/bimvoice/ | grep -iE 'ifctester|ids'
+```
+
+805 open PRs total (`--limit 900` used, none skipped by the limit). 22
+ifctester/ids-related PRs found; all 11 named in the task brief confirmed
+still open and unmerged. `bimvoice` branches enumerated; the round-1 branch
+`fix/ifctester-partof-userdefined` was found and read in full (its own
+`docs/dev-notes/ifctester-verdict-audit.md`) to avoid re-treading its
+branches: `PartOf` (all 6 relation branches x `USERDEFINED`),
+`Specification.minOccurs`/`maxOccurs` (considered, correctly not pursued),
+and a full re-read of `reporter.py` (nothing beyond #9190).
+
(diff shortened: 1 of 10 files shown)


aothms:
For all ids fixes please make sure there is a minimal .ids  .ifc pair that
demonstrates the defect fixed by the pr.

Op za 1 aug 2026, 01:28 schreef Petru Conduraru ***@***.***>:

BIMvoice:
Done, and applied across the other five IDS PRs too (#9203, #9190, #9189, #9188, #9187).

Two pairs here, one per defect, under `src/ifctester/test/fixtures/`, wired through `ids.open()` + `ifcopenshell.open()` + `Ids.validate()` so they run the same path as the CLI.

`attribute_optional_unset`, an unset `IfcWall.Description` with `cardinality="optional"`:

* Unfixed: `[FAIL]`, reason `FALSEY`
* Fixed: `[PASS]`

`attribute_logical_unknown`, `IfcPresentationLayerWithStyle.LayerOn = UNKNOWN` under `prohibited`:

* Unfixed: `[PASS]`
* Fixed: `[FAIL]`

Both verified against the unfixed baseline as well as the fix, so they genuinely fail without it.

PR #9229, opened 2026-08-02
ifcmcp: fix descriptions in ifcquery/ifcedit instead of duplicating

brunopostle:
Alternative to #8955, for #8951. This is a more intrusive fix for the same problem.

23 of 25 ifcmcp tools reach MCP clients with an empty description, because FastMCP reads each server.py wrapper's own docstring, and those wrappers had none. #8955 fixes this by writing a new docstring directly onto each wrapper. Most of those wrappers are one-line passthroughs to an IfcSession method in core.py, which itself was often already a passthrough to an ifcquery/ifcedit function - so that fix ends up tripling the same description across three layers that can drift out of sync.

This fixes the same bug by enriching the true source instead: the ifcquery/ifcedit library functions each tool delegates to, which are useful independently of MCP. core.py's IfcSession methods copy the delegate's docstring at class-definition time via a small `_use_doc()` decorator, and server.py registers each tool with its description pulled from the matching IfcSession method. Methods that aren't passthroughs (session lifecycle, generic API/shape dispatch) keep their own hand-written docs. Keeps #8955's regression test.

Also corrects one detail from #8955: `ifc_new(schema="IFC4X3_ADD2")` reports model.schema as "IFC4X3", not "IFC4X3_ADD2" as previously documented — verified directly against ifcopenshell.file().

### DIFF
--- a/src/ifcedit/ifcedit/discover.py
+++ b/src/ifcedit/ifcedit/discover.py
@@ -94,9 +94,15 @@ def list_functions(module: str) -> list[dict]:
 
 
 def function_docs(module: str, function: str) -> dict:
-    """Full documentation for a single API function.
+    """Show the full documentation for one ifcopenshell.api function.
 
-    Returns a dict with: module, function, description, params (with types/defaults/descriptions), return_type
+    Returns the summary and long description, every parameter with its type,
+    default and description, and the return type. Read this before calling
+    ``run_api()`` so that parameter names and value types are correct.
+
+    :param module: API module name, for example ``'root'``.
+    :param function: Function name within the module, for example
+        ``'create_entity'``.
     """
     fn = _get_underlying_function(module, function)
     if fn is None:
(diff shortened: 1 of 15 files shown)


mtngrj:
This looks right to me. Tripling the same prose across the wrapper, `IfcSession` and the underlying function was the weak point of #8955, and sourcing it from `ifcquery`/`ifcedit` gives it one place to live. Avoiding `functools.wraps` so the signature — and the tool schema derived from it — stays untouched is a nice detail.

I'm closing #8955 in favour of this.

One thing you may want to know, no rush: the `ci-lint` failure is Black wanting to reformat `src/ifcopenshell-python/test/api/alignment/test_referent_names.py`, which isn't among the files this PR touches, and the most recent push to `v0.8.0` shows the same job failing. So it doesn't look like it comes from this branch. `ci` passes.

Thanks for picking this up and doing it properly.

PR #9236, opened 2026-08-03
Bonsai: wall side areas wrong for baked openings and Y-oriented walls (#9235, #9237)

### LINKED ISSUE #9235: Bug: GrossSideArea == NetSideArea in Bonsai QTO (openings not accounted for)
### Bug Description

Testmodel: 

[Testmodel.ifc.txt](https://github.com/user-attachments/files/30659490/Testmodel.ifc.txt)

## Symptom

After *Perform Quantity Takeoff* in Bonsai:

| Source | GrossSideArea | NetSideArea |
|---|---|---|
| Bonsai | 9.80815 m² | 9.80815 m² |
| Solibri | 20.08 m² | 9.81 m² |

Bonsai Net ≈ Solibri Net. Bonsai Gross is wrong (equal to Net).
Applies to (almost) all walls in this model, not just this one.

---

## Repro

1. Open `tests/Testmodel.ifc` in Blender + Bonsai.
2. Select wall `0P6UZIU3HDXOpp9cgjZsYR`
   (`Basic Wall:22_OW_HSB_260mm`, storey B-05).
3. *Perform Quantity Takeoff* (rule `IFC4QtoBaseQuantitiesBlender`).
4. Read `Qto_WallBaseQuantities.GrossSideArea` / `NetSideArea`.

Standalone evidence (no Blender required):

```text
Body: Tessellation (IfcPolygonalFaceSet)
HasOpenings: []          ← no IfcRelVoidsElement
bbox L×H = 7.21 × 2.785 = 20.07985  ← Solibri Gross
util.shape.get_side_area = 9.808147 ← Solibri/Bonsai Net
disable-opening-subtractions True/False: identical (no opening relationships)
```

Model-wide (`tests/Testmodel.ifc`):

| Measurement | Value |
|---|---|
| IfcWall | 120 |
| walls with `HasOpenings` | **4** |
| walls without opening relationship | **116** |
| IfcOpeningElement total | 96 (24 on walls, 72 on slabs) |
| walls where bbox L×H − mesh side > 0.05 m² | **84** |

Conclusion: openings are mostly **baked into the tessellation**,
not present as `IfcOpeningElement` on the wall.

---

## Root cause

Bonsai v0.8.x, `bonsai.bim.module.qto.calculator`:

```python
def get_gross_side_area(obj):
    if not has_openings(obj):          # ← only IfcRelVoidsElement
        return get_net_side_area(obj)  # ← Gross becomes Net
    return get_lateral_area(..., subtract_openings=False, ...) / 2

def has_openings(obj):
    return [o for o in ifcopenshell.util.element.get_openings(element)]
```

Rule mapping (`ifc5d/IFC4QtoBaseQuantitiesBlender.json`):

```text
GrossSideArea → get_gross_side_area
NetSideArea   → get_net_side_area
```

`get_net_side_area` = lateral mesh area / 2 → correct Net (≈ Solibri).

`get_gross_side_area` short-circuits when there are no opening *relationships* →
Gross = Net, even when the mesh still has holes.

The same gate exists on `get_gross_volume` / `get_gross_surface_area`:
the gross path uses `disable-opening-subtractions`, but that only helps
when there are opening relationships to skip. With pure tessellation,
gross mesh == net mesh.

IFC4 definition (for reference):

- **GrossSideArea** — elevation view of the wall middle plane, **including** openings
  (≈ Length × Height of the full wall).
- **NetSideArea** — the same plane, **excluding** openings.

Solibri follows that definition (Gross = L×H). Bonsai only follows it when
`HasOpenings` is populated.

---

## What this is not

- Not a UI bug / rounding issue.
- Not a wrong QTO rule mapping in the installed Blender JSON
  (it correctly maps to `get_*_side_area`; the skill-doc example mapping to
  `get_gross_surface_area` is outdated/incorrect vs the real JSON).
- Not a Solibri error on this wall: L×H = 20.07985 matches Solibri 20.08.

---

### A — Fix in Bonsai `get_gross_side_area` (recommended for the QTO symptom)

When `not has_openings(obj)`: Gross ≠ blindly Net, but bbox elevation:

```python
# already in calculator.py:
# get_side_area(o) := max(get_x(o)*get_z(o), get_y(o)*get_z(o))
return get_side_area(obj)
```

Then for this wall: Gross ≈ 20.08, Net ≈ 9.81 → Solibri parity.
Without openings (full rectangular wall) Gross ≈ Net remains.

**Ceiling:** bbox L×H overestimates for L-shaped / non-rectangular
plan outlines. Upgrade: outer-contour projection instead of bbox, or
parametric Length×Height from type/occurrence.

Same consideration for GrossVolume if volume must also be correct without
opening relationships — separate, harder fix (reconstructing a mesh without
holes is not possible from pure tessellation alone).

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Fixes #9235.

`GrossSideArea` equals `NetSideArea` whenever a wall's openings are baked into its tessellation rather than modelled as `IfcOpeningElement`. @PahlawanJoey diagnosed this precisely, measured his whole model, and cross-checked against Solibri.

```python
def get_gross_side_area(obj):
    if not has_openings(obj):          # only sees IfcRelVoidsElement
        return get_net_side_area(obj)  # Gross becomes Net
```

On his model: 120 walls, only **4** with `HasOpenings`, and his named wall `0P6UZIU3HDXOpp9cgjZsYR` reports 9.808 for both, where Solibri reports 20.08 and 9.81.

### Why the obvious fix was not taken

Joey proposed falling back to the bounding-box elevation, `max(x*z, y*z)`, which matches Solibri on this model. **That regresses non-rectangular walls, which was measured rather than assumed.**

On a hand-built gable wall with no openings and a true elevation area of 18.0 m2, the bounding-box fallback returns **24.0 m2, 33% too high**. A bounding box cannot distinguish a rectangular wall with a hole from a non-rectangular wall without one.

### What this does instead

The openings are recoverable from the geometry. Joey's file contains 44 `IfcIndexedPolygonalFaceWithVoids`, and his named wall uses two of them.

Rather than reading those interior loops directly, which only works for that one representation type, this works from the tessellated mesh topology, which is general: project the wall's Y-facing faces onto the XZ plane, union them with Shapely, and add the area of any resulting **interior ring** to the already-correct `NetSideArea`.

A genuine hole produces an interior ring. A non-rectangular outline does not. So:

| case | net | gross |
|---|---|---|
| wall with a hole cut into the mesh | 11.0 | **12.0** |
| gable wall, no openings | 18.0 | **18.0**, unchanged |
| Joey's wall `0P6UZIU3HDXOpp9cgjZsYR` | 9.808 | matches Solibri |

Returning `net + void_area` is also monotonically safe: gross can never come out **below** net, which a naive "replace net with the projected outline" approach does for foreshortened or curved faces.

### A crash fixed along the way

`get_projected_area` raises `'MultiPolygon' object has no attribute 'interiors'` whenever a wall's projected faces are not a single connected polygon. That affects **12 of the 156 walls** in Joey's file. It was previously unused, so the crash was latent. The new `get_side_void_area` handles both `Polygon` and `MultiPolygon`.

### Verification

Whole-model sweep of Joey's file, all 156 walls including `IfcWallStandardCase`: **0 exceptions, 0 cases of gross below net**, 0.12s.

`test/tool/test_qto.py` adds `TestGetGrossSideAreaWithoutOpeningRelationship`, covering both the baked-opening recovery and the gable no-regression case. All 12 tests pass under real headless Blender.

---

## Second commit: the same functions used the wrong axis for Y-oriented walls (#9237)

@PahlawanJoey found a second defect in these functions, and it turned out the fix above shared it.

`get_net_side_area` and `get_gross_side_area` both call `get_lateral_area(..., main_axis="x")`, hardcoding the wall's length to local X. For a wall whose length runs along **Y**, they measure the **end face** instead of the elevation.

**And the fix in the first commit had the same assumption.** `get_side_void_area` filters `polygon.normal.y == 0` and projects `(co.x, co.z)`, so on a Y-oriented wall it inspected the wrong faces and always returned a void of zero. The baked-opening recovery silently did nothing for those walls.

### Every hardcoded-axis site

* `get_net_side_area` and `get_gross_side_area`, both forcing `main_axis="x"`
* `get_lateral_area`'s dispatch, `if main_axis_guess == "x" or main_axis == "x":`, where the existing bbox-based guess was always overridden
* `get_lateral_area`'s `"y"` branch was itself wrong: `side_axis = z_axis; top_axis = x_axis` rather than mirroring the `"x"` branch. Walls are vertical, so Z stays the height reference whichever horizontal axis is longer
* `get_side_void_area` from the first commit

Axis selection now compares only the local X and Y extents. `get_object_main_axis` compares all three, which would wrongly pick Z for a door taller than it is wide.

### Measured on the reporter's file

His six named GlobalIds, before: Net = Gross ≈ 1.149, matching his report. After: **Net 4.2437, Gross 9.5987**, matching Solibri's 4.24 and 9.60. Cross-checked by manual projection: net face 4.2437 plus void 5.355 equals 9.598, which is L × H.

Across all 120 walls in the file, 108 X-oriented and 12 Y-oriented: **the X-oriented walls change by at most 4.8e-8**, float noise. All 12 Y-oriented walls now report the corrected figures. The 42 X-oriented walls with detected baked-opening voids are unaffected, so the first commit's behaviour still holds.

### Not handled, stated plainly

A wall running at 45 degrees to both local axes reports 6.30 m² against a true 12.0 m². **That is unchanged by this commit** and was equally wrong before, being a limitation of axis-aligned face classification rather than of the axis choice.

`get_opening_area` also hardcodes `main_axis="x"`, against an opening's own OBB. Same defect class, left alone because the reporter's file contains no Y-oriented wall with a formal `IfcRelVoidsElement` opening to verify against.

### Tests

`test_y_oriented_wall_baked_opening_is_recovered` and `test_x_and_y_oriented_walls_agree`, generalising the first commit's wall builder to build along either axis. All 14 tests in `test/tool/test_qto.py` pass.

### DIFF
--- a/src/bonsai/bonsai/bim/module/qto/calculator.py
+++ b/src/bonsai/bonsai/bim/module/qto/calculator.py
@@ -815,13 +815,56 @@ def get_lateral_area(
     return area + total_opening_area
 
 
+def get_side_void_area(obj: bpy.types.Object) -> float:
+    """Total area of holes enclosed by the object's Y-facing mesh faces.
+
+    Openings can be baked directly into the body tessellation (e.g. an
+    IfcIndexedPolygonalFaceWithVoids) rather than modelled as an
+    IfcOpeningElement relationship. Such holes leave no trace in
+    ifcopenshell.util.element.get_openings, but they do leave a real gap
+    in the mesh. Projecting the Y-facing faces onto the XZ plane and
+    unioning them exposes that gap as an interior ring, so its area can be
+    recovered exactly instead of guessed from a bounding box.
+
+    Returns 0 if no such enclosed hole is found, which is also the correct
+    answer for a solid wall of any (including non-rectangular) outline.
+    """
+    assert isinstance(obj.data, bpy.types.Mesh)
+    odata = obj.data
+    shapely_polygons = []
+    for polygon in odata.polygons:
+        if polygon.normal.y == 0:
+            continue
+        polygon_tuples = []
+        for loop_index in polygon.loop_indices:
+            loop = odata.loops[loop_index]
+            co = odata.vertices[loop.vertex_index].co
+            polygon_tuples.append((co.x, co.z))
+        try:
+            shapely_polygon = Polygon(polygon_tuples)
+        except Exception:
+            continue
+        if shapely_polygon.is_valid and not shapely_polygon.is_empty:
+            shapely_polygons.append(shapely_polygon)
+
+    if not shapely_polygons:
+        return 0.0
+
+    union = unary_union(shapely_polygons)
+    parts = union.geoms if hasattr(union, "geoms") else [union]
+
+    void_area = 0.0
+    for part in parts:
+        for interior in getattr(part, "interiors", []):
+            void_area += Polygon(interior).area
+    return void_area
+
+
 def get_gross_side_area(obj: bpy.types.Object) -> float:
-    if not has_openings(obj):
-        return get_net_side_area(obj)
+    if has_openings(obj):
+        return get_lateral_area(obj, exclude_end_areas=True, subtract_openings=False, main_axis="x") / 2
 
-    gross_side_area = get_lateral_area(obj, exclude_end_areas=True, subtract_openings=False, main_axis="x") / 2
-
-    return gross_side_area
+    return get_net_side_area(obj) + get_side_void_area(obj)
 
 
 def get_net_side_area(obj: bpy.types.Object) -> float:
--- a/src/bonsai/bonsai/bim/module/qto/calculator.py
+++ b/src/bonsai/bonsai/bim/module/qto/calculator.py
@@ -770,7 +770,8 @@ def get_lateral_area(
     :param bool exclude_side_areas: , defaults to False
     :param int angle_z1: Angle measured from the positive z-axis to the normal-vector of the area. Openings with a normal_vector lower than this value will be ignored, defaults to 45
     :param int angle_z2: Angle measured from the positive z-axis to the normal-vector of the area. Openings with a normal_vector greater than this value will be ignored, defaults to 135
-    :param str main_axis: set main axis, for example a wall must have x main axis default 'x'
+    :param str main_axis: force the main (length) axis to "x", "y" or "z" instead of guessing
+        it from the object's bounding box. Defaults to "", which uses the guess.
     :return float: Lateral Area
     """
 
@@ -778,21 +779,21 @@ def get_lateral_area(
     y_axis = [0, 1, 0]
     z_axis = [0, 0, 1]
 
-    main_axis_guess = get_object_main_axis(obj)
-    if main_axis_guess == "x" or main_axis == "x":
+    resolved_axis: AxisType = main_axis if main_axis in ("x", "y", "z") else get_object_main_axis(obj)
+    if resolved_axis == "x":
         main_axis_v = x_axis
         side_axis = y_axis
         top_axis = z_axis
-    elif main_axis_guess == "z":
+    elif resolved_axis == "z":
         main_axis_v = z_axis
         side_axis = x_axis
         top_axis = y_axis
-    elif main_axis_guess == "y":
+    elif resolved_axis == "y":
         main_axis_v = y_axis
-        side_axis = z_axis
-        top_axis = x_axis
+        side_axis = x_axis
+        top_axis = z_axis
     else:
-        assert_never(main_axis_guess)
+        assert_never(resolved_axis)
 
     area = 0
     total_opening_area = 0 if subtract_openings else get_opening_area(obj, angle_z1=angle_z1, angle_z2=angle_z2)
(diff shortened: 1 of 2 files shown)


aothms:
There is no immediate reason for the use of shapely. I would much rather propose we decompose the mesh into (near-) coplanar facets using a dual graph like approach. Select the components facing Y and classify the boundary wires of these components. Also note that an opening does not necessarily create a inner wire. It can also mutate the exterior. It may be useful to consider something like a convex hull approach. We've been discussing a mesh/dual graph based framework https://github.com/IfcOpenShell/IfcOpenShell/discussions/7734 and https://github.com/IfcOpenShell/IfcOpenShell/pull/8332

BIMvoice:
You're right about the exterior mutation, and we get it wrong. Measured on a 4.0 x 3.0 x 0.2 wall with the opening baked into the mesh (no `IfcRelVoidsElement`), length along local X:

| elevation outline | true gross | v0.8.0 | this PR |
|---|---|---|---|
| window fully inside the wall | 12.0 | 11.0 | 12.0 |
| door cutting the bottom edge | 12.0 | 10.0 | 10.0 |
| slot cutting the top edge | 12.0 | 11.0 | 11.0 |
| notch cutting the wall end | 12.0 | 11.0 | 11.0 |

The interior-ring test only catches the first row. It isn't a regression, v0.8.0 returns the same wrong number, but it isn't a fix either, and I'd rather not ship a gross-area recovery that's silently right for windows and silently wrong for doors.

I measured the convex hull too, on those cases plus three solid walls with no openings at all:

| outline | true gross | convex hull |
|---|---|---|
| any of the four opening cases above | 12.0 | 12.0 |
| gable | 18.0 | 18.0 |
| L-shaped | 12.0 | 18.0 |
| stepped | 12.0 | 14.0 |

So the hull fixes every opening case and holds on the gable, but inflates a genuinely concave wall by 50%. That's your point again: from the 2D outline alone you can't tell a concave edge that came from an opening from one that was modelled that way. The dual graph is what carries that provenance, so I think you're pointing at the right thing.

On shapely, no argument about the direction, but it isn't new here. `calculator.py` already imports it on v0.8.0 (lines 31-32) and uses it in `get_projected_area` and `get_intersecting_area`. This PR adds a third caller, not a dependency.

On the framework, as far as I can tell it doesn't exist yet. #7734 is a Q&A thread with your Python prototype from 2 Mar, last activity 11 Apr. #8332 is mine, open since 7 Jul, where you made this same point and I agreed to do the cap-face / dual-graph method as a follow-up. Nothing merged, so I can't rebase onto it today.

What I'd like to do, unless you'd rather something else:

1. Cut this PR down to the axis fix (#9237) only. It's independent of all of the above and touches no shapely. Both callers hardcoded `main_axis="x"`, so a wall running along local Y reported 1.0 m2 for a 10.0 m2 elevation. The 108 X-oriented walls in the reporter's file change by at most 4.8e-8.
2. Drop the baked-opening gross recovery and leave #9235 open, tracked against the dual-graph work.

Confidence: the two tables are measured in headless Blender 5.2 against both v0.8.0 and this branch, not reasoned from the diff. I have not tested the formal `IfcRelVoidsElement` path against a boundary-cutting opening, so I don't know whether that path has the same hole.

PR #9241, opened 2026-08-03
tests: fix black drift blocking unrelated commits

BIMvoice:
One line in `test/api/alignment/test_referent_names.py` is **121 characters**, one over the repo's 120-column black limit:

```
$ black --check --line-length 120 src/ifcopenshell-python/test/api/alignment/test_referent_names.py
1 file would be reformatted.
```

It arrived with `e077390e3d` and is already on `v0.8.0`.

### Why it is worth a PR of its own

The `pre-commit` hook lints the **whole** staged diff, not just the changed hunks. So any commit whose diff happens to include this file fails the hook, even when the change has nothing to do with it.

That tripped **every one of seven** merges of `v0.8.0` into open PR branches today, each needing the same unrelated reformat before it could be committed. Fixing it once on `v0.8.0` stops it recurring for everyone.

Pure whitespace reformat, produced by running the repo's own black configuration. No semantic change.


BIMvoice:
Flagging the blast radius on this one, since it is larger than the diff suggests.

`src/ifcopenshell-python/test/api/alignment/test_referent_names.py` on current `v0.8.0` fails `black --check` with the CI's black version. Verified by pulling that file straight from `origin/v0.8.0` and running the CI-exact black against it, nothing else involved.

The effect is that **any** PR based on `v0.8.0` which does not happen to touch that file shows `lint-formatting: fail`, through no fault of its own. We audited 31 of our own open PRs today: 28 are red, every one of them for this single reason, with `compile-and-test` and `build-ifcopenshell` passing. Spot-checked ten of the job logs individually and the sole failure line in each is `would reformat .../test_referent_names.py`.

This is not specific to us. Any contributor opening a PR against `v0.8.0` right now inherits the same red check, which makes it harder to tell a real failure from noise.

This PR is green on all four checks and mergeable. Merging it clears the red from everything based on `v0.8.0` on the next run, without anyone touching those branches.

@Moult flagging for a merge decision, since it is blocking more than its own diff.

Issue #283, opened 2017-10-31
Viewing IFC's through pythonOCC


andersgb:
I'm trying to do essentially what's described in this dated blogpost: http://ifcopenshell.org/pythonOCC/example1/

This is my attempt at visualizing IFC from latest IfcOpenShell master:
```python
from __future__ import print_function
import ifcopenshell
import ifcopenshell.geom

# Initialize a graphical display window
occ_display = ifcopenshell.geom.utils.initialize_display()
settings = ifcopenshell.geom.settings()
# Open the IFC file using IfcOpenShell
ifc_file = ifcopenshell.open("/home/anders/Downloads/IfcOpenHouse.ifc")

# Display the geometrical contents of the file using Python OpenCascade
products = ifc_file.by_type("IfcProduct")
for product in products:
    if product.is_a("IfcOpeningElement"): continue
    shape = ifcopenshell.geom.create_shape(settings, product)
    if shape:
        display_shape = ifcopenshell.geom.utils.display_shape(shape)
        if product.is_a("IfcPlate"):
            # Plates are the transparent parts of the window assembly
            # in the IfcOpenHouse model
            ifcopenshell.geom.utils.set_shape_transparency(display_shape, 0.8)

print("Finished loading..")
raw_input()
```

This particular code crashes on `Representation` entity not found I think:
```
  File "view.py", line 15, in <module>
    shape = ifcopenshell.geom.create_shape(settings, product)
  File "/usr/lib/python3/dist-packages/ifcopenshell/geom/main.py", line 125, in create_shape
    repr.wrapped_data if repr is not None else None
RuntimeError: Entity not found
```

If I modify the script to e.g. only show `IfcWall` objects (that should have `Representation`), I instead get the following crash:
```
Traceback (most recent call last):
  File "view.py", line 18, in <module>
    display_shape = ifcopenshell.geom.utils.display_shape(shape)
  File "/usr/lib/python3/dist-packages/ifcopenshell/geom/occ_utils.py", line 159, in display_shape
    ais = OCC.AIS.AIS_Shape(shape)
  File "/usr/lib/python3/dist-packages/OCC/AIS.py", line 13110, in __init__
    _AIS.AIS_Shape_swiginit(self, _AIS.new_AIS_Shape(*args))
TypeError: in method 'new_AIS_Shape', argument 1 of type 'TopoDS_Shape const &'
```

Crashes seem to be the same for python2 and python3 btw.

Any ideas?

[...]

PR #9244, opened 2026-08-04
ifctester: put pass/fail at the end of generated testcase names

BIMvoice:
`ids_doc_generator.py` generates the buildingSMART IDS conformance test cases from the assertions in `test_facet.py` and `test_ids.py`. Their filenames currently lead with the expected result:

```python
basename = f"{result}-" + re.sub("[^0-9a-zA-Z]", "_", name.lower())
```

so a scenario's pass and fail cases sort far apart from each other.

[buildingSMART/IDS#283](https://github.com/buildingSMART/IDS/issues/283) asks for the result at the end so the pair sorts together:

```
before   pass-nesting_may_be_indirect.ifc
after    nesting_may_be_indirect-pass.ifc
```

That issue has been open since May 2024 and is labelled "please contribute". Since the names originate here rather than in the IDS repository, this is the place to change it.

Two occurrences, lines 98 and 184. Nothing else depends on the prefix: `basename` only feeds the `.ifc` and `.ids` file writes and the generated documentation links.

The renamed files appear in the IDS repository the next time the suite is regenerated. Happy to raise the corresponding bulk rename there if maintainers want it done sooner.

### DIFF
--- a/src/ifctester/test/ids_doc_generator.py
+++ b/src/ifctester/test/ids_doc_generator.py
@@ -95,7 +95,7 @@ class FacetDocGenerator:
         # ifc_text = "\n".join([f"{e} /* Testcase */" if e == inst else str(e) for e in f])
         lines = ifc.wrapped_data.to_string().split("\n")[7:-3]
         ifc_text = "\n".join([f"{l} /* Testcase */" if f"#{inst.id()}=" in l else l for l in lines])
-        basename = f"{result}-" + re.sub("[^0-9a-zA-Z]", "_", name.lower())
+        basename = re.sub("[^0-9a-zA-Z]", "_", name.lower()) + f"-{result}"
 
         # Write IFC to disk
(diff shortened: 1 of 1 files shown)


BIMvoice:
Closing this. It does not do what its description claims.

@andyward pointed out on [buildingSMART/IDS#283](https://github.com/buildingSMART/IDS/issues/283) that the conformance test cases are generated inside the IDS repository from fenced blocks in `Documentation/ImplementersDocumentation/TestCases/scripts.md`, not from `ids_doc_generator.py` here. I checked and he is right: this repository ships zero `.ids` files and nothing consumes the `pass-`/`fail-` prefix, so renaming it here changes nothing and does not reach their suite.

The body of this PR says "the names originate here rather than in the IDS repository", which is false.

The rename also turns out not to be cosmetic on their side: `RepositoryAutomation/build/Build.cs` anchors on the `invalid-` prefix twice, so it would need changing at the same time. That is recorded on their #283 and is a decision for them.

PR #9305, opened 2026-08-15
win: collect underscore-prefixed runtime plugin DLLs in build artifacts

### LINKED ISSUE #9301: v0.9.0 win64 build artifacts omit all runtime plugin DLLs (parse_schema_*, geometry_mapping_*) — schema creation and geometry fail
## Summary

The `v0.9.0alpha0` win64 artifacts on builds.ifcopenshell.org are missing every load-by-name runtime plugin DLL of the new modular architecture (`ifcopenshell_parse_schema_ifc*.dll`, `ifcopenshell_geometry_mapping_ifc*.dll`, extra kernels, document codecs, format writers). As shipped, `ifcopenshell-python-311-v0.9.0alpha0-83fc219-win64.zip` cannot create or open a model:

```
>>> import ifcopenshell
>>> ifcopenshell.file(schema="IFC4X3_ADD2")
RuntimeError: No schema named IFC4X3_ADD2
```

After manually supplying the `ifcopenshell_parse_schema_*.dll` set (lifted from the BonsaiViewer win64 zip, same commit), parsing/authoring works but all geometry fails:

```
RuntimeError: No geometry mapping registered for ifc4x3_add2
```

No win64 artifact contains `ifcopenshell_geometry_mapping_*.dll` at all (checked ifcopenshell-python-311, BonsaiViewer, IfcConvert, IfcGeomServer at `83fc219`).

## Evidence

Binaries in `ifcopenshell-python-311-v0.9.0alpha0-83fc219-win64.zip` (7 total):

```
_ifcopenshell_wrapper.cp311-win_amd64.pyd
ifcopenshell.geometry.dll
ifcopenshell.geometry.writer.dll
ifcopenshell.parse.dll
ifcopenshell.plugin.dll
ifcopenshell_document_rdb.dll
ifcopenshell_geometry_kernel_opencascade.dll
```

The linux64 zip at the same commit is complete (~40 plugin `.so` files: 8× `ifcopenshell_parse_schema_*`, 8× `ifcopenshell_geometry_mapping_*`, 5 kernels, per-schema document json/xml codecs, geometry format writers, trees). The pyodide modular zip likewise ships `parse_schema` and `geometry_mapping` wheels. The gap is win-specific.

## Root cause

`win/build-all-win.py` collects runtime plugins with a dot-prefix filter, in both `archive_executables` and `archive_python_package`:

```python
ifc_runtime_plugins = {
    d
    for d in (set(dlls) - dependencies)
    if d.name.startswith("ifcopenshell.") and not d.name.startswith("ifcopenshell.geometry.writer.")
}
```

But runtime plugins are underscore-named: `src/plugin/plugin.cpp` (`decorated_basename`) canonically prepends `ifcopenshell_`, and the CMake plugin targets set `OUTPUT_NAME "ifcopenshell_parse_schema_ifc${schema}"` / `"ifcopenshell_geometry_mapping_ifc${schema}"`. `startswith("ifcopenshell.")` (trailing dot) only matches the core shared libraries, so every load-by-name plugin is silently dropped from every zip.

The two underscore-named DLLs that *do* appear in the python zip (`ifcopenshell_document_rdb.dll`, `ifcopenshell_geometry_kernel_opencascade.dll`) appear to get in via `trace_runtime_dependencies` (dumpbin) as direct link-time dependencies, not via the plugin filter — which also explains why the BonsaiViewer zip contains the `parse_schema` DLLs (linked by the exe) but no `geometry_mapping` DLLs (loaded purely by name, invisible to dumpbin — as the comment in `archive_executables` itself notes).

## Suggested fix

Admit the underscore prefix in both collection sites:

```python
if d.name.startswith(("ifcopenshell.", "ifcopenshell_")) and not d.name.startswith("ifcopenshell.geometry.writer.")
```

win-arm64 uses the same script and is presumably affected identically.

## Environment

- Artifacts: `v0.9.0alpha0-83fc219` win64 from https://builds.ifcopenshell.org/
- Windows 11, Python 3.11 (Blender 5.0 bundled)
- Found while porting Bonsai civil-engineering work onto the v0.9.0 branch; with schemas manually supplied everything non-geometry passes our test suite, so the wrapper itself is fine — it's purely packaging. Happy to test fixed artifacts.

DesertSpringsCivil:
Fixes #9301.

### Problem

Runtime plugins are canonically underscore-prefixed: `decorated_basename()` in `src/plugin/plugin.cpp` prepends `ifcopenshell_`, and the plugin targets set `OUTPUT_NAME` to match — `ifcopenshell_parse_schema_ifc${schema}`, `ifcopenshell_geometry_mapping_ifc${schema}`, `ifcopenshell_geometry_kernel_${kernel}`, `ifcopenshell_geometry_writer_ifc${schema}`, `ifcopenshell_${plugin_output_name}` for the document/geometry serializers. Only the core shared libraries keep dotted names (`ifcopenshell.parse`, `ifcopenshell.geometry`, `ifcopenshell.geometry.writer`, `ifcopenshell.plugin`).

`win/build-all-win.py` collected runtime plugins with a dot-prefix filter, in both `archive_executables()` and `archive_python_package()`:

```python
if d.name.startswith("ifcopenshell.") and not d.name.startswith("ifcopenshell.geometry.writer.")
```

The trailing dot only matches the core libraries, so every load-by-name plugin was silently dropped from every win64 and win-arm64 zip.

The handful of underscore-named DLLs that did make it into the shipped artifacts (`ifcopenshell_document_rdb.dll`, `ifcopenshell_geometry_kernel_opencascade.dll`, and the `parse_schema` set in the BonsaiViewer zip) got there via `trace_runtime_dependencies()` as link-time dependencies, not via the plugin filter — which is also why the `geometry_mapping` plugins, loaded purely by name and invisible to dumpbin, are absent from every artifact.

On the shipped v0.9.0alpha0 win64 artifacts this means:

```
>>> ifcopenshell.file(schema="IFC4X3_ADD2")
RuntimeError: No schema named IFC4X3_ADD2
```

and, with the `parse_schema` DLLs supplied manually:

```
RuntimeError: No geometry mapping registered for ifc4x3_add2
```

linux64 and pyodide artifacts are unaffected.

### Change

Both collection sites now share one helper that accepts both prefixes, and the geometry-writer exclusion is extended to the underscore form so the per-schema writers keep their existing Python-package-only treatment.

### One question for review

`ifcopenshell_geometry_writer_ifc*.dll` stays excluded from the executable zips, which preserves current behaviour — but that exclusion predates the rename, and under the old naming `ifcopenshell.geometry.writer.` would have matched only the core library. If IfcConvert is meant to load the per-schema writers at runtime, they should be bundled with the executables too. Happy to change it either way.

### Testing

I don't have a Windows build environment to run the full script, so the filter was verified against a reconstruction of the installed `bin/` DLL set: all load-by-name plugins are now collected, the per-schema writers still route to the Python package only, and third-party DLLs are unaffected. Happy to test the resulting artifacts once CI produces them.

### DIFF
--- a/win/build-all-win.py
+++ b/win/build-all-win.py
@@ -53,6 +53,16 @@ QT_DEPLOYMENT_DLLS = {
     "vulkan-1.dll",
 }
 QT_CONF = "[Paths]\nPrefix = .\n"
+# Runtime plugins are canonically prefixed with 'ifcopenshell_' (see
+# decorated_basename() in src/plugin/plugin.cpp and the OUTPUT_NAME properties of
+# the plugin targets, e.g. 'ifcopenshell_parse_schema_ifc${schema}'), while the
+# core shared libraries keep the dotted 'ifcopenshell.' names. Match both so the
+# load-by-name plugins are not silently dropped from the archives.
+IFC_RUNTIME_PLUGIN_PREFIXES = ("ifcopenshell.", "ifcopenshell_")
+# Per-schema geometry writers ship with the Python package only, not next to the
+# executables. 'ifcopenshell.geometry.writer.' covers the core library, the
+# underscore form covers the per-schema plugins.
+IFC_GEOMETRY_WRITER_PREFIXES = ("ifcopenshell.geometry.writer.", "ifcopenshell_geometry_writer_")
 
 
 def run(command: list[str]) -> None:
@@ -153,6 +163,19 @@ def trace_runtime_dependencies(roots: set[Path], candidates: set[Path]) -> set[P
     return resolved
 
 
+def is_geometry_writer(file: Path) -> bool:
+    return file.name.startswith(IFC_GEOMETRY_WRITER_PREFIXES)
+
+
+def collect_ifc_runtime_plugins(dlls: set[Path], dependencies: set[Path]) -> set[Path]:
+    """IfcOpenShell plugins are loaded by name at runtime, so dumpbin cannot discover them."""
+    return {
+        d
+        for d in (dlls - dependencies)
+        if d.name.startswith(IFC_RUNTIME_PLUGIN_PREFIXES) and not is_geometry_writer(d)
+    }
+
+
 def is_qt_deployment_dll(file: Path) -> bool:
     name = file.name.lower()
     return name.startswith("qt") or name.startswith("d3dcompiler_") or name in QT_DEPLOYMENT_DLLS
@@ -223,11 +246,7 @@ def archive_executables() -> None:
     exes = {file for file in bin_files if file.suffix.lower() == ".exe"}
     dlls = {file for file in bin_files if file.suffix.lower() == ".dll"}
     dependencies = trace_runtime_dependencies(exes, dlls)
-    ifc_runtime_plugins = {
-        d
-        for d in (set(dlls) - dependencies)
-        if d.name.startswith("ifcopenshell.") and not d.name.startswith("ifcopenshell.geometry.writer.")
-    }
+    ifc_runtime_plugins = collect_ifc_runtime_plugins(dlls, dependencies)
     qt_deployment_files = collect_qt_deployment_files(install_dir)
 
     for file in sorted(exes):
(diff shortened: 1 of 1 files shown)


DesertSpringsCivil:
CI triage: both failing checks are pre-existing on the `v0.9.0` base, not introduced here — the branch's own last four `ci` runs and last three `ci-lint` runs (Aug 10–14) all fail identically. The lint failure is Black 26.5.1 wanting to reformat `pyodide/build-all-pack-wheel-local.py`, which this PR doesn't touch; this PR's only file (`win/build-all-win.py`) passes Black/ruff cleanly. Happy to rebase or adjust if anything else is wanted.

PR #9312, opened 2026-08-15
ifcdiff: compare container and aggregate by GlobalId, not by instance

ProductOfAmerica:
Running IfcDiff with `relationships=["container", "aggregate"]` on a model and an identical copy of it reports every element that has a container or an aggregate as changed. On a small IFC4 model with 8 spatial and physical elements, all 8 come back with `{"container_changed": True}` or `{"aggregate_changed": True}`, and the register is exactly the same as the one you get when a wall genuinely moves to another storey. A real move ends up indistinguishable from no change at all.

The cause is in `diff_element_relationships`: it compares the `entity_instance` objects returned by `get_container` and `get_aggregate`, but the old and new instances belong to two different `ifcopenshell.file` objects, and instance equality is identity within a single file, so the comparison is unequal even when nothing moved.

The fix compares the GlobalIds instead, with a `None` fallback so an added or removed container is still reported. This is what the `type` branch a few lines above already does. The register value stays `True`, so the output shape and the shallow-return behaviour are unchanged.

Repro (Project > Site > Building > two storeys, 4 walls contained in Level 1, copied via `to_string`/`from_string`):

| | before | after |
|---|---|---|
| identical models | 8 changed | 0 changed |
| one wall recontained to Level 2 | 8 changed | 1 changed (the moved wall) |

Three tests added to `src/ifcdiff/test.py` in the existing `setup_project()` style: `test_unchanged_container_is_not_a_change`, `test_unchanged_aggregate_is_not_a_change`, `test_changed_container`. `pytest -p no:pytest-blender test.py` is 2 failed / 7 passed before the change and 9 passed after. black and ruff at line-length 120 are clean.

This is the container/aggregate part of #4110. I'm aware #8294 and #8249 proposed the same fix and were closed by their author in favour of #9296, which is still a draft and bundles this with six other ifcdiff changes, including turning `container_changed` into a dict of old and new ids, which changes the output shape for anything reading `change_register` or the exported JSON (Bonsai's diff panel included). This PR is deliberately just the two-line behaviour fix with no shape change. If #9296 lands first this becomes redundant and I'll close it.

### DIFF
--- a/src/ifcdiff/ifcdiff.py
+++ b/src/ifcdiff/ifcdiff.py
@@ -325,11 +325,17 @@ class IfcDiff:
                     self.change_register.setdefault(new.GlobalId, {}).update({"properties_changed": diff})
                     return True
             elif relationship == "container":
-                if ifcopenshell.util.element.get_container(old) != ifcopenshell.util.element.get_container(new):
+                # Compare by GlobalId, not by instance: the two containers belong to
+                # different files, and instances from different files are never equal.
+                old_container = ifcopenshell.util.element.get_container(old)
+                new_container = ifcopenshell.util.element.get_container(new)
+                if getattr(old_container, "GlobalId", None) != getattr(new_container, "GlobalId", None):
                     self.change_register.setdefault(new.GlobalId, {}).update({"container_changed": True})
                     return True
             elif relationship == "aggregate":
-                if ifcopenshell.util.element.get_aggregate(old) != ifcopenshell.util.element.get_aggregate(new):
+                old_aggregate = ifcopenshell.util.element.get_aggregate(old)
+                new_aggregate = ifcopenshell.util.element.get_aggregate(new)
+                if getattr(old_aggregate, "GlobalId", None) != getattr(new_aggregate, "GlobalId", None):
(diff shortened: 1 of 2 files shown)


BIMvoice:
Nice catch, and the right shape of fix. I had reached the same conclusion independently in #9296, with the same comparison by GlobalId and the same None handling, so this is a genuine confirmation rather than a competing opinion.

Yours is the better PR to carry it: it is narrower, it is not a draft, and the regression tests are focused on exactly this behaviour. So I have dropped the container and aggregate fix from #9296 and left those code paths byte identical to `v0.8.0`, and confirmed with `git merge-tree` that the two branches now merge cleanly in either order. Nothing here needs to wait on anything of mine.

For anyone reviewing both: #9296 now covers six other ifcdiff defects that this one does not touch, mostly changes that go entirely undetected rather than falsely detected. The two are complementary.

ProductOfAmerica:
Appreciated, and thanks for checking the merge order both ways. Agreed the two are complementary; the under-reporting fixes in #9296 look worth having. I'll keep this one rebased if v0.8.0 moves.

PR #9330, opened 2026-08-19
Add cross-coplanar and material/style-change edge classification

sboddy:
# Add cross-coplanar & material/style-change SVG edge classification (#3742)

## Summary

Extends the existing single-object SVG edge classification system (#3668,
5 classes: outline/crease/valley/ridge/boundary) to also handle edges
shared between *different* touching elements. Previously, every product's
edges were classified independently, which produces two visible problems
in generated drawings:

1. Two elements with a genuinely coincident, matching surface (e.g. two
   walls butting together) each draw their own copy of the shared boundary
   edge -- correct individually, but a visible double-line once any
   transparency is applied.
2. Where two coincident surfaces belong to elements of *different*
   material or style, or where a single element's own section-cut face
   spans more than one material layer, there was previously no way to
   mark that boundary as anything other than an ordinary edge -- so a
   material change at a shared or internal surface was invisible in the
   drawing.

This PR adds two new edge classes computed directly on the real 3D BRep
geometry (before HLR projection, not as an SVG post-process) --
`cross-coplanar` and `mat-style-change` -- plus an opt-in duplicate-edge
merge pass for drawings where two classes end up needing to be drawn at
the same location, a run of correctness/robustness fixes found while
building and reviewing this, and two performance fixes for larger
drawings.

composite_coplanar

### New edge classification

- **`cross-coplanar`**: reclassifies whatever portion of an edge lies on
  a coincident, same-material/style face of a *different* product -- the
  genuinely-duplicate-boundary case above. Matching works at sub-edge
  granularity: a partial overlap, or several products contributing to
  different sections of the same edge, is split and classified per
  sub-range rather than requiring the entire edge to match. Off by
  default for rendering (`svg-render-cross-coplanar-edges`); the
  underlying match pass can also gate other classification behaviour
  independently (`svg-use-cross-coplanar-classification`).
- **`mat-style-change`**: covers the material/style-change case, in two
  situations -- two different products with a genuinely coincident face
  but differing material/style (cross-product), and a single product's
  own face spanning more than one layer of its material stack, e.g. a
  section cut through a wall (intra-product). Off by default
  (`svg-use-mat-style-change-classification`), since it visibly
  reclassifies existing linework.
- Both are wired through the same Bonsai property/EPset/UI pattern
  already used for the existing classification system, and are fully
  additive: with both toggles off, output is byte-identical to current
  behaviour.

coplanar_options

### Duplicate-edge merge post-process (opt-in)

A new, `bpy`-free post-process (`svg_dedup.py`) that clusters same-class
`<path>` elements by a tolerant parametric collinear-overlap test and
merges/trims duplicates down to one edge per cluster, operating on the
emitted SVG paths themselves rather than the 3D classification logic.
Off by default; only takes effect alongside edge classification.

### Correctness & robustness fixes

A number of pre-existing and newly-introduced classification bugs were
found and fixed along the way -- sign-test/convexity errors, occlusion
and hidden-edge-restoration edge cases, degenerate-geometry guards
against real-world OCCT failures, and several narrower regressions
introduced and caught within this same branch's own review process.
These are individually small; the commit log is the source of truth for
what each one covers rather than repeating them here.

### Performance

Two independent hot-path fixes for larger drawings, both measured against
a real architectural project file rather than synthetic data:

- Scoped `is_genuinely_occluded()`'s ray-cast to a view-direction-aware
  shortlist instead of testing against every loaded object in the
  drawing -- **3.8-5.8x speedup** on axis-aligned views (isometric/oblique
  views benefit less, a known geometric limitation of the shortlist
  shape).
- Multi-threaded `restore_coincident_hidden_edges()`'s own per-candidate
  work, which became the next dominant cost once the above landed.

Both are verified byte-identical in output to the pre-optimization
baseline -- pure performance work, no behaviour change.

### Minor

- Edge classification / cross-coplanar / mat-style-change UI is now
  hidden when Freestyle linework mode is active, since the two systems
  aren't meant to be used together.

## Testing

- New dedicated regression suite covering the classification logic
  (`test_svg_edge_classification.py`) and the dedup post-process
  (`test_svg_dedup.py`), both plain `pytest`, no Blender required for the
  classification suite.
- Every new toggle verified byte-identical to current behaviour when off.
- Verified against real architectural project files (not just synthetic
  fixtures) at multiple points during development, including a full
  before/after A/B on real drawings for the performance work.

## Known limitations / out of scope

- Cross-coplanar/mat-style-change classification currently only
  understands `IfcMaterialLayerSetUsage` (layered elements like walls and
  slabs) -- `IfcMaterialProfileSetUsage` (structural members) and
  `IfcMaterialConstituentSet` aren't covered yet.
- Like the rest of the edge classification system, this requires planar
  faces and straight edges -- curved geometry is gracefully skipped, not
  a crash.

## AI-assisted development

This contribution was developed with the assistance of an AI coding tool
(Claude); every commit's own message discloses this in its body per
AGENTS.md.

## A word from the human

I developed the original edge classification work which was entirely done
with AI. Likewise, this is entirely produced by AI with my guidance and
understanding (and yes, sometimes my frustration when I couldn't get
Claude to understand my points.)

Lots of visual inspections by me, and I'm pretty happy with the robustness.
Like anything this sensitive though, there may be things I didn't catch. It
is entirely opt-in though, so it will not break your current options.

I'd really appreciate some feedback on whether this works well for people
and, if not, _small_ reproduction files that I can troubleshoot.

Note that you will need to update your projects css files to take full
advantage. I have included two sets of path.* css rules. One set are for
debug, one set are a first guess at what categories of lines should look
like. A professional set of css rules would be appreciated.

I've included the synthetic test files from @theoryshaw (hope he doesn't
mind) which includes a full set of the drawings so you can see better
than the above screenshot.

One last thing. I realise this is made on top of v0.8.0. I just want to get
it pushed. I will pull v0.9.0 and see if I can merge this PR into it. It
should be possible as the original edge classification was merged, and this
simply extends it.


sboddy:
Test file and renders: [7894_Join coplanar SVG projection cells using normal vector evaluation.zip](https://github.com/user-attachments/files/31229936/7894_Join.coplanar.SVG.projection.cells.using.normal.vector.evaluation.zip)

Based on the originals theoryshaw provides in his [comment](https://github.com/IfcOpenShell/IfcOpenShell/pull/7894#issuecomment-4180203341).

sboddy:
### Follow-on: `face-intersection` edge classification

Added a third sibling edge class alongside `cross-coplanar`/`mat-style-change`: **`face-intersection`**, for pairs of planar faces belonging to different products that genuinely cross each other in 3D (e.g. a diagonal brace piercing a wall) — as opposed to being coincident/parallel, which stays `cross-coplanar`'s job. Computed via `BRepAlgoAPI_Section`, trimmed to both faces' real boundaries, with a deterministic GlobalId-based tie-break so a genuine crossing shared between two products is only ever drawn once, never doubled.

Same conventions as the existing classes: classification on by default, rendering off by default (`svg-use-face-intersection-classification` / `svg-render-face-intersection-edges` / `svg-face-intersection-tolerance`), fully additive — output is unchanged with rendering off. Verified against the same real fixture files used elsewhere in this PR, plus new unit tests.

7412f5c7d1f999a3504f874925e242932f9984d1

sboddy:
### Follow-up: simplified the face-intersection toggle, reworked the classification UI

Two changes on top of the previous `face-intersection` comment:

- **Collapsed `svg-use-face-intersection-classification` / `svg-render-face-intersection-edges` into one boolean.** Unlike `cross-coplanar`/`mat-style-change`, this pass has no outline-trimming interaction for the two-state split to preserve — it's deliberately unranked/lowest-priority in the coincident-duplicate-edge dedup, so it never affects any other edge. The two-flag version was paying this pass's real per-face-pair cost by default for zero visible effect. Now mirrors `mat-style-change`'s own existing single-boolean pattern in this file. Default also moves to fully off (was classify-on/render-off) — the feature is inert until explicitly enabled.
- **Reworked the Camera panel's edge-classification UI**: fixed a leftover reference to a property (`join_coplanar_surfaces`) from an earlier coplanar-feature iteration that no longer exists — it was spamming the console with errors on every panel redraw. All edge-classification properties moved out of the main camera panel into their own collapsible sub-panel, grouped for readability (enable / render / post-process).
Selection_464


`d4e5a04fd9`

theoryshaw:
Heads-up @sboddy — #9493 also changes `merge_linework_and_add_metadata`, replacing the key-building block with a single call and adding two methods above `is_manifold`. Different lines from your `merge_duplicate_edges` hunk, and no behavioural overlap with the edge dedup.

I checked rather than guessed: your `operator.py` hunks apply cleanly on top of that branch, so whichever lands first the other should just need line-offset adjustment, not a real resolution. (`git apply --3way` does report failures for `src/ifcgeom/ConversionSettings.h` and the `SvgSerializer` files, but those paths don't exist on `v0.9.0` — that's base drift on this PR, nothing to do with #9493.)

Worth noting the two changes attack the same visible symptom from opposite ends: #9493 widens which cuts merge into one polygon, while this trims coincident edges that were each drawn correctly by their own product. They look complementary to me, but if you see a case where a merged polygon's boundary and a deduped edge disagree, I'd like to know.

Issue #9286, opened 2026-08-11
Snap crash: TypeError 'cannot unpack non-iterable NoneType' in ray_cast_by_proximity_2d when edges fall behind the camera plane


carlopav:
### Bug Description

Placing a window (or any product placed with the polyline/draw-occurrence tool) intermittently fails with:

```
File "bonsai\tool\raycast.py", line 442, in ray_cast_by_proximity_2d
    p0x, p0y = p0
TypeError: cannot unpack non-iterable NoneType object
```

It is not tied to a particular IFC file — it depends on where the camera sits, which is why it only happens sometimes. It occurs in a normal perspective 3D view (not camera view).

**To reproduce**

1. Open a model containing wireframe snap objects (meshes with no faces, curves — grid axes, annotations, axis lines).
2. Put the viewport in perspective and move the camera so that it sits *right next to / inside* such an object, i.e. part of the object is behind the camera plane.
3. Start `Add Occurrence` (e.g. a window) and move the mouse over the wall.
4. The modal operator aborts with the traceback above.

Full traceback:

```
File "bonsai\bim\module\model\workspace.py", line 1192, in hotkey_S_A
    return bpy.ops.bim.draw_occurrence("INVOKE_DEFAULT")
File "bonsai\bim\module\model\polyline.py", line 470, in invoke
    detected_snaps = tool.Snap.detect_snapping_points(context, event, self.objs_2d_bbox, self.tool_state)
File "bonsai\tool\snap.py", line 394, in detect_snapping_points
    closest_snaps = tool.Raycast.ray_cast_and_get_closest_to_camera_snaps(context, event, objs_to_raycast)
File "bonsai\tool\raycast.py", line 884, in ray_cast_and_get_closest_to_camera_snaps
    hit_obj, hit = cls.process_wireframe_snap_obj(context, event, snap_obj, ray_origin, closest_snaps)
File "bonsai\tool\raycast.py", line 829, in process_wireframe_snap_obj
    snap_points = tool.Raycast.ray_cast_by_proximity_2d(context, event, snap_obj)
File "bonsai\tool\raycast.py", line 442, in ray_cast_by_proximity_2d
    p0x, p0y = p0
TypeError: cannot unpack non-iterable NoneType object
```

### Root cause

[...]

Issue #9206, opened 2026-07-31
Bonsai error


RobertAWgt:
### Bug Description

```
os: Windows
os_version: 10.0.26200
python_version: 3.13.13
architecture: ('64bit', 'WindowsPE')
machine: AMD64
processor: Intel64 Family 6 Model 142 Stepping 10, GenuineIntel
blender_version: 5.2.0 LTS
bonsai_version: 0.8.5
bonsai_commit_hash: None
bonsai_commit_date: None
bonsai_git_branch: None
last_actions: 
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
# operator: BIM_OT_draw_occurrence
last_error: Traceback (most recent call last):
  File "C:\Users\rober\AppData\Roaming\Blender Foundation\Blender\5.2\extensions\.local\lib\python3.13\site-packages\bonsai\bim\ifc.py", line 527, in execute_ifc_operator
    result = getattr(operator, "_modal")(context, event)
  File "C:\Users\rober\AppData\Roaming\Blender Foundation\Blender\5.2\extensions\.local\lib\python3.13\site-packages\bonsai\bim\module\model\product.py", line 236, in _modal
    self.handle_mouse_move(context, event)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "C:\Users\rober\AppData\Roaming\Blender Foundation\Blender\5.2\extensions\.local\lib\python3.13\site-packages\bonsai\bim\module\model\polyline.py", line 411, in handle_mouse_move
[...]

PR #9337, opened 2026-08-19
Snap gpu detection

brunoperdigao:
The PR changes the way objects are detected by the snapping system. It is inspired by Cad Transform and uses shaders and the GPU module to detect objects. Now it doesn't rely on the custom BVH that was causing issues and it should also improve performance in large files.


brunoperdigao:
This should solve #8187, #9003, #8008, #9286, #8207, #8119, #9206

Issue #4781, opened 2024-05-31
Text wrap doesn't seem to work with schedules that have custom css


theoryshaw:
Text wrap doesn't seem to work with schedules that have custom css

file: [GENERAL NOTES.ods](https://github.com/user-attachments/files/15517579/GENERAL.NOTES.ods)

```
 :root {
    --font-size-pt: 12;
    --font-size-px: 4.13;
    --font-width: 0.45;
}

text, tspan { /* 2.5mm */ fill: black; stroke: none; font-family: 'OpenGost Type B TT', 'DejaVu Sans Condensed', 'Liberation Sans', 'Arial Narrow', 'Arial'; }


```

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/aae95d9a-08f9-4afd-815e-dcf67df13dc5)

```
:root {
    --font-size-pt: 12;
    --font-size-px: 4.13;
    --font-width: 0.45;
}

text, tspan { /* 2.5mm */ fill: #2C5573; stroke: none; font-family: 'Century Gothic', 'OpenGost Type B TT', 'DejaVu Sans Condensed', 'Liberation Sans', 'Arial Narrow', 'Arial';}

```

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/f9e80e3f-e802-40f5-8511-45b2729e979b)

Moult:
This is because there is no way that I know of right now to detect the width of a font to know when it should break to a new line. Maybe there is some way to do it that someone might know who knows more about fonts and SVG, or some smart way to autodetect it but for now, you can edit your css to define font-width to something wider (as Century Gothic is pretty wide).

Issue #3586, opened 2023-08-15
Place the schedule on the sheet, but later change the column widths, the schedule gets truncated. 


theoryshaw:
If you place the schedule on the sheet, but later change the column widths, the schedule gets truncated. 

![image](https://github.com/IfcOpenShell/IfcOpenShell/assets/507113/5615f7cd-707e-47d1-908a-5a86437ce897)

theoryshaw:
probably related: https://github.com/IfcOpenShell/IfcOpenShell/issues/5382

theoryshaw:
https://github.com/IfcOpenShell/IfcOpenShell/pull/7971 proposed

PR #9350, opened 2026-08-23
Bonsai: draw the cell borders a schedule's spreadsheet declares

carlopav:
`schedule_ods` already collects every cell style attribute from the spreadsheet — background, font size, weight, style, color, wrap — and uses them. Borders are collected too, and discarded: every cell is drawn as a rectangle carrying one hardcoded `class_="border"` whose stroke comes from the stylesheet. `schedule_xlsx` does the same from openpyxl. A thick outline with thin inner rules, the usual look of an architectural schedule, cannot reach the drawing no matter what is set in LibreOffice or Excel.

### What changed

The inline rectangle in both branches becomes `draw_cell()`, which takes a background color and the borders **already resolved**, so it is independent of the source format. Each branch translates its own:

- `get_cell_borders()` / `parse_border()` read `.ods` style attributes;
- `get_xlsx_cell_borders()` / `parse_xlsx_border()` read openpyxl `Side` objects.

A cell that declares no border keeps the `.border` class and draws exactly as before. One that declares them gets a line per side, which a single rectangle stroke cannot express.

Width, style and color are read in any order, per the CSS `border` shorthand grammar (`<width> || <style> || <color>`), rather than assuming the order LibreOffice happens to emit. `dotted` and `dashed` become a dash pattern scaled by the line width; `double` and the relief styles draw plain.

`XLSX_BORDERS` converts Excel's fourteen named weights into widths and dash patterns. Excel names a weight instead of measuring one, so that table is a choice rather than a reading of the file, and is commented as such.

### One deliberate behaviour change

A cell that declares `border: none` now draws **no** line, where before it took the stylesheet's. This is intentional and is half the point of the change: a sheet that says a side has no border should not be given one.

Cells that declare nothing about borders are unaffected — that is the compatibility guarantee, and it covers every schedule IfcCsv writes, since it emits no cell styles at all.

### Left as it is

Adjacent cells each stroke the edge they share, so where two neighbours declare it differently the later one wins. The rectangles overlapped the same way before, with the same stroke on both sides. Spreadsheet applications resolve such conflicts by precedence, which this does not attempt.

### Testing

`src/bonsai/test/bim/module/drawing/test_scheduler.py` — 20 tests over both translators and the drawing: the untouched-cell guarantee, the `border: none` change, per-side overrides of the shorthand, token order, dash patterns, and the openpyxl side. One trap they pin down: asked for the ARGB of a theme color, openpyxl returns its own complaint *as a string* instead of raising, so the length of the value is what distinguishes a real ARGB from it.

`black` and `ruff` clean on both files.

### Notes

- Touches the same block of `schedule_ods` as #8537 (hidden columns), a few lines apart. Whichever lands second wants a trivial rebase; this change keeps the `background_color` local in place to keep the overlap as small as possible.
- Adjacent but not addressed here: #4781 (text wrap with custom CSS), #5221 (font lost when a schedule is placed on a sheet), #3586 (truncation after column widths change).
- Draft: opened for review of the approach before asking for a merge.

### AI disclosure

Per AGENTS.md: this contribution was written with the assistance of an AI coding tool, in whole — both the change to `scheduler.py` and the new test file, which carries the required header comment.

### DIFF
--- a/src/bonsai/bonsai/bim/module/drawing/scheduler.py
+++ b/src/bonsai/bonsai/bim/module/drawing/scheduler.py
@@ -21,6 +21,7 @@ import re
 import string
 from pathlib import Path
 from textwrap import wrap
+from typing import Union
 
 import openpyxl
 import openpyxl.cell  # Unnecessary, bug in typeshed.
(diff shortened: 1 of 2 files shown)


carlopav:
Superseded by #9351: a force-push left this pull request pointing at an orphaned commit, and it could not be reopened. #9351 carries the same change from the same branch, with the review comments already applied.

PR #9359, opened 2026-08-25
Modifications to build C++ with Visual Studio 2026 and the v145 toolset.

RickBrice:
I had to make some tweaks to get the dependencies and the C++ libraries to build with Visual Studio 2026. Also, my computer would crash with out of resource errors if I built the solution file with ccache, so I added back the -DUSE_CCACHE=OFF option.

I don't want to mess up anything you guys have done recently, so this is a PR instead of a direct commit.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -178,7 +178,7 @@ if((BUILD_CONVERT OR BUILD_GEOMSERVER OR BUILD_IFCPYTHON) AND(NOT BUILD_IFCGEOM)
     set(BUILD_IFCGEOM ON)
 endif()
 
-option(USE_CCACHE "Use ccache as a compiler launcher if it is found" ON)
+option(USE_CCACHE "Use ccache as a compiler launcher if it is found" OFF)
 if(USE_CCACHE)
     find_program(CCACHE_FOUND ccache)
     if(CCACHE_FOUND)
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -316,12 +316,25 @@ if (WITH_ROCKSDB)
     set(SWIG_DEFINES ${SWIG_DEFINES} -DIFOPSH_WITH_ROCKSDB)
     # See https://github.com/facebook/rocksdb/issues/981.
     if(TARGET RocksDB::rocksdb)
-        target_link_libraries(IFCOPENSHELL_RocksDB INTERFACE RocksDB::rocksdb)
+        set(IFCOPENSHELL_ROCKSDB_IMPORTED_TARGET RocksDB::rocksdb)
     elseif(TARGET RocksDB::rocksdb-shared)
-        target_link_libraries(IFCOPENSHELL_RocksDB INTERFACE RocksDB::rocksdb-shared)
+        set(IFCOPENSHELL_ROCKSDB_IMPORTED_TARGET RocksDB::rocksdb-shared)
     else()
         message(FATAL_ERROR "RocksDB found but neither RocksDB::rocksdb nor RocksDB::rocksdb-shared target exists")
     endif()
+    # Our win/build-deps.cmd builds RocksDB separately per Debug/Release config into the
+    # same install prefix, so the imported target only ever has DEBUG and RELEASE listed in
+    # IMPORTED_CONFIGURATIONS. On a multi-config generator (Visual Studio), CMake maps any
+    # unmatched build config to the *first* entry of that list, which happens to be DEBUG
+    # (RocksDBTargets-debug.cmake sorts before RocksDBTargets-release.cmake). Without an
+    # explicit mapping, RelWithDebInfo and MinSizeRel builds would end up linking the
+    # /MDd-flavored rocksdb_d.lib into an /MD (NDEBUG) binary, causing a CRT/runtime-library
+    # mismatch that depends on nothing but that alphabetical ordering.
+    set_target_properties(${IFCOPENSHELL_ROCKSDB_IMPORTED_TARGET} PROPERTIES
+        MAP_IMPORTED_CONFIG_RELWITHDEBINFO RELEASE
+        MAP_IMPORTED_CONFIG_MINSIZEREL RELEASE
+    )
+    target_link_libraries(IFCOPENSHELL_RocksDB INTERFACE ${IFCOPENSHELL_ROCKSDB_IMPORTED_TARGET})
 
     if (WITH_ZSTD)
         # @todo do we actually need the zstd include dir or rather just pass
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -178,26 +178,31 @@ if((BUILD_CONVERT OR BUILD_GEOMSERVER OR BUILD_IFCPYTHON) AND(NOT BUILD_IFCGEOM)
     set(BUILD_IFCGEOM ON)
 endif()
 
-find_program(CCACHE_FOUND ccache)
-if(CCACHE_FOUND)
-    message(STATUS "`ccache` is found, using it as a compiler launcher.")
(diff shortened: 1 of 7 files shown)


src/ifcparse/parse.cpp
```diff
@@ -1651,26 +1651,38 @@ express::base::set_attribute_value(size_t i, const T& t) {
...
 
-    data()->set_attribute_value(i, t);
-    auto new_attribute = get_attribute_value(i);
-
-    // Register inverse indices in file
-    if constexpr (std::is_same_v<T, express::base> || std::is_same_v<T, std::vector<express::base>> || std::is_same_v<T, std::vector<std::vector<express::base>>>) {
-        register_inverse_visitor visitor(*file(), *this);
-        apply_individual_instance_visitor(new_attribute, (int)i).apply(visitor);
+    // A null/empty single instance attribute (e.g. an omitted optional like
+    // OwnerHistory, ObjectPlacement or Representation) must not be persisted
```
aothms, line 1655: ... but it must be persisted probably as a blank attribute if that value previously got assigned something. So I agree with this change, but I think there needs to be a 

```c++
... } else {
    data()->set_attribute_value(i, blank{});
}
```

win/patches/OpenCOLLADA_CommonFWriteBufferFlusher_tr1.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/common/libBuffer/include/CommonFWriteBufferFlusher.h b/common/libBuffer/include/CommonFWriteBufferFlusher.h
```
aothms, line 1: I would just patch this blindly to `std::unordered_map` without the `#if defined...` there is no reason to be conservative here.

cmake/CMakeLists.txt
```diff
@@ -178,26 +178,31 @@ if((BUILD_CONVERT OR BUILD_GEOMSERVER OR BUILD_IFCPYTHON) AND(NOT BUILD_IFCGEOM)
...
-        if(CMAKE_GENERATOR MATCHES "Visual Studio")
-            file(COPY_FILE
-                ${CCACHE_FOUND} ${CMAKE_BINARY_DIR}/cl.exe
-                ONLY_IF_DIFFERENT)
-            set(CMAKE_VS_GLOBALS
-                "CLToolExe=cl.exe"
-                "CLToolPath=${CMAKE_BINARY_DIR}"
-                "UseMultiToolTask=true"
-            )
+option(USE_CCACHE "Use ccache as a compiler launcher if it is found" ON)
```
aothms:
```suggestion
option(USE_CCACHE "Use ccache as a compiler launcher if it is found" OFF)
```

I'm also fine with OFF as a default so that you do not need the parameter parsing in .bat so much

aothms (commented):
Thanks for front running this @RickBrice

RickBrice:
@aothms boost 1.86 isn't quite compatible with visual studio 2026 & the 145 toolset. Any objections to bumping to the last boost 1.92? Everything seems to work fine with it.

aothms:
> @aothms boost 1.86 isn't quite compatible with visual studio 2026 & the 145 toolset. Any objections to bumping to the last boost 1.92? Everything seems to work fine with it.

No objections

RickBrice:
@aothms I made the changes you suggested, fixed a bug in alignment_helper.cpp, and bumped to boost 1.92.

RickBrice:
@aothms in my c++ application, there is a significant slowdown between v0.8.0 and v0.9.0. It looks like it is related to building the inverse relationships and some sorting. If you are interested, we can set up a call and I can show you what's going on.

aothms:
My uneducated guess would be that it's either:

- shared_ptr/weak_ptr machinery on data access making things slow - but it looks like this is temporarily disabled see `IFOPSH_SAFE_INSTANCE` -> mostly intended for python and should be disabled anyway when doing C++ work
- granular calls between plug-in dlls now that prevent optimizations or inlining
- something silly that happened

Is it possible (with the help of AI) to produce something self contained. I prefer to have a look before jumping on a call.

RickBrice:
I'll get a self-contained example put together as soon as I can (I'm almost out of AI tokens for the week).

In the meantime, this example files loads into Blender/Bonsai very quickly based on v0.8.0 and very slowly based on v0.9.0. If I regenerate the model without the reinforcing, performance is better. When I use my C++ application to generate this model backed by v0.8.0 its quick and very slow backed by v0.9.0. When I break into the running exec, it seems to be taking all of its time sorting while creating the inverse relationships related to the reinforcing. Admittedly, the reinforcing model isn't very efficient, but that's not really the issue.

[PGSuper_Import_Model.ifc.txt](https://github.com/user-attachments/files/31520739/PGSuper_Import_Model.ifc.txt)

RickBrice:
I got Claude to generate a stand-alone application that is very similar to my actual application. Claude's analysis is:

~~~
Root cause: an O(n²) build pattern, made severe by an IfcOpenShell 0.9 regression.

Every rebar occurrence calls hierarchy_helper::addRelatedObject<T>() twice (bar→assembly IfcRelAggregates, bar→type IfcRelDefinesByType). Removing just those two calls makes build time flat (~5 ms/beam instead of ~30 s / ~8 min total), so the entire cost is there.

Each addRelatedObject call does two things that scale badly:

1. instances_by_type<T>() — a full scan of every instance in the model to locate the existing relationship. In 0.8 this was indexed (≈O(1)); in 0.9 it's a linear walk. Called once per bar while the model keeps growing → O(bars × instances).
2. Rewrites the relationship's entire member list. It reads the whole RelatedObjects collection, appends one element, and writes it all back. Adding k members to one relationship is O(k²).

Both terms are quadratic in the number of bars. The helper is built to be called once per member and re-derives everything from scratch each time — fine for a handful of members, pathological for thousands.

What made it worse in this test: giving all beams one shared IfcReinforcingBarType funneled all ~7,500 bars into a single IfcRelDefinesByType, so term (2) blew up (k→7,500) — ~16× slower than per-beam types. At the largest sizes, repeatedly copying the multi-thousand-element handle vector also hit a memory-bandwidth cliff (beams 14–15).

The fix: don't call addRelatedObject per member. Accumulate all members for a relationship in a local vector and set RelatedObjects once, and/or hold the relationship handle instead of re-finding it by scanning. That removes both the per-call scan and the repeated list rewrite, taking build time from quadratic to linear.
~~~

I thinking there are a few things I can do to optimize my implementation. 

[v9Test.zip](https://github.com/user-attachments/files/31534008/v9Test.zip)

aothms:
Thanks @RickBrice 

> 1. instances_by_type<T>() — a full scan of every instance in the model to locate the existing relationship. In 0.8 this was indexed (≈O(1)); in 0.9 it's a linear walk. Called once per bar while the model keeps growing → O(bars × instances).

But I suspect the above is not very precise, i.e I'm quite sure the quadratic nature did not change but maybe a linear factor somewhere - or there would be a severe thing I'm overlooking. So now I'm struggling to understand whether this is a real regression or just a fundamental performance limitation that was just made a little bit worse. We have the same in Python btw.

> rel.RelatedObjects += (ob,)

looks harmless, but what it does essenstially is

> obs = rel.RelatedObjects + (ob,)
> rel.RelatedObjects = [] # clear all inverses
> rel.RelatedObjects = obs # repopulate all inverses

This is really expensive and indeed quadratic when appending in a loop and indeed in the python api we have worked around exactly by accumulating objects before writing them out in a relationship.

For a long time I thought the proper fix is with our own custom aggregation type so that we can call:

> rel.RelatedObjects().push_back(ob)

But assuming the main quadratic performance impacts do indeed come from re-registering the inverses there is actually a really straightforward fix in how set_attribute_value() is currently implemented because we can do the set differences there and only perform the inverses that actually changed between the old and the new value.

See #9371

PR #9367, opened 2026-08-27
Fix data races in schema singleton init and logger causing crashes

sboddy:
## Summary
- Every generated schema's `get_schema()` (`schema_class.py`'s template, emitted into all 13 `src/ifcparse/schemas/*-schema.cpp`) lazily built its singleton with an unguarded `if (!schema) schema.reset(new X())`. Concurrent geometry-conversion worker threads hit this the first time an instance is formatted for a log message (`express::base::to_string()`'s header check), racing on the non-atomic `unique_ptr::reset()` and corrupting the heap.
- `ifcopenshell::logger::current_product()` had the same class of bug: it wrote a shared `express::base` member with no locking, called by every worker thread via `set_product()` on nearly every element — a much higher-frequency race than the schema one, and the bigger contributor to the crash in practice.
- Both are now guarded by a mutex.

## Test plan
- [x] Headless multi-threaded repro against a real-world file that reliably crashed Blender on load: 0 crashes in 70 stress runs post-fix (16 and 32 iterator threads), vs. ~50-70% crash rate pre-fix at 16 threads
- [x] All 13 regenerated/hand-patched `*-schema.cpp` files compile cleanly
- [ ] CI build


sboddy:
Segfaults trying to load a file into Bonsai with the default hybrid kernel. It could be worked around by using basic CGAL. After patch the segfault is fixed.

PR #9368, opened 2026-08-27
svg_serializer: skip elements that fail geometry conversion

sboddy:
## Summary
- A single element's geometry can fail deep inside OCCT (e.g. a degenerate surface) without that being representative of the rest of the file. Previously this propagated all the way through SWIG as an unhelpful "An unknown error occurred", aborting the whole drawing.
- `svg_serializer::write()` now catches the failure per-element, logs which element and why (including a temporary `std::cerr` mirror, since nothing currently wires up `ifcopenshell::logger`'s output for this code path), and skips just that element so the rest of the drawing still gets produced.
- Also hardens `convert_wire_to_faces()`'s face-area heuristic in `wire_utils.cpp` against the same class of rare OCCT failure — area there is only used to drop slivers from self-intersection splitting, so it now falls back to 0 rather than losing the whole face.

## Test plan
- [x] Compiles cleanly
- [x] Confirmed the catch behaves as intended against a real-world file where OCCT throws `GeomAdaptor_Surface::UContinuity`
- [ ] CI build

### DIFF
--- a/src/ifcgeom/kernels/opencascade/wire_utils.cpp
+++ b/src/ifcgeom/kernels/opencascade/wire_utils.cpp
@@ -937,7 +937,18 @@ bool ifcopenshell::geom::util::convert_wire_to_faces(const TopoDS_Wire& w, TopoD
 		}
 
 		TopoDS_Face face = mf.Face();
-		const double m = face_area(face);
+		double m;
+		try {
+			m = face_area(face);
+		} catch (const Standard_Failure& e) {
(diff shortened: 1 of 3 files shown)


sboddy:
Note that for me, drawings in Bonsai still do not work (all elements are currently skipped). I'm going to dig into why that is next, but it appears to be something in OCCT HLR itself.

sboddy:
Got drawings working. It's dependant on #9341.

PR #9371, opened 2026-08-28
Address quadratic inverse handling

aothms:


RickBrice:
I don't fully understand the internal data structures and how inverses are generated but here are my observations using this code.

- My export is noticeably faster, on par with the v0.8.0 implementation, but I don't have hard numbers to compare
- I get better performance on my C++ IFC export implementation if I bundle up all the RelatedObjects with their RelatingObject (or whatever the attributes are called) in a map, and then create the IfcRel directly one time, instead of calling file.addRelatedObject<>.
- On the import side, using Blender/Bonsai, it is glacially slow using the v0.9.0 implementation at the fix from this branch. It was just as slow without this change also. The example bridge model I shared loads in less than a minute when backed by v0.8 and after 30 minutes it's still loading when backed by v0.9

aothms:
Thanks for testing @RickBrice

1 - that's good to have confirmed

2 - that's expected, the major non-linear behaviour of inverse handling is alleviated a bit, but the non-linear storage copies etc. are still suboptimal without a proper append

3 - I cannot reproduce this using IfcConvert (can you?) and I had assumed Python would be faster because major extra indirections have been eliminated

RickBrice:
@aothms I can't reproduce it with IfcConvert either. It took 1m 6sec for DAE and 1m 17sec for OBJ output.

Here is another data point. When I open my bridge file in Bonsai, I picked the "Enable Advanced Mode" option. 
The Hybrid CGAL-OCC library load extremely slowly (10 minutes). The OpenCASCADE, CGAL, and CGAL Simple libraries all load quickly (about 1 minute).

image

aothms:
That's good to know. Based on these observations I think we have synthesised two options now. My preference is to merge both:

- I prefer to get C++ and python more alike, so having Kernel as a concept in python make sense
- To minimize code changes while still having ok perf, it probably makes sense to cache the filesystem look ups for an individual create_shape() call

I will just merge both to get everything unblocked and then we can always refine later in the future.

Issue #3072, opened 2023-04-29
Drawing generation speed optimisation


Moult:
Just to start the conversation, drawing generation is awesome, but still quite slow. For example, in the Revit sample model, a roof plan (which is simpler than other plans) takes nearly 4 minutes(!). After hiding two problematic high polycount meshes, I've got the time down to 25s on my machine, but still, 25 seconds is pretty long for what the user perceives as being a pretty empty plan.

Note that I'm only feeding in objects within the camera range (e.g. the top floor of the building, not the lower 3 floors) into the geometry iterator / drawing serialiser. What types of strategies might be done to make things faster?

Here's what Mr. ChatGPT has to say:

> I am building software that uses OpenCascade to create a HLR vector drawing of a building. However, the processing time is quite slow. What strategies might be used to speed it up?
> 
> Optimizing performance in a software that uses OpenCascade for creating hidden line removal (HLR) vector drawings can be a challenge. However, there are several strategies that you can employ to speed up the processing time:
> 
> 1. Optimize geometry: Simplify your building model by reducing the number of faces, edges, and vertices. Remove any unnecessary or overly detailed features, and merge or eliminate coplanar faces. This will reduce the computation complexity for HLR algorithms.
> 1. Use Level of Detail (LOD): Implement different levels of detail for your building model, where each level is a simplified version of the original. Select the appropriate LOD based on the zoom level or distance from the camera, which will reduce the amount of data being processed in the HLR algorithm.
> 1. Optimize the HLR algorithm: Profile and analyze the performance of the HLR algorithm you're using to identify any bottlenecks. Once identified, optimize those areas of the code by employing efficient data structures, parallel processing, or better algorithms.
> 1. Parallel processing: Utilize multi-threading or GPU acceleration to distribute the computation tasks across multiple cores or processing units. OpenCascade supports parallel processing, which can significantly speed up your HLR operations.
> 1. Caching and reusing data: Cache intermediate results when possible, and reuse them when generating HLR images from similar viewpoints. This can save time by avoiding redundant calculations.
> 1. Adaptive HLR: Implement an adaptive HLR algorithm that adjusts the level of detail based on the size of the features in the view. This can help reduce the processing time by avoiding unnecessary calculations for small, insignificant features.
> 1. Occlusion culling: Use occlusion culling techniques to exclude objects or parts of objects that are not visible from the current viewpoint. This can help reduce the number of objects being processed in the HLR algorithm.
[...]

PR #9378, opened 2026-08-29
Release per-storey HLR data after drawing

P4A-KLFSBK:
The SVG serializer emits one drawing per building storey, iterating `storey_hlr` in `finalize()`. Each entry is an `hlr_engine` holding that storey's shapes in `items_`, its `large_ortho_faces_`, its classified edge shapes and an OCCT HLR engine — but entries are never released, so peak memory is the projection input for the entire model even though the drawings are produced one at a time.

`storey_hlr` has three uses: the insert in `write(const geometry_data&)`, the lookup at the top of `draw_hlr`, and this loop. Nothing reads it afterwards, so each entry can be erased once its drawing has been emitted. Peak then tracks one storey plus the accumulated output paths, which are much smaller than the input BReps.

This frees data that has no remaining reader, so the emitted SVG is unchanged.

Not formatted with `clang-format`: `svg_serializer.cpp` is tab-indented throughout while `.clang-format` specifies LLVM/4-space, so running it would rewrite the whole file. The change follows the surrounding style instead.

---

This change was generated with the assistance of an AI coding tool. It is the result of static reading of the serializer; I have not built or profiled it.

### DIFF
--- a/src/serializers/svg_serializer.cpp
+++ b/src/serializers/svg_serializer.cpp
@@ -2430,8 +2430,12 @@ void svg_serializer::finalize() {
 		addTextAnnotations(p.first);
 	}
 
-	for (auto& p : storey_hlr) {
-		draw_hlr(drawing_metadata[{p.first, ""}].pln_3d, { p.first, "" });
+	// Each entry holds that storey's shapes and its HLR engine, which are not
+	// referenced again once the drawing has been emitted. Releasing them here
+	// keeps peak memory at one storey rather than the whole model.
+	for (auto it = storey_hlr.begin(); it != storey_hlr.end();) {
+		draw_hlr(drawing_metadata[{it->first, ""}].pln_3d, { it->first, "" });
+		it = storey_hlr.erase(it);
 	}
 
 	auto m = resize();


P4A-KLFSBK:
Context for this change, and the rest of what I found reading the serializer, is in #3072 (drawing generation speed). This PR is the memory half of it and stands on its own.

BIMvoice:
Checked the "nothing reads the map after the loop" claim against the current `v0.9.0` tip, since that is the only thing that makes erasing during iteration safe here.

`storey_hlr` is touched in exactly three places: inserted at 1625-1627 inside `write()`, read at 2188 inside `draw_hlr()`, and iterated at 2433 in `finalize()`. The only `draw_hlr` call after that loop is the section pass at 2530, which passes `express::base{}` as the key, so `drawing_name.first` is false and 2188 takes the `*hlr` branch rather than the map lookup. So the map really is dead once the 2433 loop has consumed it, and `it = storey_hlr.erase(it)` is the correct idiom for a `std::map`.

Not measured, same as the PR says, but the reasoning holds against the source.

P4A-KLFSBK:
I have now built and measured this, so the "not built or profiled" caveat at the end of the description no longer applies.

**Setup.** 0.9.0alpha0, built shared, linux/amd64, 9 GB cgroup limit, `-j 3`. Storey plans with `--plan --model --section-height-from-storeys --section-height 1.2 --svg-project --svg-segment-projection --svg-prefilter --exclude entities IfcOpeningElement`. Peak is `VmHWM` from `/proc/<pid>/status`, sampled at 4 Hz on the IfcConvert process itself; the geometry column is `VmRSS` at the `Done writing geometry` marker, i.e. immediately before `finalize()`.

**Exact hidden-line removal.** Four models:

| model | geometry | peak before | peak after | |
|---|---|---|---|---|
| 7 MB, 19 storeys | 262 MB | 1181 MB | 571 MB | 0.48x |
| 5 MB, 18 storeys | 231 MB | 8709 MB, OOM | 1611 MB | completes |
| 47 MB, 6 storeys (Schependomlaan) | 830 MB | 8758 MB, OOM | 7569 MB | completes |
| 47 MB, 16 storeys | 2470 MB | 8723 MB, OOM | 8748 MB | still OOM |

Two models that could not be converted at all now convert, with every storey drawn.

**With `--svg-poly`** the same four go 475 -> 387 MB, 633 -> 357 MB, 1379 -> 1196 MB, and 5019 -> 4742 MB. The last one exceeds the 90 minute limit I allowed it either way, so its peak is a plateau rather than a completed run.

**Control.** With projection off, `storey_hlr` is never populated, so the change must not move anything: 236 -> 244 MB, 142 -> 141 MB, 657 -> 673 MB, 901 -> 902 MB. Within noise, as expected.

**It does cost wall time with the exact algorithm.** 154 -> 192 s on the first model, about +25%, which is well above this bench's +/-1% reproducibility on peak RSS. Releasing each storey's BReps as its drawing is emitted is real destructor work, so that seems inherent rather than incidental. With `--svg-poly` the cost is not measurable (103 -> 106 s, 59 -> 56 s, 105 -> 104 s), presumably because the polyhedral data is cheaper to destroy.

**One thing I did not verify:** byte-level output equality. At `-j 3` the SVG is not deterministic here, the same binary and flags produced different SHA256s across runs, so I compared storey counts and object counts instead, which were identical in every pair. I can redo the comparison at `-j 1` if that would be useful.

These measurements were produced with the assistance of an AI coding tool.

PR #9379, opened 2026-08-29
Fix double free of section HLR state

P4A-KLFSBK:
In the deferred section loop in `svg_serializer::finalize()`, `hlr` is allocated only when both `use_hlr` and `pln` are set, but `delete hlr` runs at the end of every iteration and the pointer is never reset.

### Reachability

Narrower than "double free" on its own suggests, so worth stating plainly.

`deferred_section_data_` only ever receives `vertical_section` entries — `horizontal_plan` is pushed to `section_data_` instead — so the `sd.which() != 2` paths cannot be reached through this loop as it is populated today. The path that *is* open is `addDrawing(..., include_projection=false)`: `use_hlr` becomes false, nothing is allocated, and the iteration still runs `delete hlr`.

`hlr` is initialised to `nullptr` in the constructor (`svg_serializer.h:665`), so the first such iteration is a harmless no-op delete. The double free needs two: a projected section allocates and deletes, leaving the pointer dangling, and a following non-projected entry deletes it a second time.

Nothing in Bonsai calls `addDrawing` today — the only occurrence in the tree is a commented example in `ifcopenshell/draw.py` — so this is latent through the public C++ API rather than a crash users are currently hitting.

### The guard mismatch

`use_hlr` is initialised `true` and only assigned inside `if (sd.which() == 2)`, so a variant holding something other than a `vertical_section` would reach `boost::get<vertical_section>` on the wrong alternative, and a vertical section with a null `pln` would reach `*hlr` on a null pointer in `draw_hlr`. Neither is reachable through `deferred_section_data_` as populated today, but the existing `// @todo do we have always have pln here?` sits directly above the allocation, so guarding the call the way the allocation is guarded removes the trap rather than leaving it to the caller.

This guards the call the way the allocation is guarded, and nulls the pointer after `delete`.

`pln` already points at `section.plane`, so `pln->Position()` is the same expression the removed lines built — behaviour on the path that previously worked is unchanged.

Not formatted with `clang-format`: `svg_serializer.cpp` is tab-indented throughout while `.clang-format` specifies LLVM/4-space, so running it would rewrite the whole file. The change follows the surrounding style instead.

---

This change was generated with the assistance of an AI coding tool. It is the result of static reading of the serializer; I have not built or profiled it.

### DIFF
--- a/src/serializers/svg_serializer.cpp
+++ b/src/serializers/svg_serializer.cpp
@@ -2523,11 +2523,8 @@ void svg_serializer::finalize() {
 				write(e);
 			}
 
-			if (use_hlr) {
-				const auto& section = boost::get<vertical_section>(sd);
-				const auto& ax = section.plane.Position();
-
-				draw_hlr(ax, { express::base{}, drawing_name });
+			if (use_hlr && pln) {
+				draw_hlr(pln->Position(), { express::base{}, drawing_name });
 			}
 
 			addTextAnnotations({express::base{}, drawing_name});
@@ -2588,6 +2585,7 @@ void svg_serializer::finalize() {
 			resetScale();
 
 			delete hlr;
+			hlr = nullptr;
 		}
 	}
 


P4A-KLFSBK:
Found while reading the serializer for #3072, though this one is a plain lifetime bug rather than a performance change and is independent of anything discussed there.

BIMvoice:
Read this against the current `v0.9.0` tip (`1a03fd7df1`) rather than from the diff alone, since I could not build it. The guard mismatch is real; the reachability is narrower than the title suggests, which is worth stating so the reviewer knows what they are looking at.

**What is definitely wrong.** In `finalize()` the loop over `deferred_section_data_` allocates `hlr` only under `if (use_hlr && pln)` (line 2518) but runs `delete hlr` unconditionally at 2590, and the member is declared bare as `hlr_engine* hlr;` at `svg_serializer.h:614` with no initialiser. So any iteration that does not allocate still deletes whatever the pointer holds.

**How reachable it is today.** Horizontal entries never go into `deferred_section_data_`: `horizontal_plan` is pushed to `section_data_` (lines 1132, 2844, 2870, 2874), while `deferred_section_data_` only ever receives `vertical_section` (the elevations at 2480-2495, always `with_projection = true`, and `addDrawing()` at header 798). So the `which() != 2` branch, the `boost::get` on the wrong alternative, and the double free via a horizontal entry cannot happen from Bonsai as it stands.

The path that *is* open is `addDrawing(..., include_projection=false)`: `use_hlr` becomes false, nothing is allocated, and the first such iteration deletes an **uninitialised** pointer, or a dangling one if a projected section came first. Nothing in Bonsai calls `addDrawing` today (the only occurrence in the tree is a commented example in `ifcopenshell/draw.py`), so this is latent through the public API rather than a crash people are hitting.

**On the fix.** `hlr = nullptr` after the delete is right and makes `delete` on later iterations safe. It does not cover the first iteration, because the member starts uninitialised. I would add `hlr_engine* hlr = nullptr;` in the header alongside this, then every path is safe: `delete nullptr` is a no-op.

Net: correct hardening of real UB, worth merging, and the description would be more accurate as "latent via `addDrawing(..., false)`" than as a live double free.

P4A-KLFSBK:
Thanks for reading it against the tip rather than the diff — the reachability analysis is the useful part and I have reworked the description around it.

One correction on the header suggestion, though. `hlr` is not uninitialised. `svg_serializer` has a single constructor and its member-init list already sets it, at `svg_serializer.h:665` between `radii_begin(0)` and `namespace_prefix_("data-")`:

```cpp
		, radii_begin(0)
		, hlr(nullptr)
		, namespace_prefix_("data-")
```

The complete set of writes to the member is that initialiser, the guarded `hlr = new hlr_engine(...)` at `svg_serializer.cpp:2518`, and the `delete hlr; hlr = nullptr;` this PR completes at 2587-2588. So the first non-allocating iteration deletes `nullptr` rather than garbage. The case that does bite pre-patch is the *second* one: a projected section allocates and deletes, leaving the pointer dangling, and a following `addDrawing(..., include_projection=false)` iteration deletes it again. That is the double free, and nulling after `delete` is what closes it.

`hlr_engine* hlr = nullptr;` at the declaration would be defensive if a second constructor were ever added, but as the class stands it is redundant and would pull the header into the diff for no change in behaviour, so I have left it out. Happy to add it if a maintainer would rather have the belt and braces.

Line references checked against `v0.9.0` at `dc69c19aa7`, with the assistance of an AI coding tool.

PR #9383, opened 2026-08-30
choco: pass subprocess commands as argument lists

BIMvoice:
Item 2 of #8870, promised on 25 July and not delivered until now. `ci-bonsai-choco` has failed on every run since, most recently today (run 33294965528):

```
FileNotFoundError: [Errno 2] No such file or directory: 'git status'
```

`subprocess.check_output("git status", text=True)` with a bare string and no `shell=True` asks the OS to execute a file literally named `git status`. Same at line 25 (`"git tag -l"`) and in the `run()` helper used for the choco build steps.

Fix: argument lists for the two literal calls, and `shlex.split()` inside `run()` so its callers, which pass strings like `mono /opt/chocolatey/choco.exe pack --allow-unofficial`, keep working unchanged.

Verified by `py_compile`, black and ruff. Not verified by running the choco release itself, which needs the release runner. The failing step is the first `check_output` call, so it cannot get further than that today regardless.

Produced with AI assistance.

### DIFF
--- a/choco/bonsai/choco_release.py
+++ b/choco/bonsai/choco_release.py
@@ -14,6 +14,7 @@ import hashlib
 import os
 import pathlib
 import re
+import shlex
 import subprocess
 from typing import NoReturn
 from urllib import request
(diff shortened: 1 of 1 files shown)


BIMvoice:
Closing in favour of #8942, which fixes the same defect in the same file more completely.

#8942 converts every call site, including the `check_call(f"wget {url} --no-verbose")` that this PR leaves as a bare string, and it changes `run()` to take an argument list rather than splitting inside the helper. This PR was opened without noticing #8942 was already there, which is my mistake.

While comparing them I found one more defect and pushed the fix to #8942: the `setapikey` argument was a plain string containing `{choco_token}`, not an f-string, so chocolatey was being sent the literal placeholder instead of the token from the environment.

PR #9389, opened 2026-08-30
api/unit: leave derived Dimensions unset when creating IfcSIUnit

BIMvoice:
Two `src/ifcquery/tests/test_validate.py` tests fail on `v0.9.0` with `Attribute is derived in subtype` (run [33304472332](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/33304472332)).

`assign_unit.py` creates SI units positionally as `createIfcSIUnit(None, "LENGTHUNIT", ...)`, passing an explicit `None` for `Dimensions`, which is derived on `IfcSIUnit`. On `v0.9.0` that `None` lands in the raw slot instead of the `attribute_value_derived` sentinel, and `validate()` at `validate.py:555` flags every such unit. The STEP output is unaffected (`*` either way) and `unit.Dimensions` still resolves; only the raw slot differs.

Measured on a native `v0.9.0` build at `6f2e1aa993`, same file, three units:

| created as | raw slot passes validate |
|---|---|
| `create_entity("IfcSIUnit", UnitType=..., Name=...)` | yes |
| `createIfcSIUnit(None, "LENGTHUNIT", None, "METRE")` | no |
| `createIfcSIUnit(UnitType=..., Name=...)` | yes |

Fix here: keyword arguments that leave `Dimensions` unset in `create_metric_unit()` and `create_imperial_unit()`. RED reproduces the three `Attribute is derived in subtype` errors, GREEN `test_validate.py` 6 passed and `test/api/unit` 47 passed.

@aothms one question on the layer. The same `createIfcSIUnit(None, ...)` idiom appears in Bonsai, ifc5d, ifccityjson, ifc2ca and several test fixtures, and I have not checked whether `v0.8.0` stored the explicit `None` the same way. If you would rather the wrapper coerce a `None` written into a derived slot to the derived sentinel (or reject it), this PR becomes a tidy-up rather than the fix and I will drop it in favour of that. Otherwise I can sweep the other call sites in a follow-up.

Part of the `v0.9.0` CI tracking in #8870.

### DIFF
--- a/src/ifcopenshell-python/ifcopenshell/api/unit/assign_unit.py
+++ b/src/ifcopenshell-python/ifcopenshell/api/unit/assign_unit.py
@@ -128,10 +128,9 @@ class Usecase:
         elif unit_type == "volume":
             type_prefix = "CUBIC_"
         return self.file.createIfcSIUnit(
-            None,
-            "{}UNIT".format(unit_type.upper()),
-            ifcopenshell.util.unit.get_prefix(data["raw"]),
-            type_prefix + ifcopenshell.util.unit.get_unit_name(data["raw"]),
+            UnitType="{}UNIT".format(unit_type.upper()),
+            Prefix=ifcopenshell.util.unit.get_prefix(data["raw"]),
+            Name=type_prefix + ifcopenshell.util.unit.get_unit_name(data["raw"]),
         )
 
     def create_imperial_unit(self, unit_type: str, data: dict) -> ifcopenshell.entity_instance:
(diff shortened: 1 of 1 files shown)


aothms:
If this is a behavioural change with v0.8 then make sure that a positional None (not sure about keyword arg) is accepted for a derived attribute and also written as a derived in subtype marker (`*` ). No need to adapt the tests or API if this passed in v0.8.

BIMvoice:
Measured on 0.8.5: positional `None`, keyword `Dimensions=None`, keywords omitting it and plain `create_entity` all write `*` and validate clean. On `v0.9.0` both the positional and the keyword `None` are stored as a real null (STEP output still shows `*`, only `validate()` sees it), so this is a behavioural change in the wrapper, not the API.

Agreed then: the fix goes in `set_attribute_value_py`, which currently turns every `None` into `blank{}` without checking whether the slot is derived. #9394 mirrors `populate_derived_()` and stores `derived{}` for a derived slot, with RED/GREEN on a real `v0.9.0` build and the four cases tabulated. Closing this one; `assign_unit.py` and the tests stay as they are.

Issue #26, opened 2016-01-26
new CharacterDecoder.cpp


ninichki:
Hi Thomas,

thanks for your create work.
I just tried the 0.5.dev Version on GitHub.
If i use the last changes in CharacterDecoder.cpp, some Archicad Version 19 Ifc Modells don't work anymore.
Sometimes a lot of IfcProducts are missing, sometimes it can't find anything anymore.

[TPO BIB-Gebaeude - Testdatei CKi_Teilausschnitt.zip](https://github.com/IfcOpenShell/IfcOpenShell/files/104907/TPO.BIB-Gebaeude.-.Testdatei.CKi_Teilausschnitt.zip)

In the same file are some walls which are not correctly clipped. Maybe you know the reason why?

![image](https://cloud.githubusercontent.com/assets/1086380/12581302/0dd88c38-c436-11e5-9466-81219e6327a2.png)

kind regards,

Christian

aothms:
Hi,

Thanks for this. Indeed a blatant bug was introduced. It's fixed here https://github.com/IfcOpenShell/IfcOpenShell/commit/fe70acb541d1fec9685f63d3c00f33033f42ea37

As for issues with wall subtractions. There is are some boolean operands defined that as boundary representations that violate the standard [1]. For example for IfcClosedShell `#96789` the are some intersecting faces and other errors. This confuses Open Cascade up to the extent it is not able to perform the boolean operations. Of course there may be ways to make this more robust, but in this particular case the file is not without flaws.

[1] http://www.buildingsmart-tech.org/ifc/IFC2x3/TC1/html/ifctopologyresource/lexical/ifcclosedshell.htm

Kind regards,
Thomas

PR #9396, opened 2026-08-30
ifcgeom: keep one copy of repeated faces in faceset helper duplicate removal

BIMvoice:
Re-port of #8772 against `v0.9.0`. Related to #418 (the residual left after the segfault itself was fixed).

## Root cause

`src/ifcgeom/kernels/opencascade/faceset_helper.cpp:254` on `v0.9.0`: `wires()` returns `false` for every loop whose identity is in `duplicates_`. When an `IfcConnectedFaceSet` lists the same `IfcFace` more than once (Solibri IFC Optimizer output does this), every occurrence is dropped instead of all but one. The shell gets holes, sewing produces a non-manifold boolean operand, the opening subtraction is rejected by the non-manifold check, and the element ends up with no geometry.

## Fix

Count the redundant occurrences per loop identity (`duplicate_skips_remaining_`) and skip only that many, so exactly one copy of each repeated face is still built. Distinct-instance duplicates and the non-manifold rejection in `boolean_utils.cpp` are untouched.

## Verification

Verified on a native linux/arm64 build of `v0.9.0` at 6f2e1aa99.

Fixture: `test/input/418--walls--segfault.ifc`, wall `#2543` (its opening repeats 84 of its 350 unique faces across 518 entries). Control: wall `#26`.

Command: `ifcopenshell.geom.create_shape(settings, wall, body_representation)` with default settings, then count `geometry.verts` and `geometry.faces` and check that every edge is used by exactly two triangles.

RED (unpatched): `#2543` gives 0 verts, 0 faces.

GREEN (patched): `#2543` gives 106 verts (318 coordinates), 208 faces, every edge used exactly twice (watertight).

Control `#26` is 16 verts, 28 faces, watertight, before and after.

### DIFF
--- a/src/ifcgeom/kernels/opencascade/faceset_helper.cpp
+++ b/src/ifcgeom/kernels/opencascade/faceset_helper.cpp
@@ -108,7 +108,7 @@ ifcopenshell::geom::open_cascade_kernel::faceset_helper::faceset_helper(
 
 		vertex_mapping_.clear();
 		duplicates_.clear();
-		duplicate_skips_remaining_.clear();
+		duplicate_identities_built_.clear();
 
 		edge_use.clear();
 
@@ -176,7 +176,6 @@ ifcopenshell::geom::open_cascade_kernel::faceset_helper::faceset_helper(
 			if (edge_sets.find(edge_set_key) != edge_sets.end()) {
 				duplicate_faces++;
 				duplicates_.insert(loop->identity());
-				duplicate_skips_remaining_[loop->identity()]++;
 				continue;
 			}
             edge_sets.insert(edge_set_key);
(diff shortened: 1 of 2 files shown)


aothms:
There's two cases here:

 - loops with the same identity (these indeed have the defects)
 - loops without the same identity that after point mapping are marked as duplicate (no defect)

It seems to me that loops with the exact same identity can be handled more easily such as turning std::vector<ifcopenshell::geom::taxonomy::loop::ptr> loops; into a set.

Did you consider that?

BIMvoice:
Adopted, with one adjustment found by measuring. You are right about the two cases, and set semantics are the cleaner model for the same-identity one. Deduplicating the `loops` container itself turned out not to be enough on its own, because `wires()` receives each loop from the caller's traversal of the shell's faces, not from that container, and the instrumented build shows it is invoked three times per duplicate identity on the #418 fixture, not twice. So `65da4db8e` applies the set at the decision point instead: a `std::set` of duplicate identities that have already built once, first occurrence builds, the rest skip. The skip counter is gone, and this is also robust if the helper is ever driven over the same loops more than the constructor saw, where a consumed counter would under-build.

Distinct-identity loops that only become duplicates after point mapping behave exactly as before, no defect there as you said.

Re-verified on the native arm64 build, same fixture: wall `#2543` gives 106 verts / 208 faces, watertight (every edge used exactly twice), control wall `#26` unchanged at 16 / 28.

PR #9397, opened 2026-08-30
docker: document a native linux/arm64 build environment

BIMvoice:
Adds `docker/README-arm64.md`: how to build `v0.9.0` natively on Apple Silicon with the unmodified `docker/Dockerfile` under `--platform linux/arm64` and the `rockylinux9-arm64` dependency cache, with the measured numbers (full build with the Python 3.13 wrapper in 848 s at `-j2`, versus roughly 4.5 h for the amd64 image under QEMU) and the memory limit that forces `-j2` on a 7.75 GB Docker Desktop VM.

It also records the harness's known limitation: with static OpenCASCADE the SVG, STEP and IGES serializers fail (`Standard_NoSuchObject`, empty B-rep), the per-plug-in static OCCT problem of #9341, and building `--shared` does not help on arm64 because the cached `occt-shared-7.8.1` was compiled against glibc 2.38 and CXXABI 1.3.15, newer than Rocky Linux 9 (`IfcGeomServer` fails to link on `fmod@GLIBC_2.38`). That last point may be useful for #9341 if the arm64 cache is rebuilt.

Documentation only, no code changes. Used today to verify #9390 to #9396 and #9394 on a real build.

### DIFF
--- a/docker/README-arm64.md
+++ b/docker/README-arm64.md
@@ -146,3 +146,19 @@ not through `ifcos_env build`), so there's nothing to ignore there.
 
 Do not touch any `occt:*` / `occtdraw:*` image - unrelated to this harness,
 belongs to a different maintainer's build setup.
+
+## Known limitation: OCC-native serializers
+
+With the default static OpenCASCADE link, the SVG (`--plan`, `--model`),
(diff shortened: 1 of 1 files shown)


sboddy:
Cool!

Issue #9398, opened 2026-08-30
SVG serializer aborts with Standard_NoSuchObject in a fully static build


P4A-KLFSBK:
### Summary

A build produced by the documented Docker developer workflow (`docker/README.md` -> `./ifcos_env build`) yields an `IfcConvert` whose SVG serializer aborts on almost every element:

```
Creating geometry...
terminate called after throwing an instance of 'Standard_NoSuchObject'
```

It reproduces on any model with solids, including `test/input` files and the Duplex sample, so it is not model specific. OBJ and GLB output from the same binary are fine, which makes it look like an SVG bug rather than a build problem.

### Cause

`nix/build-all.py` defaults to static IfcOpenShell libraries and static OCCT. Each serializer plugin therefore statically links its **own** copy of OCCT, compiled `-fvisibility=hidden` so the copies are not interposable. Shapes created inside `ifcopenshell_geometry_kernel_opencascade.so` then fail handle downcasts against the SVG plugin's separate OCCT instance: `BRep_Tool::Surface` returns a surface that is not loaded, and the deep-copying transform in `svg_serializer::write` throws.

Backtrace, from a build with debug info added to the SVG plugin:

```
Standard_NoSuchObject
  GeomAdaptor_Surface::UContinuity
  Approx_CurveOnSurface::Perform
  ProjLib_ProjectedCurve::Perform
  BRep_Tool::CurveOnPlane
  BRep_Tool::CurveOnSurface
  BRepTools_TrsfModification::NewCurve2d
  BRepTools_Modifier::Rebuild
  BRepTools_Modifier::Perform
  BRepBuilderAPI_ModifyShape::DoModif
  BRepBuilderAPI_Transform::Perform
  BRepBuilderAPI_Transform::BRepBuilderAPI_Transform   <- svg_serializer.cpp, BRepBuilderAPI_Transform(..., true)
```

The exception is not caught, so one bad element ends the whole conversion.

This also explains a secondary symptom: the duplicated IfcOpenShell libraries give each plugin its own `Logger` singleton, so messages logged from inside a serializer never reach `IfcConvert`'s output.

### Why OBJ and GLB are unaffected

They consume triangulated output rather than `TopoDS_Shape`, so they never cross the OCCT instance boundary. Converting the same model to `.obj` succeeding while `.svg` aborts is a reliable way to recognise this.

### How to check a build

```
[...]

PR #9399, opened 2026-08-30
Build shared in the docker dev environment

P4A-KLFSBK:
`docker/README.md` tells developers to build with `./ifcos_env build`, which invokes `build-all.py` without `--ifcopenshell-shared`. Both CI workflows pass that flag:

- `.github/workflows/build_rocky.yml`
- `.github/workflows/build_osx.yml`

so the configuration CI builds and ships is not the one the documented developer workflow produces.

The difference is not cosmetic. Defaulting to static links a private copy of OCCT into every serializer plugin, each built `-fvisibility=hidden` so the copies are not interposable. Shapes created in `ifcopenshell_geometry_kernel_opencascade.so` then fail handle downcasts against the SVG plugin's own OCCT instance, and `IfcConvert model.ifc out.svg` aborts with `Standard_NoSuchObject` on nearly every element. OBJ and GLB output from the same binary is fine, because those serializers consume triangulated output and never cross the boundary, so the failure presents as an SVG or model problem rather than a build one.

Full diagnosis and backtrace in #9398.

To check whether a given build is affected:

```
readelf -d <install>/lib64/ifcopenshell_geometry_svg.so | grep -c "NEEDED.*libTK"
```

0 means the plugin carries its own static OCCT; a correct build reports 48.

This PR changes one line so the developer build matches CI. It does not attempt to make the fully static configuration work; if that configuration is meant to be supported, this is a workaround and #9398 is the real bug.

---

This change was generated with the assistance of an AI coding tool.

### DIFF
--- a/docker/ifcos_env
+++ b/docker/ifcos_env
@@ -155,7 +155,7 @@ function build() {
     
     docker exec -i -w "${WORKDIR}" -e PY_TGT="${PY_TGT}" -e BUILD_TARGET="${BUILD_TARGET}" "${NAMEPREFIX}-${UNIQUE_ID}" bash -c '
         set -o pipefail
-        CXXFLAGS="-O3" CFLAGS="-O3 ${DARWIN_C_SOURCE}" ADD_COMMIT_SHA=1 BUILD_CFG=Release uv run ./nix/build-all.py -v ${PY_TGT:+-$PY_TGT} --diskcleanup ${BUILD_TARGET} 2>&1 | tee build.log
+        CXXFLAGS="-O3" CFLAGS="-O3 ${DARWIN_C_SOURCE}" ADD_COMMIT_SHA=1 BUILD_CFG=Release uv run ./nix/build-all.py -v ${PY_TGT:+-$PY_TGT} --diskcleanup --ifcopenshell-shared ${BUILD_TARGET} 2>&1 | tee build.log
     '
     echo "🎒 Pack Dependencies"
     docker exec -i -w "${WORKDIR}" "${NAMEPREFIX}-${UNIQUE_ID}" bash -c '


sboddy:
There's some stuff going on in #9341 that is about how OCCT gets compiled in. I was waiting for that to get resolved. The line you are aligning here is somewhat recent to the 0.9.0 build process. The ifcos_env script is something I created to give me fast turnaround when developing C++ changes in 0.8.0. I have no problem with this change, I just don't know if the result will work properly. I think it needs the other flag (--occt-shared) or the code change from Moult in that PR so that the libraries actually work.

P4A-KLFSBK:
You are right, and the flag I picked is not sufficient on its own.

`build-all.py` describes `--ifcopenshell-shared` as "Build only IfcOpenShell's own libraries as shared (dependencies stay static)", and OCCT's link type is governed by a separate flag entirely:

```python
occt_link_type = "Shared" if ARGS.occt_shared else "Static"
```

Since `ifcos_env` passes neither `--shared` nor `--occt-shared`, this one-line change still leaves OCCT static, so every consumer keeps a private copy and its own `Standard_Type` registry — exactly the failure Andrej730 characterised in #9341 and that you then confirmed on your Bonsai profile once shared OCCT was applied. So `--occt-shared` is doing the work here, not the flag I added.

That also undercuts the check in my description: `readelf -d ... | grep -c "NEEDED.*libTK"` cannot report 48 while OCCT is static, so the build I took that number from must have had shared OCCT by some other route. Either way it is not the right check for *this* diff, and I should not have presented it as one.

Given @aothms's direction in #9341 — drop the static/shared flags from the build script and default to fully shared — a one-line flag tweak here is the wrong shape of fix, and in the interim it would need `--occt-shared` as well. I am happy to leave this parked behind #9341 rather than add a second flag that the full-shared change would then delete. If #9341 stalls and the documented docker workflow is still producing broken SVG output, I will come back and add both flags as a stopgap.

#9398 stands on its own as the underlying report either way.

Flag semantics re-checked against `v0.9.0` at `dc69c19aa7`, with the assistance of an AI coding tool.

PR #9406, opened 2026-08-31
Bonsai: editing a window/door no longer overwrites attributes of occurrences owning their own geometry

### LINKED ISSUE #9353: Bonsai: parametric window edit overwrites OverallWidth/OverallHeight on all occurrences of the type
### Bug Description

`update_window_modifier_representation()` writes the edited window's `OverallWidth` / `OverallHeight` to every occurrence of the same `IfcWindowType`, even when each occurrence owns its own geometry.

Geometry is handled correctly. It goes through `get_part_of_product()`, so an occurrence with its own `IfcProductDefinitionShape` keeps its own body, and only the type-mapped case writes to the shared `IfcRepresentationMap`. The attribute loop directly below carries no such condition:

```python
# occurrences attributes
occurrences = tool.Ifc.get_all_element_occurrences(element)
for occurrence in occurrences:
    occurrence.OverallWidth = props.overall_width / si_conversion
    occurrence.OverallHeight = props.overall_height / si_conversion
```

`get_all_element_occurrences()` returns every occurrence of the type when the element is typed, and `[element]` when it is not.

The result is a state no consumer can reconcile: sibling windows keep geometry of one size while reporting the attributes of another. Quantity take-off, schedules and IDS all read the attributes, not the body.

This is reachable through Bonsai's own default flow. `mesh.add_window` creates an untyped occurrence, so authoring several differently sized windows and typing them afterwards is the normal path — and it produces exactly the mixed state in which the loop does damage.

The module states its intended scope itself, in `_WindowEditMixin`:

> Single-object by design (window edits target the active object only).

The geometry path honours that. The attribute path does not.

`tool.Model.update_simple_openings()` iterates the same set and makes its assumption explicit in a comment: *"We assume all occurrences of the same element type (e.g. a window) will use openings of the same thickness."* That assumption only holds when the occurrences share the type's `IfcRepresentationMap` — which is precisely the condition `get_part_of_product()` already evaluates two lines earlier.

### Steps to reproduce

Verified in an empty project, IFC4, two elements, no parameter changed:

1. New IFC project.
2. `mesh.add_window`, then `bim.enable_editing_window` / `bim.finish_editing_window` with `overall_width = 0.80`, `overall_height = 1.00`. → window A.
3. `mesh.add_window` again, same cycle with `overall_width = 1.60`, `overall_height = 2.20`. → window B.
4. Create an `IfcWindowType` and assign both occurrences to it. `should_map_representations=False`, so both keep their own representations and only the attribute path is under test.
5. Select window A. `bim.enable_editing_window`, then `bim.finish_editing_window` — **without changing any parameter**.

Observed on window B:

| | `dimensions` | representation ids | `OverallWidth` × `OverallHeight` |
| --- | --- | --- | --- |
| before | 1.60 × 2.20 | 287, 330, 347 | **1.60 × 2.20** |
| after | 1.60 × 2.20 | 287, 330, 347 — unchanged | **0.80 × 1.00** |

Window A behaved as expected: its own representations were replaced (145 → 352), geometry and attributes unchanged.

Also observed on a production model, IFC4X3, with 28 `IfcWindow` occurrences on a single `IfcWindowType`, 21 distinct `OverallWidth` / `OverallHeight` pairs, each with its own body representation: one null edit rewrote the attributes of 24 of them. Body geometry of the others was untouched, confirming the same split at scale.

### Suggested fix

Gate the loop on the same condition as the geometry: write to all occurrences only when the representation is type-mapped, otherwise to the active element alone.

### Debug and Error Output

```
blender_version: 5.2.0 LTS
bonsai_version: 0.8.5
schema: IFC4 (minimal repro), IFC4X3 (production model)
```

Source checked against branch `v0.8.0` at the time of writing; the loop is unchanged there.

BIMvoice:
### Root cause

`update_window_modifier_representation()` at `src/bonsai/bonsai/bim/module/model/window.py:141` (and its sibling `update_door_modifier_representation()` at `src/bonsai/bonsai/bim/module/model/door.py:161`) wrote the edited element's `OverallWidth`/`OverallHeight` to every occurrence of the same type unconditionally:

```python
occurrences = tool.Ifc.get_all_element_occurrences(element)
for occurrence in occurrences:
    occurrence.OverallWidth = props.overall_width / si_conversion
    occurrence.OverallHeight = props.overall_height / si_conversion
```

The geometry path right above is conditional: `replace_object_ifc_representation()` resolves the edited object's Body representation, so only a type-mapped representation (shared through the `IfcRepresentationMap`) is rewritten for the siblings, while an occurrence owning its own `IfcProductDefinitionShape` keeps its own body. The attribute loop had no such condition, so editing one window corrupted the attributes of siblings that own their own geometry (typical for imported IFCs where occurrences of one `IfcWindowType` each carry their own body).

### Fix

The attribute loop now mirrors the geometry path: besides the edited element itself, only occurrences whose Model/Body representation resolves (via the mapped-representation chain) to the representation the modifier just rebuilt receive the new `OverallWidth`/`OverallHeight`. The helper `has_shared_model_body()` lives in `window.py` and is reused by `door.py`, which already imports from it.

### Door verdict

`update_door_modifier_representation()` had the exact same unconditional loop and is fixed the same way, with its own regression test.

### RED/GREEN

New tests in `src/bonsai/test/bim/module/model/test_window_door_occurrence_attrs.py`, run headless in Blender 5.2 against this worktree's code:

- RED (fix reverted, same worktree): `2 failed, 1 passed`. The window sibling with its own geometry went from `OverallWidth` 0.6 to 0.85 and the door sibling from 0.9 to 1.15 after editing the other occurrence.
- GREEN (fix applied): `3 passed`. The own-geometry sibling keeps its attributes; a type-mapped sibling still receives them (shared-geometry behavior preserved).
- Full `test/bim/module/model/` sweep with the fix: `533 passed`.

CI-exact black and ruff pass on the three touched files.

Fixes #9353

### DIFF
--- a/src/bonsai/bonsai/bim/module/model/door.py
+++ b/src/bonsai/bonsai/bim/module/model/door.py
@@ -38,7 +38,7 @@ import bonsai.tool as tool
 from bonsai.bim.module.drawing import gizmos as gizmo
 from bonsai.bim.module.drawing.gizmos import DimensionGizmoConfig
 from bonsai.bim.module.model.wall_offset_gizmos import WALL_OFFSET_GIZMO_CONFIGS
-from bonsai.bim.module.model.window import create_bm_box, create_bm_window
+from bonsai.bim.module.model.window import create_bm_box, create_bm_window, has_shared_model_body
 from bonsai.bim.parametric_lifecycle import FeatureModifierEditMixin, PickTypeMixin
 
 if TYPE_CHECKING:
@@ -158,8 +158,12 @@ def update_door_modifier_representation(obj: bpy.types.Object) -> None:
         element.OperationType = props.door_type
 
     # occurrences attributes
+    # Mirror the geometry path above: an occurrence with its own Body
+    # representation keeps its own OverallWidth/OverallHeight too.
     occurrences = tool.Ifc.get_all_element_occurrences(element)
     for occurrence in occurrences:
+        if occurrence != element and not has_shared_model_body(occurrence, model_representation):
+            continue
         occurrence.OverallWidth = props.overall_width / si_conversion
         occurrence.OverallHeight = props.overall_height / si_conversion
 
--- a/src/bonsai/bonsai/bim/module/model/window.py
+++ b/src/bonsai/bonsai/bim/module/model/window.py
@@ -50,6 +50,17 @@ V_ = tool.Blender.V_
 _G = gizmo.BaseParametricGizmoGroup
 
 
+def has_shared_model_body(
+    occurrence: ifcopenshell.entity_instance, base_representation: ifcopenshell.entity_instance
+) -> bool:
+    """Whether the occurrence's Model/Body geometry resolves to the
+    representation the modifier just rebuilt (i.e. is not its own)."""
+    representation = ifcopenshell.util.representation.get_representation(occurrence, "Model", "Body", "MODEL_VIEW")
+    if representation is None:
+        return True
+    return tool.Geometry.resolve_mapped_representation(representation) == base_representation
+
+
 def update_window_modifier_representation(context: bpy.types.Context) -> None:
     obj = context.active_object
     assert obj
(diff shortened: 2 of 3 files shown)


M-H-H-B:
Friendly ping for review. #9406 fixes #9353 for both the window and the door editor and ships regression tests (RED/GREEN, full `test/bim/module/model/` sweep passing). We are blocked on this in production: an IFC4X3 model with 28 windows on one `IfcWindowType` and 21 distinct sizes. Until it lands we cannot use the parametric window/door editor on any shared type. Is anything missing (tests, changelog, rebase) to get it merged?

Reproduction for the openings part: #9591

PR #9408, opened 2026-08-31
Honour IfcAxis2PlacementLinear Axis/RefDirection in the loft builder

RickBrice:
@aothms I'm working with more and more complex infrastructure geometry, specifically IfcSectionedSolidHorizontal that represents layers of pavements swept along an alignment with superelevation transitions. When pavements stopped/started at the ends of a bridge with a skew, make_loft() wasn't handling the differences in IfcAxis2PlacementLinear.RefDirection. An example:

image

Here is how it looks before the fixes to make_loft(). 
image

I don't fully understand the math in make_loft(), so I had Claude help out with a solution. The solution generated appears to work well. Attached is an analysis of the before and after handling of the sectioned solid geometry.

[sectioned_solid_examples.zip](https://github.com/user-attachments/files/31650576/sectioned_solid_examples.zip)

I wanted to get your input on this before merging it into the main v0.9.0 branch

Thanks.

### DIFF
--- a/src/ifcgeom/mapping/IfcSectionedSolidHorizontal.cpp
+++ b/src/ifcgeom/mapping/IfcSectionedSolidHorizontal.cpp
@@ -49,7 +49,7 @@ taxonomy::ptr mapping::map_impl(const IfcSchema::IfcSectionedSolidHorizontal& in
 	// The longitudes determine the range of the sweep and the offsets are interpolated in between
 	// sweep segments.
 	std::vector<Eigen::Vector3d> profile_offsets;
-	std::vector<std::optional<Eigen::Matrix3d>> profile_rotations;
+	std::vector<std::optional<Eigen::Matrix3d>> profile_axes;
 	std::vector<std::optional<Eigen::Vector3d>> profile_ref_directions;
 	std::vector<double> longitudes;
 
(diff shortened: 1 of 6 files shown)


src/ifcgeom/mapping/IfcSectionedSolidHorizontal.cpp
```diff
@@ -50,6 +50,7 @@ taxonomy::ptr mapping::map_impl(const IfcSchema::IfcSectionedSolidHorizontal& in
 	// sweep segments.
 	std::vector<Eigen::Vector3d> profile_offsets;
 	std::vector<std::optional<Eigen::Matrix3d>> profile_rotations;
+	std::vector<std::optional<Eigen::Vector3d>> profile_ref_directions;
```
aothms:
```suggestion
	std::vector<std::optional<Eigen::Vector3d>> profile_axes;
	std::vector<std::optional<Eigen::Vector3d>> profile_ref_directions;
```

These things (rotation + ref) together probably makes no sense. I think profile_rotations is now only used for the Z? So maybe we rename it accordingly?

aothms (approved):

src/ifcgeom/mapping/IfcSectionedSolidHorizontal.cpp
RickBrice: Thank you Sir - I will make the changes and merge the PR

PR #9413, opened 2026-09-01
ifctester: apply strict XSD pattern semantics and warn on non-portable patterns

BIMvoice:
## What

IDS `xs:pattern` restrictions in ifctester were translated with xmlschema's permissive XPath defaults: `translate_pattern(pattern)` with `anchors=True, back_references=True, lazy_quantifiers=True`. Strict XSD `xs:pattern` semantics have no anchors, no back references and no lazy quantifiers, so ifctester could disagree with conformant IDS validators on the same file.

This PR:

1. Translates patterns with `back_references=False, lazy_quantifiers=False, anchors=False` in `Restriction.__eq__` (the only `translate_pattern` call site in the package), with an `functools.cache` on the compiled regex. Patterns that are not expressible in strict XSD at all (for example back references, which raise `RegexError` under the strict translator) fall back to the previous permissive translation instead of crashing.
2. Adds a portability lint: `Restriction.parse` now emits a `UserWarning` when a loaded pattern relies on non-XSD regex features (anchors, back references or lazy quantifiers). It only warns; it never changes a validation verdict on its own.

## Behaviour change, intentional

Part 1 flips verdicts for patterns that used `^` or `$` as anchors:

| Pattern | Value | Before | After (strict XSD) |
|---|---|---|---|
| `^.{1,5}$` | `AB` | pass | fail (`^` and `$` are literal characters in XSD) |
| `^.{1,5}$` | `^AB$` | fail | pass |
| `.{1,5}` | `abc` | pass | pass |
| `.{1,5}` | `abcdef` | fail | fail |

Plain patterns behave identically in both modes because ifctester already used `fullmatch`, which matches XSD's implicit whole-value anchoring. The new warning gives IDS authors the migration signal: any pattern that would change behaviour is flagged on load.

This aligns ifctester with the XSD semantics documented in buildingSMART/IDS#449, which describes exactly this divergence between validators.

## Tests

Added to `src/ifctester/test/test_facet.py` (`TestRestriction`):

- `test_pattern_treats_anchors_as_literal_characters`: `^.{1,5}$` rejects `AB` and accepts `^AB$` (fails on unpatched code, red first).
- `test_pattern_is_implicitly_anchored`: `.{1,5}` accepts `abc` and rejects `abcdef` (regression guard, passes before and after).
- `test_pattern_portability_warnings`: `^x$`, `(a)\1` and `a+?` each warn naming the feature; `.{1,5}` and `DT[0-9]` stay silent under `warnings.simplefilter("error")`.

Run with local CPython 3.11 and ifcopenshell 0.8.5 wheels, `PYTHONPATH` pointing at this branch's `src/ifctester`. The pattern logic only depends on `xmlschema`/`elementpath`, not on the 0.9 wrapper, so this is a valid check; honesty note: 7 unrelated tests in `test_facet.py`/`test_ids.py` fail identically on unmodified `v0.9.0` in this environment because the 0.8.5 wrapper lacks `entity_instance.get_attribute_category`. Before this patch: 7 failed, 30 passed. After: 7 failed (same 7), 33 passed.

`black --check` and `ruff check` (CI versions) pass on both touched files.

AI disclosure: this PR was prepared with AI assistance and reviewed by a human before submission.

### DIFF
--- a/src/ifctester/ifctester/facet.py
+++ b/src/ifctester/ifctester/facet.py
@@ -20,13 +20,15 @@ from __future__ import annotations
 
 import builtins
 import re
-from functools import lru_cache
+import warnings
+from functools import cache, lru_cache
 from logging import Logger
 from typing import TYPE_CHECKING, Any, Literal, Optional, TypedDict, Union
 
 import ifcopenshell.util.classification
 import ifcopenshell.util.element
 import ifcopenshell.util.unit
+from elementpath.regex import RegexError
 from xmlschema.validators import identities
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


Moult:

Thanks Petru, the strict semantics are the right call. Pushed a slimmed version to your branch: `translate_pattern` imported from `elementpath.regex` where it lives (and elementpath declared as a dependency), the permissive fallback and the portability warnings dropped since ifctester does not support invalid files and can fail hard, and the cache moved onto the translation itself, which is the expensive step, by rebinding the import with `functools.cache`. Your two behaviour tests are kept; the warnings test is replaced by one asserting that back references and lazy quantifiers raise. Merging once CI is green.

PR #9416, opened 2026-09-01
bsdd: make the include_test_dictionaries default a real boolean

BIMvoice:
The default was the string `"False"`, which is truthy, so any bare `get_dictionary()` call silently sent `IncludeTestDictionaries=true` and returned test uploads nobody asked for. Bonsai is unaffected because it always passes the preference explicitly (`tool/bsdd.py:143`); only direct library users hit it.

RED/GREEN: the new offline test asserts the signature default `is False`; on unpatched code the default is the string `'False'` and the assertion fails, on this branch it passes. Verified by inspecting both versions directly. The test is offline on purpose, since the rest of `test_bsdd.py` performs live API calls.

Found while investigating why a user's bSDD test upload was invisible in Bonsai (the answer there was the separate Load Test Dictionaries and Load Preview Dictionaries preferences, both needed for a test upload, which arrives with Preview status).

### DIFF
--- a/src/bsdd/bsdd.py
+++ b/src/bsdd/bsdd.py
@@ -712,7 +712,7 @@ class Client:
         return self.get(f"api/Unit/{version}")
 
     def get_dictionary(
-        self, dictionary_uri: str = "", include_test_dictionaries: bool = "False", version: int = 1
+        self, dictionary_uri: str = "", include_test_dictionaries: bool = False, version: int = 1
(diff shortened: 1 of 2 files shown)


Moult:

Thanks Petru. Pushed to your branch: the one-line fix as you wrote it, with the test case dropped (a default value in a signature does not need a test) and the branch rebased onto current v0.9.0 as a single commit. Merging once CI is green.

PR #8513, opened 2026-07-11
Guard append_asset against an empty aggregate attribute (#7261)

BIMvoice:
Fixes #7261.

## Problem
`project.append_asset` raises `IndexError: tuple index out of range` when an appended asset has an empty aggregate attribute (e.g. an `IfcCartesianPointList2D` with `CoordList == ((),)`).

## Root cause
`get_tuple_type` descends nested tuples with `while isinstance(tuple_, tuple): tuple_ = tuple_[0]`, indexing `[0]` into an empty tuple.

## Fix
Stop descending at an empty tuple:
```python
while isinstance(tuple_, tuple) and tuple_:
    tuple_ = tuple_[0]
return type(tuple_)
```
It then returns `tuple`, which matches neither `entity_instance` nor `float` in the copy loop, so the empty aggregate is carried through unchanged and the append proceeds.

## Verification (headless)
- Empty aggregate (`CoordList=((),)`): before → IndexError; after → append succeeds, empty list preserved.
- Normal populated profile: appends unchanged.

This contribution was generated with the assistance of an AI coding tool.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Also fixes #6972 (same get_tuple_type empty-tuple crash, triggered by an invalid/empty IfcSurfaceStyles list when appending a type).

PR #9422, opened 2026-09-01
Consolidated ExtractElements fixes based on BIMvoice proposals

PahlawanJoey:
We have consolidated the open IfcPatch:ExtractElements fixes into a single implementation based on the v0.9.0 branch.

Rather than treating the individual reports and proposed fixes separately, we reviewed the overlapping issues and combined the relevant approaches into one coherent solution. This includes work based on the following PR proposals:

* #9420 — presentation layer membership
* #9421 — placement ancestor cleanup
* #9255 — IFC2X3 ports and subtype-aware type assignment
* #9050 — deferred/batched relationship, representation and layer updates
* #9184 — efficient reverse-index / safe-removal cleanup
* #9038 — IFC2X3 georeferencing preservation
* #8507 — blank-name asset handling
* #8663 — material styles through layer-set usage
* #8756 — class-compatible GlobalId reuse
* #8668 — safe entity removal handling
* #8513 — empty aggregate handling

The batching approach from #9050 also supersedes the earlier #8320 proposal and incorporates the nested uniqueness behaviour proposed in #8956.

We then added the additional fixes required by the related ExtractElements bug reports, including preservation of shared relationship members, non-IfcRoot relationship members, spatial parents and element types, shared representations, IFC2X3 MEP ports and georeferencing data, material styles, blank-named assets, GlobalId collisions and orphaned placement chains.

A large part of the design and prior investigation behind these fixes comes from @BIMvoice's proposed PRs. Full credit to @BIMvoice for identifying, analysing and proposing solutions for many of these problems. The intention here was not to replace that work, but to reconcile the overlapping proposals and bug reports into one integrated patch that can be reviewed and tested as a whole.


BIMvoice:
@PahlawanJoey this PR has no check runs at all, and that is not something you can fix from your side. Three workflow runs for `fix/extract-elements-v090` have been sitting in `action_required` since 2026-09-01, waiting for a maintainer to approve them. Until someone approves them the PR cannot report green and cannot be merged. You are not alone in this: there are 16 such runs across six pull requests, the oldest queued on 2026-08-15. I have listed them all on #8870 so it is visible in one place.

One thing I could check without the runners: I checked out your branch and ran the same black and ruff configuration `ci-lint` uses, over all nine files you touch. Both pass, with no findings. That is not a substitute for the real CI, since it says nothing about `compile-and-test`, but it does mean the formatting job is not what is holding this up. The approval is.

On how this relates to our two PRs, since the title mentions our proposals, here is what I verified rather than assumed:

* `edit_object_placement.py`: your change is byte for byte identical to #9421. Same resulting blob, `a6fbd86460`.
* `append_asset.py`: yours is a superset of #9420. It fixes the same `AssignedItems` accumulation on the default path, and additionally offers the `deferred_layer_items` and `deferred_relationship_members` accumulators for callers that want to avoid the O(n^2) rewrite.

So there is no disagreement between these PRs and no reason for them to compete. If #9422 goes green and is the version the maintainers prefer, we will close #9420 and #9421 as superseded, and I would rather that happen than have three PRs fixing one bug.

One practical note in the meantime. #9420 and #9421 were showing red until today, but that red was the `v0.9.0` baseline itself, which was broken between 2026-08-28 and 2026-09-02 and is now fixed. Merging current `v0.9.0` into the branch and pushing makes the checks re-run against a green base. #9420 came back fully green that way. That will not help here while the runs are unapproved, but it is worth doing once they are.

PR #9429, opened 2026-09-02
serializers: give face_info ownership semantics so it cannot be shallow copied (#9380)

### LINKED ISSUE #9380: face_info violates the rule of three in svg_serializer.h
`prefiltered_hlr::face_info` in `src/serializers/svg_serializer.h` has a destructor that deletes a raw pointer, but no copy constructor and no copy assignment operator:

```cpp
class face_info {
    ...
    BRepTopAdaptor_FClass2d* fclass;

    face_info(TopoDS_Shape* it, const TopoDS_Face& fa)
        : item(it), face(fa), fclass(nullptr)
    { ... }

    ~face_info() {
        delete fclass;
    }
```

It is stored by value in `std::multimap<double, face_info> large_ortho_faces_`, and `fclass` is allocated lazily inside `contains()`. Because the user-declared destructor suppresses the implicit move constructor, insertion copies, and the implicitly generated copy constructor copies the pointer shallowly.

This is currently harmless only by accident: `fclass` is still null at the moment of insertion, so the temporary's destructor deletes nothing and the copy in the map starts null as well. Any future code that copies a `face_info` after `contains()` has run would double free.

Either declaring the copy operations deleted, or holding `fclass` in a `std::unique_ptr` and letting the class be move-only, would make the intent explicit and remove the latent trap. I have not opened a PR because the safe choice depends on whether the class is ever meant to be copied, which is a design call rather than a fix.

Found while reading the serializer for #3072. Not a bug anyone can hit today, as far as I can tell — filing it so it is written down rather than rediscovered.

*This analysis was produced with the assistance of an AI coding tool.*

BIMvoice:
Fixes #9380.

## The reporter's analysis is correct

@P4A-KLFSBK is right on every point. `prefiltered_hlr::face_info` in `src/serializers/svg_serializer.h` declared a destructor that `delete`d a raw `BRepTopAdaptor_FClass2d*`, but no copy constructor and no copy assignment operator. The user-declared destructor also suppressed the implicit move constructor, so `std::multimap<double, face_info> large_ortho_faces_` was populated by copy, and the implicitly generated copy constructor copied the pointer shallowly. He is also right that it was harmless in practice, and right that the safe choice was a design call rather than a mechanical fix.

## Is it reachable today? No, it is genuinely latent

I traced every use of the type and of the container:

* There is exactly one insertion site, `large_ortho_faces_.insert({ d, face_info(&it->second, face) })` inside `prefiltered_hlr::add()`. At that moment `fclass` is still null, because the constructor leaves it null and only `contains()` allocates it.
* `fclass` is first allocated in `contains()`, which is reached only from `is_obscured_()`, which is reached only from `build()`. All `add()` calls precede `build()`.
* `is_obscured_()` iterates the multimap through iterators and calls `it->second.contains(...)` on a reference. Nothing is taken by value.
* `std::multimap` is node based. It never rehashes, rebalances by copying, or otherwise duplicates elements once they exist, so the entries are never copied after `contains()` has run.
* There is no `erase`, no assignment, no pass by value, and no other mutation of `large_ortho_faces_` while entries are live.
* The one place that puts a whole `prefiltered_hlr` into a container is `storey_hlr.insert({ storey, hlr_engine(...) })` in `svg_serializer.cpp`. That is not a copy at all: the braced temporary is an rvalue, so `insert(value_type&&)` is selected and the engine is moved into place. The copy constructor is never instantiated.

  (An earlier revision of this body called that line a copy that was safe because the engine is freshly constructed and its multimap empty. That was wrong twice over and @P4A-KLFSBK corrected it below: it is a move rather than a copy, and whether the map happens to be empty is a runtime property that has no bearing on which constructor gets instantiated.)

So there is no double free and no dangling pointer reachable in the current code. The defect is latent, exactly as reported. The value in fixing it is that the trap was one edit away and invisible at the call site.

## The fix

`fclass` is now a `std::unique_ptr<BRepTopAdaptor_FClass2d>`. This removes the destructor, which restores the implicit move constructor, and makes the copy constructor and copy assignment operator implicitly deleted. The shallow copy can no longer be written at all, rather than being merely discouraged by a comment. `contains()` is a non-const member and is called on a non-const reference, so no `mutable` and no helper indirection were needed; the lazy initialisation just becomes `std::make_unique`. The insertion site becomes `emplace(d, face_info(...))` so the element is moved into place.

I chose ownership over the explicit rule of three because copying a `face_info` has no sensible meaning here: `BRepTopAdaptor_FClass2d` is not copyable, which is why the raw pointer existed in the first place, so a correct copy constructor would have to either drop the cached classifier or rebuild it, and both are silent behaviour the call site would not expect.

## What the compiler now guarantees

Three `static_assert`s sit next to the class:

```cpp
static_assert(!std::is_copy_constructible_v<face_info>, "face_info must not be copyable");
static_assert(!std::is_copy_assignable_v<face_info>, "face_info must not be copy assignable");
static_assert(std::is_move_constructible_v<face_info>, "face_info must be movable");
```

Any future change that reintroduces a copyable `face_info`, or that accidentally makes it non-movable and therefore unusable in the multimap, now fails the build rather than compiling into a latent double free.

## Verification

Built incrementally in the aarch64 Linux build container. `svg_serializer.cpp.o` and `geometry_svg_plugin.cpp.o` were both recompiled and `ifcopenshell_geometry_svg.so` relinked, with no new warnings (the two `-Wsubobject-linkage` warnings on `svg_serializer` and the GCC 10.1 parameter-passing note are pre-existing and unrelated). Since this is a compile-time correctness change, the compiler is the proof: the fix would not build if any code path still copied a `face_info`.

To be plain about the limit of that: I could not exercise the SVG serializer at runtime in this build. With static OCCT the serializer plug-ins abort with `Standard_NoSuchObject` (see #9341), so no SVG output was produced or compared. The change is intentionally behaviour-preserving; the only runtime difference is that the classifier is now freed through `unique_ptr` instead of an explicit `delete`.

Thanks to @P4A-KLFSBK for reading the serializer carefully and writing this down instead of leaving it to be rediscovered.


P4A-KLFSBK:
Confirmed from my side, and ownership is the semantics I would have picked. `BRepTopAdaptor_FClass2d` is not copyable, which is why the raw pointer existed at all, so any hand-written copy constructor would have had to either drop the cached classifier or rebuild it, and both are silent surprises at the call site. Making the shallow copy unwriteable is a better outcome than the rule of three I suggested. The `static_assert`s are the part I would not have thought to add.

I re-read the current tip to check nothing else needs to move with it, and nothing does: `face_info` appears only as the class, its constructor, the `std::multimap<double, face_info>` member and the single insertion site; `is_obscured_` iterates with `for (auto it = lower; it != upper; ++it)` and reaches `it->second` through the iterator rather than by value; and `contains()` is called on a non-const reference, so the lazy `make_unique` needs no `mutable`.

One correction to the reasoning, not to the code. The argument that `storey_hlr.insert({ storey, hlr_engine(...) })` is safe because "that engine is freshly constructed, so its multimap is empty" is a runtime property standing in for a compile-time one — whether the map happens to be empty has no bearing on whether `std::multimap`'s copy constructor gets instantiated. What actually makes that line safe is that the braced temporary is an rvalue, so `insert(value_type&&)` is selected and `hlr_engine`'s implicit move constructor runs. The copy constructor is never instantiated.

That distinction matters for whoever touches this next, because `std::multimap` declares its copy constructor unconditionally instead of constraining it, so `std::is_copy_constructible_v<hlr_engine>` is still `true` after this change. The engine stays nominally copyable, and a stray `auto e = *hlr;` would fail deep inside `<map>` on `unique_ptr` rather than at one of your asserts. That is not a defect in this PR — the three asserts guard the type that can actually be guarded — but the engine-level guarantee someone might assume follows from them is not there.

Re-checked against `v0.9.0` at `dc69c19aa7`, with the assistance of an AI coding tool.

BIMvoice:
You are right, and thank you for writing the correction out properly rather than letting it stand.

My bullet was wrong twice, not once. I called `storey_hlr.insert({ storey, hlr_engine(...) })` "the one place that copies a whole `prefiltered_hlr`", when it is not a copy at all: the braced temporary is an rvalue, `insert(value_type&&)` is selected, and the engine is moved. Then I justified the safety with a runtime property, the map being empty, which as you say has no bearing on which constructor gets instantiated. I have corrected that bullet in the PR body and pointed at your comment.

I checked your second point rather than just accepting it, because it is the one that outlives this PR. Reduced to the shape of our types:

```cpp
struct NonCopyable { std::unique_ptr<int> p; };
struct EngineLikeOurs { std::multimap<double, NonCopyable> faces_; };

static_assert(!std::is_copy_constructible_v<NonCopyable>);   // holds
static_assert( std::is_copy_constructible_v<EngineLikeOurs>); // also holds
```

Both assertions pass on clang with `-std=c++17`. So you are right: `std::multimap` declares its copy constructor unconditionally, the trait keeps reporting `true`, and a stray `auto e = *hlr;` would fail inside `<map>` on the `unique_ptr` rather than at any of the three asserts. The guarantee a reader might infer from them stops at `face_info` and does not reach the engine.

Closing that gap is possible, and there is one trap in it. Deleting the copy constructor alone also suppresses the implicit move constructor, which would break the insert above. All four have to be declared:

```cpp
prefiltered_hlr(const prefiltered_hlr&) = delete;
prefiltered_hlr& operator=(const prefiltered_hlr&) = delete;
prefiltered_hlr(prefiltered_hlr&&) = default;
prefiltered_hlr& operator=(prefiltered_hlr&&) = default;
```

I compiled that against the reduction: copy blocked, move preserved, and `insert({key, Engine{}})` still compiles.

I have deliberately not pushed it. You framed this as not a defect in this PR and I agree, and adding four special member declarations to widen the guarantee is a scope decision rather than a fix. Happy to add it here if @aothms wants the engine-level guarantee to be real, or to leave this PR as the narrow ownership change and let your comment stand as the note for whoever touches it next. Either is fine by me.

Separately, the red checks on this PR were the `v0.9.0` baseline being broken between 2026-08-28 and 2026-09-02, not this diff. I have merged current `v0.9.0` in so they re-run against a green base.

PR #9430, opened 2026-09-02
Address tests

aothms:
@BIMvoice let's iterate here until tests are green


BIMvoice:
The ifctester skip can come off already: that failure was the tuple-versus-list guard in the chained facet filters, fixed in #9388 which merged shortly before this PR. I re-ran it on a native v0.9.0 build at `d11440af`, where `by_type` does return a tuple, and it passes. #9434 removes the skip.

On `test_package`: the `macos64` entry will not come back when builds resume, since Intel Mac builds were dropped from `build_osx.yml` in June (`45e9f763c8`) while `SUPPORTED_PLATFORMS` still lists it. The Windows half is #9423, where no Windows binary build has run since the plugin-collection fix landed and `BUILD_COMMIT` still pins `ad113e1`, so those packages ship without the schema plugins. Happy to take either of those if useful.

BIMvoice:
Two corrections to my earlier comment here, both mine.

First, I said `SUPPORTED_PLATFORMS` still lists `macos64`. It does not: @Andrej730 had already removed it in `66126d21bc`, several hours before I posted. I was relaying a reading of the file taken earlier in the day and did not re-check it against HEAD. Second, I described the work as removing a skip, but the skip had already been replaced with a non-strict `xfail` in `5aaf2107a5`. Sorry for the noise on both counts.

On the substance, the platform half is now satisfied: all 15 `ifcopenshell-python-{310..314}-...` zips and the 3 `IfcConvert` zips return 200 at `BUILD_COMMIT:=ad113e1`.

The WASM half had a second defect underneath the missing build. `WASM_TEMPLATE` interpolates `BINARY_VERSION` verbatim (`0.9.0alpha0`), but pyodide wheels are named with the normalised PEP 440 version (`0.9.0a0`), which is what `pyodide/build_pyodide.sh` stamps in. That URL could never resolve, at any pinned commit. #9443 fixes it, and the same defect in `pyodide/pack_wheel.py::get_wheel_url`.

The xfail still has to stay, because no pyodide wheel exists for `ad113e1`. Worth flagging separately: the bucket has `0.9.0a0+4e887e1` from 10 August, then two newer builds, `05e5a37` and `75ec717`, both tagged `pyemscripten_2025_0_wasm32` rather than `pyodide_2025_0_wasm32`. `pack_wheel.py` sets `USE_LEGACY_PLATFORM=1` to keep the old tag, citing pyodide/pyodide#6177, but `build_pyodide.yml` runs `build_pyodide.sh` directly, which does not set it. So the two most recent WASM builds are published under a name nothing consumes, which looks unintentional.

To get this green: run the WASM build on the commit that will be pinned, with the legacy tag, then bump `BUILD_COMMIT`. The xfail condition is keyed to `ad113e1` and clears itself on the bump.

PR #9441, opened 2026-09-02
Bonsai: resolve IFC save path to absolute before writing (fixes relative-path save failure)

### LINKED ISSUE #9440: Bonsai: "Save IFC" fails with relative path when Use Relative Path + autosave are enabled (writes to CWD)
### Summary

When **Use Relative Path** (`use_relative_project_path`) is enabled for an IFC project, saving can fail with:

```
RuntimeError: Failed to write to path: '<ProjectName>.ifc', check folder and file permissions.
```

The path in the error is **relative** (bare filename, no directory). It is not an actual permissions problem — the exporter is writing to a relative path, which the IfcOpenShell writer resolves against Blender's current working directory (typically not writable) instead of the project folder.

### Steps to reproduce

1. Open/save an IFC project and enable **Use Relative Path** on save (this stores `BIMProperties.ifc_file` as a path relative to the `.blend`).
2. Let the **autosave reminder** fire (or otherwise trigger a save that goes through `bim.save_project` via `EXEC_DEFAULT` rather than the file dialog).
3. The save fails with the error above; the IFC is not written.

### Root cause

`ExportIFC` (`bim.save_project`) only resolves the stored path to absolute inside `invoke()`:

```python
self.filepath = str(tool.Blender.ensure_blender_path_is_abs(Path(filepath)))
```

…and in the empty-string fallback in `_execute()`. But `_execute()` then writes `output_file` (derived straight from `self.filepath`) **without** an absolute-resolution step:

```python
output_file = bpy.path.ensure_ext(self.filepath, ".ifc")
output_file = Path(output_file).as_posix().replace("\\", "/")
settings = export_ifc.IfcExportSettings.factory(context, output_file, logger)
...
ifc_exporter.export()   # -> file.write(output_file)
```

Two ways a **relative** value reaches `_execute()` and bypasses the `invoke()` resolution:

1. **Autosave reminder** calls the operator with `EXEC_DEFAULT`, which skips `invoke()`:
   ```python
   current_ifc_path = props.ifc_file   # relative when use_relative_project_path is on
   bpy.ops.bim.save_project("EXEC_DEFAULT", filepath=current_ifc_path, should_save_as=False, skip_recent=True)
   ```
   The empty-string fallback does not trigger (the path is non-empty), so the relative path is written as-is.
2. The `filepath` operator property has no `SKIP_SAVE`, so a remembered relative value from a prior call can leak into a plain `execute()` (no `invoke()`).

Because `use_relative_project_path` re-stores the path relative after every successful save, the project is left in a state where the next `EXEC_DEFAULT`/autosave save fails.

### Proposed fix

Resolve `output_file` to an absolute path in `_execute()` before building the export settings, so the actual write target is always absolute regardless of how the operator was invoked. The relative form is still re-derived afterwards for storage:

```python
output_file = Path(output_file).as_posix().replace("\\", "/")
# Always resolve to absolute for the actual write; a relative stored path
# (use_relative_project_path) would otherwise be written against Blender's CWD.
output_file = Path(tool.Blender.ensure_blender_path_is_abs(Path(output_file))).as_posix().replace("\\", "/")
settings = export_ifc.IfcExportSettings.factory(context, output_file, logger)
```

### Environment

- Bonsai on Blender 5.2.1 LTS, Windows 10, Python 3.13
- Affects **v0.9.0** (verified: both the relative-path feature and the autosave reminder that calls `save_project` via `EXEC_DEFAULT` are present, and `_execute()` does not resolve the write path to absolute).

### Actual traceback

```
File ".../bonsai/bim/module/project/operator.py", in _execute
    ifc_exporter.export()
File ".../bonsai/bim/export_ifc.py", line 60, in export
    self.file.write(self.ifc_export_settings.output_file)
File ".../ifcopenshell/file.py", line 1029, in write
    self.wrapped_data.write(str(path))
File ".../ifcopenshell/ifcopenshell_wrapper.py", line 5951, in write
    return _ifcopenshell_wrapper.file_write(self, fn)
RuntimeError: Failed to write to path: 'Restaurant_E_Washington.ifc', check folder and file permissions.
```

theoryshaw:
Closes #9440.

### Problem

When **Use Relative Path** (`use_relative_project_path`) is enabled, `bim.save_project` can fail with:

```
RuntimeError: Failed to write to path: '<ProjectName>.ifc', check folder and file permissions.
```

The write target is **relative** (bare filename), so the IfcOpenShell writer resolves it against Blender's current working directory instead of the project folder. It is not an actual permissions problem.

### Root cause

`ExportIFC` only resolves the stored path to absolute inside `invoke()` (and the empty-string fallback in `_execute()`). But a relative value can reach `_execute()` and be written as-is:

- The autosave reminder calls `save_project("EXEC_DEFAULT", filepath=current_ifc_path, …)` — `EXEC_DEFAULT` skips `invoke()`, and `current_ifc_path` is relative when `use_relative_project_path` is on. The empty-string fallback doesn't fire (the path is non-empty).
- `self.filepath` has no `SKIP_SAVE`, so a remembered relative value can also leak into a plain `execute()`.

Since the relative path is re-stored after every save, the project is left in a state where the next `EXEC_DEFAULT`/autosave save fails.

### Fix

Resolve `output_file` to absolute in `_execute()` before building the export settings, so the actual write is always absolute regardless of how the operator was entered. The relative form is still re-derived afterwards for storage, so `use_relative_project_path` behavior is unchanged.

### Notes

- Verified the same buggy path exists on both `v0.8.0` and `v0.9.0`.
- Reported on Blender 5.2.1 LTS / Windows 10 / Python 3.13.

### DIFF
--- a/src/bonsai/bonsai/bim/module/project/operator.py
+++ b/src/bonsai/bonsai/bim/module/project/operator.py
@@ -2096,6 +2096,14 @@ class ExportIFC(bpy.types.Operator, ExportHelper):
         else:
             output_file = bpy.path.ensure_ext(self.filepath, ".ifc")
         output_file = Path(output_file).as_posix().replace("\\", "/")
+        # The stored IFC path can be relative (use_relative_project_path stores it that
+        # way after each save). invoke() resolves it to absolute, but EXEC_DEFAULT saves
+        # that pass filepath= directly (e.g. autosave) skip invoke(), and self.filepath
+        # has no SKIP_SAVE so a remembered relative value can leak into execute() too. A
+        # relative path here would be written against Blender's CWD and fail, so always
+        # resolve to absolute for the actual write; the relative form is re-derived for
+        # storage below.
+        output_file = Path(tool.Blender.ensure_blender_path_is_abs(Path(output_file))).as_posix().replace("\\", "/")
 
         settings = export_ifc.IfcExportSettings.factory(context, output_file, logger)
         settings.json_version = self.json_version
--- a/src/bonsai/bonsai/bim/module/project/operator.py
+++ b/src/bonsai/bonsai/bim/module/project/operator.py
@@ -2261,6 +2261,9 @@ class AutosavePrompt(bpy.types.Operator):
     bl_options = set()
 
     def invoke(self, context, event):
+        # Mark a reminder as open so the re-arming autosave timer won't stack
+        # another dialog on top while this one is still waiting for the user.
+        tool.Autosave.set_prompt_open(True)
         return context.window_manager.invoke_props_dialog(
             self, width=400, confirm_text="Save", title="Autosave Reminder"
         )
@@ -2271,6 +2274,7 @@ class AutosavePrompt(bpy.types.Operator):
         layout.label(text="Would you like to save your IFC project now?")
 
     def execute(self, context):
+        tool.Autosave.set_prompt_open(False)
         # Get current IFC path
         props = tool.Blender.get_bim_props()
         current_ifc_path = props.ifc_file
@@ -2289,6 +2293,7 @@ class AutosavePrompt(bpy.types.Operator):
         return result
 
     def cancel(self, context):
+        tool.Autosave.set_prompt_open(False)
         tool.Autosave.reset_timer()
         return {"CANCELLED"}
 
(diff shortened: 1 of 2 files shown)


theoryshaw:
@sboddy can you take a look?  Please pull, if it looks good to you.

Issue #6942, opened 2025-07-24
Instance #304 not found


ktyam301:
### Bug Description
```
Successfully loaded existing cache: 7631d237c540569cbb21339ab8f96aeb.h5.
Error: Cannot remove profile as it's still part of other IFC entities:
- IfcMaterialProfile
Writing homefile: "C:\Users\windows\AppData\Roaming\Blender Foundation\Blender\4.5\config\startup.blend" ok
Info: Startup file saved
Error: Cannot remove profile as it's still part of other IFC entities:
- IfcMaterialProfile
Starting export
Successfully loaded existing cache: f80ce50a8526af8f6d6404e7848422f4.h5.
Export finished in 0.01 seconds
Starting export
Successfully loaded existing cache: 16b88be0380dbee5b92d9cb1ca5590d2.h5.
Export finished in 0.01 seconds
Info: Saved "Test.blend"
Info: IFC Project "Test.ifc" And Current Blend File Are Saved
Starting export
Successfully loaded existing cache: 46ff645540789dd9bdb746c9f73ec114.h5.
Export finished in 0.02 seconds
Info: IFC Project "Test.ifc"  Saved
Writing homefile: "C:\Users\windows\AppData\Roaming\Blender Foundation\Blender\4.5\config\startup.blend" ok
Info: Startup file saved
Starting export
Successfully loaded existing cache: fb7c6b911cc6df84c27209a44333420d.h5.
Export finished in 0.01 seconds
Info: Saved "Test.blend"
Info: IFC Project "Test.ifc" And Current Blend File Are Saved
Starting export
Successfully loaded existing cache: 9a0d17ad0c56c79cc5d8772715c73a85.h5.
Export finished in 0.00 seconds
Info: Saved "Test.blend"
Info: IFC Project "Test.ifc" And Current Blend File Are Saved
Starting export
Successfully loaded existing cache: 6428fa889029369d1affbf63e30274dd.h5.
Export finished in 0.00 seconds
Starting export
Successfully loaded existing cache: 348976fba1239f2034013f6f38ee99df.h5.
Export finished in 0.01 seconds
Info: Saved "Test.blend"
Info: IFC Project "Test2.ifc" And Current Blend File Are Saved
Starting import process :: 0.00
Loading file :: 0.00
Calculate unit scale :: 0.00
Process context filter :: 0.00
Calculate model offset :: 0.00
Set units :: 0.00
Create project :: 0.00
Process element filter :: 0.00
[...]

PR #9448, opened 2026-09-04
Bonsai: guard stale representation item ids in the Representation Items panel (#9447)

### LINKED ISSUE #9447: Bonsai: Representation Items panel crashes with "Instance #N not found" after a .blend/.ifc round trip
### Bug Description

The Representation Items panel stores its item list in the .blend, but never checks that the ids are still valid against the loaded IFC. After a .blend/.ifc round trip (open the .blend, which reloads the .ifc next to it) the list can be replayed against a different state of the file, and the first button that touches an item raises:

```
  File "...\bonsai\tool\ifc.py", line 367, in execute
  File "...\bonsai\bim\ifc.py", line 569, in execute_ifc_operator
  File "...\bonsai\bim\module\geometry\operator.py", line 3586, in _execute
    representation_item = ifc_file.by_id(representation_item_id)
  File "...\ifcopenshell\file.py", line 836, in by_id
    return self[id]
  File "...\ifcopenshell\file.py", line 823, in __getitem__
    return entity_instance(self.wrapped_data.by_id(key), self)
  File "...\ifcopenshell\ifcopenshell_wrapper.py", line 5826, in by_id
    return _ifcopenshell_wrapper.file_by_id(self, id)
RuntimeError: Instance #1937489 not found
```

Opening the same .ifc directly (no .blend) does not produce the error - the list is then rebuilt from the file that was just loaded, so every id resolves.

### Cause

`BIMObjectGeometryProperties.items` is a `CollectionProperty` on the **object** (`bim/module/geometry/prop.py`), so `is_editing` and every `items[].ifc_definition_id` are written into the .blend. `EnableEditingRepresentationItems` is the only thing that rebuilds them. Nothing invalidates that list when the IFC is (re)loaded, so a list captured in one session can outlive the instances it points at, and the operators then call `by_id()` on a dead id with no guard.

### Affected call sites

All on `v0.9.0` (checked at e18d927b40), `src/bonsai/bonsai/bim/module/geometry/operator.py`:

| line | operator |
| --- | --- |
| 3275 | `RemoveRepresentationItem` |
| 3301 | `SelectRepresentationItem` |
| 3370 | `EnableEditingRepresentationItemStyle` (the traceback above) |
| 3396 | `EditRepresentationItemStyle` |
| 3522 | `EditRepresentationItemShapeAspect` |
| 3595 | `RemoveRepresentationItemFromShapeAspect` |
| 4200 | `EditRepresentationItemLayer` |
| 4228 | `UnassignRepresentationItemLayer` |

`UnassignRepresentationItemStyle` already resolves through `tool.Ifc.get_entity_by_id` and reports instead of raising, but it leaves the stale list in place, so the next click fails the same way.

### Suggested fix

Route these through the existing None-returning accessor `tool.Ifc.get_entity_by_id` in a small helper that, on a dead id, reports the error and rebuilds the list with the `disable_editing_representation_items` / `enable_editing_representation_items` pair that these operators already use as "reload items ui". The panel then heals itself: the user gets a message instead of a traceback, a list matching the loaded IFC, and the click works on the retry.

That is a lazy fix at the point of use. The underlying condition - per-object representation-item editing state surviving an IFC load - is still there, and the fuller cure would be to reset `is_editing`/`items` on every object when a project loads. Happy to open a PR for the guards if that direction is agreeable.

### Related

- #6942 - same class of bug (stale Blender-side IFC id after saving .blend + .ifc together), but in `tool.Geometry.get_active_representation` reading the *mesh* props rather than the items list. Closed as stale in Sept 2025 without a fix; would still crash after the guards above.
- #8924 (open) is the only open PR working on this list - it syncs `active_item_index` with the viewport but does not validate the ids, so it inherits the same crash.

### Environment

```
os: Windows 10 (10.0.19045)
blender_version: 5.2
python_version: 3.13
bonsai: local build off v0.8.0; the affected code is identical on v0.9.0 @ e18d927b40
```

theoryshaw:
Closes #9447.

### The bug

The Representation Items panel raises `RuntimeError: Instance #N not found` after a .blend/.ifc round trip - open the .blend, which reloads the .ifc beside it, then click any button in the panel:

```
  File "...\bonsai\bim\module\geometry\operator.py", line 3370, in _execute
    representation_item = ifc_file.by_id(representation_item_id)
RuntimeError: Instance #1937489 not found
```

Opening the same .ifc directly does not show it, because the list is then rebuilt from the file that was just loaded.

### Cause

`BIMObjectGeometryProperties.items` is a `CollectionProperty` on the object, so `is_editing` and every `items[].ifc_definition_id` are written into the .blend. `EnableEditingRepresentationItems` is the only thing that rebuilds them, and nothing invalidates the list when the IFC is (re)loaded - so a list captured in one session can outlive the instances it points at, and the operators call `by_id()` on a dead id with no guard.

### The change

Adds `resolve_representation_item()`, built on the existing None-returning accessor `tool.Ifc.get_entity_by_id`. On a dead id it reports the error and rebuilds the list using the `disable_editing_representation_items` / `enable_editing_representation_items` pair these operators already use as "reload items ui". The panel heals itself: a message instead of a traceback, a list matching the loaded IFC, and the click works on the retry.

Eight call sites now route through it: `RemoveRepresentationItem`, `SelectRepresentationItem`, `EnableEditingRepresentationItemStyle`, `EditRepresentationItemStyle`, `EditRepresentationItemShapeAspect`, `RemoveRepresentationItemFromShapeAspect`, `EditRepresentationItemLayer`, `UnassignRepresentationItemLayer`. `UnassignRepresentationItemStyle` already reported instead of raising, but left the stale list in place; it now refreshes too.

### Scope

This guards the point of use. The underlying condition - per-object representation-item editing state surviving an IFC load - is unchanged, and the fuller cure would be resetting `is_editing`/`items` on every object when a project loads. That is an O(objects) sweep on load, so I left it out of this PR; happy to add it if you would rather fix it there.

Note that #6942 is the same class of bug in a different call site (`tool.Geometry.get_active_representation` reading the *mesh* props) and is not covered here.

### Testing

Verified by hand in Blender on the file that reproduced the crash: the traceback is gone, the panel refreshes, and the style edit applies on the retry. No automated test added - reproducing it needs a .blend whose stored item ids disagree with the .ifc beside it.

### DIFF
--- a/src/bonsai/bonsai/bim/module/geometry/operator.py
+++ b/src/bonsai/bonsai/bim/module/geometry/operator.py
@@ -3247,6 +3247,27 @@ class DisableEditingRepresentationItems(bpy.types.Operator, tool.Ifc.Operator):
             props.is_editing = False
 
 
+def resolve_representation_item(
+    operator: bpy.types.Operator, item_id: int
+) -> Union[ifcopenshell.entity_instance, None]:
+    """Resolve an item id from the representation items UI list.
+
+    ``BIMObjectGeometryProperties.items`` is stored in the .blend, so its ids can
+    outlive the instances they point to - reopening a .blend next to its .ifc
+    replays a list that was built against a different state of the file. Refresh
+    the list and report instead of letting ``by_id`` raise
+    ``RuntimeError: Instance #N not found``.
+    """
+    if item := tool.Ifc.get_entity_by_id(item_id):
+        return item
+    operator.report(
+        {"ERROR"}, f"Representation item #{item_id} no longer exists, items list has been refreshed - try again."
+    )
+    bpy.ops.bim.disable_editing_representation_items()
+    bpy.ops.bim.enable_editing_representation_items()
+    return None
+
+
 class RemoveRepresentationItem(bpy.types.Operator, tool.Ifc.Operator):
     bl_idname = "bim.remove_representation_item"
     bl_label = "Remove Representation Item"
@@ -3270,9 +3291,9 @@ class RemoveRepresentationItem(bpy.types.Operator, tool.Ifc.Operator):
     def _execute(self, context):
         assert (obj := tool.Geometry.get_active_or_representation_obj())
         assert (element := tool.Ifc.get_entity(obj))
-        ifc_file = tool.Ifc.get()
 
-        representation_item = ifc_file.by_id(self.representation_item_id)
+        if not (representation_item := resolve_representation_item(self, self.representation_item_id)):
+            return {"CANCELLED"}
         tool.Geometry.remove_representation_item(representation_item, element)
         tool.Geometry.reload_representation(obj)
 
@@ -3298,7 +3319,8 @@ class SelectRepresentationItem(bpy.types.Operator):
         obj = tool.Geometry.get_active_or_representation_obj()
         obj_props = tool.Geometry.get_object_geometry_props(obj)
         assert obj_props.active_item
-        item = tool.Ifc.get().by_id(obj_props.active_item.ifc_definition_id)
+        if not (item := resolve_representation_item(self, obj_props.active_item.ifc_definition_id)):
+            return {"CANCELLED"}
         item_ids = self.get_nested_item_ids(item)
 
         props = tool.Geometry.get_geometry_props()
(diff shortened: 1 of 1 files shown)


Moult:

Thanks Ryan, the report in #9447 and the inventory of the eight call sites were exactly what was needed. The maintainers traced it one layer further down: the list is a cached snapshot that nothing invalidates, and the same crash happens within a single session, without any save or reload, because editing an item in item mode replaces it with a new one. So the fix in #9546 makes the list subscribe to IFC changes and rebuild itself, which also lets the eight disable/enable workarounds go, rather than guarding each call site. Closing this in favour of #9546.

PR #9449, opened 2026-09-04
ifctester: gate facet filters on the presence of a population, not its type

RickBrice:
Entity.filter() returned ifc_file.by_type()'s result unmodified when no predefinedType was given. That result is not guaranteed to be a list (some ifcopenshell builds return a tuple), and every other facet's filter() decides whether to narrow the population handed to it, or discard it and rescan the entire model, based on isinstance(elements, list). A tuple fails that check, so the very next facet in an <ids:applicability> silently ignored the entity restriction and matched any class sharing the same attribute, classification, material, etc. value.

Observed in practice: a specification scoped to
entity(IFCMATERIAL) + attribute(Name=X) also matched an unrelated IfcTendonType that happened to share Name X with the IfcMaterial, failing the specification because the IfcTendonType lacks Pset_MaterialSteel.

Adds a regression test that forces by_type() to return a tuple so it reproduces independent of what the local ifcopenshell build normally returns.

An example is shown here:
IfcTester_Error

Attached are the source IFC file and the IDS. You may think the IDS is a bit weird (and it is). I am experimenting with creating an IDS that tests the outputs of a structural bridge design are retained in the IFC model as the model progresses through project development.

[PGSuper_Import_Model.ifc.txt](https://github.com/user-attachments/files/31846475/PGSuper_Import_Model.ifc.txt)

[PGSuper_Import_Model.ids.txt](https://github.com/user-attachments/files/31846496/PGSuper_Import_Model.ids.txt)

### DIFF
--- a/src/ifctester/ifctester/facet.py
+++ b/src/ifctester/ifctester/facet.py
@@ -227,7 +227,14 @@ class Entity(Facet):
                     continue
         if self.predefinedType:
             return [r for r in results if self(r)]
-        return results
+        # `results` may come straight from ifc_file.by_type(), which is not guaranteed
+        # to be a `list` (e.g. it can be a `tuple`). Every other Facet.filter() gates
+        # on `isinstance(elements, list)` to decide whether a population has already
+        # been computed for it to narrow, versus starting a fresh, entity-type-agnostic
+        # scan of the whole model. Returning anything but a `list` here defeats that
+        # gate for the very next facet in the applicability, silently discarding this
+        # entity restriction instead of intersecting with it.
+        return list(results)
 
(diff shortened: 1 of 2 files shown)


Moult:

Thanks Rick, the report and the reproduction were exactly right. The maintainers went one layer deeper: the gate in every facet's `filter()` was asking about the *type* of the population when the question is whether one *exists*, and Petru's earlier widening to `(list, tuple)` still left that trap armed. Pushed to this branch: all six gates now read `elements is not None`, the signatures take and return a `Sequence`, and no copies are added. Your scenario is kept as a behaviour test in `test_ids.py` (no monkeypatching needed, since 0.9's `by_type` returns a tuple natively), plus one asserting that an empty population is not rescanned. You are credited as co-author. Merging once CI is green.

PR #9453, opened 2026-09-06
Gpu snap to v0.9.0

brunoperdigao:


brunoperdigao:
@Andrej730 I'm not being able to merge this PR because of those checks that are pending and never run. Is there something I should do here?

BIMvoice:
`ci` no longer runs here because this PR only touches `src/bonsai/**`. `403a23a976` (3 September) changed `pull_request:` to reuse the `push` paths filter, and `src/bonsai/**` is not in that list, so the workflow is skipped rather than failing. Before that commit `pull_request:` had no filter and ran on every PR, which is why this is new. A skipped workflow never reports a status, so a required `compile-and-test` stays pending indefinitely.

Three ways out, depending on what was intended: add `src/bonsai/**` to `ci_paths`, which runs the full C++ build on every Bonsai change and is probably not the goal; drop the filter from `pull_request` while keeping it on `push`, restoring the previous behaviour; or keep the filter and make the required check a job that always runs. `v0.8.0` is unaffected, its `pull_request:` still has no filter.

I have opened #9454 for the second option since it is the minimal restore, but happy to redo it as either of the others.

PR #9454, opened 2026-09-06
ci: run on every pull request again (fixes the merge block in #9453)

BIMvoice:
Fixes the merge block reported in #9453.

`403a23a976` changed the `pull_request` trigger from having no paths filter to reusing the `push` one:

```yaml
  push:
    paths: &ci_paths
      ...
  pull_request:
    paths: *ci_paths
```

`src/bonsai/**` is not in that list, so `ci` is now skipped entirely for any pull request that touches only Bonsai. A skipped workflow reports no status at all, rather than a passing or failing one, so a required `compile-and-test` never arrives and the pull request cannot be merged. That is what #9453 hit, and it applies to every Bonsai only pull request against `v0.9.0`.

This restores the previous behaviour for pull requests and leaves the `push` filter untouched, so pushes to branches still skip the heavy build when nothing relevant changed. The anchor stays where it is, since it is still the definition used by `push`.

Two alternatives, if either is closer to the intent:

- add `src/bonsai/**` to `ci_paths`, which keeps the filter symmetric but runs the full C++ build on every Bonsai change
- keep the filter and make the required status check a job that always runs, reporting success when the heavy jobs are correctly skipped

I am happy to redo it either way. `v0.8.0` is unaffected: its `pull_request:` still has no filter.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,10 @@ on:
       - 'cmake/**'
       - 'test/**'
       - '.github/workflows/ci.yml'
+  # No paths filter here on purpose. A skipped workflow reports no status,
+  # so a required check would stay pending forever on any PR the filter
+  # excludes (e.g. src/bonsai only). The push filter above still applies.
   pull_request:
-    paths: *ci_paths
 
 jobs:
   activate:


aothms:
I'm a bit lost what to do here. We either have more granular workflows and cache the C++ builds - so that we can run e.g only bonsai tests with the latest builds when the changed paths tell us that that is enough. Or we just blindly enable the full CI workflow as suggested on PR. Or we revisit the required status checks configs (but people are complaining here about the lack of flexibility - it's not just us). Admittedly I did not give this enough thoughts, I'm also fine with disabling required status checks if it's simply too painful.

BIMvoice:
Some numbers, since they narrow this a lot.

**The full `ci` run is already cheap.** Over the last 14 completed runs it is 5 to 6 minutes typical, 12 minutes worst case. The reason is that the C++ builds are already cached: `hendrikmuhs/ccache-action` has been in `ci.yml` since December 2025 and every configure passes `CMAKE_{C,CXX}_COMPILER_LAUNCHER=ccache`.

That is the part of your first option that already exists. Splitting into more granular workflows would still leave you paying most of those 5 minutes for checkout, dependency restore and the Python suites, so the saving is a couple of minutes per run in exchange for a permanent increase in workflow complexity and a second place for path lists to drift out of sync. That trade did not look worth it to me, but you know the runner budget better than I do.

**How often it would fire:** of the 60 most recently opened pull requests, 13 touch only paths outside the filter. So enabling it for pull requests costs roughly one extra 6 minute run per five pull requests, and in exchange those 13 stop being unmergeable.

So my suggestion is this PR as the immediate unblock, on the grounds that the caching work you already did is what makes it affordable.

If you would rather keep the filter, the usual pattern for the required check problem is a small always-running job whose only purpose is to report the required status, succeeding when the heavy jobs were legitimately skipped. That keeps the savings and keeps branch protection meaningful, at the cost of one more job definition. Happy to send that instead if you prefer it, and equally happy for this PR to be closed if you decide to relax the required checks instead.

brunoperdigao:
Hi guys, any news about this? Is there anything I can do on my side in the meantime?

BIMvoice:
@brunoperdigao here is exactly why you are stuck, and the honest answer on what you can do.

All twelve files in #9453 are under `src/bonsai/`, and `src/bonsai/**` is not in `ci_paths`. So the `ci` workflow never starts for your pull request, the `compile-and-test` contexts never report, and branch protection waits forever. Your head commit has exactly one check run on it, `lint-formatting`, which comes from `ci-lint` because that workflow does cover `src/bonsai`.

**There is nothing clean you can do from your side.** The only self-service trick is to include a change under a filtered path so the workflow fires, and you would have to leave it in the diff for the check to stay green, which is worse than waiting. This needs either #9454 merged or the required contexts relaxed.

**One thing I should say plainly about our own PR, because it would be easy to oversell it.** #9454 makes `ci` run on your pull request, which satisfies the required contexts. It does not give your change any real test coverage. `ci` does not test Bonsai: its only reference to `src/bonsai` is a commented-out `IfcConvert` block at line 207. So #9454 unblocks the merge, it does not tell you the GPU snapping work is sound.

For that, the suite that actually exercises Bonsai is `ci-bonsai-daily`, and it does list `src/bonsai/**`, but it triggers on `push` to `v0.9.0` rather than on pull requests. So Bonsai changes are tested after they land, not before. `ci-bonsai` is `workflow_dispatch` only. That means no pull request workflow tests Bonsai code today, for anyone.

@aothms that reframes your first option a little. The choice is not really granularity versus cost, because the granular Bonsai suite already exists and already has the right path list. What is missing is a `pull_request` trigger on it. So:

* If you want the immediate unblock, #9454 is `+3/-1`, green on all four contexts, and mergeable. Worth noting there is already a precedent in the repository: `ci-lint.yaml` has a bare `pull_request:` trigger with no paths filter, which is why it is the one workflow that did report on #9453. #9454 only makes `ci` consistent with it. The numbers from last week still hold: roughly one extra six minute run per five pull requests, affordable because of the ccache work you already did.
* If you want pull request coverage for Bonsai, that is a separate and more interesting change, adding a `pull_request` trigger to `ci-bonsai-daily` or a trimmed variant of it. Happy to prepare that, and it would have caught things like #9453 before merge rather than after.
* If you would rather relax the required contexts, that also solves @brunoperdigao's problem today and I will close #9454 without argument.

I am not going to push a second workflow PR at you unasked given the review load, so say which of the three you want and I will do that one.

CyrilWaechter:
Acknowledging up front that this PR is correct and minimal, and it would fix the deadlock. Posting an alternative for the same problem so you can weigh both, not to argue against this one.

**Alternative: #9498** keeps the `pull_request` path filter that `403a23a` added, but moves it into a `changes` job and skips the heavy matrix job when nothing relevant changed. A job skipped by a conditional still reports success, so the required contexts are satisfied without running the build.

The two approaches side by side:

- **#9454**: drop the filter from `pull_request` (keep it on `push`). Every PR then runs the full `ci`, including docs- and Bonsai-only PRs, and the green `compile-and-test` on a Bonsai-only PR does not actually build or test that code.
- **#9498**: keep the filter, and skip the heavy job when it doesn't apply.

Verified numbers:

- Warm full `ci` run: ~5–6 min (run [34875353859](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/34875353859)); cold dependency rebuild: ~27–28 min.
- On the Bonsai-only #9499: `changes` ⇒ `false`, `compile-and-test` **skipped**, workflow **success**: run [34875531922](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/34875531922).

Why the filtered variant may be worth the extra job: it preserves `403a23a`'s intent (don't run the C++ build for PRs that can't affect it), avoids ~6 min (up to ~28 min cold) per Bonsai/docs-only PR, avoids piling more from-source dependency builds (rocksdb/OpenCOLLADA/manifold/swig) onto unrelated PRs where they occasionally flake, and a skipped check is more honest for a Bonsai-only change than a green build that didn't exercise it.

Cost of the alternative, to be explicit: one lightweight job and `dorny/paths-filter@v4`, plus reliance on skip-by-conditional ⇒ success (validated above). If the action is the sticking point, the detector can be swapped for GitHub-maintained `actions/github-script`.

The two PRs touch the same lines in `ci.yml`, so only one should land. Either way, #9499 adds the `pull_request`-triggered Bonsai test job that both options leave open — currently non-required because that suite has pre-existing failures (`test_drawing`, `test_project` ×2). Happy to close #9498 if you prefer the simpler route here.

aothms:
On a conceptual my vote goes to https://github.com/IfcOpenShell/IfcOpenShell/pull/9498 - it's the intent that we want (and probably need) and allows to keep building more granular cached workflows; maybe next up a granular workflow for only markdown changes - or e.g changes to ifcpatch requiring only ifcpatch tests etc. What I can't judge is whether this approach is durable and flexible enough, I didn't really want to look into that. You have my vote, if others do not pitch in with different ideas feel free to merge either way at any time.

BIMvoice:
Agreed, and closing this in favour of #9498.

@CyrilWaechter your approach is better than mine and I verified the part I was not willing to assert myself. On this thread I described the skip-and-report pattern but refused to claim it worked, because whether a conditionally skipped job satisfies a required context is exactly the sort of thing I did not want to state without evidence. You produced it: run 34875531922 on the Bonsai-only #9499 shows `compile-and-test = skipped` with the workflow concluding `success`. Your timing checks out too, run 34875353859 ran 17:32:20 to 17:38:22, so 6 minutes 2 seconds warm.

Your point that a skipped check is more honest than a green build that did not exercise the code is the same objection I raised against my own PR here, so it would be odd of me to argue the other way now. #9454 satisfied branch protection by running a workflow that does not test Bonsai at all.

#9499 is the piece that actually matters. No pull request workflow tests Bonsai today, for anyone, which is the real reason @brunoperdigao was stuck. I offered to build that here and deliberately did not push it unasked given the review load, so you doing it is the better outcome.

One thing I checked in case it helped, and it did not, recorded so nobody repeats the search. The three failures keeping `bonsai-tests` non-required are:

```
test/tool/test_drawing.py::TestDrawingMaintainingSheetPosition::test_run
test/tool/test_project.py::TestGettingLinkedElementGeomSlice::test_skip_hidden_first_element
test/tool/test_project.py::TestGettingLinkedElementGeomSlice::test_skip_hidden_middle_element
```

I went through our open PRs to see whether any already fix them. None do. #9132 touches `test_drawing.py` and is titled as fixing three tests, but they are three different vacuous ones in the same file, not these.

**Offer stands on those three.** If it would help get `bonsai-tests` to required, I am happy to take them, and that seems a more useful contribution than another workflow PR. @aothms your note about more granular cached workflows makes that worth more, since a Bonsai-only path is only useful if its suite can actually gate.

PR #9455, opened 2026-09-06
Fix border-half-edge bugs in `CgalShape::Triangulate()`

FrogsInSpace:
### Two bugs, both only hit on shapes with an open/border edge:

1. **Invalid iterator compare** — `facet_to_component[...]` used `operator[]`, which default-constructs a value for a border half-edge's (unmapped) facet. Comparing that against a real iterator is UB and trips MSVC's debug assert. This only ever surfaces in Debug builds. In the best case/release builds this produces smoothing artefacts near open shells. Fixed by checking `is_border()` first and switching to `.at()`. 
(See attached ifc sample file and screenshot)

2. **Circulator hang** — `++vh` was inside the `if (!vh->is_border())` block, so a border half-edge stopped the circulator from advancing, looping forever. Fixed by moving `++vh` outside the `if`.

### Testing

Re-ran on the IFC file that reproduced the assert/hang — confirmed fixed in Debug build, triangulation output unchanged on normal shapes.

[IfcSample_Asserts_CGAL.zip](https://github.com/user-attachments/files/31876226/IfcSample_Asserts_CGAL.zip)

<br>vector-iterators-incompatible-DebugAssert

### DIFF
--- a/src/ifcgeom/kernels/cgal/CgalConversionResult.cpp
+++ b/src/ifcgeom/kernels/cgal/CgalConversionResult.cpp
@@ -507,8 +507,8 @@ void ifcopenshell::geometry::CgalShape::Triangulate(ifcopenshell::geometry::Sett
 								if ((fn2 * facet_normal) >= *smooth_treshold) {
 									normal_accum = normal_accum + fn2;
 								}
-								++vh;
 							}
+							++vh;
 						} while (vh != vh_begin);
 					}
 				}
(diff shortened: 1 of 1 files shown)


aothms (approved):
Looks good to me. Thanks!

FrogsInSpace:
Hmm...  python linter complains about three python files i never touched ?

src/ifcopenshell-python/ifcopenshell/util/unit.py
src/ifcopenshell-python/test/api/alignment/test_create.py
src/ifcopenshell-python/test/api/alignment/test_referent_names.py

dont know whats the procedure.. 
of course i could fix them, but not how to proceed

aothms:
Ah, I see now. You're trying to merge into v0.8 but we're on v0.9.0 branch for some weeks now. Have you attempted to get v0.9 up and running? It's a bit different with plugins et al. I'll remove the status checks for v0.8 because they are not aligned with the expected shape of the v0.9 status checks.

FrogsInSpace:
Ahh .. thanks that explains it
Yes, started to adapt to  v0.9.0, but there is some work to be done because of the API changes
BTW: the fix should apply to v0.9.0 too, as that  cgal kernel part did not change

aothms:
> BTW: the fix should apply to v0.9.0 too, as that cgal kernel part did not change

Agreed, we'll have to remember to merge back v0.8 into v0.9

PR #9459, opened 2026-09-08
docs: write the five empty IFC concept sections (#9456)

### LINKED ISSUE #9456: Missing documentation about IFC Concepts
### Bug Description

In https://docs.ifcopenshell.org/introduction/introduction_to_ifc.html at the bottom of the page, there are 5 headings with 5 concepts about IFC but all of which are empty.

### Attachments

_No response_

### Debug and Error Output

```true

```

BIMvoice:
Fixes #9456.

At the bottom of `introduction_to_ifc.rst` there are five `Concept N` headings plus `Self-learning IFC: how to learn more`, all of them empty. They have been empty since `cffc8b519c` on 2023-09-27, where they were added as placeholders. Nothing was lost; they were never written.

The reporter counted five. There are actually six, because `Self-learning IFC: how to learn more` is empty too. This PR writes all six.

### How the content was produced

Every printed value in the new sections comes from actually running the code against `AC20-FZK-Haus.ifc`, the sample the page already links to and already uses for its earlier examples. Nothing is illustrative or invented. For example, the spatial chain shown in Concept 2:

```
IfcWallStandardCase   Wand-Int-ERDG-4
IfcBuildingStorey     Erdgeschoss
IfcBuilding           FZK-Haus
IfcSite               Gelaende
IfcProject            Projekt-FZK-Haus
```

was produced by walking `get_container` then `get_aggregate` on that file.

The six sections cover, in the page's existing voice and using the `ifcopenshell.util.element` helpers a beginner will actually reach for:

1. **The project context.** One `IfcProject` per model, units, why unit scale matters before any number means anything, and representation contexts.
2. **Spatial decomposition.** The distinction between aggregation and containment, which is the part newcomers most often get wrong, with the chain above.
3. **Object typing.** Occurrence versus type, `get_type` and `get_types`. Includes a note that `by_type` includes subclasses, since in this model all 13 walls are `IfcWallStandardCase`.
4. **Attributes and property sets.** Why the two are different places to look, and that `Pset_` names are standardised while names like `ArchiCADProperties` are not.
5. **Material assignment.** That `get_material` returns an `IfcMaterialLayerSetUsage` rather than a material, and how to reach the layers.
6. **Self-learning.** A concrete method for exploring an unfamiliar model, plus a runnable snippet.

### A second defect fixed on the same page

While validating the file I found that the existing `seealso` at line 63 references `` `Core functionality crash course`_ ``, which has no target anywhere in the documentation. docutils reports `Unknown target name` and it renders as literal text rather than a link. It now points at the `Hello, world!` page with a `:doc:` role.

### Verification

Parsed with docutils, with the Sphinx-only directives and roles stubbed:

```
BEFORE (v0.9.0):  2 messages, both: (ERROR/3) Unknown target name: "core functionality crash course".
AFTER:            0 messages
```

I was not able to run a full Sphinx build locally, because that needs `sphinx-autoapi` to introspect the whole package. So this is a docutils structural check, not a rendered-output check. If a maintainer has a docs preview build to hand it is worth a look at the rendering, particularly the `warning`, `tip` and `note` admonitions.

Note that `ci.yml` excludes `src/ifcopenshell-python/docs/**` from its paths filter, so this PR will not trigger any CI run. That is expected and not a sign of anything being skipped incorrectly.

Opened as a draft because the wording is a judgement call and this is a beginner-facing page. Happy to cut, expand, or change the tone of any section. If you would rather the concepts stay short and point out to the reference docs instead, say so and I will trim them right back.

### DIFF
--- a/src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
+++ b/src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
@@ -61,7 +61,7 @@ attributes, and explore relationships. Your journey begins here.
 .. seealso::
 
     If you are already familiar with IFC and just want to learn how to use
-    IfcOpenShell, you can jump to the `Core functionality crash course`_.
+    IfcOpenShell, you can jump to :doc:`../ifcopenshell-python/hello_world`.
 
 Begin learning IFC
 ------------------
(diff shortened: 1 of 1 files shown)


brunopostle:
> **Written with the aid of an LLM (Claude).** The verification described below was genuinely run, not asserted: the sample file was downloaded from the link the page gives, every snippet was executed against it, and both versions of the RST were parsed with docutils. Read it with the provenance in mind all the same.

I reproduced every claim in this PR against the actual sample file rather than reading it.

## Verdict

The factual content is sound. Worth merging after four small fixes; none are structural.

**Everything printed in the new sections reproduces exactly.** Project name, all eight `IfcSIUnit` entries, `calculate_unit_scale` → `1.0`, both contexts as `Model 1e-05, Plan 1e-05`, the wall `Wand-Int-ERDG-4` with GUID `2XPyKWY018sA1ygZKgQPtU`, the full Erdgeschoss → FZK-Haus → Gelaende → Projekt-FZK-Haus chain, `get_decomposition` → 58, 13 walls / 2 wall types, `get_types` → 5, `ThermalTransmittance` 1.5, `Width` 0.24, `IfcMaterialLayerSetUsage`, and the literal `#15069=IfcMaterialLayerSet((#15067),'Leichtbeton 102890359 0.24',$)`. Every kwarg named (`include_subtypes`, `should_skip_usage`) exists with that spelling in `ifcopenshell.util.element` / `ifcopenshell.file`.

Two things the PR does not claim but which make it better than it looks: the wall it picks is `#15042`, the same one already printed in the raw STEP excerpt further up the page, and the material is `#15046` from that same excerpt. The new sections are anchored to the file the page already shows you.

The secondary fix is real. `Core functionality crash course` appears exactly once in the entire repository — in the broken reference itself. And `hello_world.rst` describes itself as "This crash course guides you through basic code snippets... low-level functionality", so it is the intended target. The docutils check reproduces: error before, clean after. The `ci.yml` claim is literally true too (`- '!src/ifcopenshell-python/docs/**'`).

## What to fix before merge

**1. The `seealso` rewrite loses the sentence.** `you can jump to :doc:`../ifcopenshell-python/hello_world`.` renders as "you can jump to **Hello, world!**." — the original prose named what the destination *is*, and the replacement does not. Keep the words and hang the link on them:

```rst
IfcOpenShell, you can jump to the :doc:`core functionality crash course
<../ifcopenshell-python/hello_world>`.
```

**2. `get_decomposition` → 58 teaches the wrong thing.** Breaking down what is actually in that set:

```
IfcOpeningElement 14, IfcAnnotation 12, IfcWallStandardCase 9,
IfcWindow 9, IfcSpace 6, IfcDoor 5, IfcBeam 1, IfcSlab 1, IfcStair 1
```

Nearly half of the 58 are openings and annotations. A beginner reading "what a place holds… 58" will believe the ground floor has 58 building elements. The sentence says "everything below it, not only its direct children", which is true but does not warn about *this*. Say that openings and annotations are in there too. Also: it returns a `set`, so `[0]` on it raises — worth half a sentence.

**3. `get_psets` returns an `id` key.** `{'ThermalTransmittance': 1.5, 'id': 15079}`. The section says "returns them as a plain dictionary" and then indexes two known keys, so nothing printed is wrong — but the next thing a reader does is iterate it, and `id` is not a property. Concept 5 already does exactly this move well with "that result is not a material"; Concept 4 should do it too.

**4. `include_subtypes=False` is a trap as written.** In this model it returns **0**, because there are no direct `IfcWall` instances. The note has just said every wall is an `IfcWallStandardCase`, so it is derivable — but the section's own style is to print results, so print this one: `# 0`. It makes the point instead of springing it.

## Minor

- Concept 3's numbers do not close: "13 occurrences, 2 types" then `get_types` → 5. The other type has 8. Say "5 of the 13" or drop the count.
- The page never tells the reader that the printed values come from the downloaded sample. It is inferable from the earlier section, but Concept 1 re-opens `'AC20-FZK-Haus.ifc'` where the earlier block used `'/path/to/your/model.ifc'` — one clause would remove the doubt.
- In Concept 5 the expected output sits on its own line inside the loop body, which reads like a commented-out statement. Everywhere else on the page results go on the same line as the `print`.

## On the draft question

You ask whether to trim the concepts back to pointers. Don't. The five headings sat empty for two years precisely because "point at the reference docs" is what the reference docs already are. The aggregation-versus-containment distinction in Concept 2 and the usage-is-not-a-material point in Concept 5 are the two things newcomers reliably get wrong, and neither is discoverable from the class listings the page already links.

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
 Concept 1: the project context
 ------------------------------
 
+Every IFC model contains exactly one **IfcProject**. It is the root of the
```
aothms: This it not necessarily true: at most one https://standards.buildingsmart.org/IFC/DEV/IFC4_3/HTML/lexical/IfcSingleProjectInstance.html 0 or more IfcProjectLibraries are also supported.

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
+model, and it answers two questions that every other element depends on: what
+units the numbers are in, and what geometric contexts exist.
```
aothms:
```suggestion
model. It provides units to numbers in the model are in and geometric contexts.
```

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
+    project = model.by_type("IfcProject")[0]
+    print(project.Name) # Projekt-FZK-Haus
+
+    for unit in project.UnitsInContext.Units:
+        if unit.is_a("IfcSIUnit"):
+            print(unit.UnitType, unit.Name) # LENGTHUNIT METRE, AREAUNIT SQUARE_METRE, ...
+
+This matters more than it may look. A length of ``0.24`` means nothing until
+you know whether the model measures in metres or millimetres. IfcOpenShell can
+work this out for you:
```
aothms:
```suggestion
```

This is already mentioned, let's keep it a bit terse.

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
+
+.. code-block:: python
+
+    for context in project.RepresentationContexts:
+        print(context.ContextType, context.Precision) # Model 1e-05, Plan 1e-05
+
+.. warning::
+
+   A valid IFC model has exactly one **IfcProject**. If you are creating a
+   model from scratch, create it first, before anything else.
```
aothms:
```suggestion
```

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
+
 Concept 2: spatial decomposition
 --------------------------------
 
+Real buildings are made of nested places: a site holds a building, a building
+holds storeys, and a storey holds rooms. IFC mirrors this with a chain of
+**IfcSite**, **IfcBuilding**, **IfcBuildingStorey** and **IfcSpace**.
+
+Two different relationships are at work here, and telling them apart is the
+single most useful thing to understand about IFC.
```
aothms:
```suggestion
Two different relationships are at work here:
```

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
+
+**Aggregation** breaks a place into smaller places. A building is *made of*
+its storeys.
```
aothms:
```suggestion
its storeys. This relationship can also be used for physical products, such as a window being made of frame and glazing.
```

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
 Concept 3: object typing
 ------------------------
 
+Most buildings repeat themselves. A house may have thirteen walls, but only
+two kinds of wall. IFC separates the thing that exists from the kind of thing
+it is.
+
+Each individual wall is an **occurrence**. The shared definition is a **type**,
+here an **IfcWallType**. Properties that are true of every wall of that kind,
+such as its construction, live on the type and are not repeated thirteen times.
```
aothms:
```suggestion
such as its construction, live on the type and are not repeated thirteen times. Occurrences can exist without types.
```

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
+    wall_type = ifcopenshell.util.element.get_type(wall)
+    print(wall_type.Name) # Leichtbeton 102890359 240
+
+    # And back the other way, every occurrence of that type:
+    print(len(ifcopenshell.util.element.get_types(wall_type))) # 5
+
+.. tip::
+
+   If you change something on the type, you change it for every occurrence of
+   that type. That is usually what you want, and occasionally a nasty surprise.
```
aothms:
```suggestion
   that type (unless overridden on other occurrences - when the same property exists there as well). That is usually what you want, and occasionally a nasty surprise.
```

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
+
+.. note::
+
+   ``by_type`` includes subclasses. In this model every wall is actually an
+   **IfcWallStandardCase**, which is a subtype of **IfcWall**, so asking for
+   ``IfcWall`` still finds all thirteen. If you ever need to exclude
+   subclasses, pass ``include_subtypes=False``.
```
aothms:
```suggestion
   subclasses, pass ``include_subtypes=False``. Note that `by_type()` looks up by schema entity type (e.g IfcWall or IfcCartesianPoint) not by IfcTypeObject (a specific type or family of walls).
```

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
+
 Concept 4: attributes and property sets
 ---------------------------------------
 
+There are two different ways an element carries information, and beginners
+often go looking in the wrong one.
```
aothms:
```suggestion
There are two ways an element carries information:
```

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
+**GlobalId**, a **Name** and a **Description**, in a defined order, whether or
+not they are filled in. You read them directly:
+
+.. code-block:: python
+
+    print(wall.GlobalId)    # 2XPyKWY018sA1ygZKgQPtU
+    print(wall.Name)        # Wand-Int-ERDG-4
+    print(wall.is_a())      # IfcWallStandardCase
+
+**Property sets** are extensible. Anyone can attach one, and this is where most
```
aothms:
```suggestion
**Property sets** are extensible. Tools and modellers can attach their own, and this is where most
```

src/ifcopenshell-python/docs/introduction/introduction_to_ifc.rst
```diff
@@ -203,19 +203,243 @@ between **IFC Classes**. In this guide, we'll focus on the five most common
...
+of the interesting data in a real model actually lives. They are grouped by
+name, and IfcOpenShell returns them as a plain dictionary:
```
aothms:
```suggestion
of the interesting data in a real model actually lives. When using the API, IfcOpenShell returns them as a plain dictionary keyed by property name:
```

aothms (changes requested):

BIMvoice:
Thank you both. All sixteen points are applied in `36763022f3`.

@aothms on the `IfcProject` cardinality, you are right and I checked the rule rather than take it on trust. In `src/ifcopenshell-python/ifcopenshell/express/rules/IFC4X3_ADD2.py`:

```python
class IfcSingleProjectInstance:
    SCOPE = 'file'
    def __call__(file):
        assert (sizeof(file.by_type('IfcProject')) <= 1) is not False
```

`<= 1`, so at most one, and a file with none passes. The opening sentence now says "at most one" and mentions that a file may also hold any number of **IfcProjectLibrary** entities. The `.. warning::` that asserted the stronger claim is gone, which your suggestion on 242 removed anyway.

Three places where I did not apply a suggestion literally, so you can push back:

1. Your 207-208 suggestion reads "It provides units to numbers in the model are in and geometric contexts." I took the intent and wrote "It provides the units the numbers in the model are in, and the geometric contexts", since the original does not quite parse.
2. Deleting 222-224 left two `.. code-block:: python` directives adjacent with no prose between them. I merged them into one block and moved `import ifcopenshell.util.unit` up to the other import, which is terser again rather than less so. Re-ran it against the sample: output unchanged.
3. Your 326 suggestion used a dash for the parenthetical. I wrote it as "unless the same property also exists on an occurrence and overrides it" to match the surrounding prose. Same meaning.

@brunopostle every one of your four reproduces here, so all four are in:

- `get_decomposition` returns a **set** of 58, and `elements[0]` raises `TypeError: 'set' object is not subscriptable`. The composition matches yours exactly: 14 `IfcOpeningElement`, 12 `IfcAnnotation`, 9 `IfcWallStandardCase`, 9 `IfcWindow`, 6 `IfcSpace`, 5 `IfcDoor`, 1 each of slab, stair and beam. Both facts are now in the text.
- `get_psets(wall)["Pset_WallCommon"]` is `{'ThermalTransmittance': 1.5, 'id': 15079}`. The `id` key is now called out, with the advice to skip it when iterating.
- `include_subtypes=False` returns `0` on this model. It is printed rather than left as a trap.
- The `seealso` keeps its sentence and hangs the link on it, as you wrote it.

Your three minor points are in too: `get_types` now reads `# 5 of the 13` (the other type has 8), the first snippet says the file is the sample downloaded above, and the layer result moved onto the `print` line to match the rest of the page.

On the draft question, that is a fair argument and I withdraw the offer to trim. The page parses with zero docutils messages, down from two on `v0.9.0` because of the pre-existing broken reference.

One limit unchanged from the original description: I still cannot run a full Sphinx build locally, since that needs `sphinx-autoapi` to introspect the package, so the admonitions have been checked structurally but not rendered.

PR #9460, opened 2026-09-09
[AI-generated, unverified] ifcparse: stop re-sorting the inverse index on every read after a write

Moult:

## Problem

Since the datamodel switch, any loop that mutates the file and then reads an inverse attribute is dramatically slower than v0.8 on large files. `ifcopenshell.api.pset.add_pset` is the common case: it reads `IsDefinedBy`, then creates an `IfcPropertySet` and an `IfcRelDefinesByProperties`.

Measured with the same script against the same 155 MB IFC4 model (201k `IfcRoot`), Python 3.13 source-tree builds:

| scenario | v0.8.0 | v0.9.0 before | v0.9.0 after |
|---|---|---|---|
| `add_pset` per call, 4000 walls in a fresh file | 0.2 ms | 0.11–0.24 ms | 0.10 ms |
| `add_pset` per call, elements of the 155 MB model | 0.4 ms | **400 ms** | 0.25 ms |

The v0.9 cost is flat per call and proportional to file size, so it is invisible in unit tests and on small files.

## Cause

`impl::inverse_index` (`src/ifcparse/storage.h`) is one flat `std::vector<inverse_record>` for the whole file with a `sorted_` flag. Every `add()` clears the flag, and every lookup (`equal_range`) calls `sort()`, which runs a full `std::sort` over every reference in the file when the flag is dirty. So each "create an instance, read an inverse" iteration re-sorts the whole reference table: O(R log R) per iteration for R references.

Isolated on the same model: 50 `IfcRelDefinesByProperties` creates cost 1.8 ms, the first inverse read afterwards 567 ms, the next read 0.04 ms. A single `x.OwnerHistory = o` followed by a read costs 607 ms.

This is separate from #9371, which fixed aggregate attribute rewrites re-registering every member and doesn't touch the lookup path.

## Change

`inverse_index` becomes two tiers, keeping the flat parse-time layout intact:

- **base**: the existing flat vector. Bulk loading appends to it and `sort()` finalizes it once, exactly as before. Lookups binary-search it. Removals tombstone the record in place instead of erasing.
- **delta**: records added after `sort()`, in an `unordered_map` bucketed by referenced id. A lookup reads the base range and then the bucket.

`compact()` folds the delta into the base (sort the delta, `inplace_merge`) and drops tombstones. It runs from the mutators once the delta or the tombstones outgrow the live base (delta capped at 2^20 records), so folding is amortised O(1) per mutation and delta memory stays bounded. A freshly loaded file has an empty delta, so memory after parsing is unchanged.

The four in-memory callers in `parse.cpp` (`instances_by_reference`, `get_inverse`, `get_inverse_indices_by_id`, `get_total_inverses`) move from an iterator range to `for_each`/`count`. The legacy map-shaped interface (`begin`/`end`/`find`/`insert`/`erase`) that `variant_map` needs is kept and materialises from both tiers.

Result order is unchanged for base records (`record_less`); delta records follow in insertion order. `instances_by_reference` and `get_inverse_indices_by_id` share the traversal, so their pairing is preserved.

`remove_source()` (used on instance deletion) is still a linear walk, as before; nothing indexes records by source id. That is a separate problem.

## Verification

- New Catch2 case in `src/ifcparse/tests/test_ifcopenshell_parse.cpp`: 600 interleaved create-then-read iterations (folding the delta several times), attribute repointing across both tiers, one source referencing the target through two attributes, deletion of a source and of the target, and tombstoning most of the base past the compaction threshold. All 5 parse test cases pass.
- Python: `test_file.py`, `test_file_inverse.py`, `test_entity_instance.py`, `test_file_gc.py`, `test_parse.py`, `test/api/pset`, `test/api/root`, `test/api/project` pass (360 passed, 24 skipped). The whole `test/api` tree plus `test_open.py`, `test_stream.py`, `test_global_id_updates.py`, `test_guid.py`, `test_instance_string_formatting.py`, `test_express_aggregate_bounds.py`: 1818 passed, 7 skipped.
- Removal benchmark (`file.remove` on 300 `IfcPropertySet`, `root.remove_product` on 100 walls, 61 MB IFC4 model): no regression.

  | scenario | v0.8.0 | v0.9.0 before | v0.9.0 after |
  |---|---|---|---|
  | `file.remove` per call, 300 `IfcPropertySet` | 0.06 ms | 1.34 ms | 1.33 ms |
  | `root.remove_product` per call, 100 walls | 176 ms | 416 ms | 282 ms |

  The remaining gap to v0.8 on removal is the linear `remove_source()` walk, which this PR leaves as it was.

Not tested: the RocksDB storage backend (untouched code path, but the tests were not run with it), Windows/macOS builds, and Bonsai.

### DIFF
--- a/.github/workflows/ci-bonsai-daily.yml
+++ b/.github/workflows/ci-bonsai-daily.yml
@@ -124,7 +124,7 @@ jobs:
           # Reregister Bonsai.
           # Note that running it in background might miss some errors
           # (e.g. tools are not registered in background mode).
-          blender --background --python IfcOpenShell/src/bonsai/scripts/reregister_bonsai.py
+          blender --background --python-exit-code 1 --python IfcOpenShell/src/bonsai/scripts/reregister_bonsai.py
 
           # Install sverchok.
           wget -q -O sverchok.zip https://github.com/nortikin/sverchok/archive/refs/heads/master.zip
@@ -171,6 +171,6 @@ jobs:
 
           cd IfcOpenShell/src/bonsai
           pip install -r requirements-dev.txt
-          blender --background --python scripts/setup_pytest.py
-          blender --python-expr "import bonsai; print(bonsai.bbim_semver); import ifcopenshell; print(ifcopenshell.version)" --background
+          blender --background --python-exit-code 1 --python scripts/setup_pytest.py
+          blender --python-exit-code 1 --python-expr "import bonsai; print(bonsai.bbim_semver); import ifcopenshell; print(ifcopenshell.version)" --background
           make test
--- a/src/ifcparse/parse.cpp
+++ b/src/ifcparse/parse.cpp
@@ -3091,11 +3091,10 @@ std::vector<express::base> file::instances_by_reference(int t) {
     std::vector<express::base> ret;
     std::visit([this, t, &ret](auto& x) {
         if constexpr (std::is_same_v<std::decay_t<decltype(x)>, impl::in_memory_file_storage>) {
-            auto range = x.byref_excl_.equal_range((uint32_t)t);
-            ret.reserve(ret.size() + (size_t)std::distance(range.first, range.second));
-            for (auto it = range.first; it != range.second; ++it) {
-                ret.push_back(instance_by_id(it->source_id));
-            }
+            ret.reserve(ret.size() + x.byref_excl_.count((uint32_t)t));
+            x.byref_excl_.for_each((uint32_t)t, [this, &ret](const impl::inverse_record& record) {
+                ret.push_back(instance_by_id(record.source_id));
+            });
         }
 #ifdef IFOPSH_WITH_ROCKSDB
         else if constexpr (std::is_same_v<std::decay_t<decltype(x)>, impl::rocks_db_file_storage>) {
@@ -3251,11 +3250,10 @@ std::vector<int> file::get_inverse_indices_by_id(int instance_id) {
         if constexpr (std::is_same_v<std::decay_t<decltype(x)>, std::monostate>) {
         } else if constexpr (std::is_same_v<std::decay_t<decltype(x)>, impl::in_memory_file_storage>) {
             handled = true;
-            auto range = x.byref_excl_.equal_range((uint32_t)instance_id);
-            return_value.reserve((size_t)std::distance(range.first, range.second));
-            for (auto it = range.first; it != range.second; ++it) {
-                return_value.push_back(it->attribute_index);
-            }
+            return_value.reserve(x.byref_excl_.count((uint32_t)instance_id));
+            x.byref_excl_.for_each((uint32_t)instance_id, [&return_value](const impl::inverse_record& record) {
+                return_value.push_back(record.attribute_index);
+            });
         } else if constexpr (std::is_same_v<std::decay_t<decltype(x)>, impl::rocks_db_file_storage>) {
 #ifdef IFOPSH_WITH_ROCKSDB
             // @todo no lower/upper_bounds() implemented yet
@@ -3320,13 +3318,12 @@ std::vector<express::entity> file::get_inverse(int instance_id, const ifcopenshe
             visit_subtypes(type->as_entity(), [&source_types](const ifcopenshell::declaration* ent) {
                 source_types[ent->index_in_schema()] = 1;
             });
-            auto range = x.byref_excl_.equal_range((uint32_t)instance_id);
-            for (auto it = range.first; it != range.second; ++it) {
-                if (it->source_entity < source_types.size() && source_types[it->source_entity] &&
-                    (attribute_index == -1 || it->attribute_index == attribute_index)) {
-                    return_value.push_back(instance_by_id(it->source_id).template as<express::entity>());
+            x.byref_excl_.for_each((uint32_t)instance_id, [this, &source_types, attribute_index, &return_value](const impl::inverse_record& record) {
+                if (record.source_entity < source_types.size() && source_types[record.source_entity] &&
+                    (attribute_index == -1 || record.attribute_index == attribute_index)) {
+                    return_value.push_back(instance_by_id(record.source_id).template as<express::entity>());
                 }
-            }
+            });
         }
 #ifdef IFOPSH_WITH_ROCKSDB
         else if constexpr (std::is_same_v<std::decay_t<decltype(x)>, impl::rocks_db_file_storage>) {
@@ -3366,10 +3363,9 @@ size_t file::get_total_inverses(int instance_id) {
     std::visit([&counted_ids, instance_id](auto& x) {
         if constexpr (std::is_same_v<std::decay_t<decltype(x)>, std::monostate>) {
         } else if constexpr (std::is_same_v<std::decay_t<decltype(x)>, impl::in_memory_file_storage>) {
-            auto range = x.byref_excl_.equal_range((uint32_t)instance_id);
-            for (auto it = range.first; it != range.second; ++it) {
-                counted_ids.insert(it->source_id);
-            }
+            x.byref_excl_.for_each((uint32_t)instance_id, [&counted_ids](const impl::inverse_record& record) {
+                counted_ids.insert(record.source_id);
+            });
         } else if constexpr (std::is_same_v<std::decay_t<decltype(x)>, impl::rocks_db_file_storage>) {
             // @todo
         }
--- a/src/ifcparse/storage.h
+++ b/src/ifcparse/storage.h
@@ -31,6 +31,7 @@ namespace rocksdb {
 #include <algorithm>
 #include <cstdint>
 #include <iterator>
+#include <limits>
 #include <map>
 #include <memory>
 #include <cstring>
@@ -218,6 +219,29 @@ namespace ifcopenshell {
             int16_t attribute_index;
         };
 
+        // Which instances reference a given instance, and through which
+        // attribute. One index serves the whole file.
+        //
+        // Two tiers keep every operation cheap without giving up the compact
+        // flat layout that parsing relies on:
+        //
+        // - base_: one flat vector. Bulk loading appends to it unsorted and
+        //   sort() finalizes it once; lookups then binary-search it. Removing
+        //   a record tombstones it in place (attribute_index set to
+        //   dead_attribute) rather than erasing, so removal doesn't shift the
+        //   vector.
+        // - delta_: records added after sort(), bucketed by referenced_id.
+        //   A lookup reads the base range and then the bucket.
+        //
+        // compact() folds the delta into the base and drops tombstones. add()
+        // and the removal methods run it once the delta or the tombstones
+        // outgrow the live base (capped by delta_fold_limit), so folding is
+        // amortised O(1) per mutation and the delta's memory stays bounded.
+        //
+        // Before the split every lookup re-sorted the entire vector if
+        // anything had been added since the previous lookup, so a loop that
+        // creates an instance and then reads an inverse cost O(R log R) per
+        // iteration on a file with R references.
         class inverse_index {
         public:
             typedef std::map<std::tuple<short, short>, std::vector<uint32_t>> legacy_bucket;
@@ -227,11 +251,21 @@ namespace ifcopenshell {
             typedef legacy_map::value_type value_type;
             typedef legacy_map::iterator iterator;
             typedef legacy_map::const_iterator const_iterator;
-            typedef std::vector<inverse_record>::const_iterator record_iterator;
 
         private:
-            mutable std::vector<inverse_record> records_;
-            mutable bool sorted_ = true;
+            typedef std::vector<inverse_record>::const_iterator base_iterator;
+
+            // Attribute indices are small and non-negative, so the minimum
+            // value can't collide with a live record.
+            static constexpr int16_t dead_attribute = std::numeric_limits<int16_t>::min();
+            static constexpr size_t delta_fold_limit = size_t(1) << 20;
+
+            // Lookups on a const index still need to finalize bulk loading.
+            mutable std::vector<inverse_record> base_;
+            mutable bool sorted_ = false;
+            size_t dead_ = 0;
+            std::unordered_map<uint32_t, std::vector<inverse_record>> delta_;
+            size_t delta_size_ = 0;
             mutable std::unique_ptr<legacy_map> materialized_;
 
             static bool record_less(const inverse_record& a, const inverse_record& b) {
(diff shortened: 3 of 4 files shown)


aothms:
I think what we're constantly seeing is tension between the two use cases. Static reads of huge files and more interactive usage.

Before agreeing on a hybrid approach with compaction, I think we should benchmark and investigate the following alternatives:

- vector during parse (the reason why we adopted vector was that the iterative insertion into the maps and rebalancing and all was expensive - iirc) -> but then after parse convert to unordered_map
- runtime choice between vector and unordered_map based on user-envisioned scenario - either static choice during initialisation or methods to convert between the representations (which this PR more or less provides); maybe even a runtime choice between vector; unordered_map; and hybrid - maybe this approach is a generalisation of the 1st proposal (start with vector -> convert to map).

Then we need realistic numbers of parse time; conversion time; insertion time; and peak memory usage (we do care about spikes, e.g a time when both representations exist at the same time for ex. - so try to eliminate this)

> The remaining gap to v0.8 on removal is the linear remove_source() walk, which this PR leaves as it was.

linear remove_source() walk - what is that?

Moult:
_(Like the PR, this reply and the benchmark behind it were produced by an AI coding tool. Dion has not verified the numbers or the code.)_

Benchmarked the alternatives on the real reference sets of three models. Each representation is built from the same records the parser produced, in its own process, with `malloc_trim` before every RSS reading and `VmHWM` for peaks. "add+read" is one record insert followed by a full read of that id, 20k iterations over uniformly random referenced ids, which is the `add_pset` shape. Lookups are 1M probes, iterating every record as the callers do, drawn either uniformly over distinct referenced ids or uniformly over records (so hubs like `IfcOwnerHistory` dominate).

| model | parse | RSS | records | distinct ids |
|---|---|---|---|---|
| TXG 61 MB | 1.39 s | 373 MB | 1.66 M | 0.81 M |
| 210_King_Merged 155 MB | 3.66 s | 892 MB | 4.88 M | 2.61 M |
| OKgate22 232 MB | 6.24 s | 1379 MB | 5.73 M | 3.51 M |

**210_King_Merged** (the 155 MB model from the PR description):

| representation | build / convert | footprint | peak over inputs | lookup by id | lookup by record | add+read |
|---|---|---|---|---|---|---|
| sorted vector (current) | 0.36 s | 73 MB | 0 | 213 ns | 0.9 µs | **413 ms** |
| vector + CSR offsets | 0.38 s | 95 MB | 0 | 23 ns | 0.7 µs | static |
| unordered_map, built incrementally in parse order | 1.04 s | 183 MB | +43 MB | 70 ns | 0.7 µs | 0.32 µs |
| unordered_map, bulk from sorted vector | 0.46 s | 174 MB | +35 MB | 72 ns | 0.7 µs | 0.32 µs |
| unordered_map, chunked from sorted vector | 0.59 s | 174 MB | **0** | 72 ns | 0.7 µs | 0.31 µs |
| hybrid (this PR) | 0.38 s | 73 MB | 0 | 190 ns | 1.7 µs | 0.52 µs |

Same three columns that matter, other two models:

| model | representation | build / convert | footprint | add+read |
|---|---|---|---|---|
| TXG | sorted vector | 0.12 s | 29 MB | 115 ms |
| TXG | unordered_map, chunked | 0.13 s | 66 MB | 0.30 µs |
| TXG | hybrid | 0.12 s | 29 MB | 0.39 µs |
| OKgate22 | sorted vector | 0.43 s | 86 MB | 372 ms |
| OKgate22 | unordered_map, chunked | 1.16 s | 264 MB | 0.36 µs |
| OKgate22 | hybrid | 0.45 s | 86 MB | 0.59 µs |

What this says about the two proposals:

**Vector during parse, convert to `unordered_map<uint32_t, std::vector<inverse_record>>` after.** Conversion costs 0.13 / 0.59 / 1.16 s on top of parse (9%, 16%, 19%), and the map's footprint is 2.3 to 3.1x the vector's: +37 / +101 / +178 MB, or 10 to 13% of total RSS. The spike is avoidable: consuming the sorted vector from the back and shrinking it in 16 steps measured zero peak over the inputs, at +0.13 s versus the naive bulk build. Incremental insertion during parse (1.04 s vs 0.36 s for the sort on King) is the cost you remember, and bulk-from-sorted avoids most of it since every bucket is allocated once at its exact size. After conversion there is one representation: no tombstones, no compaction, and removal into a target is a bucket erase.

**Hybrid.** Zero conversion cost and zero extra memory after load, add+read within 2x of the map. The price is code: two tiers, tombstones, and a compaction policy. The 2x on hub iteration is the per-record tombstone check defeating vectorization of the benchmark's sum; a `dead_ == 0` fast path removes it, and real callers spend microseconds per record materializing instances anyway.

**CSR offsets** (a `uint32_t` per instance id, +22 MB here) turn the sorted vector's binary search into a 23 ns lookup and could be added to either design.

**Runtime choice.** All of these fit behind the same six operations the PR already isolates (`add`, `remove`, `remove_source`, `erase`, `for_each`, `count`), so a `std::variant` of representations chosen at open time is mechanical. The hybrid is a superset in the sense that "convert to map" is a delta that is never folded back. I don't have a use case where the static-vector reader needs to opt out of the hybrid, since the hybrid's read path is the vector's, but I may be missing one.

If you'd rather pay the 10 to 20% parse and memory for the simpler single-representation code, I can rewrite the PR as chunked convert-after-parse. Your call on which tension you want to resolve which way.

> linear `remove_source()` walk - what is that?

On deletion, `process_deletion_inverse` does two things: `erase(id)` drops the records *into* the deleted instance (keyed by referenced id, cheap), and `remove_source(id)` drops the records *from* it, i.e. the references its own attributes made. Nothing indexes by source id, so that second step scans all R records. It is O(R) per deletion in both the pre-PR vector and the PR (and would be in the unordered_map too). On King it's 1.3 ms per `file.remove` against 0.06 ms in v0.8.0. The fix is independent of representation: at deletion time the instance's own entity-valued attributes are known, so walk them and call the keyed `remove(referenced, source, entity, attribute)` for each. Happy to do that as a follow-up PR.

<details>
<summary>Benchmark source (standalone, links libifcopenshell.parse)</summary>

```cpp
// Benchmarks inverse-index representations on a real model's reference set.
// Usage: bench_inverse <ifc> <mode>   mode: parse | sorted_vector | csr | umap_incremental | umap_bulk | umap_chunked | hybrid
#include "ifcparse/file.h"
#include "ifcparse/storage.h"
#include <chrono>
#include <cstdio>
#include <fstream>
#include <malloc.h>
#include <random>
#include <string>
#include <unordered_map>
#include <vector>

using ifcopenshell::impl::inverse_record;
using clk = std::chrono::steady_clock;

static double secs(clk::time_point a) { return std::chrono::duration<double>(clk::now() - a).count(); }

static long proc_kb(const char* key) {
    std::ifstream in("/proc/self/status");
    std::string line;
    while (std::getline(in, line)) {
        if (line.rfind(key, 0) == 0) return std::stol(line.substr(line.find_first_of("0123456789")));
    }
    return -1;
}
static void trim() { malloc_trim(0); }
static long rss_mb() { trim(); return proc_kb("VmRSS:") / 1024; }
static long hwm_mb() { return proc_kb("VmHWM:") / 1024; }

static bool record_less(const inverse_record& a, const inverse_record& b) {
    if (a.referenced_id != b.referenced_id) return a.referenced_id < b.referenced_id;
    if (a.source_entity != b.source_entity) return a.source_entity < b.source_entity;
    if (a.attribute_index != b.attribute_index) return a.attribute_index < b.attribute_index;
    return a.source_id < b.source_id;
}
struct ref_less {
    bool operator()(const inverse_record& a, uint32_t id) const { return a.referenced_id < id; }
    bool operator()(uint32_t id, const inverse_record& a) const { return id < a.referenced_id; }
};

typedef std::unordered_map<uint32_t, std::vector<inverse_record>> umap;

int main(int argc, char** argv) {
    if (argc < 3) return 1;
    const std::string path = argv[1], mode = argv[2];
    const long rss_start = rss_mb();

    auto t = clk::now();
    ifcopenshell::file f(path);
    const double t_parse = secs(t);
    auto& index = std::get<ifcopenshell::impl::in_memory_file_storage>(f.storage_).byref_excl_;
    const long rss_parsed = rss_mb();
    printf("parse            %8.2f s   rss %6ld MB (+%ld)   records %zu   max_id %u\n", t_parse, rss_parsed, rss_parsed - rss_start, index.size(), f.get_max_id());
    if (mode == "parse") return 0;

    // Parse-order proxy: the records as the parser emits them, unsorted.
    // Extracted by referenced id then shuffled, since the index is already sorted.
    std::vector<inverse_record> recs;
    recs.reserve(index.size());
    for (uint32_t id = 1; id <= f.get_max_id(); ++id) {
        index.for_each(id, [&recs](const inverse_record& r) { recs.push_back(r); });
    }
    std::vector<inverse_record> sorted = recs;  // sorted by construction
    std::shuffle(recs.begin(), recs.end(), std::mt19937(1));
    const size_t R = recs.size();
    const uint32_t max_id = f.get_max_id();
    const long rss_base = rss_mb();
    const long hwm_base = hwm_mb();
    printf("extracted        records %zu   rss %ld MB (%.0f MB per copy)\n", R, rss_base, R * sizeof(inverse_record) / 1048576.0);

    std::mt19937 rng(2);
    // Two distributions: per record (hubs like IfcOwnerHistory dominate) and per distinct referenced id.
    std::vector<uint32_t> probes_hub(1000000), probes_id(1000000);
    for (auto& p : probes_hub) p = recs[rng() % R].referenced_id;
    std::vector<uint32_t> distinct;
    for (size_t i = 0; i < R; ++i) if (i == 0 || sorted[i].referenced_id != sorted[i - 1].referenced_id) distinct.push_back(sorted[i].referenced_id);
    for (auto& p : probes_id) p = distinct[rng() % distinct.size()];
    printf("distinct referenced ids %zu   mean fan-in %.1f\n", distinct.size(), (double)R / distinct.size());

    // ---- build ----
    t = clk::now();
    std::vector<inverse_record> vec;
    std::vector<uint32_t> offsets;
    umap map;
    ifcopenshell::impl::inverse_index hybrid;
    if (mode == "sorted_vector" || mode == "csr") {
        vec = recs;
        std::sort(vec.begin(), vec.end(), record_less);
        if (mode == "csr") {
            offsets.assign((size_t)max_id + 2, 0);
            for (const auto& r : vec) ++offsets[r.referenced_id + 1];
            for (size_t i = 1; i < offsets.size(); ++i) offsets[i] += offsets[i - 1];
        }
    } else if (mode == "umap_incremental") {
        for (const auto& r : recs) map[r.referenced_id].push_back(r);
    } else if (mode == "umap_bulk") {
        // From the sorted vector: one exact-size bucket per run.
        for (size_t i = 0; i < R;) {
            size_t j = i;
            while (j < R && sorted[j].referenced_id == sorted[i].referenced_id) ++j;
            map.emplace(sorted[i].referenced_id, std::vector<inverse_record>(sorted.begin() + i, sorted.begin() + j));
            i = j;
        }
    } else if (mode == "umap_chunked") {
        // Consume the sorted vector from the back, releasing it in 16 steps so
        // both representations never fully coexist.
        std::vector<inverse_record> src = std::move(sorted);
        const size_t step = R / 16 + 1;
        while (!src.empty()) {
            size_t cut = src.size() > step ? src.size() - step : 0;
            while (cut > 0 && src[cut - 1].referenced_id == src[cut].referenced_id) --cut;
            for (size_t i = cut; i < src.size();) {
                size_t j = i;
                while (j < src.size() && src[j].referenced_id == src[i].referenced_id) ++j;
                map.emplace(src[i].referenced_id, std::vector<inverse_record>(src.begin() + i, src.begin() + j));
                i = j;
            }
            src.resize(cut);
            src.shrink_to_fit();
        }
    } else if (mode == "hybrid") {
        hybrid.reserve(R);
        for (const auto& r : recs) hybrid.add(r.referenced_id, r.source_id, r.source_entity, r.attribute_index);
        hybrid.sort();
    }
    const double t_build = secs(t);
    const long hwm_build = hwm_mb();
    // Free the inputs so the representation's own footprint is visible.
    std::vector<inverse_record>().swap(recs);
    std::vector<inverse_record>().swap(sorted);
    const long rss_rep = rss_mb();
    printf("build %-10s %8.2f s   footprint %6ld MB   peak during build +%ld MB over inputs\n", mode.c_str(), t_build, rss_rep - rss_base + (long)(2 * R * sizeof(inverse_record) / 1048576), hwm_build - hwm_base);

    // ---- lookups: iterate every record, as the real callers do ----
    size_t hits = 0;
    const auto visit = [&](uint32_t id) {
        if (mode == "sorted_vector") {
            auto r = std::equal_range(vec.begin(), vec.end(), id, ref_less{});
            for (auto it = r.first; it != r.second; ++it) hits += it->source_id;
        } else if (mode == "csr") {
            for (uint32_t i = offsets[id]; i < offsets[id + 1]; ++i) hits += vec[i].source_id;
        } else if (mode == "hybrid") {
            hybrid.for_each(id, [&hits](const inverse_record& r) { hits += r.source_id; });
        } else {
            auto it = map.find(id);
            if (it != map.end()) for (const auto& r : it->second) hits += r.source_id;
        }
    };
    t = clk::now();
    for (auto id : probes_id) visit(id);
    printf("lookup by id     %8.0f ns/lookup   (1M probes, uniform over distinct ids)\n", secs(t) / probes_id.size() * 1e9);
    t = clk::now();
    for (auto id : probes_hub) visit(id);
    printf("lookup by record %8.0f ns/lookup   (1M probes, uniform over records)\n", secs(t) / probes_hub.size() * 1e9);

    // ---- interleaved add + read (the add_pset pattern) ----
    const int N = (mode == "sorted_vector") ? 10 : 20000;
    uint32_t next_id = max_id + 1;
    t = clk::now();
    for (int i = 0; i < N; ++i) {
        const uint32_t target = probes_id[i];
        const inverse_record r{target, next_id++, 1, 4};
        if (mode == "sorted_vector") {
            vec.push_back(r);
            std::sort(vec.begin(), vec.end(), record_less);
            auto q = std::equal_range(vec.begin(), vec.end(), target, ref_less{});
            for (auto it = q.first; it != q.second; ++it) hits += it->source_id;
        } else if (mode == "csr") {
            break;  // static structure: no insertion path on its own
        } else if (mode == "hybrid") {
            hybrid.add(r.referenced_id, r.source_id, r.source_entity, r.attribute_index);
            hybrid.for_each(target, [&hits](const inverse_record& r) { hits += r.source_id; });
        } else {
            map[target].push_back(r);
            for (const auto& q : map.find(target)->second) hits += q.source_id;
        }
    }
    if (mode != "csr") printf("add+read         %8.2f us/iteration over %d iterations\n", secs(t) / N * 1e6, N);
    printf("final            rss %ld MB   peak %ld MB   (%zu)\n", rss_mb(), hwm_mb(), hits);
    return 0;
}
```
</details>

aothms:
Ok, it seems hybrid (this pr) is the proper way forward. I indeed wasn't a fan of the increased code complexity. But if I have to choose between the tradeoffs, memory is the major deciding factor to me. @Moult I'm ok on both PRs.

Issue #6432, opened 2025-03-24
D&D Phase 1 Item 10 - Open multiple drawings in viewport


jes-r:
Hi,

The tenth (and for now last) item in the Drawings & Documentation proposal: https://community.osarch.org/discussion/2640/drawings-documentation-donation-proposal#latest

So when working on a model it would be very helpful to be able to see e.g. both section drawing view and plan drawing view at the same time. Or 3D and plan or whatever. However, when switching to a plan view, all 3D viewports switch to that plan view, see video below. 

(dont minde the missing text, im testing some code xD)

You see all views change to the new items, though I'm a bit suprised here the 3D views stay in place. Though ideally I'd want 1 view plan and the other a section view and be able to choose what changes when!

Would be very nice to be able to control this!

theoryshaw:
Here's the OSArch funding campaign for this issue, if interested in helping to finance the development of this feature. For every dollar that is raised, OSArch will match that dollar from their general funding pool (if available).
https://opencollective.com/osarch/projects/bonsai-open-multiple-drawings

franklincg:
I'm interested in working on this funded OSArch item. The Open Collective project still shows a balance for #6432, and I don't see an assignee or an open PR for it.

Before I start implementation, could you confirm that the allocation is still active/claimable given the 365-day note on the funding page? If so, my plan is to first isolate the drawing/view state that is currently shared across Blender 3D viewports, then implement the smallest viewport-local approach that lets different viewports show different drawing/3D views without changing the others, with regression coverage where practical.

I use AI-assisted development tools as part of my workflow and will disclose that in any resulting PR.

theoryshaw:
Very cool.  Yes it's still available.  Have at it. :)

franklincg:
Thanks — confirmed. I’m starting on this now and will keep the
implementation scoped to independent per-viewport drawing/view state, with
focused regression coverage where practical. I’ll post the PR once the
behavior is validated.

On Wed, 09 Sep 2026 05:10:23 -0700, Ryan Schultz ***@***.***
wrote:

theoryshaw left a comment (IfcOpenShell/IfcOpenShell#6432)
<https://github.com/IfcOpenShell/IfcOpenShell/issues/6432#issuecomment-5601597276>

Very cool. Yes it’s still available. Have at it. :)

—
Reply to this email directly, view it on GitHub
[...]

PR #9464, opened 2026-09-09
Bonsai: preserve drawing cameras per viewport

franklincg:
## Summary

This is a focused first step toward #6432: keep camera-mode 3D viewports on their current drawing camera when another viewport activates a different drawing.

Before changing `scene.camera`, Bonsai now pins other camera-perspective viewports that are still following the scene camera to the previous camera using `SpaceView3D.use_local_camera`. The viewport that invoked drawing activation is preferred as the active target, while existing user-local cameras and non-camera perspective views are left unchanged.

## Scope

This addresses the camera-switching part of #6432. Bonsai still has drawing activation paths that change object visibility and IFC representations at scene/view-layer level, so this PR does **not** claim to make all drawing state viewport-local yet. I kept this change narrow rather than mixing a broader visibility/representation redesign into the camera fix.

## Testing

- `git diff --check`
- Both changed Python files compiled successfully inside Blender 4.5 (`COMPILE_OK`).
- Added focused regression coverage for preserving the previous scene camera in other camera-mode viewports and leaving local/perspective views untouched.
- I attempted the focused Blender pytest run, but this local Blender Python does not have `pytest` installed (`ModuleNotFoundError: No module named 'pytest'`), so the repository CI remains the authoritative test run.

## AI assistance

OpenAI Codex was used to help investigate, implement, and review this change. I reviewed the resulting code and tests and take responsibility for the contribution.

Refs #6432

### DIFF
--- a/src/bonsai/bonsai/tool/blender.py
+++ b/src/bonsai/bonsai/tool/blender.py
@@ -146,13 +146,16 @@ class Blender(bonsai.core.tool.Blender):
 
     @classmethod
     def activate_camera(cls, obj: bpy.types.Object) -> None:
-
         area = cls.get_view3d_area()
         assert area
-        assert isinstance((space := area.spaces[0]), bpy.types.SpaceView3D)
+        assert isinstance((space := area.spaces.active), bpy.types.SpaceView3D)
         is_local_view = space.local_view is not None
 
         assert bpy.context.screen and bpy.context.scene
+        previous_camera = bpy.context.scene.camera
+        if previous_camera and previous_camera != obj:
+            # Other camera views would otherwise follow the global scene camera change.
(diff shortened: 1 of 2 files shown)


franklincg:
Thanks for confirming #6432 is still funded. The first focused camera-localization slice is now in this PR and is mergeable. GitHub Actions is currently waiting on the standard maintainer approval for an external contributor. Could a maintainer approve the workflow and review this slice? I kept the PR deliberately narrow: it prevents other camera-mode 3D viewports from following the scene camera when a different drawing is activated, while preserving existing local cameras and perspective views. For the funded item, please also let me know whether you consider this camera-independence slice an acceptable first funded milestone, or whether you want the visibility / representation state made viewport-local before any payout. I can continue with the next bounded slice based on that answer.

PR #9465, opened 2026-09-09
Document the quoting rule in the selector syntax

brunopostle:
An unquoted pset, prop, key or value may not contain any of , . = > < * ! or whitespace. This was never stated, and the only motivation the docs gave for quoting was the "Level 3" example having a space, so there was nothing to suggest that Pset_WallCommon.ThermalTransmittance=1.5 is a syntax error while FireRating=2HR is fine. Decimal numbers are the common way to hit it.

State the rule where the reader meets values, in a new subsection under each of the two value tables. The two grammars exclude different characters, so the element value section states its own set and only the delta. Note that all of these characters are a syntax error except the comma, which is read as the filter separator instead, so Name=Foo,IfcWall silently means "named Foo and an IfcWall" rather than matching the literal value.

Also fix the Query filter example, which had the same problem in the file itself: query:types.count=0 does not error, it is silently parsed as a property filter for a count property in a property set named query:types, and matches nothing. Quote the keys, as the other query examples already do.

### DIFF
--- a/src/ifcopenshell-python/docs/ifcopenshell-python/selector_syntax.rst
+++ b/src/ifcopenshell-python/docs/ifcopenshell-python/selector_syntax.rst
@@ -56,11 +56,13 @@ Filtering is typically used to select any IFC element or type.
 
     "``IfcWall, Pset_WallCommon.FireRating=2HR``", "Any 2 hour fire rated wall"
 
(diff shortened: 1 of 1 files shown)


Moult:
No longer needed right because the other PR went through?

brunopostle:
I think there were two pull requests, the other one included this commit and fixed the decimal quirk and updated the doc. The other one was merged so this one can be closed.

PR #9474, opened 2026-09-11
[AI-generated, unverified] ifcparse: inline the file reader accessors and pre-size the inverse index

Moult:
> **This PR was written entirely by an AI coding tool (Claude Code) and has not been verified by a human.** The maintainer who opened it has not reviewed the code, the design, or the benchmark methodology.

The commit from #9473 that needs no design discussion, split out so it can merge on its own ("makes sense, no impact on complexity").

The SWIG `-fastproxy -fastdispatch` change that was here too has been withdrawn from both PRs: it breaks `ifcopenshell.util.scripts.validate_stub` (and its CI test), which parses the generated `ifcopenshell_wrapper.py` with `ast` and compares each method's `def` signature against the `.pyi` stub. Fast-proxy replaces those defs with `name = _swig_new_instance_method(...)` assignments, so there is nothing to compare. Keeping it would need the stub validator to learn a second method form and lose the signature check, which is a separate discussion.

## Inline the file reader accessors, pre-size the inverse index

`full_buffer_impl::size()`, `get()`, `get_u32()` and `get_u64()` (and their `mmap_impl` twins) were defined out of line, so every character the lexer read crossed a call boundary carrying its own bounds check. Callgrind put the three at 5.6% of parse self time on a 58 MB model; inlining them lets the compiler hoist the checks out of the scanning loops, which is worth more than their own cost. The streamer's inverse vector is also reserved from the file size (about one record per 32 bytes of SPF on real models) and shrunk once the bulk load is sorted, so the doubling copies and the capacity slack go away.

Parse time, C++ `file` constructor, 12-core Linux box, Python 3.13 source-tree build (v0.9.0 at 148ff0285):

| model | before | after |
|---|---|---|
| TXG 58 MB | 1.35 s | 1.15 s |
| 210_King_Merged 148 MB | 3.70 s | 3.09 s |
| OKgate22 232 MB | 6.18 s | 5.30 s |

End to end from Python (`ifcopenshell.open`): 1.40 → 1.22 s, 3.70 → 3.24 s, 6.27 → 5.52 s. Memory unchanged. The removed exported symbols mean the Python wrapper must be rebuilt against the library (a stale wrapper fails to import with `undefined symbol: ...full_buffer_impl4sizeEv`).

## Verification

5 Catch2 parse cases; Python `test_file.py`, `test_file_inverse.py`, `test_entity_instance.py`, `test_file_gc.py`, `test_parse.py`, `test/api/pset`, `test/api/root`, `test/api/project`: 360 passed, 24 skipped. Not tested: Windows/macOS builds, Bonsai.

## Across more files

Same measurement (C++ `file` constructor, single thread, 12-core Linux box) on eleven further models of 10 MB to 788 MB supplied privately for benchmarking (named by size; nothing from their contents is reported here):

| model | v0.9.0 head (148ff0285) | with this commit | change |
|---|---|---|---|
| TXG 58 MB | 1.39 s | 1.19 s | −14% |
| 210_King 147 MB | 3.77 s | 3.22 s | −15% |
| OKgate22 231 MB | 6.21 s | 5.41 s | −13% |
| 10 MB | 0.23 s | 0.19 s | −17% |
| 74 MB | 1.62 s | 1.36 s | −16% |
| 81 MB | 1.77 s | 1.49 s | −16% |
| 107 MB | 2.42 s | 2.04 s | −16% |
| 249 MB | 5.62 s | 4.82 s | −14% |
| 279 MB | 4.81 s | 3.86 s | −20% |
| 315 MB | 7.55 s | 6.44 s | −15% |
| 349 MB | 7.79 s | 6.67 s | −14% |
| 523 MB | 8.91 s | 7.10 s | −20% |
| 603 MB | 10.97 s | 8.86 s | −19% |
| 787 MB | 17.20 s | 14.78 s | −14% |

Memory is identical before and after on every file (the change touches no data structure that survives the parse). Both builds are the same tree at the two commits, built with the same flags, schema plugins included; each file was parsed once per build in alternation.

### DIFF
--- a/src/ifcparse/file_reader.cpp
+++ b/src/ifcparse/file_reader.cpp
@@ -75,33 +75,6 @@ full_buffer_impl::full_buffer_impl(const std::string& content, const caller_fed_
     , size_(content.size()) {
 }
 
-size_t full_buffer_impl::size() const { return size_; }
-
-char full_buffer_impl::get(size_t pos) const {
-    if (pos >= buf_.size()) {
-        throw std::out_of_range("get out of range");
-    }
-    return buf_[pos];
-}
-
-uint64_t full_buffer_impl::get_u64(size_t pos) const {
-    if (pos + sizeof(uint64_t) > buf_.size()) {
-        throw std::out_of_range("get_u64 out of range");
-    }
-    uint64_t value;
-    std::memcpy(&value, buf_.data() + pos, sizeof(value));
-    return value;
-}
-
-uint32_t full_buffer_impl::get_u32(size_t pos) const {
-    if (pos + sizeof(uint32_t) > buf_.size()) {
-        throw std::out_of_range("get_u32 out of range");
-    }
-    uint32_t value;
-    std::memcpy(&value, buf_.data() + pos, sizeof(value));
-    return value;
-}
-
 void full_buffer_impl::push_next_page(const std::string& data) {
     buf_.insert(buf_.end(), data.begin(), data.end());
     size_ = buf_.size();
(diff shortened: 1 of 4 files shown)


Moult:
Closing: per the discussion on #9473, this commit becomes the first of a fresh, minimal series that will replace #9473 as a whole, so there is no need to review it separately. (Comment written by an AI coding tool.)

Issue #1860, opened 2021-11-08
blenderbim brickschema dependencies


brunopostle:
Latest blenderbim requires `brickschema`, this package is very specific about which versions of its dependencies it requires. 

For example fedora 34 and later ship with `xlrd-2.0.1`, but all versions of `brickschema` require `xlrd < 2.0.0`. Similarly, I can't satisfy required versions of `rdflib`, `flask`, `jinja2`, `typer`, `click-spinner`, `requests`, and others.

Installing `brickschema` with pip fails for the same reasons.

I've hacked-up rpm packages of `owlrl`, `pyshacl` and `brickschema` that install without full dependency checking, so at least I can build IfcOpenShell and blenderbim packages again, but the `brickschema` functionality is probably broken.

Can `brickschema` be made an optional dependency? i.e. import it within a try/except

Moult:
Good point - also a good reminder that my current method of bundling via Makefile could be significantly improved with perhaps a requirements.txt file with explicit versions, and perhaps optional vs mandatory vs dev deps (but I'm not too familiar with Python packaging so I'll have to read up on it).

PR #9479, opened 2026-09-12
Reallow Bonsai without brickschema

brunopostle:
6b62723 added a module-level `from brickschema.persistent import Changeset` to `src/bonsai/bonsai/tool/brick.py`, which made `brickschema` a hard dependency again - Bonsai fails to load without it.

This moves the import into the `TYPE_CHECKING` block, where the other `brickschema` imports already live, so it is only used for annotations.

See also #1860.


Moult:

Closed and reopened only to trigger the renamed test workflows, which are now required checks and had never run on this head. Merging once they are green; no change to the content is needed.

PR #9489, opened 2026-09-12
dev_environment.py: copy the runtime libraries next to the wrapper

### LINKED ISSUE #9488: dev_environment.py drops 0.9.0's runtime libraries when linking, breaking the dev env with "No schema named IFC4"
## Bug Description

`dev_environment.py` copies the compiled dependencies into the repo with a single glob:

```python
# src/bonsai/scripts/dev_environment.py
for path in PACKAGE_PATH.glob("ifcopenshell/*_wrapper*"):
```

and then replaces `site-packages/ifcopenshell` with a symlink to `src/ifcopenshell-python/ifcopenshell`.

That glob was written when the wrapper was self-contained. Since 0.9.0 it isn't: the package ships runtime libraries alongside it which are loaded **by name at runtime** rather than linked. From `ifcopenshell-0.9.0a260909-py313-none-win_amd64.whl`:

```
ifcopenshell/_ifcopenshell_wrapper.cp313-win_amd64.pyd      <- matches *_wrapper*
ifcopenshell/ifcopenshell.geometry.dll                      <- does not match
ifcopenshell/ifcopenshell.geometry.writer.dll               <- does not match
ifcopenshell/ifcopenshell.parse.dll                         <- does not match
ifcopenshell/ifcopenshell.plugin.dll                        <- does not match
ifcopenshell/ifcopenshell_document_rdb.dll                  <- does not match
ifcopenshell/ifcopenshell_geometry_kernel_opencascade.dll   <- does not match
```

Only the wrapper is copied into the repo. The symlink then shadows the originals still sitting in site-packages, so they become unreachable and the wrapper comes up without the libraries it needs. It surfaces as:

```
RuntimeError: No schema named IFC4
```

Comparing what the glob selects against what the package actually contains:

| package | `*_wrapper*` selects | present and needed |
|---|---|---|
| `ifcopenshell-0.8.6a260708-py313` | 2 | 2 |
| `ifcopenshell-0.9.0a260909-py313` | 2 | 8 |

### To reproduce

1. Install a Bonsai 0.9.0 build (Blender 5.1+).
2. Run `python src/bonsai/scripts/dev_environment.py`, answering with that Blender version.
3. Launch Blender. Bonsai fails to register with `RuntimeError: No schema named IFC4`.

### Relationship to #9423

Same error text, different cause, and they are independent.

#9423 is the Windows **packaging** step failing to collect `ifcopenshell_parse_schema_ifc*.dll` into the published zip. This issue is the **dev-link script** failing to copy runtime libraries that are present in the package.

They compound rather than overlap: once #9423 is fixed and the schema libraries ship, this glob will still drop them on every `dev_environment.py` run, so a linked dev environment would stay broken even with a correct package.

Not Windows-specific in principle — the same omission applies to `.so` / `.dylib` on Linux and macOS.

### Suggested fix

Select on file type instead of on the `_wrapper` name, so anything compiled in the package folder is copied:

```python
library_suffixes = (".dll", ".pyd", ".so", ".dylib")
for path in PACKAGE_PATH.glob("ifcopenshell/*"):
    if not path.is_file():
        continue
    # Type stubs are checked into the repo, don't overwrite them.
    if path.suffix.lower() == ".pyi":
        continue
    if path.suffix.lower() not in library_suffixes and "_wrapper" not in path.name:
        continue
    ...
```

Checked against both packages above: identical selection for 0.8.6 (no behaviour change), and 8 files instead of 2 for 0.9.0. Future `ifcopenshell_parse_schema_ifc*.dll` files are then collected by the same `.dll` rule.

I have this working locally and am happy to open a PR.

## Debug and Error Output

```
OS:              Windows 10 19045
Blender:         5.1 (bundled Python 3.13)
Bonsai:          0.9.0-alpha260909
Script:          src/bonsai/scripts/dev_environment.py @ v0.9.0 (23f3874de1)

Files in site-packages/ifcopenshell matching the current glob "ifcopenshell/*_wrapper*":
  _ifcopenshell_wrapper.cp313-win_amd64.pyd
  ifcopenshell_wrapper.py

Compiled files in the same folder that are NOT matched, and so are lost behind the symlink:
  ifcopenshell.geometry.dll
  ifcopenshell.geometry.writer.dll
  ifcopenshell.parse.dll
  ifcopenshell.plugin.dll
  ifcopenshell_document_rdb.dll
  ifcopenshell_geometry_kernel_opencascade.dll
```

theoryshaw:
Closes #9488.

The compiled dependencies were collected with a `ifcopenshell/*_wrapper*` glob, written when the wrapper was a single self-contained binary.

Since 0.9.0 it isn't. The package ships runtime libraries alongside it — `ifcopenshell.parse.dll`, `ifcopenshell.geometry.dll`, the per-schema `ifcopenshell_parse_schema_ifc*.dll` and friends — which are loaded by name at runtime rather than linked, so none of them match that glob. Only the wrapper was copied into the repo, and the symlink that replaces the package folder then shadowed the originals left behind in site-packages. The wrapper came up without the libraries it needs and failed at import with `RuntimeError: No schema named IFC4`, a long way from the cause.

This selects on file type instead of on the name. Checked against both generations:

| package | before | after |
|---|---|---|
| `ifcopenshell-0.8.6a260708-py313` | 2 files | 2 files (unchanged) |
| `ifcopenshell-0.9.0a260909-py313` | 2 files | 8 files |

No behaviour change on 0.8.6. On 0.9.0 it picks up the six runtime libraries that were being dropped. The schema libraries are collected by the same `.dll` rule once they ship, and `.so` / `.dylib` keep Linux and macOS from hitting the same gap.

Worth noting this is independent of #9423 despite the identical error text: that one is the Windows packaging step failing to collect `ifcopenshell_parse_schema_ifc*.dll` into the published zip, this one is the dev-link script failing to copy runtime libraries that *are* present. They compound — once #9423 ships fixed packages, this glob would still drop the schema libraries on every `dev_environment.py` run.

---

**Stacked on #9240**, which touches the same region of this file — please merge that one first. Until it lands this PR's diff shows both commits; it narrows to the single commit here once #9240 is merged.

### DIFF
--- a/src/bonsai/scripts/dev_environment.py
+++ b/src/bonsai/scripts/dev_environment.py
@@ -19,6 +19,7 @@ import argparse
 import shutil
 import subprocess
 import sys
+import time
 import urllib.request
 from pathlib import Path
 from typing import Union
@@ -95,6 +96,44 @@ BLENDER_VERSION_INT = tuple(map(int, BLENDER_VERSION.split(".")))
 PYTHON_VERSION = "3.13" if BLENDER_VERSION_INT >= (5, 1) else "3.11"
 PACKAGE_PATH = BLENDER_PATH / rf"extensions/.local/lib/python{PYTHON_VERSION}/site-packages"
 
+PARTIAL_LINK_ERROR = """
+Failed while replacing '{path}'.
+
+The extension is now only partially linked and Blender won't be able to import it until
+this completes. This is usually a file lock - close Blender if it's running. An antivirus
+scan of a freshly installed extension can also hold files for a while.
+
+Re-running this script is safe and will finish the job.
+"""
+
+
+def remove_path(path: Path) -> None:
+    """Remove `path` so it can be replaced by a symlink."""
+    if path.is_dir() and not path.is_symlink():
+        shutil.rmtree(path)
+    # Check `is_symlink` in case if it's a broken symlink.
+    elif path.is_file() or path.is_symlink():
+        path.unlink()
+
+
+def remove_path_with_retries(path: Path, attempts: int = 5, delay: float = 2.0) -> None:
+    """`remove_path`, retrying while the files are still locked by another process.
+
+    `rmtree` is only reached on the first setup, when the extension binaries have just been
+    written and an antivirus scanner may still be holding them. On Windows that surfaces as
+    PermissionError (WinError 5, not the sharing violation you might expect) and usually
+    clears within a few seconds.
+    """
+    for attempt in range(1, attempts + 1):
+        try:
+            remove_path(path)
+            return
+        except PermissionError:
+            if attempt == attempts:
+                raise
+            print(f"'{path}' is locked, retrying in {delay:.0f}s ({attempt}/{attempts - 1})...")
+            time.sleep(delay)
+
 
 def main() -> None:
     global REPO_PATH
@@ -137,6 +176,8 @@ def main() -> None:
         path.unlink()
     subprocess.check_call(("git", "checkout", "--", symlinks_glob), cwd=REPO_PATH)
 
+    # Note this has to stay ahead of the symlinking below: it's the step that gets the compiled
+    # wrapper out of site-packages before that folder is deleted and replaced by a symlink.
     if args.skip_binaries:
         print("Skipping copying compiled dependencies to the repo...")
     else:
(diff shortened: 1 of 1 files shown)


theoryshaw:
Superseded by b71cac0771 ("Bonsai dev_environment.py: copy all IfcOpenShell binaries, not just the wrapper"), which fixes this properly — closing.

That commit takes a better approach than this PR did. Rather than matching binaries by name or suffix, it copies everything in the installed package that isn't tracked in the repo:

```python
output = subprocess.check_output(("git", "ls-files"), cwd=repo_package_path, text=True)
tracked = {line.split("/")[0] for line in output.splitlines()}
```

No list to keep up to date as the build layout changes, which is exactly the failure mode this PR would have had. It also covers two things I missed: writing *through* a repo symlink (clobbering a local build the symlink pointed at), and refusing the "I compile it myself" mode when the repo has no binaries.

Confirming it against a current install, since the scale of the problem grew a lot in the meantime — `0.9.0-alpha260920` on Windows/Blender 5.1 now ships **87** compiled libraries next to the wrapper (8 `ifcopenshell_parse_schema_ifc*.dll`, 27 OpenCASCADE `TK*.dll`, plus per-schema geometry mappers and writers), 105 MB in total. The old `*_wrapper*` glob selected 2 of them.

Thanks for the more durable fix.

PR #9492, opened 2026-09-13
[AI-generated, unverified] ifcparse: faster, smaller open(): one tokenizer, references in the slots, lazy loading, parallel and paged parsing

Moult:
> **This PR was written entirely by an AI coding tool (Claude Code) and has not been verified by a human.** The maintainer who opened it has not reviewed the code, the design, or the benchmark methodology.

This is the fresh series asked for in #9473: one tokenizer, minimal commits in the order (1) inlining and minimal optimisations, GlobalId map, heap allocation reduction; (2) template context on the existing tokenizer; (3) unresolved reference storage and lazy loading; (4) multithreading; plus the SWIG fast-proxy with `validate_stub` taught to read signatures from the compiled functions, two commits from profiling the lazy open, and the tokenizer restructured as `scan(Consumer&)`, of which every pass is now a consumer, the parser included (below). It replaces #9473, which stays open only as the reference for the numbers posted there. 24 commits; net +4420 −2041 lines, of which +3788 −1930 outside tests (#9473 added 2169). Commits 15 and 16 take on the serial phases left after the parallel parse (the inverse-record sort, the per-worker merge); commits 17 to 20 are @aothms's: an MSVC export fix and three tokenizer commits that put the index scan at the byte scanner's throughput; commit 21 takes the paged index scan a step further on the same idea; commits 22 and 23 make the parser itself a consumer of `scan()` and remove `next()` and the token object; commit 24 restores the streamer's stop point after the `;`. The tables below are the whole branch at c64353c07 against v0.9.0 head (commit 24 changes only where the streamer stops). `dense_id_map` is not included. `guid_map.h` is gone as reviewed: the GlobalId index is the spelled-out `std::unordered_map<std::array<char, 22>, ...>` on the storage class.

## Before and after

C++ `file` constructor on fourteen models: three public (TXG, 210_King, OKgate22) and eleven supplied privately for benchmarking, identified by size only. Wall clock on a 12-core Linux box, one run each; memory is anonymous RSS after the parse, peak is VmHWM; percentages are against v0.9.0 head (148ff0285) in the same row. Every run logged zero parser messages. Single runs on this box vary by up to about 10% between sessions, so differences smaller than that between tables are not significant.

| model | instances | v0.9.0 head, 1 thread | this PR, 1 thread | this PR, 12 threads | this PR, paged, 12 threads | this PR, lazy, 1 thread | this PR, lazy, 12 threads |
|---|---|---|---|---|---|---|---|
| TXG 58 MB | 917,995 | 1.39 s, 357 MB (peak 463) | 0.86 s (-38%), 274 MB (-23%), peak 366 (-21%) | 0.33 s (-76%), peak 396 MB | 0.34 s, peak 387 MB (-16%) | 0.31 s, 146 MB (-59%) | 0.24 s (-83%), 212 MB |
| 210_King 147 MB | 2,731,958 | 3.77 s, 877 MB (peak 1138) | 2.11 s (-44%), 654 MB (-25%), peak 871 (-23%) | 0.77 s (-80%), peak 933 MB | 0.79 s, peak 838 MB (-26%) | 0.86 s, 392 MB (-55%) | 0.64 s (-83%), 476 MB |
| OKgate22 231 MB | 3,825,882 | 6.21 s, 1364 MB (peak 1743) | 4.26 s (-31%), 1036 MB (-24%), peak 1348 (-23%) | 1.25 s (-80%), peak 1438 MB | 1.25 s, peak 1260 MB (-28%) | 1.23 s, 541 MB (-60%) | 0.97 s (-84%), 660 MB |
| 10 MB (private) | 209,548 | 0.23 s, 61 MB (peak 90) | 0.13 s (-43%), 47 MB (-23%), peak 76 (-16%) | 0.05 s (-78%), peak 78 MB | 0.06 s, peak 83 MB (-8%) | 0.05 s, 33 MB (-46%) | 0.05 s (-78%), 46 MB |
| 74 MB (private) | 1,162,030 | 1.62 s, 444 MB (peak 573) | 0.97 s (-40%), 340 MB (-23%), peak 449 (-22%) | 0.36 s (-78%), peak 485 MB | 0.36 s, peak 458 MB (-20%) | 0.36 s, 164 MB (-63%) | 0.27 s (-83%), 240 MB |
| 81 MB (private) | 1,527,802 | 1.77 s, 443 MB (peak 584) | 1.05 s (-41%), 333 MB (-25%), peak 451 (-23%) | 0.38 s (-79%), peak 471 MB | 0.38 s, peak 442 MB (-24%) | 0.38 s, 197 MB (-56%) | 0.30 s (-83%), 268 MB |
| 107 MB (private) | 2,152,533 | 2.42 s, 581 MB (peak 792) | 1.32 s (-45%), 433 MB (-25%), peak 591 (-25%) | 0.48 s (-80%), peak 622 MB | 0.50 s, peak 567 MB (-28%) | 0.56 s, 281 MB (-52%) | 0.42 s (-83%), 359 MB |
| 249 MB (private) | 3,884,378 | 5.62 s, 1335 MB (peak 1752) | 3.26 s (-42%), 1022 MB (-23%), peak 1378 (-21%) | 1.26 s (-78%), peak 1478 MB | 1.29 s, peak 1280 MB (-27%) | 1.21 s, 550 MB (-59%) | 0.98 s (-83%), 680 MB |
| 279 MB (private) | 625,577 | 4.81 s, 720 MB (peak 1044) | 3.01 s (-37%), 669 MB (-7%), peak 974 (-7%) | 0.91 s (-81%), peak 1015 MB | 0.98 s, peak 790 MB (-24%) | 0.60 s, 102 MB (-86%) | 0.61 s (-87%), 184 MB |
| 315 MB (private) | 5,618,442 | 7.55 s, 1791 MB (peak 2333) | 4.47 s (-41%), 1346 MB (-25%), peak 1778 (-24%) | 1.48 s (-80%), peak 1923 MB | 1.52 s, peak 1657 MB (-29%) | 1.51 s, 737 MB (-59%) | 1.17 s (-85%), 909 MB |
| 349 MB (private) | 6,430,306 | 7.79 s, 1913 MB (peak 2551) | 4.47 s (-43%), 1442 MB (-25%), peak 1913 (-25%) | 1.57 s (-80%), peak 1989 MB | 1.60 s, peak 1689 MB (-34%) | 1.72 s, 833 MB (-56%) | 1.32 s (-83%), 955 MB |
| 523 MB (private) | 1,056,071 | 8.91 s, 1320 MB (peak 1913) | 5.66 s (-36%), 1232 MB (-7%), peak 1790 (-6%) | 1.72 s (-81%), peak 1851 MB | 1.84 s, peak 1380 MB (-28%) | 1.11 s, 170 MB (-87%) | 1.14 s (-87%), 244 MB |
| 603 MB (private) | 6,579,063 | 10.97 s, 2289 MB (peak 3081) | 6.81 s (-38%), 1889 MB (-17%), peak 2598 (-16%) | 2.29 s (-79%), peak 2714 MB | 2.37 s, peak 2160 MB (-30%) | 2.26 s, 851 MB (-63%) | 1.78 s (-84%), 1015 MB |
| 787 MB (private) | 16,232,582 | 17.20 s, 4304 MB (peak 5811) | 9.87 s (-43%), 3174 MB (-26%), peak 4254 (-27%) | 3.66 s (-79%), peak 4454 MB | 3.68 s, peak 3719 MB (-36%) | 3.99 s, 2047 MB (-52%) | 3.12 s (-82%), 2310 MB |

Python `ifcopenshell.open` on the same fourteen models, wall clock and RSS growth over the bare interpreter as a multiple of file size. This PR's default open is the parallel strict parse; v0.8.5 is the last release; v0.9.0 head is a wrapper built from 148ff0285 with the same options.

| model | v0.8.5 | v0.9.0 head | this PR, default | this PR, lazy |
|---|---|---|---|---|
| TXG 58 MB | 2.81 s, 10.6x file size | 1.45 s, 6.2x file size | 0.33 s, 5.0x file size (-77%) | 0.25 s, 3.7x file size (-83%) |
| 210_King 147 MB | 7.14 s, 11.1x file size | 3.87 s, 6.0x file size | 0.82 s, 4.7x file size (-79%) | 0.67 s, 3.3x file size (-83%) |
| OKgate22 231 MB | 11.44 s, 10.4x file size | 6.46 s, 5.9x file size | 1.25 s, 4.7x file size (-81%) | 0.97 s, 2.8x file size (-85%) |
| 10 MB (private) | 0.50 s, 12.6x file size | 0.25 s, 6.5x file size | 0.06 s, 5.1x file size (-76%) | 0.06 s, 4.9x file size (-76%) |
| 74 MB (private) | 3.56 s, 10.6x file size | 1.68 s, 6.1x file size | 0.36 s, 5.0x file size (-79%) | 0.28 s, 3.3x file size (-83%) |
| 81 MB (private) | 3.57 s, 10.4x file size | 1.82 s, 5.5x file size | 0.36 s, 4.3x file size (-80%) | 0.31 s, 3.3x file size (-83%) |
| 107 MB (private) | 5.16 s, 11.4x file size | 2.49 s, 5.5x file size | 0.50 s, 4.2x file size (-80%) | 0.44 s, 3.3x file size (-82%) |
| 249 MB (private) | 11.93 s, 9.8x file size | 5.78 s, 5.4x file size | 1.28 s, 4.4x file size (-78%) | 1.02 s, 2.7x file size (-82%) |
| 279 MB (private) | 9.65 s, 3.2x file size | 5.13 s, 2.6x file size | 0.93 s, 2.6x file size (-82%) | 0.63 s, 0.7x file size (-88%) |
| 315 MB (private) | 14.74 s, 10.8x file size | 7.82 s, 5.8x file size | 1.49 s, 4.6x file size (-81%) | 1.20 s, 2.9x file size (-85%) |
| 349 MB (private) | 16.52 s, 11.1x file size | 8.37 s, 5.5x file size | 1.56 s, 4.2x file size (-81%) | 1.31 s, 2.7x file size (-84%) |
| 523 MB (private) | 18.01 s, 3.1x file size | 9.18 s, 2.5x file size | 1.73 s, 2.4x file size (-81%) | 1.14 s, 0.5x file size (-88%) |
| 603 MB (private) | 22.52 s, 6.5x file size | 11.37 s, 3.9x file size | 2.31 s, 3.1x file size (-80%) | 1.80 s, 1.7x file size (-84%) |
| 787 MB (private) | 36.95 s, 11.5x file size | 18.12 s, 5.5x file size | 3.58 s, 4.1x file size (-80%) | 3.12 s, 3.0x file size (-83%) |

## The commits

Lines are per commit, added and removed, with the tests counted separately so the cost of each item is visible.

| # | commit | code +/− | tests +/− | what, and which item of the requested order |
|---|---|---|---|---|
| 1 | e0b6122d1 | +51 −62 | +0 −0 | (1) minimal optimisations: the reader accessors inline, the inverse vector pre-sized (was #9474) |
| 2 | 64805b409 | +94 −14 | +33 −12 | (1) the GlobalId index as `std::unordered_map<std::array<char, 22>, ...>` on the storage, with `variant_map` converting from `std::string` at the file interface; only a 22-character GlobalId can be stored or found |
| 3 | d0ac88881 | +35 −28 | +0 −0 | (1) heap allocation reduction: one block per attribute array, `std::optional` storage |
| 4 | d647050bc | +64 −0 | +0 −0 | (1) the paged reader caches the page its cursor is on, revalidated by an eviction counter |
| 5 | a026fc698 | +49 −18 | +0 −0 | (1) the SWAR helpers, the reader's cursor accessors, paged `size()` and the page-cache check forced inline; found with callgrind |
| 6 | 9829ebf00 | +102 −9 | +48 −0 | (2) `spf_lexer::next<Policy>()`: `full_tokens` is the parser, `index_tokens` ends strings without decoding them and passes over numbers, enumerations and binaries; `character_decoder::skip()` is the same state machine with collection compiled out; also fixes a comment that follows a token without whitespace |
| 7 | 91622b97b | +148 −16 | +92 −0 | (3) references stay in the slot as the existing `instance_reference` / `reference_or_simple_type` aggregate until every instance is read, mixed lists included; no side table for the parse |
| 8 | a92bebd73 | +352 −7 | +154 −0 | (3) lazy loading through `next<index_tokens>()`; offsets in one lazy-only sorted vector, nothing added to `instance_data` for the full parse |
| 9 | b1cba757b | +499 −91 | +91 −3 | (4) parallel parsing; the instance-header loop is shared with the lazy index; the split-point scan tracks strings and comments |
| 10 | 43686bde8 | +22 −2 | +19 −1 | (4) opt-in full parse through the paged reader |
| 11 | 1d950365a | +110 −3 | +0 −0 | SWIG `-fastproxy -fastdispatch`; `validate_stub` rebuilds signatures from the compiled functions' docstrings |
| 12 | 97c83f029 | +27 −6 | +0 −0 | (profile) keyword, enumeration and binary text inside an attribute list not copied by the index |
| 13 | f24e9a6bc | +242 −161 | +8 −1 | (profile) the lazy index chunked over the same split points as the parallel parse, merged in file order; serial is the one-chunk case |
| 14 | 17ec5d450 | +519 −350 | +37 −0 | (consumer model) the tokenizer body as `scan(Consumer&)` in `spf_scan.h`, callbacks inlining into the loop; `next()` kept as the one-token consumer for the pull parser; the lazy index is a consumer |
| 15 | aaeeecb69 | +46 −1 | +0 −0 | (serial phases) `inverse_index::sort_records()`: LSD radix sort on `referenced_id`, `record_less` within each run of equal ids, same order as before |
| 16 | 788d25767 | +112 −28 | +0 −0 | (serial phases) each parse and index worker sorts its own records in place, builds its own by-type lists and reads its own GlobalIds; the main thread merges sorted runs (`inverse_index::merge_sorted`) |
| 17 | 3b808a893 | +1 −1 | +0 −0 | (aothms) MSVC: the streamer's explicit full-buffer instantiation exported instead of the class template |
| 18 | cb66a2dda | +22 −15 | +0 −0 | (aothms) a string-pool entry only when the consumer keeps text or the token is a name; discarded literals allocate nothing |
| 19 | 188319996 | +21 −17 | +0 −0 | (aothms) the SWAR probes compiled out for consumers that discard text, a scalar delimiter check instead: index tokens are mostly shorter than a word |
| 20 | 5f4ecdc30 | +41 −6 | +75 −5 | (aothms) `#name` parsed with `from_chars` straight from the reader's span, the token loop kept for whitespace, page boundaries and invalid spellings; tests with one-byte pages and the pushed reader |
| 21 | 8623e9054 | +27 −0 | +0 −0 | a discarded literal walked over the reader's span with a local pointer instead of a bounds and page check per byte; the per-byte loop only at a page boundary |
| 22 | dacadaa0e | +760 −387 | +0 −0 | the parser as a consumer of `scan()`: `list_consumer` with a frame stack in place of the recursive attribute reader; `instance_header_consumer` (the `#name = KEYWORD (` window as consumer states) for the parallel chunk, the lazy index and the streamer; the header section a consumer; `next()` no longer called |
| 23 | c64353c07 | +405 −705 | +75 −89 | `next()`, the one-token consumer, the `token` object and the two policies removed; the scan body moved from `spf_scan.h` into `parse.h` next to the lexer; the tests reference a decoding consumer |
| 24 | f0e08d8b9 | +39 −3 | +0 −0 | the streamer's `read_instance()` takes the `;` after the instance again (the streaming loop asks `has_semicolon()` before each read); CI's `test_chunked_stream` caught the regression of commit 22 |

Each commit message carries its own numbers on the three public models.

## One tokenizer

There is no byte-level scanner left. Every pass is a consumer of `scan()`: the parser (`list_consumer`, a frame stack where the pull parser recursed), the lazy index (`attribute_consumer`, counting parentheses and commas on the operator callbacks), and above both the instance-header consumer, which finds `#name = KEYWORD (`, looks declarations up once per keyword, passes over a bypassed instance's attribute list and slides past a stray keyword, and hands the list to the parser's or the index's consumer through a scan of its own; the serial parse, the parallel chunks, the lazy index and the streamer all go through it. There is no token object. Whatever the consumer decodes, a string ends at the byte the decoder's state machine says it ends at; the equivalence test includes `''`, `\X2\` and `\S\'`, an apostrophe used as the page character, which any quote-counting byte scan (including the one in #9473) ended early.

The one place that looks at raw bytes is the split-point scan for parallel parsing, because tokenizing the file serially to find the split points would leave nothing to parallelise. It applies three rules: a string starts and ends at a quote and cannot span a line, and a comment runs from `/*` to `*/`; a split is a `#` that starts a line outside both. Getting a string's end wrong can only lose a candidate, never accept a wrong one, since no string contains a newline. The equality test puts a comment holding a fake instance and a string holding `/*` between the chunks. Comments in the DATA section therefore no longer force the serial path.

## Throughput of the stages

Single thread, whole file, each column a consumer of `scan()` and nothing built: the full tokenizer decodes everything, as the parser sees it; the index consumer keeps keywords but ends strings without decoding them and passes over values; the `scan()` consumer keeps nothing, as the lazy index; the paged columns read through 64 KB pages with a 4 MB cache.

| model | full tokenizer, everything decoded, in memory | same, paged | index consumer, keywords kept, in memory | scan() consumer, in memory | scan() consumer, paged | full parse, 1 thread | paged parse, 1 thread | lazy open + one attribute of every instance |
|---|---|---|---|---|---|---|---|---|
| TXG 58 MB | 432 MB/s | 360 MB/s | 535 MB/s | 1002 MB/s | 750 MB/s | 67 MB/s | 0.84 s (-2%) | 0.31 + 0.40 s |
| 210_King 147 MB | 465 MB/s | 378 MB/s | 519 MB/s | 927 MB/s | 680 MB/s | 70 MB/s | 2.18 s (+3%) | 0.86 + 0.97 s |
| OKgate22 231 MB | 449 MB/s | 366 MB/s | 541 MB/s | 985 MB/s | 725 MB/s | 54 MB/s | 4.31 s (+1%) | 1.23 + 1.76 s |
| 10 MB (private) | 452 MB/s | 353 MB/s | 503 MB/s | 1026 MB/s | 716 MB/s | 77 MB/s | 0.14 s (+8%) | 0.05 + 0.07 s |
| 74 MB (private) | 435 MB/s | 356 MB/s | 539 MB/s | 1044 MB/s | 779 MB/s | 76 MB/s | 0.97 s (+0%) | 0.36 + 0.51 s |
| 81 MB (private) | 381 MB/s | 346 MB/s | 482 MB/s | 1083 MB/s | 817 MB/s | 77 MB/s | 1.06 s (+1%) | 0.38 + 0.52 s |
| 107 MB (private) | 421 MB/s | 343 MB/s | 475 MB/s | 1069 MB/s | 841 MB/s | 81 MB/s | 1.36 s (+3%) | 0.56 + 0.68 s |
| 249 MB (private) | 474 MB/s | 396 MB/s | 558 MB/s | 1026 MB/s | 802 MB/s | 76 MB/s | 3.24 s (-1%) | 1.21 + 1.58 s |
| 279 MB (private) | 223 MB/s | 190 MB/s | 262 MB/s | 984 MB/s | 763 MB/s | 93 MB/s | 3.04 s (+1%) | 0.60 + 2.72 s |
| 315 MB (private) | 463 MB/s | 384 MB/s | 533 MB/s | 1073 MB/s | 818 MB/s | 70 MB/s | 4.44 s (-1%) | 1.51 + 2.22 s |
| 349 MB (private) | 431 MB/s | 357 MB/s | 498 MB/s | 1070 MB/s | 807 MB/s | 78 MB/s | 4.46 s (-0%) | 1.72 + 2.25 s |
| 523 MB (private) | 220 MB/s | 186 MB/s | 260 MB/s | 973 MB/s | 781 MB/s | 92 MB/s | 5.78 s (+2%) | 1.11 + 5.05 s |
| 603 MB (private) | 319 MB/s | 267 MB/s | 368 MB/s | 1059 MB/s | 827 MB/s | 89 MB/s | 6.84 s (+0%) | 2.26 + 4.01 s |
| 787 MB (private) | 457 MB/s | 378 MB/s | 502 MB/s | 1093 MB/s | 848 MB/s | 80 MB/s | 9.97 s (+1%) | 3.99 + 4.94 s |

Reading across: the index consumer is 10–25% faster than the full tokenizer and the `scan()` consumer two to four times faster again, at 925–1095 MB/s in memory and 680–850 MB/s through pages; through pages the full tokenizer runs at 78–91% of its in-memory speed, the per-byte page checks showing once the token object is gone; the full parse through pages costs −2% to +8% on one thread and 0–20% at 12 threads, with peak memory down by the size of the file. A lazy open on one thread is 20–42% of the strict single-thread parse and holds 130–170 bytes per instance; on 12 threads it is under the default (parallel full) open on every model but the smallest, where both take 0.05 s. Reading every attribute afterwards costs 40–90% of the strict parse again, so lazy is for touching a fraction of the model.

## What lazy loading is

One pass over the DATA section through the tokenizer's index policy builds everything that indexes the file: a shell per instance (name and declaration, no attribute array), the complete inverse index with attribute indices, the GlobalId map and the by-type lists. No attribute value is decoded. The first time anything touches an instance's attributes, `ensure_loaded()` seeks the retained paged reader to the instance and runs the same `load_attributes()` the full parse runs, with inverse registration off, then resolves that instance's references from its own slots. A modified instance is materialised first, so writing works. The offset of each instance's attribute list lives in one sorted vector that exists only in lazy mode. Materialising from several threads at once is not safe.

## Where the lazy open's time went, and what was done about it

Callgrind on the lazy open of TXG, on this branch against the byte scanner of #9473: the index pass is 0.28 s through `next<index_tokens>()` against 0.055 s for the byte scan, and the rest of a lazy open (shells, inverse sort, GlobalIds) is the same 0.33 s in both. The samples in the tokenizer: 37% in the index-policy `next()` itself, 6.7% in `file_reader::peek()` as a non-inlined call per byte from the whitespace skip, about 5% copying keyword and enumeration text the index never reads, 2% parsing names. The byte scanner did none of that: seven characters through one SWAR mask. Roughly 17 ns per token times 13 million tokens is the gap. Three hypotheses, each measured on the three public models (TXG / 210_King / OKgate22):

| change | index tokenizer, TXG | lazy open, 1 thread | lazy open, 12 threads |
|---|---|---|---|
| before | 236 MB/s | 0.61 / 1.73 / 2.86 s | (serial) |
| (a) reader cursor accessors forced inline (folded into commit 5; the full parse gains ~2% too) | 261 MB/s | 0.58 / 1.73 / 2.86 s | |
| (b) attribute-list policy: keyword, enumeration and binary text not copied (commit 12) | 293 MB/s | 0.57 / 1.65 / 2.78 s | |
| (c) the index chunked over threads, name table reserved before the merge (commit 13) | | 0.55 / 1.59 / 2.11 s | 0.37 / 1.01 / 1.34 s |
| for comparison: the default (parallel full) open | | | 0.44 / 1.17 / 1.91 s |

(a) and (b) recover about a fifth of the tokenizer gap: the per-token call is the cost, not what the token carries (measured per token type on synthetic files: 9–16 ns per token whatever the token is, strings the cheapest). (c) is what puts lazy back under the default open. Skipping non-structural tokens inside `next()` was also tried and reverted: on real files the structural tokens and names are the majority, so it changed nothing measurable.

Commit 14 then inverts the tokenizer as asked: the body of `next()` becomes `scan(Consumer&)` in `spf_scan.h`, handing each token to the consumer's callbacks (which inline into the loop) instead of building a token; `next<Policy>()` is kept as the consumer that stops after one token, so the recursive pull parser is untouched; the lazy index is `attribute_consumer`. Index pass, nothing decoded, in memory: 214–247 MB/s through `next()` → 266–343 MB/s through `scan()` (the byte scanner it replaces: 700–1300 MB/s); lazy open on one thread TXG / King / OKgate22 0.55 / 1.59 / 2.11 → 0.54 / 1.57 / 2.07 s. The full tokenizer through the adapter was unchanged, and end to end commit 14 was an architecture change, not a performance one: the lazy open moved within noise, the strict parse and memory not at all, because the strict parse still pulled tokens through the adapter. Commits 18 to 20 (@aothms) then take the index pass from 270–340 to 780–970 MB/s in memory (the table above), the full tokenizer gaining 30–54%; a lazy open on one thread drops by about a fifth, the strict parse within noise, because its time is in the parser. Commit 21 removes the paged reader's per-byte checks from the discarded-literal loop (the index consumer's scalar path): index scan a further 11–21% in memory and 17–28% through pages on four models, the opens within noise. Commits 22 and 23 (@aothms's last request) then make the parser a consumer too and remove `next()`, the one-token adapter and the token object: the strict parse −4–8%, the lazy open on one thread −8–10% (its header loop was the last pull-based piece), peak memory unchanged, four models best of three.

## SWIG fast-proxy

`-fastproxy` binds each wrapped method directly to its compiled function instead of a generated Python def; `-fastdispatch` shortens overload dispatch. Reading `is_a()`, `id()` and an attribute of every one of TXG's 918k instances: 2.36 s → 2.15 s (−8%). `validate_stub` compared the stub against the wrapper's def signatures; a fast-proxy wrapper has assignments, so it now rebuilds each signature from the compiled function's autodoc docstring with the rule SWIG itself uses (one prototype with literal defaults → named parameters; overloads or a non-literal default such as an enum → `*args`), checked against every def of a non-fast-proxy wrapper: 773 of 773 identical. The autodoc feature moves above the SWIG library includes so the iterator and container classes carry prototypes too.

## Verification

- 17 Catch2 parse cases (`BUILD_IFCOPENSHELL_PARSE_TESTS=ON`): a consumer that decodes nothing against one that decodes everything, token for token, the escapes above included; references resolved in place for scalars, lists, nested lists, mixed selects, missing and bypassed names; a lazily opened file against the full parse, per-instance `to_string()`, inverse counts and GlobalId lookups, on the fixture and on an inline file, plus the stray-keyword and fallback cases; `scan()` against `next()` token for token across one-byte, two-byte and larger pages and the pushed reader, and the invalid-name spellings still throwing; serial, 5-thread, paged and paged 5-thread equality on a 12 MB replicated fixture with comments and `/*`-strings between the chunks.
- Python: the whole `test/` directory of ifcopenshell-python as CI runs it, `test/test_lazy.py` included: 2537 passed, 37 skipped, 1 xfailed, with the default parallel parse and the fast-proxy wrapper; black and ruff clean. Four existing tests looked walls up by GlobalIds such as "id1"; they now use real 22-character ids, and `test_file.py` additionally checks that a short key is rejected.
- Not tested: RocksDB backend (two `ensure_loaded()` calls added in the serializer, paths otherwise untouched), Windows/macOS, Bonsai (fast-proxy changes what the wrapper's methods are at the Python level: assignments instead of defs, so anything that introspects them the way `validate_stub` did would need the same change). The Python wrapper must be rebuilt against the library.


src/ifcparse/guid_map.h
```diff
@@ -0,0 +1,154 @@
...
+    };
+    static bool is_guid_sized(const std::string& s) {
+        return s.size() == std::tuple_size<guid_key>::value;
+    }
+    static guid_key to_key(const std::string& s) {
+        guid_key key;
+        std::memcpy(key.data(), s.data(), key.size());
+        return key;
+    }
+    typedef std::unordered_map<guid_key, V, guid_key_hash> map_type;
```
aothms: I would prefer if we just make this the type in the ifcopenshell::file/storage class and remove this file. I find that more illustrative of what we do and its limitations (`std::array<char, 22>`) than introducing a seemingly opaque type that also might hint at a numeric map instead of a textual map.

src/ifcparse/parse.cpp
```diff
@@ -179,6 +179,14 @@ bool parse_num_(const char* pStart, size_t size, T& val) {
 
 } // namespace
 
+// These helpers sit on the tokenizer's innermost loop; left to the
```
aothms: I find this a bit ugly. Shall we just make them macros?

src/ifcparse/parse.cpp
```diff
@@ -2451,6 +2458,71 @@ std::optional<std::tuple<size_t, const ifcopenshell::declaration*, shared_pointe
 
 template class IFC_PARSE_API ifcopenshell::instance_streamer<file_reader<full_buffer_impl>>;
 
+void ifcopenshell::impl::in_memory_file_storage::resolve_instance_references(const shared_pointer_type& data, const std::vector<unsigned>& bypassed) {
```
aothms: This is logic we already had. Can we generalize it with the existing call site with some templated abstraction? The datatype is now the same.

src/ifcparse/parse.cpp
```diff
@@ -2523,6 +2552,255 @@ void ifcopenshell::impl::in_memory_file_storage::resolve_instance_references(con
...
+            size_t guid_begin = 0, guid_end = 0;
+            while (depth > 0) {
+                token t = lexer.next<index_tokens>();
+                if (!t) {
+                    failure = "file ends inside an instance";
+                    failure_offset = attributes_offset;
+                    break;
+                }
+                if (t.is_operator()) {
+                    if (t.value_char == '(') {
```
aothms: We already have this loop in `in_memory_file_storage::load()` and that one also accounts for: 

const bool retain_value = attribute_index_within_data < expected_size;

Which is important for the symmetry in inverses when records with the wrong length are encountered in the file.

This is exactly the reason why I want to centralize this logic because it has been battle tested with 15 years of broken data.

It also does register_inverse() already.

Please see if we can generalize the loop in while loop in_memory_file_storage::load() using a templated member for (a) full storage reads (b) lazy scans.

I feel like we're almost there in terms of centralizing the logic where possible.

aothms:
Thanks for your patience I think with the last comments addressed we're nearly at a place where important logic is reused rather than reintroduced :)

aothms:
It does seem that lazy got quite a bit slower, slower than non-lazy?

model | v0.8.5 | v0.9.0 head | this PR, default | this PR, lazy
-- | -- | -- | -- | --
349 MB (private) | 16.52 s, 11.1x file size | 8.37 s, 5.5x file size | 2.40 s, 4.2x file size (-71%) | 3.51 s, 2.4x file size (-58%)
523 MB (private) | 18.01 s, 3.1x file size | 9.18 s, 2.5x file size | 2.13 s, 2.4x file size (-77%) | 3.91 s, 0.3x file size (-57%)
603 MB (private) | 22.52 s, 6.5x file size | 11.37 s, 3.9x file size | 3.30 s, 3.2x file size (-71%) | 5.14 s, 1.4x file size (-55%)
787 MB (private) | 36.95 s, 11.5x file size | 18.12 s, 5.5x file size | 5.70 s, 4.0x file size (-69%) | 8.14 s, 2.7x file size (-55%)

once we have the PR in a state of good human comprehension let's hypothesise what we can do optimising with a profiler and really see where the allocation of samples is different wrt to the fully inlined scan blob from the other PR.

Moult:
_(AI-written, not verified by Dion.)_

**guid_map.h** is gone. The GlobalId index is now the spelled-out `std::unordered_map<std::array<char, 22>, express::base, guid_key_hash>` on the storage class, with a two-line `guid_key()` next to it that says which strings qualify; `variant_map` converts from `std::string` at the file's interface (and a string of the wrong length is simply never found or stored). Folded into commit 2, which is now +94 −14 outside tests.

**Lazy slower than non-lazy.** Two things were mixed in that table: the default column is the 12-thread parallel parse and the lazy open was single-threaded; and lazy did get slower than #9473 by exactly the difference between the byte scanner and the tokenizer. I profiled it as you suggested before changing anything. Callgrind on the lazy open of TXG, this branch against the byte scanner: the index pass is 0.28 s through `next<index_tokens>()` against 0.055 s for the byte scan, and the rest of a lazy open (shells, inverse sort, GlobalIds) is the same 0.33 s in both. Inside the tokenizer the samples were 37% in the index-policy `next()` itself, 6.7% in a non-inlined `file_reader::peek()` called per byte from the whitespace skip, about 5% copying keyword and enumeration text the index never reads, 2% parsing names; the byte scanner did none of that, it looked at seven characters through one SWAR mask. About 17 ns per token times 13 million tokens is the gap. Three hypotheses, each measured (TXG / 210_King / OKgate22, lazy open on one thread):

| change | index tokenizer, TXG | lazy open, 1 thread | lazy open, 12 threads |
|---|---|---|---|
| before | 236 MB/s | 0.61 / 1.73 / 2.86 s | |
| (a) the reader's cursor accessors forced inline, folded into commit 5; the full parse gains ~2% too | 261 MB/s | 0.58 / 1.73 / 2.86 s | |
| (b) a third policy for the inside of an attribute list: keyword, enumeration and binary text not copied (commit 12) | 293 MB/s | 0.57 / 1.65 / 2.78 s | |
| (c) the index chunked over the same split points as the parallel parse, name table reserved before the merge (commit 13) | | 0.55 / 1.59 / 2.11 s | 0.37 / 1.01 / 1.34 s |
| the default (parallel full) open, for comparison | | | 0.44 / 1.17 / 1.91 s |

(a) and (b) recover about a fifth of the tokenizer gap: the per-token call is the cost, not what the token carries, and while `next()` returns a token per call it will not reach the byte scanner's 1 GB/s. (c) is what puts lazy back under the default. The remaining single-thread lever, not tried, is the consumer-driven `next()` you mentioned earlier: hand tokens to a callable so the index's loop can inline into the tokenizer.

All fourteen models, this branch, C++ (memory after the open):

| model | default open (parallel full) | lazy, 1 thread | lazy, 12 threads |
|---|---|---|---|
| TXG 58 MB | 0.43 s | 0.53 s, 147 MB | 0.36 s, 203 MB |
| 210_King 147 MB | 1.16 s | 1.56 s, 388 MB | 0.99 s, 453 MB |
| OKgate22 231 MB | 1.89 s | 2.10 s, 542 MB | 1.34 s, 622 MB |
| 10 MB (private) | 0.08 s | 0.10 s, 33 MB | 0.07 s, 45 MB |
| 74 MB (private) | 0.49 s | 0.61 s, 163 MB | 0.38 s, 223 MB |
| 81 MB (private) | 0.51 s | 0.69 s, 194 MB | 0.44 s, 250 MB |
| 107 MB (private) | 0.72 s | 1.02 s, 275 MB | 0.63 s, 339 MB |
| 249 MB (private) | 1.95 s | 2.35 s, 551 MB | 1.58 s, 645 MB |
| 279 MB (private) | 1.09 s | 1.72 s, 102 MB | 0.82 s, 164 MB |
| 315 MB (private) | 2.20 s | 2.86 s, 730 MB | 1.82 s, 842 MB |
| 349 MB (private) | 2.41 s | 3.19 s, 816 MB | 1.99 s, 920 MB |
| 523 MB (private) | 2.08 s | 3.15 s, 170 MB | 1.53 s, 233 MB |
| 603 MB (private) | 3.31 s | 4.53 s, 851 MB | 2.63 s, 973 MB |
| 787 MB (private) | 5.75 s | 7.43 s, 2028 MB | 4.89 s, 2191 MB |

From Python, lazy is now 4–11% faster than the default open on every model at 55–70% of its memory. Full tables, the per-commit code and test line counts, and the profile numbers are in the PR body; 13 commits, net +1982 −176 lines (+1542 −164 outside tests).

aothms:
> (a) and (b) recover about a fifth of the tokenizer gap: the per-token call is the cost, not what the token carries, and while next() returns a token per call it will not reach the byte scanner's 1 GB/s.

What I think would be interesting to benchmark is on what token types the difference is most pronounced. We can also skip converting the reals to float for example on the lazy path (if not already). But it could well be that a major driver for the performance difference is the string handling, for cases such as (as you noted):

> a quote-counting byte scan ends a string early at \S\', which the shared decoder does not

So maybe benchmark the difference on files with:
 - mixed content
 - string performance (e.g a large amount of IfcPerson instances with interesting names - also large amounts of empty strings)
 - real performance (e.g a large amount of IfcCartesianPointList3D)
 - handling of aggregates (IfcPolyline?)
 - nested types performance (IfcTrimmedCurve?)

Because profiling otherwise averages everything out. This will indicate more clearly where the difference exists.

As I said I have a suspicion its the strings and the character decoder that kicks in to determine the string literal - in case of complex encoding, which is often not there. Maybe we can SWAR or otherwise handle this more efficiently. Is see you already added back character_decoder::skip() (which we previously had) which is a step in the right direction.

---

I don't know if the consumer model matches the structure well, but otherwise maybe a lexer::next_n(vector&) call can also be investigated - or maybe even just exhaust tokenize one page at a time.

Or maybe omit tokens in next() that we don't care about for lazy. Because you can keep the structural ones such as `(` `)` `,` and just skip over all the non-identifier tokens within `next()`. `,` will increment attribute count regardless of whether we saw an attribute value.

Moult:
_(AI-written, not verified by Dion.)_

Measured per token type as you suggested, on 50 MB synthetic files (IfcPerson with doubled quotes, `\S\`, `\X2\` and empty strings; IfcCartesianPoint and IfcCartesianPointList3D; IfcPolyline of 20 references; IfcTrimmedCurve with inline IfcParameterValue) plus TXG.

The columns: **byte scan** is the standalone scanner from #9473, for reference. The other three are the one `spf_lexer::next<Policy>()` of this branch under its three compile-time policies: **full** (`full_tokens`) decodes everything and is what the parser uses; **index** (`index_tokens`) ends a string without decoding it and returns numbers, enumerations and binaries as a literal with only a position, keeping keyword text, and is what the lazy index uses for the instance headers `#name = KEYWORD(`; **attribute** (`attribute_tokens`) is the same but does not copy keyword text either, and is what the lazy index uses inside each attribute list, so it is the policy under which most of a lazy index runs. Every policy sees the same tokens (the token count is the same for all three); they differ in what is decoded and copied. **ns per token** is wall time over token count under the attribute policy.

| file | tokens | byte scan | full tokenizer | index policy | attribute policy | ns per token |
|---|---|---|---|---|---|---|
| strings | 18.9 M | 936 MB/s | 166 MB/s | 263 MB/s | 309 MB/s | 8.6 |
| reals | 10.5 M | 1002 MB/s | 195 MB/s | 238 MB/s | 291 MB/s | 16.4 |
| aggregates | 13.2 M | 1023 MB/s | 202 MB/s | 232 MB/s | 272 MB/s | 13.9 |
| nested | 12.8 M | 1317 MB/s | 221 MB/s | 260 MB/s | 301 MB/s | 13.0 |
| TXG (mixed) | 13.3 M | 990 MB/s | 219 MB/s | 273 MB/s | 301 MB/s | 14.5 |

Two things fall out. Strings are not the driver: the string-heavy file is the fastest per byte and the cheapest per token, because `character_decoder::skip()` runs the same state machine without collecting and a long token is a cheap token. And the cost is flat, 9–16 ns per token whatever the token is; reals look expensive per byte only because they are short tokens.

I then tried your last suggestion, passing over the non-identifier tokens inside `next()` and returning only `(` `)` `,` `;`, names and strings. It drops the returned tokens by 13% on TXG and 34% on the reals file, and changes throughput by a few percent either way (TXG 301 → 290 MB/s, reals 291 → 284, nested 301 → 306), with lazy opens unchanged. The token counts explain it: on real files the structural tokens and the names are the majority, so most calls remain. Reverted, since it adds a loop for nothing measurable. `next_n()` or a page at a time would meet the same limit: the token still has to be produced.

So the honest reading is that the gap to the byte scanner is the per-token structure itself, not any token type, and closing it single-threaded would need the attribute structure (depth, commas) folded into the skip loop, which is the byte scanner in another place. I would rather not do that. What the branch does instead is chunk the index over threads (commit 13), which puts lazy under the default open on every model; the fourteen-model column is in the previous comment and the PR body.

For the record, the synthetic files also confirm the decoder cost you had in mind is real but small: with strings decoded (full tokenizer) the string file is the slowest at 166 MB/s; with them only ended it is the fastest.

aothms:
You're right. Never realized that ifc models are so sparse on literal values. On Duplex_A I get to 12%.

Token category | Count | Structural count | Ratio non-structural
-- | -- | -- | --
References (#123) | 80408 | 80408 |  
Comma | 70149 | 70149 |  
Opening parenthesis ( | 47977 | 47977 |  
Closing   parenthesis ) | 47977 | 47977 |  
Entity/type/keyword identifiers | 28736 | 28736 |  
Semicolon   ; | 27538 | 27538 |  
Equals = | 27529 | 27529 |  
Real   numbers | 19337 |   |  
$ null/unset | 10366 |   |  
Strings | 9071 |   |  
Enumerations (.FOO.) | 5708 |   |  
Integers | 68 |   |  
`*` omitted values | 9 |   |  
  | 374873 | 330314 | 0.118864

But I think the fact that all tokens are more or less in the same order of magnitude tells us that the overhead is likely in the inability to optimise across the next() boundary or the allocation of the token object etc.

So let's go ahead with the inversion to a Consumer model.

`template <typename Consumer> lexer::scan(Consumer& c) ...`

Make the function body exactly that of lexer::next() but wrap it with a while loop and do not allocate Token objects but call the appropriate function - other than that exact equal. Then remove next().

```C++
struct lazy_index_consumer {
    static constexpr bool decode_strings = false;
    static constexpr bool decode_numbers = false;
    // ...

    uint32_t depth = 0;
    uint32_t attribute = 0;

    FORCEINLINE void operator_(size_t pos, char c) {
        switch (c) {
        case '(':
            ++depth;
            break;
        case ')':
            --depth;
            break;
        case ',':
            if (depth == 1)
                ++attribute;
            break;
        }
    }

    FORCEINLINE void identifier(size_t pos, uint32_t id) {
        // directly build inverse index
    }
};

struct full_parse_consumer {
    static constexpr bool decode_strings = true;
    static constexpr bool decode_numbers = true;

    IFC_FORCEINLINE void operator_(size_t, char);
    IFC_FORCEINLINE void identifier(size_t, uint32_t);
    IFC_FORCEINLINE void string(size_t, std::string_view);
    ...
};
```

---

Also a completely different subject, but since SPF structure is so regular and IFC entity types follow typical usage patterns we should really try and investigate profile guided optimizations. If a local test confirms this helps, please investigate a separate PR where we build this into the build_win / build_rocky / build_osx github action build pipelines.

Moult:
_(AI-written, not verified by Dion.)_

**The consumer model** is in as commit 14. The body of `next()` is now `spf_lexer::scan(Consumer&)` in `spf_scan.h`: the same code wrapped in a loop, handing each token to the consumer's callbacks (`operator_`, `identifier`, `string`, `keyword`, `enumeration`, `binary`, `boolean`, `integer`, `real`, `literal`) instead of building a token; each callback returns whether to go on, and the consumer's constexpr flags say what is decoded (`decode_strings`, `decode_values`, `keep_keywords`). It lives in the header, with the SWAR helpers and number parsing it needs, so the consumer inlines into the loop. The lazy index is `attribute_consumer`: depth and attribute index from the operators, every name straight into the inverse index, the bounds of the first attribute if it is a string, done at the closing semicolon; the `attribute_tokens` policy is gone.

One deviation from what you wrote, and I'd rather state it than hide it: `next()` is not removed. It is kept as the consumer that stops after one token, six lines over `scan()`. The attribute reader, the header parser and the streamer's three-token lookahead pull tokens from inside recursive calls, 32 sites; inverting them into callbacks is a state-machine rewrite of the strict parser, which the profile never asked for (its time is in decoding and allocation). One tokenizer body, the index gets the inlined loop, the parser keeps pulling; the full tokenizer through the adapter measures the same as before (TXG 202–210 MB/s against 195–219). If you want the parser inverted too, that is its own PR.

Tokenizing 50 MB files with nothing decoded, in memory: the index policy through `next()` 214–247 MB/s, `scan()` with the inlined consumer 314–403 MB/s (TXG 247 → 345; the byte scanner it replaces 700–1300). On the fourteen models the index pass now runs at 266–343 MB/s in memory and 200–261 MB/s through 64 KB pages. It moves the lazy open only a little, because the index pass is no longer most of it: on one thread TXG / 210_King / OKgate22 0.55 / 1.59 / 2.11 → 0.54 / 1.57 / 2.07 s; on 12 threads 0.37 / 1.01 / 1.36 s against 0.44 / 1.19 / 1.94 s for the default open. Full tables, per-commit line counts and the profile section are in the PR body; 14 commits, net +2427 −415 lines (+1950 −403 outside tests). CI is green.

**PGO**, tested locally as you suggested: IfcParse and the schema plugins built with `-fprofile-generate`, trained by parsing TXG, 210_King and OKgate22 once each, rebuilt with `-fprofile-use`, against a plain build of the same tree. Median of three runs:

| model | plain, 1 thread | PGO, 1 thread | plain, 12 threads | PGO, 12 threads |
|---|---|---|---|---|
| TXG 58 MB (in the training set) | 1.03 s | 0.96 s (−7%) | 0.44 s | 0.39 s (−11%) |
| 210_King 147 MB (training) | 2.70 s | 2.46 s (−9%) | 1.17 s | 1.03 s (−11%) |
| OKgate22 231 MB (training) | 5.03 s | 4.71 s (−6%) | 1.90 s | 1.69 s (−11%) |
| 107 MB private (not in training) | 1.68 s | 1.53 s (−9%) | 0.72 s | 0.62 s (−13%) |
| 523 MB private, large instances (not in training) | 6.80 s | 6.55 s (−4%) | 2.12 s | 2.05 s (−4%) |

So 6–13% on ordinary models, about 4% on one dominated by long real lists, and it holds out of sample. Two things to know before putting it in the pipelines: a first attempt showed no gain at all because the `.gcda` files are keyed by the object path and the profile-use build was in a different directory, so GCC silently used nothing (`-Wno-missing-profile` hid it; keep that warning on in CI). And the `-fprofile-use` compiles of the generated schema files took enough memory that my machine's watchdog killed the build twice at four and two jobs; on a 7 GB CI runner that needs `-j1` for those units or a larger runner. If that is acceptable I'll open a separate PR for build_rocky, build_osx and build_win with a checked-in training model; the numbers above are the expected size of the win.

Moult:
_(AI-written, not verified by Dion.)_

A clarification on commit 14, so the numbers are not misread: it is an architecture change, not a performance one. Same fourteen models, the full matrix before the commit against the one after:

| | before → after |
|---|---|
| index pass in isolation (tokenizer with nothing decoded) | 206–273 → 266–343 MB/s, +20–30% on every model |
| lazy open, 1 thread | −6% to +2%, mean −1.2% |
| lazy open, 12 threads | −1% to +11%, mean +1.8% |
| strict parse, 1 thread | −1% to +7%, mean about +2% |
| memory, all modes | identical |

Run-to-run scatter on this machine is about ±3%, so the open times are flat and the strict parse is unchanged by construction (`next()` is the adapter). The isolated gain does not reach the open because the tokenizer pass is now a small share of it: on TXG 0.17 s of a 0.54 s single-threaded lazy open, the rest being the shells, the inverse sort and the GlobalId map, which the commit does not touch; at 12 threads that pass is divided by twelve and the serial merge dominates.

What the commit does deliver is the shape you asked for: one tokenizer body, the lazy index inlined into it as a consumer instead of a policy returning a token per call, and one policy fewer. The PR body's profile section says the same. The end-to-end before/after tables at the top of the PR body are unchanged by it.

Moult:
_(AI-written, not verified by Dion.)_

Two more commits at the end of the branch, 15 and 16. They are experiments, kept separate so they are easy to drop: the series up to commit 14 is the one under review, and we assume 15 and 16 are discarded unless you want them. Nothing above them changed and the numbers in the PR body up to commit 14 are as they were; this comment's numbers are on top of those.

**Why these two.** With the tokenizer inlined, a lazy open of TXG on one thread (0.54 s) breaks down roughly as follows, from the callgrind profile and the isolated timings:

| phase | cost | notes |
|---|---|---|
| tokenizer pass (scan) | ~0.17 s | parallelises |
| 918k shell allocations | ~0.06 s | one malloc each |
| 918k inserts into the name table | ~0.07 s | `unordered_map` |
| 1.66 M inverse records: append and sort | ~0.10 s | introsort on 12-byte records |
| 337k GlobalId inserts | ~0.04 s | hash of 22 bytes each |
| by-type lists, offsets | ~0.03 s | |

Everything below the first row was serial even on 12 threads, which is why the lazy open only went from 0.54 s to 0.37 s on twelve cores, and the same serial merge is what held the strict parse at 2–3x on 12 cores: of its 0.44 s on TXG, roughly 0.3 s was the merge and the sort. So the two levers with contained changes were the sort and the merge.

**Commit 15, radix sort of the inverse records.** `inverse_index::sort_records()`: a stable LSD radix sort on `referenced_id`, 11 bits per pass and as many passes as the largest id needs, then `record_less` within each run of equal ids, so the order is exactly what `std::sort` gave; small inputs still use `std::sort`. 46 lines in one function. Measured in isolation against commit 14, five models, best of three:

| model | strict, 1 thread | strict, 12 threads | lazy, 1 thread | lazy, 12 threads |
|---|---|---|---|---|
| TXG 58 MB | 1.03 → 0.96 s (−7%) | 0.43 → 0.36 s (−16%) | 0.52 → 0.47 s (−10%) | 0.36 → 0.29 s (−19%) |
| 210_King 147 MB | 2.68 → 2.46 s (−8%) | 1.16 → 0.96 s (−17%) | 1.57 → 1.34 s (−15%) | 1.00 → 0.78 s (−22%) |
| OKgate22 231 MB | 5.00 → 4.70 s (−6%) | 1.89 → 1.67 s (−12%) | 2.05 → 1.83 s (−11%) | 1.33 → 1.11 s (−17%) |
| 107 MB (private) | 1.68 → 1.55 s (−8%) | 0.73 → 0.60 s (−18%) | 1.02 → 0.90 s (−12%) | 0.62 → 0.50 s (−19%) |
| 523 MB (private, few large instances) | 6.91 → 6.71 s (−3%) | 2.10 → 2.01 s (−4%) | 2.97 → 2.96 s (0%) | 1.49 → 1.42 s (−5%) |

**Commit 16, sort and index per worker, merge sorted runs.** Each parser (and index) worker sorts its own records, builds its own by-type lists and reads its own GlobalIds; the main thread appends the lists, inserts names and GlobalIds, and merges the sorted runs pairwise (`inverse_index::merge_sorted`, O(n log k)) instead of concatenating and sorting. Instance order, GlobalId precedence and record order are unchanged; the equality tests cover it. The workers sort their runs in place (no radix buffer on the worker's arena, no shrink: the merge copies them anyway). About 140 lines, mostly moved. In isolation against commit 14:

| model | strict, 1 thread | strict, 12 threads | lazy, 1 thread | lazy, 12 threads |
|---|---|---|---|---|
| TXG 58 MB | +1% | 0.43 → 0.36 s (−16%) | +2% | 0.36 → 0.27 s (−25%) |
| 210_King 147 MB | +1% | 1.16 → 0.89 s (−23%) | −1% | 1.00 → 0.74 s (−26%) |
| OKgate22 231 MB | 0% | 1.89 → 1.41 s (−25%) | +2% | 1.33 → 1.06 s (−20%) |
| 107 MB (private) | +1% | 0.73 → 0.55 s (−25%) | +1% | 0.62 → 0.48 s (−23%) |
| 523 MB (private) | −1% | 2.10 → 1.92 s (−9%) | −2% | 1.49 → 1.38 s (−7%) |

Single-threaded it is neutral by construction. A third experiment, deferring the lazy GlobalId map to first use, gained 0–8% on the 12-thread lazy open only and is not included.

**Both together**, against commit 14, all fourteen models:

| model | strict, 1 thread | strict, 12 threads | lazy, 1 thread | lazy, 12 threads | peak memory, strict 12 threads | Python default open | Python lazy open |
|---|---|---|---|---|---|---|---|
| TXG 58 MB | 1.08 → 0.99 s (-8%) | 0.44 → 0.37 s (-16%) | 0.54 → 0.47 s (-13%) | 0.37 → 0.28 s (-24%) | 381 → 392 MB | 0.44 → 0.37 s (-16%) | 0.37 → 0.28 s (-24%) |
| 210_King 147 MB | 2.73 → 2.50 s (-8%) | 1.19 → 0.90 s (-24%) | 1.57 → 1.35 s (-14%) | 1.01 → 0.75 s (-26%) | 903 → 933 MB | 1.19 → 0.90 s (-24%) | 1.01 → 0.76 s (-25%) |
| OKgate22 231 MB | 5.06 → 4.78 s (-6%) | 1.94 → 1.40 s (-28%) | 2.07 → 1.91 s (-8%) | 1.36 → 1.09 s (-20%) | 1380 → 1434 MB | 1.93 → 1.42 s (-26%) | 1.36 → 1.11 s (-18%) |
| 10 MB (private) | 0.18 → 0.15 s (-17%) | 0.08 → 0.06 s (-25%) | 0.10 → 0.09 s (-10%) | 0.07 → 0.06 s (-14%) | 81 → 78 MB | 0.09 → 0.07 s (-22%) | 0.08 → 0.06 s (-25%) |
| 74 MB (private) | 1.26 → 1.13 s (-10%) | 0.53 → 0.41 s (-23%) | 0.60 → 0.56 s (-7%) | 0.42 → 0.32 s (-24%) | 485 → 479 MB | 0.50 → 0.43 s (-14%) | 0.40 → 0.33 s (-18%) |
| 81 MB (private) | 1.30 → 1.22 s (-6%) | 0.52 → 0.42 s (-19%) | 0.69 → 0.63 s (-9%) | 0.45 → 0.37 s (-18%) | 478 → 470 MB | 0.55 → 0.44 s (-20%) | 0.45 → 0.37 s (-18%) |
| 107 MB (private) | 1.74 → 1.60 s (-8%) | 0.75 → 0.57 s (-24%) | 1.03 → 0.91 s (-12%) | 0.64 → 0.49 s (-23%) | 608 → 622 MB | 0.74 → 0.57 s (-23%) | 0.63 → 0.50 s (-21%) |
| 249 MB (private) | 4.26 → 3.86 s (-9%) | 1.92 → 1.42 s (-26%) | 2.32 → 2.01 s (-13%) | 1.61 → 1.15 s (-29%) | 1438 → 1478 MB | 1.91 → 1.45 s (-24%) | 1.60 → 1.16 s (-28%) |
| 279 MB (private) | 3.77 → 3.71 s (-2%) | 1.13 → 1.04 s (-8%) | 1.61 → 1.57 s (-2%) | 0.83 → 0.77 s (-7%) | 1005 → 1015 MB | 1.13 → 1.06 s (-6%) | 0.83 → 0.79 s (-5%) |
| 315 MB (private) | 5.59 → 5.12 s (-8%) | 2.24 → 1.70 s (-24%) | 2.88 → 2.47 s (-14%) | 1.86 → 1.37 s (-26%) | 1852 → 1915 MB | 2.24 → 1.74 s (-22%) | 1.82 → 1.40 s (-23%) |
| 349 MB (private) | 5.69 → 5.30 s (-7%) | 2.41 → 1.82 s (-24%) | 3.18 → 2.81 s (-12%) | 2.02 → 1.52 s (-25%) | 1967 → 1979 MB | 2.44 → 1.81 s (-26%) | 2.01 → 1.54 s (-23%) |
| 523 MB (private) | 6.94 → 6.91 s (-0%) | 2.11 → 2.04 s (-3%) | 2.99 → 2.87 s (-4%) | 1.51 → 1.40 s (-7%) | 1825 → 1851 MB | 2.14 → 1.99 s (-7%) | 1.53 → 1.43 s (-7%) |
| 603 MB (private) | 8.55 → 7.89 s (-8%) | 3.23 → 2.67 s (-17%) | 4.38 → 3.99 s (-9%) | 2.61 → 2.12 s (-19%) | 2649 → 2710 MB | 3.29 → 2.68 s (-19%) | 2.64 → 2.14 s (-19%) |
| 787 MB (private) | 12.48 → 11.27 s (-10%) | 5.62 → 4.14 s (-26%) | 7.28 → 6.18 s (-15%) | 4.85 → 3.55 s (-27%) | 4384 → 4454 MB | 5.69 → 4.18 s (-27%) | 4.88 → 3.55 s (-27%) |

**Memory.** Single-threaded nothing changes. At 12 threads the peak is one transient copy of the records higher, for the merge: +1–4% strict, +0–10% lazy (the 787 MB model: 2684 → 2863 MB). Resident memory after the open is within ±2% once the allocator's free pages are trimmed, and 0–7% higher before trimming, which is glibc keeping the freed per-worker runs on the worker arenas (`MALLOC_ARENA_MAX=1` removes the difference). A first form of commit 16 ran the radix sort inside the workers; its buffers stayed on those arenas and resident memory after a 12-thread open was 3–10% higher, so the workers now sort their runs in place with `std::sort`, which costs nothing at 12 threads (the runs are a twelfth of the records each) and keeps the radix sort for the single-threaded paths.

If you would rather keep the PR at commit 14, say so and I will drop these two; they stand on their own and can come back later.

aothms:
I still have these numbers in my head:

> Index pass, nothing decoded, in memory: 214–247 MB/s through next() → 266–343 MB/s through scan() (the byte scanner it replaces: 700–1300 MB/s); 

Where do we stand now? I don't measure a major difference with the first PR, testing on Windows MSVC. But my machine is hard to use for benchmarking because it constantly throttles and has heterogeneous cpu cores. Pushed some fixes and further ideas for msvc.

Moult:
> I still have these numbers in my head

I was wondering too where those improvements "show up" and when I asked AI, it basically said it doesn't because of https://github.com/IfcOpenShell/IfcOpenShell/pull/9492#issuecomment-5672016858 - so switching to the consumer model doesn't seem to make a significant performance difference, just a code architectural change.

> Where do we stand now?

All the latest numbers are in the first post: https://github.com/IfcOpenShell/IfcOpenShell/pull/9492#issue-5443106218 - you can see the edit history (which I've been checking to make sure we don't go backwards in terms of performance).

Moult:
_(AI-written, not verified by Dion.)_

Where we stand, measured here on your four commits (head 5f4ecdc30) against the branch as it was before them (788d25767), gcc 15 -O3 on Linux, best of three on a quiet 12-core box. The tokenizer is header-only, so the "before" bench binary was compiled against the old headers and the "after" one against yours; the opens go through the respective libraries.

The first table is the tokenizer in isolation: megabytes of DATA-section text per second through the tokenizer alone, no parser behind it, nothing built. "Index scan" is the tokenizer driven by the index consumer, the one the lazy open uses (token boundaries, names and string ends, nothing decoded); "full tokenizer" is every token decoded through `next()`, as the parser sees them. Before is the branch at 788d25767, after is your head 5f4ecdc30.

| model | index scan, in memory | index scan, paged | full tokenizer, in memory | full tokenizer, paged |
|---|---|---|---|---|
| TXG 58 MB | 342 → 910 MB/s (+166%) | 270 → 562 MB/s (+108%) | 210 → 274 MB/s (+30%) | 221 → 317 MB/s (+43%) |
| 210_King 147 MB | 273 → 873 MB/s (+220%) | 216 → 517 MB/s (+139%) | 193 → 297 MB/s (+54%) | 192 → 330 MB/s (+72%) |
| OKgate22 231 MB | 319 → 915 MB/s (+187%) | 250 → 537 MB/s (+115%) | 205 → 286 MB/s (+40%) | 210 → 324 MB/s (+54%) |
| 107 MB (private) | 284 → 940 MB/s (+231%) | 226 → 583 MB/s (+158%) | 194 → 274 MB/s (+41%) | 197 → 305 MB/s (+55%) |

The index pass is now at the byte scanner's throughput (700–1300 MB/s on these models in #9473) while staying the one tokenizer, and the full tokenizer gains 30–54%, with your three tokenizer commits together.

End to end:

| model | lazy, 1 thread | lazy, 12 threads | strict, 1 thread | strict, 12 threads |
|---|---|---|---|---|
| TXG 58 MB | 0.42 → 0.35 s (-17%) | 0.26 → 0.24 s (-8%) | 0.88 → 0.88 s (+0%) | 0.35 → 0.35 s (+0%) |
| 210_King 147 MB | 1.20 → 0.95 s (-21%) | 0.70 → 0.66 s (-6%) | 2.17 → 2.15 s (-1%) | 0.83 → 0.81 s (-2%) |
| OKgate22 231 MB | 1.66 → 1.35 s (-19%) | 1.01 → 0.95 s (-6%) | 4.38 → 4.37 s (-0%) | 1.31 → 1.32 s (+1%) |
| 107 MB (private) | 0.81 → 0.64 s (-21%) | 0.45 → 0.42 s (-7%) | 1.38 → 1.37 s (-1%) | 0.51 → 0.52 s (+2%) |

The lazy open on one thread drops by a fifth; at 12 threads the scan is a smaller share of the open (the index build and the merge are the rest), and the strict parse does not move because its time is in the parser, not the tokenizer, as before.

In numbers, for TXG: the lazy open on one thread was 0.42 s, of which the scan was about 0.17 s (58 MB at 342 MB/s) and the index build (shells, name table, inverse records and their sort, GlobalIds, by-type lists) the other 0.25 s. At 910 MB/s the scan is 0.064 s, so the predicted open is 0.31 s and the measured one 0.35 s: the scan's gain is all there is to gain from it, and the index build is now 80% of a single-thread lazy open. At 12 threads the scan was already spread over the cores, about 0.015 s of the 0.26 s, so tripling it is worth a hundredth of a second; what is left there is the serial merge. The strict parse does not use the index scan at all, and the full tokenizer's 30–54% in isolation would be about 0.07 s of TXG's 0.88 s parse, which is within the noise once it is interleaved with instance construction and reference resolution. Note that these are best-of-three timings from a short session, and that single runs on this box vary by up to about 10% between sessions, so compare percentages within a table rather than times across tables.

On "no major difference with the first PR": that matches what we see. Against #9473 the lazy open on one thread is about level (TXG 0.43 s there, 0.35 s now with your commits), because a lazy open is dominated by building the index (shells, name table, inverse records, GlobalIds), not by the scan, and the strict parse is the same parser in both PRs. The differences of this series are against v0.9.0 head (the tables in the body) and in the structure: one tokenizer, references in place, no second scanner.

Your tests pass here (18 Catch2 cases, 2.1 M assertions), CI is green on your head. One small thing noticed reading the pool change: a consumer with `decode_values` true and `keep_keywords` false would now dereference the null text pointer; a `static_assert` that the first implies the second would document the contract. The PR body is now re-measured at your head on all fourteen models, with commits 15 and 16 folded into the one before/after.

aothms:
> The index pass is now at the byte scanner's throughput (700–1300 MB/s on these models in #9473) while staying the one tokenizer, and the full tokenizer gains 30–54%, with your three tokenizer commits together.

Thanks that's exactly what we hoped for!

> 210_King 147 MB 	273 → 873 MB/s (+220%) 	216 → 517 MB/s (+139%) 	193 → 297 MB/s (+54%) 	192 → 330 MB/s (+72%)

This is a bit inconsistent though: How come paging is slower for lazy, but faster for full parse. It's not a major concern, but it would help if we understand this. Might be a bug somewhere or unrealised optimisation potential.

I still need to review the multi-threading commits. These are the largest. Will take me some time.

Moult:
_(AI-written, not verified by Dion.)_

Not a bug: two effects with opposite signs, and which one wins depends on how much work the consumer does per byte. Callgrind with cache simulation on TXG, 58 MB, one thread:

| mode | instructions | data refs | last-level cache misses | measured |
|---|---|---|---|---|
| index scan, in memory | 1.56 G | 394 M | 1.91 M | 899 MB/s |
| index scan, paged | 2.53 G | 792 M | 0.15 M | 544 MB/s |
| full tokenizer, in memory | 4.16 G | 1.31 G | 1.91 M | 261 MB/s |
| full tokenizer, paged | 4.36 G | 1.39 G | 0.15 M | 304 MB/s |

**Paging costs instructions.** Every `peek()` and `increment()` on the paged reader checks the cursor against the file size and the current page (`cached_()`: four comparisons, two of them loads through the implementation pointer) where the in-memory reader checks one bound and indexes. The index consumer now walks discarded literals with the scalar loop (commit 19), one `peek()` and one `increment()` per byte, so the paged index scan executes 63% more instructions and twice the data references of the in-memory one; the scan function itself goes from 0.99 G to 1.76 G instructions. The full tokenizer takes the SWAR path, eight bytes per check, and does far more work per byte anyway (string pool, from_chars, the token object), so the same checks are 5% of its instructions.

**Paging saves cache misses.** The in-memory scan streams the whole buffer through the cache once: 1.9 M last-level misses on TXG, one per line, for both consumers. Through 64 KB pages the bytes were just copied in by `read()` and are hot when scanned: 0.15 M misses. The hardware prefetcher hides most of the streaming cost, but not all of it when the tokenizer's own stores compete, and for the full tokenizer that residual is the 5–15% by which paged wins. The index scan gets the same cache saving, but it is dominated by the extra instructions.

So the paged index scan is the unrealised optimisation: the discarded-literal loop should not pay a page check per byte. Commit 21 (8623e9054, pushed on top of yours) gives that loop the treatment your name fast path has: it takes the reader's `span()` and walks it with a local pointer to the next delimiter, handing over to the per-byte loop only when a span ends first (a page boundary). Callgrind on TXG: index scan 1.56 G → 1.21 G instructions in memory, 2.53 G → 1.76 G through pages; your 18 Catch2 cases pass, the one-byte-page case included. What is left of the paged gap is the operators (one `read()` each, 7.4 M of the 13.3 M tokens), the string decoder's `skip()` and the whitespace skip, which still go byte by byte through the reader; those could take the same route. Timed against your head, best of three on a quiet box:

| model | index scan, in memory | index scan, paged | full tokenizer, in memory | full tokenizer, paged | lazy, 1 thread | lazy, 12 threads | strict, 1 thread |
|---|---|---|---|---|---|---|---|
| TXG 58 MB | 927 → 1073 MB/s (+16%) | 569 → 709 MB/s (+25%) | 278 → 278 MB/s (+0%) | 323 → 322 MB/s (-0%) | 0.35 → 0.34 s (-3%) | 0.24 → 0.24 s (+0%) | 0.86 → 0.87 s (+1%) |
| 210_King 147 MB | 882 → 982 MB/s (+11%) | 539 → 631 MB/s (+17%) | 300 → 302 MB/s (+1%) | 337 → 335 MB/s (-1%) | 0.93 → 0.92 s (-1%) | 0.64 → 0.64 s (+0%) | 2.11 → 2.12 s (+0%) |
| OKgate22 231 MB | 933 → 1073 MB/s (+15%) | 548 → 667 MB/s (+22%) | 291 → 291 MB/s (+0%) | 329 → 328 MB/s (-0%) | 1.34 → 1.31 s (-2%) | 0.94 → 0.95 s (+1%) | 4.28 → 4.34 s (+1%) |
| 107 MB (private) | 952 → 1149 MB/s (+21%) | 594 → 760 MB/s (+28%) | 278 → 278 MB/s (+0%) | 312 → 312 MB/s (+0%) | 0.64 → 0.62 s (-3%) | 0.42 → 0.41 s (-2%) | 1.34 → 1.37 s (+2%) |

The index scan gains 11–21% in memory and 17–28% through pages, the full tokenizer is untouched, and the opens move within noise, as the arithmetic above says they should: the scan is about 15% of a single-thread lazy open now. It is a small, self-contained commit at the end of the branch; drop it if you would rather not have it.

aothms:
One last thing that I would propose. Can we also have the full parse use spf_scan functionality. We detect the pattern `{name} '=' {keyword}` (same circular buffer) but do so in the consumer - same as the while loop in file::read(). So the code gets even more centralised as a consequence and we realise higher degree of code reuse and can then remove spf_lexer::next(). Likely benefits performance as well. I would suggest to move spf_scan.h inline in the same file next to the lexer.

Moult:
_(AI-written, not verified by Dion.)_

Done, in two commits on the branch:

- **22, the parser as a consumer of `scan()`.** `list_consumer` reads an attribute list from just past its `(` through the `)` that closes it; the recursion of `load_attributes()` → `read_direct_aggregate()` → `load()` is an explicit stack of frames, one per open parenthesis: a list frame fills the instance's slots (an entity's list, or a typed value such as `IFCLABEL('x')` read on its owner's behalf), an aggregate frame accumulates a `direct_aggregate`, a skip frame counts parentheses for a value beyond the schema's attribute count. Inverse records, unresolved references, the attribute-count warning, the enumeration and unknown-type messages and their offsets are made as before. `instance_header_consumer` is the top level of the DATA section, your three-token window as consumer states: `#name`, `=`, the keyword (looked up once per spelling), then `(` hands the list to a visit callback, which scans the list with its own consumer and returns; the outer scan continues from where that stopped, so the lazy index keeps its own consumer with its own flags. It serves the parallel chunk, the lazy index and the streamer (one instance per `read_instance()`, so the serial parse, the streaming wrapper and the RocksDB serializer see what they saw). The header section is a consumer too, `has_semicolon()`, `semicolon_count()` and `decode_spf_string()` one small consumer each. `next()` has no caller left after this commit.
- **23, the removal.** `next()`, the one-token consumer, the `token` object with its accessors and the two token policies go; the tokenizer body moves from `spf_scan.h` into `parse.h` under the lexer, with a local enum where it used the token type's constants; the tests that used `next()` as the reference scan with a consumer that decodes everything.

17 Catch2 cases (per-instance equality against the fixtures, lazy, parallel and paged equality, your page and invalid-name cases) and 2296 pytest cases pass, the validate suite with its message offsets included. Three differences, all on malformed input, are in commit 22's message: a token other than `(` after a header entity's name is an error instead of being swallowed, an unknown type in the serial parse is logged once per spelling instead of per instance, and an invalid token inside a typed value ends the parse with INVALID_SYNTAX instead of being logged and skipped.

Commit 21 → commit 23 on the same models, best of three on a quiet box:

| model | strict, 1 thread | strict, 12 threads | paged, 1 thread | paged, 12 threads | lazy, 1 thread | lazy, 12 threads | peak, strict 1 thread |
|---|---|---|---|---|---|---|---|
| TXG 58 MB | 0.90 → 0.83 s (-8%) | 0.35 → 0.33 s (-6%) | 0.91 → 0.85 s (-7%) | 0.36 → 0.34 s (-6%) | 0.35 → 0.32 s (-9%) | 0.25 → 0.24 s (-4%) | 365 → 366 MB |
| 210_King 147 MB | 2.22 → 2.11 s (-5%) | 0.83 → 0.80 s (-4%) | 2.23 → 2.15 s (-4%) | 0.87 → 0.80 s (-8%) | 0.93 → 0.86 s (-8%) | 0.66 → 0.64 s (-3%) | 871 → 871 MB |
| OKgate22 231 MB | 4.50 → 4.23 s (-6%) | 1.35 → 1.24 s (-8%) | 4.48 → 4.29 s (-4%) | 1.35 → 1.28 s (-5%) | 1.32 → 1.22 s (-8%) | 0.96 → 0.95 s (-1%) | 1347 → 1347 MB |
| 107 MB (private) | 1.40 → 1.33 s (-5%) | 0.52 → 0.49 s (-6%) | 1.44 → 1.35 s (-6%) | 0.54 → 0.50 s (-7%) | 0.62 → 0.56 s (-10%) | 0.42 → 0.41 s (-2%) | 590 → 590 MB |

The strict parse gains 4–8% on one thread, the lazy open 8–10% on one thread (its header loop was the last pull-based piece), and peak memory does not move. The two commits together are +1165 −1092 lines outside tests: one code path, the same size as before.

Moult:
_(AI-written, not verified by Dion.)_

CI caught one regression of commit 22 that my local set had not covered: `test_chunked_stream`. The streaming loop asks `has_semicolon()` before every `read_instance()`, so `read_instance()` has to consume the `;` after the instance it returns, which the pull version did as its last token and the consumer version did not; a chunked stream therefore ended after its first page. Commit 24 gives the instance-header consumer a one-instance mode that takes the `;` and stops there (a token other than `;` in that place is left for the next scan). The whole `test/` directory now passes as CI runs it: 2537 passed, 37 skipped, 1 xfailed.

aothms:
Can we move the multi-threading (too little time to review) out of this PR into a new one - and then see what a rebased version of this PR minus multi-threading looks like to see if we can merge it in for 0.9.0?

Moult:
_(AI-written, not verified by Dion.)_

Done. #9492 is now the series without the multithreading, rebased onto v0.9.0 (90c350e35): 21 commits, +3061 −1739 lines outside tests, and it merges cleanly (the rebase itself had no conflicts with upstream; the four commits that touched the parser files since our base are elsewhere in it). What went out are the three threading commits (parallel parse, parallel lazy index, per-worker sort and merge) and with them `file::parse_threads()`, the split-point scan, the worker helpers on the reader and the index, and the 5-thread test cases; what stayed of the first of them is the lazy index's shared instance-header loop, which is serial. Your MSVC export fix turned out to be already upstream and dropped out as empty. The lazy index is in its inline form again, with no per-chunk output record. Each commit the rebase stopped on was built and tested before continuing, and the tip passes 17 Catch2 cases and the whole Python `test/` directory (2557 passed).

The multithreading is #9587, three commits on top of this branch: the parallel parse (with the sorted-run merge folded in, since nothing else used the concatenate-and-sort form), the parallel lazy index, and `open(threads=...)`; the thread count is configurable through `file::parse_threads()`, `IFCOPENSHELL_PARSE_THREADS` or that argument, one per core capped at 16 by default. Its body carries the 12-thread columns that used to be here.

The tables in this PR's body are the ones measured before the split, minus the threaded columns, since nothing in the serial paths changed; the Python table is re-measured on this branch, whose default open is now the single-threaded parse (-34% to -46% against v0.9.0 head; the lazy open -72% to -87%).

Moult:
_(AI-written, not verified by Dion.)_

CI fails on both PRs in ifcquery's clash tests, and it is the SWIG fast-proxy commit meeting something that arrived upstream after our base: `%feature("kwargs") ifcopenshell::geom::tree;` (from the tree-backends work on the 22nd) and ifcquery calling `clash_intersection_many(..., tolerance=..., check_all=True)`. Without `-fastproxy`, SWIG emits a Python `def` with named parameters for a keyword-enabled method and calls the compiled function positionally, so keywords work. With `-fastproxy`, the `def` is replaced by the compiled function itself, which is positional-only: `TypeError: tree_clash_intersection_many() takes no keyword arguments`. SWIG 4.3.1 has no per-class opt-out from fast-proxy, so anything that adopts `%feature("kwargs")` from now on would break the same way.

Three ways out, and I would like your call before rewriting the branch again:

1. Drop the fast-proxy commit from both PRs (my recommendation). It is self-contained (the CMake flag, an include-order move in `IfcPython.i`, and the stub validator reading signatures from docstrings; 110 lines) and its gain was about 8% on attribute-heavy Python loops; nothing in the parser depends on it.
2. Keep it and add SWIG's global `-keyword`, which makes the compiled wrappers accept keywords by name; SWIG cannot do that for overloaded methods, of which this wrapper has many, so it warns and leaves those positional. Works for the tree methods today, wide blast radius.
3. Keep it and hand-write Python shadows for the keyword-enabled methods. Works, but a special case that recurs with every method that gains the feature.

If 1, the serial PR becomes 20 commits and the multithreading PR is rebased on top; the numbers in the bodies are unaffected except the one sentence about the fast-proxy gain.

aothms:
My vote is on 3. if it's actually only the tree. I'm not super confident on why this only affects the tree. I cannot imagine we use keyword args that little. But I do know there is a work-around exactly there. Where we unpack the kwargs into a linear sequence of args: https://github.com/IfcOpenShell/IfcOpenShell/blob/895328d0c6102451c2ed14379be01dbd007510d0/src/ifcopenshell-python/ifcopenshell/geom/main.py#L437-L440 So adding it for the other tree methods isn't more ugly per se.

Moult:
_(AI-written, not verified by Dion.)_

It is only the tree: `%feature("kwargs") ifcopenshell::geom::tree;` is the one keyword-argument declaration in the wrapper (IfcGeomWrapper.i:661); every other method is `*args`, and for those fast-proxy changes nothing. And the Python `tree` class already does what you describe for `select`, `select_box`, `clash_collision_many` and `clash_clearance_many`; the two methods without a shadow were `clash_intersection_many` (the one ifcquery and ifcclash call with keywords) and `select_ray`.

Since that is independent of this branch, it goes in as its own small PR against v0.9.0, #9594, adding those two shadows in the same style; #9492 and #9587 rebase onto it once it is in, with the fast-proxy commit unchanged.

PR #9497, opened 2026-09-14
Handle null Styles when loading surface styles

CyrilWaechter:
Fixes a `TypeError: 'NoneType' object is not iterable` raised while loading a project whose IFC file contains an `IfcSurfaceStyle` with a NULL `Styles` set.

## Cause

`tool.Style.get_style_elements` iterated `style.Styles` unconditionally. While `IfcSurfaceStyle.Styles` is mandatory (`SET [1:?]`) in IFC, files produced by some applications (e.g. Revit-derived styles named `color <rgb>`) and files merged/appended from them can contain `$`. During `load_project_elements`, `import_ifc.create_style` calls `get_style_elements`, which aborted the import.

The same unguarded iteration existed in `get_surface_shading_style` and `get_surface_texture_style`, which are reached when editing such a material. `import_presentation_styles` already used the defensive `... or []` pattern.

## Changes

- Guard the `style.Styles` iteration in `get_style_elements`, `get_surface_shading_style`, and `get_surface_texture_style`.
- Add `TestGetStyleElements` covering both the `IfcSurfaceStyle` entity path and the Blender-material path used during import.

## Testing

`pytest test/tool/test_style.py` -> 35 passed, 1 skipped (Blender 5).

This PR is AI-generated.


CyrilWaechter (commented):
Reviewed.

PR #9498, opened 2026-09-14
Report required CI checks for skipped paths

CyrilWaechter:
`compile-and-test (ON)` / `(OFF)` are required checks, but they come from `ci.yml`, whose top-level `paths:` filter excludes `src/bonsai/**`. A workflow skipped by path filtering leaves its required checks in a "Pending" state forever, so Bonsai-only pull requests are permanently blocked (and show only `lint-formatting`). Identical pattern on #9479, #9489, #9491, #9494.

## Change

- Remove the top-level `paths:` filter so `ci.yml` always triggers.
- Add a lightweight `changes` job (`dorny/paths-filter@v4`) that reproduces the previous list of relevant paths.
- Gate the heavy `compile-and-test` matrix job with `if: needs.changes.outputs.code == 'true'`.

Per GitHub's documented behavior, a job skipped by a conditional reports Success, whereas a workflow skipped by path filtering does not. So:

- library paths changed -> full build runs exactly as before;
- only `src/bonsai/**` (or docs, etc.) changed -> `compile-and-test` is skipped and its required checks pass, unblocking the PR.

Build reuse is unchanged: the existing `ccache` action already reuses compiled objects across runs.

## Verification

This PR only touches `.github/workflows/ci.yml`, which is in the filtered list, so the full `compile-and-test` run is exercised here. The skip path will be exercised by a later `src/bonsai/**`-only PR.

This PR is AI-generated.


CyrilWaechter:
Posted the full comparison over on #9454 so the alternatives sit in one place: https://github.com/IfcOpenShell/IfcOpenShell/pull/9454#issuecomment-5676397372

Short version: #9454 and this PR touch the same lines in `ci.yml`, so only one should land. This is the variant that keeps the `pull_request` path filter (skips the heavy build on Bonsai/docs-only PRs); #9454 drops the filter for PRs instead. Happy to close this one if maintainers prefer the simpler route.

Moult:

Thanks @CyrilWaechter, this is the approach that landed. #9531 renamed `ci.yml` to `test-ifcopenshell.yml`, which is why this branch conflicts, so #9533 carries your `changes`-job pattern onto the renamed file, adds the same gate to `Test Bonsai` so it can become a required check, and adds a `workflow_dispatch` trigger. Closing this one in favour of #9533, with your `some-with-excludes` note and path list kept as they were.

PR #9499, opened 2026-09-14
Add Bonsai tests to pull requests

CyrilWaechter:
Stacked on #9498.

Bonsai tests only run today in `ci-bonsai-daily.yml`, which triggers on `push` to `v0.9.0` (and `workflow_dispatch`) — never on pull requests. So a PR that touches `src/bonsai/**` gets only `lint-formatting` before merge; regressions surface afterwards.

## Change

Add `.github/workflows/ci-bonsai-pr.yml`, triggered on `pull_request` for `src/bonsai/**` (and itself) plus `workflow_dispatch`:

1. Check out the PR.
2. Check out the public `IfcOpenShell/bonsai_unstable_repo` and install the latest Bonsai build to provide a working Blender/ifcopenshell environment (no secrets needed, so fork PRs work).
3. Download Blender 5.2.0.
4. Run the Bonsai tests from the PR source (`make test-tool`).

The tests resolve `import bonsai` to the PR's source tree (pytest inserts `src/bonsai` on `sys.path`, since `test/` is a package), so the PR code is what gets exercised.

## Deliberately not required yet

The Bonsai suite currently has pre-existing failures unrelated to any single PR (`ci-bonsai-daily` is red on `v0.9.0`). This check is added **non-required** so it can be validated and the failures addressed before it becomes a gate.

This PR is AI-generated.


Moult:

Superseded by #9527, which builds the extension from the pull request's own checkout so changes to Bonsai's code are what gets tested, runs all three suites, and moves the ifcsverchok smoke test there as well. Thanks for the groundwork here, the shape of the job is yours.

Issue #3052, opened 2023-04-22
When i try and delete this particular object, BB freezes. 


theoryshaw:
When i try and delete this particular object, BB freezes. 
file: [merged.ifc.txt](https://github.com/IfcOpenShell/IfcOpenShell/files/11301252/merged.ifc.txt)

![2023-04-22_07-00-29_Blender_blender](https://user-images.githubusercontent.com/507113/233783488-23d2f92e-81b5-400b-b3a7-10329ad0f6e2.gif)

Moult:
Surprisingly it's actually working, it's just working incredibly slowly until it freezes. If you ordered a pizza and watched a Seinfeld episode it'll probably finish.

The root cause is that removing an element X that is referenced in a very long list is very slow. Potentially @aothms knows of a more fundamental fix on the IfcOpenShell end, but for now the solution recognises that is a problem and preemptively clears those huge list attributes.

Moult:
Probably worth mentioning that the object you're trying to delete has 10,000 polygons which is completely ridiculous in the first place, but alas that's what Revit tends to do: https://wiki.osarch.org/index.php?title=Revit_setup_for_OpenBIM/Revit_and_IFC_Geometry#Reducing_curves_in_profiles_that_are_swept.2C_revolved.2C_or_blended

The worst offender is a little glazed panel detail further along with over 30,000 polygons. Thankfully this fix now means that it now merely takes seconds to delete these objects. Hooray!

PR #9502, opened 2026-09-15
[AI-generated, unverified] ifcopenshell-python: decide remove_deep2 containment in C++ without materializing inverses

Moult:

**Base: `v0.9.0`.** First opened stacked on #9492 (`open-perf-v2`) because that is where it was built and verified; it doesn't depend on that PR's changes, so it was rebased onto `v0.9.0` (the commits cherry-pick cleanly) and retargeted on 2026-09-17. It lands after #9508, whose fixed-width RocksDB keys this branch's one hand-built `v|` prefix follows.

## Problem

`root.remove_product` spends most of its time in `util.element.remove_deep2`, and most of *that* in deciding whether a subelement may be purged: "is every instance that references it inside the subgraph?". The check was

```python
len(set(ifc_file.get_inverse(subelement)) - subgraph_set) == 0
```

which wraps every referencing instance into a Python `entity_instance` before comparing. For shared instances that is enormous: a representation context is referenced once per representation in the file, so purging one representation wrapped ~100k instances (9.5M `__hash__` calls over 300 products on a 155 MB model) to conclude "not exclusively mine". The start element is checked the same way against `also_consider`, by traversing each considered element and scanning the result with `entity_instance.__eq__` (990k calls in the same run).

## Change

- `inverse_index::all_sources(id, pred)` visits the live records referencing `id` (base tier and delta) and stops at the first source `pred` rejects.
- `file::all_referencing_instances(instance_id, pred)` dispatches that per storage backend the way `instances_by_reference` does: the in-memory index through `all_sources`, RocksDB through the same `v|<id>|` prefix seek, stopping at the first source `pred` rejects.
- Two thin `%extend` helpers on `file` build the id set and call it: `_all_inverses_within(e, ids)` for one instance, and `_ids_referenced_only_within(ids)`, which returns the subset of `ids` referenced only from within `ids` — one crossing in, one out, early exit per id in C++.
- `remove_deep2` calls `_ids_referenced_only_within` **once per invocation**, before the loop, and the per-subelement test becomes a set lookup. That is exact for the whole loop: the only mutation inside it (clearing large aggregates of instances about to be deleted, #3052) removes references whose *source* is inside the subgraph, and "referenced only from inside" can't change by removing inside references. The start-element check uses `_all_inverses_within` against the `also_consider` ids.

Why once per call and not per candidate: the first version called the predicate per candidate, rebuilding the C++ id set from the Python list each time — O(subgraph²). On models with very large representation subgraphs that was *slower* than the baseline (a 243 MB model 340→400 ms/call, an 85 MB one 432→1220), while winning 25× elsewhere. The once-per-call form has no regression on any of the fourteen models below.

**Where the decisions differ from the old code, and why that is a fix.** On the 155 MB model the two make identical decisions (same surviving ids after 300 `remove_product` calls). On TXG, removing the first 200 products, this PR keeps 39 instances the old code deleted: 13 `IfcPolyline`s and their 26 points. Each of those polylines is referenced by *two* `IfcGeometricCurveSet`s — one inside the purged subgraph through a 14-element `Elements` list, one outside it. The old loop clears aggregates longer than 10 on instances it is about to delete (#3052), which drops the polyline's live inverse count to 1, and its `get_total_inverses(subelement) < 2` shortcut then purges the polyline although its one remaining referencer is outside the subgraph — `process_deletion_` goes on to prune it out of the surviving curve set. The shortcut's premise ("0 or 1 inverses guarantees the subelement only exists in this subgraph") only holds before the loop starts mutating. This PR decides containment against the references as they were before the loop, so an outside reference is never lost sight of, and those 39 instances survive. The `also_consider` check changes in the same direction: it is now "every reference into the start element comes from a considered element", where the old code counted considered elements and compared with the record count, so an element referenced twice by one considered element was previously not purged.

No magic numbers: the decision compares record sources against an id set, not counts against a threshold.

## Verification

- After `root.remove_product` on 300 products of the 155 MB model the set of surviving instance ids is the same (2,720,623 ids, same SHA-1 of the sorted list) on this branch and on `open-perf-v2`, and the same as on `v0.9.0` (243f0f32d) without #9492. On TXG the 39-instance difference described above is the only one, and it is the old code deleting shared geometry.
- `util/test_element.py`, `test_file.py`, `test_file_inverse.py`, `test/api/{root,geometry,pset,feature,material,type,spatial,aggregate}`: 875 passed on this branch and on `open-perf-v2`. **Correction:** that local run had, it turned out, imported a different ifcopenshell build (an auto-loaded `pytest_blender` plugin redirects the import); rerun with plugin autoload disabled and the import verified, the same suites pass on a tree with this PR applied. The stub failure CI reported had two parts, both fixed since: `file::all_referencing_instances` takes a `std::function` and, being a public method of `file`, was being wrapped by SWIG — it is now hidden from SWIG (`#ifndef SWIG`), since only the compiled `%extend` helpers use it; the two `%extend` helpers are underscore-private and need no stub entry (`validate_stub.py` passes). The `./IfcOpenHouse` / `./IfcAdvancedHouse` abort (`double free or corruption (out)`) in the earlier CI run was not this PR: `open-perf-v2` reproduced the identical abort on its own on 2026-09-16 (c64353c07) and was green again on 2026-09-17 (f0e08d8b9); this branch is rebased onto that.

## Benchmark

`root.remove_product` per call over the first 300 `IfcProduct`s (spatial elements last), Python 3.13 source-tree Release builds, single thread, one run each on a quiet 12-core Linux box. Private models supplied for benchmarking, identified by size only.

| model | `open-perf-v2` | this PR |
|---|---|---|
| TXG 61 MB | 52.1 ms | 23.3 ms (-55%) |
| 210_King 155 MB | 51.0 ms | 17.7 ms (-65%) |
| OKgate22 243 MB | 314.6 ms | 323.2 ms (+3%) |
| 10 MB (private) | 9.5 ms | 5.8 ms (-39%) |
| 78 MB (private) | 1285 ms | 62.5 ms (-95%) |
| 85 MB (private) | 377.5 ms | 204.1 ms (-46%) |
| 112 MB (private) | 319.2 ms | 66.9 ms (-79%) |
| 262 MB (private) | 50.3 ms | 27.2 ms (-46%) |
| 292 MB (private) | 3.3 ms | 2.5 ms (-24%) |
| 330 MB (private) | 3299 ms | 171.7 ms (-95%) |
| 366 MB (private) | 4059 ms | 1257 ms (-69%) |
| 548 MB (private) | 5.8 ms | 4.3 ms (-26%) |
| 632 MB (private) | 201.0 ms | 27.7 ms (-86%) |
| 825 MB (private) | 403.8 ms | 95.4 ms (-76%) |

Removing every product of a 51 MB IFC4 model (5,597 products, 1,876 of them already gone as cascaded ports/openings by the time the loop reaches them): 260 s → 30 s, 46.5 → 5.4 ms/call.

The remaining cost of `remove_product` after this change is `file.remove()` itself — of which ~85% is `remove_type_ref`'s linear scan of the per-type instance vector (a separate PR) — and generic SWIG attribute access. OKgate22 and the 366 MB model stay expensive in absolute terms for those reasons, not this check.

RocksDB: built with `WITH_ROCKSDB=ON` (RocksDB 10.4.2). On a 51 MB model converted with `convert_path_to_rocksdb`, `_all_inverses_within` and `_ids_referenced_only_within` return the same answers through the RocksDB branch as through the in-memory index for 400 random instances × three id sets each plus 40 real subgraphs (1,240 checks, 0 mismatches), and `test_validate_rocksdb.py` passes. Note that `root.remove_product` on a RocksDB-backed file segfaults inside `file.remove()` on `open-perf-v2` *without* this PR — the crash is in the deletion itself, after the containment decisions — so end-to-end removal could not be compared there; that is a pre-existing problem this PR neither causes nor fixes.

Not tested: lazy loading (`open(lazy=True)`; the helpers only read the inverse index, which the lazy index builds, but no run was made), Windows/macOS, and Bonsai.


src/ifcwrap/IfcParseWrapper.i
```diff
@@ -316,6 +317,46 @@ private:
...
+		}
+		const std::unordered_set<uint32_t> allowed(ids.begin(), ids.end());
+		const auto referenced_id = (uint32_t)e_.id();
+		return std::visit([referenced_id, &allowed](auto& x) -> bool {
+			if constexpr (std::is_same_v<std::decay_t<decltype(x)>, ifcopenshell::impl::in_memory_file_storage>) {
+				return x.byref_excl_.all_sources(referenced_id, [&allowed](uint32_t source_id) {
+					return allowed.count(source_id) != 0;
+				});
+			} else {
+				throw ifcopenshell::exception("_all_inverses_within is only implemented for in-memory storage");
```
aothms, line 335: This is not an option and also not necessary. There are plenty of variants/adapters around these maps so that compatibility is possible.

Moult: Agreed and fixed in ba070e063: the dispatch now lives in `file::all_referencing_instances(instance_id, pred)`, implemented for both storage backends the same way `instances_by_reference` is (the in-memory index via `inverse_index::all_sources`, RocksDB via the same `v|<id>|` prefix seek, early-exiting on the first rejected source). The two SWIG helpers just build the id set and call it — no `std::visit` or storage internals in the `.i`, no throw. I could only run the in-memory path here (`WITH_ROCKSDB=OFF`); the RocksDB branch mirrors `instances_by_reference` line for line but is untested.

Dion edit: I've told AI to build with rocksdb on and test properly

Moult: Built with `WITH_ROCKSDB=ON` since: the RocksDB branch returns the same answers as the in-memory index on a 51 MB model converted to RocksDB (400 instances × 3 id sets + 40 subgraphs, 0 mismatches), and `test_validate_rocksdb.py` passes. Separately, `file.remove()` on a RocksDB-backed file segfaults on `open-perf-v2` without this PR, so end-to-end removal can't be compared there — pre-existing, noted in the description; I'll investigate it.

Moult: Follow-up on the RocksDB deletion segfault: it was five separate gaps (uint32/size_t decode mismatch in `process_deletion_inverse`, `attribute_value::size()` ignoring the storage model, `create()` / `recalculate_id_counter()` / `byid_.erase()` all being `@todo` stubs). Fixed in #9508; with it, `remove_product` on a RocksDB-backed TXG leaves the same surviving ids and inverse counts as the in-memory file, with and without this PR.

aothms (approved):

PR #9508, opened 2026-09-15
[AI-generated, unverified] ifcparse: make deleting and creating instances work on RocksDB-backed files

Moult:

**Base: `v0.9.0`.** First opened stacked on #9492 (`open-perf-v2`) because that is where it was built and verified; it doesn't depend on that PR's changes, so it was rebased onto `v0.9.0` (the commits cherry-pick cleanly) and retargeted on 2026-09-17. #9502 builds one `v|` key prefix by hand and will switch to `rocksdb_key::inverse_prefix` when it lands after this.

## Problem

`file.remove()` on a RocksDB-backed file segfaulted (found while verifying #9502 with `WITH_ROCKSDB=ON`). Reducing it turned up five separate gaps, each of which makes editing a RocksDB-backed file either crash or silently do nothing:

1. **`rocks_db_file_storage::process_deletion_inverse()`** decoded the `v|` inverse-record values as `size_t`, while the serializer, `register_inverse()`, `unregister_inverse()` and `instances_by_reference()` all write and read them as `uint32_t`. A 4-byte value read as 8 bytes yields no elements, `std::find` returns `end()`, and `vals.erase(end())` is undefined behaviour — the segfault. The same function also computed its `DeleteRange` upper bound from an iterator that is invalid whenever the instance has no inverse records.
2. **`attribute_value::size()`** applied its visitor straight to the in-memory slot array regardless of `storage_model_`, so on RocksDB it hit the variant's "Invalid variant index" fallthrough. Since `set_attribute_value()` calls it for every aggregate, *any* aggregate assignment on a RocksDB instance threw (`rel.RelatedElements = [...]`), not just the rewrite that deletion performs.
3. **`rocks_db_file_storage::create()`** was a stub returning an empty `express::base{}` (the body commented out under a `@todo`), so anything that creates an instance — `update_owner_history` via `util.element.copy()` in `remove_product`, but equally `file.create_entity()` — dereferenced a null handle.
4. **`max_id_`** is only initialised by the in-memory parse paths. A RocksDB-opened file starts at 0, so a working `create()` would have handed out `#1` and `add_type_ref()` would have overwritten an existing instance's type record — silent corruption. `file::recalculate_id_counter()` was itself a commented-out `@todo`.
5. **`byid_.erase(id)`** was a no-op on RocksDB: both `set_to_map_transformer::erase()` and `rocksdb_set_view::erase()` were `@todo` stubs returning 0. After a "successful" removal the instance's attribute keys stayed in the database and its handle stayed in `instance_cache_`, so `entity_names()` still listed it, `by_id()` still returned it, and a fresh open of the database threw `Instance #<id> not found` while enumerating.

## Change

- `process_deletion_inverse()`: decode as `uint32_t`, `std::remove` every occurrence guarded on "found", and derive the range end from the prefix itself (the same prefix with its separator incremented, `'|' + 1`), so no iterator is needed to find it.
- `attribute_value::size()`: branch on `storage_model_` like its sibling accessors; on RocksDB dispatch on `type()` and return the element count of the deserialized aggregate, or -1 for non-aggregates, matching `size_visitor`. The `argument_type` → stored-type correspondence is expressed once as `argument_storage_type_t<A>` (`argument_type` enumerates `type_variant_parameter_pack` in order; `static_assert`s pin it), and `size()` now returns `size_t` — its only caller already took `size_t`.
- `rocks_db_file_storage::create()`: mirrors `in_memory_file_storage::create()` — make the `instance_data` with `rocks_db_attribute_storage`, cache it, `add_type_ref()`. Attributes are written to the database by `set_attribute_value()` and read back on access, so the cache only keeps the handle's identity stable; `assert_existance()` can reload the instance from the type record.
- `file::recalculate_id_counter()`: implemented per backend (in-memory: max over `byid_`; RocksDB: a `SeekForPrev` to the last key under `i|`, see the key format below). The RocksDB `create()` runs it once before its first `fresh_id()`.
- `rocksdb_set_view::erase(key)`: `DeleteRange` over every key under `prefix + key + "|"`, and `set_to_map_transformer::erase()` forwards to it. `file::remove_entity()` and `unbatch()` don't go through that per-instance path, though: they call `file::erase_instances_(ids)`, which on RocksDB is `rocks_db_file_storage::erase_instances()` — one `WriteBatch` of `DeleteRange`s and one lock to drop the cached handles for every instance of the batch (for in-memory storage it erases from `byid_` as before).

**Key format** (review): numeric key segments are now fixed-width 16-digit lowercase hex, produced and parsed by `key_to_string`/`key_from_string` in `rocksdb_map_adapter.h`, with named builders (`rocksdb_key::attribute`, `type_record`, `inverse`, `inverse_prefix`, `type_list`, `upper_bound`) that the storage, `entity_instance_data.cpp` and the serializer all use instead of assembling `"i|" + std::to_string(id) + ...` by hand. Keys therefore sort numerically: an instance's records are laid out in id order and the largest id is the last key under `i|`, which is what `recalculate_id_counter()` seeks to. The two dormant `to_string_fixed_width()` helpers are gone. **This changes the on-disk layout: a database converted before this PR has to be re-converted with `convert_path_to_rocksdb`.**

Editing RocksDB-backed files, including creating instances, is intended for that backend (per the maintainer), so item 3 is a required part of the fix rather than a design choice: without it every API that creates an instance dereferences a null handle.

## Verification

Built with `WITH_ROCKSDB=ON` (RocksDB 10.4.2), Python 3.13 source-tree Release build on top of `open-perf-v2`.

- Minimal repro (a storey, a wall, a point and a containment rel, written to SPF, converted with `convert_path_to_rocksdb`): removing an instance with no references, one with an inverse, and one with outgoing references each went segfault / `Invalid variant index` / segfault before; all three now remove, the instance is gone from `entity_names()` and `by_id()` both in-session and after reopening the database, and the referencing rel's list is pruned.
- Aggregate assignment probe: `rel.RelatedElements = [w1]`, `= []`, `= None` on a RocksDB instance all succeed (the first two threw before).
- Creation probe: `create_entity()` on a RocksDB-opened file returns an instance with a fresh id above the existing maximum, `len()`/`is_a()` work, attribute writes persist across re-read, existing instances are intact, `util.element.copy()` works.
- End to end on TXG (61 MB IFC4): `root.remove_product` on the first 200 products by id, on the file opened from SPF and on the same file converted to RocksDB, leaves the **same 899,312 surviving ids and the same inverse counts on 500 sampled survivors**, both with the current `remove_deep2` and with #9502's. Before this PR the RocksDB run crashed; with only items 1–4 fixed it "succeeded" while leaving 918,000 instances, i.e. deleted nothing.
- `test_streaming_rocksdb_and_simpletyperefs.py` (its `test_rocks` now queries the hex key layout), `test_validate_rocksdb.py`, `test_file.py`, `test_file_inverse.py`, `util/test_element.py`, `test/api/{root,geometry,pset,feature,material,type,spatial,aggregate}` and `util/scripts/test_validate_stub.py`: 920 passed, 1 skipped, 1 failed — the failure is `test_validate_stub`, because the local tree also carries #9502's two new wrapper methods, which are missing from the `.pyi` stub (the same failure #9502's own CI shows; not caused by this PR). Note on method: an earlier revision of this description quoted 913 passed from a pytest run that, it turned out, had imported a different ifcopenshell build (an auto-loaded `pytest_blender` plugin redirects the import); the numbers above were taken with plugin autoload disabled and the import verified. In-memory deletion decisions on the 155 MB model are unchanged (same surviving-id fingerprint).

Not measured: performance of any of this — `size()` on RocksDB deserializes the aggregate (as reading it did already); `recalculate_id_counter()` is one seek. Not tested: Windows/macOS, Bonsai, RocksDB versions other than 10.4.2.


src/ifcparse/entity_instance_data.cpp
```diff
@@ -294,8 +294,38 @@ bool attribute_value::isNull() const
...
+    }
+#ifdef IFOPSH_WITH_ROCKSDB
+    else {
+        // Same answers as size_visitor: element count for aggregates, -1 otherwise.
+        switch (type()) {
+        case Argument_EMPTY_AGGREGATE:
+        case Argument_AGGREGATE_OF_EMPTY_AGGREGATE:
+            return 0;
+        case Argument_AGGREGATE_OF_INT:
+            return (unsigned int)dispatch_get_<std::vector<int64_t>>(array_, storage_model_, instance_name_, entity_or_type_, index_).size();
```
aothms: A bit tangential, but can we create a template look up of 

`Argument_AGGREGATE_OF_INT` <-> `std::vector<int64_t>` - just so that for the untrained eye it looks a bit more coherent.

I also wonder how large the surface area would be if we change unsigned int -> size_t for all size() related calls. I've been beginning to migrate and apparently we're at the boundary here between old and new.

src/ifcparse/file.cpp
```diff
@@ -206,22 +214,21 @@ void ifcopenshell::impl::rocks_db_file_storage::process_deletion_inverse(const e
...
         auto prefix = "v|" + std::to_string(id) + "|";
-        auto it = std::unique_ptr<rocksdb::Iterator>(db->NewIterator(rocksdb::ReadOptions()));
-        it->Seek(prefix);
-        while (it->Valid()) {
-            it->Next();
-            if (!it->key().starts_with(prefix)) {
-                break;
-            }
-        }
+        auto upper_bound = prefix;
```
aothms: It took me a sec because we use `}` also for the placeholder annotation "v|{id}|"

Can we instead do something like 

```
auto prefix = "v|" + std::to_string(id) + "|";
auto upper_bound = "v|" + std::to_string(id) + std::string('|' + 1, 1);
```

(untested)

aothms (approved):
Thanks there were indeed some leftovers, really cool once proper rocksdb-backed storage becomes fully functional.

src/ifcparse/entity_instance_data.cpp
Moult: Done in 78b7285e1: `argument_storage_type_t<A>` in `instance_data.h` maps an `argument_type` member back to its `type_variant_parameter_pack` element (they enumerate the same list in the same order — `static_assert`s pin that), and `size()` goes through one `aggregate_size_<A>()` helper per case. On `size_t`: the surface is tiny — `attribute_value::size()` had exactly one C++ caller (`can_have_significant_inverse_attribute_intersection`, already `size_t`) and no wrapper use — so this PR makes the switch; the non-aggregate sentinel is now `(size_t)-1`.

src/ifcparse/file.cpp
Moult: Agreed, the `}` was doing double duty. Both places (`process_deletion_inverse` and `rocksdb_set_view::erase`) now build the upper bound as prefix + `std::string(1, '|' + 1)`, and the comments no longer use the `{id}` notation.

src/ifcparse/parse.cpp
```diff
@@ -3294,16 +3294,32 @@ template void ifcopenshell::impl::in_memory_file_storage::read_from_stream(file_
...
+            for (const auto& p : x.byid_) {
+                k = std::max(k, (unsigned int)p.first);
+            }
         }
-    }
-    max_id_ = (unsigned int)k;
-    */
+#ifdef IFOPSH_WITH_ROCKSDB
+        else if constexpr (std::is_same_v<std::decay_t<decltype(x)>, impl::rocks_db_file_storage>) {
+            // Keys sort as text, so the largest id can't be found by seeking;
```
aothms: Better address this now. I'm fine with e.g a fixed-width hex encoding of a size_t (better than binary because of human readability).

This might help also with locality of the data (caching, block reads, etc.) on typical export idioms.

src/ifcparse/file.cpp
```diff
@@ -154,7 +154,15 @@ ifcopenshell::impl::rocks_db_file_storage::rocks_db_file_storage(const std::stri
...
     , file(ffile)
     , instance_ids_(db.get(), "i|")
-    , instance_by_name_(&instance_ids_, [this](size_t v) { return assert_existance(v, entityinstance_ref); })
+    , instance_by_name_(
+        &instance_ids_,
+        [this](size_t v) { return assert_existance(v, entityinstance_ref); },
+        [this](size_t v) {
+            // The instance's keys are gone from the database; drop the
+            // cached handle so lookups don't keep resolving it.
+            std::lock_guard<std::mutex> lock(instance_cache_mutex_);
```
aothms: This looks like a performance nightmare, individual calls into a std::function with a mutex. Shouldn't we batch this higher up in the call site rather than this weird coupling into the map type with a function?

src/ifcparse/parse.cpp
Moult: Done in 221ef964e: numeric key segments are now fixed-width 16-digit lowercase hex via `key_to_string`/`key_from_string` in `rocksdb_map_adapter.h`, with named builders under `rocksdb_key::` used by the storage, `entity_instance_data.cpp`, `read_schema()` and the serializer (the dormant `to_string_fixed_width` helpers are gone). Keys sort by id, so `recalculate_id_counter()` is one `SeekForPrev` to the last key under `i|`. It's an on-disk format change — databases converted before this commit need re-converting; the PR description says so. #9502 builds one `v|` prefix by hand and will need `rocksdb_key::inverse_prefix` once this lands.

src/ifcparse/file.cpp
Moult: Agreed. The transformer hook is gone; `remove_entity()` and `unbatch()` now call `file::erase_instances_(ids)`, which on RocksDB is `rocks_db_file_storage::erase_instances()` — one `WriteBatch` of `DeleteRange`s and one lock over the whole batch. The set view keeps a plain `erase()` for the adapter's sake, but the file doesn't go through it per instance.

aothms (approved):

PR #9518, opened 2026-09-17
Build OpenCASCADE shared on every platform, and test the packages for it

Moult:

Release packages statically link a private copy of OpenCASCADE into every geometry plug-in. Each copy has its own `Standard_Type` registry, typeinfo and allocator, so a `TopoDS_Shape` created by the kernel plug-in is misread by the next plug-in that touches it: `geom.tree().select()` returns nothing or raises `RuntimeError: An unknown error occurred`, the SVG serializer raises or aborts with `Standard_NoSuchObject`, and Bonsai drawings come out blank (#9341, #9398, the static-OCCT cluster in `ci-bonsai-daily`). Geometry conversion alone stays bit-identical, and CI builds against one shared system OCCT, so every workflow was green while every package on builds.ifcopenshell.org was broken.

Supersedes #9341. The parts of that PR that overlapped with @Andrej730's merged `--occt-shared` plumbing are dropped; what remains, plus the pieces needed to make it hold on every platform, is here.

## The fix: OCCT is shared everywhere

- **Linux/macOS** — `nix/build-all.py` now builds OCCT shared by default (`--occt-static` opts out; wasm implies it since it has no shared libraries). `package-zip-archives.py` stages the shared OCCT whenever one was built. No workflow ever passed `--occt-shared`, which is why the plumbing never reached a package.
- **Windows** — OCCT is built with `BUILD_LIBRARY_TYPE=Shared`; the unconditional `add_definitions(-DHAVE_NO_DLL)` in `win/patches/V7_8_1.patch` is scoped to static builds, since it empties `Standard_EXPORT` and a shared OCCT would otherwise export nothing. The install dir is renamed `opencascade-shared-7.8.1` so the `build-outputs` cache cannot hand back the static build under the same name (mirrors the `-shared-` infix `build-all.py` uses for the same reason). The TK DLLs are installed next to the plug-ins via `$<TARGET_RUNTIME_DLLS>`; `build-all-win.py` then picks them up through its existing `dumpbin` trace, so the archiver needs no change.
- **macOS packaging** — with `CREATE_BUNDLE` no plug-in or core dylib gets an `LC_RPATH`, so IfcConvert's plug-ins could not find a shared OCCT. `package-zip-archives.py` now adds `@loader_path` to every staged library and `@executable_path` to the flat executables, re-signs them (arm64 refuses unsigned images), and gains an `otool -L` dependency check to match the Linux `ldd` one. `BonsaiViewer.app` also stages the OCCT dylibs into `Frameworks/`.
- **Headless machines** — `OpenCASCADE_LIBRARIES` from OCCT's config is the complete toolkit list, Visualization included. Static, that is free; shared, every entry becomes a hard load dependency and `TKV3d` drags in libGL/libEGL. `FindOpenCASCADE.cmake` now records only referenced toolkits (`--as-needed` on GNU ld, `-dead_strip_dylibs` on Apple ld). Measured on the kernel plug-in: 58 → 14 `libTK` entries, libGL/libEGL gone.

## Why it can never ship silently again

- **`test/package/`** — pytest module that exercises exactly the cross-plug-in hand-offs (kernel → brep tree via `add_element` and via `tree(file)`, kernel → SVG serializer mirroring Bonsai's drawing generator), on a fixture from the main tree. `run_against_package.py` extracts a built `ifcopenshell-python-*.zip` (preserving the SONAME symlinks `zipfile` would otherwise flatten) and runs the tests with the packaged wrapper on `PYTHONPATH`. Every build workflow runs it against the zip it just produced, before uploading. Verified locally against real packages from S3:

  | package | result |
  |---|---|
  | `15deb1c` linux64 (static OCCT, v0.9.0 branch, 2026-08-26) | 3 failed, all `RuntimeError: An unknown error occurred` |
  | `06c3b4f` linux64 (shared OCCT) | 3 passed |
  | local source build against system OCCT | 3 passed |

## Things to know

- The first CI run per platform pays a full OCCT rebuild (~30-60 min) and adds `cache-occt-shared-7.8.1` / `cache-opencascade-shared-7.8.1` to the `build-outputs` branches; the static caches stay behind unused and can be deleted there.
- Windows and macOS could only be verified structurally from Linux (link flags, DLL/dylib layout, `otool`/`dumpbin` behaviour). The package test on each runner is the verification; expect a round or two on those workflows.
- `src/ifcopenshell-python/Makefile` still pins `BUILD_COMMIT` for `ci-bonsai-daily`; bumping it after the first green package build clears the static-OCCT test cluster there.


Moult:

Linux x64 is green end to end on this branch: shared OCCT built and cached, packages built, the new package test passed all three cases against the packaged 3.13 wrapper, and the zips uploaded (`ifcopenshell-python-*-v0.9.0alpha0-25c1170-linux64.zip`). https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35190953443

Linux ARM, macOS and Windows are in progress. macOS and Windows run on a test branch that merges #9519, since v0.9.0 head cannot currently build on either for unrelated reasons.

Moult:

Linux ARM64 is green too, package test 3/3 on the packaged wrapper: https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35190955661

Moult:

macOS (arm64) is green on the test branch that merges #9519: shared OCCT built, dylibs staged with `@loader_path` rpaths and re-signed, the `otool -L` check found nothing unresolved, and the package test passed 3/3 against the packaged 3.13 wrapper. https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35195603605 (a first attempt failed on a transient GitHub 500 while fetching the wgpu-native release asset; unrelated.)

Windows is the last platform still running.

cmake/FindOpenCASCADE.cmake
```diff
@@ -59,6 +59,29 @@ if(NOT OCC_INCLUDE_DIR AND NOT OCC_LIBRARY_DIR)
...
+        # toolkits whose symbols are actually referenced.
+        #
+        # On GNU ld this is deliberately not closed with --no-as-needed: CMake
+        # emits the imported targets' INTERFACE_LINK_LIBRARIES (where OCCT lists
+        # libGL/libEGL) after this item, so closing the bracket would switch
+        # recording back on for exactly the libraries this exists to exclude.
+        # The flag is a no-op for static archives, so it is safe for both link
+        # types; wasm is excluded only because wasm-ld has no shared OCCT to
+        # trim and need not see the GNU ld spelling at all.
+        if(APPLE)
```
aothms, line 80:
```suggestion

        # Windows / MSVC doesn't have this option and behaves like --as-needed by default.
        if(APPLE)
```

aothms (approved):

cmake/FindOpenCASCADE.cmake
Moult: _Written by Claude (AI) on behalf of @Moult._

Applied, thanks.

Moult:

Windows is green on both architectures (x64 and ARM64): OCCT built shared under `opencascade-shared-7.8.1`, the TK DLLs shipped next to the plug-ins, and the package test passed 3/3 against the packaged 3.12 wrapper on each. https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35198313147

That completes the verification on every platform this branch touches:

| platform | run | package test |
|---|---|---|
| Linux x64 | [35190953443](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35190953443) | 3/3 |
| Linux ARM64 | [35190955661](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35190955661) | 3/3 |
| macOS arm64 | [35195603605](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35195603605) | 3/3 |
| Windows x64 + ARM64 | [35198313147](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35198313147) | 3/3 each |
| wasm | [35187503124](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35187503124) | n/a (single module), builds |

The Windows and macOS runs used a test branch merging #9519, which is now in v0.9.0; this branch is rebased onto it. The packages from these runs are on S3 under commits `25c1170` (Linux), `9018228` (macOS) and `ef015c9` (Windows). Nothing changed in the source since apart from Thomas's comment nit on the cmake finder.

PR #9519, opened 2026-09-17
Fix v0.9.0 head build regressions on wasm, Windows and macOS

Moult:

Three regressions on v0.9.0 head that no build workflow had run into, found while verifying #9518 on every platform. Each is reproduced on plain v0.9.0 and is independent of #9518.

1. **wasm** — ae2f27c3a replaced the global `include_directories(${Boost_INCLUDE_DIRS})` with a `Boost::headers` target appended to `Boost_LIBRARIES`. The wasm branch of IfcParse's link line never included `Boost_LIBRARIES`, so it lost the include path: `src/ifcparse/exception.h:25:10: fatal error: 'boost/lexical_cast.hpp' file not found`. Reproduced: https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35187209175 (no Pyodide build had run since 2026-08-28).
2. **Windows** — 6dbc4e28d raised the minimum CGAL to 5.4. The Windows dependency cache is keyed on the install directory name and only ever adds to it, so the unversioned `cgal` directory kept serving CGAL 5.2.3: `Could not find a configuration file for package "CGAL" that is compatible ... version: 5.2.3`. The directory is now `cgal-5.5.5` (the TODO in the installer) and recorded in the deps cache like the other installers.
3. **macOS** — Homebrew no longer publishes bottles for `macos-14`, so `brew install qt` compiles Qt and QtWebEngine from source and fails in the dependency step. Runner bumped to `macos-15`.


Moult:
Ping @Andrej730

Moult:

Pyodide build is green on this branch: https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35187503124 (the same workflow fails on plain v0.9.0, run 35187209175). Windows and macOS runs with these fixes are in progress on a test branch merged with #9518.

aothms (approved):

PR #9520, opened 2026-09-17
[AI-generated, unverified] ifcparse: stop scanning a type's instance list when deleting an instance

Moult:

**Stacked on #9492** (base branch `open-perf-v2`, one commit). Independent of #9502 and #9508.

**Revision note:** the first revision of this PR used a dead-count + compact-on-read scheme ("lazy" below). After review it was replaced by the sorted-vector alternative suggested there; the three-way measurements that led to that are kept below.

## Problem

After #9460/#9467 made the inverse index cheap to update, `file.remove()` still cost ~0.19 ms per call on a 155 MB model, and a `py-spy --native` profile put **72% of that in `remove_type_ref()`**: it ran `std::remove` over the entire per-type instance vector, comparing `express::base` handles — a `data()` dereference each — so removing one `IfcPropertySet` scanned all 73k of them. O(instances of the type) per deletion, i.e. quadratic for bulk deletion of a common type; the same shape the inverse index had.

## Change

Each type's `std::vector<express::base>` is kept **sorted by id**. Nothing else about the lists changes: no counters, no tombstones, readers copy them as before.

- Loading fills the lists in file order; `sort_type_lists()` establishes the order once, after `byref_excl_.sort()`, on both the parse and the lazy-index paths. A list already in id order (the common case) is only checked. Otherwise the ids are read once and `(id, position)` pairs are sorted — no instance data is touched per comparison — with the lists sorted in parallel, largest first. A file that isn't in id order costs a few tens of milliseconds to open (the 155 MB merged model: 0.81 → 0.82 s).
- `add_type_ref()` appends a fresh id (they only grow) or inserts an explicit one at its sorted position.
- `remove_type_ref()` finds the instance by binary search — equal ids sit together, so it picks the one that is this handle — and erases it: a memmove of handles, no dereferences.
- `by_type()` now returns a type's instances in id order rather than file order. There was no documented guarantee of the previous order.
- #9492's per-worker population paths are untouched (they append, the sort comes after). The RocksDB storage keeps its own `remove_type_ref()`.

## Measurements

Per call, Python 3.13 source-tree Release builds, single thread, one run each on a quiet 12-core Linux box. `base` = `open-perf-v2` (f0e08d8b9); `lazy` = this PR's first revision; `sorted` = this PR. `file.remove` over the first 300 `IfcPropertySet`, `batch()`/`unbatch()` over the next 2000, `root.remove_product` over the first 300 `IfcProduct` (spatial elements last), each from a fresh open; `open()` is the median of three.

| model | scenario | base | lazy (#9520) | sorted |
|---|---|---|---|---|
| TXG 61 MB | `file.remove` ×300 psets, ms/call | 0.189 | **0.012** | **0.018** |
| TXG 61 MB | batched ×2000, ms/call | 0.177 | **0.011** | **0.017** |
| TXG 61 MB | `remove_product` ×300, ms/call | 49.0 | **29.9** | **30.9** |
| TXG 61 MB | `open()`, s | 0.34 | 0.33 | 0.34 |
| 210_King 155 MB | `file.remove` ×300 psets, ms/call | 0.167 | **0.006** | **0.013** |
| 210_King 155 MB | batched ×2000, ms/call | 0.168 | **0.009** | **0.011** |
| 210_King 155 MB | `remove_product` ×300, ms/call | 48.6 | **38.9** | **39.8** |
| 210_King 155 MB | `open()`, s | 0.81 | 0.78 | 0.82 |
| OKgate22 243 MB | `file.remove` ×300 psets, ms/call | 0.492 | **0.021** | **0.038** |
| OKgate22 243 MB | batched ×2000, ms/call | 0.488 | **0.020** | **0.036** |
| OKgate22 243 MB | `remove_product` ×300, ms/call | 306.7 | **7.3** | **17.9** |
| OKgate22 243 MB | `open()`, s | 1.26 | 1.24 | 1.28 |

Interleaved: one `file.remove` then one `by_type` of the same type, 300 times; ms per iteration (a single `by_type` read of that list alone is shown for scale).

| model | type (list size) | read alone | base | lazy (#9520) | sorted |
|---|---|---|---|---|---|
| 210_King 155 MB | IfcCartesianPoint (504,190) | 207.0 ms | 303.0 | 314.3 (+4%) | 301.9 (-0%) |
| 210_King 155 MB | IfcPropertySet (73,804) | 32.3 ms | 32.2 | 35.1 (+9%) | 32.7 (+2%) |
| OKgate22 243 MB | IfcPropertySet (216,302) | 84.4 ms | 107.2 | 119.3 (+11%) | 109.1 (+2%) |
| TXG 61 MB | IfcPropertySet (77,269) | 34.0 ms | 34.1 | 36.6 (+7%) | 34.3 (+1%) |

Reading the tables: both schemes remove the scan (`file.remove` 13–28× faster than base); lazy is ~2× ahead of sorted on bulk removal because its removal is O(1) where sorted's `erase` memmoves the tail of the list — most visible on OKgate22's `remove_product`, whose representation-item lists run into the millions. But on the interleaved case lazy is 4–11% *slower* than base (its compaction does a `byid_` lookup per element, more than the old scan's pointer compare), while sorted sits at base; and sorted carries no hidden invariant for another call site to forget. That is the trade the review asked for and the sorted scheme is what this PR now carries.

The first revision's table over all fourteen models (three public, eleven private, identified by size only) is in this PR's edit history; the sorted scheme was measured on the three public models above.

## Verification

- Catch2 (`test_ifcopenshell_parse.cpp`), all 19 cases: removals from the front, middle and back keep the rest sorted and a read sees none of them; an explicitly reused id is inserted in place and names only the new instance; batched removal of everything left of a type removes it from `types_begin()..types_end()` while another type stays; a file written in descending id order lists each type in ascending id order after loading.
- After `root.remove_product` on the first 300 products of TXG, `by_type()` counts for every class and `types()` agree exactly with what `entity_names()` says survives.
- Deletion decisions unchanged: after 300 `remove_product` on the 155 MB model with targets chosen by id, the surviving-id fingerprint is identical to `open-perf-v2`'s. (Choosing targets by `by_type()` position gives a different set on this branch, because that order changed.)
- Python (`PYTEST_DISABLE_PLUGIN_AUTOLOAD=1`, import verified): `test_streaming_rocksdb_and_simpletyperefs.py`, `test_validate_rocksdb.py`, `util/test_element.py`, `test/api/{root,geometry,pset,feature,material,type,spatial,aggregate}`, `test_file.py`, `test_file_inverse.py`, `test_file_gc.py`: 923 passed, 25 skipped.

Not tested: Windows/macOS, Bonsai, downstream code that relied on `by_type()` returning file order.


aothms:
It's not a bad design, but as always I'm a bit weary of introducing types with hidden assumptions. It's easy to at another callsite forget about the thombstone.

> binary search by id (the lists are in parse order, not id order

There is no documented requirement on the order of the by_id list and in fact by order by id is maybe even a feature. Can you investigate this alternative? It does come of course with an impact on maintaining the order with add() and it's not O(1) removal either (unless we write back nullptr - maybe also an option - not sure when to compact).

> Interleaving one removal with one read of that type costs the same as before; nothing gets worse.

I'm not sure here. I think this needs a benchmark of interleaving removal + by_type read. The per-element cost looks to be higher. This is where a sorted vector might be the proper compromise.

Moult:
Investigated the sorted-vector alternative and measured base vs the dead-count scheme vs sorted (tables in the description). You were right about interleaving: the dead-count compaction is 4–11% slower than base on alternating remove + `by_type` (a `byid_` lookup per element beats the old pointer compare), while the sorted vector sits at base. Sorted gives up ~2× to dead-count on bulk removal (`erase` memmoves the tail) but is still 13–28× ahead of base, and there's no invariant for another call site to forget — so the PR now carries the sorted vector (65722bac4). The one cost I hit was `open()` on a file that isn't in id order (+0.5 s on the 155 MB merged model with a dereferencing comparator); sorting `(id, position)` pairs, lists in parallel, brought that to +0.01 s. `by_type()` now returns id order, which was undocumented before. The RocksDB `remove_type_ref` (linear over a value blob) is untouched.

aothms:
I don't think we can do this:

```C++
const size_t threads = std::min<size_t>(lists.size(), std::max(1u, std::thread::hardware_concurrency()));
```

I don't think it's good CPU custodianship somewhere deep in a library to just hog all threads without the user having any influence on it. People can also be opening multiple files in parallel.

For now let's take the 0.5 sec penalty and then we look at holistically bringing in multi-threading configuration for 0.9.1. @Moult thoughts?

aothms (approved):

PR #9527, opened 2026-09-17
ci: test Bonsai on pull requests, publish-only nightly

Moult:

Nothing tested `src/bonsai` on pull requests: `ci.yml` runs the IfcOpenShell Python suites but cannot run Blender, and `ci-bonsai-daily` only tested the nightly package after publishing it, on Linux, against whatever the latest release happened to be, a day after the change.

**`ci-bonsai-test.yml`** (new) builds the extension from the checkout with `make dist` (IfcOpenShell binaries come from the pinned `BUILD_COMMIT`, so no C++ build), installs it into a fresh Blender 5.2, smoke-tests ifcsverchok next to it, and runs `test-core`, `test-tool` and `test-bim` as separate steps so the job shows which suite broke. Triggers on PRs touching `src/bonsai/**` and the packages the extension bundles, plus manual dispatch; superseded runs on the same PR are cancelled.

**`ci-bonsai-daily.yml`** keeps only what it is for: build the zips, publish the release, regenerate the extension index. Runs queue rather than overlap and are never cancelled, since a run cancelled mid-publish leaves a release with zips missing. `activate` is renamed `set-timestamp`.

Both check out submodules: the wheel is built from the checkout's `ifcopenshell/` directory and `simple_spf` and `mvd` are submodules under it, so the nightly's ifcopenshell wheel shipped without either package while the PyPI wheels carry both. Both use host Python 3.13, the same as Blender's.

Supersedes #9499, which installed the previous night's unstable zip rather than the PR's code and ran `test-tool` only. First run of the new job on this PR matched last night's daily exactly (389 / 2+858 / 2+1438, same four known failures).


Moult:

First run of the workflow on this PR: https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35283501520

| suite | result |
|---|---|
| test-core | 389 passed |
| test-tool | 2 failed, 858 passed |
| test-bim | 2 failed, 1438 passed |

Identical to last night's ci-bonsai-daily on the same commit (run 35276254325), down to the same four failures: `TestGettingLinkedElementGeomSlice::test_skip_hidden_{first,middle}_element`, `test_create_a_mep_transition`, and the ifcjson export test (`wrapped_data`, fixed upstream in Andrej730/IFC2JSON_python#1). So the job reproduces the daily's verdict on a PR, without the build matrix or the publish steps.

PR #9534, opened 2026-09-18
nix: package every shared library once and only the OCCT toolkits in use

Moult:

## Why

The Bonsai Linux zip went from 144 MB (v0.8) to 234 MB (v0.9). 68 MB of that is the `ifcopenshell` wheel (43 → 111 MB), and **55 MB of the wheel is the OCCT toolkits stored three times**:

```
libTKBool.so        5.1 MB
libTKBool.so.7.8    5.1 MB   <- the only name anything loads
libTKBool.so.7.8.1  5.1 MB
```

The build zip on S3 is fine — `zip -y` keeps these as symlinks (65 MB). But a wheel cannot hold symlinks, so by the time `src/ifcopenshell-python/Makefile` has turned the zip into a wheel, every link is a full copy. macOS has the identical layout (`libTKBool.dylib -> .7.8.dylib -> .7.8.1.dylib`, 96 symlinks).

## What

Everything is in `nix/package-zip-archives.py`. There are two independent changes; the first is worth 55 MB, the second 2.4 MB.

### 1. Stage each library once, under its SONAME

The old code kept symlinks **on purpose**. This predicate is not a bug:

```python
if not (runtime_file.is_symlink() or runtime_file.is_file()):
    continue
```

It skips only what is neither a symlink nor a file. Symlinks were then copied as symlinks (`follow_symlinks=False`), and `ensure_soname_links()` even created the SONAME link when it was missing. That is a correct design for a zip; it only goes wrong once the zip becomes a wheel.

Flipping the predicate alone would break the package: only the real file `libTKBool.so.7.8.1` would be left, while every binary asks the loader for `libTKBool.so.7.8` — which was one of the symlinks. (In practice `ensure_soname_links()` would re-create that link and the wheel would hold two copies instead of three.) So the fix is two lines working together:

```python
if runtime_file.is_symlink() or not runtime_file.is_file():          # skip the symlinks ...
    continue
...
dest_file = dest / (get_soname(runtime_file) or runtime_file.name)   # ... and name the real file by its SONAME
```

The SONAME (install name on macOS) is the only name used at run time: every `DT_NEEDED` / `LC_LOAD_DYLIB` entry refers to `libTKx.so.7.8` / `libTKx.7.8.dylib`. The unversioned `libTKx.so` only serves the linker at build time, and nothing refers to `.so.7.8.1`. IfcOpenShell's own libraries and plug-ins have SONAME == file name (or none, for the Python extension module, where the file name is kept), so they are unaffected.

What follows from that:

- **`get_soname()`** is the `readelf` lookup lifted out of `ensure_soname_links()` so staging can use it, plus a macOS branch (`otool -D`). The old code needed no macOS lookup because macOS relied entirely on the copied symlinks.
- **`ensure_soname_links()`** now calls `get_soname()` and otherwise behaves as before. It stays because the Qt staging for the viewer executables still uses it, and that package never becomes a wheel.
- **`shutil.copy()` loses `follow_symlinks=False`**, and the rpath loop loses its `is_file()` / `is_symlink()` guards: nothing copied is a symlink any more. The Linux and macOS rpath branches merge into one loop.

No change is needed in the Makefile that builds the wheel.

### 2. Drop the dependency libraries nothing links to

- **`stage_runtime_payload()` now returns the paths it accounts for.** `copied_files` is what this call wrote and gets its rpath patched, as before. `staged_files`, the return value, also includes files an earlier call already placed — the existing comment explains that the same library can be installed in two places — and pruning needs the complete list.
- **`get_needed_libraries()`** reads a binary's needed-library entries (`readelf -d` on Linux, `otool -L` on macOS).
- **`prune_unused_libraries()`** starts from every binary in the package that did not come from a dependency directory (the wrapper, `libifcopenshell.*`, the plug-ins, the executable), follows the needed-library entries through the dependency libraries, and deletes what is never reached.
- **The two call sites** collect the returned lists for the dependency directories only. IfcOpenShell's own libraries are never candidates: plug-ins are loaded by file name at run time and nothing links to them, so pruning them would remove them wrongly.

For OCCT that is 21 of the 48 toolkits (`TKBin*`, `TKXml*`, `TKStd*`, `TKTObj`, `TKDEGLTF/OBJ/PLY/STL/VRML`, `TKFeat`, `TKMeshVS`, `TKRWMesh`, `TKXMesh`). The Windows package already ships exactly the same 27 toolkits.

This part carries the only real risk in the PR: a library that is loaded by name at run time but never linked would be dropped. The STEP/IGES test below and the existing `ldd` check are there to catch that. It is cleanly separable and can be split into its own PR if preferred.

## Measured

Re-packaging the current `8003d43` Linux build through the modified script:

| | before | after |
|---|---|---|
| native files in the python package | 206 (96 symlinks) | 89 (0 symlinks) |
| OCCT toolkits | 48 × 3 names | 27 |
| wheel payload | 112 MB | **55 MB** |
| IfcConvert zip | 50.9 MB | 48.5 MB |

`ldd` finds no missing dependency in either package, the re-packaged `IfcConvert` converts to `.stp`, and all of `test/package` passes against the re-packaged python zip.

## Tests

New `test/package/test_package_contents.py` (runs in the four build workflows through `run_against_package.py`):

- no symlinked libraries, and no two libraries with identical contents — both fail against today's package, pass after;
- the STEP and IGES serializers write a file. They link to more OCCT toolkits than anything else (data exchange → application framework → visualization), so they are the first to notice a toolkit that was wrongly left out.

## Verified in CI

The build workflows are `workflow_dispatch` only, so they were dispatched on this branch at `c1c1ca5`. Both are green: [macOS](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35342223582) (the first real run of the `otool -D` / `otool -L` branches) and [Linux](https://github.com/IfcOpenShell/IfcOpenShell/actions/runs/35342226449). All 7 `test/package` tests pass on each, including the four new ones.

Comparing the python 3.13 zips those runs uploaded against the `8003d43` ones:

| | symlinks | OCCT library names | wheel payload |
|---|---|---|---|
| Linux before | 96 | 144 | 112.2 MB |
| Linux after | 0 | 27 | **55.0 MB** |
| macOS arm64 before | 96 | 144 | 92.4 MB |
| macOS arm64 after | 0 | 27 | **47.0 MB** |

("Wheel payload" = every file deflated, symlinks expanded to full copies, which is what the wheel build does to them.) The zips themselves shrink only by the pruned toolkits, since they already held symlinks: python 61.9 → 59.3 MB on Linux, 52.2 → 50.2 MB on macOS; IfcConvert 48.5 → 46.3 MB and 37.4 → 35.6 MB.

The Linux dependency check still reports missing libraries for the `BonsaiViewer` and `IfcViewerMinimal` packages. All of them are system X11/GTK/CUPS/SQL-client libraries wanted by Qt plug-ins (`libxcb-cursor.so.0`, `libgtk-3.so.0`, `libcups.so.2`, `libpq.so.5`, ...); none is an OCCT or IfcOpenShell library, and the python and IfcConvert packages report nothing.

Not run: Linux arm and Windows. Windows doesn't use this script, but its build workflow does run `test/package`, so the new tests will run there too. Reading the current `8003d43` Windows zip: no duplicate DLLs, 27 toolkits, and both the STEP and IGES plug-ins are present, so the contents tests pass; the STEP/IGES serializer test has not actually been run on Windows.


aothms (approved):

aothms:
For me it's ok. The alternative would be to rewrite the dt_needed tables (pathelf) and remove the symlinks altogether. Or maybe if you tweak the cmake version detection (for every dependency library then... that's painful) you can force it not to link to a symlink at the source.

PR #9536, opened 2026-09-18
fixes ifcclash add_collision_objects

RickBrice:
iterator.get() refuses a second call to get() before iterator.next()


aothms:
Correct. This is new. get() returns a unique ptr now moving the data out of
the iterator
https://github.com/IfcOpenShell/IfcOpenShell/blob/c9118cfd93fdb375fa8977b001b23715172de7ed/src/ifcgeom/iterator.h#L300

Op za 19 sep 2026, 08:14 schreef Dion Moult ***@***.***>:

Issue #9364, opened 2026-08-26
v0.9.0 alpha wheels ship no shared libraries: Bonsai fails to load on all platforms


BIMvoice:
### Bug Description

The `bonsai-0.9.0-alpha260825` builds fail to load on **all three platforms**. The Python wheel ships `_ifcopenshell_wrapper` but none of the shared libraries it links against, so importing `ifcopenshell` raises immediately and Blender reports `FATAL ERROR: Unable to load Bonsai`.

Anyone on the unstable channel who updates today gets a Bonsai that will not start.

Reproduced on macOS. Confirmed by inspecting the published archives that Linux and Windows have the same defect.

### Error on macOS

```
ImportError: dlopen(.../ifcopenshell/_ifcopenshell_wrapper.cpython-313-darwin.so, 0x0002):
  Library not loaded: @rpath/ifcopenshell_document_rdb.dylib
  Reason: tried: '.../site-packages/ifcopenshell/ifcopenshell_document_rdb.dylib' (no such file),
          '/Applications/Blender.app/Contents/Resources/lib/ifcopenshell_document_rdb.dylib' (no such file)

ImportError: IfcOpenShell not built for 'darwin/64bit/python3.13'

FATAL ERROR: Unable to load Bonsai
```

### It is not one missing library, it is all of them

`otool -L` on the installed wrapper shows six `@rpath` dependencies, and **none** are present in the wheel:

```
@rpath/ifcopenshell_document_rdb.dylib
@rpath/libifcopenshell.geometry.writer.dylib
@rpath/ifcopenshell_geometry_kernel_opencascade.dylib
@rpath/libifcopenshell.geometry.dylib
@rpath/libifcopenshell.parse.dylib
@rpath/libifcopenshell.plugin.dylib
```

### All three platforms are affected

I downloaded the three published `py313` archives from the release and extracted the nested `ifcopenshell` wheel from each. Every one contains **exactly one** native file, the wrapper itself, and no libraries:

| Platform | Native files in wheel |
|---|---|
| `macos-arm64` | 1, `_ifcopenshell_wrapper.cpython-313-darwin.so` |
| `linux-x64` | 1, `_ifcopenshell_wrapper.cpython-313-x86_64-linux-gnu.so` |
| `windows-x64` | 1, `_ifcopenshell_wrapper.cp313-win_amd64.pyd` |

And the same dependencies are recorded in the Linux and Windows binaries:

```
[...]

PR #9544, opened 2026-09-19
cmake: use underscores in the core shared library names (#9328)

### LINKED ISSUE #9328: v0.9.0 library names
### Bug Description

The plugin library files have been renamed with underscores instead of dots as expected. And the build time is down from 110 minutes to 25 minutes  (aarch64) and from 120 minutes to 45 minutes (x86_64) - this is comparing the ifcviewer-wgpu branch from two weeks ago with the v0.9.0 branch from today, so I don't know what has changed.  

..but there are still issues with the system library files:

``` bash
      31 /usr/lib64/libifcopenshell.geometry.so -> libifcopenshell.geometry.so.0.8 
      33 /usr/lib64/libifcopenshell.geometry.so.0.8 -> libifcopenshell.geometry.so.0.8.6
  778272 /usr/lib64/libifcopenshell.geometry.so.0.8.6
   66584 /usr/lib64/libifcopenshell.geometry.writer.so
      28 /usr/lib64/libifcopenshell.parse.so -> libifcopenshell.parse.so.0.8
      30 /usr/lib64/libifcopenshell.parse.so.0.8 -> libifcopenshell.parse.so.0.8.6
  986128 /usr/lib64/libifcopenshell.parse.so.0.8.6
      29 /usr/lib64/libifcopenshell.plugin.so -> libifcopenshell.plugin.so.0.8
      31 /usr/lib64/libifcopenshell.plugin.so.0.8 -> libifcopenshell.plugin.so.0.8.6
  111496 /usr/lib64/libifcopenshell.plugin.so.0.8.6
```

1. These library files still use dots instead of underscores
2. The symlinks are setup correctly for geometry, parse and plugin, but geometry.writer is created as a single unversioned .so file
3. The SONAME number is 0.8.6, but I'm expecting 0.9.0

brunopostle:
Fixes #9328. As discussed there, the four core shared libraries kept dotted names while everything loaded by name is `ifcopenshell_*`.

| before | after |
| --- | --- |
| `ifcopenshell.parse` | `ifcopenshell_parse` |
| `ifcopenshell.geometry` | `ifcopenshell_geometry` |
| `ifcopenshell.geometry.writer` | `ifcopenshell_geometry_writer` |
| `ifcopenshell.plugin` | `ifcopenshell_plugin` |

## More than four `OUTPUT_NAME`s

Several places match these names by prefix or pattern, so a rename changes their behaviour without any literal old name appearing. That is how the packaging gaps behind `No schema named IFC4` (#9301, #9364, #9423) happened, so I went through every consumer:

- **Plugin discovery (`src/plugin/plugin.cpp`).** `ifcopenshell_geometry_writer` starts with `ifcopenshell_geometry_`, which is what the generic geometry serializer scan matches. Where libraries have no `lib` prefix (Windows) the core library would be picked up and loaded as a plugin. It only survives today because the dot in `ifcopenshell.geometry.writer` keeps it out of the match. `discover()` now skips the four core libraries.
- **`win/build-all-win.py`.** The writer filter ended in a separator, so it would stop recognising the core writer, which would then ship next to the executables. It now matches the core library and the per-schema plugins.
- **`ifcviewer-minimal`.** The macOS bundle `EXCLUDE REGEX` was keyed on the dotted name. Kept so it excludes the same file as before.
- `nix/package-zip-archives.py` literal, and comments in the Makefile and viewer CMake files.

I found no other references to the old names in the repo (CMake, workflows, conda, pyodide, docs, tests, the Python loader), including after rebasing onto the current `v0.9.0`.

## Interaction with open PRs

- #9427 edits the same line in `nix/package-zip-archives.py`, and its `IFC_GEOMETRY_WRITER_PREFIXES` ends in `_`, which won't match the renamed core writer (`ifcopenshell_geometry_writer.dll` / `libifcopenshell_geometry_writer.so`). The prefix needs to be `ifcopenshell_geometry_writer`.
- #9442 checks for `ifcopenshell.parse` / `ifcopenshell.geometry`. After this change that check would match nothing and could pass silently, so it needs the new names.

Whichever lands first, the other needs a small update. Happy to do it.

## Testing

- Built the real `plugin.cpp` before and after, and called `discover("geometry_")` on a directory of Windows-style (unprefixed) filenames: before, the core `ifcopenshell_geometry_writer` is returned as a candidate; after, it isn't, and all real plugins (kernels, `geometry_writer_ifc4`, `svgfill`, serializers) still are.
- Ran the real filter functions from `win/build-all-win.py` on both name sets. With the old script and new names the core writer ships next to the executables and isn't treated as a writer; with this change the result is identical to the pre-rename baseline.
- Both scripts byte-compile.

Not tested: a complete build and install of the renamed libraries on Windows, macOS or Linux. A Fedora RPM build is in progress. It would be worth confirming on each platform that a clean install imports, e.g. `ifcopenshell.file(schema="IFC4")`. The nix writer exclusion still matches nothing because of the `lib` prefix; #9427 fixes that separately and I left it alone.


src/plugin/plugin.cpp
```diff
@@ -134,6 +134,20 @@ namespace {
 		return {decorated_basename(basename)};
 	}
 
+	// The core shared libraries share the ifcopenshell_ prefix with the plugins and are
+	// installed next to them, but they are linked rather than loaded by name and export no
+	// plugin entry points. Where libraries have no platform prefix (Windows) a prefix scan
+	// such as "geometry_" would otherwise match ifcopenshell_geometry_writer and try to
+	// load it as a plugin. Keep in sync with the OUTPUT_NAME of the core library targets.
+	bool is_core_library(const std::string& basename) {
```
aothms, line 142: did we consider a check on currently loaded libraries instead? `GetModuleHandleW` / `dl_iterate_phdr` - maybe this is not the best time to start thinking about this, but ultimately maybe harcoded strings is not the most robust way forward.

aothms (approved):

brunopostle:
I'd be happy to merge this sooner rather than later, but it has the risk of breaking the wheels and bonsai packages again, which I can't test from here

PR #9549, opened 2026-09-20
cmake: read the nlohmann_json version from multi-header and 3.11+ installations

Moult:
_Written by Claude (AI) on behalf of @Moult. Please review before merging: this touches the CMake dependency detection, and @Moult would like @aothms or @Andrej730 to check it._

## Symptom

The daily conda builds (`publish-ifcopenshell-conda.yml`) have failed on all three platforms since 2026-09-17, after eleven green days, with:

```
CMake Error at .../FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find nlohmann_json: Found unsuitable version "..", but required
  is at least "3.6.1" (found $PREFIX/include)
```

## Cause

a5dec54719 ("cmake: add minimum required json version", 2026-09-16) made `Findnlohmann_json.cmake` parse the version macros out of `nlohmann/json.hpp`. That only works for single-header installations older than 3.11:

1. **Multi-header installations keep the macros elsewhere.** conda-forge builds `nlohmann_json` with `JSON_MultipleHeaders=ON` ([feedstock build.sh](https://github.com/conda-forge/nlohmann_json-feedstock/blob/main/recipe/build.sh)), and since 3.11 that layout defines `NLOHMANN_JSON_VERSION_*` in `nlohmann/detail/abi_macros.hpp`, not in `json.hpp`. The parse yields `".."`.
2. **Since 3.11 the macro lines end in a `// NOLINT(modernize-macro-to-enum)` comment**, so the `$`-anchored regex doesn't match even on the single header. The version then comes out as the whole `#define` line. Ubuntu 22.04 ships 3.10.5, which is why `test-ifcopenshell` stayed green; Ubuntu 24.04 ships 3.11.3 and would fail the same way.

The finder tries `find_package(... CONFIG)` first, which would have worked, but `conda/build.sh`, `conda/build.bat` and the pixi presets in `cmake/CMakePresets.json` all pass `JSON_INCLUDE_DIR`, which skips config mode and forces the header parse. So conda and pixi hit the same bug.

## Fix

- Read the macros from `nlohmann/detail/abi_macros.hpp` when it exists, else from `nlohmann/json.hpp`.
- Allow trailing text after the number (`([0-9]+).*$`).
- The three copy-pasted `file(STRINGS)`/`string(REGEX REPLACE)` blocks become one `foreach` over `MAJOR MINOR PATCH`.

An alternative would be to drop `JSON_INCLUDE_DIR` from the conda/pixi configs so config mode is used. That would fix conda and pixi but leave the finder broken for any 3.11+ install found by path, so the finder fix seemed the right place.

Also bumps the stale `version` in `pixi.toml` (was `0.8.4`, the lock doesn't embed it).

## Verification

Isolated CMake harness (`include(utilities)` + `find_package(nlohmann_json 3.6.1 REQUIRED)` with `JSON_INCLUDE_DIR` set) against the real headers from the nlohmann/json releases:

| Layout | Old finder | New finder |
|---|---|---|
| 3.12.0 multi-header (conda-forge) | `".."` → error | 3.12.0 |
| 3.10.5 multi-header (pre-`abi_macros.hpp`) | 3.10.5 | 3.10.5 |
| 3.11.3 single header (Ubuntu 24.04) | whole `#define` line → error | 3.11.3 |
| 3.6.1 single header (Windows CI) | 3.6.1 | 3.6.1 |

Not run: an actual conda or pixi build. The conda workflow has `workflow_dispatch`, so it can be triggered right after merging instead of waiting for the nightly.

https://claude.ai/code/session_01Gpj1MBesJMgsZnupzFqjtE


cmake/Findnlohmann_json.cmake
```diff
@@ -21,13 +21,17 @@ find_path(json_header_path "nlohmann/json.hpp" HINTS "${JSON_INCLUDE_DIR}")
...
-    set(nlohmann_json_VERSION "${json_version_major}.${json_version_minor}.${json_version_patch}")
+    # Multiple headers installations (e.g. conda-forge) keep the version macros in a separate header since 3.11.
+    set(json_version_header "${json_header_path}/nlohmann/detail/abi_macros.hpp")
+    if(NOT EXISTS "${json_version_header}")
+        set(json_version_header "${json_header_path}/nlohmann/json.hpp")
+    endif()
+    foreach(component MAJOR MINOR PATCH)
+        file(STRINGS "${json_version_header}" json_version_line REGEX "^#define NLOHMANN_JSON_VERSION_${component} ")
+        # Since 3.11 the macros are followed by a NOLINT comment.
+        string(REGEX REPLACE "^#define NLOHMANN_JSON_VERSION_${component} ([0-9]+).*$" "\\1" json_version_${component} "${json_version_line}")
```
aothms:
```suggestion
        string(REGEX REPLACE "^#define NLOHMANN_JSON_VERSION_${component} ([0-9]+)(\s*//.*)?$" "\\1" json_version_${component} "${json_version_line}")
```

I would prefer something like above (untested) so that we're not silently dropping something unexpected

aothms (approved):

cmake/Findnlohmann_json.cmake
Moult: _Written by Claude (AI) on behalf of @Moult._

Agreed, applied in the amended commit with one change: CMake's regex syntax has no `\s`, so the suggestion as written fails at parse time with `Invalid character escape '\s'`. The pushed version uses `[ \t]*` instead:

```cmake
"^#define NLOHMANN_JSON_VERSION_${component} ([0-9]+)([ \t]*//.*)?$"
```

Re-ran the harness:

| Layout | Result |
|---|---|
| 3.12.0 multi-header (`   // NOLINT(...)`) | 3.12.0 |
| 3.11.3 single header (`   // NOLINT(...)`) | 3.11.3 |
| 3.6.1 single header (no comment) | 3.6.1 |
| `#define ... 9 /* something else */` | `Could NOT find nlohmann_json: Found unsuitable version "#define NLOHMANN_JSON_VERSION_MAJOR 3 /* ...` |

So an unexpected trailer now surfaces as a visible configure error rather than being dropped.

PR #9553, opened 2026-09-20
Reformat Python with ruff format

Moult:

The mechanical follow-up to #9552, based on that branch. **Merge #9552 first**; GitHub will then retarget this PR to `v0.9.0`, after which it may need a rebase before merging.

## The commit

**Reformat Python with ruff format** — 181 files, +944/−862, produced by

    git ls-files -z '*.py' | xargs -0 ruff format --force-exclude

with ruff 0.16 and no manual edits. Anyone can reproduce it on the #9552 branch. The changes are the known black/ruff deviations: trailing pragma comments (`# ty:ignore`, `# noqa`) no longer count towards the line length, implicit string concatenations are joined when they fit, and multi-item `with` statements are parenthesised.

## `.git-blame-ignore-revs` comes afterwards, on purpose

The reformat commit's SHA changes with every rebase and with a rebase-merge, so an entry inside this PR can never name the commit that ends up on `v0.9.0`. Once this is merged, a one-line follow-up adds the final SHA to `.git-blame-ignore-revs`, and `git config blame.ignoreRevsFile .git-blame-ignore-revs` makes `git blame` skip it.

Worth timing the merge for when the active branches are quiet: any open PR touching one of the 181 files gets formatting conflicts on rebase.

## Verification

- After the reformat, `ruff format --check` on tracked files reports 1,808 files already formatted.
- CI's syntax check (`compileall -W error`) passes on the tree, except for gitignored `external_dependencies/` clones, which are untouched and absent in CI.

https://claude.ai/code/session_01Gpj1MBesJMgsZnupzFqjtE


Moult:

Folded into #9552 as its second commit. Splitting the tooling switch from the reformat cannot work with a blocking lint: #9552's `ruff format --check` fails on the black-formatted tree until this reformat lands, and this could not land first.

Issue #9468, opened 2026-09-09
Bonsai: removing a drawing from a sheet silently leaves it in the layout when data-id no longer matches the reference's STEP id


theoryshaw:
### Summary

`bim.remove_drawing_from_sheet` removes the `IfcDocumentReference` from the model but leaves the drawing in the sheet's layout SVG, whenever the layout's `data-id` no longer matches the reference's current STEP id. It fails silently — no error, and the layout file is still rewritten, so it looks like something happened.

The result is a model and a layout that disagree. Bonsai's own Sheets UI shows the drawing gone (it reads the IFC), while the layout SVG still places it — and so does anything built from that layout.

### Cause

`SheetBuilder.remove_drawing` matches on the STEP entity id:

https://github.com/IfcOpenShell/IfcOpenShell/blob/e1be433207bee6f539b8f295dca561b60b00379a/src/bonsai/bonsai/bim/module/drawing/sheeter.py#L224-L229

```python
for g in layout_root.findall(f"{SVG}g"):
    if g.attrib.get("data-id") == str(reference.id()):
        layout_root.remove(g)
        break

layout_tree.write(layout_path)
```

STEP ids are not stable across a re-serialisation of the file — merging projects, or any round trip through a tool that renumbers entities, changes them. The layout keeps whatever ids were current when each drawing was added, so after a renumbering `data-id` points at nothing. The loop matches nothing, `break` is never reached, and `layout_tree.write()` runs regardless: the file's mtime changes and it gets reformatted, while its content is unchanged.

The same `<g>` element already carries a stable key, written two functions above:

```python
view.attrib["data-id"] = str(reference.id())
view.attrib["data-drawing"] = drawing.GlobalId
```

And `update_drawing_sizes` in the same file already prefers it:

```python
drawing = ifc_file.by_guid(drawing_view.attrib.get("data-drawing"))
```

So `remove_drawing` is the outlier — it is the one place keyed on the unstable half of the pair.

### What I observed

On a project that had been through a merge:

- Sheet had 9 drawings placed in its layout. I removed 5 of them via the Sheets UI.
- The model afterwards referenced 4 — correct, and Bonsai's UI showed 4.
- The layout SVG still contained all 9, with its mtime updated to the moment of the removal.
- Every `data-id` in that layout (`3672697`, `3729889` … `3729896`) matched no entity in the file; the live `IfcDocumentReference`s were `#1373231` … `#1380359`.

[...]

PR #9555, opened 2026-09-20
Bonsai: a web API for tools that display and edit sheets without building them

### LINKED ISSUE #9554: Bonsai: a web API for tools that display and edit sheets without building them
### The gap

Bonsai renders a sheet by compositing its layout into one SVG (`SheetBuilder.build`). That is the right artefact for issuing a drawing set, but a poor one for a tool that *displays* sheets continuously — a review or redlining tool, say:

- the built sheet is a rasterised composite, often megabytes, regenerated wholesale on every change;
- nothing in it says which drawing a mark sits over, so an annotation cannot be anchored to the model.

The layout is the better source. It is a few KB of structure and links, it carries each drawing's GlobalId in `data-drawing`, and another tool can render it directly. What it cannot do by itself is fill in the templates — `{{Name}}`, `{{Identification}}`, `{{Scale}}`, and a titleblock's fields all come from the model, and a tool reading the saved `.ifc` sees neither unsaved edits nor a model that has never been saved.

### What such a tool needs from Bonsai

**1. Read the values a build would fill in, without building.** From the model in memory, so unsaved edits are included, and from the same methods `build()` uses so the two cannot drift.

**2. Write them back through Bonsai, not into the IFC.** Blender holds the model in memory: a write to the file is invisible to it and lost on its next save. Several of these fields are also more than an attribute — renaming a sheet moves its layout and its built sheet, renaming a drawing moves its SVG and relinks every layout that places it. Only Bonsai does that. Which fields can be written, and why not when they cannot (a scale comes from the drawing's camera; revisions come from the repository's tags), is likewise Bonsai's to answer — a tool deciding for itself would drift from what Bonsai accepts.

**3. A connection that is there when asked.** The web server exists for pages a person opens; a tool reading continuously needs it up without a browser having been opened first, and needs it to survive a project load.

### Proposed

Three requests on the existing `sourcePage: "sheets"` web operator:

- `getTemplateValues` — every sheet's titleblock data and each placed drawing's or document's view-title data. Sheets keyed by their layout's path, placements by the file they place, drawings also by GlobalId — never by the layout's `data-id`, which is a STEP id and does not survive a re-serialised model (#9468).
- `getEditableFields` — for the placeholders a template actually uses, what each holds and whether Bonsai can write it, with a reason when it cannot.
- `setTemplateValue` — applies values through Bonsai's own operations (`rename_sheet`, `update_drawing_name`, `rename_reference`, `document.edit_information`), and answers with the layout's path afterwards, since renaming a sheet moves it.

Along with the reliability work that turned out to be needed for any of it to be usable: connection state read from the client rather than the scene (loading a project calls `wm.read_homefile`, which replaced it), a persistent request timer that survives a file load, no second client or second server on reconnect, one failing request no longer unregistering the timer, dead entries pruned from `running_pid.json`, and an opt-in *Keep Web Connection* preference so a tool does not depend on someone having opened a web page.

### Context

This is implemented and in daily use in [SketchSpace](https://github.com/OpeningDesign/SketchSpace), a self-hosted multi-page whiteboard for redlining drawing sets. It renders Bonsai layouts live, writes drawing positions back into the layout, and now lets view-title and titleblock values be edited from the sheet — every edit going to Bonsai rather than to the file.

Nothing proposed here is specific to that tool: it is a read/write API for sheet template values, useful to anything that shows a drawing set without building it.

A PR follows.

theoryshaw:
Closes #9554.

A read/write API over the existing web socket for tools that display a drawing set from its layouts rather than from built sheets, and let its view-title and titleblock values be edited from there. Three commits: the reading half and the connection work it needs, then the writing half, then each sheet's site and building.

### Reading — `getTemplateValues`

Answers with every sheet's titleblock data and each placed drawing's or document's view-title data, from the model in memory so unsaved edits are included.

The data comes from the same methods `build()` uses, which are extracted for the purpose — `get_titleblock_data`, `get_drawing_view_title_data`, `get_document_view_title_data` — so what a tool renders and what a build renders cannot drift. That extraction is the only change to how sheets are built, and it is behaviour-preserving.

Sheets are keyed by their layout's path, placements by the file they place, drawings also by GlobalId. Not by the layout's `data-id`: that is a STEP id and does not survive a re-serialised model (#9468). Values are text as pystache renders them, except the booleans and row lists that templates test and iterate (`as_template_data`).

### Writing — `getEditableFields` and `setTemplateValue`

An edit has to come through Bonsai rather than be written into the IFC. Blender holds the model in memory, so a write to the file is invisible to it and lost on its next save. Several of these fields are also more than an attribute: renaming a sheet moves its layout and its built sheet, renaming a drawing moves its SVG and relinks every layout that places it. So each goes through the operation Bonsai itself uses — `core.rename_sheet`, `core.update_drawing_name`, `core.rename_reference`, or `document.edit_information`.

- `getEditableFields` takes the placeholders a template actually uses and answers, for each, what it holds and whether it can be written — with a reason when it cannot, which the caller shows as given. Deciding that here, beside the operations that would apply it, is what keeps it in step with what `setTemplateValue` accepts.
- `setTemplateValue` applies values all-or-nothing, and answers with the layout's path afterwards: renaming a sheet moves it, so the path the caller named is already gone and only Bonsai knows the new one.
- A view is named by its drawing's GlobalId or by the file it places — the two handles that survive a re-serialised model.

Failures are reported to the caller rather than only printed: a remote tool has no other way to learn why nothing happened. Blender is also repainted after an applied edit, since a web operator runs on a timer rather than an event and its panels would otherwise keep drawing the names they last drew (`tool.Blender.redraw_all_areas`).

### A sheet's site and building

IFC gives a project no address of its own, and a model can hold several sites and buildings, so a titleblock that shows an address has to mean "this sheet's". Titleblocks can now use `{{SiteName}}`, `{{SiteDescription}}`, `{{SiteAddress}}` (the whole address on one line) and its parts (`{{SiteAddressLines}}`, `{{SitePostalBox}}`, `{{SiteTown}}`, `{{SiteRegion}}`, `{{SitePostalCode}}`, `{{SiteCountry}}`), and the same set for `{{Building…}}`.

- **A sheet names its site and building** with an `IfcRelAssociatesDocument` from the sheet to each: ordinary IFC that survives a re-serialised model and is removed with the sheet by `document.remove_information`. The Sheets panel gains **Site** and **Building** dropdowns for the active sheet (`bim.link_sheet_site`, `bim.link_sheet_building`); linking one replaces any other.
- **Unlinked, only a single answer is used:** a linked building's own site, the model's only top-level site, or the site's only building. With several to choose from, the fields are blank rather than guessed. `IfcRelAggregates.RelatedObjects` is a set, so "the first" would be whatever order the file happens to hold. A sheet linked to two sites (the generic Documents panel can do that) counts as linked to none for the same reason, and its dropdown says so.
- **Editable through `setTemplateValue`:** the names, descriptions and address parts, and the links themselves, which `getEditableFields` answers as a field with `options`. An element without an address gets one; clearing every part removes it, since an empty `IfcPostalAddress` breaks WR1. Renaming a site or building renames its object and refreshes the spatial tree, as the attribute panel does. Unset values render blank rather than `None`: "None, None" in an address reads as a fault.

**Web edits are now undoable.** `setTemplateValue` applied its edit outside any operator, so Ctrl+Z skipped it and undid whatever came before, and Bonsai's panels kept drawing their cached data. It now checks the values first (`check_template_values`) and applies them as an internal IFC operator (`bim.edit_sheet_template_values`). A refusal comes back as its reason and leaves no undo step; an applied edit is one undo step, like any made in Blender.

### Connection work

None of the above is usable without this, and each was found by a tool reading continuously rather than a person opening a page:

- **A project load silently stopped every answer.** `bim.load_project` calls `wm.read_homefile`, which replaces the scene — and the connected flag lived on the scene while the request timer was registered non-persistently. Connection state is now judged by the client, the timer is persistent, and the props are restored on `load_post`.
- **Reconnecting made a second client** beside the dead one, or a second server on a fresh port. `ConnectToWebsocketServer` now returns early when already connected.
- **One failing request unregistered the timer,** silently ending every later one. The loop no longer dies on an exception.
- **`running_pid.json` only ever lost an entry to an explicit kill,** so quitting Blender left one behind and it gained a dead server per session. Dead ports are dropped when a server starts, and a clean exit removes its own.
- **A new `open_browser` option** on the connect operator, and an opt-in **Keep Web Connection** preference (off by default) that connects in the background without opening a page, so a tool does not depend on someone having opened one first.

### Testing

`src/bonsai/test/tool/test_sheeter.py` — 54 tests. The first 24 cover `as_template_data`'s conversions, which view a request names, which fields may be written and why not, and that each field goes to the operation that owns it (including that setting a sheet's name alone does not blank its identification, and that a refused field leaves nothing written). The other 30 cover the site and building: resolving an unlinked sheet, leaving the fields blank when there is no single answer, conflicting links, linking and relinking, editing names and addresses, and refusing before writing.

The suite passes on this branch with an IfcOpenShell 0.9 build (Blender 5.1). The third commit was developed on a 0.8 build, where in a Blender session the dropdowns, linking, and Ctrl+Z undoing an edit sent over the web were also checked; this branch's copy differs from that one only where the branches differ (it has no revisions table).

Used daily against real drawing sets by [SketchSpace](https://github.com/OpeningDesign/SketchSpace), which is what these requests were written for, though nothing in them is specific to it.


theoryshaw:
Worth pairing with #9557 for anyone trying this out, though the two are independent and touch no common files.

`setTemplateValue` renames through Bonsai, which moves the layout and built sheet at once while the `.ifc` only catches up when the model is saved. A rename made from a browser is therefore made by someone who cannot save it, and reopening the model then leaves it naming files that are not there — `bim.open_layout` raises, `create_sheets` skips the sheet. #9557 puts those files back.

This PR is usable without it; the failure is just unpleasant to recover from by hand.

PR #9557, opened 2026-09-20
Bonsai: put back sheet files renamed in a session that was not saved

### LINKED ISSUE #9556: Bonsai: renaming a sheet or drawing without saving leaves the model pointing at files that no longer exist
Renaming a sheet or a drawing moves files on disk immediately, but the new names reach the IFC only when the model is saved. Close without saving, reopen, and the model names files that are no longer there.

### To reproduce

1. Open a project with a sheet that has a drawing on it.
2. Rename the sheet (or the drawing) in the Drawings panel. Its layout and built sheet are moved on disk at once; a renamed drawing's SVG is moved too, and every layout placing it is relinked.
3. Close Blender **without saving the IFC**.
4. Reopen the model and run **Open Layout** on that sheet.

**Expected:** the layout opens, or Bonsai says what is wrong.

**Actual:** `FileNotFoundError` out of the operator (paths shortened):

```
  File ".../bonsai/bim/ifc.py", line 569, in execute_ifc_operator
  File ".../bonsai/bim/module/drawing/operator.py", line 3056, in _execute
    sheet_builder.update_sheet_drawing_sizes(sheet)
  File ".../bonsai/bim/module/drawing/sheeter.py", line 211, in update_sheet_drawing_sizes
    layout_tree = ET.parse(layout_path)
  ...
  File ".../python/Lib/xml/etree/ElementTree.py", line 562, in parse
    source = open(source, "rb")
FileNotFoundError: [Errno 2] No such file or directory:
  '...\\layouts\\A01 - PLANS SECTIONS AND ELEVTIONS.svg'
```

**Create Sheets** reaches the same state more quietly: `validate_sheet_files` finds the layout missing and the sheet is skipped, so a sheet that looks fine in the UI is simply not built.

Seen on Blender 5.2.1, IFC4, on a `v0.8.0`-era build; the code involved is unchanged on `v0.9.0`.

### Why it happens

The rename is applied in two places that do not commit together:

- **On disk, immediately** — `core.rename_sheet` moves the layout and the built sheet; `core.update_drawing_name` moves the drawing's SVG and rewrites the link in every layout that places it.
- **In the IFC, only on save** — the `IfcDocumentReference` locations are changed in memory.

So an unsaved session leaves the files under their new names and the saved model under the old ones. Nothing is lost — the work is all still on disk — but the model cannot find any of it, and nothing says which file is which.

It is easy to hit, because renaming sheets is a tidying-up activity and tidying up is exactly when someone closes without saving.

### Suggested direction

The model in front of you is the authority on what its sheets are called, and Bonsai already names a sheet's files after it. So the files can be put back under the names the open model uses, and the person told that it happened — rename again and save to keep a new name.

The matching has to be careful, since it moves files. A candidate is an `.svg` in the same folder that no sheet refers to and that demonstrably belongs to this sheet: it places exactly that sheet's drawings, or carries the sheet's number and shares a drawing with it. Only an unambiguous match should be moved, and nothing another sheet or drawing still expects. A drawing's moved file can be found through the layout, which still ties each drawing's GlobalId to the file it was moved to.

Failing all that, `bim.open_layout` should report a missing layout rather than raise.

A PR follows.

theoryshaw:
Closes #9556.

Renaming a sheet moves its layout and built sheet; renaming a drawing moves its SVG and relinks it in every layout placing it. All of that happens on disk at once, while the new names reach the IFC only when it is saved. Reopen the model without having saved and it points at files that are not there: `bim.open_layout` died with `FileNotFoundError`, and `create_sheets` skipped the sheet without saying why.

Those files are now put back under the names the open model uses. The model in front of you decides what its sheets are called, and Bonsai names a sheet's files after it — so the rename that was never saved is the thing to undo, not the model. The operator reports what was moved: *rename again, and save, to keep a new name.*

### How a moved file is recognised

It moves files, so it only moves one it is sure about.

**A sheet's file** (its layout, or its built sheet) is an `.svg` in the same folder that **no sheet refers to** and that belongs to this sheet — it places exactly the sheet's drawings, or it carries the sheet's number and shares at least one drawing with it. Two sheets with no drawings at all match on number alone. Only an unambiguous match is moved; if two candidates fit, none is.

**A drawing's file** is found through the layout, which still ties each drawing's GlobalId to the file it was moved to, so the layout itself says where the SVG went. Every layout placing that drawing is relinked, not only the first, and a drawing's raster underlay moves with it.

Every sheet is checked, whichever one the operator was about: one unsaved session can have renamed several, and the model names them all.

`bim.open_layout` also reports a missing layout as an error now instead of raising, so the case this does not cover is still a message rather than a traceback.

### Scope

`OpenLayout`, `OpenSheet` and `CreateSheets` call it before they do anything else, since those are the three that need the files to be where the model says.

Nothing is deleted and nothing belonging to another sheet or drawing is taken. A file that cannot be matched is left alone.


theoryshaw:
One more reason this matters, which was not obvious when I wrote it.

#9555 lets a tool edit a sheet's or drawing's name through Bonsai. That turns this from a bug you hit by being careless into one you hit by using the feature: the rename moves the files at once, the `.ifc` is only correct once someone saves, and **the person renaming is not the person who can save**. They are in a browser; Blender is on someone else's screen, or the same screen an hour later.

So the two-phase commit that causes this stops being a slip and becomes the normal shape of a remote edit. Without something like this, the recovery is to work out which file is which and rename them back by hand — from a traceback that only names the file that is missing.

The two PRs are independent and can merge in either order; this one does not touch anything #9555 touches.

